PR #2, opened 2023-08-14
Improve renderHTML and TexPlugin

mscolnick:
* This improves the TexPlugin to grab the already rendered HTML from the mounted element instead of re-rendering inside the shadow dom.
* this also changes RenderHTML from a react component to function, which simplifies the WebComponent created in `registerReactComponent`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ fe-check: fe-lint fe-typecheck
 .PHONY: fe-test
 # test frontend
 fe-test:
-	cd frontend; pnpm test
+	cd frontend; CI=true pnpm test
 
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -33,7 +33,7 @@ fe-check: fe-lint fe-typecheck
 .PHONY: fe-test
 # test frontend
 fe-test:
-	cd frontend; pnpm test
+	cd frontend; CI=true pnpm test
```
akshayka, line 36: Nice, thanks.

akshayka (approved):
Looks great, love the simplification.

PR #4, opened 2023-08-15
Fix favicon / manifest.json

mscolnick:


pyproject.toml
```diff
@@ -43,7 +43,6 @@ dev = [
     "mypy~=1.4.1",
     "ruff~=0.0.275",
     # For docs
-    "pypandoc_binary",
```
mscolnick, line 46: hitting some errors with  `pip install -e ".[dev]"` otherwise. having it out doesn't block development. we wil need to add somewhere to manually install this or an alternative

akshayka: Can you post the error?

If the error is from just the binary, we should at least replace this with `pypandoc` (which installs the library without the pandoc binary).

mscolnick:
recheck

akshayka (approved):

PR #5, opened 2023-08-15
Prevent autosave of 0 and wait until some code is loaded

mscolnick:


frontend/src/core/saving/useAutoSave.ts
```diff
@@ -32,5 +31,5 @@ export function useAutoSave(opts: {
         clearTimeout(autosaveTimeoutId.current);
       }
     };
-  }, [codes, config.save, connStatus.state, onSave, needsSave]);
```
akshayka, line 35: I think taking out `codes` breaks autosave. When I type rapidly for a while then stop, autosave doesn't end up saving the final output (even though needsSave is true) (autosave delay 1s).

If I modify your change to include codes, it works as intended.

This is the behavior when `codes` is included as a dependency. I modeled it after VS Code's autosave behavior.
- `needsSave` is false
- user types a character `needsSave` -> true, new timeout is created
- user keeps typing rapidly ... so `codes` changes and old timeouts are cancelled and new ones are queued
- finally user stops typing, autosave delay expires and save request is actually sent

Without `codes` as a dep:
- `needsSave` is false
- user types a character `needsSave` -> true, new timeout is created
- user keeps typing rapidly for a long time ... timeout expires and onSave is called. But `needsSave` is still true so this hook doesn't get triggered again ...? Not sure.

akshayka (commented):
Nice catch on the race condition! But the change to the autosave hook breaks autosave I believe. See inline comment

frontend/src/core/saving/useAutoSave.ts
mscolnick: I'll revert this part, I thought it wasn't being used, but your explanation makes sense. I can add a comment

akshayka (approved):

PR #7, opened 2023-08-16
Wide mode

mscolnick:
This change adds a private API for per-app configuration:

Configuration is specified via keyword arguments to the App object. This is a private API (for now). Users can configure view the settings menu in the marimo frontend.
- Edit-mode only handler for saving app config added.
- To start we have just one setting, the width of the app (normal or full). For example,

```python
import marimo

```


akshayka (approved):
LGTM!

PR #8, opened 2023-08-16
App config

akshayka:
This change adds a private API for per-app configuration:

- Configuration is specified via keyword arguments to the App object. This is a private API (for now).
- Edit-mode only handler for saving app config added.

To start we have just one setting, the width of the app (normal or full). For example,

```python
import marimo

```

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -19,7 +19,8 @@
     </script>
     <marimo-filename hidden>{{ filename }}</marimo-filename>
     <marimo-mode data-mode={{ mode }} hidden></marimo-mode>
-    <marimo-config data-config='{{ config }}' hidden></marimo-config>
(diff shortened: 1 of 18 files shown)


frontend/index.html
```diff
@@ -19,7 +19,8 @@
     </script>
     <marimo-filename hidden>{{ filename }}</marimo-filename>
     <marimo-mode data-mode={{ mode }} hidden></marimo-mode>
-    <marimo-config data-config='{{ config }}' hidden></marimo-config>
+    <marimo-user-config data-config='{{ user_config }}' hidden></marimo-user-config>
```
akshayka, line 22: Saw no reason not to split the configs into two tags.

mscolnick (approved):

PR #11, opened 2023-08-19
mo.stop; interrupt => state invalidation

akshayka:
- Add a function `stop(predicate, output)` that stops execution of a cell when `predicate` is `False`

When `predicate` is `True`, this function stops execution of the current cell and makes `output` its output. Any descendants of this cell that were previously scheduled to run will not be run, and their defs will be removed from program memory.

Example:

```python
mo.stop(form.value is None, mo.md("**Submit the form to continue.**"))
```

- Invalidate descendant state on `MarimoInterrupt` (manual stop), for consistency with how other exceptions are handled

- Refactor runtime

### DIFF
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -1,5 +1,14 @@
 # Misc
 
+## Control flow
+Use `mo.stop` to halt execution of a cell, and optionally output an object.
+This function is useful for validating user input.
+
+```{eval-rst}
+.. autofunction:: marimo.stop
+```
+
+## Debugging
 Use `mo.defs` and `mo.refs` to get the global definitions and references of
 a cell.
 
--- a/docs/api/misc.md
+++ b/docs/api/misc.md
@@ -8,6 +8,10 @@ This function is useful for validating user input.
 .. autofunction:: marimo.stop
 ```
 
+```{eval-rst}
+.. autoclass:: marimo.MarimoStopError
+```
+
 ## Debugging
 Use `mo.defs` and `mo.refs` to get the global definitions and references of
 a cell.
--- a/frontend/src/core/kernel/messages.tsx
+++ b/frontend/src/core/kernel/messages.tsx
@@ -5,10 +5,16 @@ import { CellId } from "../model/ids";
 export type MarimoError =
   | { type: "syntax"; msg?: string }
   | { type: "interruption"; msg?: string }
-  | { type: "exception"; msg: string; raising_cell?: number }
-  | { type: "cycle"; edges: Array<[number, number]> }
-  | { type: "multiple-defs"; name: string; cells: number[] }
-  | { type: "delete-nonlocal"; name: string; cells: number[] }
+  | {
+      type: "exception";
+      exception_type: string;
+      msg: string;
+      raising_cell?: CellId;
+    }
+  | { type: "ancestor-stopped"; msg: string; raising_cell: CellId }
+  | { type: "cycle"; edges: Array<[CellId, CellId]> }
+  | { type: "multiple-defs"; name: string; cells: CellId[] }
+  | { type: "delete-nonlocal"; name: string; cells: CellId[] }
   | { type: "unknown"; msg?: string };
 
 export type OutputMessage =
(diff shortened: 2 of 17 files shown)


mscolnick:
here is a decent smoke_test (if you want to include in that folder). the UI renders ok in present mode, but the cells throw errors

```python
import marimo

__generated_with = "0.1.1"
app = marimo.App()


@app.cell
def __():
    import marimo as mo
    import time
    return mo, time


@app.cell
def __(mo):
    secret = mo.ui.text(label="Type a valid password: ")
    secret
    return secret,


@app.cell
def __(mo, secret):
    # Validation 1
    # This cell just depends on the secret
    mo.stop(len(secret.value) < 8, mo.md("Must have length 8").callout(kind="warn"))

    success_1 = True
    return success_1,


@app.cell
def __(mo, secret):
    # Validation 2
    # This cell just depends on the secret
    mo.stop("$" not in secret.value, mo.md("Must contain a **$**").callout(kind="warn"))

    success_2 = True
    return success_2,


@app.cell
def __(mo, secret, success_1):
    # Validation 3
    # This cell depends on the scrent and first validation
    mo.stop("7" not in secret.value and success_1, mo.md("Must contain the number 7").callout(kind="warn"))

    success_3 = True
    return success_3,


@app.cell
def __(mo, success_1, success_2, success_3):
    # This depends on all the validations, and not the secret
    _success = success_1 and success_2 and success_3
    mo.stop(not _success)

    mo.md("Secret is correct!").callout(kind="success")
    return


if __name__ == "__main__":
    app.run()
```

akshayka:
Thanks for the smoke test.

1. I'll remove the traceback that's printed when `mo.stop()` stops a cell.
2. Descendants of a stopped cell should somehow communicate to the user (in edit mode) that the cell didn't run because one of its ancestors was stopped. Otherwise I can see things getting confusing. Perhaps the exception message is not quite right, and instead we need a message that says something like "This cell didn't run because one of its ancestors was stopped with `mo.stop`" (that shows only in edit mode).

akshayka:
> Thanks for the smoke test.
> 
> 1. I'll remove the traceback that's printed when `mo.stop()` stops a cell.
> 2. Descendants of a stopped cell should somehow communicate to the user (in edit mode) that the cell didn't run because one of its ancestors was stopped. Otherwise I can see things getting confusing. Perhaps the exception message is not quite right, and instead we need a message that says something like "This cell didn't run because one of its ancestors was stopped with `mo.stop`" (that shows only in edit mode).
> 
> EDIT: Actually, I think we should keep the traceback, to communicate to developers (ie in edit mode) that a cell was stopped. Imagine if library code had a `mo.stop` (likely not good practice) and there was no indication that it had been triggered. That would be difficult to debug.

image

mscolnick:
I think the communication is good, but it looks like other Exceptions while this is not really an exception but instead is intentional

akshayka:
> I think the communication is good, but it looks like other Exceptions while this is not really an exception but instead is intentional

Are you talking about the traceback or the "Ancestor stopped" output?
- I think we need the traceback. We want users to be able to catch these exceptions and handle them. (Again, if library code raises.)
- For `Ancestor stopped`, I can experiment with removing the red outline and just using regular paragraph color for the output. It is good for it to stand out visually, though: in a future change, we need to disable execution of cells whose ancestors raised exceptions (a regular exception or a `MarimoStopError`). Otherwise the program can get in an inconsistent change.

mscolnick:
I think both look a bit more alarming than the user is intending - and requires user intervention when it probably doesn't. I think if the user wanted to raise an error, they could do that manually instead of stop. 

Regardless, a hill I won't die on and any suggestions are not breaking changes so we can iterate on the presentation after feedback.

mscolnick (dismissed):

akshayka:
> I think both look a bit more alarming than the user is intending - and requires user intervention when it probably doesn't. I think if the user wanted to raise an error, they could do that manually instead of stop. 

Makes sense. I'll at least remove the error outline from the ancestor stopped.

akshayka:
@mscolnick take another look?

mscolnick (dismissed):

PR #15, opened 2023-08-21
Update hover area to extend left and right

mscolnick:


akshayka (commented):
looks good except e2e test is failing

akshayka:
> looks good 

Actually just ran `marimo edit` and seeing this:

image

> except e2e test is failing

click events on run button are getting intercepted by the tray div

akshayka (approved):

PR #17, opened 2023-08-22
Don't print a traceback on mo.stop()

akshayka:
Analogous to `sys.exit()`.


akshayka:
@mscolnick Came around to your suggestion 😄

mscolnick (approved):

PR #25, opened 2023-08-23
mo.image

mscolnick:
Add mo.image

- common props are exposed 
- common styles are pulled out too
- you can pass in arbitrary styles as well
- created a helper class for building divs, images
- tests


marimo/_plugins/stateless/image.py
```diff
@@ -0,0 +1,55 @@
...
+        src="https://marimo.io/logo.png",
+        alt="Marimo logo",
+        width=100,
+        height=100,
+        rounded=True,
+    )
+    ```
+    **Args.**
+
+    - `src`: the URL of the image
```
akshayka, line 35: What do you think about letting people use images on their filesystem as well? We could read these as bytes and create a data URL out of them.

akshayka (commented):
Haven't gotten to do a thorough review yet, but I played with it and it feels really nice! Lot more satisfying than having to write the HTML manually.

Left one question about using local files. Will come back and review source later.

akshayka (approved):
Awesome! (The HTMLBuilder utility is cool too.)

I think this change accidentally adds an empty refresh.py smoke test. But otherwise looks good!

PR #28, opened 2023-08-24
Center error boundary

mscolnick:


frontend/src/css/index.css
```diff
@@ -23,8 +23,10 @@ body {
 
 body,
 #root {
-  height: 100%;
+  height: 100vh;
   width: 100%;
+  display: flex;
+  flex-direction: column;
```
akshayka, line 29: Just curious, what motivated these changes?

mscolnick: it was so the children could `flex` itself larger (i find flex easier to deal with). the other option is to add `position:absolute; height: 100%` instead of flex

akshayka (approved):

PR #29, opened 2023-08-24
Set default values for 'autosave' to 'after_delay' and 'autosave_delay' to 5000 in both frontend and MarimoConfig

mscolnick:


akshayka (dismissed):
Thanks! I would keep 1000 so it feels responsive. But otherwise lgtm.

PR #41, opened 2023-08-30
fix: focus on the first cell when editing on startup

mscolnick:


akshayka (approved):
This still occasionally doesn't work. I was trying `marimo edit ui.py` (UI tutorial) and it worked maybe 5/6 times. Maybe request another frame? Anyway it's better than before ...

PR #44, opened 2023-08-30
feat: add debounce to slider and number input

mscolnick:
- smoke test included to easily verify

API
- `debounce` is a true/false instead of a delay. ideally we abstract away too many details and we can always at a `debounce_delay` later.

Impl
- for slider, the debounce isn't a time, but rather on mouse up
- for number, the default is 200ms
- although they are different implementations, they achieve the same thing so i think we should keep the naming (`debounce`) consistent

### DIFF
--- a/frontend/src/plugins/impl/NumberPlugin.tsx
+++ b/frontend/src/plugins/impl/NumberPlugin.tsx
@@ -1,10 +1,11 @@
 /* Copyright 2023 Marimo. All rights reserved. */
-import { useEffect, useId, useRef } from "react";
+import { useId, useRef } from "react";
 import { z } from "zod";
 
 import { IPlugin, IPluginProps, Setter } from "../types";
 import { Input } from "../../components/ui/input";
 import { Labeled } from "./common/labeled";
+import { useDebounceControlledState } from "@/hooks/useDebounce";
 
 type T = number;
 
@@ -13,6 +14,7 @@ interface Data {
   stop: T;
   step?: T;
   label: string | null;
+  debounce: boolean;
 }
 
 export class NumberPlugin implements IPlugin<T, Data> {
@@ -24,6 +26,7 @@ export class NumberPlugin implements IPlugin<T, Data> {
     start: z.number(),
     stop: z.number(),
     step: z.number().optional(),
+    debounce: z.boolean().default(false),
   });
 
   render(props: IPluginProps<T, Data>): JSX.Element {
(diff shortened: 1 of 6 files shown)


frontend/src/plugins/impl/NumberPlugin.tsx
```diff
@@ -44,30 +47,16 @@ interface NumberComponentProps extends Data {
...
-    if (inputRef.current !== null) {
-      inputRef.current.addEventListener(
-        "wheel",
-        (e) => {
-          if (e.target === inputRef.current) {
-            e.preventDefault();
-          }
-        },
-        { passive: false }
-      );
```
akshayka, line 65: Can you disable/ignore scrolling again? This was based on feedback

marimo/_plugins/ui/_impl/input.py
```diff
@@ -54,6 +54,7 @@ class number(UIElement[Optional[Numeric], Optional[Numeric]]):
     - `stop`: the maximum value of the interval
     - `step`: the number increment
     - `value`: default value
+    - `debounce`: whether to debounce the number picker
```
akshayka, line 57: "whether to debounce (rate-limit) value updates from the frontend"?

Or something like that.

marimo/_plugins/ui/_impl/input.py
```diff
@@ -125,6 +128,8 @@ class slider(UIElement[Numeric, Numeric]):
     - `stop`: the maximum value of the interval
     - `step`: the slider increment
     - `value`: default value
+    - `debounce`: whether to debounce the slider to only send
+        the value on mouseup
```
akshayka, line 132: "whether to debounce the slider to only send the value on mouse-up or drag-end" ?

akshayka (commented):
The debouncing feels great! Thanks for the smoke test.

One comment about restoring some behavior, and a couple comments on the docstrings.

frontend/src/plugins/impl/NumberPlugin.tsx
mscolnick: i moved it to:
```
        onWheel={(e) => {
          e.preventDefault();
        }}
```

were you hitting this issue still with that change? i tried to repro and couldnt

akshayka: Oh weird. I thought I was. But now I can't repro ... my bad.

akshayka (approved):

PR #54, opened 2023-08-31
feat: mo.ui.microphone to record audio and mo.audio to play audio

mscolnick:
Screenshot 2023-08-31 at 4 31 05 PM


marimo/_plugins/stateless/audio.py
```diff
@@ -0,0 +1,76 @@
...
+def audio(
+    src: Union[str, io.IOBase],
+) -> Html:
+    """Render an audio file as HTML.
+
+    **Example.**
+
+    ```python3
+    mo.audio(
+        src="/recording.wav"
```
akshayka, line 23: Let's use a URL. Otherwise some users might think you can pass a path to a file on your local filesystem.

Maybe

```python3
mo.audio(
    src="https://upload.wikimedia.org/wikipedia/commons/8/8c/Ivan_Ili%C4%87-Chopin_-_Prelude_no._1_in_C_major.ogg"
)
```

marimo/_plugins/ui/_impl/microphone.py
```diff
@@ -0,0 +1,48 @@
...
+class microphone(UIElement[str, io.BytesIO]):
+    """
+    An element that can be used to record audio via the user's browser.
+    The user must grant permission to use the microphone.
+
+    **Example.**
+
+    ```python
+    mic = mo.ui.microphone()
+    ```
```
akshayka, line 22: Can you include an example of using the value?

Perhaps
```python3
recording = mo.audio(mic.value)
recording
```

marimo/_plugins/ui/_impl/microphone.py
```diff
@@ -0,0 +1,48 @@
...
+
+    **Attributes.**
+
+    - `value`: The blob of the recorded audio, as a bytes object.
```
akshayka, line 26: "as an `io.BytesIO` object". A bytes object is something different (https://docs.python.org/3/library/stdtypes.html#bytes-objects)

docs/api/layouts/index.md
```diff
@@ -23,9 +23,18 @@ certain way.
...
   marimo.accordion
   marimo.callout
-  marimo.image
   marimo.hstack
   marimo.vstack
   marimo.tabs
   marimo.tree
 ```
+
+There are also a set of utility elements that make it easier to build HTML components.
```
akshayka, line 32: Good call ... later, when we have `mo.video` and `mo.ui.webcam`, we can split these out into another page.

akshayka (commented):
Looks great. Just a couple of small comments about the docstrings. Nice Python code by the way.

akshayka (approved):

PR #58, opened 2023-08-31
Update LICENSE

tschm:
Make sure your name is in there...

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -176,18 +176,7 @@
 
    END OF TERMS AND CONDITIONS
 
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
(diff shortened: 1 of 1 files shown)


tschm:
recheck

akshayka:
Thanks @tschm!

I'm looking into this. My current understanding is that the Appendix is not actually part of the license, but instead a recommendation to include the boilerplate below it as the header of individual files. I got this understanding from

https://www.apache.org/licenses/LICENSE-2.0.html#apply

This interpretation is congruent with how official Apache projects are set up, e.g. Apache Spark: https://github.com/apache/spark/blob/master/LICENSE

We have some automation that prepends a copyright notice but not the appendix/header. Perhaps we should do that ...

tschm:
@akshayka you might be perfectly right. But a copyright notice in each *.py file could be a bit too much...

akshayka:
Closing as per our discussion. We should perhaps follow up and prepend the license notice to each file ...

PR #59, opened 2023-08-31
feat: justify API

akshayka:
- `mo.center()`, `mo.left()`, `mo.right()`
- added for convenience: easier to remember (and read) than `mo.as_html(...).center()`


docs/api/layouts/justify.md
```diff
@@ -0,0 +1,13 @@
+# Stacks
```
mscolnick, line 1: Typo?

akshayka: oops thanks

mscolnick (approved):

PR #65, opened 2023-09-05
feat: Download PNG 

mscolnick:
* download full page as PNG
* download output as PNG
* create a Cell Action Menu where the drag handle is
* change the accent background and foreground colors

### DIFF
--- a/frontend/src/css/globals.css
+++ b/frontend/src/css/globals.css
@@ -40,11 +40,11 @@
     --secondary: 210 40% 96.1%;
     --secondary-foreground: 222.2 47.4% 11.2%;
 
-    /* tailwind blue-500 */
-    --accent: 217 91% 60%;
+    /* blue-3 */
+    --accent: 209 100% 96.5%;
+    /* blue-12 */
+    --accent-foreground: 216 71% 23%;
 
-    /* white */
-    --accent-foreground: 100 100% 100%;
     --destructive: 0 100% 70%;
     --destructive-foreground: 210 40% 98%;
     --error: 0 77% 64%;
--- a/frontend/src/css/globals.css
+++ b/frontend/src/css/globals.css
@@ -16,6 +16,8 @@
 @import url("@radix-ui/colors/amber-dark.css");
 @import url("@radix-ui/colors/yellow.css");
 @import url("@radix-ui/colors/yellow-dark.css");
+@import url("@radix-ui/colors/blue.css");
+@import url("@radix-ui/colors/blue-dark.css");
 
 @tailwind base;
 @tailwind components;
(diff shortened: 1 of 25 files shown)


akshayka:
Cool!

Some questions/comments about the UI. I can think through some of these too.
- How do we decide what goes in the actions menu in the top-right, versus controls on the bottom left? Download button could have been an additional control; conversely keyboard shortcuts could go in actions menu in top right.
- When I take a screenshot, the content is left-aligned instead of center-aligned. Is this intentional? It's different from how it appears in the browser
- The cell-specific download button shouldn't show up for cells that have no output. But maybe we should have a dropdown menu for cell actions instead of adding a button to the GUI ...

mscolnick:
For the first point, let's chat live. 

The other 2 are bugs, I'll fix them.

frontend/src/css/globals.css
```diff
@@ -40,11 +40,11 @@
     --secondary: 210 40% 96.1%;
     --secondary-foreground: 222.2 47.4% 11.2%;
 
-    /* tailwind blue-500 */
-    --accent: 217 91% 60%;
+    /* blue-3 */
+    --accent: 209 100% 96.5%;
+    /* blue-12 */
+    --accent-foreground: 216 71% 23%;
```
mscolnick, line 46: @akshayka, i think this looks better (lighter background on darker foreground), but let me know if you prefer the other way

mscolnick: we can chat live about this one and cycle through the scale together

frontend/src/editor/cell/cell-actions.tsx
```diff
@@ -0,0 +1,255 @@
...
+            className="dark dark-theme w-full bg-card"
+            tabIndex={-1}
+          >
+            <div className="text-xs text-foreground-muted flex flex-col text-center">
+              <span>
+                <span className="text-foreground font-semibold">Drag </span>to
+                move cell
+              </span>
+              <span>
+                <span className="text-foreground font-semibold">Click </span>to
```
akshayka, line 194: Can we allow the user to open the menu with `Enter` too, if they have tabbed to focus on it?

Not urgent, can do in a follow-up, up to you

akshayka (commented):
This is really great! I like both the global menu and the cell-specific menu, the actions there and how they look.

frontend/src/css/globals.css
mscolnick: ive reverted the accent change - but the accent probably shouldnt be dark. i think this color got switched up at some point and we should instead make `primary` our dark blue. ill handle this in a follow up

frontend/src/editor/cell/cell-actions.tsx
```diff
@@ -0,0 +1,255 @@
...
+                open menu
+              </span>
+            </div>
```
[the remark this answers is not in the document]
mscolnick: ill change back to our white - but i quite liked the dark, so lets chat live about making all tooltips dark

frontend/src/editor/cell/cell-actions.tsx
mscolnick: ill do it in a followup

akshayka (approved):

Issue #64, opened 2023-09-05
Embedded PDF viewer


handrew:
### Description

Would love a `marimo.ui` component which allows me to embed PDFs in a Marimo notebook, and ideally allows me to determine which part of the PDF to start at.

### Suggested solution

Open to what you guys come up with. 

### Alternative

Have considered just printing plain text.

### Additional context

Context is that I am building an LLM-enabled information retrieval app.

daviddai-evenup:
@akshayka I'm unable to render local pdfs (see screenshot below), but it works fine when using a web URL. Interestingly I'm able to render local images, so it doesn't seem to be a file access issue. Do you know what might be causing this?

<img width="696" alt="Image" src="https://github.com/user-attachments/assets/de115bcf-dcf5-4917-963e-fb3b7edf122d" />

akshayka:
This function doesn't support local filepaths. You have to pass it an open file. See the code example in the thread above

PR #66, opened 2023-09-05
feat: add PDF with initial page number

mscolnick:
https://github.com/marimo-team/marimo/issues/64

Screenshot 2023-09-05 at 6 07 19 PM

### DIFF
--- a/marimo/_plugins/stateless/pdf.py
+++ b/marimo/_plugins/stateless/pdf.py
@@ -1,0 +1,76 @@
+# Copyright 2023 Marimo. All rights reserved.
+from __future__ import annotations
+
+import base64
+import io
+from typing import Any, Optional, Union
+
+from marimo._output.builder import h
+from marimo._output.hypertext import Html
+from marimo._output.rich_help import mddoc
+from marimo._output.utils import create_style
+
+
+@mddoc
+def pdf(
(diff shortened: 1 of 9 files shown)


marimo/_plugins/stateless/pdf.py
```diff
@@ -0,0 +1,76 @@
...
+from marimo._output.rich_help import mddoc
+from marimo._output.utils import create_style
+
+
+@mddoc
+def pdf(
+    src: Union[str, io.IOBase],
+    initial_page: Optional[int] = None,
+    width: Optional[Union[int, str]] = "100%",
+    height: Optional[Union[int, str]] = "70vh",  # arbitrary, but good default
```
akshayka, line 19: `Optional` means can be `None`, not can be omitted Is that intended here? If not then should remove `Optional`

akshayka (commented):
Nice! Just one comment

marimo/_plugins/stateless/pdf.py
mscolnick: yea I wanted the user to be allowed to unset the default if it's useful. I personally don't think they will want to or need to , but it could come from and outside value

akshayka (approved):

PR #68, opened 2023-09-06
feat: mo.download button

mscolnick:
Screenshot 2023-09-05 at 9 31 35 PM


marimo/_plugins/stateless/download.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    ```
+
+    **Args.**
+
+    - `data`: The data to download. Can be a string, bytes,
+        or a file-like object.
+    - `filename`: The name of the file to download.
+        If not provided, the name will be guessed from the data.
+    - `mimetype`: The mimetype of the file to download.
```
akshayka, line 49: Our users won't know what a mimetype is. Can you include some examples in the docstring? And add "If not provided, the name will be guessed from the filename".

akshayka (approved):
Approved except for 1 comment

akshayka (approved):

PR #75, opened 2023-09-07
Add microphone to sidebar

akshayka:
add top padding to demo iframe


docs/_static/css/custom.css
```diff
@@ -29,6 +29,7 @@ iframe.demo {
   height: 250px;
   border: 1px solid var(--color-background-border);
   border-radius: 5px;
+  padding-top: 1rem;
```
mscolnick, line 32: padding came form the `components.py` actually. it was modified there from an old version of marimo which had the large padding on it

PR #94, opened 2023-09-12
add event listeners to send copy/paste to vscode

mscolnick:


akshayka (approved):
LGTM. If it's not too much trouble, can we also handle Cut/Paste, i.e. intercept and propagate `"cut"` events?

PR #99, opened 2023-09-12
experimental: add config block for experimental, support feature-flags on the FE, add theming

mscolnick:
toml could look like this:
```toml
...

[experimental]
theming = true
```
Screenshot 2023-09-12 at 2 46 47 PM
Screenshot 2023-09-12 at 2 46 42 PM
Screenshot 2023-09-12 at 2 46 38 PM


akshayka (approved):
Nice! Looks great.

As an aside, in the future I think dark mode, or more generally the selected theme, could be part of the user config. So the theme would be applied to all possible browsers/frontends. Not relevant for this change though.

mscolnick:
@akshayka i did consider that - i think i will make that change once it comes out of experimental. 

i was only hesitant at first, because most applications use localstorage or cookies for dark-mode, but since we have user config set up nicely, i do think it makes sense to keep it in there.

PR #105, opened 2023-09-13
cell config, disabled cells

mscolnick:
This change adds per-cell configuration to marimo:
- exposed to the user through the frontend's cell action menu
- written to the `.py` file via the `App.cell` decorator

This change also implements our first cell configuration, `disabled`:
- A disabled cell can't be run, manually, through UI components, or anything else
- A disabled cell's descendants can't be run
- When re-enabled, a cell/its descendants are re-run if they became stale (if at least one of their ancestors ran while they were disabled/disabled transitively)

Disabling cells is useful when you're editing the notebook and don't want to accidentally trigger expensive computation.
- Edit and run a cell's ancestor without running the disabled cell.
- Disabling a cell lets you add code to the graph without running it.

Notebook actions added:
- Disable all cells
- Enable all cells

(Edited by Akshay)


mscolnick (commented):
@akshayka looks great from my end

frontend/src/App.tsx
```diff
@@ -57,6 +58,7 @@ export const App: React.FC<AppProps> = ({ userConfig, appConfig }) => {
   const [viewState, setViewState] = useAtom(viewStateAtom);
   const [filename, setFilename] = useState(getFilenameFromDOM());
   const [savedCodes, setSavedCodes] = useState<string[]>([""]);
+  const [savedConfigs, setSavedConfigs] = useState<CellConfig[]>([]);
```
mscolnick, line 61: in a followup, for QOL, i'll try to move this stale state to the reducer (less logic in the components)

frontend/src/components/icons/multi-icon.tsx
```diff
@@ -1,15 +1,28 @@
 /* Copyright 2023 Marimo. All rights reserved. */
 import React, { PropsWithChildren } from "react";
 
+import "./multi-icon.css";
+
+interface MultiIconProps {
+  layerTop?: boolean;
```
mscolnick, line 7: nice addition

PR #119, opened 2023-09-18
show notebook actions in command pallete

mscolnick:


frontend/src/editor/CommandPallette.tsx
```diff
@@ -94,6 +118,9 @@ export const CommandPallette = () => {
           {HOTKEYS.iterate().map((shortcut) =>
             renderShortcutCommandItemIfNotRecent(shortcut)
           )}
+          {notebookActionsWithoutHotkeys.map((action) =>
```
akshayka, line 121: Should we track notebook commands in recents, too? Doesn't have to be in this PR.

akshayka (approved):
LGTM modulo lint

frontend/src/editor/CommandPallette.tsx
mscolnick: good call, i can do it in this PR

akshayka (approved):

PR #158, opened 2023-09-26
feat: Interactive Matplotlib Plots

akshayka:
This change makes it possible to render matplotlib plots using matplotlib's interactive backend.

image

A new public API symbol is added: `marimo.mpl.interactive`

This function takes a figure or axes object and returns` Html` that renders the plot.
- Each figure is served by its own webserver
- When a cell is invalidated by the kernel, its interactive figure servers are torn down (if any)


marimo/__init__.py
```diff
@@ -28,6 +28,7 @@
     "image",
     "left",
     "md",
+    "mpl_interactive",
```
mscolnick, line 31: You recommended mpl.interactive which I actually like more than the underscore

marimo/_plugins/stateless/mpl.py
```diff
@@ -0,0 +1,344 @@
...
+    shutdown_event: Optional[asyncio.Event] = None
+
+
+class InteractiveMplRegistry:
+    """Registry of figures created by each cell.
+
+    Allows the kernel to tear down all figures for a given cell.
+    """
+
+    _registry: dict[CellId_t, list[CleanupHandle]] = {}
```
mscolnick, line 251: should this be a set of CleanupHandle to prevent duplicates? if that is even possible

akshayka: It's tricky because you'd want set membership to depend on the id of the shutdown event, but that might not exist when the handle is first registered. `CleanupHandle.shutdown_event` starts off as `None` and is later re-assigned to an `Event` in the server thread, for compatibility with Python 3.8 and 3.9 (creating the event in the main thread lead to an error). 

Right now a duplicate Event won't ever be registered. I agree for defensibility a set would be nice ... but I think in this case it's not worth the extra effort (i.e. synchronizing to wait until the shutdown event is created before registering).

mscolnick (approved):

PR #162, opened 2023-09-28
mo.ui.altair_chart

mscolnick:
Although `plot` is more popular than `chart`, since out initial usage is a wrapper around "vega-**charts**" and "altair **charts**", it felt a bit confusing to switch back to `plot`.

* adds `mo.ui.chart` wrapper that can take in a altair chart or vega-spec and will render in the output. we do this auto-magically:
  * we auto add features to the vega-spec to make it interactive and selectable
  * we propagate selection back to the Kernel and create a filtered dataframe from the initial data
* adds a `VegaPlugin` which takes in a vega-spec (created from `mo.ui.chart`). there is a lot of memoization going on in the FE to avoid re-renders. 
* `mo.ui.chart` requires altair and pandas. we can keep them optional, unless using charts
* if a consumer passes in a chart with selection/interactivity enabled, we dont remove it and only append to it.

Follow-up
* instead of passing in the full data into the altair spec. we can create a csv with pandas and lazy-load. vega will load the data on the FE and using a csv we can reduce the file size by ~80% since it columnar


frontend/src/plugins/core/componentFactory.tsx
```diff
@@ -241,6 +241,7 @@ export function registerReactComponent<T>(plugin: IPlugin<T, unknown>): void {
 
     disconnectedCallback() {
       this.observer.disconnect();
+      this.root?.unmount();
```
mscolnick, line 244: this was a pretty big memory leak. i am going to pull this out into a different PR, with the debug tools

docs/api/inputs/charts.md
```diff
@@ -0,0 +1,79 @@
...
+# Display the chart and its data filtered by the selection
+mo.vstack([chart, chart.value.head()])
+```
+
+## Learning Altair and Vega-Lite
+
+If you're new to **Altair** or **Vega-Lite**, we highly recommend exploring the [Altair documentation](https://altair-viz.github.io/) and [Vega-Lite documentation](https://vega.github.io/vega-lite/). **Altair** simplifies the process of creating **Vega-Lite** charts in Python and is the tool we utilize in many of our examples. **Vega-Lite** an exceptional tool for creating interactive charts and serves as the backbone for marimo's reactive charting capabilities.
+// OR
+If you're new to Vega-Lite, we highly recommend exploring the [Vega-Lite documentation](https://vega.github.io/vega-lite/). Vega-Lite is an exceptional tool for creating interactive charts and serves as the backbone for marimo's reactive charting capabilities. Additionally, we suggest familiarizing yourself with [Altair](https://altair-viz.github.io/), a Python library specifically designed for generating Vega-Lite charts. **Altair** simplifies the process of creating Vega-Lite charts in Python and is the tool we utilize in many of our examples.
+
```
mscolnick, line 48: [ ] need to pick one of these paragraphs

frontend/src/editor/output/JsonOutput.tsx
```diff
@@ -77,7 +77,7 @@ function inferBestFormat(data: unknown): "tree" | "raw" {
 const LEAF_RENDERERS = {
   "image/": (value: string) => <ImageOutput src={value} />,
   "video/": (value: string) => <VideoOutput src={value} />,
-  "text/html": (value: string) => <HtmlOutput html={value} inline={true} />,
+  "text/html": (value: string) => <HtmlOutput html={value} />,
```
mscolnick, line 80: something to watch out for, i removed the inline so charts in a json would expand fully, but this may break other things. easy to revert, because people may not even want to put charts in a json

frontend/src/plugins/impl/vega/VegaPlugin.tsx
```diff
@@ -0,0 +1,155 @@
...
+        onNewView={handleNewView}
+      />
+    </>
+  );
+};
+
+const actions = {
+  source: false,
+  compiled: false,
+};
```
mscolnick, line 155: bunch of debug stuff in here - its all ignored in production, so planned to leave it in

akshayka:
This is amazing!!

A couple of quirks I found.

1. When I `hstack`, I don't see the plot in normal width
image

2. When I `hstack` in full-width, the plot doesn't stretch
image

akshayka:
> 2. When I `hstack` in full-width, the plot doesn't stretch

If I start in full-width, it does stretch. But if I start in normal width then switch to full-width, it doesn't stretch

akshayka:
Something to think about: should an empty selection yield an empty dataframe or the full data?

Right now it yields the full data. But somehow I expected it to yield an empty dataframe. I think this is because I expected my code to be able to distinguish between when a user has selected something (dataframe non-empty) and when they haven't (dataframe empty)

image

akshayka:
@mscolnick Here's the issue with `bar` selection. I selected a bar, and the dataframe is correctly updated but the plot doesn't show the selection.

result (10)


```python
import marimo

__generated_with = "0.1.21"
app = marimo.App()


@app.cell
def __():
    import marimo as mo
    return mo,


@app.cell
def __(mo):
    import altair as alt
    from vega_datasets import data

    chart = alt.Chart(data.cars()).mark_point().encode(
        x='Horsepower',
        y='Miles_per_Gallon',
        color='Origin',
    )

    chart = mo.ui.chart(chart)
    return alt, chart, data


@app.cell
def __(chart, mo):
    mo.vstack([chart, chart.value])
    return


@app.cell
def __(mo):
    mo.md("**Bar selection issue.**")
    return


@app.cell
def __(alt, data, mo):
    hist = alt.Chart(data.cars()).mark_bar().encode(
        x=alt.X("Miles_per_Gallon:Q", bin=True), y="count()"
    )
    hist = mo.ui.chart(hist)
    return hist,


@app.cell
def __(hist, mo):
    mo.vstack([hist, hist.value])
    return


if __name__ == "__main__":
    app.run()
```

akshayka:
@mscolnick Interval selection appears to ignore and unset the field selection. 

[Screencast from 10-09-2023 04:40:12 PM.webm](https://github.com/marimo-team/marimo/assets/1994308/34878bc2-39ee-474d-bcfe-072b599d72d0)

marimo/_plugins/ui/_impl/chart.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Selection for binned charts is not yet implemented
+        if _has_binning(vega_spec):
+            sys.stderr.write(
+                "Binning + selection is not yet supported in "
+                "marimo.ui.chart.\n"
+                "If you'd this feature, please file an issue: "
+                "https://github.com/marimo-team/marimo/issues\n"
+                "In the meantime, use `pd.cut` to bin data in Python "
+                "before creating the chart.\n"
```
akshayka, line 125: If it's short, can you include an example how to do this? I've never seen that function. Googled and it wasn't apparent to me how to use it.

mscolnick: we can just remove this

akshayka (approved):
This is amazing!!

PR #163, opened 2023-09-28
fix: unmount react component when web-component unmounts

mscolnick:


akshayka (dismissed):
LGTM modulo failing lint

PR #164, opened 2023-09-29
feat: marimo.output

akshayka:
Functions for writing to the output area.

- `mo.output.replace`
- `mo.output.append`
- `mo.output.clear` (which is just `mo.output.replace(None)`)


marimo/_runtime/output/_output.py
```diff
@@ -0,0 +1,71 @@
...
+        channel="output",
+        mimetype=output.mimetype,
+        data=output.data,
+        cell_id=cell_id,
+        status=None,
+    )
+
+
+@mddoc
+def replace(value: object) -> None:
```
akshayka, line 26: Thoughts on naming?

I know we talked about `mo.output()` and `mo.output.append()`, but I thought this was clearer. I also considered `write` instead of `replace`, but opted for `replace` because it's more descriptive.

frontend/src/core/state/cell.ts
```diff
@@ -22,6 +22,7 @@ export function transitionCell(
       // the cell's current code if the user modified it.
       break;
     case "running":
+      console.log("setting runstart ts to", message.timestamp);
```
mscolnick, line 25: remove?

frontend/src/editor/Cell.tsx
```diff
@@ -142,7 +144,18 @@ const CellComponent = (
...
+    runStartTimestamp !== null &&
+    output.timestamp > runStartTimestamp;
+  console.log("output ts?", output?.timestamp);
+  console.log("runstarttimestamp ?", runStartTimestamp);
+  console.log("output received?", outputReceivedWhileRunning);
+  const loading =
+    (status === "running" && !outputReceivedWhileRunning) ||
+    status === "queued";
+  console.log("status:", status);
+  console.log("loading:", loading);
```
mscolnick, line 158: remove all logs

marimo/_runtime/output/_output.py
mscolnick: i like the naming that you landed on - also reflecting on the `mo.output()` `mo.output.append()`, some people get confused (and so does IDE docs) when having something callback and has further properties

mscolnick (dismissed):
looks good pending removing console.logs

i'll follow up with a unit test and e2e test too

frontend/src/core/state/cell.ts
akshayka: sorry, these were debugging logs that I forgot to remove

frontend/src/editor/Cell.tsx
```diff
@@ -143,6 +145,19 @@ const CellComponent = (
 
   const needsRun = edited || interrupted;
   const loading = status === "running" || status === "queued";
+  // output may or may not be refreshed while a cell is running, so
+  // we need to check if an output was received
+  const outputReceived =
+    output !== null &&
+    runStartTimestamp !== null &&
+    output.timestamp > runStartTimestamp;
+  const outputStale =
```
akshayka, line 154: @mscolnick fyi decoupled output staleness from loading status. Reading your PR I realized it was incorrect for me to modify `loading`, since it's used to control runnability of cells

PR #165, opened 2023-09-29
Add outputs e2e test

mscolnick:


frontend/src/editor/Cell.tsx
```diff
@@ -186,18 +181,25 @@ const CellComponent = (
     [editorView, prepareToRunEffects]
   );
 
-  // Hack to provide the value of `running` to Code Mirror's EditorView.
```
mscolnick, line 189: this was as small drive-by change. but using `useEvent`, we pass the latest value of loading

PR #169, opened 2023-09-30
feat: capture stdout and stderr utilities

akshayka:
Utilities to capture stdout and stderr.


akshayka:
@mscolnick Not sure if these are worth including ... on the fence. Could just include the implementations as snippets in the docs. Let me know what you think.

mscolnick (dismissed):

PR #170, opened 2023-09-30
wip: pull out data from html components to avoid json parsing/unparsing and decoding

mscolnick:
in order to pass data to stateful and stateless plugin, we `json-stringify` and `html-encode` them into strings to put them into `data-*` attributes. instead this PR creates a "data locator" for each attribute so that it can be looked up in a map that is passed along with the message:
for example
```
data: <marimo-table data-rows="12345" />
data_store: {
  "12345": {...}
}
```

this should improve performance, message size, debuggability, and remove some encoding/decoding hacks

open issues: 
- [ ] i've regressed the Inf, -Inf, NaN issue when python creates JSON for `float('nan'), float('inf')`. but the fix should be less hacky than before
- [ ] accessing context in stateless plugins..feels...wrong. will need a better way to do this
- [ ] we sometimes create more info when the data value was a small string or number, should we special case small values or certain value types? my hunch is no, since they are already small, the extra size is still small

### DIFF
--- a/frontend/src/core/kernel/ui-data-registry.tsx
+++ b/frontend/src/core/kernel/ui-data-registry.tsx
@@ -1,0 +1,69 @@
+/* Copyright 2023 Marimo. All rights reserved. */
+import { Logger } from "@/utils/Logger";
+import { CellId } from "../model/ids";
+import { TypedString } from "../model/typed";
+import { Objects } from "@/utils/objects";
+import { repl } from "@/utils/repl";
+
+export type DataLocatorId = TypedString<"DataLocatorId">;
+
+/**
+ * Registry to track UIElement values.
+ */
+export class UIElementDataRegistry {
+  /**
+   * Shared singleton instance.
+   */
+  static readonly INSTANCE = new UIElementDataRegistry();
+
+  private constructor() {
+    repl(this, "UI_ELEMENT_DATA_REGISTRY");
+  }
+
+  // We want fast lookups by DataLocatorId and fast removals by CellId.
+  private data = new Map<DataLocatorId, unknown>();
+  private removeHandlers = new Map<CellId, () => void>();
+
+  /**
+   * Set the datastore for a Cell.
+   */
+  setValue(
+    cellId: CellId,
+    dataStore: Record<DataLocatorId, unknown> | undefined
+  ) {
+    // If there is already a remove handler for this cell, call it to remove
+    // the old values.
+    this.removeHandlers.get(cellId)?.();
+
+    if (!dataStore) {
+      return;
+    }
+
+    for (const [name, value] of Objects.entries(dataStore)) {
+      const dataLocatorId = name as DataLocatorId;
+      this.data.set(dataLocatorId, value);
+    }
+    this.removeHandlers.set(cellId, () => {
+      Objects.keys(dataStore).forEach((dataLocatorId) => {
+        this.data.delete(dataLocatorId);
+      });
+    });
+  }
+
+  /**
+   * Get the value for a DataLocatorId
+   */
+  getValue<T>(dataLocatorId: DataLocatorId): T {
+    if (!this.data.has(dataLocatorId)) {
+      Logger.warn(`No value found for ${dataLocatorId}`);
+    }
+    return this.data.get(dataLocatorId) as T;
+  }
+
+  /**
+   * Remove the value for a UIElement.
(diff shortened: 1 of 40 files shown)


marimo/_runtime/runtime.py
```diff
@@ -609,12 +610,17 @@ def _run_cells_internal(self, cell_ids: set[CellId_t]) -> set[CellId_t]:
...
+                flat_data_store = {
+                    k: v for data_store in data_stores.values() for k, v in data_store.items()
+                }
                 CellOp.broadcast_output(
                     channel="output",
                     mimetype=formatted_output.mimetype,
                     data=formatted_output.data,
                     cell_id=cell_id,
                     status=cell.status,
+                    data_store=flat_data_store,
```
akshayka, line 623: Why is this the only place that the data stores are sent?

marimo/_runtime/runtime.py
```diff
@@ -609,12 +610,17 @@ def _run_cells_internal(self, cell_ids: set[CellId_t]) -> set[CellId_t]:
                 if formatted_output.traceback is not None:
                     with self._install_execution_context(cell_id):
                         sys.stderr.write(formatted_output.traceback)
+                data_stores = get_context().ui_element_registry._data_stores
+                flat_data_store = {
+                    k: v for data_store in data_stores.values() for k, v in data_store.items()
+                }
```
akshayka, line 616: If I understand correctly, this sends all plugin data stores, across all cells. If so: Why are all data stores sent, instead of just ones that are created by this cell and/or referenced in this output?

- Could this be a performance regression? Eg sending tons of data created by multiple cells, even though none of those data are output in this cell.

frontend/src/core/kernel/ui-data-registry.tsx
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Set the datastore for a Cell.
+   */
+  setValue(
+    cellId: CellId,
+    dataStore: Record<DataLocatorId, unknown> | undefined
+  ) {
+    // If there is already a remove handler for this cell, call it to remove
+    // the old values.
+    this.removeHandlers.get(cellId)?.();
```
akshayka, line 36: Could this accidentally delete data locators that this cell doesn't "own"? I thought about it for a while and now I think the answer is "no" ... but wanted to double check with you. If I understand correctly the dataStore for a cell contains all data locators, not just ones created by this cell; that's why I originally thought that this could lead to freeing resources that you didn't want to free.

marimo/_runtime/output/_output.py
```diff
@@ -17,6 +17,7 @@ def write_internal(cell_id: CellId_t, value: object) -> None:
         channel="output",
         mimetype=output.mimetype,
         data=output.data,
+        data_store={},
```
akshayka, line 20: Why don't we need to send the data store here?

akshayka (commented):
Left some comments. I'm a bit confused by ownership of data locator IDs.

- My understanding is that every plugin has a data store, which is a map from data locator to value. That makes sense.

- It also seems that every cell has a collection of data stores, but my understanding is that this is just all data stores created across all cells. I'm confused why that is. I think the following would be more intuitive: when a cell finishes executing, all data locator-value pairs that it created, and only those data-locator-value pairs, are sent over. (You can get the current cell id with `get_context().kernel.execution_context.cell_id`)

Some additional documentation would help me.

Can talk in person. Totally possible I'm missing something simple.

akshayka:
Do we need to update the re-instantiation logic to send data stores in addition to UI element values? Just asking because I'm not sure. https://github.com/marimo-team/marimo/blob/c8a50d5476c1419eb9e5a93e2d64f65ad92b6598/frontend/src/core/websocket/useMarimoWebSocket.tsx#L97-L103

> this should improve performance,

Was the performance particularly bad with vega plots? I haven't seen many performance issues so far. Though matplotlib plots sometimes flicker.

>  we sometimes create more info when the data value was a small string or number, should we special case small values or certain value types? my hunch is no, since they are already small, the extra size is still small

I wouldn't special case, perf won't be noticeable.

mscolnick:
it's still a wip. the data stores are meant to be per cell and logic should be just on the request context of a single cell (not across cells). that should answer most of your questions

akshayka:
Closing since we haven't had activity in some time. We can revisit later if needed.

PR #179, opened 2023-10-03
Shared memory for virtual files

akshayka:
This change adds "virtual files" — files served by the marimo server, accessed via a route `/@file/{filename}`.
- This enables plugins like `mo.pdf` and `mo.image` to take URLs to virtual files instead of binary data in their HTML attributes
- Care is taken to make virtual filenames unique within kernel thread (by using random filenames) and across kernel threads (by prepending a thread ID)
- The mimetype of the file is inferred from its extension
- We use Python's `shared_memory` module to make the virtual file data accessible to the server (it's registered in the kernel) — this is really only needed in edit mode, in which the kernel is a separate process instead of a thread. In run mode, we could in the future add a simpler path just use (lock-protected) global memory ...
- Virtual files are tied to the cells that create them: they (and their associated shared memory) are destroyed when a cell is re-run, deleted, or otherwise invalidated

Because this change adds OS-specific logic, it also updates our unit tests and CLI tests to run on windows + macOS

### DIFF
--- a/marimo/_output/data/data.py
+++ b/marimo/_output/data/data.py
@@ -1,0 +1,48 @@
+# Copyright 2023 Marimo. All rights reserved.
+import io
+import random
+import string
+from typing import Union
+
+from marimo._runtime.context import get_context
+from marimo._runtime.virtual_file import VirtualFile, VirtualFileLifecycleItem
+
+
+def pdf(
+    data: io.IOBase,
+) -> VirtualFile:
+    """Create a virtual file from a PDF.
+
+    **Args.**
+    - data: The data the represents the PDF
+
+    **Returns.**
+    A `VirtualFile` object.
+    """
+    filename = _random_filename("pdf")
+    item = VirtualFileLifecycleItem(filename, "application/pdf", data)
+    get_context().virtual_file_registry.add(item)
+    return item.to_virtual_file()
+
+
+def image(
+    data: io.IOBase,
+    ext: string = "png",
+) -> VirtualFile:
+    """Create a virtual file from an image.
+
+    **Args.**
+    - data: The data the represents the image
+
+    **Returns.**
+    A `VirtualFile` object.
+    """
+    filename = _random_filename(ext)
+    item = VirtualFileLifecycleItem(filename, f"image/{ext}", data)
+    get_context().virtual_file_registry.add(item)
+    return item.to_virtual_file()
+
+
+def _random_filename(ext: string) -> str:
+    basename = "".join(random.choices(string.ascii_letters, k=8))
+    return f"{basename}.{ext}"
--- a/marimo/_output/data/data.py
+++ b/marimo/_output/data/data.py
@@ -1,6 +1,7 @@
 # Copyright 2023 Marimo. All rights reserved.
+import random
 import string
-import uuid
+import threading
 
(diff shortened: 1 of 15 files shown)


akshayka:
@mscolnick This shoudl work now, but haven't tested on Windows. Still a draft but can talk through it offline

marimo/_runtime/virtual_file.py
```diff
@@ -2,49 +2,94 @@
...
     url: str
+    filename: str
+    buffer: bytes
 
-    def __init__(self, filename: str) -> None:
-        raise NotImplementedError
+    def __init__(self, filename: str, mimetype: str, buffer: bytes) -> None:
+        self.filename = filename
+        self.mimetype = mimetype
+        self.url = f"/@file/{filename}/{mimetype}"
```
mscolnick, line 26: this produces something like `http://localhost:2718/@file/NiAlthcx.pdf/application/pdf`
vega (and possibly some other libraries) will look at the extension of the url to infer the mime.

thoughts on using the extension to convert to mime using `import mimetypes; mimetypes.guess_type`?

marimo/_server/api/virtual_file.py
```diff
@@ -0,0 +1,31 @@
...
+            # doing it in one line yields a 'released memoryview ...'
+            # because shared_memory has built in ref-tracking + GC
+            shm = shared_memory.SharedMemory(name=key)
+            buffer_contents = bytes(shm.buf)
+        except FileNotFoundError as err:
+            raise tornado.web.HTTPError(
+                HTTPStatus.NOT_FOUND,
+                reason="File not found",
+            ) from err
+        self.set_header("Content-Type", mimetype)
```
mscolnick, line 30: i also meant to add a Cache-Control:
```
self.set_header("Cache-Control", "max-age=86400")
```

mscolnick (dismissed):

mscolnick (approved):

marimo/_runtime/virtual_file.py
```diff
@@ -2,49 +2,128 @@
...
+        tries = 0
+        max_tries = 100
+        while registry.has(filename) and tries < max_tries:
+            filename = random_filename(self.ext)
+            tries += 1
+        if tries > max_tries:
+            raise RuntimeError(
+                "Failed to add virtual file to registry. "
+                "This is a bug in marimo. Please file an issue."
+            )
```
akshayka, line 68: @mscolnick I decided to force virtual file names to be unique, since we don't have a use-case for deterministic names yet. Let me know if this is ok?

marimo/_output/data/data.py
```diff
@@ -0,0 +1,35 @@
...
+
+    **Args.**
+
+    - data: Image data in bytes
+
+    **Returns.**
+
+    A `VirtualFile` object.
+    """
+    item = VirtualFileLifecycleItem(ext="pdf", buffer=data)
```
mscolnick, line 33: I think this ext is wrong

marimo/_runtime/virtual_file.py
mscolnick: Yea that's totally fine. I agree we don't need to pre optimize that

mscolnick (dismissed):

mscolnick (approved):

PR #184, opened 2023-10-06
feat: format on save

mscolnick:
format on save when the user saves the code. default to `False` in case they dont have a formatter installed

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -146,7 +146,9 @@ export const App: React.FC<AppProps> = ({ userConfig, appConfig }) => {
       .then(() => {
         if (userInitiated) {
           toast({ title: "Notebook saved" });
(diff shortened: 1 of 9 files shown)


frontend/src/App.tsx
```diff
@@ -143,8 +144,9 @@ export const App: React.FC<AppProps> = ({ userConfig, appConfig }) => {
       layout: getSerializedLayout(),
     })
       .then(() => {
-        if (showToast) {
+        if (userInitiated) {
           toast({ title: "Notebook saved" });
+          formatAll(updateCellCode);
```
akshayka, line 149: @mscolnick Check config to make sure `format_on_save` is enabled?

mscolnick: oh yea, duh, will add

akshayka (approved):

PR #190, opened 2023-10-09
improvement: support full-width input UI elements

mscolnick:
Added to these elements
- mo.ui.text,
- mo.ui.text_area
- mo.ui.button
- mo.ui.number
- mo.ui.dropdown
- mo.ui.multiselect
- mo.ui.date

Note in the smoke test, full width does not apply when its parent is flex (e.g. a hstack). There is a fixable solution by making it `flex: 1` and then updating a few components to be `display: contents`. but this can break other elements. easier to show live, but my guess is we can support this later

### DIFF
--- a/frontend/src/components/ui/combobox.tsx
+++ b/frontend/src/components/ui/combobox.tsx
@@ -34,6 +34,7 @@ interface ComboboxCommonProps<TValue> {
   search?: string;
   onSearchChange?: (search: string) => void;
   emptyState?: React.ReactNode;
+  className?: string;
 }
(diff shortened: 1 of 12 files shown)


akshayka:
> Note in the smoke test, full width does not apply when its parent is flex (e.g. a hstack). There is a fixable solution by making it flex: 1 and then updating a few components to be display: contents. but this can break other elements. easier to show live, but my guess is we can support this later

I think this is fine.

akshayka (approved):

PR #192, opened 2023-10-09
improvement: use VirtualFile in mo.download()

mscolnick:


marimo/_output/data/data.py
```diff
@@ -33,3 +43,45 @@ def image(data: bytes, ext: str = "png") -> VirtualFile:
     item = VirtualFileLifecycleItem(ext=ext, buffer=data)
     get_context().cell_lifecycle_registry.add(item)
     return item.virtual_file
+
+
+def any_data(data: Union[str, bytes, io.BytesIO], ext: str) -> VirtualFile:
```
akshayka, line 48: nite: `data`'s type should be `Optional[Union[...]]` since `data` can be `None`

akshayka: Looks like it should also include `io.FileIO`? What types are you trying to cover?

marimo/_runtime/virtual_file.py
```diff
@@ -34,15 +34,22 @@ class VirtualFile:
...
-        self.url = f"/@file/{filename}"
+        self.url = url
         self.buffer = buffer
 
 
+EMPTY_VIRTUAL_FILE = VirtualFile(
+    filename="empty.txt",
+    url="/@file/empty.txt",
+    buffer=b"",
+)
```
akshayka, line 47: Do you need to pre-load this into the registry?

marimo/_output/data/data.py
mscolnick: ill update to include both of those. ideally i cover all common types. not sure what will exist, but str and bytes cover most of what i've hit. ill also add `pd` here in the future as well

marimo/_output/data/data.py
```diff
@@ -33,3 +43,45 @@ def image(data: bytes, ext: str = "png") -> VirtualFile:
...
+    if data is None:
+        return EMPTY_VIRTUAL_FILE
+
+    if is_data_empty(data):
+        return EMPTY_VIRTUAL_FILE
+
+    # Base64 encoded data
+    if isinstance(data, str) and data.startswith("data:"):
+        buffer = data.split(",")[1]
+        item = VirtualFileLifecycleItem(ext=ext, buffer=buffer)
```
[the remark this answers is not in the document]
mscolnick: ok, ill convert or cast

marimo/_runtime/virtual_file.py
mscolnick: i can - it never actually gets called from mo.download, but in future it may

marimo/_output/data/data.py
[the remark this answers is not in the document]
akshayka: I don't think casting will work, the buffer needs to be `bytes` (or otherwise implement the buffer protocol)  because it gets stored in a `memoryview` (https://docs.python.org/3/library/stdtypes.html?#memoryview). If it doesn't raise an error it's probably undefined behavior

marimo/_output/data/data.py
```diff
@@ -33,3 +44,52 @@ def image(data: bytes, ext: str = "png") -> VirtualFile:
...
+
+
+def any_data(data: Union[str, bytes, io.BytesIO], ext: str) -> VirtualFile:
+    """Create a virtual file from any data.
+    It can be a string, bytes, or a file-like object.
+    For external URLs, these are passed through as-is.
+
+    **Args.**
+
+    - data: Data in bytes
```
akshayka, line 56: nit: "Data in bytes, or a string URL"

akshayka (approved):

PR #197, opened 2023-10-10
improvement: auto-encoding pandas/numpy, mo.ui.table takes in Dataframe

mscolnick:


marimo/_plugins/ui/_impl/table.py
```diff
@@ -0,0 +1,153 @@
...
+        disable
+    - `label`: text label for the element
+    - `on_change`: optional callback to run when this element's value changes
+    """
+
+    _name: Final[str] = "marimo-table"
+
+    def __init__(
+        self,
+        data: TableData,
```
akshayka, line 88: I know it's kind of ugly, but can you inline this type? That makes the docs / autocomplete more useful to the user.

marimo/_plugins/core/json_encoder.py
```diff
@@ -2,11 +2,54 @@
...
+    _is_pandas_installed = True
+except ImportError:
+    pass
+
 
 class WebComponentEncoder(JSONEncoder):
     """Custom JSON encoder for WebComponents"""
 
     def default(self, obj: Any) -> Any:
+        # Handle numpy objects
```
akshayka, line 26: nice!

marimo/_plugins/ui/_impl/table.py
```diff
@@ -0,0 +1,153 @@
...
+    ```
+
+    **Attributes.**
+
+    - `value`: the selected values, or `None` if no selection.
+    - `data`: the table data
+
+    **Initialization Args.**
+
+    - `data`:  a list of values representing a column, or a list of dicts
```
akshayka, line 70: update doc to mention that `data` can be a dataframe? I'd start with that.

"A pandas dataframe, or a list of values representing a column, or a list of dicts ...

marimo/_plugins/ui/_impl/table.py
```diff
@@ -0,0 +1,153 @@
...
+        {'first_name': 'Dwight', 'last_name': 'Schrute'}
+      ],
+      label='Users'
+    )
+    ```
+
+    ```python
+    # df is a Pandas dataframe
+    table = mo.ui.table(
+        data=df.to_dict('records'),
```
akshayka, line 56: Remove `to_dict`

marimo/_plugins/ui/_impl/table.py
```diff
@@ -0,0 +1,153 @@
...
+LOGGER = _loggers.marimo_logger()
+
+Numeric = Union[int, float]
+
+if TYPE_CHECKING:
+    import pandas as pd
+
+TableData = Union[
+    list[dict[str, str]],
+    Sequence[str | int | float | bool | MIME | None],
```
akshayka, line 30: Why do we need `list[dict[str, str]]`?

marimo/_plugins/ui/_impl/table.py
akshayka: I think that might also make the type error in the docs CI go away. Otherwise you can replace the `|` with `Union`

mscolnick: switched to union, and inline here (where its used in the docs), but kept everywhere else DRYed up

marimo/_plugins/ui/_impl/table.py
mscolnick: technically our values in the dicts need to be homegenious. this makes the typings super ugly, so i removed this and just type ignored the response from `to_dicts`

akshayka (dismissed):

PR #200, opened 2023-10-12
improvement: Custom altair data_transformer, load 1 million points

mscolnick:
* add custom altair data_transformer for csv and json (default to csv)
* add smoke_test for 1 million flights
* fix shared memory buffer to use the write buffer offset


akshayka (approved):
Looks great!

I pushed a change to rename the `figure` arg to `altair_chart` back to `chart`. My bad. I had changed it earlier in the previous PR, when I was experimenting with renaming to plot. Forgot to change it back.

PR #205, opened 2023-10-12
feat: add widths to flex

mscolnick:


akshayka (approved):

marimo/_plugins/stateless/flex.py
```diff
@@ -132,6 +147,9 @@ def hstack(
     - `align`: Align items vertically: start, end, center, or stretch.
     - `wrap`: Wrap items or not.
     - `gap`: Gap between items as a float in rem. 1rem is 16px by default.
+    - `widths`: "equal" to give items equal width; or a list of relative widths
```
akshayka, line 150: @mscolnick I added this for convenience, but let me know if it doesn't make sense ...

PR #212, opened 2023-10-14
Rename loading to status

akshayka:
Rename `mo.loading` to `mo.status`, because we may include other elements in this namespace that aren't strictly related to loading, such as pre-styled callouts (`mo.status.info`, `mo.status.error`, ...).

Since we no longer have an imperative API (`mo.loading.start`), we can easily nest other elements in this namespace.

Merge when ready. Once this is merged, I would count `ms/loading` as "merge when ready" as well.


mscolnick (approved):
I like the change!

PR #217, opened 2023-10-16
improvement: improve table performance by using virutal file for pandas data

mscolnick:
1. only use virtual file when we have a pandas dataframe. if its just a list of dicts, then we pass it along as-is. i dont want to conditionally convert it to pandas if its installed, nor write our own `to_csv` without pandas
2. by converting to_csv, we drop the NaNs again, which i think it a good thing as it makes the table more readable


marimo/_smoke_tests/dashboard.py
```diff
@@ -0,0 +1,413 @@
+# Copyright 2023 Marimo. All rights reserved.
```
mscolnick, line 1: this may be better in `examples`

akshayka:
> by converting to_csv, we drop the NaNs again, which i think it a good thing as it makes the table more readable

What do you mean by drop? Are they encoded as `null`, or are all rows with `NaN`s dropped? Former might be okay but I don't think we want the latter -- will confuse users if they are doing exploratory data analysis and there are rows missing.

mscolnick:
spoke offline: just the values are encoded with `null`s, so no rows are dropped

akshayka (approved):

PR #218, opened 2023-10-16
max width and max height utilities

akshayka:
Call `.max_width(pixels)` or `.max_height(pixels)` on any HTML object to constrain its max width or height.

e.g.,

```python
mo.hstack([...]).max_height(height=400)
```


akshayka:
Closing in favor of a more general `style` escape hatch, since it can be difficult to anticipate desired behavior (eg, with overflow or other things). will open a PR for that later

PR #223, opened 2023-10-18
mo.ui.dataframe

mscolnick:
Ready for first pass of a design-review

* add new plugin `mo.ui.dataframe`
* the `.value` is the filtered dataframe after all the transforms have been applied
* 1 tab to construct the transforms. another tab to view the generated python code
* in future PRs, can add more transformations but picked the top 5


frontend/package.json
```diff
@@ -59,6 +59,7 @@
     "@textea/json-viewer": "^3.2.2",
     "@types/js-cookie": "^3.0.5",
     "@types/react-grid-layout": "^1.3.4",
+    "@uiw/react-codemirror": "^4.21.19",
```
mscolnick, line 62: this basically binds codemirror to react. we do this for the cell's editors ourselves, but it can get messy so rather just use a library instead of do it again or merge those 2 codepaths

frontend/src/components/ui/table.tsx
```diff
@@ -71,7 +71,7 @@ const TableHead = React.forwardRef<
   <th
     ref={ref}
     className={cn(
-      "h-12 px-4 text-left align-middle font-medium text-muted-foreground [&:has([role=checkbox])]:pr-0",
+      "h-12 px-2 text-left align-middle font-medium text-muted-foreground [&:has([role=checkbox])]:pr-0",
```
mscolnick, line 74: i reduced the padding of the table too - happy to discuss this live

frontend/src/css/Cell.css
```diff
@@ -269,7 +269,7 @@
 }
 
 .output-area {
-  padding: 0.25rem 1.85rem;
+  padding: 0.25rem 1rem;
```
mscolnick, line 272: - [ ] padding reduction

frontend/src/editor/output/Outputs.css
```diff
@@ -9,8 +9,8 @@
 
 /* Just applied to vertical layout */
 .output-area .output {
-  margin-top: 1rem;
-  margin-bottom: 1rem;
+  margin-top: 0.5rem;
```
mscolnick, line 12: - [ ] padding reduction

frontend/src/plugins/impl/data-frames/DataFramePlugin.tsx
```diff
@@ -0,0 +1,112 @@
...
+                {typeof value === "string" ? (
+                  <Label>{value}</Label>
+                ) : (
+                  Array.isArray(value) && <Label>{value.join(", ")}</Label>
+                )}
+              </div>
+            </div>
+          );
+        })}
+      </div> */}
```
mscolnick, line 78: will be removed before merging

marimo/_server/api/model.py
```diff
@@ -37,6 +47,31 @@ def _build_value(value: Any, cls: Type[T]) -> T:
...
+            and first_arg_type.value == value
+        ):
+            return first_arg_type  # type: ignore[return-value]
+        if value not in arg_types:
+            raise ValueError(
+                f"Value '{value}' does not fit any type of the literal"
+            )
+        return value  # type: ignore[no-any-return]
+    elif type(cls) == type(Enum) and issubclass(cls, Enum):
+        return cls(value)  # type: ignore[return-value]
```
mscolnick, line 74: supports unions, enums, and discriminate unions

akshayka:
At first I thought this was a bug, with transform form disappearing when you delete a transform. But it turns out that just no transform is selected. I think this is an edge case when you delete the last transform. In that case can you focus on the new last transform / the one right before it?


[Screencast from 10-18-2023 05:08:24 PM.webm](https://github.com/marimo-team/marimo/assets/1994308/a71ca33f-3a8b-45e7-b682-c1d2ba9b15eb)

akshayka:
We should design what happens when a user enters an invalid transform, eg mean when there's a object/string column:

image

Some options:
- set dataframe.value to `None`, right now it just doesn't update on error
- try to block invalid transforms from the frontend based on what's known about dtypes

akshayka:
UX comment: it would be nice if it were possible to drag and reorder transforms. but we can punt that for later, probably more pressing things to do ...

akshayka:
**Dataframe preview.**
Maybe we do add dataframe previews in the initial version; showed the current plugin to a user and they were confused. I can help. From the backend, it shouldn't be much work (< 1 day). Not sure about the frontend. We can discuss and prioritize tomorrow.

The UI would be:

```
[tabs: transform/code]
[dataframe preview underneath]
```

since it would help to be able to see the dataframe as it's transformed.

frontend/src/plugins/impl/data-frames/panel.tsx
```diff
@@ -0,0 +1,240 @@
...
+                onSelect={(evt) => {
+                  evt.stopPropagation();
+                  window.open(
+                    "https://github.com/marimo-team/marimo/issues/new?title=New%20dataframe%20transform:&labels=enhancement&template=feature_request.yaml",
+                    "_blank"
+                  );
+                }}
+              >
+                <span className="underline text-primary text-xs cursor-pointer">
+                  Request a transform
```
akshayka, line 222: Love this! 🥳

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -0,0 +1,88 @@
...
+                # "describe-row-headers": described.index.to_list(),
+                "total": len(df),
+            },
+        )
+
+    def _convert_value(self, value: Dict[str, Any]) -> pd.DataFrame:
+        if value is None:
+            return self._data
+        transformations = parse_raw(value, Transformations)
+        return apply_transforms(self._data, transformations)
```
akshayka, line 88: - Protect with a `try/except Exception as e` in case the frontend sends an invalid transform
- Catch the exception and write to stderr? `sys.stderr.write("Error applying dataframe transform: %s\n\n" % str(e))`

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -0,0 +1,88 @@
...
+import inspect
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Final,
+    Optional,
+)
+
+import pandas as pd
```
akshayka, line 13: Need to protect with  `if TYPE_CHECKING` to fix failing CLI test

akshayka (commented):
Super cool!!

Left some small review comments on the Python.

Feel free to ignore the type errors in `handlers.py`.

Left some design/UX comments in the PR discussion thread. Let's talk tomorrow about the dataframe preview via RPCs.

frontend/src/css/Cell.css
mscolnick: reverting this

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -0,0 +1,107 @@
...
+            return self._data
+
+        try:
+            transformations = parse_raw(value, Transformations)
+            return apply_transforms(self._data, transformations)
+        except Exception as e:
+            sys.stderr.write(
+                "Error applying dataframe transform: %s\n\n" % str(e)
+            )
+            return self._data
```
akshayka, line 107: If I understand correctly — each time the user adds an additional transform in the UI, all the previous transforms are re-applied? If true, that could get pretty expensive, since some pandas operations can take a long time.

Ideally, when a new transform is added to the end of the transform list (in the frontend), only the new transform would have to be applied. (ie just transform `self._value`) However when a transform is removed, I guess all the transforms would need to be re-applied. Could be done by having the `UIElement` instance keep track of the history of transforms, and diffing with the transforms sent to it?

marimo/_plugins/ui/_impl/dataframes/handlers.py
```diff
@@ -0,0 +1,186 @@
...
+        return df.agg(dict_of_aggs)  # type: ignore[arg-type]
+
+
+def apply_transforms(
+    df_prev: "pd.DataFrame", transforms: Transformations
+) -> "pd.DataFrame":
+    if transforms.transforms is None or len(transforms.transforms) == 0:
+        return df_prev
+    # defensive copy
+    df = df_prev.copy()
```
akshayka, line 173: Can we remove this copy? Pandas operations aren't in-place by default, ie they already make a copy.

akshayka:
Having the preview together with the transforms is a great UX improvement, nice work

akshayka:
@mscolnick Seeing an issue when I groupby year:

output on right is correct, but output on left (rendered through the data-table plugin) is not 
image

```
Exception in callback BaseAsyncIOLoop._handle_events(23, 1)
handle: <Handle BaseAsyncIOLoop._handle_events(23, 1)>
  File "/opt/homebrew/Cellar/python@3.9/3.9.13_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/akshay/envs/marimo/lib/python3.9/site-packages/tornado-6.2-py3.9-macosx-12-arm64.egg/tornado/platform/asyncio.py", line 206, in _handle_events
    handler_func(fileobj, events)
  File "/Users/akshay/src/marimo-team/marimo/marimo/_server/sessions.py", line 294, in reader
    self.socket.write_op(op=op, data=data)
  File "/Users/akshay/src/marimo-team/marimo/marimo/_server/sessions.py", line 74, in write_op
    return self.write_message(
  File "/Users/akshay/envs/marimo/lib/python3.9/site-packages/tornado-6.2-py3.9-macosx-12-arm64.egg/tornado/websocket.py", line 338, in write_message
    message = tornado.escape.json_encode(message)
  File "/Users/akshay/envs/marimo/lib/python3.9/site-packages/tornado-6.2-py3.9-macosx-12-arm64.egg/tornado/escape.py", line 75, in json_encode
    return json.dumps(value).replace("</", "<\\/")
  File "/opt/homebrew/Cellar/python@3.9/3.9.13_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Timestamp is not JSON serializable
```

akshayka:
Another issue similar to the one I found above. I changed the second group by from name to mpg, but the output on the left didn't update

image

marimo/_plugins/ui/_impl/dataframes/handlers.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        Applies the given transformations to the dataframe.
+        """
+        # If differs from the original dataframe, then we need to start from
+        # the original dataframe.
+        if df is not self._original_df:
+            self._original_df = df
+            self._snapshot_df = apply_transforms(df, transform)
+            self._transforms = transform.transforms
+            return self._snapshot_df
```
akshayka, line 213: Can this actually happen, or is this just defensive? If it can't happen, should it be an error? Wondering if it might silently mask bugs in the future.

marimo/_plugins/ui/_impl/dataframes/handlers.py
```diff
@@ -0,0 +1,261 @@
...
+        else:
+            self._snapshot_df = apply_transforms(self._original_df, transform)
+            self._transforms = transform.transforms
+            return self._snapshot_df
+
+    def _is_superset(self, transforms: Transformations) -> bool:
+        """
+        Checks if the new transformations are a superset of the existing ones.
+        """
+        if len(self._transforms) == 0:
```
akshayka, line 236: nit: Python convention is to just write

```python
if not self._transforms:
   ...
```

Instead of explicitly checking length.

https://peps.python.org/pep-0008/

image

marimo/_plugins/ui/_impl/dataframes/handlers.py
```diff
@@ -0,0 +1,261 @@
...
+    raise AssertionError(f"Unhandled value: {value} ({type(value).__name__})")
+
+
+def _coerce_value(dtype: Any, value: Any) -> Any:
+    import numpy as np
+
+    return np.array([value]).astype(dtype)[0]
+
+
+class TransformsContainer:
```
akshayka, line 189: Nice!

akshayka (commented):
Some small comments on the Python code. Some correctness comments left on the PR discussion.

marimo/_plugins/ui/_impl/dataframes/handlers.py
mscolnick: can we enforce this with linting?

marimo/_plugins/ui/_impl/dataframes/handlers.py
mscolnick: yea i can throw an error instead

marimo/_plugins/ui/_impl/dataframes/handlers.py
```diff
@@ -0,0 +1,261 @@
...
+        return df.agg(dict_of_aggs)  # type: ignore[arg-type]
+
+
+def apply_transforms(
+    df_prev: "pd.DataFrame", transforms: Transformations
+) -> "pd.DataFrame":
+    if transforms.transforms is None or len(transforms.transforms) == 0:
+        return df_prev
+    # make a new reference
+    df = df_prev
```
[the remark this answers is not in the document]
mscolnick: yea this was leftover since i thought i was mutating the dataframes, ill clean it up

marimo/_plugins/ui/_impl/dataframes/handlers.py
akshayka: There might be a rule in ruff, but I couldn't find it.

frontend/src/plugins/impl/data-frames/DataFramePlugin.tsx
```diff
@@ -109,6 +113,11 @@ export const DataFrameComponent = ({
           <CodePanel dataframeName={dataframeName} transforms={value} />
         </TabsContent>
       </Tabs>
+      {error && (
+        <div className="text-error border-[var(--red-6)] bg-[var(--red-2)] text-sm p-2 border">
+          {prettyError(error)}
+        </div>
+      )}
```
akshayka, line 120: I don't see this being rendered when I enter an invalid transform.

image

akshayka:
Thanks for adding the error alert + popup. It's great!

@mscolnick found another issue. When I remove a filter, the preview isn't updated. See video 👇

[filter_not_removed.webm](https://github.com/marimo-team/marimo/assets/1994308/dc15861c-e26a-4f38-a4f9-db983cf59332)
.

akshayka:
@mscolnick Another UX issue with editing filters. Switched from a filter on `object` type to `float64` type and got this error:

image

akshayka:
@mscolnick Some more thoughts on filter.

- When a filter value is empty (even if column name and operator are set), the preview should show the unfiltered (full) data. Otherwise it's hard for user to know what values to enter. Right now if you enter a value but then delete it, the filtered data is still shown.
- For string/object filter with equality, should we have a dropdown of possible options, instead of free-form text? ou'd be able to select an option or unselect it (ie unset the filter). This would make it easier for users to filter (they'd know the universe of options), and would make it possible for them to temporarily unset the filter / make it possible to distinguish between the empty string and an unset filter. Would require an RPC, or passing that information along at plugin creation time. I think it would make the UX nicer.

akshayka:
@mscolnick

- transform request: select columns. Very common operation. You might take a subset of columns before doing a groupby/agg, for example.

In code, it's `df[["column_one", "column_two"]]`.

In particular, for the cars dataset, I'd like to get the mean MPG per origin. Not sure how to do that with the operations that are currently exposed. With select, would be `select MPG`, `group by origin agg mean`

- in `aggregation`: should there be an option to select all columns?

akshayka:
@mscolnick When you apply a numerical agg function like `mean` to a grouped dataframe, I think Pandas drops non-numeric columns before doing the agg, so there's no error:

image

But our plugin doesn't do that and instead raises an error. Maybe we should just match Pandas behavior and drop non-numeric columns first?

mscolnick:
@akshayka 

```
cars.groupby("Origin").agg("mean")
```

fails for me either way

docs/api/inputs/dataframe.md
```diff
@@ -0,0 +1,28 @@
+# Dataframe
+
+The dataframe UI element will output a visual editor to apply "transforms" to a dataframe. The resulting dataframe will be show below the transform editor. The UI output also includes the generated Python used to generate the resulting dataframe. You can programmatically access the resulting dataframe by accessing the element's `.value` attribute.
```
akshayka, line 3: Fixed some typos.

"The dataframe UI element outputs a visual editor to apply "transforms" to a dataframe, such as filtering rows, applying group-bys and aggregations, and more. The transformed dataframe is shown below the transform editor. The UI output also includes the generated Python used to generate the resulting dataframe, which you can copy paste into a cell. You can programmatically access the resulting dataframe by accessing the element's `.value` attribute."

akshayka (dismissed):

PR #227, opened 2023-10-19
Remote Function Registry + ui.table download function

akshayka:
This change adds a function registry for RPCs to the backend, adding the following:
- namespaced registry
- server route for invoking a function; request includes a unique function call ID/token so frontend can match return values to invocations
- websocket op for returning a result or indicating an error status
- functions are not specialized to UI elements, any Python object can register an RPC, including stateless plugins

Additionally, this PR adds a function for `mo.ui.table` for downloading the data as CSV/JSON.

Comments/room for improvement:
- For UI elements, the namespace is the object ID
- This isn't made explicit to the frontend, so the abstraction is a little leaky

### DIFF
--- a/marimo/_messaging/ops.py
+++ b/marimo/_messaging/ops.py
@@ -193,6 +193,16 @@ class CellOp(Op):
 
 
 @dataclass
+class FunctionCallResult(Op):
+    """Result of calling a function."""
+
+    function_call_id: str
+    return_value: JSONType
+    # True if the function call succeeded, False if it was aborted
+    success: bool
+
+
+@dataclass
 class RemoveUIElements(Op):
     """Invalidate UI elements for a given cell."""
 
--- a/marimo/_messaging/ops.py
+++ b/marimo/_messaging/ops.py
@@ -196,6 +196,8 @@ class CellOp(Op):
 class FunctionCallResult(Op):
     """Result of calling a function."""
 
+    name: ClassVar[str] = "function-call-result"
+
     function_call_id: str
     return_value: JSONType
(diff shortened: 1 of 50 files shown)


marimo/_utils/parse_dataclass.py
```diff
@@ -38,12 +38,12 @@ def _build_value(value: Any, cls: Type[T]) -> T:
             }
```
akshayka, line 38: Moved this file to get rid of a circular dependency

marimo/_messaging/ops.py
```diff
@@ -192,6 +192,16 @@ def broadcast_error(
         ).broadcast()
 
 
+@dataclass
+class FunctionCallResult(Op):
+    """Result of calling a function."""
+
+    function_call_id: str
+    return_value: JSONType
```
mscolnick, line 200: i wonder if we should allow returning a structured error. something that is human/UI friendly:
```
{
  title: "",
  message: "",
}
```

and up to the component to figured out how it wants to display it

marimo/_runtime/runtime.py
```diff
@@ -886,6 +889,36 @@ def get_ui_initial_value(self, object_id: str) -> Any:
...
+        else:
+            with self._install_execution_context(cell_id=function.cell_id):
+                try:
+                    return True, function(request.args)
+                except MarimoInterrupt:
+                    LOGGER.debug(
+                        "Function call (%s) was interrupted by the user",
+                        request,
+                    )
+                except Exception as e:
```
mscolnick, line 913: i think it would be valuable to return these errors in the response

PR #232, opened 2023-10-19
improvement: add row headers

mscolnick:
Screenshot 2023-10-19 at 4 54 31 PM


akshayka (commented):
Awesome, looks great. Only thing outstanding is special casing initial state for `pagination` and `page_size`

akshayka (approved):

PR #233, opened 2023-10-20
Utility class for fn that take 0 args

akshayka:
Adds a utility class to for RPCs that take 0 args.

### DIFF
--- a/marimo/_runtime/functions.py
+++ b/marimo/_runtime/functions.py
@@ -14,6 +14,12 @@ T = TypeVar("T")
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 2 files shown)


marimo/_runtime/functions.py
```diff
@@ -13,6 +13,12 @@
 T = TypeVar("T")
 
 
+@dataclasses.dataclass
+class EmptyArgs:
```
akshayka, line 17: Function using this could look like

```python
def rpc(self, args: EmptyArgs):
  del args  # unused
   # ...
```

mscolnick (approved):

PR #234, opened 2023-10-20
feat: mo.video

mscolnick:


marimo/_plugins/stateless/video.py
```diff
@@ -0,0 +1,74 @@
...
+    rounded: bool = False,
+    style: Optional[dict[str, Any]] = None,
+) -> Html:
+    """Render an video as HTML.
+
+    **Example.**
+
+    ```python3
+    mo.video(
+        src="https://www.youtube.com/watch?v=dQw4w9WgXcQ",
```
akshayka, line 31: lol.

Youtube videos don't work with `<video>` tags though, you have to embed them in an iframe and change `watch?v=` to `embed/` in the URL

- Could automagically handle youtube videos. For the user that'd be nice. But there's a risk that YouTube changes their URL structure, embedding policy, etc.
- Could raise a warning if URL is for YouTube (since that'll be a common use case).
- Could just fail silently (like right now)

marimo/_plugins/stateless/video.py
```diff
@@ -0,0 +1,74 @@
...
+        controls=False,
+    )
+    ```
+
+    **Args.**
```
akshayka, line 36: Add boolean `loop`?

marimo/_plugins/stateless/video.py
```diff
@@ -0,0 +1,74 @@
...
+
+    - `src`: the URL of the video or a file-like object
+    - `controls`: whether to show the controls
+    - `muted`: whether to mute the video
+    - `autoplay`: whether to autoplay the video.
+        the video will only autoplay if `muted` is `True`
+    - `width`: the width of the video
+    - `height`: the height of the video
+    - `rounded`: whether to round the corners of the video
+    - `style`: a dictionary of CSS styles to apply to the video
```
akshayka, line 46: I'd suggest not exposing `style` at the moment. Just in case we decide to have default styles to make this plugin render consistently across browsers, like our other plugins.

marimo/_plugins/stateless/video.py
```diff
@@ -0,0 +1,74 @@
...
+    **Returns.**
+
+    `Html` object
+    """
+    # Convert to bytes right away since can only be read once
+    if isinstance(src, io.BufferedReader):
+        src.seek(0)
+        src = src.read()
+
+    resolved_src = io_to_data_url(src, fallback_mime_type="video/png")
```
akshayka, line 57: Is `video/png` a valid mime type?

marimo/_plugins/stateless/video.py
mscolnick: i dont think we should do any magic. i just changed the src

marimo/_plugins/stateless/video.py
mscolnick: nope, copy paste

akshayka (approved):

PR #240, opened 2023-10-23
fix: autocomplete regressions

akshayka:
- fixes a bug which prevented user from being able to select (eg copy) text in a completion window
  - blur logic for closing completion window already handled by cell component
- fix a bug in which kernel stopped sending "no completions" op
- make tooltips appear above code completion

### DIFF
--- a/frontend/src/core/codemirror/completion/hints.ts
+++ b/frontend/src/core/codemirror/completion/hints.ts
@@ -75,7 +75,8 @@ export function hintTooltip() {
     // Clear tooltips on blur
     EditorView.domEventObservers({
       blur: (event, view) => {
-        closeCompletion(view);
+        // Only close tooltip, not view; blur for completion handled by
+        // cell editor, so that completion text is selectable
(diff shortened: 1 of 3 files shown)


frontend/src/core/codemirror/completion/hints.ts
```diff
@@ -61,10 +60,6 @@ export function hintTooltip() {
           message: result,
           exactName: fullWord,
         });
-        // Close the completion tooltips
```
akshayka, line 64: just show tooltip above code completion using CSS

frontend/src/css/codemirror.css
```diff
@@ -55,6 +55,11 @@
   padding: 0.625rem 0.875rem;
 }
 
+/* Hover tooltips get highest priority in display. */
+.cm .cm-tooltip.cm-tooltip-hover {
+  z-index: 1000;
```
akshayka, line 60: feels hacky

mscolnick (approved):

PR #244, opened 2023-10-23
Fix priority of closeBracketsKeymap

akshayka:
- it was previously getting overriden by deleteCharBackward


frontend/src/core/codemirror/keymaps/keymaps.ts
```diff
@@ -17,6 +18,7 @@ export function keymapBundle(
   switch (config.preset) {
     case "default":
       return [
+        keymap.of(closeBracketsKeymap),
```
mscolnick, line 21: you can also do `Prec.high` on this, or `Prec.low` on the others

mscolnick: its less fragile than ordering

mscolnick (dismissed):

PR #247, opened 2023-10-25
Dataframe styling adjustments

akshayka:


mscolnick (approved):

mscolnick:
The groupby and aggregate can look a bit weird with flex-end

image
image

akshayka:
> The groupby and aggregate can look a bit weird with flex-end
> 
> image image

Thanks for catching this! I've fixed this.

- I was trying to make the trash icon for deleting a form row appear inline with the controls. I think it mostly works now.
- To make it work I changed the `FormMessage` to be a tooltip in one place. Not tied to using a tooltip, but otherwise I was finding it hard to align everything. It's possible I regressed something, could use your help if I did. 
- Finally, I could use your help in repositioning the Drop NA for GroupBy, if possible. It'd be nice if it showed up in the bottom left or the top left, not in line with columns and aggregations as it does now.

image

frontend/src/components/ui/form.tsx
```diff
@@ -176,6 +178,24 @@ const FormMessage = React.forwardRef<
...
+const FormMessageTooltip = ({ className }: { className: string }) => {
+  const { error } = useFormField();
+  const body = error && error?.message ? String(error?.message) : null;
+
+  if (!body) {
+    return null;
+  }
+
+  return (
+    <TooltipProvider>
```
akshayka, line 190: Not sure if this is the right place for the `Provider`. I was getting an error saying that `Tooltip` must be used in a provider. But come to think of it maybe that's because I was running in storybook

PR #263, opened 2023-10-26
dataframe: load values from backend and distaply in better dropdowns

mscolnick:
* add RPC to `mo.ui.dataframe` to load column values. this loads all values instead of at that particular transform or filter. this adds complexity and could be confusing to users.
* refactor the inputs in form.tsx to be more re-usable and use better inputs (selects, multi-select, etc)
* make panels the same height to avoid content shift
* improving some border styling

### DIFF
--- a/marimo/_plugins/ui/_impl/dataframes/dataframe.py
+++ b/marimo/_plugins/ui/_impl/dataframes/dataframe.py
@@ -5,6 +5,7 @@ import inspect
 import sys
 from typing import TYPE_CHECKING, Any, Callable, Dict, Final, List, Optional
 
+
 if TYPE_CHECKING:
     import pandas as pd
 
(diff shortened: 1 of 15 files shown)


marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -109,6 +120,29 @@ def get_dataframe(self, _args: EmptyArgs) -> GetDataFrameResponse:
             row_headers=get_row_headers(self._value),
         )
 
+    def get_column_values(
```
akshayka, line 123: Nice, very clean! Gives confidence in the RPC framework.

akshayka:
@mscolnick I merge ms/improve-select-inputs because it said merge when ready, but now I see that caused a merge conflict here. Sorry about that

akshayka (changes requested):
Found a bug that causes output to vanish, see video in PR thread

akshayka:
@mscolnick Here's the bug

[filter-in-bug.webm](https://github.com/marimo-team/marimo/assets/1994308/79182289-2712-424c-b33f-fe0ded00fc26)

marimo/_plugins/ui/_impl/dataframes/dataframe.py
mscolnick: yep, great job here!

mscolnick:
@akshayka, thanks, i knew exactly where to fix it. i saw it before and then regressed it in my final cleanup. should be good for another reviews, thanks!

akshayka (approved):

PR #264, opened 2023-10-26
improve dark mode

mscolnick:
in the next PR, ill move `theme: "light" | "dark"` to UserConfig


akshayka (approved):
Looks fantastic! I think it's a coincidence but dark mode looks vaguely Halloween-esque, I think it's very festive ha

PR #267, opened 2023-10-27
feat: add outline as a panel while in edit-mode

mscolnick:
* this is just in editor mode, until we get more feedback in app-mode / how this could be handled (app-config, always, `mo.toc()`, etc)
* we don't handle duplicate headers between different cells, we just go to the first header. i think this is ok as we likely want the user to define unique headers. i'd propose we can handle this based on feedback

Screenshot 2023-10-27 at 1 09 23 PM


akshayka:
Bug in scrolling to heading in long documents

[scroll-to-heading-bug.webm](https://github.com/marimo-team/marimo/assets/1994308/44538bf0-c4f7-4048-8064-82294fb0cef4)

akshayka (approved):
Conditionally approved modulo nested headings with same title --- nice work!!

PR #268, opened 2023-10-27
Add scatter to smoke test

akshayka:
-100k is slow with select

@mscolnick included a scatter plot in the charts smoke test. it's not really usable at 100k points right now -- haven't looked into where the slowness is coming from


mscolnick:
just tested this same chart in the vega editor, so the slowness is due to the rendering. we do render on canvas, so could be the redrawing of the canvas with each selection. worth exploring https://github.com/hex-inc/vegafusion for 100k+

PR #269, opened 2023-10-27
feat: add BrowserConfig to UserConfig

akshayka:
Let user specify which web browser to use.

Implements the feature requested in #265

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -89,6 +89,18 @@ class DisplayConfig(TypedDict, total=False):
     theme: Literal["light", "dark"]
 
 
+class BrowserConfig(TypedDict, total=False):
+    """Settings related to web browsers.
+
+    **Keys.**
+
+    - `browser`: `"default"` or a browser registered with Python's
+        webbrowser module (e.g., `"firefox"` or `"google-chrome"`)
+    """
+
+    browser: Literal["default"] | str
+
+
 @mddoc
 class MarimoConfig(TypedDict, total=False):
     """Configuration for the marimo editor.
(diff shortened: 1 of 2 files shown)


marimo/_config/config.py
```diff
@@ -89,6 +89,18 @@ class DisplayConfig(TypedDict, total=False):
...
+class BrowserConfig(TypedDict, total=False):
+    """Settings related to web browsers.
+
+    **Keys.**
+
+    - `browser`: `"default"` or a browser registered with Python's
+        webbrowser module (e.g., `"firefox"` or `"google-chrome"`)
+    """
+
+    browser: Literal["default"] | str
```
akshayka, line 101: List of supported values is here: https://docs.python.org/3/library/webbrowser.html#webbrowser.register. Figured it was better to just type as `str` rather than have to maintain exhaustive `Literal` type.

mscolnick, line 101: I wonder if this could go in runtime config? Or maybe we rename it to launch config?

akshayka: How about [server] instead? We'll likely add other server config, such as the port

mscolnick: Yea I like that

marimo/_config/config.py
```diff
@@ -83,12 +83,24 @@ class DisplayConfig(TypedDict, total=False):
...
     **Keys.**
 
-    - `theme`: one of `"light"` or `"dark"`
+    - `theme`: `"light"` or `"dark"`
     """
 
     theme: Literal["light", "dark"]
 
 
+class ServerConfig(TypedDict, total=False):
```
akshayka, line 92: @mscolnick I renamed this

mscolnick (approved):
can you confirm the frontend doesn't lose this configuration when using the Settings UI? I'm not sure if the read-then-write is lossy on a new top level config section

PR #271, opened 2023-10-28
Update code editor to clear tooltips on 'Escape' key press, adjust code cell active line contrast in dark mode and optimize tooltip height

mscolnick:
* clear tooltip on Esc and don't lose focus
* improve dark mode contrast
* make tooltip height responsive, With This One Neat Trick™️

### DIFF
--- a/frontend/src/css/codemirror.css
+++ b/frontend/src/css/codemirror.css
@@ -1,12 +1,8 @@
-.cm-selectionBackground {
+.light .cm-selectionBackground {
   /* This is the background for selected text when focused,
   but we want to use it when not in focus either because we
   will often lose focus during find/replace or in another editor */
   background-color: #d7d4f0 !important;
-}
-
-.dark .cm-selectionBackground {
-  background-color: #4d4d4d !important;
 }
 
 /* -- Gutters -- */
@@ -30,7 +26,6 @@
 /* -- Tooltips: code completion and type hints -- */
 
 .cm .cm-tooltip {
-  font-family: var(--monospace-font);
   font-size: 0.875rem;
   border: none;
   box-shadow: var(--light-shadow);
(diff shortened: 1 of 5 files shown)


frontend/src/css/codemirror.css
```diff
@@ -30,7 +26,6 @@
 /* -- Tooltips: code completion and type hints -- */
 
 .cm .cm-tooltip {
-  font-family: var(--monospace-font);
```
mscolnick, line 33: i find this MUCH easier to read and adds more content to the same area of tooltip

akshayka:
@mscolnick with/without monospace I still see formatting (ragged spacing) like your first screenshot:

image

Somehow the function docstring (not autocomplete) has better spacing for this example, but that's true with and without monospace:

your branch:
image

on main:
image


I believe the spacing you're trying to fix is due to `white-space: pre-wrap`, not the monospace font. If you remove `white-space: pre-wrap` some docstrings look better (like `md`), but others become totally illegible (`np.array`), so we can't remove that unfortunately.

frontend/src/css/codemirror.css
akshayka: nice, i do think it's easier to read. only thing is that the prose and code look to be different sizes (not sure if they actually are), with the code looking bigger — possible to shrink the code size or increase the prose size?

akshayka: I'm a little split actually on which font style I prefer

mscolnick:
im not sure what ragged spacing you mean, but i think its more legible/easier to read when not everything is monospace and its more similar to vscode and other IDEs (not sure why we would want to differ in that)

mscolnick:
oh is it the new lines? i wasnt trying to fix that but that is actually just a python docstring issue. we should definitely fix that (its a weird assumption that the FE should know the BE only uses a certain width of characters). ill revert this font change in the meantime, and create a ticket

frontend/src/css/codemirror.css
mscolnick: if you are split, i think we should get rid of it. i think it looks much better and similar to existing IDEs. but there is a bug in the python docstring -> markdown in that it creates some extra new lines. ill revert for now

mscolnick:
for reference, this is what vscode does and is much more legible

image

mscolnick:
sorry i was looking at `mo.ui.array` instead of `np.array`. ill revert the font changes for now

akshayka (approved):

PR #276, opened 2023-10-30
fix: fix stale values in DictPlugin and allow re-submitting in FormPlugin

mscolnick:


frontend/src/plugins/impl/DictPlugin.tsx
```diff
@@ -64,12 +64,11 @@ const Dict = ({
         // If the UI element firing the event is not in the dict, ignore it
         return;
       }
-
-      // Tell Python to update the dict at the given key with the new value
-      setValue(() => {
-        const updates: Record<string, unknown> = {};
-        updates[key] = e.detail.value;
-        return updates;
```
mscolnick, line 72: this was a bug that lost previous values

frontend/src/plugins/core/registerReactComponent.tsx
```diff
@@ -144,13 +118,15 @@ function PluginSlotInternal<T>(
...
-    },
-    [setValue, shouldDispatchInput]
-  );
+  // When the value changes, send an input event
+  const setValueAndSendInput = useEvent((value: SetStateAction<T>): void => {
+    setValue((prevValue) => {
+      const updater = Functions.asUpdater(value);
+      const nextValue = updater(prevValue);
+      hostElement.dispatchEvent(createInputEvent(nextValue, hostElement));
+      return nextValue;
```
mscolnick, line 127: by dispatching here (and instead of the next render loop), we are able to resubmit since React otherwise skips re-renders when `setValue` is `===` to the previous value

akshayka (approved):

marimo/_plugins/ui/_impl/array.py
```diff
@@ -96,7 +96,10 @@ def elements(self) -> Sequence[UIElement[JSONType, object]]:
     def _convert_value(self, value: dict[str, JSONType]) -> Sequence[object]:
         if self._initialized:
             for k, v in value.items():
-                self._elements[int(k)]._update(v)
+                element = self._elements[int(k)]
+                # only call update if the value has changed
+                if element and element._value != v:
+                    element._update(v)
```
mscolnick, line 102: @akshayka, one of the bugs that was fixed (sending only partial of the dictionary plugin) surfaced another bug which was calling `._update` on all elements of `mo.ui.array` (or `mo.ui.dict`) which would call the `on_change` handler. i noticed this with this code which would call `on_click` when it wasnt really clicked

```python
    dict = mo.ui.dictionary(
        {
            "slider": mo.ui.slider(1, 10),
            "text": mo.ui.text("type something!"),
            "array": mo.ui.array(
                [
                    mo.ui.button(value=0, on_click=lambda v: v + 1)
                    for _ in range(3)
                ],
                label="buttons",
            ),
        }
    )
```

PR #281, opened 2023-10-30
websocket smoke test

mscolnick:


akshayka (approved):
LGTM. If it's not too much trouble, could you include a markdown string in the smoke test with like two or three sentences of context of what the desired behavior is?

Issue #260, opened 2023-10-26
Plot blinking: plotly&matplotlib interactive


proItheus:
### Describe the bug

I suspect that's very much the same as #251, but maybe not for `mpl.interactive`?

Besides blinking, `mpl.interactive` will also use a new port every time I drag the slider, and stop responding after a few changes.

![Screenshot_20231026_154844](https://github.com/marimo-team/marimo/assets/50539150/a11f54b9-af6d-46a1-84f5-c1afc4b10fb6)

### Version

0.1.38

### Code to reproduce

```python
import marimo

__generated_with = "0.1.38"
app = marimo.App()


@app.cell
def __(exp):
    exp
    return


@app.cell
def __(altp, exp, mo, mplplot, plotlyplot, snsplot):
    mo.vstack(
        [
            mo.md(f"$$y=x^{exp.value}$$"),
            mo.tabs(
                {
                    "altair": mo.vstack([altp, altp.value]),
                    "plotly": plotlyplot(exp.value),
                    "seaborn": snsplot(exp.value),
                    "matplotlib": mplplot(exp.value),
                }
            ),
        ]
    )
    return


@app.cell
def __():
    import marimo as mo
    from plotly import express as px
    import altair as alt
    import numpy as np
    import pandas as pd
    import seaborn as sns
    from matplotlib import pyplot as plt
    from functools import cache
    return alt, cache, mo, np, pd, plt, px, sns


@app.cell
def __(mo):
    exp = mo.ui.slider(1, 8, 1, label="exp: ")
    return exp,


@app.cell
def __(cache, np):
    @cache
    def prep_xy(exp):
        x = np.linspace(0, 5, 100)
        return x, x**exp
    return prep_xy,


@app.cell
def __(alt, cache, mo, pd, prep_xy):
    @cache
    def altplot(exp):
        x, y = prep_xy(exp)
        data = pd.DataFrame({"x": x, "y": y})
        return mo.ui.altair_chart(
            alt.Chart(data).mark_line().encode(x="x", y="y"),
            chart_selection="interval",
        )
    return altplot,


@app.cell
def __(cache, prep_xy, px):
    @cache
    def plotlyplot(exp):
        x, y = prep_xy(exp)
        return px.line(x=x, y=y)
    return plotlyplot,


@app.cell
def __(cache, prep_xy, sns):
    @cache
    def snsplot(exp):
        x, y = prep_xy(exp)
        return sns.lineplot(x=x, y=y)
    return snsplot,


[...]

PR #282, opened 2023-10-31
feat: add marimo-plotly plugin to reduce flashing, improve performance, and reactivity

mscolnick:
Partial fix for https://github.com/marimo-team/marimo/issues/260

Screenshot 2023-10-30 at 9 14 35 PM


frontend/src/plugins/impl/plotly/PlotlyPlugin.tsx
```diff
@@ -0,0 +1,55 @@
...
+  host: HTMLElement;
+}
+
+export const PlotlyComponent = ({ figure, host }: PlotlyPluginProps) => {
+  return (
+    <Plot
+      {...figure}
+      layout={{
+        ...figure.layout,
+        autosize: true,
```
akshayka, line 42: Can the user still manually specify a width and height in Python, or will this override it? Sorry, I'm not familiar with the Plotly.js API. Sometimes manually specified width/height are important (eg, if the data you're plotting requires a certain aspect ratio)

akshayka (approved):
Left 1 comment, just want to double check users can specify their own width/height for the plot. From your smoke test it seems like the answer is "yes", but I haven't run it so just wanted to double check.

Otherwise LGTM, thanks!

frontend/src/plugins/impl/plotly/PlotlyPlugin.tsx
mscolnick: the docs says that this is only applied if the height, width is `undefined`. ill double check

PR #288, opened 2023-11-01
refactor notebook state to better support collaboration

mscolnick:
* split up `CellState` into `CellData` and `CellRuntime`
* move utils and functions into cells.ts
* update tests


akshayka (approved):
Looks great!

PR #289, opened 2023-11-01
fix ui.array updating logic: check against frontend value

akshayka:
Fixes array updates to check value frontend sent against last value received from frontend (instead of against Python value). Fixes a bug in which `mo.ui.array()` couldn't hold `mo.ui.dataframe` and other complex types.


mscolnick (commented):
should we write a test for this given I regressed this unknowingly?

mscolnick (dismissed):

mscolnick (approved):

PR #295, opened 2023-11-03
Dispatch resize event when config width changes 

mscolnick:
this makes charts/plots update their size


akshayka (dismissed):
LGTM except lint

PR #303, opened 2023-11-04
experimental: start of static html

mscolnick:
Initial implementation of static HTML. This currently does not handle virtual files or function calls. This is behind an experimental feature-flag `staticExport`

In order to test:
1. change `ENABLE_LOCAL_ASSETS` to `true`
2. add `static_export = True` to `~/marimo.toml`
3. good example is `marimo edit marimo/_smoke_tests/layout.py`


marimo/_server/server.py
```diff
@@ -136,6 +136,13 @@ def post(self) -> None:
         shutdown()
 
 
+class CorsStaticFileHandler(tornado.web.StaticFileHandler):
+    def set_default_headers(self) -> None:
+        self.set_header("Access-Control-Allow-Origin", "*")
+        self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+        self.set_header("Access-Control-Allow-Methods", "POST, GET, OPTIONS")
```
akshayka, line 143: just for my own knowledge, why was this needed?

akshayka:
FYI I think the latest frontend is not on npm yet? When I try a simple app (`mo.md("# hello")`), I see a blank page and the following errors in the console:

image

akshayka (approved):
nice work! this looks like it was challenging than I thought it would be

PR #315, opened 2023-11-07
feat: dependency graph

mscolnick:
Adds a panel for a vertical dependency tree. the edges are along the side only when a node it clicked. 
This is using the library https://reactflow.dev

Some alternatives:
* add a button for more of a tree view (TB or LR). Could additional focus on an element so there is an inverted-tree above and a tree below
* add more controls for zoom/pan

### DIFF
--- a/frontend/e2e-tests/layout-grid.spec.ts
+++ b/frontend/e2e-tests/layout-grid.spec.ts
@@ -14,7 +14,7 @@ test("can run Grid layout", async ({ page }) => {
   await expect(page.getByText("Grid Layout")).toBeVisible();
 
   // Type in search box
-  await page.getByRole("textbox").first().fill("hello");
+  await page.getByRole("textbox").last().fill("hello");
 
   // Verify dependent output updated
   await expect(page.getByText("Searching hello")).toBeVisible();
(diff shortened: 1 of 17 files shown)


akshayka:
Two comments on scrolling.

1. I noticed I can scroll past the start and end of the graph. Possible to stop scrolling once you reach the start/end? See video.
2. The scrollable area seems to be only on top of the array of cells. If my mouse is in the panel but not on top of a cell, it doesn't scroll. It would be nice if there were a buffer around the panel that reacted to scroll.

[Screencast from 11-08-2023 03:06:24 PM.webm](https://github.com/marimo-team/marimo/assets/1994308/2a902dd4-7350-461f-980b-fce3c31adcd7)

akshayka:
Design wise I think this is great for the first release! As we iterate on the design in the future, here are some things to consider:

1. Can we communicate why an edge exists? What are the refs/defs flowing on each edge?
2. When the user clicks on a cell in the main editor, if the graph panel is open, should we scroll to and focus the corresponding cell in the graph panel?
3. Pivoting -- users might want to pivot between app sort (default) and topological sort
4. Cell-specific views  --- how can a user "drill-down" or narrow focus to a single cell? This narrowed focus would show only the ancestors and descendants of the selected cell (possibly supporting pivoting between app sort and topological sort).

akshayka (approved):
I think this is great! LGTM except for two points about scrolling (see PR conversation) -- if it's not much effort, would prefer fixing those before merging. If it's too much effort can leave for later.

I also left a comment for things to consider as we iterate on the design in the future.

PR #316, opened 2023-11-07
fix: exclude tooltip completion type in hover

mscolnick:


frontend/src/core/codemirror/completion/hints.ts
```diff
@@ -39,6 +39,7 @@ export function hintTooltip() {
           position: endToken,
           message: result,
           exactName: fullWord,
+          excludeTypes: ["tooltip"],
```
mscolnick, line 42: this is the main change

akshayka (approved):

PR #320, opened 2023-11-08
Fix: Don't Delete Cached Virtual Files

akshayka:
This PR includes an important fix to the management of virtual file lifecycles.

Previously, all of a cell's virtual files were deleted at the end of its lifecycle. This broke cached HTML that referenced those virtual files (e.g., with a `@functools.cache` decorator).

This change adds reference counting to virtual files, to prevent them from being deleted when objects exist referencing them. We assume that only HTML objects will reference virtual files, and we implement the ref-counting by string matching on HTML text: on object creation, we search the text to find its virtual files and increment their refcounts, on deletion we decrement these refcounts. String matching was the simplest solution we could find that is compatible with using f-string interpolation of HTML text into other HTML text, which is common in marimo (e.g., f-strings with `mo.md`).

This change also gives `CellLifecycleItems` the option to persist into the next lifecycle. This is needed for virtual files, since when theirt refcount is greater than 0 we can't delete them, but instead must try to delete them in the next lifecycle.

### DIFF
--- a/marimo/_output/hypertext.py
+++ b/marimo/_output/hypertext.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Literal, final
 
 from marimo._output.mime import MIME
 from marimo._output.rich_help import mddoc
+from marimo._output.utils import flatten_string
 
 if TYPE_CHECKING:
     from marimo._plugins.core.web_component import JSONType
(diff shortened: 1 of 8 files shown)


marimo/_output/hypertext.py
```diff
@@ -51,7 +52,57 @@ class Html(MIME):
...
+
+        # Virtual File Refcounting
+        #
+        # HTML elements are responsible for maintaining the reference counts
+        # of virtual files: virtual files cannot be disposed when HTML elements
+        # reference them. For example, a user might cache HTML referencing a
+        # virtual file if they create it using functools.cache.
+        #
+        # flatten the text to make sure searching isn't broken by newlines
+        flat_text = flatten_string(self._text)
```
mscolnick, line 81: for `mo.ui.altair_chart` - it may have a virtual file in the `data` which gets base64 encoded. could this make `virtual_filename in flat_text` fail?

akshayka: I've checked manually and the file is there in plain text, here's a snippet of what's there:

`{&quot;url&quot;: &quot;/@file/3726-108546-7FiyZh2K.csv&quot;, `

akshayka: Filename is alphanumeric plus dashes and period. We only check the basename (without the bytes), in this case we'd be checking the string `108546-7FiyZh2K.csv`. So it should be fine?

mscolnick (approved):

PR #338, opened 2023-11-13
Add envinfo command to CLI for debugging purposes

mscolnick:


akshayka (commented):
Pretty cool!

Can we change the name from `envinfo` to just `env`, for short?

FYI `pip` also has `pip inspect`. That dumps a lot more stuff though, and I suppose doesn't include things we care about like browser versions.

mscolnick:
@akshayka, yea it has a lot more info that we don't need and missing some others. and ours may change over time

marimo/_cli/envinfo.py
```diff
@@ -0,0 +1,139 @@
...
+        "OS Version": platform.release(),
+        "Python Version": platform.python_version(),
+    }
+
+    binaries = {
+        "Chrome": _get_chrome_version() or "--",
+        "Node": _get_node_version() or "--",
+    }
+
+    requirements = _get_pip_list()
```
akshayka, line 26: This will work for now, but eventually we'll be on `conda`, at which point we might add a flag that controls whether this command looks at `conda` or `pip`. Just pointing it out, no action needed

PR #341, opened 2023-11-13
Fix: truncate dataframes rows, columns in output

### LINKED ISSUE #339: Showing Pandas DataFrame crashes marimo
### Describe the bug

When I display a DataFrame with a lot of rows, e.g. [California housing dataset](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_california_housing.html) directly, marimo is getting really unresponsive. 

I would expect to add a pagination or show the first and last few rows (like Jupyter does). 

If you can point me in the right direction, I'm happy to help!

### Version

0.1.47

### Code to reproduce

```python
import marimo
 
__generated_with = "0.1.47"
app = marimo.App()
 
 
@app.cell
def __():
    from sklearn.datasets import fetch_california_housing
    return fetch_california_housing,
 
 
@app.cell
def __(fetch_california_housing):
    data = fetch_california_housing(as_frame=True)
    return data,
 
 
@app.cell
def __(data):
    data.get("data")
    return
```

akshayka:
Previously the formatter showed the entire dataframe, which can make the frontend slow and unresponsive. This change sets pandas options for `display.max_rows` and `display.max_columns` to 10 and 20.

Users can override these options using `pd.set_option` or the pandas option context manager. Or they can output the HTML themselves: `mo.Html(df.to_html())`.

image

Fixes #339

### DIFF
--- a/marimo/_output/formatters/pandas_formatters.py
+++ b/marimo/_output/formatters/pandas_formatters.py
@@ -15,20 +15,28 @@ class PandasFormatter(FormatterFactory):
     def register(self) -> None:
         import pandas as pd
 
+        pd.set_option("display.max_rows", 10)
+        pd.set_option("display.max_columns", 20)
+
         from marimo._output import formatting
 
         @formatting.formatter(pd.DataFrame)
         def _show_dataframe(df: pd.DataFrame) -> tuple[str, str]:
+            max_rows = pd.get_option("display.max_rows")
+            max_columns = pd.get_option("display.max_columns")
             # Flatten the HTML to avoid indentation issues when
             # interpolating into other HTML/Markdown with an f-string
             return (
                 "text/html",
-                flatten_string(df.to_html()),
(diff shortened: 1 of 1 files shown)


marimo/_output/formatters/pandas_formatters.py
```diff
@@ -15,20 +15,28 @@ def package_name() -> str:
     def register(self) -> None:
         import pandas as pd
 
+        pd.set_option("display.max_rows", 10)
+        pd.set_option("display.max_columns", 20)
```
mscolnick, line 19: this is just the new default right? and it is overridable by a user?

mscolnick (approved):

marimo/_output/formatters/pandas_formatters.py
akshayka: Yes it's overridable, just the new default.

PR #343, opened 2023-11-13
feat: Update cli.py run/edit from github or external files

mscolnick:
Update cli.py run/edit from github or external files

can do things like:
```
marimo edit https://github.com/marimo-team/marimo/issues/339
marimo edit https://raw.githubusercontent.com/marimo-team/marimo/main/examples/optimization/regularization_and_sparsity.py
marimo edit https://github.com/marimo-team/marimo/blob/main/examples/optimization/regularization_and_sparsity.py
```

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -15,6 +15,7 @@ import click
 from marimo import __version__, _loggers
 from marimo._ast import codegen
 from marimo._cli import ipynb_to_marimo
+from marimo._cli.file_path import validate_name
 from marimo._server.server import start_server
 from marimo._utils.url import is_url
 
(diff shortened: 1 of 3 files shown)


akshayka (commented):
This is really nice — nice tests too. Feels great to pass in a Github URL.

I updated the code to only make a temporary directory when needed.

Some thoughts about using temporary files for `edit`:
- It makes sense for `run`, since the user can't modify the app when using `run`
- But for `edit`, the user can modify the notebook. I can anticipate people editing, modifying, saving, then losing work.

One idea on how to improve the experience for `edit` from a URL:

- `marimo edit URL`
- [prompt at terminal] Please name your file: ...
- once user gives filename, then we start it up

Perhaps that's too much friction. Another idea:

- marimo edit URL` just chooses a filename by itself:
-  basename of the URL (if that's a Python file), uniquified with the standard `(number)` trick that operating systems do.
- In the case of a Github issue, maybe it's the issue title (uniquified), or maybe it's just a random string of words (`magical_colorful_unicorn.py`)

marimo/_cli/cli.py
```diff
@@ -184,9 +184,11 @@ def edit(
     name: Optional[str] = None,
 ) -> None:
     if name is not None:
-        temp_dir = tempfile.TemporaryDirectory()
         # Validate name, or download from URL
-        name = validate_name(name, allow_new_file=True, temp_dir=temp_dir)
+        # The second return value is an optional temporary directory. It is
```
akshayka, line 188: It turns out `delete=True/False` is only available in Python 3.12. In older versions the lifetime of the directory on disk is bound to the object (similar to the shared memory object), so I just changed `validate_name` to return the temporary directory to the caller, who keeps it around.

akshayka:
Left a comment about what happens with URLs and edit, but I think we can improve later, not urgent

PR #344, opened 2023-11-14
Add support for notebook logs and additional panel type in editor

mscolnick:
Screenshot 2023-11-14 at 10 09 48 AM

### DIFF
--- a/frontend/src/components/editor/chrome/panels/logs-panel.tsx
+++ b/frontend/src/components/editor/chrome/panels/logs-panel.tsx
@@ -70,10 +70,7 @@ function formatLog(log: CellLog) {
   const timestamp = formatLogTimestamp(log.timestamp);
 
   const color = levelColor[log.level];
-  let level = log.level.toUpperCase();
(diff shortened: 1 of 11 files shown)


akshayka (commented):
This is really nice. I find it helps with debugging when I'm in the "app view" mode — don't have to toggle back to editing to see console output. I could even see people using it to good effect in run mode, though let's design that later.

Just a couple of small changes this PR:
1. Change INFO to STDOUT
2. Change ERROR to STDERR

It's a little more clear that way. Otherwise funny things happen — for example, Python sends all `logging` messages (including INFO) to `STDERR` by default.

The log panel would be even more useful if each message had a cell hyperlink. But I guess that gets tricky because you can delete cells. Could do in a future PR if you think it makes sense.

mscolnick:
@akshayka, updated with your comments.

i can do the cell link in a followup. maybe ill toast if the cell is deleted

akshayka (approved):
🎉

PR #347, opened 2023-11-14
Update tooltip position to absolute

mscolnick:
currently there is a tradeoff, and I spent some time trying for no tradeoff, but was unable to find a solution.

when `position: absolute` is enabled, we can select text within the tooltip without losing focus of the editor.
when `position: fixed` is enabled, the autocomplete and hover tooltip take up a bit more space up to the edges thus utilizing more space.


akshayka (approved):
I think this is definitely an improvement! To be honest I couldn't notice the space utilization

PR #349, opened 2023-11-15
fix: date filtering/selection for mo.ui.altair_chart

mscolnick:


marimo/_plugins/ui/_impl/altair_chart.py
```diff
@@ -104,6 +103,17 @@ def _filter_dataframe(
     return df
 
 
+def _coerce_value(dtype: Any, value: Any) -> Any:
+    # If dtype is datetime[ns], then we need to convert the value
+    # from milliseconds (which is what vega returns for dates)
+    if dtype == "datetime64[ns]":
```
akshayka, line 109: @mscolnick Do we need to do this for other units as well, or is `ns` special?

akshayka (approved):

marimo/_plugins/ui/_impl/altair_chart.py
mscolnick: hmm, yea im not sure what the other units can be. but a quick scan shows `datetime64[ns, tz]` is also valid, so something i can cover as well

Issue #353, opened 2023-11-15
mo.ui.plotly: send tooltip back in selection


akshayka:
### Description

One of our users was using reactive plotly.

Their plot had some tooltips, and they would like the tooltips to be sent back with selection.

### Suggested solution

Include a "tooltip" key in the entries of `fig.value`

```
{"x": ..., "y": ..., "tooltip": ...}
```

### Additional context

The user chose plotly (and was very happy to see it supported!) because it gives more flexibility than altair — they were plotting a graph with nodes and edges, which altair/vega does not support as far as I can tell.

PR #357, opened 2023-11-15
Add indexes property to mo.ui.plotly, and update points to use original column names

mscolnick:
Fixes https://github.com/marimo-team/marimo/issues/353

* Add indexes property to mo.ui.plotly - this is a great escape hatch to filter an original dataset in case not all the data is passed to plotly)
* update points to use original column names, instead of `x`, `y`

### DIFF
--- a/marimo/_plugins/ui/_impl/plotly.py
+++ b/marimo/_plugins/ui/_impl/plotly.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
 #     "field1": [min, max],
 #     "field2": [min, max],
 #   },
-#  "indexes": int[],
+#  "indices": int[],
 # }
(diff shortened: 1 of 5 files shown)


marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -120,6 +121,14 @@ def points(self) -> List[Dict[str, Any]]:
             return []
         return self._selection_data["points"]  # type:ignore
 
+    @property
+    def indexes(self) -> List[int]:
```
akshayka, line 125: nit: let's make this "indices"

akshayka (approved):
LGTM except for renaming "indexes" to "indices"

PR #360, opened 2023-11-15
feat: Add version update check and print in 'marimo edit'

mscolnick:
image

### DIFF
--- a/marimo/_cli/upgrade.py
+++ b/marimo/_cli/upgrade.py
@@ -20,7 +20,7 @@ class MarimoCLIState:
     last_checked_at: Optional[str] = None
 
 
-def check_for_updates():
+def check_for_updates() -> None:
     try:
         _check_for_updates_internal()
     except Exception:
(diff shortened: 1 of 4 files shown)


marimo/_cli/upgrade.py
```diff
@@ -0,0 +1,119 @@
...
+    if not state.latest_version:
+        # We couldn't get the latest version, so do nothing
+        return
+
+    # Compare versions and warn if there's a new version
+    if current_version and version.parse(state.latest_version) > version.parse(
+        current_version
+    ):
+        #         Update available 0.1.35 -> 0.1.123.
+        # Run "fly version upgrade" to upgrade.
```
akshayka, line 63: Remove comment?

akshayka (commented):
Pretty robust solution, left a couple comments.

One thing. Once people start installing from conda, I wonder how we'll print a reliable upgrade message. Might be `conda update marimo` instead of `pip` ... maybe there are some hacks?

mscolnick:
> One thing. Once people start installing from conda, I wonder how we'll print a reliable upgrade message. Might be conda update marimo instead of pip ... maybe there are some hacks?

I think we can just say `pip install or conda install` in the message. and likely will have other reasons we ask for their preferred package manager and store in user config, then can use that

akshayka (approved):

PR #363, opened 2023-11-16
feat: Add new data frame transformations for shuffl and sample

mscolnick:


akshayka (commented):
For `sample`, can you add the option to toggle sampling with replacement (defaults to `False`)? Using the pandas argument `replace`

akshayka (approved):

PR #364, opened 2023-11-16
fix: flush empty outputs

akshayka:
This change fixes a bug that prevented `mo.spinner()` from being cleared on context manager exit (previously it was only cleared at the end of cell execution, by the cell's last expression -- but when running in a callback there is no "last expression").


mscolnick (dismissed):

akshayka:
@mscolnick FYI added a unit test. It's kind of messy

PR #371, opened 2023-11-17
fix: json parse to be more performant for non-edge cases

mscolnick:


frontend/src/utils/json/json-parser.ts
```diff
@@ -5,6 +5,14 @@
  * This also handles NaN, Infinity, and -Infinity.
  */
 export function jsonParseWithSpecialChar<T>(value: string): T {
+  // This regex handling is expensive and often not needed.
+  // We try to parse with JSON.parse first, and if that fails, we use the regex.
+  try {
+    return JSON.parse(value) as T;
```
mscolnick, line 11: @akshayka this is the only change that matter

akshayka (approved):

PR #374, opened 2023-11-18
docs: recipes

akshayka:
Add recipes for a variety of common tasks.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -1,11 +1,13 @@
 # Recipes
 
-This page includes code snippets or "**recipes**" for a variety of common tasks:
-use them as building blocks or examples when making your own notebooks.
+This page includes code snippets or "**recipes**" for a variety of common tasks.
+Use them as building blocks or examples when making your own notebooks.
 
-_In the recipes, **each code block represents a cell**._
+In these recipes, **each code block represents a cell**.
  
-## Show an output conditionally
+## Control Flow
+
+### Show an output conditionally
 
 **Task.** Only show an output if a condition is `True`.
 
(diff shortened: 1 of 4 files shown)


docs/recipes.md
```diff
@@ -0,0 +1,309 @@
...
+This page includes code snippets or "**recipes**" for a variety of common tasks.
+Use them as building blocks or examples when making your own notebooks.
+
+In these recipes, **each code block represents a cell**.
+ 
+## Control Flow
+
+### Show an output conditionally
+
+**Task.** Only show an output if a condition is `True`.
```
mscolnick, line 12: I think recipes should be more "use-case" oriented and include "why" you could do this:

"For example, you may want to wait for user input or lazily load data"

docs/recipes.md
```diff
@@ -0,0 +1,309 @@
...
+
+1. Use an `if` expression to choose which output to show.
+
+```python
+# condition is a boolean, True of False
+condition = True
+"condition is True" if condition else None
+```
+
+### Run a cell on a timer
```
mscolnick, line 24: Same here: maybe add "why"

- Load data on an interval
- Run a job on an interval

docs/recipes.md
```diff
@@ -0,0 +1,309 @@
...
+button
+```
+
+3. Show an output after the button is clicked.
+
+```python
+mo.md("#" + "🍃" * button.value) if button.value > 0 else None
+```
+
+### Run a cell when a buton is pressed
```
mscolnick, line 284: These button examples are great, could be nice to include a couple one-liners for use cases

mscolnick (approved):

Issue #386, opened 2023-11-22
`from x import *` results in invalid Python file


akshayka:
### Describe the bug

Using `import *` results in an invalid `.py` file: `import *` can only be used at module level, but marimo puts it in a function.

We may have to disallow `*` imports in marimo. They are not great practice anyway, since they add invisible names to globals that marimo cannot track

cc @bmeyers

### Environment

```
{
  "marimo": "0.1.56",
  "OS": "Linux",
  "OS Version": "6.2.0-36-generic",
  "Python Version": "3.10.12",
  "Binaries": {
    "Chrome": "119.0.6045.159",
    "Node": "v18.13.0"
  },
  "Requirements": {
    "black": "23.3.0",
    "click": "8.1.3",
    "jedi": "0.19.0",
    "pymdown-extensions": "10.3",
    "tomlkit": "0.12.1",
    "tornado": "6.3.3",
    "typing_extensions": "4.8.0"
  }
}
```

akshayka:
Fixed by #387

PR #387, opened 2023-11-22
fix: disallow `import *` in marimo notebooks

akshayka:
Python disallows `import *` everywhere except at the module-level. We store marimo code in functions, so we can't support `import *`. Star imports are also not great because they introduce globals that marimo can't easily track

This fixes a bug in which a user could create a notebook with a `star` import but then couldn't re-open it due to the syntax error.


akshayka:
Fixes #386

PR #404, opened 2023-11-27
Update completion configuration to include activate_on_hover

mscolnick:


akshayka:
- I noticed that this disables cursor tooltips too. But I think those should be linked to completion on typing, not completion on hover?
- As an aside, since we fixed hover tooltips to no longer show in function contexts, I think this isn't very necessary anymore. But we can still make it configurable I suppose!

PR #407, opened 2023-11-27
feat: Add leafmap formatter

mscolnick:


marimo/_output/formatters/leafmap_formatters.py
```diff
@@ -0,0 +1,34 @@
...
+    def register(self) -> None:
+        import leafmap  # type: ignore[import-not-found]
+
+        from marimo._output import formatting
+
+        @formatting.formatter(leafmap.Map)
+        def _show_dataframe(lmap: leafmap.Map) -> tuple[str, str]:
+            # 540px is the pixel height that makes the map fit in the
+            # notebook without scrolling
+            html = mo_data.html(lmap.to_html(height="540px"))
```
akshayka, line 23: Can we respect user width/height, if set? Formatters should do that as a general rule I think. You can access via `lmap.layout`.

PR #408, opened 2023-11-28
experimental: export as html improvements and bug fixes

mscolnick:
* fix special characters or html tags in output by encoding/decoding static dumps as base64
* improve Base64 types with branded strings
* add e2e test for kitchen sink - will export as html and take a photo of the export 
* fix height expansion on some elements when set to 100% 
* include `__resizeIframe` in the html dump

### DIFF
--- a/frontend/src/components/editor/renderers/grid-layout/styles.css
+++ b/frontend/src/components/editor/renderers/grid-layout/styles.css
@@ -21,7 +21,7 @@
 }
 
 /* for images we expand the image to fill the grid item */
-.output img {
+.react-grid-item .output img {
   height: 100%;
   object-fit: contain;
 }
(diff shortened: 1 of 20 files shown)


frontend/src/plugins/impl/plotly/PlotlyPlugin.tsx
```diff
@@ -91,7 +91,7 @@ export const PlotlyComponent = memo(
             range: evt.range,
           });
         })}
-        className="w-full h-full"
+        className="w-full"
```
mscolnick, line 94: this caused the plotly bug

frontend/src/components/editor/renderers/grid-layout/styles.css
```diff
@@ -21,7 +21,7 @@
 }
 
 /* for images we expand the image to fill the grid item */
-.output img {
+.react-grid-item .output img {
```
mscolnick, line 24: this caused the matplotlib bug

akshayka (approved):

PR #412, opened 2023-11-28
feat: add readCode API endpoint, download code in static export

mscolnick:


frontend/src/components/data-table/download-actions.tsx
```diff
@@ -48,12 +49,7 @@ export const DownloadAs: React.FC<DownloadActionProps> = (props) => {
               if (!downloadUrl) {
                 return;
               }
-              const link = document.createElement("a");
-              link.href = downloadUrl;
-              link.setAttribute("download", "download");
-              document.body.append(link);
-              link.click();
-              link.remove();
+              downloadByURL(downloadUrl, "download");
```
mscolnick, line 52: just some DRYing up

akshayka (dismissed):
LGTM except we should communicate error when user tries to export an unnamed file

PR #414, opened 2023-11-29
docs: improve batch recipe + faq

akshayka:
Improve the docs recipes and FAQs for creating outputs with a dynamic number of UI elements. Also add a FAQ for "Why aren't my `on_change` handlers being called?`.

In the future, we can experiment with error messages that point people to the FAQ/recipes if they have unbound UI elements with handlers, and/or experiment with `mo.ui.vstack` and `mo.ui.hstack`.


mscolnick (approved):
this is much clearer, great addition

PR #422, opened 2023-11-30
feat: support markdown syntax

mscolnick:
This feature adds "pseudo" markdown support. Whenever the python code is `mo.md(...)`, we let the user toggle to markdown which includes syntax highlighting and other language features. The code is still python under the hood and we provide a `transformIn/transformOut` to convert from python<>markdown behind the scenes. 

This feature is also designed in a way to add more language providers in future (e.g. sql)

Screenshot 2023-12-01 at 12 28 03 PM

Screenshot 2023-12-01 at 12 27 57 PM


frontend/src/components/editor/cell/code/language-toggle.tsx
```diff
@@ -0,0 +1,50 @@
...
+  if (!canUseMarkdown && languageAdapter !== "markdown") {
+    return null;
+  }
+
+  const otherLanguage = languageAdapter === "markdown" ? "python" : "markdown";
+  const Icon = languageAdapter === "markdown" ? PythonIcon : MarkdownIcon;
+
+  return (
+    <div className="absolute top-0 right-4 z-20 hover-action">
+      <Tooltip content={`Switch to ${startCase(otherLanguage)}`}>
```
akshayka, line 35: Terminology nit.

```suggestion
      <Tooltip content={`View as ${startCase(otherLanguage)}`}>
```

mscolnick:
I originally ignored it for simplicity but i can remove that line

frontend/src/core/codemirror/language/markdown.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    // Multiline code
+    const start = `mo.md(\n\t${prefix}"""\n`;
+    const end = `\n\t"""\n)`;
+    return [start + indentOneTab(escapedCode) + end, start.length + 1];
+  }
+
+  isSupported(pythonCode: string): boolean {
+    return regexes.some(([, regex]) => regex.test(pythonCode));
```
akshayka, line 84: Actually the check was helpful. Without the check now we match cells containing two mo.md() lines, which then garbles the code when you switch to markdown and back.

How about this?

```suggestion
    const markdownLines = pythonCode                                             
        .split("\n")                                                               
        .map((line) => line.startsWith("mo.md("))                                  
        .filter(Boolean);                                                          
    if (markdownLines.length > 1) {                                              
        // more than line starting with mo.md(; as a heuristic,                    
        // don't show "view as markdown"                                           
        return false;                                                              
    }                 
    return regexes.some(([, regex]) => regex.test(pythonCode));
```

frontend/src/core/codemirror/language/markdown.ts
```diff
@@ -0,0 +1,178 @@
...
+    const escapedCode = code.replaceAll('"""', '\\"""');
+    if (isOneLine) {
+      const start = `mo.md(${prefix}"""`;
+      const end = `""")`;
+      return [start + escapedCode + end, start.length];
+    }
+
+    // Multiline code
+    const start = `mo.md(\n\t${prefix}"""\n`;
+    const end = `\n\t"""\n)`;
```
akshayka, line 79: More tabs that should be replaced with 4 spaces.

```suggestion
    const start = `mo.md(\n    ${prefix}"""\n`;
    const end = `\n    """\n)`;
```

frontend/src/core/codemirror/language/markdown.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  transformOut(code: string): [string, number] {
+    // Get the quote type from the last transformIn
+    const prefix = upgradePrefixKind(this.lastQuotePrefix, code);
+
+    const isOneLine = !code.includes("\n");
+    const escapedCode = code.replaceAll('"""', '\\"""');
+    if (isOneLine) {
+      const start = `mo.md(${prefix}"""`;
```
akshayka, line 72: Single-line => Single-quote

```suggestion
      const start = `mo.md(${prefix}"`;
```

frontend/src/core/codemirror/language/markdown.ts
```diff
@@ -0,0 +1,178 @@
...
+      const end = `""")`;
```
akshayka, line 73:
```suggestion
      const end = `")`;
```

mscolnick:
@akshayka nice catch - i was able to handle the formatting change

frontend/src/core/codemirror/language/extension.ts
```diff
@@ -117,6 +118,8 @@ function updateLanguageAdapterAndCode(
       ),
       // Clear history
       historyCompartment.reconfigure([]),
+      // Let downstream extensions know that this is a formatting change
+      formattingChangeEffect.of(true),
```
akshayka, line 122: cool! very slick

frontend/src/core/codemirror/language/markdown.ts
```diff
@@ -0,0 +1,179 @@
...
+        return [dedent(unescapedCode), offset];
+      }
+    }
+
+    return [pythonCode, 0];
+  }
+
+  transformOut(code: string): [string, number] {
+    // Get the quote type from the last transformIn
+    const prefix = upgradePrefixKind(this.lastQuotePrefix, code);
```
akshayka, line 67: Did some more dogfooding and upgrading was troublesome when writing LaTeX, where it's common to have curly braces.

Let's just not upgrade to f-strings and let the user manually handle it. In the future if we really want to, we could have a UI for choosing the prefix in the markdown view (but we definitely don't need it now).

akshayka (approved):
I have one comment regarding not upgrading to f-strings, due to clashing with LaTeX.

Another much more minor thing I noticed is that when the cell is just one line long, the "View as markdown" button is dangerously close to the delete cell button ... could maybe just position the "View as markdown" button slightly more to the left.

Not related to the code, but if you could update the PR description with some explanatory text, that would be great — just as a record of the feature, to make our changelog more useful and also to help other developers.

Otherwise, LGTM!

akshayka (approved):
This is fantastic. Great work!

PR #437, opened 2023-12-05
fix: fix RuntimeState to be able to recover from failed ui element updates

mscolnick:
Previously if `sendComponentUpdates` failed with a 403 or any other network request, the app would not be able to send any more updates. This makes sure we can recover from those


akshayka (approved):
Great catch!

PR #439, opened 2023-12-06
fix: scroll in cell's console output

mscolnick:
Previously you could not scroll in the console area of a cell due to overlapping elements. This fixes that.


akshayka (commented):
Can you add a PR description that explains what this PR fixes? Before without the change, and with it? It's not clear to me from the title + diff.

PR #440, opened 2023-12-06
improvement: option to show code in static html

mscolnick:
* always show a button in the static export to view the code
* code is shown **above** 
* defaults to hidden, but can expose this by a query-param in future if requested
* option to copy each code on hover

Screenshot 2023-12-06 at 3 11 11 PM

### DIFF
--- a/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
+++ b/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
@@ -3,7 +3,6 @@ import { memo, useRef, useState } from "react";
 import { CellRuntimeState } from "@/core/cells/types";
 import { CellId, HTMLCellId } from "@/core/cells/ids";
 import { OutputArea } from "@/components/editor/Output";
-import clsx from "clsx";
 import { ICellRendererPlugin, ICellRendererProps } from "../types";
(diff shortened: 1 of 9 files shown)


akshayka (commented):
Awesome! Good idea to always include code and expose via button. Simpler than having two export types.

Can we show code below? This would be consistent with marimo edit -- since we think of code as "captions" for output. We'd just need some formatting to associate each output with its cell. We could do what we already do in edit -- the box-shadow/border.

frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
```diff
@@ -19,24 +22,36 @@ const VerticalLayoutRenderer: React.FC<VerticalLayoutProps> = ({
...
+        <div className="fixed m-4 left-0 bottom-0">
+          <Button
+            onClick={() => setShowCode((prev) => !prev)}
+            data-testid="show-code"
+          >
+            <Code2Icon className="w-4 h-4 mr-2" />
+            {showCode ? "Hide code" : "Show code"}
+          </Button>
+        </div>
+      )}
```
akshayka, line 54: This is great! Can you put it in the blue banner instead of here, to the left of the "run or edit" button?

akshayka (approved):
LGTM except for one comment!

Issue #356, opened 2023-11-15
Is it possible to recover overwritten marimo notebooks (`.py` files)? 


tdiao:
### Describe the bug

Is it possible to recover overwritten marimo notebooks (`.py` files)? I launched two different scripts in the same port and one overwrote the other, it seems.

Longer version/more details:
I was working in VS code (version 1.84.2) with a remote ssh connection. In the terminal of the ssh machine, I typed `marimo edit remote_script.py` which launched localhost:2718 (the default) in my *local* browser. I edited and saved there, with the "Notebook saved" notification on the bottom right corner popping up often.

I have, at the same time, another VS code window open that is local. When I typed `marimo edit different_script.py` in the terminal (of this local window) where `different_script.py` is a blank script, it launched to localhost:2718 in my local browser again by default, because I meant to but forgot to specify a port. Unfortunately, this action seemed to have overwritten the `remote_scripty.py` in my remote session.

I have tried a few different ways to recover the saving history, including looking at VS code Timeline/local history. The most recent save is my latest commit, ~2 days ago. I am able to see the edit history of my other scripts in the past two days though.

Thank you!

### Version

0.1.43

tdiao:
Hi @akshayka, thanks for the quick and compassionate response. The loss is not huge and I am almost done rebuilding the progress.

Your understanding is correct. I meant to reproduce the error too and will do that as soon as I can. 

From what I recall, launching the tutorials in the same port did not lead to this overwriting issue. Instead, it displayed some error message on the browser page which I don't exactly recall. So it could be the case that this happens only when starting a blank new script. But mostly speculation now; more on that when I can.

akshayka:
Hi @tdiao. We've implemented a fix (in #367) and released it as part of version `0.1.53`. Thanks so much once again for the thorough report!

I'll close this issue for now. But please write back and re-open if you're still running into issues.

PR #443, opened 2023-12-10
fix: don't check server token in run mode

akshayka:
This change modifies the server to only validate the `ServerToken` when running as a notebook (edit mode), and not when running as an app (run mode).

**The fix.**
This fix allows a frontend in run mode to connect to a server that died and was restarted without refreshing the page.

**Context.**
The `ServerToken` was introduced to prevent a frontend started for one notebook (eg, marimo edit locally) from accidentally clobbering/mutating another notebook served on the same port but a different server (see https://github.com/marimo-team/marimo/issues/356). Since run mode is read-only, no such risk exists.


marimo/_server/api/code_complete.py
```diff
@@ -17,9 +17,10 @@ class CodeComplete:
     cell_id: CellId_t
 
 
-class CodeCompleteHandler(ValidatedHandler):
-    """Complete a path to subdirectories and Python files."""
+class CodeCompleteHandler(EditHandler):
+    """Complete a code fragment."""
 
+    @sessions.requires_edit
```
akshayka, line 23: We don't need code completion in run mode.

marimo/_server/api/directory_autocomplete.py
```diff
@@ -13,9 +14,10 @@ class DirectoryAutocomplete:
     prefix: str
 
 
-class DirectoryAutocompleteHandler(ValidatedHandler):
+class DirectoryAutocompleteHandler(EditHandler):
     """Complete a path to subdirectories and Python files."""
 
+    @sessions.requires_edit
```
akshayka, line 20: Don't need directory autocompletion in run mode.

marimo/_server/api/instantiate.py
```diff
@@ -19,7 +20,7 @@ class Instantiate:
     values: List[Any]
 
 
-class InstantiateHandler(ValidatedHandler):
+class InstantiateHandler(tornado.web.RequestHandler):
```
akshayka, line 23: Doesn't require edit

marimo/_server/api/set_cell_config.py
```diff
@@ -17,7 +18,18 @@ class SetCellConfig:
...
-class SetCellConfigHandler(ValidatedHandler):
+class SetCellConfigHandler(tornado.web.RequestHandler):
+    # This mutates cell config in the kernel.
+    #
+    # NB: Checking server token?
+    #
+    # - To prevent a frontend from accidentally setting the config of a kernel
+    #   it wasn't started for (eg, frontend created locally on a port that's
+    #   forwarded to a remote server for a different marimo app), we could
+    #   require a server token.
```
akshayka, line 29: Not a big concern, since this doesn't write to disk. Just might cause a confusing bug, but as an edge case fine to ignore for now I think

marimo/_server/api/set_ui_element_value.py
```diff
@@ -18,7 +19,7 @@ class SetUIElementValue:
     values: List[Any]
 
 
-class SetUIElementValueHandler(ValidatedHandler):
+class SetUIElementValueHandler(tornado.web.RequestHandler):
```
akshayka, line 22: Doesn't require edit

mscolnick (approved):

marimo/_server/api/set_cell_config.py
mscolnick: why not  extend the `EditHandler` that validate the server token? we don't use this handler is run mode afaik

akshayka: I noticed it's used in run mode before instantiating actually. I did extend EditHandler but got an error when trying marimo run, which is how I found out. I guess we shouldn't need to call this in run mode tho, as long as backend is parsing and setting the configs. Can investigate in another change

mscolnick (approved):

PR #449, opened 2023-12-12
fix: check frontend's app matches server's app in run mode

akshayka:
This change makes the server token in run mode a hash of the code, to prevent frontends made for one app from trying to instantiate on a server serving a different app. This still allows frontends to connect to a server that died and came back as long as the server is serving the same app.

### DIFF
--- a/marimo/_server/sessions.py
+++ b/marimo/_server/sessions.py
@@ -218,7 +218,7 @@ class IOSocketHandler(tornado.websocket.WebSocketHandler):
         self.status = ConnectionState.CLOSED
         mgr = get_manager()
         if mgr.mode == SessionMode.RUN:
-            # TODO: keep alive
+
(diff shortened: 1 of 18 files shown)


marimo/_server/sessions.py
```diff
@@ -403,13 +403,25 @@ def __init__(
...
+            # frontends that it didn't create from connecting to it and
+            # executing edit-only commands (such as overwriting the file).
+            self.server_token = str(uuid4())
+        else:
+            # Because run-mode is read-only, all that matters is that
+            # the frontend's app matches the server's app.
+            assert app is not None
+            self.server_token = str(
+                hash("".join(code for code in app._codes()))
+            )
```
akshayka, line 423: This is the main change

mscolnick (approved):

Issue #448, opened 2023-12-12
Progress bar cell marked stale in preview mode


delennc:
### Describe the bug

<img width="614" alt="Screenshot 2023-12-11 at 6 04 51 PM" src="https://github.com/marimo-team/marimo/assets/29712687/03449c2d-bcdb-4379-a382-b841be4e0d6b">

Progress bar is marked stale in preview mode when cell is running, but not in edit mode

### Environment

{
  "marimo": "0.1.64",
  "OS": "Darwin",
  "OS Version": "22.6.0",
  "Python Version": "3.11.4",
  "Binaries": {
    "Chrome": "120.0.6099.71",
    "Node": "v20.3.1"
  },
  "Requirements": {
    "black": "23.10.1",
    "click": "8.1.7",
    "jedi": "0.19.1",
    "pymdown-extensions": "10.3.1",
    "tomlkit": "0.12.1",
    "tornado": "6.3.3",
    "typing_extensions": "4.8.0"
  }
}

### Code to reproduce

```python
import marimo

__generated_with = "0.1.64"
app = marimo.App()


@app.cell
def __():
    import time
    return time,


@app.cell
def __():
    import marimo as mo
    return mo,


@app.cell
def __(mo, time):
    for i in mo.status.progress_bar(range(10)):
        time.sleep(1)
    return i,


if __name__ == "__main__":
    app.run()
```

akshayka:
Thanks for reporting!

Most likely in our frontend we consider outputs stale until the cell has finished running. With streaming outputs this isn't right anymore — as soon as an output arrives during a run, the output should no longer be stale

PR #454, opened 2023-12-12
fix: fix incorrect staleness indicator of outputs in run mode

mscolnick:
Fixes https://github.com/marimo-team/marimo/issues/448

* Refactor output staleness check in Cell component to use shared function
* Test said function

This issue existed since we render cells slightly differently in run mode vs edit mode. We do this to avoid extra logic like drag handlers and edit UI, so ideally we DRY up logic between the two

### DIFF
--- a/frontend/src/components/editor/code/readonly-python-code.tsx
+++ b/frontend/src/components/editor/code/readonly-python-code.tsx
@@ -1,6 +1,5 @@
 /* Copyright 2023 Marimo. All rights reserved. */
 import { memo } from "react";
-import { python } from "@codemirror/lang-python";
 import CodeMirror, {
   EditorView,
   ReactCodeMirrorProps,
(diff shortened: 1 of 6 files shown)


frontend/src/components/editor/code/readonly-python-code.tsx
```diff
@@ -25,7 +25,7 @@ export const ReadonlyPythonCode = memo(
           theme={theme === "dark" ? "dark" : "light"}
           height="100%"
           editable={true}
-          extensions={[python(), EditorView.lineWrapping]}
+          extensions={[customPythonLanguageSupport(), EditorView.lineWrapping]}
```
mscolnick, line 28: @akshayka snuck this in here too

akshayka (approved):

PR #463, opened 2023-12-13
improvement: add --include-code to marimo run

mscolnick:
Add a command-line flag to include the code when running a marimo notebook as an app. This enables a button in the app that allows the user to toggle between notebook view and app view:

`marimo run notebook.py --include-code`

### DIFF
--- a/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
+++ b/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
@@ -1,5 +1,5 @@
 /* Copyright 2023 Marimo. All rights reserved. */
-import { memo, useRef, useState } from "react";
+import React, { memo, useRef, useState } from "react";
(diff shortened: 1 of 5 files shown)


akshayka (commented):
- Like you mentioned, since the `show code` button overlaps with the waiting spinner, how about we put it in the top right?

- I think we should have the option to include code and show it by default, include code and hide it by default, and not include code at all. I envision the following API:

`marimo run -code={shown,hidden,stripped}`

with the first two including code, showing the toggle button, but controlling its default visibility, and `stripped` not including code at all (what we do today). The default value will be stripped.

I can update the Python according to this API, since you're handing off to me

akshayka:
As we discussed offline, decided against providing options for default visibility of code, and will punt to query params in a future PR. Instead, I just renamed `--show-code` to `--include-code`.

akshayka (approved):

PR #484, opened 2023-12-28
feat: allow renaming cells

mscolnick:
- add a way to rename cells in the cell content menu
- if a cell is named, it is show below
- drive-by: fix a small bug where we allowed downloading a png of a cell's empty output

### DIFF
--- a/frontend/src/components/editor/actions/name-cell-input.tsx
+++ b/frontend/src/components/editor/actions/name-cell-input.tsx
@@ -1,0 +1,120 @@
+/* Copyright 2023 Marimo. All rights reserved. */
+import { Input } from "@/components/ui/input";
+import { getCellNames, useCellActions } from "@/core/cells/cells";
+import { CellId } from "@/core/cells/ids";
+import {
+  DEFAULT_CELL_NAME,
+  normalizeName,
+  getValidName,
+} from "@/core/cells/names";
+import { useOnMount } from "@/hooks/useLifecycle";
+import { cn } from "@/utils/cn";
+import { Events } from "@/utils/events";
+import React, { useRef, useState } from "react";
+
+interface Props {
+  className?: string;
+  value: string;
+  onChange: (newName: string) => void;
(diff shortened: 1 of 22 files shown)


akshayka:
Sweet! This is cool.

I like putting the `name` text input in the cell menu, and I like the label being editable when shown. Both feel natural and intuitive.

I think we should tweak the position of the label so that it doesn't consume additional vertical space. I've suggested an alternative here: https://www.figma.com/file/6LzExkKpkgdcoqKuL7Pf1L/Cell-Name-Options?type=whiteboard&node-id=0-1

akshayka:
The cell links in the errors panel appear to no longer work. Clicking on the link should highlight the cell

[Screencast from 01-02-2024 10:41:37 AM.webm](https://github.com/marimo-team/marimo/assets/1994308/029eb3cb-d1b6-4860-97b5-ef536677f162)

akshayka:
Can we add a tooltip to the cell name input in the console area? "Edit this cell's name" if the cell is already named by the user (name != `__`), otherwise "Give this cell a name".

frontend/src/components/editor/actions/name-cell-input.tsx
```diff
@@ -0,0 +1,120 @@
...
+  className: string;
+}> = ({ value, cellId, className }) => {
+  const { updateCellName } = useCellActions();
+  const inputProps = useCellNameInput(value, (newName) =>
+    updateCellName({ cellId, name: newName })
+  );
+
+  return (
+    <span
+      className={cn("outline-1 outline-cyan-500/40", className)}
```
akshayka, line 69: Let's have the cyan outline show up on hover as well, to make it more obvious that it's editable

akshayka:
When auto-save is on, I believe renaming a cell should register a save with the kernel. I got confused why my name changes weren't persisted when testing

akshayka (approved):
LGTM

PR #489, opened 2023-12-31
improvement: use virtual file for image

mscolnick:
Use virtual files for `mo.image`. This is useful if a user has uploaded a large image from the file dropdown and then wants to display it.


marimo/_plugins/stateless/conftest.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2023 Marimo. All rights reserved.
```
mscolnick, line 1: this is copy-paste from another conftest. is there a better way to share this?

akshayka: For now, not really. I need to refactor our tests to live in their own directory, then we'll just put the `conftest.py` at the top level of `_tests`. 

In the meantime can you move this file to `marimo/_plugins/`?

marimo/_plugins/stateless/test_image.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_image_bytes_io(k: Kernel, exec_req: ExecReqProvider) -> None:
+    k.run(
+        [
+            exec_req.get(
+                """
+                import io
+                import marimo as mo
+                bytestream = io.BytesIO(b"\x89PNG\r\n\x1a\n\x00\x00\x00")
```
akshayka, line 22: I tried testing locally on Ubuntu, and Linux doesn't like this bytestring, saying that it contains `null` bytes (which causes the test to fail). In particular cell registration fails with

```
(UnknownError(msg='Traceback (most recent call last):
File "/home/akshay/src/marimo-team/marimo/marimo/_runti...compile(source, filename, mode, flags,
ValueError: source code string cannot contain null bytes\n', type='unknown'),
```

akshayka: The same error affects `test_image_bytes`

akshayka: More context: https://github.com/python/cpython/issues/96670#issuecomment-1249054660

So it seems like we can't hardcode the bytestring as a string literal. Instead you could download the image and convert it to bytes at runtime?

marimo/_plugins/stateless/conftest.py
mscolnick: sounds good - we can refactor in a followup. im pretty sure you already have a ticket for this

marimo/_plugins/stateless/test_image.py
mscolnick: i just updated with `b"hello"`. it doesnt need to be an actual image

akshayka (approved):

PR #490, opened 2023-12-31
refresh with ISS API

liquidcarbon:
An example showing the usage of refresh widget to fetch new data and update plots of the International Space Station location above Earth.

There's probably a few things to clean up but wanted to get this going in 2023. :)

### DIFF
--- a/examples/refresh.py
+++ b/examples/refresh.py
@@ -20,7 +20,7 @@ def __(
             n_points_slider,
             refresher,
             mo.md("## ISS Positions | [`marimo.ui.altair_chart`](https://docs.marimo.io/api/plotting.html#marimo.ui.altair_chart)\n---"),
-            mo.as_html(chart)
+            mo.as_html(chart).style({"width": "700px"})
         ], align="center"),
         mo.vstack([
             mo.md("## Data | [`marimo.as_html`](https://docs.marimo.io/api/html.html)`(pd.DataFrame)`\n---"),
--- a/examples/refresh.py
+++ b/examples/refresh.py
@@ -1,6 +1,6 @@
 import marimo
 
-__generated_with = "0.1.63"
+__generated_with = "0.1.68"
 app = marimo.App(width="full")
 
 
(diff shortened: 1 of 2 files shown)


liquidcarbon:

akshayka:
Awesome! Thanks so much for this example.

One small request: Can you move `refresh.py` into the `examples` directory?

When I try out the notebook, I don't see the location of the ISS; I just see an empty map. Maybe this is expected, but just thought I'd mention it in case it's not:

image

Thanks again for the really neat example!

liquidcarbon:
For some reason it works fine on Marimo cloud ( @mscolnick helped me set it up ).

The same code, using `mo.ui.altair_chart`, like you note, is not showing the active updated chart.  It is showing up though with `mo.as_html(chart)` or just `chart`, but the divs are arranged differently and the styling doesn't show up as intended.

What does `ui.altair_chart` do, am I using it wrong?

examples/refresh.py
```diff
@@ -0,0 +1,126 @@
...
+    refresher,
+):
+    mo.hstack([
+        mo.vstack([
+            mo.md("## Settings | [`marimo.ui.slider`](https://docs.marimo.io/api/inputs/slider.html), [`marimo.ui.refresh`](https://docs.marimo.io/recipes.html#run-a-cell-on-a-timer)\n---"),
+            refresh_interval_slider,
+            n_points_slider,
+            refresher,
+            mo.md("## ISS Positions | [`marimo.ui.altair_chart`](https://docs.marimo.io/api/plotting.html#marimo.ui.altair_chart)\n---"),
+            mo.as_html(chart)
```
akshayka, line 23: The width not being respected is a bug on our part I think. Oops! In the meantime, this suggestion should fix it:

```suggestion
            mo.as_html(chart).style({"width": "700px"})
```

akshayka:
> What does ui.altair_chart do, am I using it wrong?

`ui.altair_chart` makes the chart reactive, so if you have a scatter plot you could select points and get them back as a dataframe. But it's not needed to display an altair chart. So you're not using it wrong per se, but you don't need it here.

> The same code, using mo.ui.altair_chart, like you note, is not showing the active updated chart. 

This is a bug on our part; oops!

> It is showing up though with mo.as_html(chart) or just chart, but the divs are arranged differently and the styling doesn't show up as intended.

Also a bug on our part. I left a suggestion inline with a workaround. The width looks fine, and I can see the ISS position (so cool!)

Thanks!

akshayka (approved):
# 🌌 🛰

LGTM! This is awesome!

PR #491, opened 2024-01-01
Refactor tests into their own directory

akshayka:
This simplifies sharing of pytest fixtures across tests, and helps brings our repo into agreement with modern Python conventions for code and test organization.


akshayka:
Looking into failing tests

PR #502, opened 2024-01-03
Relax validation of args/returns on cells

akshayka:
This change removes validation of a cell's args and returns on notebook startup. The validation made spot-editing a notebook onerous, and it wasn't needed because marimo auto-fixes errors on save.

This change also makes it somewhat easier to write a notebook from scratch in your text editor, because you can just omit the refs and defs. Eg, you could open the following file with marimo edit:

```python3
import marimo
app = marimo.App()

@app.cell
def imports():
  import marimo as mo

@app.cell
def greetings():
  mo.md("# Hello, world!")
```

and marimo would autofix it for you on save.

This change was motivated by feedback from users, who wanted to be able to easily make edits to their notebooks in their editor of choice.


mscolnick (dismissed):
Probably worth having a test case for this. Also, I assume it still fails for multiple defs? Or unknown refs? 

Maybe a test case for that as well?

akshayka:
> Probably worth having a test case for this. Also, I assume it still fails for multiple defs? Or unknown refs?
> 
> Maybe a test case for that as well?

Yea, app.run() still fails for multiple defs, unknown refs. Added tests, thanks.

marimo forms the dag by just reading the cell code, ignoring args/rets. The args and rets are inserted by marimo on save so the cells can be executed as functions, but they aren't needed to resolve refs/defs.

PR #505, opened 2024-01-04
fix: Handle banned names when renaming a cell

mscolnick:
- Handle banned names when renaming a cell
- Fix console warnings that are shown during development


akshayka (approved):
Looks great!

Issue #507, opened 2024-01-05
Columns with same name silently not shown in `mo.ui.table`


akshayka:
### Describe the bug

Columns with the same name are silently not shown by mo.ui.table. However, if you check the downloaded CSV, the data is there.

![image](https://github.com/marimo-team/marimo/assets/1994308/efcd475f-a5c0-4bba-9744-155e124c3945)

cc @delennc

### Environment

```
{
  "marimo": "0.1.69",
  "OS": "Linux",
  "OS Version": "6.2.0-39-generic",
  "Processor": "x86_64",
  "Python Version": "3.10.12",
  "Binaries": {
    "Chrome": "119.0.6045.159",
    "Node": "v18.13.0"
  },
  "Requirements": {
    "black": "23.3.0",
    "click": "8.1.3",
    "jedi": "0.18.2",
    "pymdown-extensions": "10.0.1",
    "tomlkit": "0.12.1",
    "tornado": "6.2",
    "typing_extensions": "4.4.0"
  }
}
```

### Code to reproduce

```python
import marimo

__generated_with = "0.1.69"
app = marimo.App()


@app.cell
def __():
    import pandas as pd
    return pd,


@app.cell
def __():
    import numpy as np
    return np,


@app.cell
def __():
    import marimo as mo
    return mo,


@app.cell
def __(np, pd):
    df_with_duplicate_columns = pd.DataFrame(
        np.array([[1, 2], [4, 5]]), columns=["a", "a"]
    )
    df_with_duplicate_columns
    return df_with_duplicate_columns,


@app.cell
def __(df_with_duplicate_columns, mo):
    mo.ui.table(df_with_duplicate_columns)
    return


if __name__ == "__main__":
    app.run()
```

PR #511, opened 2024-01-05
fix: handle duplicate column names in CSV data

mscolnick:
Fixes https://github.com/marimo-team/marimo/issues/507

Make each csv column name unique before converting to JS objects, by appending zero-width-spaces

Other options:
* larger refactor of passing around objects in a columnar format
* change each column name to index, and create a mapping table to its key
* append `_1`, `_2` to each

image


akshayka (commented):
Since our users use Pandas a lot, we should strive for the table UI and the dataframe to be identical. If that's too much work, in the interim I would prefer just raising an `ValueError` in `mo.ui.table` if there are duplicate column names. 

> use zero padding spaces to make them unique so they look the same, but are unique by string comparison

I think this could work! But up to you.

For context: for the user who reported this bug, the main point of confusion was the fact that the column headers in the table UI didn't match the columns in the dataframe (they were using `mo.ui.table` to sanity check transformations).

PR #514, opened 2024-01-05
misc: feedback dialog

mscolnick:
This adds a feedback dialog for users to send anonymous feedback

Screenshot 2024-01-05 at 3 26 48 PM


akshayka (approved):
This is awesome!!

Should we have an additional input for people to (optionally) provide their email address? In case they want us to get in touch

PR #516, opened 2024-01-06
improvement: sortable variable table

mscolnick:
This refactors the variable table to use the tanstack headless-UI table library (we use it for our `mo.ui.table` already). This makes it easy to add sorting and flexible columns. 

This also unlocks virtualization or column resizing if we ever want that.


akshayka (approved):
This is great!

PR #551, opened 2024-01-13
Update callout kinds

pavelzw:
The documentation doesn't mention `"alert"` but instead uses `"danger"`.

image


akshayka (approved):
Thanks!

PR #554, opened 2024-01-13
improvement: make footer more visible

akshayka:
This change makes the footer bar and the footer buttons more noticeable by:

- adding a slight shadow,
- increasing the size of the icons,
- moving the help icon next to the panel icons (it's not a panel, but it's more visible there and the discongruity is ok)

I also tried experimenting with turning the icons black or blue, and floating them in the left as a sidebar instead of a footer. The sidebar could be a better long-term solution -- right now we don't really have any design elements to bring the user's eyes to the bottom of the screen -- but would require some finessing to get right. Hopefully this PR combined with better docs is sufficient  ... we can wait and see how users respond.

### DIFF
--- a/frontend/src/components/editor/chrome/wrapper/footer.tsx
+++ b/frontend/src/components/editor/chrome/wrapper/footer.tsx
@@ -23,14 +23,14 @@ export const Footer: React.FC = () => {
   const errorCount = useAtomValue(cellErrorCount);
 
   return (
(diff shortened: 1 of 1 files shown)


mscolnick (approved):
could also try changing the background of the footer too to one level darker

akshayka:
> could also try changing the background of the footer too to one level darker

I tried but one or even two levels darker wasn't very noticeable. Inverting the color with respect to the theme (black foot, white icons in light, opposite in dark) is very noticeable and looks interesting ... but if I went down that path I'd spend too long tweaking. Might explore again later.

PR #555, opened 2024-01-14
Improve docstring for mo.image

akshayka:
Improve example: when opening from a local file, assign the output to a variable and output it.

### DIFF
--- a/marimo/_plugins/stateless/image.py
+++ b/marimo/_plugins/stateless/image.py
@@ -23,9 +23,10 @@ def image(
 ) -> Html:
     """Render an image as HTML.
 
-    **Example.**
+    **Examples.**
 
     ```python3
+    # Render an image from a url
     mo.image(
         src="https://marimo.io/logo.png",
         alt="Marimo logo",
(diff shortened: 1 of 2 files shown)


marimo/_plugins/stateless/image.py
```diff
@@ -23,24 +23,30 @@ def image(
...
         height=100,
         rounded=True,
     )
+    ```
 
+    ```python3
+    # Render an image from a local file
     with open("logo.png", "rb") as file:
-        mo.image(src=file)
+        image = mo.image(src=file)
```
mscolnick, line 42: any downside to auto-opening when the file is a local path?

akshayka: Nope! I think we should do that. Just need to write some logic to detect if it's a URL or not.

akshayka: Can leave for another PR? Since we'll keep the file support anyway.

mscolnick: i guess, wouldnt we want to do the heavy lifting for the user? seems like a suboptimal example otherwise?

akshayka: done

akshayka:
The unit test isn't working on Windows. Tried a number of things — not sure why it's failing.

mscolnick (approved):

PR #564, opened 2024-01-14
feat: mo.mermaid

mscolnick:
This add `mo.mermaid`. 

Screenshot 2024-01-14 at 10 47 50 AM
Screenshot 2024-01-14 at 10 47 14 AM

* I tried initially making this a recipe by using `mo.Html` and esm imports, but was not working properly with reloads with the current way to render scripts.

### DIFF
--- a/frontend/src/plugins/impl/CodeEditorPlugin.tsx
+++ b/frontend/src/plugins/impl/CodeEditorPlugin.tsx
@@ -56,7 +56,7 @@ const CodeEditorComponent = (props: CodeEditorComponentProps) => {
   const minHeight = props.minHeight ? `${props.minHeight}px` : "70px";
 
   return (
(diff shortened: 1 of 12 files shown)


frontend/src/plugins/impl/CodeEditorPlugin.tsx
```diff
@@ -56,7 +56,7 @@ const CodeEditorComponent = (props: CodeEditorComponentProps) => {
   const minHeight = props.minHeight ? `${props.minHeight}px` : "70px";
 
   return (
-    <Labeled label={props.label} align="top">
+    <Labeled label={props.label} align="top" fullWidth={true}>
```
mscolnick, line 59: small bug fix

akshayka (approved):
Nice! (Took the liberty of touching up a docstring but that's it.)

PR #581, opened 2024-01-16
mo.ui.data_voyager

mscolnick:
This adds mo.ui.data_voyager which is inspired by https://vega.github.io/voyager/. 

It uses https://github.com/vega/compassql for chart recommendations and guides the user into exploring various charts and drilling down. 

Screenshot 2024-01-16 at 10 03 03 AM

Screenshot 2024-01-16 at 10 03 19 AM

I expect a fair amount of back-and-forth comments on the UI/UX. There are also some features not added, but can be added in this PR if preferred: 
* sorting
* scale (e.g. log)
* filtering a fields values
* grouping/filtering the types of recommendations (qualitative, temporal)
* stack offset for row/column encodings
* suggesting flipping encoding. currently all suggested are additive and we don't suggest replacements.

Things to note:
* we don't add all time fns, there are a lot. i am aiming to add the common ones
* The license from https://github.com/vega/voyager will need to be attributed, but we don't bring this library as a dep. The code from there has been simplified and modified, but still asks that we include the License.


akshayka (commented):
This is fantastic! The ability to very rapidly make plots is super nice -- really levels up EDA.

Some questions:

1. I noticed that the plugin is under `mo.ui`. What is it's value? When I played with it the value was always an empty dictionary.
2. What does this button do? I couldn't figure it out.

image

marimo/_plugins/ui/_impl/data_explorer.py
```diff
@@ -0,0 +1,55 @@
...
+
+import marimo._output.data.data as mo_data
+from marimo._output.rich_help import mddoc
+from marimo._plugins.ui._core.ui_element import UIElement
+
+
+@mddoc
+class data_explorer(UIElement[Dict[str, Any], Dict[str, Any]]):
+    """
+    Quickly explore a DataFrame.
```
akshayka, line 18:
```suggestion
    Quickly explore a DataFrame with automatically suggested visualizations.
```

tests/_plugins/ui/_impl/test_data_explorer.py
```diff
@@ -0,0 +1,24 @@
...
+HAS_DEPS = DependencyManager.has_pandas()
+
+
+@pytest.mark.skipif(not HAS_DEPS, reason="optional dependencies not installed")
+def test_data_explorer() -> None:
+    import pandas as pd
+
+    # Create kernel and give the execution context an existing cell
+    mocked = MockedKernel()
+    mocked.k.execution_context = ExecutionContext("test_cell_id", False)
```
akshayka, line 20: Oh creating a `MockedKernel` like this can lead to problems because it won't get cleaned up. I added a fixture for this called `executing_kernel`.

`def test_data_explorer(executing_kernel: Kernel) -> None:`

tests/_plugins/ui/_impl/test_table.py
```diff
@@ -82,3 +74,21 @@ def test_normalize_data(executing_kernel: Kernel) -> None:
...
+
+
+HAS_DEPS = DependencyManager.has_pandas()
+
+
+@pytest.mark.skipif(not HAS_DEPS, reason="optional dependencies not installed")
+def test_normalize_data_pandas() -> None:
+    # Create kernel and give the execution context an existing cell
+    mocked = MockedKernel()
+    mocked.k.execution_context = ExecutionContext("test_cell_id", False)
```
akshayka, line 86: Same thing here re using `executing_kernel` fixture

akshayka (commented):
Two small comments re: a text fixture, otherwise LGTM!

akshayka (approved):

PR #591, opened 2024-01-17
feat: documentation panel

mscolnick:
This stores the completion response when inside a function call in the documentation tab. This is re-using an old PR when we tried to put this in a popover. 

This documentation panel isn't perfect. Some improvements that could happen in this PR or later:
1. Parse certain packages that we know are RST using an RST parser (e.g. numpy)
2. Populate the panel when we move the cursor to a function (not just the arguments of it)

Screenshot 2024-01-16 at 9 01 01 PM


akshayka:
Nice!

> Populate the panel when we move the cursor to a function (not just the arguments of it)

I think we implement this feature in this PR. I tried it out and it's a bit awkward / non-obvious having to type in order to get the docs to show up

mscolnick:
@akshayka, now we request using the same logic as the tooltip - should be pretty sufficient. some variables have pretty weird/poor docs

akshayka:
Nice! Just tried it out for a bit. 

I like that docs show up on cursor move now. However when I move the cursor around (typically by clicking), maybe 20% of the time the docs don't update or I see "No documentation". Here's an example 

image

> some variables have pretty weird/poor docs

I guess the issue with using tooltips is that now we get "documentation" for variables, whereas really all we're seeing is the variable name and its type. If we want to avoid this I don't mind creating a new route in the backend that's only for docstrings.

akshayka (approved):

PR #597, opened 2024-01-17
chore: decouple sessions from tornado

mscolnick:
Decoupling sessions.py from any tornado logic. This is so we can switch to another webframework much more easily

### DIFF
--- a/marimo/_server/model.py
+++ b/marimo/_server/model.py
@@ -1,0 +1,51 @@
+# Copyright 2024 Marimo. All rights reserved.
+import abc
+from enum import Enum
+from multiprocessing.connection import Connection
+from typing import Any, Callable
+
+
+class ConnectionState(Enum):
+    """Connection state for a session"""
+
+    OPEN = 0
+    CLOSED = 1
+
+
+class SessionMode(Enum):
+    """Session mode for a session"""
+
+    # read-write
+    EDIT = 0
+    # read-only
+    RUN = 1
+
+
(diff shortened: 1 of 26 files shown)


marimo/_server/model.py
```diff
@@ -0,0 +1,51 @@
...
+class SessionMode(Enum):
+    """Session mode for a session"""
+
+    # read-write
+    EDIT = 0
+    # read-only
+    RUN = 1
+
+
+class SessionHandler(metaclass=abc.ABCMeta):
```
mscolnick, line 24: this abstraction isn't perfect, but does remove the coupling to torando. the arguments passed should be sufficient for other connection types

marimo/_server/server_handler.py
```diff
@@ -0,0 +1,269 @@
...
+    def on_start(
+        self,
+        mode: SessionMode,
+        connection: Connection,
+        check_alive: Callable[[], None],
+    ) -> None:
+        self.mode = mode
+        if mode == SessionMode.RUN:
+            loop = tornado.ioloop.IOLoop.current()
+            loop.remove_handler(connection.fileno())
```
akshayka, line 234: Isn't this clean up logic, which happens in `on_stop`? Why is this here?

mscolnick: incorrectly copied from the previous code. fixed

akshayka (approved):

Issue #558, opened 2024-01-14
Ctrl-[ does not act as Escape in vim mode


inducer:
### Describe the bug

Somehow, I've trained myself to use Ctrl-[ as escape in vim and vim-likes (much closer to the home row than the actual Escape key, given that I have Caps mapped to Ctrl). As you may be able to guess,  this throws a bit of a wrench in my muscle memory. This seems like a hopefully easy thing to fix that would make Marimo easier to adopt for me. As additional context, this was on Firefox in Linux. I have Vimium (a browser extension for more Vim-like browsing) installed, but I did turn it off for Marimo. Other web pages seem to be able to capture Ctrl-[ just fine.

### Environment

{
  "marimo": "0.1.76",
  "OS": "Linux",
  "OS Version": "6.5.0-5-amd64",
  "Processor": "",
  "Python Version": "3.12.1",
  "Binaries": {
    "Chrome": "--",
    "Node": "v18.19.0"
  },
  "Requirements": {
    "black": "23.9.1",
    "click": "8.1.7",
    "jedi": "0.19.1",
    "pymdown-extensions": "10.7",
    "tomlkit": "0.12.1",
    "tornado": "6.3.3",
    "typing_extensions": "4.8.0"
  }

### Code to reproduce

(none)

mscolnick:
@inducer - seems like part of your issue got stripped 
<img width="837" alt="image" src="https://github.com/marimo-team/marimo/assets/2753772/71023b9b-e4ba-43f1-87fe-2d4036bee8c9">

mscolnick:
also, fwiw, I cannot produce this on Mac/Chrome, so likely an OS issue with Linux

inducer:
> seems like part of your issue got stripped

Thanks! Edited. Not sure where it went, I distinctly remember typing the rest of it... :shrug:

akshayka:
I can reproduce it on Linux/Chrome. I will look into it

PR #611, opened 2024-01-20
fix: move dd on cell in Vim mode, so it doesnt conflict with other commands

mscolnick:
Fixes: https://github.com/marimo-team/marimo/issues/558

This was only affecting Linux users, and not Mac users.


akshayka:
We tried several times to get this working but weren't able to

PR #614, opened 2024-01-21
improvement: form customization options and validation

mscolnick:
- add configurable submit button (tooltip, text)
- add configurable reset button (visible, tooltip, text)
- add validation via RPC functions
- add tooltip to mo.ui.button

For designs, you can check out: https://marimo-storybook-git-ms-form-features-marimo.vercel.app/?path=/story/formwrapper--primary (and the 4 other variations)


marimo/_plugins/ui/_impl/input.py
```diff
@@ -1049,19 +1059,56 @@ def __init__(
...
+        submit_button_label: str = "Submit",
+        submit_button_tooltip: Optional[str] = None,
+        submit_button_disabled: bool = False,
+        clear_on_submit: bool = True,
+        show_clear_button: bool = False,
+        clear_button_label: str = "Clear",
+        clear_button_tooltip: Optional[str] = None,
+        validate: Optional[
+            Callable[[Optional[JSONType]], Optional[str]]
+        ] = None,
```
mscolnick, line 1073: this is repeated between `class form`. is there a way to consolidate this without making a nested dataclass and keeping it typesafe with docs?

marimo/_plugins/ui/_impl/input.py
```diff
@@ -1049,19 +1071,56 @@ def __init__(
         self,
         element: UIElement[JSONTypeBound, T],
         *,
+        bordered: bool = True,
+        loading: bool = False,
+        submit_button_label: str = "Submit",
+        submit_button_tooltip: Optional[str] = None,
+        submit_button_disabled: bool = False,
+        clear_on_submit: bool = True,
```
akshayka, line 1079: Can this default to `False`? In the common case where the user is using `form` to gate submission of UI elements, it's helpful to see what was last submitted. The transition from `stale` to `submitted` already gives a visual cue to the user that the form was submitted.

```suggestion
        clear_on_submit: bool = False,
```

marimo/_plugins/ui/_impl/input.py
akshayka: Not that I know of unfortunately

akshayka (commented):
This is really great! The validation RPC is very nice. Just one comment about flipping the default value for `clear_on_submit`.

akshayka (approved):

PR #615, opened 2024-01-21
feat: emojis autcomplete in markdown

mscolnick:
Just for fun - emoji auto-complete when in markdown mode via `:`.

I pull emojis from a CDN to avoid the cost when bundling (although could lazy load), which should be a fine tradeoffs since when offline, it is ok to miss this feature. (although happy to bundle/lazy-load)


akshayka (approved):
❤️! LGTM. Idea/and one issue below.

- Thoughts about using Noto Color Emoji for the emoji font, to standardize across machines? We could also pull this from a CDN: https://fonts.google.com/noto/specimen/Noto+Color+Emoji?query=emoji

- One issue,  up to you to determine whether worth fixing now (not sure how much effort). I can only navigate the completion option with the up arrow key, the down key doesn't work (tested on macOS)

image

PR #638, opened 2024-01-23
test: test WebComponentEncoder

mscolnick:


akshayka (commented):
Now that we have testing set up for optional dependencies, I think the following changes would work:

- add numpy to `testoptional`
- modify these tests to be skipped if NumPy/Pandas are not installed
- move to `tests` folder

akshayka (approved):

PR #654, opened 2024-01-25
Resumable sessions

mscolnick:
This PR makes sessions "resumable". 
When a user is editing, we don't close the sessions when the browser closes (i.e. when the websocket loses connection).
When they refresh the page (or close the page and come back minutes later), we resume the last open session and stream any outputs or computation that happened while the browser was closed. 
We also show a banner when this reconnection happens, giving the user an easy way to restart the kernel. 

This PR also:
* add an endpoint `/restart_session`
* removes `msw` from the FE (this was unused)
* add `TypedConnection`, `KnownMimeTypes`, and `CellChannel` for stronger typing

Still TODO:
- [x] design polish of the banner
- [x] playwright e2e test


marimo/_messaging/mimetypes.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Marimo. All rights reserved.
+from typing import Literal
+
+# It is convenient to write mimetypes a string,
```
akshayka, line 4:
```suggestion
# It is convenient to write mimetypes as strings,
```

twrightsman:
Wow, just tried this out! Worked great with my minimal sleep loop test. Thank you for implementing this!

marimo/_server/sessions.py
```diff
@@ -212,49 +220,110 @@ def create(
...
+            await self.session_consumer.write_operation(operation)
 
     def close(self) -> None:
-        self.session_handler.on_stop(self.kernel_manager.kernel_connection)
+        # Could be no consumer if we already disconnect, but the session
+        # is running in the background
+        if self.session_consumer is not None:
+            self.session_consumer.on_stop()
         self.kernel_manager.close_kernel()
+        self.message_distributor.stop()
```
akshayka, line 316: One of these two methods needs to remove the reader on the multiprocessing connection from the event loop:

```python
asyncio.get_running_loop().remove_reader(self.kernel_connection.fileno())
```

Failure to remove the reader causes the server to hang on exit on Windows (I was only able to free up the terminal by using the Windows Task Manager to forcibly kill the Python process).

Additionally, the reader must be removed _before_ the kernel connection is closed (which currently happens in `kernel_manager.close_kernel()`, otherwise an `OSError` will be thrown

marimo/_utils/distributor.py
```diff
@@ -0,0 +1,68 @@
...
+            try:
+                response = self.input_connection.recv()
+            except (EOFError, StopIteration):
+                break
+            for consumer in self.consumers:
+                consumer(response)
+
+    def start(self) -> Disposable:
+        """Start distributing the response."""
+        asyncio.get_event_loop().add_reader(
```
akshayka, line 52: `get_event_loop()` is deprecated starting in Python 3.12, since it silently returns a new event loop if one is not installed. 

The docs recommend replacing this function with `get_running_loop()`, which raises a `RuntimeError` when no event loop is found, though that may make some of your unit tests fail if they don't install an event loop ...

tests/_server/api/endpoints/test_resume_session.py
```diff
@@ -0,0 +1,252 @@
...
+
+def test_restart_session(client: TestClient) -> None:
+    with client.websocket_connect("/ws?session_id=123") as websocket:
+        data = websocket.receive_json()
+        assert_kernel_ready_response(data, create_response({}))
+        # Send save request
+        client.post(
+            "/api/kernel/restart_session",
+            headers=headers("123"),
+        )
```
akshayka, line 236: Moving this out of the `with` block fixes one of the failing `Windows` tests. Does that work or does that break the logic of the test?

akshayka (commented):
I tested on Windows and found a bug that prevented the server from shutting down (had to force quit from the task manager) -- the kernel multiprocessing connection reader needs to be removed from the event loop. Please see my inline comment for options on where/how to do this.

I also left comment for fixing one of the failing Windows tests, but I have not yet been able to diagnose why the others are failing.

marimo/_utils/distributor.py
mscolnick: i can do this in a followup (give `main` has the same problem), assuming I can get the windows tests to pass without the refactor

tests/_server/api/endpoints/test_resume_session.py
mscolnick: shouldnt effect the logic, with resumed sessions. moving it out

marimo/_server/sessions.py
mscolnick: good call. i originally did remove the reader, but removed that recently. i added it back

marimo/_server/session/session_view.py
```diff
@@ -0,0 +1,162 @@
...
+        next_.timestamp = previous.timestamp
+
+    if next_.output is None:
+        next_.output = previous.output
+
+    return next_
+
+
+def as_list(value: Union[Any, Optional[Any], list[Any]]) -> list[Any]:
+    if not value:
```
akshayka, line 160: I think this is what you want? `not value` will evaluate to `True` for `False`-y objects

```suggestion
    if value is None:
```

akshayka (dismissed):
Just one small comment

akshayka (approved):

PR #663, opened 2024-01-26
chore: create cell ids on the backend, and change to random alpha

mscolnick:
This changes `CellId` generation to be unique 4 letters instead of a counter. These IDs still only exist at runtime and not persistent anywhere (which is desired)

This is necessary for resumable sessions since we don't track cell order changes in the BE view of a session, so on if the user moves cells and refreshes, it can get in a weird state.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -235,10 +235,11 @@ class CellManager:
         self._cell_data: dict[CellId_t, CellData] = {}
         self._cell_data = OrderedDict()
         self.unparsable = False
+        self.random_seed = random.Random(42)
 
     def create_cell_id(self) -> CellId_t:
         # 4 random letters
-        return "".join(random.choices(string.ascii_letters, k=4))
+        return "".join(self.random_seed.choices(string.ascii_letters, k=4))
 
     def cell_decorator(
         self,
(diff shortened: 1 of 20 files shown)


akshayka (commented):
Makes sense to move to random IDs. I just want to make sure that instantiate requests that have object ids + values still work (which can happen in run mode).

For that to work, the cell IDs have to be deterministic for each run of the notebook. I think this can be achieved by setting a random seed.

marimo/_ast/app.py
```diff
@@ -232,13 +234,11 @@ class CellManager:
...
         self._cell_data: dict[CellId_t, CellData] = {}
         self._cell_data = OrderedDict()
-        self._cell_id_counter = 0
         self.unparsable = False
 
     def create_cell_id(self) -> CellId_t:
-        cell_id = str(self._cell_id_counter)
-        self._cell_id_counter += 1
-        return cell_id
+        # 4 random letters
```
akshayka, line 240: I believe we need to set a random seed for instantiate requests that have object IDs/values to work. I suggest:

- create a `random.Random()` instance with a fixed seed in `App's` constructor: e.g., `self.rng = random.Random(42)`.
- use this instance (instead of the module-level `random`) in `create_cell_id`: `self.rng.choices(...)`

mscolnick: sounds good, will do

akshayka (approved):

akshayka (approved):

Issue #658, opened 2024-01-25
In vim mode allow j/k to move between cells like up/down


peterwilton-ttam:
### Description

Using the vim keymap in normal mode, I'd like to be able to navigate between the current cell and the adjacent cells with j/k when I reach the bottom/top of the cell. This seems to be how up/down is set up in the normal keymap, so it seems like it should be possible with the vim keymap.

I'm coming from Jupyter, and Jupyter's various vim keybinding plugins have always supported this, so it's been a small hurdle as I get used to Marimo.

### Suggested solution

If possible, mirror the code that allows this to happen with up/down.

Also, IMO it would be nice if, for both j/k and up/down, navigating between cells with j/k/up/down either consistently moves the cursor to be beginning of the line in the destination cell or to the same column as in the source cell. Right now (with up/down), it moves to the column last used in the destination cell, which is often the last or first column because up/down moves to the beginning/end of the line before moving between cells. If possible I'd remove this behavior too and just have up/down/j/k move to the next cell rather than the end of the line. vim never moves to the end/beginning of the line with j/k.

### Alternative

The current keybindings of cmd+shift+j/k works well enough after some muscle memory adjustment, but I think allowing j/k to move between cells would make it a bit more fluid.

### Additional context

I'm loving Marimo so far. Thanks!

akshayka:
Hi, thanks for the detailed feedback! As a vim user myself I'd also like this. We'll add it to our task list.

mscolnick:
@peterwilton-ttam, thanks for the detailed write-up. We've made changes in #667 to fix this. The new behavior will be:

* Allows `j/k` to focus below/above when at the last line (and also works when the code is hidden)
* When you are at the last line and not the last character, Vim will automatically move you to the last character. This is the default in the [Vim extension](https://github.com/replit/codemirror-vim) we are using and the same behavior in https://colab.research.google.com/ - we can consider changing this in another PR
* When moving below we place the cursor at the first line, first character of the destination cell
* When moving above we place the cursor at the last line, first character of the destination cell

This feels the best to me. Let me know if you have strong opinions against this.

peterwilton-ttam:
That seems great. Thanks so much for the quick response and changes!

Regarding these specifics:

> - When moving below we place the cursor at the first line, first character of the destination cell
> - When moving above we place the cursor at the last line, first character of the destination cell

I think that's a good solution. It's predictable and no inconvenience.

PR #667, opened 2024-01-26
improvement: focus-up/focus-down cells with vim

mscolnick:
This:
* Allows `j`/`k` to focus below/above when at the last line (and also works when the code is hidden)
* This improves `j`/`k` and `Up/Down` to focus on the first or last line, when come from above or below (instead of the previous line)
* add padding to the bottom of the cells when editing to you can scroll the bottom cell to the middle (request in Discord)
* fix the `scrollActiveLineIntoView`. this was probably regressed a while ago when we added the app footer

Fixes https://github.com/marimo-team/marimo/issues/658

### DIFF
--- a/frontend/src/components/editor/renderers/CellArray.tsx
+++ b/frontend/src/components/editor/renderers/CellArray.tsx
@@ -90,7 +90,12 @@ export const CellArray: React.FC<CellArrayProps> = ({
   const cells = flattenNotebookCells(notebook);
 
   return (
-    <VerticalLayoutWrapper invisible={invisible} appConfig={appConfig}>
+    <VerticalLayoutWrapper
+      // 'pb' allows the user to put the cell in the middle of the screen
+      className="pb-[40vh]"
+      invisible={invisible}
+      appConfig={appConfig}
+    >
       <NotebookBanner />
       {cells.map((cell) => (
         <Cell
--- a/frontend/src/core/cells/scrollCellIntoView.ts
+++ b/frontend/src/core/cells/scrollCellIntoView.ts
@@ -5,11 +5,17 @@ import { CellId, HTMLCellId } from "./ids";
 import { CellHandle } from "@/components/editor/Cell";
 import { CellConfig } from "./types";
 
-export function focusAndScrollCellIntoView(
-  cellId: CellId,
-  cell: RefObject<CellHandle>,
-  config: CellConfig
-) {
+export function focusAndScrollCellIntoView({
+  cellId,
+  cell,
+  config,
+  codeFocus,
+}: {
+  cellId: CellId;
+  cell: RefObject<CellHandle>;
+  config: CellConfig;
+  codeFocus: "top" | "bottom" | undefined;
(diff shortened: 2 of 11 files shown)


frontend/src/components/editor/renderers/CellArray.tsx
```diff
@@ -90,7 +90,12 @@ export const CellArray: React.FC<CellArrayProps> = ({
   const cells = flattenNotebookCells(notebook);
 
   return (
-    <VerticalLayoutWrapper invisible={invisible} appConfig={appConfig}>
+    <VerticalLayoutWrapper
+      // 'pb' allows the user to put the cell in the middle of the screen
+      className="pb-[40vh]"
```
mscolnick, line 95: requested by @baggiponte, in the discord a while back

frontend/src/core/cells/scrollCellIntoView.ts
```diff
@@ -24,7 +30,30 @@ export function focusAndScrollCellIntoView(
...
+    } else if (codeFocus === "bottom") {
+      // If codeFocus is bottom, move the cursor to the bottom of the editor,
+      // but front of the last line.
+      const lastLine = editor.state.doc.line(editor.state.doc.lines);
+      editor.dispatch({
+        selection: {
+          anchor: lastLine.from,
+          head: lastLine.from,
+        },
+      });
```
mscolnick, line 55: focus the correct line when coming from above/below

frontend/src/core/codemirror/extensions.ts
```diff
@@ -86,10 +87,9 @@ export function scrollActiveLineIntoView() {
       // Only scroll if there is an active line
       if (activeLines.length === 1) {
         const activeLine = activeLines[0] as HTMLElement;
-        smartScrollIntoView(activeLine, {
-          top: 30,
-          bottom: 120,
-        });
+        const appEl = document.getElementById("App");
+        invariant(appEl, "App not found");
+        smartScrollIntoView(activeLine, { top: 30, bottom: 150 }, appEl);
```
mscolnick, line 92: this was a bug that i regressed when adding the app footer a while back. this makes it so your last line it not at the bottom, but rather 150px from the bottom

frontend/src/core/codemirror/keymaps/vim.ts
```diff
@@ -0,0 +1,40 @@
...
+        run: (ev) => {
+          if (isAtStartOfEditor(ev) && isInVimNormalMode(ev)) {
+            callbacks.focusUp();
+            return true;
+          }
+          return false;
+        },
+      },
+    ]),
+  ];
```
mscolnick, line 39: this works in Mac - i'd be curious if it fails again in Linux

frontend/src/components/editor/renderers/CellArray.tsx
baggiponte: Oh noice!

akshayka (approved):

PR #668, opened 2024-01-26
feat: Support __file__ in cells

mscolnick:


marimo/_runtime/runtime.py
```diff
@@ -588,6 +592,12 @@ def mutate_graph(
 
     def _run_cells(self, cell_ids: set[CellId_t]) -> None:
         """Run cells and any state updates they trigger"""
+
+        # Set __file__ to the globals
+        # we do this dynamically because the user may change the filename
+        # of the notebook
+        self.globals["__file__"] = self.app_metadata.filename
+
```
akshayka, line 600: Let's do this at kernel construction time, to match what the Python interpreter does (that way we'll also match the behavior when running as a script, e.g. if a user did something weird like overwriting `__file__`). It's okay for this to be stale if the user changes the filename.

akshayka (commented):
nice -- i'd prefer doing this at notebook construction time though, prefer not to set variables at runtime. i think it's okay if it's stale when the user changes the filename. or we could do it at runtime I guess, with a caveat -- see my other comment

marimo/_runtime/runtime.py
akshayka: If we really want to update it at runtime, we can check to make sure the user hasn't written to `__file__` themselves. If they have, then don't update; otherwise, fine to update.

mscolnick: made a change. i don't do it dynamically on each run, but i do set `__file__` whenever it gets changes via a Request

marimo/_runtime/runtime.py
```diff
@@ -845,6 +851,13 @@ def set_cell_config(self, request: SetCellConfigRequest) -> None:
                 dataflow.transitive_closure(self.graph, cells_to_run)
             )
 
+    def set_app_metadata(self, request: UpdateAppMetadataRequest) -> None:
+        """Update app metadata."""
+        self.app_metadata = request.metadata
+
+        # Set __file__ to the globals
+        self.globals["__file__"] = self.app_metadata.filename
```
akshayka, line 859: Don't overwrite `__file__` if the user has defined it (double underscore variables are not local). 

```suggestion
        if "__file__" not in self.graph.definitions:
            self.globals["__file__"] = self.app_metadata.filename
```

marimo/_server/sessions.py
```diff
@@ -333,12 +350,21 @@ def create_session(
...
             return s
         else:
             return self.sessions[session_id]
 
+    def _get_filename(self) -> Optional[str]:
+        if self.filename is None:
+            return None
+        try:
+            return os.path.abspath(self.filename)
+        except AttributeError:
```
akshayka, line 365:
```suggestion
        except Exception as e:
            LOGGER.error("Failed to get absolute path for filename %s", str(self.filename))       
```

tests/_runtime/test_globals.py
akshayka, line 1: Let's put this in `test_runtime.py`, to follow the convention of one test file for each Python module.

marimo/_server/sessions.py
```diff
@@ -322,6 +334,11 @@ def rename(self, filename: Optional[str]) -> None:
         Should be called if an api call renamed the current file on disk.
         """
         self.filename = filename
+        self.app_metadata.filename = self._get_filename()
+        for _, session in self.sessions.items():
```
akshayka, line 338:
```suggestion
        for session in self.sessions.values():
```

akshayka (commented):
Looks mostly good, just a few comments

marimo/_runtime/runtime.py
akshayka: On the other hand, then the frontend would need to register a rename as a run, and the backend needs to send a completed run when  it's done renaming, which is kind of weird.

After giving it some thought, I vote for only writing to __file__ at kernel creation time, to not violate our guarantee that program state and outputs are consistent. I think it's simpler, matches CPython, and good enough for most use cases.

akshayka (approved):
Looks good to me except for the one unresolved comment about moving the test

PR #669, opened 2024-01-26
fix: dont move between cells if auto-complete is open

mscolnick:
Previously you can accidentally go to the next cell when using arrow-keys inside the auto-complete


akshayka (approved):
🎉

PR #674, opened 2024-01-27
feat: add --host to CLI

mscolnick:


marimo/_server/api/lifespans.py
```diff
@@ -102,8 +102,9 @@ async def lsp(app: Starlette) -> AsyncIterator[None]:
 
 @contextlib.asynccontextmanager
 async def open_browser(app: Starlette) -> AsyncIterator[None]:
-    session_mgr = get_manager()
-    url = f"http://localhost:{session_mgr.port}"
+    host = app.state.host
+    port = app.state.post
```
akshayka, line 106:
```suggestion
    port = app.state.port
```

marimo/_cli/cli.py
```diff
@@ -318,6 +339,13 @@ def recover(name: str) -> None:
     type=int,
     help="Port to attach to.",
 )
+@click.option(
+    "--host",
+    default="127.0.0.1",
```
akshayka, line 344: @mscolnick Can we continue to use `localhost` as the default? Or at least print `localhost` at the CLI when `host==127.0.0.1`? I think that's friendlier for our users.

akshayka (commented):
Left 1 comment. CI was failing, maybe due to a typo, which I fixed

marimo/_cli/cli.py
mscolnick: Most modern web frameworks have moved from `localhost` to explicitly `127.0.0.1`. I've seen this on both Python (starlette, fastapi) and Node (vite). I am sure there is good reason, so I would advise to not make localhost the default.

For printing, most every case localhost will be aliased to `127.0.0.1`, but it is possible that the user had changed the alias. I'm fine to still print `localhost` as the default if you think it is friendlier

akshayka (approved):

PR #675, opened 2024-01-27
Add testcore, testoptional deps to pyproject.toml

akshayka:
- To facilitate CI testing with minimal possible set of dependencies.
- Refactor pandas/altair tests to only run when deps installed
- Updates CI to run tests with and without the optional dependencies, to make sure we don't accidentally require an optional dep.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ testcore = [
 
 testoptional = [
     # For testing mo.ui.chart
+    "altair>=5.0.0",
     "pandas>=1.3.0",
     "pandas-stubs>=1.3.0",
     # polars 0.19.13 requires building maturn from source, but we don't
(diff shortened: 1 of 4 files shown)


mscolnick (approved):

pyproject.toml
```diff
@@ -92,6 +92,23 @@ dev = [
...
+    "pytest-codecov~=0.5.1"
+]
+
+testoptional = [
+    # For testing mo.ui.chart
+    "pandas>=1.3.0",
+    "pandas-stubs>=1.3.0",
+    # polars 0.19.13 requires building maturn from source, but we don't
+    # have the rust toolchain installed on CI
+    "polars==0.19.12",
```
mscolnick, line 109: do we need altair?

akshayka: added, thanks

PR #700, opened 2024-01-31
feat: file system

mscolnick:
This PR adds a file browser panel. Currently it only lets you click-to-copy or drag-n-drop onto a cell's editor which gives a snippet to read the file in Python. This will however allow us to do more things in future such as compose marimo notebooks or open new notebooks.

* Adds `FileSystem` and `OSFileSystem`. In future, we can connect to other backends. 
* Adds starlette endpoints
* Add a FE panel 
* Adds a new codemirror extension for drag-n-drop


akshayka (commented):
Pretty cool! Two comments to address.

- [ ] in `endpoints/file_explorer.py`, let's wrap all the filesystem calls in a `try/except` (could except `OSError`). Operating systems are finicky, so in case one of these calls doesn't work, we'd still want the server to be alive.
- [ ] For the automatically generated `open` code, we don't want users to accidentally leave their files open. Let's generate `with` statements instead. Examples:

```python
with open("path/to/my/file.txt", "r") as _f:
   ...
```

```python
with open("path/to/my/file.json", "r") as _f:
    _data = json.load(_f)
```

A suggestion:

- [ ] For the "Copied to clipboard" toast, let's use the following text: "Code to open the file has been copied to your clipboard. You can also ..." When I read the banner I thought the contents of the file were copied to my clipboard. 

An idea, up to you whether to implement now or later:

- [ ] For marimo notebooks, use a different, notebook-like icon.

mscolnick:
@akshayka 

1. Already handled by our middleware
2/3 Done
4. Will do in a followup when marimo apps mean something in the sidebar

Issue #595, opened 2024-01-17
reload app when file updated externally


alefminus:
### Description

I have a marimo app running on a computer and I want to be able to replace the underlying app.py notebook and have the existing "marimo run" instance notice the updated file and reload the connected browsers (a single one in my case).

The use case is a computer used as a wall mounted dashboard where I want to update the dashboard ocasionally.

Alternatively I can kill the existing "marimo run" process but then I need to restart the client since there is no refresh on it either (it shows the "blood spots" background of a killed kernel, but does not connect when "marimo app" is run listening on the same port again).

### Suggested solution

use inotify (linux specific, can poll or use similar API on other platforms) to check updates to the notebook.
On update, rerun the notebook (wholesale, do not attempt to discover the changed cells and only run them - that would be cool but also potentially wrong) and notify all clients to reload as well.
This can be behind a flag, i.e.

```
marimo app --reload-on-update
```

### Alternative

Restarting the "marimo app" process and the corresponding browser tab.

mscolnick:
This sounds like a useful feature. (I would vote `--watch` as the flag)

@alefminus, is it just the the file running that you want reloaded? or would you potentially want to include downstream files (e.g. a csv, log file, etc). Although, you could probably put a reload/refresh on those files within your marimo notebook itself.

In the meantime, curious if this would work on Linux (haven't tested since I'm on a Mac)
`while inotifywait -e close_write app.py; marimo run app.py; done`

PR #722, opened 2024-02-02
feat: file watcher in run mode (--watch)

mscolnick:
Closes https://github.com/marimo-team/marimo/issues/595

This adds a filewatcher to `marimo run --watch`. If `watchdog` is installed, we use that. Otherwise we poll every 1s. We can update this poll via config in future depending on feedback.


akshayka (dismissed):
LGTM! modulo playwright

PR #755, opened 2024-02-08
Take completion queue off the kernel's main thread ...

akshayka:
... and drain the completion queue on each completion.

This PR takes the completion queue off the kernel's main thread. Instead, it adds a completion queue to the server's QueueManager, and completion requests are enqueued by the server endpoint (so that they reach the completion worker without going through the kernel).

As a result, completions now work even when the kernel is executing code.

Additionally, this PR changes the completion worker to drain its queue when retrieving completions, processing only the newest completion requests and ignoring ones that it had received but has not yet processed.

As a result, stale completions that take a long time to process (such as for numpy, torch, tensorflow, ...) no longer waste the completion worker's time (previously we would process all requests, even ones the frontend no longer needed).

### DIFF
--- a/marimo/_server/sessions.py
+++ b/marimo/_server/sessions.py
@@ -59,7 +59,7 @@ SessionId = str
 class QueueManager:
     """Manages queues for a session."""
 
-    def __init__(self, use_multiprocessing: bool, provide_completions: bool):
+    def __init__(self, use_multiprocessing: bool):
         context = mp.get_context("spawn") if use_multiprocessing else None
 
         # Control messages for the kernel (run, set UI element, set config, etc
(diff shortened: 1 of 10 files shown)


marimo/_server/sessions.py
```diff
@@ -59,13 +59,25 @@
...
+
+        # Code completion requests are sent through a separate queue
+        if provide_completions:
+            self.completion_queue: QueueType[
+                requests.CompletionRequest
+            ] | None = (
+                context.Queue() if context is not None else queue.Queue()
+            )
+        else:
+            self.completion_queue = None
```
mscolnick, line 79: nit: is it expensive to make a queue that never gets used? i wouldnt imagine so. the code would look a lot cleaner / easier to read without `provide_completions`

mscolnick (dismissed):

mscolnick (approved):

PR #776, opened 2024-02-12
chore: slimmed down interfaces for streams

mscolnick:


marimo/_messaging/streams.py
```diff
@@ -198,14 +206,26 @@ def writelines(self, sequence: Iterable[str]) -> None:  # type: ignore[override]
...
+        self._watcher.start()
+        return
+
+    def stop(self) -> None:
+        self._watcher.stop()
+        return
+
+    def pause(self) -> None:
+        self._watcher.pause()
+        return
```
akshayka, line 219: These (start, stop, pause) shouldn't be public methods, since this class is exposed to the user (by means of patching `sys.stdout`). Same goes for `Stderr`.

marimo/_messaging/types.py
```diff
@@ -1,6 +1,63 @@
...
+    def start(self) -> None:
+        pass
+
+    @abc.abstractmethod
+    def stop(self) -> None:
+        pass
+
+    @abc.abstractmethod
+    def pause(self) -> None:
+        pass
```
akshayka, line 37: I'm not sure these (start, stop, pause) are useful methods for a generic `Stdout` type — they are sort of tied to the implementation of the pipe-based `Stdout` that we have today.

Instead, if you do want to introduce a generic abstract `Stdout` type, why not have methods that `Stdout`s must support? These would be the public methods that are on our `Stdout` implementation today.

Same for `Stderr`.

marimo/_messaging/streams.py
```diff
@@ -45,7 +51,7 @@
 STD_STREAM_MAX_BYTES = int(os.getenv("MARIMO_STD_STREAM_MAX_BYTES", 1_000_000))
 
 
-class Stream:
+class ConcurrentStream(Stream):
```
akshayka, line 54: Instead of `ConcurrentStream`, how about something more descriptive like `PipeStream` or `ThreadSafeStream`? Same for Stdout/Stderr

akshayka (commented):
A few comments

marimo/_messaging/streams.py
mscolnick: fixed

marimo/_messaging/types.py
mscolnick: fixed

marimo/_messaging/types.py
```diff
@@ -1,6 +1,39 @@
...
+    """
+
+    cell_id: Optional[CellId_t] = None
+
+    @abc.abstractmethod
+    def write(self, op: str, data: Dict[Any, Any]) -> None:
+        pass
+
+
+class Stdout:
```
akshayka, line 26: Since we're using this as the expected type in various places, let's add the other public methods that Stdout should have. Same for Stderr and Stdin. In particular the public methods that exist on ThreadsafeStd*

akshayka (commented):
One comment, otherwise looks good

akshayka (approved):

PR #780, opened 2024-02-13
experimental: wasm/pyodide support

mscolnick:


marimo/_runtime/runtime.py
```diff
@@ -1079,6 +1079,36 @@ def instantiate(self, request: CreationRequest) -> None:
...
+            FunctionCallResult(
+                function_call_id=request.function_call_id,
+                return_value=ret,
+                status=status,
+            ).broadcast()
+            CompletedRun().broadcast()
+        elif isinstance(request, DeleteRequest):
+            self.delete(request)
+        elif isinstance(request, StopRequest):
+            return None
```
akshayka, line 1108: `StopRequest` should make the Kernel's control loop quit. This is used for stopping kernels in run mode.

marimo/_runtime/runtime.py
```diff
@@ -1214,34 +1244,10 @@ def sigterm_handler(signum: int, frame: Any) -> None:
...
-                status=status,
-            ).broadcast()
-            CompletedRun().broadcast()
-        elif isinstance(request, DeleteRequest):
-            kernel.delete(request)
-        elif isinstance(request, StopRequest):
-            break
-        else:
-            raise ValueError(f"Unknown request {request}")
+        kernel.handle_message(request)
```
akshayka, line 1247: Should break out of this loop on a `StopRequest`

marimo/_utils/typed_connection.py
```diff
@@ -1,15 +1,20 @@
...
+from typing import TYPE_CHECKING, Generic, TypeVar
+
+if TYPE_CHECKING:
+    from multiprocessing.connection import Connection
 
 T = TypeVar("T")
 
 
-class TypedConnection(Generic[T], Connection):
+class TypedConnection(Generic[T]):
```
akshayka, line 12: Why the change to this class? I didn't see it used in the `pyodide` code path

akshayka (approved):
LGTM except for Playwright

marimo/_utils/typed_connection.py
mscolnick: i wasnt able to mock out `Connection` and this gets used at file/module import, instead of lazily. this refactor just made it more lazy (and switching to composition over inheritance)

PR #795, opened 2024-02-17
feat: top-level await

akshayka:
This change makes it possible to use top-level `await` in marimo cells.

**Motivation.**

The main motivation for this change is to enable users to reliably install packages when running WebAssembly notebooks: the `micropip.install()` function is asynchronous, and must be awaited. Unfortunately, when running under `pyodide`, one cannot simply use `asyncio.run()` or any other similar method to block on completions of coroutines ([see Pyodide docs](https://pyodide.org/en/stable/usage/api/python-api/webloop.html#pyodide.webloop.WebLoop)). In contrast, `await`-ing the coroutine does the right thing.

With this change, users can write code like:

```python
import micropip
await micropip.install("plotly")
import plotly
```

As an alternative, I explored enabling WebAssembly stack-switching and using Pyodide's [`syncify`](https://pyodide.org/en/stable/usage/api/python-api/webloop.html#pyodide.webloop.WebLoop) method, which should block until a `Future` is completed. This would have required fewer changes to our code; unfortunately, this feature is not yet well-supported across browsers, and I couldn't even get it to work in Chrome.

For regular notebooks that don't use WebAssembly, top-level await makes it a bit easier to use async functions. ([Jupyter and IPython added support for this feature 6 years ago](https://blog.jupyter.org/ipython-7-0-async-repl-a35ce050f7f7).) However, due to an implementation decision (see next section), this does mean that library code or notebook code that calls `asyncio.get_running_loop().run_until_complete()` or `asyncio.run()` will no longer work -- Jupyter/ipykernel has the same problem.

**Design/Implementation/Tradeoffs.**

A few changes were made to support this feature:

1. Cells that use top-level await are serialized as coroutine functions (`async def ...`) in the file format.
2. The compiler was modified to allow top-level await, using Python's compiler flags.
3. An async cell execution method was added; everything that depends on it was also converted to coroutine functions, in particular the cell runner's `run` method and various `Kernel` methods.
4. In non-WASM notebooks, the Kernel's control loop now runs in an asyncio event loop. I decided to have the event loop alive for the entirety of the control loop (what Jupyter does), instead of starting and stopping it for each cell execution (what IPython terminal does) ([see docs](https://ipython.readthedocs.io/en/stable/interactive/autoawait.html#using-autoawait-in-a-notebook-ipykernel)).

Point number 4 has some tradeoffs that we should consider:

- ~background tasks started by user code *might* work, although they might crash the kernel if they try to access `stdout/stderr` when no execution context is installed~ Tasks made with `create_task()` are only scheduled if the user yields at some point (eg, by sleeping with `asyncio`); this behavior should not be relied upon ...
- `asyncio.get_running_loop().run_until_complete()` or `asyncio.run()` will no longer work in user code. If we want to support these functions, we could instead do what IPython terminal does -- only run user code in an event loop if the user code is a coroutine function. I am not sure if it will be possible to do this when running under pyodide, however.

**Other notes.**

- Verified that `pdb`'s code introspection still work. ~I did a spot-check and `pdb` still appears to be working (top-level await caused issues for `pdb` in IPyKernel, causing it to show the wrong stack frame/code) -- this should be double checked.~ 
- User interrupts are translated into `asyncio CancellationError`s. Unfortunately, the interrupt is not immediate, (`e.g.`, when interrupting an `asycio.sleep()` call, the `sleep` task finishes before the surrounding coroutine is cancelled), but still the interrupt does generally the right thing.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -1245,17 +1245,22 @@ def launch_kernel(
         else:
             signal.signal(signal.SIGTERM, sigterm_handler)
 
-    while True:
-        try:
-            request = control_queue.get()
-        except Exception as e:
-            # triggered on Windows when quit with Ctrl+C
-            LOGGER.debug("kernel queue.get() failed %s", e)
-            break
-        LOGGER.debug("received request %s", request)
-        if isinstance(request, StopRequest):
-            break
-        asyncio.run(kernel.handle_message(request))
+    # The control loop is asynchronous only because we allow
+    # user code to use top-level await; nothing else is awaited.
+    async def control_loop():
+        while True:
+            try:
+                request = control_queue.get()
+            except Exception as e:
+                # triggered on Windows when quit with Ctrl+C
+                LOGGER.debug("kernel queue.get() failed %s", e)
+                break
+            LOGGER.debug("received request %s", request)
+            if isinstance(request, StopRequest):
+                break
+            await kernel.handle_message(request)
+
+    asyncio.run(control_loop())
 
     if stdout is not None:
         stdout._watcher.stop()
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -176,6 +176,11 @@ class Kernel:
     Args:
 
     - cell_configs: initial configuration for each cell
+    - app_metadata: metadata about the notebook
+    - stream: object used to communicate with the server/outside world
+    - stdout: replacement for sys.stdout
+    - stderr: replacement for sys.stderr
+    - stdin: replacement for sys.stdin
     - input_override: a function that overrides the builtin input() function
     """
 
@@ -841,9 +846,12 @@ class Kernel:
     ) -> None:
         """Run cells and their descendants.
 
+
         The cells may be cells already existing in the graph or new cells.
         Adds the cells in `execution_requests` to the graph before running
         them.
+
+        Cells may use top-level await, which is why this function is async.
         """
 
         await self._run_cells(
@@ -1245,8 +1253,6 @@ def launch_kernel(
         else:
             signal.signal(signal.SIGTERM, sigterm_handler)
 
-    # The control loop is asynchronous only because we allow
-    # user code to use top-level await; nothing else is awaited.
     async def control_loop() -> None:
         while True:
             try:
@@ -1260,6 +1266,10 @@ def launch_kernel(
                 break
             await kernel.handle_message(request)
 
+    # The control loop is asynchronous only because we allow user code to use
+    # top-level await; nothing else is awaited. Don't introduce async
+    # primitives anywhere else in the runtime unless there is a *very* good
+    # reason; prefer using threads (for performance and clarity).
     asyncio.run(control_loop())
 
     if stdout is not None:
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -1,6 +1,7 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
+import asyncio
 import builtins
 import contextlib
 import dataclasses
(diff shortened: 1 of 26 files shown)


tests/_ast/codegen_data/test_generate_filecontents_async.py
```diff
@@ -0,0 +1,30 @@
...
+    xx = 1
+    await asyncio.sleep(1)
+    return x, xx
+
+
+@app.cell
+def three(asyncio, x):
+    async def _():
+        await asyncio.sleep(x)
+    return
```
mscolnick, line 26: should we test unparsable async code too? i assume it doesnt add the async, but still a good case to cover

mscolnick: also maybe a test case for when someone adds await, without async (e.g. they manually edited the file), would it recorrect itself?

tests/_runtime/test_virtual_file.py
```diff
@@ -185,19 +189,21 @@ def create_vfile(arg):
...
-    k.run([make_vfile])
+    await k.run([make_vfile])
     # the previous vfile should not be in the registry
     assert vfile not in ctx.virtual_file_registry.registry
     assert len(ctx.virtual_file_registry.registry) == 1
 
 
-def test_cached_vfile_disposal(k: Kernel, exec_req: ExecReqProvider) -> None:
-    k.run(
+async def test_cached_vfile_disposal(
```
mscolnick, line 203: i thought we had to add `@pytest.mark.asyncio` to all the tests. was there a config change that allowed us to remove this? should we remove other instances of `@pytest.mark.asyncio`?

akshayka: There was a config change that enables us to omit this. It's probably fine to keep the existing marks.

https://pytest-asyncio.readthedocs.io/en/latest/concepts.html#auto-mode

Configured in pyproject.toml

tests/_ast/codegen_data/test_generate_filecontents_async.py
akshayka: Added a test for the first comment. For the second comment, the module fails to load with a syntax error (`await` has to be used inside an `aync` function, unless a special compiler flag is set), so didn't add a test for that.

akshayka:
Outstanding tasks/issues:

- [ ] attempting to interrupt execution of a coroutine breaks the kernel
- [ ] tasks created with `create_task` are only executed when a foreground async task is running. In particular this doesn't print any output:

```python
import marimo

__generated_with = "0.2.5"
app = marimo.App()


@app.cell
def __():
    import asyncio
    return asyncio,


@app.cell
def __():
    async def aprint(msg):
        print(msg)
    return aprint,


@app.cell
def __(aprint, asyncio):
    asyncio.create_task(aprint("hello"))
    return


if __name__ == "__main__":
    app.run()
```

marimo/_runtime/runtime.py
```diff
@@ -1236,17 +1253,24 @@ def sigterm_handler(signum: int, frame: Any) -> None:
...
-            LOGGER.debug("kernel queue.get() failed %s", e)
-            break
-        LOGGER.debug("received request %s", request)
-        if isinstance(request, StopRequest):
-            break
-        kernel.handle_message(request)
+    async def control_loop() -> None:
+        while True:
+            try:
+                request = control_queue.get()
```
mscolnick, line 1259: is there any benefit in moving some of the queues to async queues?

mscolnick: i guess this queue is across threads so nvm

mscolnick (approved):

marimo/_runtime/runtime.py
akshayka: I thought about that. But I wasn't really sure how to do it easily, bc we're still communicating across processes. Using an async queue will probably also have a slight performance hit, though probably negligible.

PR #811, opened 2024-02-21
fix: pickle-ability of functions/classes

akshayka:
Replace the `__main__` module with a dummy module, and use this dummy module's `__dict__` as our globals, to fix an issue with pickling (only functions defined at module scope can be pickled).

I ran into this issue when testing top-level await. Another one of our users also found this issue and mentioned it in a blog post.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -202,6 +202,8 @@ class App:
 
         self._maybe_initialize()
 
+        # No need to provide `file`, `input_override` here, since this
+        # function is only called when running as a script
         with patch_main_module_context() as module:
             glbls = module.__dict__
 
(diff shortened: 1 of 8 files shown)


marimo/_ast/app.py
```diff
@@ -200,28 +201,34 @@ def run(self) -> tuple[Sequence[Any], dict[str, Any]]:
...
-            tuple(outputs[cid] for cid in self._cell_manager.valid_cell_ids()),
-            # omit defs that were never defined at runtime, eg due to
-            # conditional definitions like
-            #
-            # if cond:
-            #   x = 0
-            {name: glbls[name] for name in self._defs if name in glbls},
-        )
+
+        with patch_main_module_context() as module:
```
mscolnick, line 205: do you need to add the file and input_override here?

mscolnick (dismissed):

marimo/_ast/app.py
akshayka: This function is only called when running as a script (for now), so no we don't need to. Added a comment

Issue #826, opened 2024-02-23
Enhancement: Pin scroll to bottom of console ouotput area


akshayka:
### Description

When printing many lines in a cell, the console output area's scroll is pinned to the top, making it difficult to see new lines as they are printed. For example:

```python
import time
for i in range(1000):
    time.sleep(0.1)
    print(i)
```

This issue was raised in Discord, by a user training a model with PyTorch.

### Suggested solution

Pin scroll to the bottom of the console output area.

PR #828, opened 2024-02-23
improvement: Add scroll lock to bottom of console output

mscolnick:
Closes https://github.com/marimo-team/marimo/issues/826

```
// Keep scroll at the bottom if it is within 120px of the bottom,
// so when we add new content, it will lock to the bottom
// This won't handle large jumps in the scroll position
// if there is a lot of content added at once.
```

### DIFF
--- a/frontend/src/components/editor/code/readonly-python-code.tsx
+++ b/frontend/src/components/editor/code/readonly-python-code.tsx
@@ -25,7 +25,7 @@ export const ReadonlyPythonCode = memo(
     const [hideCode, setHideCode] = useState(initiallyHideCode);
 
     return (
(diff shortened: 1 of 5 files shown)


akshayka (approved):
Thanks! LGTM

FYI I did find this https://stackoverflow.com/questions/18614301/keep-overflow-div-scrolled-to-bottom-unless-user-scrolls-up

and it seemed to work ... but it would require us to change `display: flow-root` (which we use to prevent floated elements from extending out of the output area) to `display: flex` for `.console-output-area`, in `Cell.css`

mscolnick:
Oh that's a much better solution. I can actually get it to work with that. Thanks for the pointer

PR #841, opened 2024-02-27
Fix: form internal value

### LINKED ISSUE #840: `ui.form()` initial value not updated when re-rendered
### Describe the bug

This issue was raised on Discord by @ztawil.

When a `form` element is re-rendered without unmounting/remounting, its initial value in the frontend isn't updated, causing it to return the incorrect value on submit.

This can be seen by populating the form's initial value with a UI element's value, changing the original UI element, and submitting the form.

### Environment

```
{
  "marimo": "0.2.8",
  "OS": "Linux",
  "OS Version": "6.5.0-15-generic",
  "Processor": "x86_64",
  "Python Version": "3.10.12",
  "Binaries": {
    "Chrome": "121.0.6167.184",
    "Node": "v18.13.0"
  },
  "Requirements": {
    "black": "23.3.0",
    "click": "8.1.3",
    "jedi": "0.19.1",
    "pymdown-extensions": "10.7",
    "starlette": "0.36.1",
    "tomlkit": "0.12.3",
    "typing_extensions": "4.9.0",
    "uvicorn": "0.27.0.post1"
  }
}
```

### Code to reproduce

```python
import marimo
 
__generated_with = "0.2.8"
app = marimo.App()
 
 
@app.cell
def __():
    import marimo as mo
    return mo,
 
 
@app.cell
def __(mo):
    mo.md("Modify the below text input to re-render the form, then hit submit")
    return
 
 
@app.cell
def __(mo):
    input = mo.ui.text()
    input
    return input,
 
 
@app.cell
def __(input, mo):
    input_form = mo.ui.text(value=input.value).form()
    input_form
    return input_form,
 
 
@app.cell
def __(input_form):
    input_form.value
    return
 
 
if __name__ == "__main__":
    app.run()
```

akshayka:
This PR handles an edge case in which the form's wrapped element changes its value without firing an event, due to Python re-creating the form

Fixes #840

### DIFF
--- a/frontend/src/plugins/impl/FormPlugin.tsx
+++ b/frontend/src/plugins/impl/FormPlugin.tsx
@@ -63,7 +63,7 @@ export const FormPlugin = createPlugin("marimo-form")
       clearButtonLabel: z.string().default("Clear"),
       clearButtonTooltip: z.string().optional(),
       shouldValidate: z.boolean().optional(),
-    }),
+    })
   )
   .withFunctions<Functions>({
     validate: rpc
@@ -127,7 +127,7 @@ export const FormWrapper = <T,>({
           const response = await validate({ value: newValue }).catch(
             (error_) => {
               setError(error_.message ?? "Error validating");
-            },
+            }
           );
           if (response != null) {
             setError(response);
@@ -178,7 +178,7 @@ export const FormWrapper = <T,>({
               >
                 {clearButtonLabel}
               </Button>,
-              clearButtonTooltip,
+              clearButtonTooltip
             )}
           {withTooltip(
             <Button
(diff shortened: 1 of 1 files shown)


frontend/src/plugins/impl/FormPlugin.tsx
```diff
@@ -225,6 +225,14 @@ const Form = ({
     setInternalValue(UI_ELEMENT_REGISTRY.lookupValue(elementId));
   }, [elementId]);
 
+  // Edge case: the Form plugin may be re-created in Python with the same
+  // wrapped `elementId`, meaning the value of the wrapped element
+  // can change without the plugin generating an event
+  const wrappedValue = UI_ELEMENT_REGISTRY.lookupValue(elementId);
+  if (UI_ELEMENT_REGISTRY.lookupValue(elementId) !== internalValue) {
```
mscolnick, line 232: could this be`if(wrappedValue !== internalValue)`

akshayka: Oops yep. fixed

PR #842, opened 2024-02-27
improvement: support completions in wasm

mscolnick:
Adds code completion to the wasm runtime. This is done on the same thread, but with another async task.

- we add a flag for `docstrings` limit to prevent searching for docstrings past 5 for performance reasons


akshayka (approved):
Nice!

PR #843, opened 2024-02-27
feat: import and run named cells

akshayka:
This PR makes it possible to import named cells from one notebook into another, and to run the imported cells programmatically to obtain their output and definitions.

For example:

```python
from my_notebook import my_cell

# refs is any subset of the cell's refs
refs: dict[str, Any] | None = { ... }
output, defs = my_cell.run(**refs)
```

This helps make notebooks more testable: you can import named cells in a `pytest` module, run it, and check their outputs or defs. It also also helps with code re-use, making it possible to re-use cells across notebooks, especially if the cells are written in a functional way.

The main changes are in `cell.py`: we've introduced a new public `Cell` class. Calling its `run()` method executes the cell's ancestors, and then the cell itself, returning a tuple of its output and a `Mapping` of def names to values. The caller can supply any subset of the cell's refs as keyword arguments to short-circuit / prevent computation of ancestors and run the cell with custom arguments.

The returned `defs` `Mapping` is an internal `_Namespace` object; the runtime has been updated so that `set-ui-element-value` requests on objects in namespaces trigger reactive execution of cells that reference the namespace. 

This feature has some limitations.

- Interacting with the returned output doesn't trigger reactivity inside the cell's containing `App` -- so if the output contains UI elements that have `on_change` handlers that set reactive state, the state updates will just be dropped / won't trigger reactive execution. Full reactivity is left for later, when we add `App` composition
- No caching is implemented. This means if two cells have common ancestors, running them without providing `refs` will recompute their ancestors. This is a conscious choice for correctness, but we can add an option to enable caching in the future.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -12,16 +12,12 @@ from typing import (
     Iterable,
     Literal,
     Optional,
-    Union,
-    cast,
 )
 
 from marimo import _loggers
 from marimo._ast.cell import (
+    Cell,
     CellConfig,
-    CellFunction,
-    CellFuncType,
-    CellFuncTypeBound,
     CellId_t,
     execute_cell_async,
 )
@@ -76,8 +72,8 @@ class CellData:
     # Cell config
     config: CellConfig
 
-    # Callable cell, or None if cell was not parsable
-    cell_function: Optional[CellFunction[CellFuncType]]
+    # The original cell, or None if cell was not parsable
+    cell: Optional[Cell]
 
 
 @mddoc
@@ -107,15 +103,12 @@ class App:
 
     def cell(
         self,
-        func: Optional[CellFuncTypeBound] = None,
+        func: Callable[..., Any] | None = None,
         *,
         disabled: bool = False,
         hide_code: bool = False,
         **kwargs: Any,
-    ) -> Union[
-        Callable[[CellFuncType], CellFunction[CellFuncTypeBound]],
-        CellFunction[CellFuncTypeBound],
-    ]:
+    ) -> Cell | Callable[[Callable[..., Any]], Cell]:
         """A decorator to add a cell to the app
 
         This decorator can be called with or without parentheses. Each of the
@@ -167,8 +160,8 @@ class App:
             return
 
         # Add cells to graph
-        for cell_id, cell_fn in self._cell_manager.valid_cells():
-            self._graph.register_cell(cell_id, cell_fn.cell)
+        for cell_id, cell in self._cell_manager.valid_cells():
+            self._graph.register_cell(cell_id, cell._cell)
         self._defs = self._graph.definitions.keys()
 
         try:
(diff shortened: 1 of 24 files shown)


marimo/_ast/cell.py
```diff
@@ -148,111 +144,211 @@ def set_status(self, status: CellStatusType) -> None:
...
+          is the cell's last expression and `defs` is a `dict` mapping the
+          cell's defined names to their values.
+        """
+        assert self._app is not None
+        if self._is_coroutine():
+            return self._app.run_cell_async(cell=self, kwargs=refs)
+        else:
+            return self._app.run_cell_sync(cell=self, kwargs=refs)
+
+    def __call__(self, *args: Any, **kwargs: Any) -> None:
```
akshayka, line 333: Decided against implementing this to avoid confusion, given that the signature of `run` doesn't match the signature of the cell in the file format.

marimo/_ast/cell.py
```diff
@@ -148,111 +144,211 @@ def set_status(self, status: CellStatusType) -> None:
...
+        ret = cell.run()
+        if isinstance(ret, Awaitable):
+            output, defs = await ret
         else:
-            n_args += 1
-
-    parameters = list(signature.parameters.keys())
-    return_names = sorted(defn for defn in cell.defs)
-
-    def _prepare_args(*args: Any, **kwargs: Any) -> dict[Any, Any]:
```
akshayka, line 189: This was an early attempt at runnable cells that was unused

marimo/_ast/cell.py
```diff
@@ -148,111 +144,211 @@ def set_status(self, status: CellStatusType) -> None:
...
 
-CellFuncType = Callable[..., Optional[Tuple[Any, ...]]]
-# Cumbersome, but used to ensure function types don't get erased in decorators
-# or creation of CellFunction
-CellFuncTypeBound = TypeVar(
-    "CellFuncTypeBound",
-    bound=Callable[..., Optional[Tuple[Any, ...]]],
-)
+@dataclasses.dataclass
+class Cell:
```
akshayka, line 148: The new `Cell` class, to be added to the public API. It also replaces `CellFunction` throughout our codebase

marimo/_plugins/ui/_core/registry.py
```diff
@@ -71,22 +73,34 @@ def _has_parent_id(
...
+        #
+        # Also introspects _Namespace objects, including the name of the
+        # _Namespace if it contains `object_id`
+        bindings: set[str] = set()
+        for name, value in glbls.items():
+            if isinstance(value, UIElement) and self._has_parent_id(
+                value, object_id
+            ):
+                bindings.add(name)
+            elif (
```
akshayka, line 91: This is the update that allows `set-ui-element-value` requests on objects inside a `_Namespace` returned by `cell.run()` to trigger reactive execution based on the containing `_Namespace` object.

There's a unit test in `test_runtime.py` that demonstrates this

marimo/_ast/app.py
```diff
@@ -238,6 +240,22 @@ async def _run_async(self) -> tuple[Sequence[Any], dict[str, Any]]:
...
         return asyncio.run(self._run_async())
 
+    async def _run_cell_async(
+        self, cell: Cell, kwargs: dict[str, Any]
+    ) -> tuple[Any, _Namespace]:
+        self._maybe_initialize()
+        output, defs = await self._runner.run_cell_async(
+            cell._cell.cell_id, kwargs
+        )
+        return output, _Namespace(defs)
```
mscolnick, line 250: i like the api for what is returned. its simple for the 99% case (just needing `output`) but `_Namespace` can be extended in future if we need to pass more info

marimo/_ast/cell.py
mscolnick: should we add a doc for the `Cell` class?

mscolnick (dismissed):

mscolnick (approved):

PR #844, opened 2024-02-27
improvement: improved shortcut symbols

mscolnick:
Use more symbols where possible. And add tooltip over the symbols

## After
Screenshot 2024-02-26 at 1 00 08 PM

## Before 
Screenshot 2024-02-26 at 1 00 17 PM


akshayka (dismissed):
Sweet! I like the tooltips too.

By the way, can we add a `Keyboard Shortcuts` button in the notebook actions menu, after `Command Palette`/before `Open Documentation`? Just realized the only way to access shortcuts now is through the Command Palette

image

Issue #852, opened 2024-02-28
Large integers lose precision in `mo.ui.table`


akshayka:
### Describe the bug

*Reported by one of our users*

Large numbers are rounded by `mo.ui.table`, which confused one of our users who had a column with very large integers and needed their values to be displayed precisely:

![image](https://uploads.linear.app/63f65126-890b-4810-98cb-b7f8eaee40ee/7fca09de-bfa2-4b23-aa5a-8967e4f1aac8/5e996dc0-1370-4055-884a-fba68735c5d4?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzYzZjY1MTI2LTg5MGItNDgxMC05OGNiLWI3ZjhlYWVlNDBlZS83ZmNhMDlkZS1iZmEyLTRiMjMtYWE1YS04OTY3ZTRmMWFhYzgvNWU5OTZkYzAtMTM3MC00MDU1LTg4NGEtZmJhNjg3MzVjNWQ0IiwiaWF0IjoxNzA5MDkwNzIwLCJleHAiOjMzMjc5NjUwNzIwfQ.rPFDqPaJZD-n5cY28aYyg4M6TE22cMDv1ZiwCe0pajs)

### Environment

Version `0.2.8`

### Code to reproduce

WASM notebook: [https://marimo.app/l/l2go45](https://marimo.app/l/l2go45)

```python
import marimo
 
__generated_with = "0.2.9"
app = marimo.App()
 
 
@app.cell
def __():
    import marimo as mo
    return mo,
 
 
@app.cell
def __():
    import pandas as pd
    return pd,
 
 
@app.cell
def __(pd):
    df = pd.DataFrame({"data": [912312851340981241284]})
    df
    return df,
 
 
@app.cell
def __(df, mo):
    table = mo.ui.table(df)
    table
    return table,
 
 
if __name__ == "__main__":
    app.run()
```

PR #853, opened 2024-02-28
fix: support big-ints in datatable

mscolnick:
Closes https://github.com/marimo-team/marimo/issues/852

Previously we used things like `parseFloat` and `parseInt` (via vega-utils) to parse our data. But this is lossy unfortunately, instead we are now using `BigInt`


marimo/_smoke_tests/bugs/852.py
```diff
@@ -0,0 +1,41 @@
...
+
+__generated_with = "0.2.9"
+app = marimo.App()
+
+
+@app.cell
+def __(mo):
+    mo.md(
+        """
+    # Bug 852
```
mscolnick, line 11: i do this often - i think it would be good practice to commit these

PR #858, opened 2024-02-28
improvement: add upload file (dnd and button), render csv in a table

mscolnick:
1. Add uploading a file (up to 50mb) via drag-n-drop or a button
2. Add a csv viewer that renders in our data-table.


akshayka (commented):
Cool! I tried opening a CSV but I think there was an uncaught exception somewhere, causing the frontend to be replaced with an error:

image

akshayka (approved):

PR #865, opened 2024-02-28
fix: opening notebooks from the UI

mscolnick:
We had multiple sources of truth for `filename` and `path`, and renaming the filename got `SessionManager` out of state. this makes `path` a computer property based on `filename`


frontend/src/components/editor/file-tree/file-explorer.tsx
```diff
@@ -163,10 +163,10 @@ const Show = ({ node }: { node: NodeApi<FileInfo> }) => {
     <span
       className="flex-1"
       onClick={(e) => {
-        e.stopPropagation();
         if (node.data.isDirectory) {
           return;
         }
+        e.stopPropagation();
```
mscolnick, line 169: fix so you can click the whole row to expand a folder

akshayka (approved):

PR #868, opened 2024-02-29
improvement: flush sync in wasm instead of async

mscolnick:
Pass down a message callback to flush sync instead of using async-iterators + queue to handle a message. This is actually a simpler implementation, but provides the benefit that std out and stream outputs appear right away instead of at the end of the cell's code.

### DIFF
--- a/marimo/_plugins/ui/_impl/altair_chart.py
+++ b/marimo/_plugins/ui/_impl/altair_chart.py
@@ -227,7 +227,11 @@ class altair_chart(UIElement[ChartSelection, "pd.DataFrame"]):
                 vega_spec["autosize"] = "fit-x"
 
         # Selection for binned charts is not yet implemented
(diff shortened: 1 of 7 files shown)


marimo/_plugins/ui/_impl/altair_chart.py
```diff
@@ -227,7 +227,11 @@ def __init__(
                 vega_spec["autosize"] = "fit-x"
 
         # Selection for binned charts is not yet implemented
-        if _has_binning(vega_spec):
+        has_chart_selection = chart_selection is not False
+        has_legend_selection = legend_selection is not False
+        if _has_binning(vega_spec) and (
+            has_chart_selection or has_legend_selection
+        ):
```
mscolnick, line 234: this was a small unrelated fix to remove an incorrect warning

PR #893, opened 2024-03-01
experimental: ai completion

mscolnick:


akshayka (commented):
Pretty cool ...! Some initial feedback.

- For hotkey ... how about `Meta+Shift+e`?

- When I accept completions (or hit `x` when a completion is available), my CodeMirror editor collapses/vanishes:

image

- UX: When I have the option to "Accept" a completion (like below), hitting `Enter` should accept the completion, instead of re-sending the completion

- `ESC` should close the AI completion window

- The "Accept" button is kind of hard to see

- "Accept" should be accompanied by a "Reject" button that rejects the completion, clears the diff, but keeps the completion window open (with the user's prompt still typed in it)

- For the diff: Can we show an inline diff instead of side-by-side? We could have an option to view the diff side-by-side, but I think it should be inline by default to support the default app width as well as smaller screens

- I wonder if we should try to design a system prompt that produces code that's marimo compatible. For example, instead of `plt.show()` at the end of a cell, in marimo we use `plt.gca()`

image

- Is the context for the completion prompt the entire notebook, or just the current cell? I think it should be the whole notebook

- We should think about how to make this available in WASM notebooks. In particular, how will users specify their API keys? I know you've given this a bit of thought. Can discuss offline ...

akshayka (dismissed):
# 🤖 🎶

image

PR #894, opened 2024-03-02
pyodide: lazily install black on first format

akshayka:
This makes code formatting work in WASM notebooks


mscolnick (approved):
nice super cool we can support that

PR #907, opened 2024-03-04
improvement: mock micropip when not running under pyodide

akshayka:
This change makes a mocked micropip available when not running under Pyodide. The mocked micropip's functions are noops (they just print warnings to the user that micropip is unavailable when not running under WASM). Still, this is better than outright failing, and is a step toward compatibility with Pyodide and CPython notebooks.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -223,7 +223,8 @@ class Kernel:
         self.state_updates: dict[State[Any], CellId_t] = {}
 
         # an empty string represents the current directory
-        exec("import sys; sys.path.append('')", self.globals)
+        patches.patch_micropip(self.globals)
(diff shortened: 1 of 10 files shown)


marimo/_runtime/runtime.py
```diff
@@ -222,8 +223,10 @@ def __init__(
         # was invoked. New state updates evict older ones.
         self.state_updates: dict[State[Any], CellId_t] = {}
 
+        if not is_pyodide():
+            patches.patch_micropip(self.globals)
         # an empty string represents the current directory
-        exec("import sys; sys.path.append('')", self.globals)
+        exec("import sys; sys.path.append(''); del sys", self.globals)
```
akshayka, line 229: we were accidentally leaking `sys` before, drive-by fix

mscolnick (approved):

PR #908, opened 2024-03-04
feat: mo.carousel and full screen slides

mscolnick:
Add `marimo.carousel`. 

Currently we don't expose any config, but provide some good defaults (fullscreen, keyboard nav, point nav, scroll dots)

Screenshot 2024-03-05 at 12 45 27 PM


akshayka (commented):
This is very cool!

I like the API. UX is also pretty good.

Can we lighten the gray bars on the side?

Additionally, after selecting all the data in the table, I managed to get in a weird state where an output from a previous slide overflowed into the current slide (this slide is only supposed to show "Thank You"). I think this only happened when using the arrow keys / keyboard to move between slides.

image

When I quit fullscreen with ESC, the text still says "Exit Fullscreen":

image

PR #919, opened 2024-03-06
improvement: patch `IPython.display.display` to call `mo.output.append`

akshayka:
This improves compatibility with code written for IPython.

Additionally, this adds a formatter for `IPython.display.HTML` -- we already some `display.HTML` objects by special casing `_repr_html_`, but this adds support for URLs. However, iframe'd URLs are not resizing; left a TODO.

Smoke test:

image

### DIFF
--- a/marimo/_output/formatters/ipython_formatters.py
+++ b/marimo/_output/formatters/ipython_formatters.py
@@ -1,0 +1,53 @@
+# Copyright 2024 Marimo. All rights reserved.
+from __future__ import annotations
+
+from typing import Any
+
+from marimo._messaging.mimetypes import KnownMimeType
+from marimo._output import builder
+from marimo._output.formatters.formatter_factory import FormatterFactory
+
+
+class IPythonFormatter(FormatterFactory):
+    @staticmethod
+    def package_name() -> str:
+        return "IPython"
+
+    def register(self) -> None:
+        import IPython.display
+
+        from marimo._output import formatting
+        from marimo._runtime.output import _output
+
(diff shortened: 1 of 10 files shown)


marimo/_output/formatters/ipython_formatters.py
```diff
@@ -0,0 +1,60 @@
...
+
+        def unpatch() -> None:
+            IPython.display.display = old_display
+
+        @formatting.formatter(IPython.display.HTML)
+        def _format_html(
+            html: IPython.display.HTML,
+        ) -> tuple[KnownMimeType, str]:
+            if html.url is not None:
+                # TODO(akshayka): resize iframe not working
```
mscolnick, line 47: this wont work for CORS - not sure if thats what you were testing, but that would be once reason. 

if you do same origin, it should work. you could make marimo a proxy for URLs (but i think that could be done later)

pyproject.toml
```diff
@@ -194,6 +195,7 @@ exclude = [
     'marimo/_tutorials/',
     'marimo/_smoke_tests/',
 ]
+warn_unused_ignores=false
```
mscolnick, line 198: ahhhhh! this was the worst without this since you couldn't do a blanket ignore b/c in another test matrix it would fail. and other matrices would have different ignores..

mscolnick (approved):

pyproject.toml
akshayka: haha yea it was getting really cumbersome

PR #921, opened 2024-03-08
fix: pyodide interrupts

akshayka:
This change gets interrupts working in WASM notebooks. The implementation uses `SharedArrayBuffer`, [as recommended by Pyodide](https://pyodide.org/en/stable/usage/keyboard-interrupts.html), which only works in secure contexts.

An example where interrupts now work:

```python
while True:
  ...
```

For some reason, `time.sleep()` cannot be interrupted -- this tripped me up for a while, because I thought my interruption implementation was incorrect.

### DIFF
--- a/frontend/src/core/pyodide/worker/types.ts
+++ b/frontend/src/core/pyodide/worker/types.ts
@@ -18,7 +18,7 @@ import type {
 export interface RawBridge {
   put_control_request(operation: object): Promise<string>;
   put_input(input: string): Promise<string>;
-  interrupt(): Promise<string>;
+  interrupt(buffer: Uint8Array): Promise<string>;
   code_complete(request: CodeCompletionRequest): Promise<string>;
   read_code(): Promise<{ contents: string }>;
   format(request: FormatRequest): Promise<FormatResponse>;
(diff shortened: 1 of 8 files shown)


marimo/_pyodide/pyodide_session.py
```diff
@@ -321,9 +315,9 @@ def launch_pyodide_kernel(
     )
 
     if is_edit_mode:
-        from marimo._output.formatters.formatters import register_formatters
-
-        register_formatters()
```
mscolnick, line 326: was this just not needed before?

frontend/src/core/pyodide/worker/types.ts
```diff
@@ -18,7 +18,7 @@ import type {
 export interface RawBridge {
   put_control_request(operation: object): Promise<string>;
   put_input(input: string): Promise<string>;
-  interrupt(): Promise<string>;
+  interrupt(buffer: Uint8Array): Promise<string>;
```
mscolnick, line 21: can we remove this?

mscolnick (approved):

marimo/_pyodide/pyodide_session.py
akshayka: I'm going to add this back, for both edit/run mode

frontend/src/core/pyodide/worker/types.ts
akshayka: oops yes

PR #925, opened 2024-03-09
fix: fix keyboard controls with carousel

### LINKED ISSUE #923: Arrows spin the carousel when trying to move cursor inside mo.ui.code_editor()
### Describe the bug

When there is a code_editor inside the carousel, using arrows inside that code_editor spins the carousel as well. I assume it is unintended.

### Environment

WASM notebook.
Brave [Version 1.63.169 Chromium: 122.0.6261.111 (Official Build) (64-bit)]

### Code to reproduce

_No response_

mscolnick:
- only use keyboard nav in fullscreen 
- prevent keypresses while inside components (e.g. `mo.ui.code_editor`) from moving the carousel

Fixes #923

### DIFF
--- a/frontend/src/plugins/layout/CarouselPlugin.tsx
+++ b/frontend/src/plugins/layout/CarouselPlugin.tsx
@@ -61,6 +61,11 @@ const CarouselComponent = ({
   const [isFullscreen, setIsFullscreen] = React.useState(false);
 
   useEventListener(document, "fullscreenchange", () => {
+    if (document.fullscreenElement) {
+      el.current?.swiper.keyboard.enable();
+    } else {
+      el.current?.swiper.keyboard.disable();
+    }
     setIsFullscreen(!!document.fullscreenElement);
   });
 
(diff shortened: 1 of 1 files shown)


frontend/src/plugins/layout/CarouselPlugin.tsx
```diff
@@ -99,6 +104,15 @@ const CarouselComponent = ({
...
             <div
+              onKeyDown={(e) => {
+                // If the target is from a marimo element, stop propagation
+                if (
+                  e.target instanceof HTMLElement &&
+                  e.target.tagName.toLocaleLowerCase().startsWith("marimo-")
+                ) {
+                  e.stopPropagation();
+                }
+              }}
```
akshayka, line 115: Could we do this with `onMouseDown` too and then re-enable `simulateTouch=True`?

akshayka (dismissed):

frontend/src/plugins/layout/CarouselPlugin.tsx
mscolnick: i just tried - got pretty close. it works with sliders, but not with altair, so lets just keep it off

PR #928, opened 2024-03-09
fix: function calls in wasm, add error state to pyodide bootstrap

mscolnick:
- Fix function calls in wasm (like from `mo.ui.dataframe`)
- add error state to pyodide bootstrap when bootstrap fails 

Screenshot 2024-03-08 at 7 06 13 PM

### DIFF
--- a/frontend/src/core/pyodide/bridge.ts
+++ b/frontend/src/core/pyodide/bridge.ts
@@ -110,6 +110,9 @@ export class PyodideBridge implements RunRequests, EditRequests {
     if (event.data.type === "initialized") {
       this.initialized.resolve();
     }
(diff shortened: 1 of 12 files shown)


frontend/src/core/pyodide/bridge.ts
```diff
@@ -286,9 +289,7 @@ export class PyodideBridge implements RunRequests, EditRequests {
   };
 
   sendFunctionRequest = async (request: SendFunctionRequest): Promise<null> => {
-    await this.putControlRequest({
```
mscolnick, line 289: this was the bug

akshayka (approved):

PR #934, opened 2024-03-11
improvement: patch urllib in bootstrap to support networking

akshayka:
Increases load time a bit, but makes `pd.read_csv` work out of the box. We could make this lazy later to optimize performance, but not worth doing right now.

With this change, `pd.read_csv(url)` works out of the box

### DIFF
--- a/frontend/src/core/pyodide/worker/bootstrap.ts
+++ b/frontend/src/core/pyodide/worker/bootstrap.ts
@@ -41,10 +41,17 @@ export async function bootstrap() {
         "${marimoWheel}",
         "markdown",
         "pymdown-extensions",
+        "pyodide_http",
       ],
       deps=False
       );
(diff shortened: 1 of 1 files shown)


mscolnick (approved):
nice

frontend/src/core/pyodide/worker/bootstrap.ts
```diff
@@ -41,11 +41,18 @@ export async function bootstrap() {
...
         "markdown",
         "pymdown-extensions",
+        "pyodide_http",
       ],
       deps=False
       );
     `);
 
+  // Patch the built-in urllib
+  pyodide.runPython(`
```
mscolnick, line 51: Do you know how long this takes? I assume milliseconds but this could be lazy as well

PR #935, opened 2024-03-11
dev release workflow for marimo python

akshayka:
This adds a workflow that creates a dev release on every push to main. The release is uploaded to test.pypi.org.

The uploaded version has the form `X.Y.dev<n-commits-since-last-tag>+g<short-sha>`.

I believe the latest dev release should then be available via `pip` with

`pip install --index-url https://test.pypi.org/simple/ --pre marimo`

This release doesn't publish new frontend packages; I'm not sure how we want to handle that.

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -7,7 +7,7 @@ on:
       - main
 jobs:
   upload_dev_wheel:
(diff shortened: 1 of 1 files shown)


mscolnick:
nice this is great!

PR #961, opened 2024-03-16
feat: package installation from pypi

akshayka:
This PR adds a basic UI to install packages using pip from the frontend. When packages are detected as missing (we use a `ModuleNotFoundError` as a heuristic), a list of missing packages is presented to the user, along with an option to install them using pip.

This is mostly for convenience -- it can be helpful when you spin up a new environment for prototyping or for testing a notebook someone shared with you, by eliminating the need to quit a marimo notebook, install the missing packages, and restart it.

If the user chooses to install, the UI starts installing the packages one by one, showing which package is currently being installed, which failed to install, and which succeeded. Installation of a package triggers re-running of the cell that imports it

Some notes:
- If an install of a module fails (perhaps the "package" was actually just a missing local module), the user won't be re-prompted to install it.
- We use a heuristic to convert from module names to package names, but this won't ever be perfect.
- We refuse to install packages if we detect the user has installed marimo in system Python.

Room for improvement:
- Specifying package versions
- Specifying package manager
- Sending console logs to the log viewer (they just print to the terminal)

image
image
image


marimo/_runtime/packages/package_manager.py
```diff
@@ -0,0 +1,134 @@
...
+
+from marimo._ast.cell import CellId_t
+from marimo._runtime.dataflow import DirectedGraph
+from marimo._runtime.packages.module_name_to_pypi_name import (
+    MODULE_NAME_TO_PYPI_NAME,
+)
+from marimo._utils.platform import is_pyodide
+
+
+class PackageManager:
```
mscolnick, line 18: could we make this an abstract class, and then have `PipPackageManager`. may feel like a pre-optimization, but we know its coming down the the road and will force a cleaner design

marimo/_runtime/packages/module_name_to_pypi_name.py
```diff
@@ -0,0 +1,758 @@
+# Copyright 2024 Marimo. All rights reserved.
+# Mapping from a module name to the corresponding package name # on PyPI
+MODULE_NAME_TO_PYPI_NAME = {
```
mscolnick, line 3: can this be lazy (maybe put under a function)? its not super large but we often won't need to read it

frontend/src/core/alerts/state.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Right now we only have one type of alert.
+ */
+interface AlertState {
+  packageAlert: MissingPackageAlert | InstallingPackageAlert | null;
+}
+
+const { reducer, createActions } = createReducer(
+  () => ({ packageAlert: null }) as AlertState,
+  {
+    addPackageAlert: (
```
mscolnick, line 44: i think you should create a shared typed `Identified` that extends `MissingPackageAlert`. This way you can share it with the message type too

```ts
type Identified<T> =  {id: string} & T

export interface MissingPackageAlert {
  kind: "missing";
  packages: string[];
  isolated: boolean;
}

export interface InstallingPackageAlert {
  kind: "installing";
  packages: PackageInstallationStatus;
}
```

```ts
interface AlertState {
  packageAlert: Identified<MissingPackageAlert> | Identified<InstallingPackageAlert> | null;
}
```

and in most cases, you just need `MissingPackageAlert`

also, i would create the uuid in the reducer (better for more logic there since its more testable than react components.

```ts
      return {
        ...state,
        packageAlert: {id: generatedUUID(), ...alert}
      };
```

frontend/src/components/editor/package-alert.tsx
```diff
@@ -0,0 +1,252 @@
...
+      </div>
+    );
+  } else {
+    logNever(packageAlert);
+    return null;
+  }
+};
+
+function getInstallationStatusElements(packages: PackageInstallationStatus) {
+  const statuses = Object.entries(packages).map(([_, status]) => status);
```
mscolnick, line 164: perf doesn matter here, but since you are just checking existence, you could use `new Set(Object.entries(packages).map(([_, status]) => status))`

frontend/src/components/editor/package-alert.tsx
```diff
@@ -0,0 +1,252 @@
...
+      titleIcon: <PackageXIcon className="w-5 h-5 inline-block mr-2" />,
+      description: "See terminal for error logs.",
+    };
+  }
+}
+
+const ProgressIcon = ({
+  status,
+}: {
+  status: "queued" | "installing" | "installed" | "failed";
```
mscolnick, line 198: could re-use `PackageInstallationStatus[string]`

marimo/_runtime/packages/package_manager.py
akshayka: I started doing this, but it felt premature. My preference is to abstract and refactor when we need to

mscolnick: I agree, we should avoid pre-optimizations, but this 1) makes it easier to test, and 2) welcomes contributions for other managers 

Also 3) don't we need this in this PR for micropip/wasm?

akshayka: Ok, sounds good

marimo/_runtime/packages/package_manager.py
```diff
@@ -0,0 +1,26 @@
...
+        """Canonicalizes a module name to a package name."""
+        ...
+
+    @abc.abstractmethod
+    def package_to_module(self, package_name: str) -> str:
+        """Canonicalizes a package name to a module name."""
+        ...
+
+    @abc.abstractmethod
+    async def install(self, package: str) -> bool:
```
mscolnick, line 21: is it worth installing multiple at once? its more performant and possibly more correct because of version compatibility

marimo/_server/api/endpoints/editing.py
```diff
@@ -77,3 +78,13 @@ async def stdin(request: Request) -> BaseResponse:
     app_state.require_current_session().put_input(body.text)
 
     return SuccessResponse()
+
+
+@router.post("/install_missing_packages")
+@requires("edit")
```
mscolnick, line 84: should we do this on `read` too when in pyodide?

marimo/_server/api/endpoints/editing.py
```diff
@@ -77,3 +78,13 @@ async def stdin(request: Request) -> BaseResponse:
     app_state.require_current_session().put_input(body.text)
 
     return SuccessResponse()
+
+
+@router.post("/install_missing_packages")
```
mscolnick, line 83: can we add one test in the endpoint tests for this? should be really easy

marimo/_runtime/packages/package_manager.py
akshayka: > possibly more correct because the package manager can find the right versions compatible across all the packages requested to be installed

Good point regarding correctness. Unfortunately we may lose the progress notifications in the UI, though, if we install multiple at once.

marimo/_server/api/endpoints/editing.py
akshayka: added a TODO

mscolnick (approved):

PR #972, opened 2024-03-21
fix: graceful shutdown for uvicorn==0.29.0

akshayka:
Fixes a bug in graceful shutdown caused by version 0.29.0 of uvicorn (which fixed a bug we were previously compensating for)


akshayka:
Fixed CLI tests, but broke other tests ... looking into it ...

mscolnick (approved):

PR #982, opened 2024-03-21
feat: Backspace on empty cell deletes cell

lmondada:
In normal mode, I suggest deleting a cell when it is empty backspace is pressed. If my memory serves me right, this feature exists in Pluto.jl. See if you like it, feel free to close this if you are not interested -- this was an excuse to familiarise myself with the codebase more than anything else.

I've had to add an argument to the `deleteCell` callback as removing the first cell results in a counter-intuitive UI (the cursor cannot move to the previous cell, as would be expected). I don't like the fact that `deleteCell(false)` isn't very explicit about what it is doing -- would you prefer that `deleteCell` takes a `{ deleteIfFirst?: boolean }` object? Or can you think of a better way of doing this? 

EDIT: I've turned the argument into an object, the resulting code is clearer imo. Feel free still to make further suggestions!

### DIFF
--- a/frontend/src/components/editor/cell/code/cell-editor.tsx
+++ b/frontend/src/components/editor/cell/code/cell-editor.tsx
@@ -88,13 +88,13 @@ const CellEditorInternal = ({
   const loading = status === "running" || status === "queued";
   const { sendToTop, sendToBottom } = useCellActions();
 
-  const handleDelete = useEvent(() => {
+  const handleDelete = useEvent((deleteIfFirst?: boolean) => {
     // Cannot delete running cells, since we're waiting for their output.
     if (loading) {
       return false;
     }
 
-    deleteCell({ cellId });
+    deleteCell({ cellId, deleteIfFirst });
     return true;
   });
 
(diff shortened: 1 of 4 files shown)


lmondada:
I think the second commit made the code clearer

akshayka:
@lmondada, thanks for contributing a PR!

We used to support `Backspace` for cell deletion, but later changed the hotkey to `Shift+Backspace` due to feedback we received. A user was holding backspace absentmindedly in an empty cell, perhaps after deleting some code, only to find that marimo continued deleting code in previous cells, causing him to accidentally delete several cells. Here's a video illustrating what I mean:

[Screencast from 03-21-2024 02:07:14 PM.webm](https://github.com/marimo-team/marimo/assets/1994308/eabab63f-add7-47e3-ad87-5067d1940e7e)

With `Shift+Backspace`, this doesn't happen, since the combination of these keys doesn't delete characters in an editor.

So unless we solve the issue illustrated in the video, I would prefer to keep the current hotkey.

lmondada:
Ah right, didn't realise there was shift+delete! Yes that's definitely just as good, and avoids the case you mention. I will close this.

Issue #992, opened 2024-03-22
Little bugs / Nix compatibility


dmadisetti:
### Describe the bug

Just a few things using marimo under nix that seem to be broken, these are more general code health things- so not dedicated nix specific fixes.

Rest assured that if marimo works under the pathological env that is nix, it will probably work pretty much everywhere!

### Environment

nixpkgs 0.3.4

### Code to reproduce

 - [x] `marimo env` crashes because pip is not guaranteed to be installed
 https://github.com/marimo-team/marimo/blob/38909fec3a56dfa5938c20a9e1b9801feb38c298/marimo/_cli/envinfo.py#L49-L84

proposal, tell user that there's nothing you can do OR extract from `import sys; sys.modules`

<details><summary>relevant nix stack trace</summary>

```
  Traceback (most recent call last):
  File "/nix/store/nr79v4j901yhlz29pwklbpxgz54jxzpn-python3.11-marimo-0.3.4/bin/.marimo-wrapped", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File "/nix/store/hza59ra675z8m5dc7kvms4j798h91i3f-python3-3.11.8-env/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/hza59ra675z8m5dc7kvms4j798h91i3f-python3-3.11.8-env/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/nix/store/hza59ra675z8m5dc7kvms4j798h91i3f-python3-3.11.8-env/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/hza59ra675z8m5dc7kvms4j798h91i3f-python3-3.11.8-env/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #999, opened 2024-03-25
bug fix: support env / status without pip (#992)

dmadisetti:
fixes expectation of pip being installed reported in #992

Exposed this data over the API, because I thought it would be nice to see from the notebook - but UI decisions are beyond anything I want to contribute


mscolnick (approved):
nice stuff!

mscolnick:
I can look into exposing this in the ui

PR #1006, opened 2024-03-27
feat: mo.query_params

### LINKED ISSUE #951: Support for URL parameters
### Description

It would be great if it were possible to pass url parameters to a running notebook. This would make it possible to create a shareable URL that would take the recipient right to where they need to be. For instance, by pre-filling a textbox, or providing a date-range.

### Suggested solution

The easiest (to implement) solution would be to simply expose the url parameters through a dictionary-like object in the `marimo` namespace. While something inspired by FastAPI would be nice, it would probably also be overkill - we're building quick interactive notebooks, doing some validation on input isn't that inconvenient.

Example of how the API could work. Example URL: `https://example.com/notebook?seed=42&country=us`

```
> marimo.query_params
{
    'seed': '42',
    'country': 'us'
}
```

### Alternative

_No response_

### Additional context

Streamlit uses a very similar approach: https://docs.streamlit.io/library/api-reference/utilities/st.query_params

mscolnick:
This adds the ability pass query_params to python and keep it in sync with the URL state.

```python
@app.cell
def __(mo):
    query_params = mo.query_params()
    return query_params,


@app.cell
def __(mo, query_params, random):
    # In another cell
    print(random.randint(1, 50))
    search = mo.ui.text(
        value=query_params["search"] or "",
        on_change=lambda v: query_params.set("search", v),
    )
    search
    return search,


@app.cell
def __(mo):
    toggle = mo.ui.switch(label="Toggle me")
    toggle
    return toggle,


@app.cell
def __(query_params, toggle):
    # change the value of a query param, and watch the next cell run automatically
    query_params["has_run"] = toggle.value
    return


@app.cell
def __(mo):
    new_value = mo.ui.text(label="Text to add")
    return new_value,


@app.cell
def __(mo, new_value, query_params):
    append_button = mo.ui.button(
        label="Add to query param",
        on_click=lambda _: query_params.append("list", new_value.value),
    )
    replace_button = mo.ui.button(
        label="Replace in query param",
        on_click=lambda _: query_params.set("list", new_value.value),
    )
    mo.hstack([new_value, append_button, replace_button])
    return append_button, replace_button
```

Closes #951 

- [x] Reactivity
- [x] Test wasm support
- [x] Docs
- [x] Better example for saving state lots of state in the URL params for read mode

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -15,6 +15,7 @@
   outputs
   control_flow
   html
+  query_params
   state
   cell
   debugging
@@ -40,6 +41,7 @@ Use the marimo library in marimo notebooks (`import marimo as mo`) to
 | {doc}`outputs`       | Modify cell outputs, redirect console output              |
 | {doc}`control_flow`  | Control how cells execute                                 |
 | {doc}`html`          | Manipulate HTML objects                                   |
+| {doc}`query_params`  | Access and set query parameters with `mo.query_params`    |
 | {doc}`state`         | Synchronize multiple UI elements with `mo.state`          |
 | {doc}`cell`          | Run cells defined in another notebook                     |
 | {doc}`debugging`     | Debugging utilities                                       |
--- a/docs/api/query_params.md
+++ b/docs/api/query_params.md
@@ -1,0 +1,9 @@
+# Query Parameters
+
+Use `mo.query_params` to access query parameters passed to the notebook. You can also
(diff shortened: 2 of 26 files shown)


akshayka (dismissed):
LGTM!

The flickering in the smoke test was actually due to a cycle among query parameter getting and setting, so that's expected and not an issue.

PR #1009, opened 2024-03-27
fix: don't crash kernel when package managers are missing

akshayka:
Fixes a task mentioned in #992

### DIFF
--- a/marimo/_runtime/packages/pypi_package_manager.py
+++ b/marimo/_runtime/packages/pypi_package_manager.py
@@ -56,12 +56,26 @@ class PypiPackageManager(PackageManager):
 class PipPackageManager(PypiPackageManager):
     name = "pip"
 
+    def is_manager_installed(self) -> bool:
+        try:
+            return (
+                subprocess.run(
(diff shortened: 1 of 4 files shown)


marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -86,5 +111,16 @@ async def install(self, package: str) -> bool:
...
+    def is_manager_installed(self) -> bool:
+        try:
+            return (
+                subprocess.run(
+                    ["rye", "--help"], capture_output=True
+                ).returncode
+                == 0
+            )
+        except Exception:
+            return False
```
mscolnick, line 123: seems like `[self.name, "--help"]` is pretty standard - maybe we just included that in the super class?

mscolnick (dismissed):

marimo/_runtime/packages/pypi_package_manager.py
akshayka: Good idea. Also made the implementation more robust.

PR #1020, opened 2024-03-28
fix local file reading for mo.audio

akshayka:
Fixes #623

### DIFF
--- a/marimo/_plugins/core/media.py
+++ b/marimo/_plugins/core/media.py
@@ -39,9 +39,7 @@ def io_to_data_url(
     if src is None:
         return None
 
-    if isinstance(src, io.BytesIO):
-        # clone before reading, so we don't consume the stream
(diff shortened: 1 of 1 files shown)


marimo/_plugins/core/media.py
```diff
@@ -39,9 +39,7 @@ def io_to_data_url(
     if src is None:
         return None
 
-    if isinstance(src, io.BytesIO):
-        # clone before reading, so we don't consume the stream
-        src = io.BytesIO(src.getvalue())
```
mscolnick, line 44: do we still want this? otherwise we can only read it once

akshayka: Added a tell() and seek back to original position (files don't support `getvalue()`).

mscolnick (approved):
Nice fix

PR #1024, opened 2024-03-29
improvement: Cmd+p to open cell actions , restore focus back to cell

mscolnick:
This is a pretty nice QOL improvement. You can run any cell action without leaving your keyboard, and we restore focus back


akshayka (approved):
Awesome!

PR #1021, opened 2024-03-29
feat: enable toggling to md mode automatically via mod-shift-m

riyavsinha:
Features demonstrated in video:
- markdown toggle in cell menu
- markdown toggle doesnt show in cell menu when non-markdown-convertible code
- markdown shortcut works (mod-shift-m)

However, it seems that when a cell is created, the `editorView` is null until _something_ is typed. For this reason, I only know how to get this working after typing some whitespace character first, not sure how to actually initialize the EditorView.

Fixes #1003 

https://github.com/marimo-team/marimo/assets/33243383/4dca535f-7611-4594-9613-136213a2bfd7

PR #1027, opened 2024-03-29
fix: one source of truth for language adapter state

mscolnick:
Should help unblock: https://github.com/marimo-team/marimo/pull/1021

cc @riyavsinha


akshayka (approved):
Thanks!

PR #1034, opened 2024-03-30
avoid server error when encountering broken symlinks

fuenfundachtzig:
When there is a bad symbolic link in a directory, [this line](https://github.com/marimo-team/marimo/blob/413bc9dbb20aa7be86b6cf661606134583ae5e6b/marimo/_server/files/os_file_system.py#L46) will lead to an "Internal Server Error". ("Bad" can either be a recursive link (ln -s broken) or a link pointing to a non-existing file (ln -s broken notthere).)

### DIFF
--- a/marimo/_server/files/os_file_system.py
+++ b/marimo/_server/files/os_file_system.py
@@ -35,7 +35,14 @@ class OSFileSystem(FileSystem):
                 ):
                     continue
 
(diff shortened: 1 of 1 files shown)


fuenfundachtzig:

mscolnick (approved):

mscolnick:
thank you for the fix!

PR #1035, opened 2024-03-30
feat: add pixi package manager

akshayka:
Make it possible to install packages using pixi via the frontend

Also cleanup types on frontend


mscolnick (approved):
So simple adding a new one!

Issue #1004, opened 2024-03-27
\n in mo.md literally interperted in markdown mode


dmadisetti:
### Describe the bug

Clicking on the markdown mode cell button transforms the literal `\n` to a breakline

![image](https://github.com/marimo-team/marimo/assets/2689338/dc74652a-122f-4564-a303-0ea3f4565877)
![image](https://github.com/marimo-team/marimo/assets/2689338/29e4201f-200e-4f5b-b507-83c7876af55d)

changing back

![image](https://github.com/marimo-team/marimo/assets/2689338/1a8edef0-7195-4af9-8dec-c08a3e37e4ea)

### Environment

0.3.4

### Code to reproduce

```python
mo.md(
    r"""
    Some people denote Poisson's ratio as $n$, $\nu$, or even $\mathllap{}\cdot\mathllap{\alpha}$
    """
)
```
Breaks on the \nu

try it here:
https://marimo.app/l/z5epex

fuenfundachtzig:
The problem seems to be caused by [this line](https://github.com/dmnd/dedent/blob/e634f0e59279796e9bb1019836593cde22ab5835/src/dedent.ts#L76) at the end of the dedent function itself.

![image](https://github.com/marimo-team/marimo/assets/8006302/48a5b761-f3f7-4dc2-abfc-1eb221d546a5)

#1038 proposes to switch to dedent-js which doesn't have this problem.

PR #1038, opened 2024-03-31
Address issue 1004 ("\n in mo.md literally interperted in markdown mode")

fuenfundachtzig:
This only replaces the `dedent` library by `dedent-js`.

For details, see https://github.com/marimo-team/marimo/issues/1004.

### DIFF
--- a/frontend/src/core/codemirror/language/__tests__/markdown.test.ts
+++ b/frontend/src/core/codemirror/language/__tests__/markdown.test.ts
@@ -105,7 +105,7 @@ describe("MarkdownLanguageAdapter", () => {
       expect(innerCode).toBe(
         `# Hello, Markdown!\nmo.md(\n    '''\n    # Hello, Markdown!\n    Use marimo's "md" function to embed rich text into your marimo\n    '''\n)`,
       );
-      expect(offset).toBe(18);
+      expect(offset).toBe(9);
     });
 
     it("simple markdown", () => {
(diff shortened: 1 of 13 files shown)


dmadisetti:
dedent-js hasn't been maintained in 10yrs, I think dendent is the correct dependency

Maybe just open an issue in dedent? Alternatively, it's a trivial dep and could be moved to just utils with the patch

akshayka:
> dedent-js hasn't been maintained in 10yrs, I think dendent is the correct dependency
> 
> 
> 
> Maybe just open an issue in dedent? Alternatively, it's a trivial dep and could be moved to just utils with the patch

I wasn't aware it wasn't maintained.

I'd be fine with either of those solutions!

dmadisetti:
I think this package works: https://github.com/jridgewell/string-dedent

It's listed as the example in the proposal to add String.dedent to JavaScript, so would probably get polyfill support too if the proposal goes through

akshayka:
@fuenfundachtzig would you like to update this PR to use [string-dedent](https://github.com/jridgewell/string-dedent) instead?

fuenfundachtzig:
Sure, can do!

fuenfundachtzig:
Now using [string-dedent](https://github.com/jridgewell/string-dedent).

frontend/src/core/codemirror/language/__tests__/markdown.test.ts
```diff
@@ -119,22 +119,22 @@ describe("MarkdownLanguageAdapter", () => {
       const pythonCode = 'mo.md("""   \n# Title\nContent\n   """)';
       const [innerCode, offset] = adapter.transformIn(pythonCode);
       expect(innerCode).toBe("# Title\nContent");
-      expect(offset).toBe(13);
+      expect(offset).toBe(9);
     });
 
-    it("should handle space around the f=strings", () => {
+    it("should handle space around the f-strings", () => {
```
dmadisetti, line 125: it("doesn't do this")

Looks like a quick copypaste from a different test, will update in my pr

dmadisetti (commented):
LGTM, I added some tests in https://github.com/fuenfundachtzig/marimo/pull/1
and they pass for me.

akshayka (approved):
Woohoo! Thank you @fuenfundachtzig and @dmadisetti.

Issue #1030, opened 2024-03-29
Go to executing cell when clicking the sand clock


Eloitor:
### Description

Go quickly to the cell that is currently executing

### Suggested solution

The spinning sand clock could be a great place to click to go to the cell that is currently being executed.

### Alternative

Have some  keyboard shortcut for the same functionality.

PR #1040, opened 2024-03-31
improvement: click hourglass to move view to current cell

fuenfundachtzig:
Addresses #1030

### DIFF
--- a/frontend/src/core/cells/cells.ts
+++ b/frontend/src/core/cells/cells.ts
@@ -733,6 +733,20 @@ export function notebookIsRunning(state: NotebookState) {
   );
 }
 
+export function notebookScrollToRunning(state: NotebookState) {
+  // find cell that is currently in "running" state
+  const { cellIds } = store.get(notebookAtom);
+  const cellIdx = Object.values(state.cellRuntime).findIndex(
+    (cell) => cell.status === "running",
+  );
+  // scroll to editor view of this cell
+  const view = getCellEditorView(cellIds[cellIdx]);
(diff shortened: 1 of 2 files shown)


frontend/src/core/cells/cells.ts
```diff
@@ -733,6 +733,20 @@ export function notebookIsRunning(state: NotebookState) {
   );
 }
 
+export function notebookScrollToRunning(state: NotebookState) {
+  // find cell that is currently in "running" state
+  const { cellIds } = store.get(notebookAtom);
+  const cellIdx = Object.values(state.cellRuntime).findIndex(
+    (cell) => cell.status === "running",
+  );
```
mscolnick, line 741: i think you can avoid wiring the state through, and might be able to change this to:

```tsx
  const { cellRuntime } = store.get(notebookAtom);
  const cell = Objects.entries(cellRuntime).find(
    ([cellId, runtime]) => runtime.status === "running",
  )
  if (!cell) {
     return;
  }
  const view = getCellEditorView(cell[0]);
  view?.dispatch({
    selection: { anchor: 0 },
    effects: [EditorView.scrollIntoView(0, { y: "center" })],
  });
```

fuenfundachtzig: Looks good, thanks for the suggestion :) I'll give it a try.

fuenfundachtzig: I updated the PR to your suggestion. Thanks for the improvement :)

PR #1067, opened 2024-04-04
fix: sort routes in create_asgi_app

mscolnick:
Before if you had a root app and nested app, order mattered. This fixes it an lazily mounts the applications after ordering them.

### DIFF
--- a/marimo/_server/asgi.py
+++ b/marimo/_server/asgi.py
@@ -101,7 +101,7 @@ def create_asgi_app(
     # We call the entrypoint `root` instead of `filename` incase we want to
     # support directories or code in the future
     class Builder(ASGIAppBuilder):
-        def __init__(self):
+        def __init__(self) -> None:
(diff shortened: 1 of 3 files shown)


marimo/_smoke_tests/bugs/1064.py
```diff
@@ -0,0 +1,54 @@
+import marimo
```
mscolnick, line 1: unrelated but wanted to commit to the repo

marimo/_server/asgi.py
```diff
@@ -101,7 +101,14 @@ async def root():
     # We call the entrypoint `root` instead of `filename` incase we want to
     # support directories or code in the future
     class Builder(ASGIAppBuilder):
+        def __init__(self):
```
akshayka, line 104:
```suggestion
        def __init__(self) -> None:
```

akshayka (approved):
LGTM, with 1 suggestion to fix type check

PR #1069, opened 2024-04-04
feat: snippets panel

mscolnick:
Screenshot 2024-04-08 at 10 57 27 AM

Add a new Helper panel for snippets. Snippets are generated from `marimo/_snippets/data`, each snippet as their own file.

Also adds a form to contribute a snippet which we can add ourselves (or someone can create a PR).


akshayka (approved):
Awesome! I think people will find this very helpful. I'm looking forward to adding snippets in the future too :)

PR #1082, opened 2024-04-06
refactor parse_dataclass.py

fuenfundachtzig:
This PR proposes a small refactoring to avoid the need for an unconditional `except` clause to catch all exceptions. Instead, additional checks are introduced and only `TypeError` is ignored. (If there are additional expected exceptions that can safely be ignored, these would need to be added.)

### DIFF
--- a/marimo/_utils/parse_dataclass.py
+++ b/marimo/_utils/parse_dataclass.py
@@ -61,8 +61,11 @@ def _build_value(value: Any, cls: Type[T]) -> T:
         for arg_type in arg_types:
             try:
                 return _build_value(value, arg_type)  # type: ignore # noqa: E501
-            except Exception:
+            # catch expected exceptions when conversion fails
+            except (TypeError, ValueError):
                 continue
+            except:
+                raise
         raise ValueError(f"Value '{value}' does not fit any type of the union")
     elif origin_cls is Literal:
         # if its a single Literal of an enum, we can just return the enum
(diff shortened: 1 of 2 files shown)


fuenfundachtzig:
As the end-to-end tests do not work for me locally, this PR may need further fixes.

fuenfundachtzig:
This now took quite a bit longer than I expected but I hope it is useful. It should (hopefully) not change the behavior but make this part of the code easier to understand for newcomers (like me).

marimo/_utils/parse_dataclass.py
```diff
@@ -87,12 +90,24 @@ def _build_value(value: Any, cls: Type[T]) -> T:
...
+
+    if not isinstance(values, dict):
+        raise ValueError(
+            "value passed to build_dataclass needs to be a dictionary"
+        )
+
     types = get_type_hints(cls)
 
+    snake_cased_values = {to_snake(k): v for k, v in values.items()}
+    if not snake_cased_values.keys() <= types.keys():
```
mscolnick, line 104: this is probably a good thing for us to have. any chance you can add a test case for a subset as well as a superset (and the superset should pass)

fuenfundachtzig: Implicitly, it is already tested in `test_parse_dataclass.py`:
```
        nested = Nested(config=ConfigTwo(gpu=True))
        parsed = parse_raw(serialize(nested), Nested)
```
will fail here when `cls` is `ConfigOne` (first part of `Union[ConfigOne, ConfigTwo]` that it is tested against) because `ConfigOne` expects `types.keys() = dict_keys(['disabled'])` but `nested.config`, i.e. `ConfigTwo`, only provides `gpu` as key in `snake_cased_values`. 

And in `test_build_optional()` it passes (with `types.keys()` being `dict_keys(['x'])`, `snake_cased_values.keys()` being empty).

(We could add more explicit tests, of course.)

mscolnick: ah great, ok thanks - yea we can add explicit tests but maybe in a followup

mscolnick (approved):

PR #1084, opened 2024-04-06
performance: pyodide code completion

akshayka:
- discard stale code completion requests
- don't send code completion requests while the pyodide kernel is running


mscolnick (approved):
nice fixes

Issue #1080, opened 2024-04-05
Column-major tables


akshayka:
`mo.ui.table` expects row-oriented data, which works well in many but not all contexts. In ML and other data applications, data is often stored in columnar formats (parquet, arrow, ...).

We should support columnar data in `mo.ui.table`. A simple way to get started would be to support `pyarrow.Table`. Selection on the frontend would be on columns instead of rows. 

- [x] Support pyarrow tables
- [ ] Add an option to select by column instead of row

mscolnick:
Going to close this as we support pyarrow now. We can re-open if column selection is requested

PR #1091, opened 2024-04-08
feat: support pyarrow and column major data in mo.ui.table

mscolnick:
Part of https://github.com/marimo-team/marimo/issues/1080

Support pyarrow and column-major data. So data can be:
```json
 {
            "A": [1, 2, 3],
            "B": ["a", "b", "c"],
}
```

This works in with selection too - so the selected value is in the same shape. (which also fixes a bug for a primitive list of values and selection)

This refactors a lot of the code into table managers each implementing the `TableManager` interface - this change cleans up the code, improves the testing:

```python
class TableManager(abc.ABC, Generic[T]):
    def __init__(self, data: T) -> None:
        self.data = data
    def to_data(self) -> JSONType:
    def to_csv(self) -> bytes:
    def to_json(self) -> bytes:
    def select_rows(self, indices: list[int]) -> TableManager[T]:
    def get_row_headers(self) -> list[tuple[str, list[str | int | float]]]:
    def is_type(value: Any) -> bool:
```


marimo/_plugins/ui/_impl/table.py
```diff
@@ -129,6 +135,7 @@ def __init__(
             ListOrTuple[dict[str, JSONType]],
             "pd.DataFrame",
             "pl.DataFrame",
+            "pa.Table",
```
akshayka, line 138: Also supports `dict[str, ListOrTuple[JSONType]]` ?

akshayka (approved):

marimo/_plugins/ui/_impl/table.py
mscolnick: oh yep, updating

akshayka (approved):

Issue #292, opened 2023-11-01
Reloading symbols imported with `from ... import ....`


akshayka:
We don't have a good way for users to reload names imported using `from ... import ...`. We recommend that users reload modules using `importlib.reload()`, but that doesn't work for names.

In the interim, we could recommend

```python
import mymodule
from mymodule import name

import importlib
importlib.reload(mymodule)
``` 

But I find that I need to run `reload(mymodule)` twice before `name` is updated with what's on disk; this is a bug, or at least very unintuitive.

PR #1092, opened 2024-04-09
feat: autoreload

### LINKED ISSUE #998: Hot-reload dependencies on file changes
### Description

I am hoping to use Marimo as my "main" script, and develop more complicated logic in an accompanying Python library. To do this most effectively, I need some mechanism for re-loading imports.

In Jupyter, I would use something like https://ipython.readthedocs.io/en/stable/config/extensions/autoreload.html

### Suggested solution

Enable import blocks to be re-loaded on dependent file change.

### Alternative

_No response_

### Additional context

_No response_

akshayka:
This PR implements autoreloading of Python modules (including hot-reloading of objects). The implementation is based on IPython's autoreload extension, and in particular follows the popular `%autoreload 2` settings.

Autoreloading is not reactive, i.e., marimo does NOT watch files and automatically re-execute affected cells. Instead, when autoreloading is enabled, marimo hot-reloads modules before executing the cell.

In the future, if there's interest, we can support reactive autoreloading.

- [x]  Unit tests

Closes #998 , #292

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -318,17 +318,12 @@ notebook menu in the top right and click "Restart kernel".
 
 ### How do I reload modules?
 
-To reload modules, use
-[`importlib.reload()`](https://docs.python.org/3/library/importlib.html#importlib.reload):
+Enable automatic reloading of modules via the runtime settings in your
+marimo installation's user configuration. (Click the "gear" icon in the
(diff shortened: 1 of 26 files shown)


mscolnick (dismissed):
sweet!

PR #1095, opened 2024-04-09
Change test order in Makefile

fuenfundachtzig:
This PR just changes the order in which tests are executed when running `make test`, moving the python unit tests (pytest) before end-to-end tests (playwright).


akshayka (approved):
Thank you!

Issue #546, opened 2024-01-13
Support code formatting using `ruff format`


pavelzw:
### Description

It would be nice to be able to use `ruff format` instead of `black`. It's designed as a faster drop-in replacement for black, see https://docs.astral.sh/ruff/formatter/

### Suggested solution

Check if ruff is installed and if so, use `ruff format` instead of `black`.

baggiponte:
I thought of that too but I think the problem is that ruff does not offer a Python API while black does. @akshayka what if marimo used black or ruff in the environment and installing black + isort would be optional (like `marimo[format]`)? The only issue I see is that I mostly install ruff as a pre-commit hook or as an external dependency, but we could just configure the path in the configs. It could be an entrypoint to create an API to support plugins for formatting.

PR #1118, opened 2024-04-12
feat: format with ruff if provided, black is not

mscolnick:
We still prefer `black` formatting, but if they have `ruff` we fallback to that before showing the warning. 

Perf for each subprocess seems ok, but because we launch a new subprocess for each, I'd still prefer to use `black`

Fixes https://github.com/marimo-team/marimo/issues/546 , as long as the user removes `black`

### DIFF
--- a/marimo/_utils/formatter.py
+++ b/marimo/_utils/formatter.py
@@ -55,6 +55,9 @@ class RuffFormatter(Formatter):
                     capture_output=True,
                     check=True,
                 )
+                if process.returncode != 0:
+                    raise FormatError("Failed to format code with ruff")
+
                 formatted = process.stdout.decode()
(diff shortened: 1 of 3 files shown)


marimo/_utils/formatter.py
```diff
@@ -18,6 +20,51 @@ def format(self, codes: CellCodes) -> CellCodes:
...
+                        "--line-length",
+                        str(self.line_length),
+                        "-",
+                    ],
+                    input=code.encode(),
+                    capture_output=True,
+                    check=True,
+                )
+                formatted = process.stdout.decode()
+                formatted_codes[key] = formatted.strip()
```
akshayka, line 59: Before doing this, check to make sure the `process.returncode` is `0` (ie success)

PR #1198, opened 2024-04-23
improvement: colorize traceback

mscolnick:
Colorize traceback with pygments

1. im pretty sure this is not the write place to do this. there are other stderrs that are not tracebacks
2. i took a first stab at colors, but don't have the best eye for designing this

Screenshot 2024-04-22 at 11 40 10 PM
Screenshot 2024-04-22 at 11 38 27 PM

### DIFF
--- a/marimo/_messaging/streams.py
+++ b/marimo/_messaging/streams.py
@@ -13,7 +13,6 @@ from marimo import _loggers
 from marimo._ast.cell import CellId_t
 from marimo._messaging.cell_output import CellChannel
 from marimo._messaging.console_output_worker import ConsoleMsg, buffered_writer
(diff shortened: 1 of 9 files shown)


marimo/_messaging/streams.py
```diff
@@ -255,7 +256,7 @@ def write(self, data: str) -> int:
                 ConsoleMsg(
                     stream=CellChannel.STDERR,
                     cell_id=self._stream.cell_id,
-                    data=data,
+                    data=highlight_trace_back(data),
```
mscolnick, line 259: this is not the right place, but im not quite sure where is the right place

akshayka (approved):

PR #1209, opened 2024-04-24
improvement: add cell deletion toast to undo, and add to command pallete and notebook actions

### LINKED ISSUE #1192: cell deletion undo stack
### Description

After deleting a cell I'd like to undo it.

### Suggested solution

Keep an undo stack of operations at the cell level - this is not trivial since it is not clear how it is affected by other operations. At the cell level I guess also move up/down topmost/bottommost need to be in it.

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
Fixes #1192

* add cell deletion toast to undo
* add to command pallete
* add to notebook actions

### DIFF
--- a/frontend/islands/__demo__/output.html
+++ b/frontend/islands/__demo__/output.html
@@ -8,9 +8,9 @@
         <meta name="description" content="a marimo app" />
         <title>🏝️</title>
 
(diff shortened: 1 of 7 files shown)


akshayka (approved):
Love this! I really like undo button in the toast.

PR #1220, opened 2024-04-24
improvement: tab to start completion

akshayka:
requested in notion feedback; hit it myself a few times too.


mscolnick (approved):
nice - smart way of figuring out if should do completion or tab

PR #1224, opened 2024-04-25
Add guard around click import for WASM

dmadisetti:
Alternatively, add click to micropip install, but this is better


akshayka (approved):
Thanks!

Issue #1065, opened 2024-04-04
Add logarithmic scale support for Slider


AsgerJ95:
### Description

Log scale Slider: volume, voice pitch, voice speed, etc.

### Suggested solution

Would be great to see log10 scale support working.

### Alternative

A slider with values in dB, which is represented linearly, then read the selected value x and calculate the power afterwards.

mscolnick:
I think a more long-term solution would just allowing custom `steps` and you can create the log scale in python.

We use [Radix slider](https://www.radix-ui.com/primitives/docs/components/slider) which does not support non-linear steps. When we add this feature, 
1) we can make do some stuff under the hood to make `min=0, max=steps.length`, and convert `value=steps[idx]`
2) we can to switch to [Ranger](https://tanstack.com/ranger/latest/docs/framework/react/api/custom-steps) which does support custom steps.

PR #1225, opened 2024-04-25
fix: #1065 Add custom values to slider

metaboulie:
Screenshot 2024-04-25 at 19 02 47

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -96,6 +96,7 @@
     "lodash-es": "^4.17.21",
     "lucide-react": "^0.338.0",
     "lz-string": "^1.5.0",
+    "mathjs": "^12.4.2",
     "mermaid": "^10.8.0",
     "partysocket": "0.0.25",
     "plotly.js": "^2.29.1",
--- a/frontend/src/utils/numbers.ts
+++ b/frontend/src/utils/numbers.ts
@@ -1,7 +1,29 @@
 /* Copyright 2024 Marimo. All rights reserved. */
+import { format } from "mathjs";
+
 export function prettyNumber(value: number): string {
   return value.toLocaleString(undefined, {
     minimumFractionDigits: 0,
     maximumFractionDigits: 2,
   });
+}
+
+export function prettyScientificNumber(value: number): string {
+  // Handle special cases first
+  if (value === 0) {
+    return "0";
+  } // Avoid displaying -0
+  if (Number.isNaN(value)) {
+    return "NaN";
+  }
+  if (!Number.isFinite(value)) {
+    return value > 0 ? "Infinity" : "-Infinity";
+  }
+  // Only format if the value is not an integer
+  if (value % 1 !== 0) {
+    // Use mathjs for formatting with embedded numbers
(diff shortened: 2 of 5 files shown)


frontend/package.json
```diff
@@ -96,6 +96,7 @@
     "lodash-es": "^4.17.21",
     "lucide-react": "^0.338.0",
     "lz-string": "^1.5.0",
+    "mathjs": "^12.4.2",
```
mscolnick, line 99: i think youll want to run `pnpm install` in `cd frontend`, and commit the changes to the lockfile

frontend/src/utils/numbers.ts
```diff
@@ -1,7 +1,31 @@
...
+  if (Number.isNaN(value)) {
+    return "NaN";
+  }
+  if (!Number.isFinite(value)) {
+    return value > 0 ? "Infinity" : "-Infinity";
+  }
+
+  return Math.trunc(value) === 0
+    ? // No integer part, use scientific notation
+      format(value, { notation: "auto", precision: 2 })
```
mscolnick, line 25: the browser has this built in now. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#scientific

would that work instead and avoid the extra dep? we can lowercase the `e` aswell to match this formatting. 


```ts
new Intl.NumberFormat("en-US", {
  notation: "scientific",
}).format(987654321);
```

tests/_plugins/ui/_impl/test_input.py
```diff
@@ -75,6 +84,18 @@ def test_slider_invalid_bounds() -> None:
     assert "Invalid bounds" in str(e.value)
 
 
+def test_slider_exclusive_args() -> None:
```
mscolnick, line 87: can we add a test for empty `steps` (not sure how we want to handle) and invalid bounds (value is not in steps)

tests/_plugins/ui/_impl/test_input.py
```diff
@@ -114,6 +137,13 @@ def test_range_slider_init() -> None:
         # cast to floats
         assert isinstance(num, float)
 
+    slider = ui.range_slider(steps=[1, 3, 6, 10, 17, 20], value=[3, 17])
```
mscolnick, line 140: can we add a test for empty `steps` (not sure how we want to handle) and invalid bounds (value is not in steps)

frontend/package.json
mscolnick: maybe can remove this with the suggestion below

frontend/src/utils/numbers.ts
metaboulie: I think there is a bug for this, the slider would disappear after some dragging, so I retain to `mathjs`

frontend/package.json
metaboulie: I've updated the lockfile, but don't know whether it's right

mscolnick (approved):
awesome! this is really great

PR #1231, opened 2024-04-26
fix: include console outputs in exported html

akshayka:
This fixes the HTML exporter to include console outputs.

However, even though the HTML now contains non-empty `cellConsoleOutputs` (verified manually on a simple example), the console outputs are still not shown in the browser.

~@mscolnick, would appreciate your help with getting the frontend to render the outputs.~

This is only a partial (backend) fix, the frontend still doesn't render the captured console outputs, since it renders in run mode.

Here's my simple test case:

```python
import marimo

app = marimo.App()

@app.cell
def printer():
    print("hello, world!")
    return
```

### DIFF
--- a/marimo/_server/export/utils.py
+++ b/marimo/_server/export/utils.py
@@ -42,17 +42,6 @@ async def run_app_then_export_as_html(
 
     config = UserConfigManager()
     session = await run_app_until_completion(file_manager, cli_args)
-    # Process console messages
-    #
-    # TODO(akshayka): A timing issue with the console output worker
-    # might still exist; the better thing to do would be to flush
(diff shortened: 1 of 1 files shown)


marimo/_server/export/utils.py
```diff
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+from multiprocessing import Process
```
mscolnick, line 5: just confirming - does this import make wasm fail? i think we import this file into wasm

akshayka: No it doesn't, the multiprocessing module is available in Pyodide, even though some of its features don't work.

https://marimo.app/l/hzanca

PR #1247, opened 2024-04-26
feat: autoinstall packages in micropip

mscolnick:
this autoinstalls packages in micropip which will help reduce the errors when in run mode


akshayka (approved):
nice!

PR #1250, opened 2024-04-26
cleanup:  remove mode "island"

mscolnick:


frontend/src/core/dom/marimo-tag.ts
```diff
@@ -32,7 +32,6 @@ const domBasedMarimoSettings: MarimoSettings = {
   },
 };
 
-// We have limited access to the DOM, so defaults are provided.
 const islandsBasedMarimoSettings: MarimoSettings = {
   getMarimoVersion: () => {
     if (import.meta.env.VITE_MARIMO_VERSION === undefined) {
```
dmadisetti, line 37: Since not accessible outside of ISLAND mode, I think you can remove this

marimo/_islands/island_generator.py
```diff
@@ -242,7 +242,6 @@ def render_head(
 
             <marimo-version data-version="{version_override}" hidden="">
```
dmadisetti, line 243: so this tag could be removed too

PR #1253, opened 2024-04-26
fix UnicodeDecodeError

fuenfundachtzig:
Fixes the error
```
...
  File "[home]/ProjekteAktuell/marimo/marimo/_server/files/os_file_system.py", line 96, in _is_marimo_file
    return "app = marimo.App(" in file.read()
  File "/usr/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 1535: invalid continuation byte
```
that will occur when trying to open the file browser in a directory that contains a file that is not utf8-encoded.

To reproduce the error, run the following cell in a marimo notebook and reload the file browser:
`open("babylon.py", "wb").write(b"\xe4")`

### DIFF
--- a/marimo/_server/files/os_file_system.py
+++ b/marimo/_server/files/os_file_system.py
@@ -92,8 +92,8 @@ class OSFileSystem(FileSystem):
         if not path.endswith(".py"):
             return False
 
-        with open(path, "r") as file:
-            return "app = marimo.App(" in file.read()
+        with open(path, "rb") as file:
+            return b"app = marimo.App(" in file.read()
 
     def open_file(self, path: str, encoding: str | None = None) -> str:
         try:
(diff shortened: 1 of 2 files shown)


mscolnick (dismissed):
thanks for the fix!

mscolnick:
@fuenfundachtzig , could you actually add a small test under `def test_get_details_marimo_file(self)`  in `test_os_file_system.py` for a non-utf encoded file?

akshayka (dismissed):

fuenfundachtzig:
I have added a test. `unittest` doesn't support `@pytest.mark.parametrize` that would be very handy here, and as we're already using `pytest` we could switch to pytest completely. If you want, I can make a follow-up PR to refactor `test_os_file_system.py` to rely on `pytest` only.

mscolnick (approved):

mscolnick:
Thanks for the change! Being able to use parameterized sounds nice, if you'd like to make the change, but no need to.

PR #1257, opened 2024-04-27
removed unittest from test_os_file_system.py

fuenfundachtzig:
Follow-up PR to #1253 to stream-line my added unit test.


mscolnick (approved):
thank you!

PR #1258, opened 2024-04-27
fix: properly handle iframes in marimo export

### LINKED ISSUE #1255: Bokeh plots show up as raw html instead of chart on export
### Describe the bug

When exporting a notebook to html that contains bokeh charts, it displays as raw html as opposed to the char.

### Environment

```{
  "marimo": "0.4.6",
  "OS": "Darwin",
  "OS Version": "23.2.0",
  "Processor": "arm",
  "Python Version": "3.11.9",
  "Binaries": {
    "Browser": "124.0.6367.92",
    "Node": "v21.7.3"
  },
  "Requirements": {
    "click": "8.1.7",
    "importlib-resources": "6.4.0",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pymdown-extensions": "10.8",
    "pygments": "2.17.2",
    "tomlkit": "0.12.4",
    "uvicorn": "0.29.0",
    "starlette": "0.37.2",
    "websocket": "missing",
    "black": "24.4.2"
  }
}
```

### Code to reproduce

```
import marimo

__generated_with = "0.4.6"
app = marimo.App()


@app.cell
def __():
    import marimo as mo
    import polars as pl
    return mo, pl


@app.cell
def __(pl):
    pl.DataFrame(
        {
            "time": range(0, 10),
            "value": range(0, 10)
        }
    ).plot.line(x="time", y="value")
    return


if __name__ == "__main__":
    app.run()
```

`marimo export html test.py -o test.html --no-include-code`

mscolnick:
Fixes #1255

1. We properly add `filename` to the virtual file so it can guess the mime type which is needed when creating an data-uri
2. We use `srcdoc` over `src` for better cors support (so resizing the iframe on load works)

### DIFF
--- a/marimo/_runtime/virtual_file.py
+++ b/marimo/_runtime/virtual_file.py
@@ -128,7 +128,7 @@ class VirtualFileLifecycleItem(CellLifecycleItem):
         filename = random_filename(self.ext)
         if context is None or not context.virtual_files_supported:
             self._virtual_file = VirtualFile(
-                filename="", buffer=self.buffer, as_data_url=True
+                filename=filename, buffer=self.buffer, as_data_url=True
             )
             return
 
--- a/docs/guides/coming_from_jupyter.md
+++ b/docs/guides/coming_from_jupyter.md
@@ -3,6 +3,5 @@
 If you're coming from Jupyter, this guide will help you get started
 with marimo. Fun fact: the guide is itself a marimo notebook!
 
-
-<iframe src="https://marimo.app/l/z0aerp?embed=true" width=800 height=1200>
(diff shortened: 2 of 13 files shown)


marimo/_runtime/virtual_file.py
```diff
@@ -128,7 +128,7 @@ def create(self, context: "RuntimeContext" | None) -> None:
         filename = random_filename(self.ext)
         if context is None or not context.virtual_files_supported:
             self._virtual_file = VirtualFile(
-                filename="", buffer=self.buffer, as_data_url=True
+                filename=filename, buffer=self.buffer, as_data_url=True
```
akshayka, line 131: nice, thank you for this fix

akshayka (approved):

PR #1264, opened 2024-04-29
Make app title configurable

fuenfundachtzig:
This PR adds a field in the configuration that allows the user to change the title (HTML <title>) displayed in the browser.

### DIFF
--- a/frontend/src/components/app-config/app-config-form.tsx
+++ b/frontend/src/components/app-config/app-config-form.tsx
@@ -13,7 +13,9 @@ import {
   AppConfig,
   AppConfigSchema,
   APP_WIDTHS,
+  AppTitleSchema,
 } from "../../core/config/config-schema";
+import { Input } from "../ui/input";
 import { NativeSelect } from "../ui/native-select";
 import { useAppConfig } from "@/core/config/config";
 import { saveAppConfig } from "@/core/network/requests";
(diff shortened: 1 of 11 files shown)


frontend/src/components/app-config/app-config-form.tsx
```diff
@@ -81,6 +83,28 @@ export const AppConfigForm: React.FC = () => {
             </FormItem>
           )}
         />
+        <FormField
+          control={form.control}
+          name="apptitle"
```
mscolnick, line 88: can we change the field to just `title`?

marimo/_ast/app.py
```diff
@@ -55,6 +55,7 @@ class _AppConfig:
     """
 
     width: Literal["normal", "medium", "full"] = "normal"
+    apptitle: Optional[str] = None
```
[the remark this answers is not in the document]
mscolnick: it would be `codegen.generate_filecontents`  and `codegen.get_app`

frontend/src/components/app-config/app-config-form.tsx
fuenfundachtzig: I found just "title" a bit ambiguous? But I am happy to change this if you prefer title. (Any alternative suggestions also welcome, of course.)

marimo/_ast/app.py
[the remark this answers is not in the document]
fuenfundachtzig: I have added some tests. (Doing so I found there is already quite a large number of tests :grin:)

frontend/src/components/app-config/app-config-form.tsx
mscolnick: sorry for the late response - maybe `app_title` then, instead of without a space

mscolnick: after that, i think everything looks good! thanks for the awesome contribution

fuenfundachtzig: > sorry for the late response - maybe `app_title` then, instead of without a space

Sounds good, I will change to that.

mscolnick (approved):
thanks for the awesome contrib @fuenfundachtzig! 🚀

PR #1285, opened 2024-05-01
Fix: Issue#822 ui.dataframe can adjust the number of rows displayed

metaboulie:
Fix Issue#822
Besides:
make some changes to the slider and range_slider, cleaning the codespace.
Lint check and tests are done.


tests/_plugins/ui/_impl/test_input.py
```diff
@@ -115,44 +116,24 @@ def test_slider_exclusive_args() -> None:
...
+    assert slider.steps == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 
     slider = ui.range_slider(1, 10, value=[2, 5])
     assert slider.start == 1
     assert slider.stop == 10
-    assert slider.step is None
+    assert slider.step == 1
     assert slider.value == [2, 5]
-    assert slider.steps is None
+    assert slider.steps == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
mscolnick, line 132: I can see how this could clean up the code, but this also is much less performant. For example, if we want a range from 1-10000000 but step 1, we have to pass about more data than just the start/stop. 

I'd rather keep start/stop if we can.

marimo/_plugins/ui/_impl/input.py
```diff
@@ -562,13 +517,22 @@ def __init__(
                 raise ValueError("A radio group cannot have repeated options.")
             options = {option: option for option in options}
         self.options = options
+        if not inline and chunk is not None:
+            chunk = None
+        if (
+            inline
```
mscolnick, line 523: what happens if it's out of bounds? If nothing happens on the frontend, it's probably ok to still pass it for correctness

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -119,6 +121,10 @@ def __init__(
         self._transform_container = TransformsContainer(df)
         self._error: Optional[str] = None
 
+        if page_size not in range(
+            1, len(df) + 1
+        ):  # check if page_size is valid
+            page_size = min(len(df) + 1, 5)
```
mscolnick, line 127: it might be ok to pass the page size for correctness and no need to clamp it. It's really a max size and the frontend should handle when it's above the df

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -82,13 +82,15 @@ class dataframe(UIElement[Dict[str, Any], "pd.DataFrame"]):
...
     - `df`: the DataFrame or series to transform
+    - `page_size`: the number of rows to show in the table
     """
 
     _name: Final[str] = "marimo-dataframe"
 
     def __init__(
         self,
         df: pd.DataFrame,
+        page_size: Optional[int] = 5,
```
mscolnick, line 93: Placing this here instead of the end is technically a breaking change - we should place this after the on_change

marimo/_plugins/ui/_impl/dataframes/dataframe.py
metaboulie: yeah, this is unnecessary

tests/_plugins/ui/_impl/test_input.py
metaboulie: The render of the slider is really slow

PR #1295, opened 2024-05-01
refactor: lazy-load top-level pages, reduce initial page load

mscolnick:


akshayka (approved):
LGTM but haven't looked at Playwright

PR #1297, opened 2024-05-01
fix: sanitize stdout and stderr

### LINKED ISSUE #1270: unsanitized html fragment causing createElement failure
### Describe the bug

Opening a notebook resulted in "Something went wrong" message with the text:

```
Failed to execute 'createElement' on 'Document': The tag name provided ('core::iter::adapters::genericshunt<i,r') is not a valid name.
```

It seems the left angular bracket there is being used as a tag name, or another character is invalid. The string itself is produced by the panic of a polars call, but I don't believe it is material to the bug itself - marimo should handle any python output.

I'll try to create a minimal reproducer, reporting hoping the error itself might be enough.

marimo 0.4.7

### Environment

{
  "marimo": "0.4.7",
  "OS": "Linux",
  "OS Version": "6.8.7-300.fc40.x86_64",
  "Processor": "",
  "Python Version": "3.12.2",
  "Binaries": {
    "Browser": "--",
    "Node": "v20.12.2"
  },
  "Requirements": {
    "click": "8.1.7",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pygments": "2.17.2",
    "tomlkit": "0.12.4",
    "uvicorn": "0.29.0",
    "starlette": "0.37.2",
    "websocket": "missing",
    "black": "24.4.2"
  }
}

### Code to reproduce

none yet.

mscolnick:
Fixes #1270, again


marimo/_messaging/cell_output.py
```diff
@@ -37,16 +39,30 @@ class CellOutput:
     def asdict(self) -> dict[str, Any]:
         return asdict(self)
 
+    def __post_init__(self) -> None:
+        # For plain text outputs, we sanitize the message
+        # if it's already code highlighted.
+        if (
+            self.mimetype == "text/plain"
+            and isinstance(self.data, str)
+            and not is_code_highlighting(self.data)
```
akshayka, line 48: should this be `and is_code_highlighting`?

akshayka (commented):
doesn't look quite right

image

mscolnick:
Maybe quotes need to be sanitized too?

marimo/_messaging/cell_output.py
akshayka: @mscolnick this feels brittle and adds a linear-time string search to every text output.

Instead, I think we may just want to send the right mimetype when writing a traceback. I can try to make a patch.

PR #1299, opened 2024-05-02
Plumb mimetypes to console messages

akshayka:
console already supported mimetypes, this just plumbs them down


tests/_runtime/test_cell_runner.py
```diff
@@ -78,5 +77,4 @@ async def test_base_exception_caught(
     )
     with capture_stderr() as buffer:
         await runner.run(er.cell_id)
-    # pygments inserts HTML in traceback
-    assert 'line <span class="m">3</span>' in buffer.getvalue()
```
akshayka, line 82: This was a bug actually, fixed in this PR

PR #1301, opened 2024-05-02
fix: disconnect from consumer if there is a stale connection

mscolnick:
Sometimes a proxy can hang onto to a connection, but the client is already disconnected (and trying to reconnect). This disconnects if there is still a connection

### DIFF
--- a/marimo/_server/api/endpoints/ws.py
+++ b/marimo/_server/api/endpoints/ws.py
@@ -257,6 +257,8 @@ class WebsocketHandler(SessionConsumer):
             existing_session = mgr.get_session(session_id)
             if existing_session is not None:
                 LOGGER.debug("Reconnecting session %s", session_id)
+                # In case their is a lingering connection, close it
+                existing_session.disconnect_consumer()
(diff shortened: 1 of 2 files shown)


marimo/_server/api/endpoints/ws.py
```diff
@@ -257,6 +257,8 @@ async def get_session() -> Session:
             existing_session = mgr.get_session(session_id)
             if existing_session is not None:
                 LOGGER.debug("Reconnecting session %s", session_id)
+                # In case their is a lingering connection, close it
+                existing_session.maybe_disconnect_consumer()
```
akshayka, line 261: Just checking, does this interfere with waking from sleep?

akshayka (approved):
LGTM, just 1 question

marimo/_server/api/endpoints/ws.py
mscolnick: this actually fixes a case where the client did go to sleep but the backend did not disconnect (if through a proxy)

PR #1303, opened 2024-05-02
feat: show usage stats (memory, cpu) in the footer

mscolnick:


akshayka (commented):
I like this!

`psutil` isn't part of the Python standard library. I don't mind adding it as a dependency -- it has no dependencies itself and is widely used.

mscolnick:
Yea i didnt realize `psutil` wasn't available. i didnt install it through, so i wonder if it was brought in transitively (in which it will be brought in anyways). if we can add it, i think its worth it

akshayka (approved):

akshayka:
@mscolnick idk if storybook and playwright are just flaky?

mscolnick:
@akshayka I'll look into the flakes. This is pretty low priority so can merge later

PR #1304, opened 2024-05-02
Improvement: annotate cycles with variable names

### LINKED ISSUE #847: Naming in cycle resolution
### Description

I am trying to migrate a notebook with a cycle. It is hard to debug especially when the cycle is long.

### Suggested solution

Could the edges of the dependency graph be annotated with the variable names? For example 
```
This cell is in a cycle:
cell-5 -> cell-0
cell-0 -> cell-5
```

Change to 
`cell-5 (var a...) -> cell-0`?
Or it could be shown graphically in the dependency explorer, on the edges, maybe even just on hover?

### Alternative

_No response_

### Additional context

_No response_

fuenfundachtzig:
Fixes #847

### DIFF
--- a/frontend/src/components/editor/output/MarimoErrorOutput.tsx
+++ b/frontend/src/components/editor/output/MarimoErrorOutput.tsx
@@ -63,11 +63,12 @@ export const MarimoErrorOutput = ({
           <Fragment key={idx}>
             <p className="mt-4">{"This cell is in a cycle:"}</p>
             <ul className="list-disc">
-              {error.edges.map((edge) => (
+              {error.edges_with_vars.map((edge) => (
                 <li className={liStyle} key={`${edge[0]}-${edge[1]}`}>
                   <CellLinkError cellId={edge[0]} />
+                  {" -> "} {edge[1].length == 1 ? edge[1] : edge[1].join(", ")}{" "}
                   {" -> "}
-                  <CellLinkError cellId={edge[1]} />
+                  <CellLinkError cellId={edge[2]} />
                 </li>
               ))}
             </ul>
--- a/frontend/src/components/editor/output/MarimoErrorOutput.tsx
+++ b/frontend/src/components/editor/output/MarimoErrorOutput.tsx
@@ -66,8 +66,11 @@ export const MarimoErrorOutput = ({
               {error.edges_with_vars.map((edge) => (
                 <li className={liStyle} key={`${edge[0]}-${edge[1]}`}>
                   <CellLinkError cellId={edge[0]} />
-                  {" -> "} {edge[1].length == 1 ? edge[1] : edge[1].join(", ")}{" "}
-                  {" -> "}
+                  <span className="text-muted-foreground">
+                    {" -> "}
+                    {edge[1].length == 1 ? edge[1] : edge[1].join(", ")}
+                    {" -> "}
+                  </span>
                   <CellLinkError cellId={edge[2]} />
                 </li>
               ))}
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 import threading
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Optional, Set, Tuple
 
 from marimo import _loggers
 from marimo._ast.cell import (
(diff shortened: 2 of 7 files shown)


akshayka:
@fuenfundachtzig, thank you for doing this; definitely an improvement. I'll plan to review once it's moved out of draft, but let me know if you want eyes earlier.

fuenfundachtzig:
> @fuenfundachtzig, thank you for doing this; definitely an improvement. I'll plan to review once it's moved out of draft, but let me know if you want eyes earlier.

Thanks a lot :) There is a failing test, let me look into that first (might have time over the weekend).

fuenfundachtzig:
The test fails because I am using a set in https://github.com/marimo-team/marimo/blob/976b324753f0ed43b460fde701ac567e59bf4f6f/marimo/_runtime/dataflow.py#L22

which becomes part of the cell output if there is a `CycleError` in https://github.com/marimo-team/marimo/blob/976b324753f0ed43b460fde701ac567e59bf4f6f/marimo/_messaging/errors.py#L13

making the `CellOutput` no longer JSON-serializable in https://github.com/marimo-team/marimo/blob/5886681eb6df516e7df4a235536deeaf1262ed80/marimo/_server/templates/templates.py#L106.

I could either
* simply change the definition of `EdgeWithVar` to use a list instead of a set, or
* add a custom handler to `json.dumps` that converts set to lists

Any preferences?

frontend/src/components/editor/output/MarimoErrorOutput.tsx
```diff
@@ -63,11 +63,12 @@ export const MarimoErrorOutput = ({
           <Fragment key={idx}>
             <p className="mt-4">{"This cell is in a cycle:"}</p>
             <ul className="list-disc">
-              {error.edges.map((edge) => (
+              {error.edges_with_vars.map((edge) => (
                 <li className={liStyle} key={`${edge[0]}-${edge[1]}`}>
                   <CellLinkError cellId={edge[0]} />
+                  {" -> "} {edge[1].length == 1 ? edge[1] : edge[1].join(", ")}{" "}
```
akshayka, line 69:
```suggestion
                   <span className="text-muted-foreground">                       
                      {" - "} {edge[1].length == 1 ? edge[1] : edge[1].join(", ")} 
                      {" -> "}                                                     
                   </span>  
```

marimo/_runtime/dataflow.py
```diff
@@ -19,6 +19,9 @@
     from collections.abc import Collection
 
 Edge = Tuple[CellId_t, CellId_t]
+# EdgeWithVar uses a list rather than a set for the variables linking the cells
+#   as sets are not JSON-serializable (required by static_notebook_template())
```
akshayka, line 23:
```suggestion
# as sets are not JSON-serializable (required by static_notebook_template()).
# The first entry is the source node; the second entry is a list of defs from
# the source read by the destination; and the third entry is the destination
# node.
```

marimo/_runtime/validate_graph.py
```diff
@@ -57,8 +57,19 @@ def check_for_cycles(graph: DirectedGraph) -> dict[CellId_t, list[CycleError]]:
     errors = defaultdict(list)
     for cycle in graph.cycles:
         nodes_in_cycle: set[CellId_t] = set(sum(cycle, ()))
+        # before reporting the cells in the cycle to the user,
+        #   we first annotate the cycle with the variable names
+        #   that link its cells
```
akshayka, line 62: Super small nit on comment formatting.

```suggestion
        # before reporting the cells in the cycle to the user,
        # we first annotate the cycle with the variable names
        # that link its cells
```

tests/_runtime/test_runtime.py
```diff
@@ -32,11 +32,11 @@
...
 
 
-def _check_edges(error: Error, expected_edges: Sequence[Edge]) -> None:
+def _check_edges(error: Error, expected_edges: Sequence[EdgeWithVar]) -> None:
     assert isinstance(error, CycleError)
-    assert len(error.edges) == len(expected_edges)
+    assert len(error.edges_with_vars) == len(expected_edges)
     for edge in expected_edges:
-        assert edge in error.edges or (edge[1], edge[0]) in error.edges
+        assert edge in error.edges_with_vars or reversed(edge) in error.edges_with_vars
```
akshayka, line 39: I believe the reversed logic wasn't needed originally; removed it and these tests
still pass on my machine. Checking for the presence of a reversed edge is no longer
correct, since the variables that flow on the edges change when you flip the direction.

```suggestion
        assert edge in error.edges_with_vars
```

akshayka (commented):
This is great! I left just a few small comments and suggestions.

fuenfundachtzig:
Thanks for the suggestions! I have implemented all of them. You're right about the reversed... I had wondered about that but didn't want to break anything I might be overlooking.

I have also decided to sort the variables when printing them.

akshayka (approved):
Thanks!

PR #1315, opened 2024-05-04
fix: add markdown sanitization for malicious notebooks

dmadisetti:
Was looking over the merge this morning and realized it would be possible to "inject" code blocks.

I don't think this could be used that maliciously since running a notebook is already pretty risky. Maybe injecting a <script> tag that would only appear during conversion.

Either way, it's a bug. If you can think of more unsafe behavior let me know. Might be a cool chance to use one of those LLM fuzzers.

### DIFF
--- a/tests/_cli/snapshots/marimo_for_jupyter_users.md.txt
+++ b/tests/_cli/snapshots/marimo_for_jupyter_users.md.txt
@@ -66,25 +66,17 @@ reactive execution model: interactions automatically trigger execution of
 cells that refer to them.
 
 In contrast, Jupyter's lack of reactivity makes IPyWidgets difficult to use.
+## Shareable as apps
 
-````{.python.marimo}
-mo.md(
-    rf"""
-    ## Shareable as apps
+marimo notebooks can be shared as read-only web apps: just serve it with
 
-    marimo notebooks can be shared as read-only web apps: just serve it with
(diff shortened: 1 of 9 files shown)


tests/_cli/snapshots/marimo_for_jupyter_users.md.txt
```diff
@@ -66,17 +66,25 @@ reactive execution model: interactions automatically trigger execution of
...
+````{.python.marimo}
+mo.md(
+    rf"""
+    ## Shareable as apps
+
+    marimo notebooks can be shared as read-only web apps: just serve it with
+
+    ```marimo run your_notebook.py```
 
-```marimo run your_notebook.py```
```
dmadisetti, line 73: Actually, I think this is fine- fixing and checking

tests/_cli/snapshots/unsafe-app.md.txt
```diff
@@ -0,0 +1,68 @@
...
+mo.md(f"""
+    Not markdown
+    ```{{python}}
+    print("1 + 1 = {1 + 1}")
+    ```
+""")
+````
+
+Nested fence
+````text
```
mscolnick, line 43: i didnt even know 4 ticks was a way to nest 3 ticks

mscolnick (approved):
great test cases

PR #1322, opened 2024-05-06
fix: allow int columns in mo.ui.dataframe

mscolnick:
Changes to allow for integer columns in `mo.ui.dataframe`. The diff is larger than expected since numeric-ness is lost when using js maps, so need to switch to a `List` when serializing and an es `Map<>` on the frontend

* other typing improvments
* brute-force snapshot test
* add test cases for integer columns


frontend/.eslintrc.cjs
```diff
@@ -115,6 +115,18 @@ module.exports = {
     "@typescript-eslint/no-unsafe-call": "off",
     "@typescript-eslint/no-unsafe-member-access": "off",
     "@typescript-eslint/no-unsafe-return": "off",
+    // "@typescript-eslint/strict-boolean-expressions": [
```
mscolnick, line 118: leaving this in here since its good to turn on in development, but there are lots of noise that is not worth fixing in this PR

akshayka (approved):

PR #1332, opened 2024-05-07
feat: marimo edit/run markdown.md

dmadisetti:
Cool, I just had this one waiting in a branch, so rebased and tests pass

Open/ Save/ Run work. Known issues are frontend:
- [x] marimo really doesn't want to rename anything not *.py
- [x] Home doesn't pick these up
- [x] no option to download .md files

I feel like this is still a bit of a hack, and could do with a refactor from someone with a deeper understanding of the layout. Feel free to directly contribute to this branch or just use it for inspiration. I really have 2/3 concerns prior to getting this into shape:

1. I think that `_cli/convert` has gotten to big. Might be worth trying to merge this with codegen. Maybe:

```
     codegen/
           __init__.py
           markdown/
                 __init__.py
                 extensions.py
                 parsers.py
           python/
                 __init__.py
```
and moving some of the `exporter.py` logic over too

2. I don't really get the app life cycle. New cells don't automatically create a `cellimpl`? I followed what codegen was doing, but it felt like I was doing something wrong. I didn't dig too deep- but I figured that it was either a deeper issue, or I was missing some way of getting app state.

3. I may have starting to abuse markdown extension system too much. Might be fine as is, but also might be worth explicitly override parts of markdown to make things more transparent.

I think these are all points that either you @akshayka or @mscolnick might have more insight into / if you think this is worth supporting at the moment.

---

Anyway, I left a `markdown.md` in root for you all to play with, if you are interested. Which is a brief write up of the markdown work I just put in. `marimo edit markdown.md` should work


mscolnick:
this is awesome - ill branch off this and make those frontend changes.

regarding the app lifecycle, it just recently got refactored by @akshayka which may/may not help this. historically a lot of state was (and still is) stored in the frontend, and so the cells data-structures are really only initialized and never really updated. part of it, i think, is because those don't need to be updated, until maybe now.

akshayka:
> and so the cells data-structures are really only initialized and never really updated. part of it, i think, is because those don't need to be updated,

Yea, that's right. The `app` object is just used to read in the cells and send their code/config to the kernel process, which has a graph of `CellImpl`s.

mscolnick:
Regarding: "no option to download .md files"

@dmadisetti here is the feat: https://github.com/marimo-team/marimo/pull/1339

dmadisetti:
Super cool!

mscolnick:
@dmadisetti might have a hopefully small merge conflict - but should make this PR much easier

the frontend is currently behind a feature flag - you can either run in development mode or add to your marimo.toml:

```
[experimental] 
markdown = true
```

akshayka:
> My other concerns were just code health and file project structure- so I'm OK with this if you are.

Yup, I'm OK with it.

> I can move the markdown.md to tutorials and write tests for marimo run/edit markdown.md

That sounds great. By the way, the `markdown.md` tutorial is very well written, nice work.

dmadisetti:
:tada: sorry for the commit spam, CI was doing its job.

> That sounds great. By the way, the `markdown.md` tutorial is very well written, nice work.

Thanks! I just followed the style in the others

Issue #717, opened 2024-02-01
Convert to Jupyter notebook or Python script


King-of-Infinite-Space:
### Description

Currently `marimo convert` converts Jupyter notebook to marimo notebook but can't convert it back.

Also, although marimo notebooks can be executed as scripts, it is still useful to export it to a python script (rather than having the code wrapped in functions). Jupyter supports exporting as python script through `nbconvert`.

These features would make it easier for people to integrate marimo into their existing workflows. Thanks!

### Suggested solution

Support `marimo convert  your_notebook.py > your_notebook.ipynb` command.
Add "Export to Python script" option.

scls19fr:
Even if we could lose some features it will be "fair" that Marimo try to do it's best to convert into Jupyter notebook.

round-trip conversion between Jupyter notebook and Marimo notebook wil lead to some features loss... that's life!

You said that Jupyter doesn't support out of order execution

We can consider that the export should consider one cell (maybe first, maybe latest a "input" for data flow).

Moreover there is not only on kernel for Jupyter.

Most people use ipykernel (see https://docs.jupyter.org/en/latest/projects/kernels.html ) but there is also ipyflow https://github.com/ipyflow/ipyflow which is a reactive kernel for Jupyter.

mscolnick:
hey @King-of-Infinite-Space - we haven't added "export to Jupyter", but we do have "export to a flat Python script"

```bash
# prints to stdout
marimo export script notebook.py > notebook.script.py
# writes to the outfile
marimo export script notebook.py --output notebook.script.py
# writes to the outfile and watch notebook.py for changes
marimo export script notebook.py --watch --output notebook.script.py
```

dmadisetti:
I want to add that with export to markdown you can also do

```bash
# prints to stdout
marimo export md notebook.py > notebook.qmd
# writes to jupyter notebook, maybe with a bit of post processing (replace .python.marimo with just python)
quarto convert notebook.qmd
```

King-of-Infinite-Space:
Thanks for following up. I noticed that the generated python script uses `# ---` to separate the cells. I suggest using `# %%`, because VSCode will detect them as code cells ([doc](https://code.visualstudio.com/docs/python/jupyter-support-py)). This is also the format used by VSCode via `Jupyter: Export to Python Script`, for example
```python
# %%
import numpy as np

# %%
print(np.exp(1))



```
[...]

PR #1350, opened 2024-05-09
fix: # %% in export script

akshayka:
Make exported scripts compatible with VSCode notebook format

For #717

### DIFF
--- a/marimo/_server/export/exporter.py
+++ b/marimo/_server/export/exporter.py
@@ -115,7 +115,7 @@ class Exporter:
         ]
         code = (
             f'\n__generated_with = "{__version__}"\n\n'
-            + "\n\n# ---\n\n".join(codes)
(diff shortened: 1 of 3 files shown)


marimo/_server/export/exporter.py
```diff
@@ -115,7 +115,7 @@ def export_as_script(
         ]
         code = (
             f'\n__generated_with = "{__version__}"\n\n'
-            + "\n\n# ---\n\n".join(codes)
```
King-of-Infinite-Space, line 118: Thanks for implementing this. I'd add `# %%` before the first cell also.

mscolnick: I updated to be right before each cell, and included on the first cell

mscolnick (approved):

King-of-Infinite-Space (approved):
Looks great, thank you!

PR #1356, opened 2024-05-10
feat: mo.routes()

mscolnick:
This adds `mo.routes` to change the layout based on the url hash or pathname

```python
    @app.cell
    def __():
        mo.routes({
            "#/home": mo.md("# Home"),
            "#/about": mo.md("# About"),
            "#/contact": mo.md("# Contact"),
        })
        return

    @app.cell
    def __():
        mo.sidebar(
            [
                mo.md("# marimo"),
                mo.nav_menu(
                    {
                        "#/home": f"{mo.icon('lucide:home')} Home",
                        "#/about": f"{mo.icon('lucide:user')} About",
                        "#/contact": f"{mo.icon('lucide:phone')} Contact",
                        "Links": {
                            "https://twitter.com/marimo_io": "Twitter",
                            "https://github.com/marimo-team/marimo": "GitHub",
                        },
                    },
                    orientation="vertical",
                ),
            ]
        )
        return
```

PR #1359, opened 2024-05-11
improvement: hold on to unnamed UI elements

akshayka:
Changes the semantics of the UIRegistry to hold on to unnamed UI elements.

This makes function call requests work on unnamed UI elements.

The registry now keeps track of all the UI elements registered to a cell, and exposes a method to unregister them. The runtime unregisters all UI elements for a cell when cell state is invalidated.

STATUS: DRAFT.

[ ] Think through second-order effects.
[ ] Might need to change destructor logic for UI elements.
[ ] Update existing tests, rewrite new ones.

### DIFF
--- a/marimo/_plugins/ui/_core/registry.py
+++ b/marimo/_plugins/ui/_core/registry.py
@@ -2,8 +2,7 @@
 from __future__ import annotations
 
 import sys
-import weakref
-from typing import Any, Dict, Iterable, Mapping, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Mapping, TypeVar, Union
 
 if sys.version_info < (3, 10):
     from typing_extensions import TypeAlias
@@ -11,9 +10,11 @@ else:
     from typing import TypeAlias
 
 from marimo._ast.app import _Namespace
-from marimo._ast.cell import CellId_t
 from marimo._plugins.ui._core.ui_element import UIElement
 from marimo._runtime.context import get_context
+
+if TYPE_CHECKING:
+    from marimo._ast.cell import CellId_t
 
 UIElementId = str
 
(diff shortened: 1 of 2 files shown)


akshayka:
@mscolnick if you want to check if the `FunctionCall` in #1356 request error is due to not holding onto unnamed UI elements, you can test with this branch

akshayka:
@mscolnick I remember now why we didn't do this. This interacts badly with cached UI elements (we have a unit test that exercises this).

In particular it isn't obvious how to handle the lifecycle of UI elements:

It seems intuitive that when a cell is re-run, we purge UI elements it created from the registry — otherwise we'll have a memory leak. That's what I did in this PR.

But the user may have held onto a reference to the UI element (with `functools.cache`, in a list, ...). If we purged UI elements from the registry on re-run, UI registry lookups would fail with cryptic errors.

Here's the unit test that fails:

```python
  async def test_cached_element_still_registered(                                  
      k: Kernel, exec_req: ExecReqProvider                                         
  ) -> None:                                                                       
      await k.run(                                                                 
          [                                                                        
              exec_req.get("import functools"),                                    
              exec_req.get("import marimo as mo"),                                 
              exec_req.get(                                                        
                  """                                                              
                  @functools.lru_cache                                             
                  def slider():                                                    
                      return mo.ui.slider(1, 10)                                   
                  """                                                              
              ),                                                                   
              (construct_slider := exec_req.get("s = slider()")),                  
          ]                                                                        
      )                                                                            
      # Make sure that the slider is registered                                    
      s = k.globals["s"]                                                           
      assert get_context().ui_element_registry.get_object(s._id) == s              
                                                                                   
      # Re-run the cell but fetch the same slider, since we're using               
      # functools.cache. Make sure it's still registered!                          
      await k.run([construct_slider])                                              
      assert get_context().ui_element_registry.get_object(s._id) == s 
```

Maybe there's a way to work around this, by tracking references/pointers in a clever way, but if there's a solution it isn't obvious to me.

mscolnick:
@akshayka this worked on my PR

PR #1360, opened 2024-05-11
feat: adding reactive html export

gvarnavi:
Big fan of the latest islands feature!

This PR extends the `_server` and `_cli` export functionality to return an interactive HTML using marimo islands.

CLI usage (note CLI args available in regular html export are currently ignored):
```bash
marimo export html notebook.py --reactive --no-include-code -o notebook-reactive.html
```

Low-level python usage (e.g. what I use to define ObservableHQ data-loaders):
```python
from marimo._server.export import run_app_then_export_as_reactive_html
from marimo._utils.marimo_path import MarimoPath
from bs4 import BeautifulSoup
import asyncio

path = MarimoPath("src/data-files/notebook.py")
html, _ = asyncio.run(
    run_app_then_export_as_reactive_html(
        path, include_code=False
    )
)

soup = BeautifulSoup(html, features="html.parser")
body = soup.body.decode_contents().strip()
print(body)
```

Happy to include CLI tests if this looks good!


gvarnavi:

gvarnavi:
recheck

mscolnick:
I like this - we should probably expose a low-level python api that isn't private (don't want to break APIs) and hopefully the smallest api (just a path?)

mscolnick:
or maybe we don't need the low-level api? would you still use it, if this CLI addition was added?

gvarnavi:
I would probably still call this from the low-level api since I suspect most SSGs (e.g. Observable Framework) won't handle the html as is, and one might as-well manipulate it further with python (e.g. using BeautifulSoup as above).

marimo/_cli/export/commands.py
```diff
@@ -101,6 +102,13 @@ async def start() -> None:
     type=bool,
     help="Include notebook code in the exported HTML file.",
 )
+@click.option(
+    "--reactive/--no-reactive",
+    default=False,
+    show_default=True,
+    type=bool,
+    help="Whether to export react HTML file, using marimo islands.",
```
akshayka, line 110: Is this supposed to say "reactive HTML file"? If so, "Whether to export a reactive HTML file; if enabled, the exported HTML will run Python"

We can omit `using marimo islands` from the description, since that is an implementation detail.

akshayka (commented):
This is cool! We should definitely have a way to export WASM-powered HTML. Just yesterday someone asked me for this.

A few comments.

1. I tried using the CLI export on the `intro.py` tutorial and noticed that some styles were off (notebook width, codemirror editors, and KaTeX didn't render are a few things I noticed). `marimo export` should look good out of the box.

2. If we add a CLI, we could omit the low-level API. When integrating with Observable, you could just run the `marimo` CLI (either in a Python script using `subprocess`, or in a shell script, etc), unless I am missing something.

3. It seems odd that `cli-args` is ignored when `--reactive` is passed. That sort of suggests to me that this should be its own `export` option, eg `marimo export wasm` or `marimo export reactive-html`.

If we are not prepared to do the work required for (1) yet, then we could opt for a small low-level API and revisit the CLI later ...

marimo/_cli/export/commands.py
gvarnavi: Ah yes, good catch! I can push the change.

gvarnavi:
Thanks for the prompt response!

> 1. I tried using the CLI export on the `intro.py` tutorial and noticed that some styles were off (notebook width, codemirror editors, and KaTeX didn't render are a few things I noticed). `marimo export` should look good out of the box.

This uses `MarimoIslandStub.render()` directly - but agreed we should perhaps look at the tweaking some inline css styles or similar. FWIW, when I use this to [embed on Observable](https://gvarnavi.observablehq.cloud/marimo-framework/marimo-islands) KaTeX seems to work?

> 2. If we add a CLI, we could omit the low-level API. When integrating with Observable, you could just run the `marimo` CLI (either in a Python script using `subprocess`, or in a shell script, etc), unless I am missing something.

Sure yeah - to clarify, I'm perfectly happy calling the low-level "private" API as above (which presumably the CLI will call as-well) or, like you said, running the CLI tool and then piping the output to a different process for post-processing. 

> 3. It seems odd that `cli-args` is ignored when `--reactive` is passed. That sort of suggests to me that this should be its own `export` option, eg `marimo export wasm` or `marimo export reactive-html`.

`marimo export reactive-html` was my initial thought, and then I realized I was essentially copying a lot of the html code and switched to this. Happy to go back if you think it's cleaner.

marimo/_server/export/__init__.py
```diff
@@ -67,6 +67,58 @@ async def run_app_then_export_as_html(
...
+    file_manager = file_router.get_file_manager(file_key)
+
+    # Create Marimo islands
+    stubs = []
+    generator = MarimoIslandGenerator()
+    for cell_data in file_manager.app.cell_manager.cell_data():
+        stubs.append(
+            generator.add_code(
+                cell_data.code,
+                display_code=include_code,
```
dmadisetti, line 90: I think you could use `show_code` from the notebook cell

marimo/_server/export/__init__.py
```diff
@@ -67,6 +67,58 @@ async def run_app_then_export_as_html(
...
+
+    rendered_stubs = []
+    for stub in stubs:
+        stub._internal_app = generator._app
+        stub._session_view = session_view
+        rendered_stubs.append(stub.render())
+
+    head = generator.render_head()
+    body = "\n".join(rendered_stubs)
+    html = f"""
```
dmadisetti, line 107: Potential for user-provided templates, but I think this is fine for now

dmadisetti (commented):
Cool! So Observable data loaders use require a `*.html.py*` right? Or since invoking from marimo, are you using a `*.html.sh`?

gvarnavi:
> Cool! So Observable data loaders use require a `*.html.py*` right? Or since invoking from marimo, are you using a `*.html.sh`?

@dmadisetti Indeed, the simplest case would be a data-loader called `marimo-intro-cli.html.sh` with the contents
```bash
#!/bin/bash
marimo export html src/data-files/intro-notebook.py --reactive --no-include-code
```

which one can call as follow in an Observable Framework md file:
````md
```js
const marimo_html = FileAttachment("data/marimo-intro-cli.html").html();
```

<div style="max-width:740px; margin: 0 auto;">
  <div id="intro-notebook"> ${marimo_html.body} </div>
</div>
````

Added the [repo](https://github.com/gvarnavi/marimo-framework) on git, if you wanted to have a look. Deployed site can be found [here](https://gvarnavi.observablehq.cloud/marimo-framework/).

**Comment by dmadisetti** [comment]
If the data loader just goes off the shebang, I think you should be able to do

```python
#!/user/bin/env marimo  export html --reactive --no-include-code

import marimo as mo
# normal app here
# ...
```

gvarnavi:
To clarify, I don't necessarily need this PR for my purposes - I'm perfectly happy using the snippet I added in `export.__init__.py` as my Observable dataloader. This was mostly because I wanted to contribute to the codebase and figured others might also like a reactive HTML export!

Pushed the changes I was working on before your comment @mscolnick - I think it might capture atleast some of @akshayka concerns, if you want to check it out.

Regardless of the cli, a static `MarimoIslandGenerator.from_file()` does sound like a good idea.

marimo/_server/export/__init__.py
```diff
@@ -72,6 +72,7 @@ async def run_app_then_export_as_reactive_html(
     include_code: bool,
 ) -> tuple[str, str]:
     from marimo._islands.island_generator import MarimoIslandGenerator
+    from marimo._server.api.utils import parse_title
```
dmadisetti, line 75: Other export uses: https://github.com/marimo-team/marimo/blob/69830587f343f839d777155a870056d010044ea6/marimo/_server/export/utils.py#L30

but parse_title looks like dupe logic anyway 🤔

dmadisetti:
I think this makes sense; islands are a little bit hidden otherwise. Regarding the limited format, I think this and normal export HTML - could be expanded to take a template, but that probably doesn't need to be added until it's needed.

akshayka:
@gvarnavi , thanks for clarifying, and thanks for contributing back to us! We definitely appreciate it :)

@dmadisetti, ok, good point -- wasm export is hard to find/low-level with just islands.

If I understood the above discussion, it will be sufficient to just add reactive HTML export to the CLI; that will enable both gvarnavi's use case as well as let less advanced users benefit from WASM exports. If this is correct, then I support adding reactive HTML export to the CLI.

@gvarnavi, I took a look and the styling does look better -- thank you. Three comments:

1. **Margins.** Below I've pasted a side-by-side comparison of `marimo run intro.py` (left) and `marimo export html --reactive --no-include-code intro.py` (right). Notice that the right-hand export is missing vertical margins. Is this possible to fix?

image

2. **Reactivity**. When testing locally, the exported HTML is not actually reactive. Is this just a me problem? Here's what I see in the console:

```
Not allowed to load local resource: blob:null/049a4c1b-06df-4771-a0fa-50275133aa6b
```

3. **CLI Args.** Why not support CLI args? If it's too hard to support, we should at least raise an Exception if `cli_args` are non-empty when exporting reactive HTML.

dmadisetti:
Oh yikes. # 3 might be from a work around I did to get a local cross site Web Worker. Never tested from a non-served file. This might work in islands bridge: https://gist.github.com/walfie/a80c4432bcff70fb826d5d28158e9cc4

But will probably need to tinkering. I know Myles has taken a pass at that web worker too: https://github.com/marimo-team/marimo/blob/69830587f343f839d777155a870056d010044ea6/frontend/src/core/islands/bridge.ts#L51

mscolnick:
@dmadisetti @akshayka , im pretty sure you can't run a webworker from `file://` (or at least ours) - needs to be a hosted server

akshayka:
(cc @gvarnavi)

> im pretty sure you can't run a webworker from file:// (or at least ours) - needs to be a hosted server

Oh shoot -- I wasn't aware of that. I guess that's for security?

If that's the case we might as well not include this in the CLI, and just recommend that people use `marimo.io` or islands -- can already imagine all the questions we'd get from our users about why their reactive exports aren't working.

@mscolnick I'll leave the decision up to you on how to proceed.

gvarnavi:
> im pretty sure you can't run a webworker from file:// (or at least ours) - needs to be a hosted server

Ah indeed, hadn't realized either since I was mostly testing directly on Observable - which is hosted.

Hmm, perhaps the easiest solution then would be a static `MarimoIslandGenerator.from_file()`  class method, and some additional documentation/examples pointing users to it, as @mscolnick suggests -- I can add that sometime tomorrow.

dmadisetti:
> If that's the case we might as well not include this in the CLI, and just recommend that people use `marimo.io` or islands -- can already imagine all the questions we'd get from our users about why their reactive exports aren't working.

If the major concern is communication to users, then I think this could be mitigated with a alert dialog with something like:

> [!CAUTION]
marimo requires Web Workers to run reactively, and as such marimo islands will not work with local files. Please host this file.

Because not allowing export doesn't get rid of the problem that islands won't work locally. But addressing enables this export, and documents the edgecase

gvarnavi:
Pardon the delay, just got round to this.  

Added a static `MarimoIslandGenerator.from_file()` method per @mscolnick suggestion. This indeed simplifies the islands html generation a fair bit. e.g. I use the following snippet for Observable.
```python
import asyncio

from marimo import MarimoIslandGenerator

generator = MarimoIslandGenerator.from_file("src/data-files/dislocation-fields.py")
app = asyncio.run(generator.build())
body = generator.render_body()
print(body)
```

Note: the new `render_body()` and `render_html()` functions are nice-to-have but not necessary utility functions.

Also, I've kept the cli export for now (and switched it to using the simpler static method above), until we reach a consensus on whether or not to include it. I like @dmadisetti's warning idea -- perhaps this should be implemented at the [islands js file](https://github.com/marimo-team/marimo/blob/main/frontend/src/core/islands/main.ts) though? As in check if the file is served from a local file (perhaps using [window.location.protocol](https://developer.mozilla.org/en-US/docs/Web/API/Location/protocol) or similar), and issue a warning if so?

marimo/_islands/island_generator.py
```diff
@@ -175,6 +179,42 @@ def __init__(self, app_id: str = "main"):
         self._app_id = app_id
         self._app = InternalApp(App())
         self._stubs: List[MarimoIslandStub] = []
+        self._config = _AppConfig()
+
+    @staticmethod
+    def from_file(
```
mscolnick, line 185: great addition, thanks!

marimo/_islands/island_generator.py
```diff
@@ -313,6 +355,100 @@ def render_head(
             """
         ).strip()
 
+    def render_body(
```
mscolnick, line 358: i like this add too

marimo/_islands/island_generator.py
```diff
@@ -313,6 +355,100 @@ def render_head(
...
+            width = self._config.width
+            if width == "normal":
+                max_width = "740px;"
+            elif width == "medium":
+                max_width = "1110px;"
+            else:
+                max_width = "none;"
+
+        if style is None:
+            style = f"margin: {margin} max-width: {max_width}"
```
mscolnick, line 394: i think you need a `;` between this: `margin: {margin}; max-width: {max_width}`

marimo/_cli/export/commands.py
```diff
@@ -101,6 +102,16 @@ async def start() -> None:
     type=bool,
     help="Include notebook code in the exported HTML file.",
 )
+@click.option(
```
mscolnick, line 105: if you aren't using this yet, maybe we can remove it until i add a warning when opening it.
(you can keep `run_app_then_export_as_reactive_html` still if you want)

mscolnick (commented):
looks good! i think if we can remove exposing the CLI `--reactive/--no-reactive` for now, until i can add the warning. the other additions are great and easy enough to make the CLI 

with;

```python
    import os

    from marimo._islands.island_generator import MarimoIslandGenerator

    generator = MarimoIslandGenerator.from_file(
        path.absolute_name, display_code=include_code
    )
    await generator.build()
    html = generator.render_html()
```

marimo/_cli/export/commands.py
gvarnavi: sounds good, will revert.

mscolnick (approved):

gvarnavi:
Great, thanks for the review @mscolnick!

mscolnick:
force merging - looks like unrelated tests are failing on main

PR #1375, opened 2024-05-14
improvement: seperate worker for saving

mscolnick:
This adds a second web-worker for saving. This does add to the overhead since now we need to install pyodide on 2 workers since the python interpreter is not shared between them


akshayka (approved):
LGTM. How much overhead does this add? Hopefully it doesn't double the start-up time?

mscolnick:
@akshayka about 2x the heap. but doesn't seem to effect start-up time on my computer

PR #1393, opened 2024-05-16
fix: enable rpcs unnamed UI elems

akshayka:
- if the output of a cell contains a UIElement, store a reference to it
- the stored reference is released when a cell's state is invalidated
  (on rerun, deletion, etc)

We check for UI elements by using flatten -- covers bare UI elements
and lists, tuples, dicts. Custom datastructures not supported but
should cover almost all real cases, including the very common
practice of ending a cell with `mo.ui.table()`.

- [x] unit tests


mscolnick (approved):
worth adding a test?

PR #1397, opened 2024-05-16
fix: random ids for ui elements made in functions

akshayka:
RPC triggered creation of a UI element (`altair_chart`), which then incorrectly evicted an existing UI element from the registry.


mscolnick (approved):

akshayka:
@mscolnick if this looks good to you, feel free to merge it into your branch. I noticed a unit test is failing, but it at least works in the smoke test for me.

PR #1398, opened 2024-05-17
improvement: markdown shortcuts

mscolnick:
Adds a few markdown shortcuts

```
Bold (Mod-b)
Italic (Mod-i)
Convert to Link (Mod-k)
Convert to Ordered list (Mod-Shift-7)
Convert to Unordered list (Mod-Shift-8)
Convert to Blockquote (Mod-Shift-9)
Convert to Code (Mod-Shift-0)
```

Also improves pasting with a URL when selected and wrapping with a tick (`)


akshayka (commented):
This feels great!

Found a couple bugs though.

1. I'm unable to copy paste URLs into the markdown editor (nothing happens). Pasting non urls still works.

2. I really like the "to code" shortcut. For fenced blocks (triple ticks), can we add a newline at the beginning and end of the block?

For example, starting from

```
def foo():
  ...
```

Instead of

\`\`\`def foo():
  ...\`\`\`

the shortcut should yield


\`\`\`
def foo():
  ...
\`\`\`

akshayka (approved):

PR #1401, opened 2024-05-17
improvement: start cell editor in markdown if is markdown-compatible

mscolnick:
Open to discussing if we want this. but this will make every cell that is markdown compatible, start with the markdown editor


dmadisetti:
I'm a fan. Any more thoughts on inlined markdown? ( I thought there was an issue, but I can't find it) Like double click to edit.
Or could have 3 "language modes", python, markdown, and rendered, removing the output

mscolnick:
@dmadisetti  Seems doable to combine the markdown editor and the markdown outputs - could be under a setting. I can explore next week

PR #1404, opened 2024-05-17
feat: auto-import marimo for markdown

mscolnick:
This will auto-import `import marimo as mo` after toggling markdown

### DIFF
--- a/frontend/src/components/editor/actions/useCellActionButton.tsx
+++ b/frontend/src/components/editor/actions/useCellActionButton.tsx
@@ -25,12 +25,12 @@ import {
   EyeOffIcon,
   SparklesIcon,
 } from "lucide-react";
-import { ActionButton } from "./types";
+import type { ActionButton } from "./types";
 import { MultiIcon } from "@/components/icons/multi-icon";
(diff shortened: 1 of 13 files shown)


akshayka (commented):
To avoid a flash of `NameError` -> new cell -> re-running the markdown, I suggest  specializing this only to markdown actions in the frontend, and doing the logic in the frontend as well, without running any code.

When the user clicks "Create Markdown" or "View as Markdown" ...

1. Check for `import[ \t]+marimo[ \t]+as[ \t]+mo` across all cells.
2. Check for `mo` in the variables table.
3. If `import marimo` is not found, and `mo` is not in the variables table, then add a new cell with `import marimo as mo` to the top of the notebook. If `autorun on startup` is enabled, run it.
4. Add the markdown cell.

1 and 2 should be inexpensive operations, so I wouldn't worry about overhead.

This hotfix is really only needed for markdown anyway. When typing Python, a `NameError` for `mo` is readily understood and fixed by the user.

akshayka (commented):
Nice, feels pretty good. Except the newly added cell shouldn't be marked as stale when it's autorun:

image

PR #1405, opened 2024-05-17
fix: formatting when no formatter is found

akshayka:
- just print plaintext of str(obj): this is better for some objects, plus pprint can fail when str() doesn't
- without this, attempting to outupt duckdb.read_csv() failed

image


mscolnick (dismissed):

akshayka:
Converting to draft -- `plain_text` not working well, might just need to sanitize/escape HTML in string

PR #1407, opened 2024-05-18
Update tutorials

akshayka:
- mention markdown toggle in markdown tutorial
- mention runtime settings in dataflow/intro

### DIFF
--- a/marimo/_tutorials/intro.py
+++ b/marimo/_tutorials/intro.py
@@ -189,13 +189,13 @@ def __(mo):
         """
         ## 2. UI elements
 
-            Cells can output interactive UI elements. Interacting with a UI
-            element **automatically triggers notebook execution**: when
-            you interact with a UI element, its value is sent back to Python, and
-            every cell that references that element is re-run.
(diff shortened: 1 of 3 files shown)


marimo/_tutorials/intro.py
```diff
@@ -164,15 +186,16 @@ def __(mo):
...
-        every cell that references that element is re-run.
+            Cells can output interactive UI elements. Interacting with a UI
+            element **automatically triggers notebook execution**: when
+            you interact with a UI element, its value is sent back to Python, and
+            every cell that references that element is re-run.
 
-        marimo provides a library of UI elements to choose from under
-        `marimo.ui`.
+            marimo provides a library of UI elements to choose from under
+            `marimo.ui`.
```
akshayka, line 198: @mscolnick Is it possible the view as markdown toggle is adding these indents?

mscolnick: seems likely - could be because the original was on the same line. 

this shouldnt happen if we always start in markdown - and only would if the user moves text to the same line as the opening `"""`

Issue #1126, opened 2024-04-12
Go to cell defining something


Eloitor:
### Description

Quickly go to the cell defining a variable or a function.

### Suggested solution

- An item in the right click menu.
- A keyboard shortcut.

robmck1995:
I didn't factor https://github.com/marimo-team/marimo/pull/1365 in when making these changes.

Here's the plan I used to make the change:
https://glide.agenticlabs.com/sharing?userId=65bae76b7fbfb1c20ac1f3b8&taskId=ghM1mqX

Do we think this needs to be updated based on the new functionality of `goToDefinition`? It looked like from my testing that `goToDefinition` just focused on the first instance of the variable within the cell, not across cells?

PR #1411, opened 2024-05-19
Go to definition functionality added to hotkey and context menu

robmck1995:
Fixes https://github.com/marimo-team/marimo/issues/1126

For this one, I set the hotkey to Mod+Shift+, but let me know if there's a better one.

I'm unsure if this is structured correctly, specifically, the utility functions specified in `cell-editor.tsx`. Let me know if there is a better way to structure it. Basically, I wasn't sure how to ensure we could use the same functionality with the right-click menu and the hotkeys. The solution works, but there is a bit of code repetition.

Created this with the help of [Glide](https://glide.agenticlabs.com)! The change in this document only covers adding the hotkey, I manually added the right-click after as it was a small change:
https://glide.agenticlabs.com/sharing?userId=65bae76b7fbfb1c20ac1f3b8&taskId=ghM1mqX


frontend/src/components/editor/cell/cell-context-menu.tsx
```diff
@@ -81,6 +87,21 @@ export const CellActionsContextMenu = ({ children, ...props }: Props) => {
...
+    {
+      label: "Go to Definition",
+      icon: <SearchIcon size={13} strokeWidth={1.5} />,
+      handle: () => {
+        const { getEditorView } = props;
+        const editorView = getEditorView();
+        if (editorView) {
+          const variableName = getWordUnderCursor(editorView?.state);
+          const focusCellId = getCellIdOfDefinition(variables, variableName);
+          if (focusCellId) {
```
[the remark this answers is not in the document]
robmck1995: I think that `getWordUnderCursor` should always return some string. And `getCellIdofDefinition` will only return a value if the variable is defined in the variables array so I think this should not cause errors, but not 100% sure about this

frontend/src/components/editor/cell/code/cell-editor.tsx
```diff
@@ -394,3 +418,29 @@
...
+export const getCellIdOfDefinition = (
+  variables: Variables,
+  variableName: string,
+) => {
+  const variable = variables[variableName as VariableName];
+  if (!variable || variable.declaredBy.length === 0) {
+    return null;
+  }
+  const focusCellId = variable.declaredBy[0];
+  return focusCellId;
```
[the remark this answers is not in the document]
mscolnick: yea this is ok - you can define something twice, but its an error and going to the first one will be ok

frontend/src/core/hotkeys/hotkeys.ts
```diff
@@ -258,6 +258,11 @@ const DEFAULT_HOT_KEY = {
     group: "Navigation",
     key: "Mod-Shift-s",
   },
+  "cell.goToDefinition": {
+    name: "Go to Definition",
+    group: "Navigation",
+    key: "Mod-Shift-,",
+  },
```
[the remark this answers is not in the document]
mscolnick: this is fine

frontend/src/core/cells/cells.ts
```diff
@@ -231,6 +231,25 @@ const {
       scrollKey: null,
     };
   },
+  focusCellAtDefinition: (
```
mscolnick, line 234: Even though its kinda similar to `focusCell`, I think we can make this a function outside of the reducer in `core/codemirror/go-to-definition`. You can access the store by `store.get(atomName)`.

frontend/src/components/editor/cell/code/cell-editor.tsx
```diff
@@ -410,3 +434,29 @@ const CellCodeMirrorEditor = React.forwardRef(
 CellCodeMirrorEditor.displayName = "CellCodeMirrorEditor";
 
 export const CellEditor = memo(CellEditorInternal);
+
+export const getWordUnderCursor = (state: EditorState) => {
```
mscolnick, line 438: i think we can move all this logic to another file `core/codemirror/go-to-definition.ts`. 
we should be able to just expose one function `goToDefinition(view: EditorView)`

frontend/src/components/editor/cell/cell-context-menu.tsx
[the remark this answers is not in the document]
mscolnick: i can play around with some cases - vscode does always show you `Go To Definition`, but does no-op when you don't have a work under it

frontend/src/components/editor/cell/code/cell-editor.tsx
robmck1995: Perfect

frontend/src/core/codemirror/cells/extensions.ts
```diff
@@ -52,6 +52,7 @@ export function cellMovementBundle(
     toggleHideCode,
     aiCellCompletion,
   } = callbacks;
+  const variables = useVariables()
```
mscolnick, line 55: you can use `store.get(variables)` and can likely push this into `goToDefinition`

robmck1995: Will add, thanks for the tip

mscolnick (approved):
nice contrib!

PR #1416, opened 2024-05-20
feat: add duckdb formatter, update table header styles

mscolnick:
* add duckdb formatter
* update table header styles

image

### DIFF
--- a/marimo/_output/formatters/duckdb_formatters.py
+++ b/marimo/_output/formatters/duckdb_formatters.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from marimo._dependencies.dependencies import DependencyManager
 from marimo._messaging.mimetypes import KnownMimeType
 from marimo._output.formatters.formatter_factory import FormatterFactory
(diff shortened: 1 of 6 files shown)


marimo/_output/formatters/duckdb_formatters.py
```diff
@@ -0,0 +1,33 @@
...
+        ) -> tuple[KnownMimeType, str]:
+            if DependencyManager.has_polars():
+                return table(con.pl(), selection=None)._mime_()
+            if DependencyManager.has_pandas():
+                return table(con.df(), selection=None)._mime_()
+
+            return (
+                "text/html",
+                str(con),
+            )
```
akshayka, line 33: This needs to be plain text (pre-formatted). `mo.plain_text` does the right thing. Otherwise you get an unaligned/unreadable ASCII table.

akshayka (commented):
Comment on fixing formatter when polars/pandas not installed

akshayka (commented):
image

PR #1447, opened 2024-05-22
feat: admonition in markdown

### LINKED ISSUE #1340: Admonitions?
### Description

I noticed you all do this in the docs quite a bit, and I'm a fan of them too.

I think github native supports them:

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

mkdocs has the following notation:

```
!!! note
    Hello
```

which is what the markdown extension triggers on: https://github.com/Python-Markdown/markdown/blob/ec8c305fb14eb081bb874c917d8b91d3c5122334/markdown/extensions/admonition.py#L53

but also has `???` for collapsable.

Looks like the implementation in the docs.marimo.io actually comes from rst support?

### Suggested solution

Something to consider; callouts by themselves are a little loud

`callout(title="", body="", kind="", foladable=False)` could make a admonition

### Alternative

callouts are cool- and they're left as is

### Additional context

Was just converting a notebook and thought about it again. I sort of put the beginnings of md interop in `exporter/utils`, but forgot to strip to out

mscolnick:
## 📝 Summary

Adds admonition formatting with markdown and stylings

Fixes #1340 

## 🔍 Description of Changes

Screenshot 2024-05-22 at 2 25 50 PM

### DIFF
--- a/frontend/src/css/admonition.css
+++ b/frontend/src/css/admonition.css
@@ -1,0 +1,98 @@
+.admonition {
+  display: flex;
+  --admonition-padding: 1em;
+  --admonition-border-radius: 5px;
+  --admonition-title-font-weight: bold;
+  --admonition-title-text-transform: capitalize;
+  --admonition-bg: var(--blue-1);
+  --admonition-border: var(--blue-6);
+  --admonition-heading-color: var(--blue-11);
+
+  padding: var(--admonition-padding);
+  border-left: 4px solid;
+  margin-bottom: 1em;
+  background-color: var(--admonition-bg);
+  border-color: var(--admonition-border);
+
+  &::before {
+    margin-top: 0.35rem;
+    margin-right: 1rem;
+  }
+}
+
+.admonition-title {
+  font-weight: var(--admonition-title-font-weight);
+  text-transform: var(--admonition-title-text-transform);
+  color: var(--admonition-heading-color);
+
+  .paragraph {
+    font-weight: initial;
+    @apply text-foreground;
+  }
+}
+
+/* Icons from https://lucide.dev/icons/ */
+
+.admonition.info,
+.admonition.note {
+  --admonition-bg: var(--blue-1);
+  --admonition-border: var(--blue-6);
+  --admonition-heading-color: var(--blue-11);
+
+  &::before {
+    content: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" opacity="60%" width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="admonition-icon lucide lucide-info"><circle cx="12" cy="12" r="10"/><path d="M12 16v-4"/><path d="M12 8h.01"/></svg>');
+  }
+}
+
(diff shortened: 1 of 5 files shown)


mscolnick:
cc @dmadisetti - thoughts? should i update some markdown export/import logic too from the callouts?

marimo/_output/md.py
```diff
@@ -55,6 +55,8 @@ def _md(text: str, apply_markdown_class: bool = True) -> Html:
             "toc",
             # Footnotes
             "footnotes",
+            # Admonitions
+            "admonition",
```
dmadisetti, line 59: Oh sick, I thought the parser was FE markdown. Adding the extension and styles looks awesome.
No markdown integration needed, since it's just pure markdown as is.

Just out of curiosity, what does adding `pymdownx.details` look like?
https://facelessuser.github.io/pymdown-extensions/extensions/details/
Since pymdownx is already a dep

frontend/src/css/admonition.css
```diff
@@ -0,0 +1,96 @@
...
+
+  @apply tracking-wide;
+
+  .paragraph {
+    font-weight: initial;
+    @apply text-foreground tracking-normal;
+  }
+}
+
+/* Icons from https://lucide.dev/icons/ */
```
dmadisetti, line 32: "under ISC"

dmadisetti (approved):
I think this is awesome and closes #1340

You could add a callout hook as syntatic sugar. Pretty easy to add for markdown conversion since the logic is already in: https://github.com/marimo-team/marimo/blob/b71e1d9aa2584bf4718c988a5b7556f0b32ee2d1/marimo/_server/export/utils.py#L78
"kind" just needs to be "admonition".

But going back would be a bit of a pain. It's probably better to strip the logic in utils.

dmadisetti:
Just an idea, but 

```
def _md(text: str, apply_markdown_class: bool = True, extensions : Optional[list[]] = None) 
```

might be interesting. Could allow people to choose their admonition of choice, like for github:
 - https://github.com/dahlia/markdown-gfm-admonition/tree/main
(I think this is small enough to fold into marimo's own extensions)
 - or that rst flavour you use in docs

Just thinking aloud, esp since github has its own flavor and md notebooks already look OK in github: https://github.com/marimo-team/marimo/blob/main/marimo/_tutorials/markdown_format.md

but additionally, python formatting doesn't work on gh with `{.python.marimo}`.
So that could be addressed completely separately with a `marimo export github-md file.py`, which could provide gh specific admonations and valid gh code blocks, and maybe add a badge like colab

akshayka (approved):
This is great!

mscolnick:
@dmadisetti maybe instead of choosing one, we could support both. and then the user doesnt need to pass them down. 

Also for github rendering python.marimo code blocks, could we change the tag/fence to

```
\```python {marimo}
def foo():
   ...
\```
```

Specifically `python {marimo}` works well:

```python {marimo}
def foo():
   ...
```

PR #1460, opened 2024-05-23
fix: boolean parsing for pandas and ui.table

### LINKED ISSUE #1459: ui.table renders `False` bools as `true`
### Describe the bug

`ui.table()` is incorrectly rendering `False` values as `True`

image

### Environment

0.6.2

### Code to reproduce

```python
import marimo

__generated_with = "0.6.2"
app = marimo.App()


@app.cell
def __():
    import marimo as mo
    return mo,


@app.cell
def __():
    import pandas as pd
    return pd,


@app.cell
def __(pd):
    df = pd.DataFrame({"boolean": [True, False, True, False]})
    df
    return df,


@app.cell
def __(df, mo):
    t = mo.ui.table(df)
    t
    return t,


if __name__ == "__main__":
    app.run()
```

akshayka:
Implement a custom boolean parser on the frontend that accounts for the fact that pandas serialized bools start with uppercase letters.

Fixes #1459

### DIFF
--- a/frontend/src/css/admonition.css
+++ b/frontend/src/css/admonition.css
@@ -25,6 +25,7 @@
 
   .paragraph {
     font-weight: initial;
+
     @apply text-foreground tracking-normal;
   }
 }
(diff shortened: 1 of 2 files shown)


akshayka:
cc @mscolnick

PR #1466, opened 2024-05-23
feat: add output replace_by_index function

### LINKED ISSUE #1417: Allow replacement of cell outputs by index
### Description

I would like to be able to replace a certain index of a cell output. My use case is that I want to be able to have a progress bar at output index 0, and occasionally update a plot at output index 1. If I clear the output to show the new plot, I lose the progress bar.

I would be happy to submit a PR if there's an API for this that you like.

### Suggested solution

I've been using this, which seems to work okay (modified from the functions in [_output.py](https://github.com/marimo-team/marimo/blob/main/marimo/_runtime/output/_output.py)):

```python
def replace_output_idx(value: object, idx: int):
    ctx = get_context()
    if ctx.execution_context is None:
        return

    if len(ctx.execution_context.output) < idx:
        raise Exception
    elif len(ctx.execution_context.output) == idx:
        ctx.execution_context.output.append(formatting.as_html(value))
    else:
        current_fig = ctx.execution_context.output[idx]
        ctx.execution_context.output[idx] = formatting.as_html(value)
    write_internal(cell_id=ctx.execution_context.cell_id, value=vstack(ctx.execution_context.output))
```

### Alternative

Another approach that would be a little bit more work is to make `mo.output` behave more like a list with a `__getitem__` method, and maybe an `insert` method:

```python
mo.output.append(1)

mo.output[0] = 2

mo.output.insert(1, 3)

mo.output.clear()
```

### Additional context

_No response_

seantur:
## 📝 Summary

This PR adds an output function `replace_at_index` which allows a specific index of the output to be replaced.

The intended use case is something like this:

```python
import marimo as mo
import time

for i in mo.status.progress_bar(range(10)):
    time.sleep(0.5)
    mo.output.replace_at_index(mo.md(f'Iteration: {i}'), 1)
```

Closes #1417 

## 🔍 Description of Changes

This adds the `mo.output.replace_at_index` function, which allows a specific index of the output to be replaced. If the index is equal to the length of the output, this is equivalent to an append, so that the use case above works.

## 📋 Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://discord.gg/JE7nhX6mD8), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable). #1417 

## 📜 Reviewers

@mscolnick

### DIFF
--- a/frontend/e2e-tests/output.spec.ts
+++ b/frontend/e2e-tests/output.spec.ts
@@ -25,5 +25,9 @@ test("it can clear and append output", async ({ page }) => {
   // Test that Cleared does not exist
   await expect(page.getByText("Cleared!")).not.toBeVisible();
 
+  // Test that Replaced by index is visible and To be replaced is not
+  await expect(page.getByText("To be replaced.")).not.toBeVisible();
+  await expect(page.getByText("Replaced by index!")).toBeVisible();
+
   await takeScreenshot(page, __filename);
 });
--- a/frontend/e2e-tests/py/output.py
+++ b/frontend/e2e-tests/py/output.py
@@ -55,5 +55,12 @@ def __(loop_append, mo):
     return
 
 
+@app.cell
+def __(mo):
+    mo.output.append("To be replaced.")
+    mo.output.replace_at_index(mo.md("Replaced by index!"), 0)
+    return
+
+
 if __name__ == "__main__":
     app.run()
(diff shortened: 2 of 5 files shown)


mscolnick (approved):
looks great! Thanks for the contribution

PR #1469, opened 2024-05-23
feat: add download as html or png in Run mode

mscolnick:
## 📝 Summary

Add download as html or png in Run mode. 

Note: this is always enabled and not configurable


jevalentini:
Hey, big fan of marimo, was trying to use WASM to embed it in a website I'm developing. I see this new button isn't quite responsive. Are you planning it on making it not visible as a config or put it in the end of the notebook or maybe a bit above.

image

akshayka:
@jevalentini , thanks for reporting -- I've opened a new issue for making it responsive: https://github.com/marimo-team/marimo/issues/1577

mscolnick:
@TobiasEnergyMachines it is not currently possible - if you'd like happy to accept a contribution to remove it based on url-configs (if thats something that would work for you).

otherwise this issue https://github.com/marimo-team/marimo/issues/1577 was going to make it minimal when smaller. maybe instead we can just always make it minimal

Issue #1464, opened 2024-05-23
Scratchpad


PatrickCleary:
### Description

I find Notebooks become cluttered with non-essential code while iterating. This code is mostly from testing and checking variables.

### Suggested solution

I've always felt that a scratchpad in Notebooks would solve the above issue in addition to some others. Three use cases:

1. Check the value of a variable
2. Draft a cell while simultaneously being able to scroll through notebook and look at previous cells.
3. Test code without altering variables

Quick mockup (after making this I realized it may belong on the left side with the other tab options):
![image](https://github.com/marimo-team/marimo/assets/46229349/2c28cfa7-7014-477d-a9c4-e54e586f9814)

mscolnick:
@PatrickCleary thanks for the feature request (and great design). this is something requested before. we can add this. is this something you would be interested in contributing or pairing on? if not, i am sure i can get to it at some point

PatrickCleary:
I would take a crack at it! Would you be able to share some quick context on how you'd recommend approaching?

mscolnick:
@PatrickCleary 

I think we can put this in the sidebar for now as a new helper panel. We can add a new endpoint for run_scratch_code which runs Python with the existing global variables but does not add itself to the dependency graph. The scratch doesn't need to be reactive either. It can return the console or visual outputs via the websocket infrastructure we have -  we can add new ws messages. 

Feel free to implement as much as you want - and then we can fill in the gaps. 

I really like your version history with arrow up/down too - up to you if you want to implement that in the first iteration.

PatrickCleary:
Started on this. Left one question on a draft PR and also if someone wanted to give it a quick glance to make sure I'm on the right track that would be great.

mscolnick:
Closed by https://github.com/marimo-team/marimo/pull/1821

Thanks for your help @PatrickCleary!

PR #1475, opened 2024-05-23
Starting work for Scratchpad 

PatrickCleary:
## 📝 Summary

WIP for #1464 
Screenshot 2024-05-23 at 7 21 39 PM

## 🔍 Description of Changes

I just made a quick panel with an icon. Started looking into the backend. Had some questions.

## 📋 Checklist
**Will fill this out when the PR is closer to ready.**
- [ ] I have read the [contributor guidelines](../CONTRIBUTING.md).

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -11,7 +11,12 @@ import {
   useRef,
 } from "react";
 
-import { saveCellConfig, sendRun, sendStdin } from "@/core/network/requests";
+import {
+  saveCellConfig,
+  sendRun,
+  sendRunScratchpad,
+  sendStdin,
+} from "@/core/network/requests";
 import { autocompletionKeymap } from "@/core/codemirror/cm";
 import { UserConfig } from "../../core/config/config-schema";
 import { CellConfig, CellData, CellRuntimeState } from "../../core/cells/types";
@@ -104,6 +109,7 @@ export interface CellProps
    */
   allowFocus: boolean;
   userConfig: UserConfig;
+  isScratchpad?: boolean;
 }
 
 // TODO(akshayka): a component for displaying/editing the cell's name.
@@ -144,6 +150,7 @@ const CellComponent = (
     userConfig,
     config: cellConfig,
     name,
+    isScratchpad,
   }: CellProps,
   ref: React.ForwardedRef<CellHandle>,
 ) => {
@@ -203,8 +210,9 @@ const CellComponent = (
     }
 
     const code = prepareToRunEffects();
+    const runCellFunction = isScratchpad ? sendRunScratchpad : sendRun;
 
-    await sendRun([cellId], [code]).catch((error) => {
+    await runCellFunction([cellId], [code]).catch((error) => {
       Logger.error("Error running cell", error);
     });
   });
--- a/frontend/src/components/editor/chrome/wrapper/footer.tsx
+++ b/frontend/src/components/editor/chrome/wrapper/footer.tsx
@@ -79,7 +79,13 @@ export const Footer: React.FC = () => {
       >
         {renderIcon("snippets")}
       </FooterItem>
-
+      <FooterItem
+        tooltip="Scratchpad"
+        selected={selectedPanel === "scratchpad"}
+        onClick={() => openApplication("scratchpad")}
+      >
+        {renderIcon("scratchpad")}
+      </FooterItem>
       <FeedbackButton>
         <FooterItem tooltip="Send feedback!" selected={false}>
           <MessageCircleQuestionIcon className="h-5 w-5" />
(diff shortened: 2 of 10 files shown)


marimo/_server/api/endpoints/execution.py
```diff
@@ -108,10 +108,21 @@ async def run_cell(
     app_state.require_current_session().put_control_request(
         body.as_execution_request()
     )
+    print("Why does this not print?")
```
PatrickCleary, line 111: Just wanted some input here. I was trying to figure out how the backend works. I was running the backend locally with `marimo edit --headles --no-token`

I feel confident it was running because if I killed the server the frontend gave a warning. Probably a silly question, but not sure why this didn't print when I ran cells.

dmadisetti: `/run` is hit when a cell executes, not just with the server up.
Does that explain the issue? Not sure otherwise, wasn't able to repro locally

frontend/src/components/scratchpad/scratch-pad.tsx
```diff
@@ -0,0 +1,5 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+import type React from "react";
+export const ScratchPad: React.FC = () => {
+  return <p className="text-red-500">TO DO: Put a cell here</p>;
```
dmadisetti, line 4: Cool! Excited to see where this goes

marimo/_server/api/endpoints/execution.py
akshayka: On server up, `instantiate` is hit.

When you click the run button for an individual cell, `/run` is hit (and your message should print, at least it does for me).

For the backend, we can add a new endpoint — something like `run_scratchpad` — that just takes the Python code to run, since the code won't be registered as a cell/node in the kernel's graph.

PatrickCleary: yes, sorry, was running a cell and still no print. I'll try it again tomorrow...


I have a `run_scratchpad` in this PR in the same file. That is what you mean right?

frontend/src/components/editor/Cell.tsx
```diff
@@ -104,6 +109,7 @@ export interface CellProps
    */
   allowFocus: boolean;
   userConfig: UserConfig;
+  isScratchpad?: boolean;
```
mscolnick, line 112: i think rather than using this `<Cell` for the scratchpad - maybe we can re-use different smaller components like `<OutputArea/>`

i am happy to take on splitting this up though if you prefer

frontend/src/components/editor/chrome/wrapper/footer.tsx
```diff
@@ -79,7 +79,13 @@ export const Footer: React.FC = () => {
...
         {renderIcon("snippets")}
       </FooterItem>
-
+      <FooterItem
+        tooltip="Scratchpad"
+        selected={selectedPanel === "scratchpad"}
+        onClick={() => openApplication("scratchpad")}
+      >
+        {renderIcon("scratchpad")}
+      </FooterItem>
```
mscolnick, line 88: we moved some footer items to a sidebar, so you will likely have a merge conflict here FYI

frontend/src/components/scratchpad/scratch-pad.tsx
```diff
@@ -0,0 +1,91 @@
...
+    unfoldAll,
+    sendToBottom,
+    sendToTop,
+    setStdinResponse,
+  } = useCellActions();
+  const cells = flattenNotebookCells(notebook);
+  const testCell = cells[0];
+  return (
+    <div className="p-2">
+      <Cell
```
mscolnick, line 49: I think ideally we can create a single scratch cell that doesn't exist in the notebook graph but can still be executed. I think we can hard-code an ID (i.e. `__scratch__` so that the backend knows to run this cell, but not include its definitions as part of the dependency graph. 

(we can own some of that backend logic if you prefer)

mscolnick (commented):
looks great so far! excited for this feature. let us know how we can help/pair/plug-in

frontend/src/components/scratchpad/scratch-pad.tsx
PatrickCleary: I was kinda hoping someone would drop in here and tell me what I was doing wrong 😆 

Gonna take a look at OutputArea rn

PatrickCleary: How would you recommend creating this cell? It seems `createNewCell` wouldn't really be appropriate because it has logic like assigining ID and adding to the array which we want to avoid. Should I add a new action for creating scratchpad cells?

mscolnick:
Thanks for the initial work on this @PatrickCleary! This was finished off in https://github.com/marimo-team/marimo/pull/1821

PR #1476, opened 2024-05-24
UX improvement: panel footer icons -> sidebar

akshayka:
This change adds a sidebar with panel icons, moving the panel icons from the footer to the sidebar (except for the errors icon) while keeping the footer.

This has two effects:

1. It makes the panel icons much more discoverable. We received feedback that they were not visible.
2. It frees up space in the footer for more UI elements -- in the future, we can surface options like runtime config in the footer, instead of hiding it in the settings menu.

**Closed**

image

**Open**

image

### DIFF
--- a/frontend/src/components/editor/chrome/wrapper/sidebar.tsx
+++ b/frontend/src/components/editor/chrome/wrapper/sidebar.tsx
@@ -1,0 +1,103 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+import React, { PropsWithChildren } from "react";
+import { MessageCircleQuestionIcon } from "lucide-react";
+import { cn } from "@/utils/cn";
+import { useChromeActions, useChromeState } from "../state";
+import { Tooltip } from "@/components/ui/tooltip";
+import { FeedbackButton } from "../components/feedback-button";
+import { PANEL_ICONS, PanelType } from "../types";
+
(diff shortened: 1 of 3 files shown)


frontend/src/components/editor/chrome/wrapper/sidebar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const { selectedPanel } = useChromeState();
+  const { openApplication } = useChromeActions();
+
+  const renderIcon = (type: PanelType, className?: string) => {
+    const Icon = PANEL_ICONS[type];
+    return <Icon className={cn("h-5 w-5", className)} />;
+  };
+
+  return (
+    <div className="h-full py-4 px-1 flex flex-col items-start text-muted-foreground text-md border-t border-border select-none no-print text-sm z-50 absolute">
```
mscolnick, line 20: Do you like the border top?

mscolnick (dismissed):
Looks good! One question about the border-top, may look better without it?

frontend/src/components/editor/chrome/wrapper/sidebar.tsx
akshayka: I can't actually tell the difference with and without the border. It was a holdover from the footer — will remove.

mscolnick (approved):

PR #1498, opened 2024-05-27
feat: adding initialization indicator to marimo islands

### LINKED ISSUE #1468: Marimo Islands Quirks (Pyodide initialization)
### Describe the bug

Currently, the MarimoIslandGenerator tries to build the app before producing the rendered HTML stubs. This has a number of consequences:
- A marimo island like the one below, which tries to import a package with micropip that is not locally-installed will pre-render with errors until pyodide gets initialized. See deployed example [here](https://gvarnavi.observablehq.cloud/marimo-framework/marimo-pyodide-bug).
```python
# imports
import marimo as mo
import sys

if "pyodide" in sys.modules:
  import micropip
  await micropip.install('cowsay')

import cowsay
cowsay.get_output_string('cow', 'Hello, Marimo!')
```
- If the marimo app during export is successful, the HTML output will include all outputs
  - This can be nice, since you get a static view of the app until pyodide initializes
  - Is slightly inconsisent with the rest of marimo (which does not store outputs, and instead only saves the necessary .py files)

- There is no indication to the user when pyodide has initialized and is ready to use

### Suggested Solution Discussion
The simplest solution to address all these would be to simply not pre-render the HTML and show an initialization progress indicator (similar to marimo.app). This does take away the nice static view we now get until the app initializes, so perhaps instead we can still show this if the app runs through without any errors, and otherwise show the initialization progress indicator.

Happy to make this PR once we've agreed on the best approach!

### Environment

```
{
  "marimo": "0.6.3",
  "OS": "Linux",
  "OS Version": "6.5.0-35-generic",
  "Processor": "x86_64",
  "Python Version": "3.12.3",
  "Binaries": {
    "Browser": "--",
    "Node": "v20.11.1"
  },
  "Requirements": {
    "click": "8.1.7",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pygments": "2.18.0",
    "tomlkit": "0.12.5",
    "uvicorn": "0.29.0",
    "starlette": "0.37.2",
    "websocket": "missing",
    "black": "24.4.2"
  }
}
```

gvarnavi:
## 📝 Summary

This PR adds a `include_initialization_indicator` flag to the `MarimoIslandsGenerator.render_body()` function to indicate to the user when the Pyodide initialization is done.

Fixes #1468.

## 🔍 Description of Changes

Specifically, it prepends the following div in the rendered stubs:

```html
  <div class="marimo" style="--tw-bg-opacity: 0;">
    <div class="flex flex-col items-center justify-center">
      <svg
        xmlns="http://www.w3.org/2000/svg"
        width="24"
        height="24"
        viewBox="0 0 24 24"
        fill="none"
        stroke="currentColor"
        stroke-width="1"
        stroke-linecap="round"
        stroke-linejoin="round"
        class="size-20 animate-spin text-primary"
      >
        <path d="M21 12a9 9 0 1 1-6.219-8.56"></path>
      </svg>
      <div>Initializing...</div>
    </div>
  </div>
```
which uses the `animate-spin` class loaded with the marimo css to display something like the following:
Screenshot from 2024-05-27 13-34-18

Note this can be used with or without the asynchronous app build. Consider the following code:

```python
import asyncio
from marimo import MarimoIslandGenerator

generator = MarimoIslandGenerator.from_file("notebook.py")
#app = asyncio.run(generator.build())
body = generator.render_body(include_initialization_indicator=True)
```
If the app is not built, then only the above indicator is displayed (i.e. no static html prerender), see [here](https://gvarnavi.observablehq.cloud/marimo-framework/marimo-pyodide-bug). Note this also addresses points 1 & 2 in #1468.

If the app is built as usual, then this will appear temporarily on the top of the static html prerender, see [here](https://gvarnavi.observablehq.cloud/marimo-framework/marimo-intro).

## 📋 Checklist

## 📜 Reviewers

@mscolnick @dmadisetti

### DIFF
--- a/marimo/_islands/island_generator.py
+++ b/marimo/_islands/island_generator.py
@@ -353,10 +353,57 @@ class MarimoIslandGenerator:
             """
         ).strip()
 
+    def render_init_island(self) -> str:
+        """
+        Renders a static html MarimoIsland str which displays a spinning
+        initialization loader while Pyodide loads and disappears once
+        the kernel is ready to use.
+        """
+
+        init_cell_id = self._app.cell_manager.create_cell_id()
+        init_input = "<marimo-cell-code hidden> '' </marimo-cell-code>"
+        init_output = """
+        <div class="marimo" style="--tw-bg-opacity: 0;">
+          <div class="flex flex-col items-center justify-center">
+            <svg
+              xmlns="http://www.w3.org/2000/svg"
+              width="24"
+              height="24"
+              viewBox="0 0 24 24"
+              fill="none"
+              stroke="currentColor"
+              stroke-width="1"
+              stroke-linecap="round"
+              stroke-linejoin="round"
+              class="size-20 animate-spin text-primary"
+            >
+              <path d="M21 12a9 9 0 1 1-6.219-8.56"></path>
+            </svg>
+            <div>Initializing...</div>
+          </div>
+        </div>
+        """
+        init_island = dedent(
+            f"""
+            <marimo-island
+                data-app-id="{self._app_id}"
+                data-cell-id="{init_cell_id}"
+                data-reactive="{json.dumps(True)}"
+            >
+                <marimo-cell-output>
+                {init_output}
+                </marimo-cell-output>
+                {init_input}
+            </marimo-island>
+            """
+        ).strip()
+
+        return init_island
+
     def render_body(
         self,
         *,
-        include_initialization_indicator: bool = True,
+        include_init_island: bool = True,
         max_width: Optional[str] = None,
         margin: Optional[str] = None,
         style: Optional[str] = None,
@@ -366,7 +413,7 @@ class MarimoIslandGenerator:
         This should be included in the <body> tag of the page.
 
         *Args:*
-        - include_initialization_indicator (bool): It True, adds init loader.
+        - include_init_island (bool): It True, adds initialization loader.
         - max_width (str): CSS style max_width property.
         - margin (str): CSS style margin property.
         - style (str): CSS style. Overrides max_width and margin.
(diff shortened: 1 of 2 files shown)


marimo/_islands/island_generator.py
```diff
@@ -380,6 +376,46 @@ def render_body(
...
+                    data-reactive="{json.dumps(True)}"
+                >
+                    <marimo-cell-output>
+                    {dummy_output}
+                    </marimo-cell-output>
+                    {dummy_input}
+                </marimo-island>
+                """
+            )
+            rendered_stubs = [dummy_stub] + rendered_stubs
```
dmadisetti, line 417: Not really clear why you are changing the order? Shouldn't you be adding dummy stub with real code input? instead of both?

dmadisetti (commented):
+1 I like that it's a flag, I don't know what the default should be, but this does make sense.

Maybe Optional[bool] and the default could be mimetype informed? UI always loading, image and it can be displayed normally. HTML w/o UI can be displayed normally, an expected error with a loading: https://github.com/marimo-team/marimo/blob/6e8ccd936ba5761cf3179bda45d061513e2f1296/marimo/_islands/island_generator.py#L513

Just need to catch and manage the mimetype. Just an idea

gvarnavi:
So currently this is at the `app` level, similar to how marimo.app shows an initialization indicator when first loaded. So, when enabled, the code adds a single dummy input with an empty string and a dummy output with the icon. After initialization, the empty string input evaluates -- leading the icon to disappear.

Presumably we could have something like this for each cell. In which case, you're correct that there's no dummy input and we can add the dummy output using a mime-aware default. I think that might be too much though? Like apps can have many cells..

dmadisetti:
Ah ok. I got you

I think you can just conditionally do the output here:

https://github.com/marimo-team/marimo/blob/6e8ccd936ba5761cf3179bda45d061513e2f1296/marimo/_islands/island_generator.py#L124C1-L126C34

Opposed to making a new cell. When render kicks in it will replace the loading with the correct value.

---

But also, I was thinking that this doesn't account for writing islands code directly (which is feasible right now). We could move the logic to the front end by having a `data-include-initialization-indicator` (if empty, default true) like here:

https://github.com/marimo-team/marimo/blob/6e8ccd936ba5761cf3179bda45d061513e2f1296/marimo/_islands/island_generator.py#L122

Then, backfill the initially empty output with the loading indicator. I don't remember how the logic is laid out - but you'd have to do it prior to cell registration since, otherwise, you'd end up with a race condition between the backfill and the pyodide actually loading.

gvarnavi:
> Opposed to making a new cell.

The reason for a new cell is to allow for the (presumably fairly common) use-case where one likes the static pre-rendered html output, but also to inform users when Pyodide has finished loading. See e.g. [here](https://gvarnavi.observablehq.cloud/marimo-framework/marimo-intro).

> I was thinking that this doesn't account for writing islands code directly (which is feasible right now).

Not sure how much of a concern this is. Like if you're writing islands by hand - you can always copy the initialization code stub boilerplate too. We can make this a little easier by writing a `self.render_initialization_island()` utility (and then call it inside `self.build_body()` to avoid duplicating code).

> We could move the logic to the front end.

This was my initial idea in #1468, but I think the css solution is quite elegant!

mscolnick (approved):

marimo/_islands/island_generator.py
```diff
@@ -370,7 +413,7 @@ def render_body(
         This should be included in the <body> tag of the page.
 
         *Args:*
-
+        - include_init_island (bool): It True, adds initialization loader.
```
dmadisetti, line 416: *If True

dmadisetti:
Ok, got you. I thought it was replacing with loading for every island. Yeah, I agree I don't think this is needed for hand written islands.

I think this is a nice work around. The UI elements are still tweakable during the load though, the loader could go with a message?

But I'm alright with this as is- there can always be future tweaks. Thanks!

dmadisetti (approved):

PR #1505, opened 2024-05-28
fix: turbo cache busting for production builds

mscolnick:


akshayka (approved):

akshayka (commented):
Looks like the CLI tests can't find the assets directory

PR #1514, opened 2024-05-28
feat: run_button

akshayka:
## 📝 Summary

Adds `mo.ui.run_button()`, which can be used to trigger computations when a button is pressed and ONLY when a button is pressed.

## 🔍 Description of Changes

When the button is pressed, its value is set to `True` and its descendants are run. After its descendants have run, marimo automatically sets the button's value back to False. In this way, cells can gate computations on `run_button.value` -- updates to other variables won't spuriously cause the cell to run.

Caveat: The button's value is only reset to `False` when the runtime is configured to autorun cells, since the lazy runtime doesn't know when the button's descendants will/have been run. This should be fine in practice, since `ui.run_button` has no real use in lazy kernels anyway.

The implementation reuses the frontend plugin for `mo.ui.button()`.

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/docs/api/inputs/run_button.md
+++ b/docs/api/inputs/run_button.md
@@ -17,7 +17,7 @@
 
     @app.cell
     def __():
-        mo.stop(b.value, "Click `run` to submit the slider's value")
+        mo.stop(not b.value, "Click `run` to submit the slider's value")
 
         s.value
         return
--- a/docs/api/inputs/run_button.md
+++ b/docs/api/inputs/run_button.md
@@ -1,0 +1,31 @@
+# Run Button
+
+```{eval-rst}
+.. marimo-embed::
+    @app.cell
+    def __():
+        b = mo.ui.run_button()
+        b
+        return
+
+    @app.cell
+    def __():
+        s = mo.ui.slider(1, 10)
+        s
+        return
+
+
+    @app.cell
+    def __():
+        mo.stop(b.value, "Click `run` to submit the slider's value")
+
+        s.value
+        return
+```
+
+```{eval-rst}
+.. autoclass:: marimo.ui.run_button
+  :members:
+
+  .. autoclasstoc:: marimo._plugins.ui._impl.run_button.run_button
+```
--- a/marimo/_plugins/ui/_core/ui_element.py
+++ b/marimo/_plugins/ui/_core/ui_element.py
@@ -379,6 +379,10 @@ class UIElement(Html, Generic[S, T], metaclass=abc.ABCMeta):
         if self._on_change is not None:
             self._on_change(self._value)
 
+    def _on_update_completion(self) -> None:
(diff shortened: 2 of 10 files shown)


marimo/_plugins/ui/_core/ui_element.py
```diff
@@ -379,6 +379,10 @@ def _update(self, value: S) -> None:
         if self._on_change is not None:
             self._on_change(self._value)
 
+    def _on_update_completion(self) -> None:
+        """Callback to run after the kernel has processed a value update."""
+        return
```
akshayka, line 384: This method acts as a post execution hook; the runtime calls it after setting UI element values.

marimo/_plugins/ui/_impl/run_button.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if isinstance(ctx, KernelRuntimeContext) and ctx.lazy:
+            # Resetting to False in lazy kernels makes the button pointless,
+            # since its value hasn't been read by downstream cells on update
+            # completion.
+            #
+            # The right thing to do would be to somehow set to False after
+            # all cells that were marked stale because of the update were run,
+            # but that's too complicated.
+            return
```
akshayka, line 115: Special case for lazy kernels, since we don't when descendants will run

docs/api/inputs/run_button.md
```diff
@@ -0,0 +1,31 @@
...
+    @app.cell
+    def __():
+        s = mo.ui.slider(1, 10)
+        s
+        return
+
+
+    @app.cell
+    def __():
+        mo.stop(b.value, "Click `run` to submit the slider's value")
```
mscolnick, line 20: should this be `mo.stop(not b.value)`.

not sure if this hides to much, but could put a stop on the button too?

```python
b.stop(fallback=None)

// Run thing
```

or maybe `b.gate()`, `b.clicked()`

marimo/_plugins/ui/_impl/run_button.py
```diff
@@ -0,0 +1,116 @@
...
+class run_button(UIElement[Any, Any]):
+    """
+    A button that can be used to trigger computation.
+
+    **Example.**
+
+    ```python
+    # a button that when clicked will have its value set to True;
+    # any cells referencing that button will automatically run.
+    button = mo.ui.button()
```
mscolnick, line 22: `mo.ui.submit_button()`

mscolnick (dismissed):

docs/api/inputs/run_button.md
akshayka: Oops yes it should.

Something like that could be nice — maybe `b.gate()` ... might add in a follow-up

mscolnick (approved):

PR #1522, opened 2024-05-30
feat: support ibis and the dataframe protocol

mscolnick:
## 📝 Summary

This adds support for Ibis and the [dataframe protocol](https://data-apis.org/dataframe-protocol/latest/) in `mo.ui.table` and `mo.ui.data_explorer` as well as the new datasources panel.

This is using the pyarrow interchange to convert to a pyarrow table, so we can use leverage our existing infrastructure downstream.

It isn't quite necessary to try to convert to pandas or polars from the dataframe protocol since most of the libraries that implement the dataframe protocol (i.e. ibis) already require pyarrow. If we wanted to add this later, we ccan

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/types.py
+++ b/marimo/_plugins/ui/_impl/tables/types.py
@@ -1,0 +1,11 @@
+# Copyright 2024 Marimo. All rights reserved.
+from __future__ import annotations
+
+from typing import Any, Protocol, runtime_checkable
+
+
+@runtime_checkable
+class DataFrameLike(Protocol):
+    def __dataframe__(
+        self, nan_as_null: bool = False, allow_copy: bool = True
+    ) -> Any: ...
--- a/frontend/src/core/datasets/state.ts
+++ b/frontend/src/core/datasets/state.ts
@@ -34,9 +34,12 @@ const {
       seen.add(key);
       return true;
     });
+    const sortedTables = dedupedTables.sort((a, b) => {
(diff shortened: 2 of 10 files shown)


marimo/_plugins/ui/_impl/tables/types.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2024 Marimo. All rights reserved.
+from __future__ import annotations
+
+from typing import Any, Protocol, runtime_checkable
+
+
+@runtime_checkable
```
akshayka, line 7: nice!

akshayka (approved):

szst11:
Hi there,
cool for implementing that new feature! 
Could it be, that the datasources panel automaticly retrieves informations from every datasource? As an Ibis-table can be a connection to a very large databasetable this could take a while and it would be nice to surpress this on demand.

akshayka:
@szst11 : 

> Could it be, that the datasources panel automaticly retrieves informations from every datasource? As an Ibis-table can be a connection to a very large databasetable this could take a while

Thanks for the question. Did you run into this issue while working on a notebook, or is this a hypothetical?

szst11:
> @szst11 :
> 
> > Could it be, that the datasources panel automaticly retrieves informations from every datasource? As an Ibis-table can be a connection to a very large databasetable this could take a while
> 
> Thanks for the question. Did you run into this issue while working on a notebook, or is this a hypothetical?

yes, my old notebook hangs when opening with this version.

PR #1553, opened 2024-06-04
feat: mo.ui.table() column chart previews

mscolnick:
## 📝 Summary

Add columns previews to the top of each column. 

- Shows nulls as orange
- Doesn't chart strings
- Uses bar and binning for ints/floats/dates

Bug fixes:
- don't wrap table cells, this helps prevent it from its height jumping
- add a loading table (to also help with height jumping)
- fail more gracefully on integer column names (altair doesnt allow this)

## 🔍 Description of Changes

Screenshot 2024-06-04 at 10 33 31 AM
Screenshot 2024-06-04 at 10 30 51 AM

Future ideas:
-  Make expandable and configure hiding to start
- Dropdown action to add altair chart code for better preview

### DIFF
--- a/frontend/src/components/charts/chart-skeleton.tsx
+++ b/frontend/src/components/charts/chart-skeleton.tsx
@@ -1,0 +1,49 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+import React from "react";
+
+interface Props {
+  seed: string;
+  width: number;
+  height: number;
+}
+
+// Simple hash function to convert a string to a number
+const hashString = (str: string) => {
+  let hash = 0;
+  for (let i = 0; i < str.length; i++) {
+    const char = str.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash = Math.trunc(hash); // Convert to 32bit integer
+  }
+  return hash;
+};
+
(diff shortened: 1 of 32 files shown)


akshayka:
This appears to regress `mo.ui.dataframe` (doesn't render):

Error in console: 

image

Code repro:

```python
import marimo

__generated_with = "0.6.13"
app = marimo.App()


@app.cell
def __():
    from vega_datasets import data
    return data,


@app.cell
def __(data):
    cars = data.cars()
    return cars,


@app.cell
def __():
    import marimo as mo
    return mo,


@app.cell
def __(cars, mo):
    mo.ui.dataframe(cars)
    return


if __name__ == "__main__":
    app.run()
```

akshayka:
For dataframes truncated by `mo.ui.table`, the chart preview and column summaries in `mo.ui.table` appear to show a preview of the truncated data instead of the original data. The data explorer panel in contrast correctly shows a preview of the original data.

Here, notice that the chart preview stops at "9,999", even though the max value is "30,000".
image

mscolnick:
1. First issue is fixed 
2. The min/max/unique are now operating on the full (non-clipped) dataframe, but we hide the charts when above >20k

akshayka (approved):

PR #1567, opened 2024-06-06
user config: display.width

akshayka:
Adds a display.width configuration to the user config that controls the default app width for new notebooks, with a default value of "medium".

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -11,7 +11,6 @@ from typing import (
     Callable,
     Iterable,
     Iterator,
-    Literal,
     Mapping,
     Optional,
 )
@@ -31,6 +30,7 @@ from marimo._ast.errors import (
     MultipleDefinitionError,
     UnparsableError,
 )
+from marimo._config.config import WidthType
 from marimo._dependencies.dependencies import DependencyManager
 from marimo._messaging.mimetypes import KnownMimeType
 from marimo._messaging.types import NoopStream
@@ -54,8 +54,7 @@ class _AppConfig:
     a single marimo program.
     """
 
-    # TODO(akshayka): rename normal -> compact
-    width: Literal["normal", "medium", "full"] = "normal"
+    width: WidthType = "normal"
     app_title: Optional[str] = None
 
     # The file path of the layout file, relative to the app file.
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -54,7 +54,7 @@ class _AppConfig:
     a single marimo program.
     """
 
-    width: WidthType = "normal"
+    width: WidthType = "compact"
     app_title: Optional[str] = None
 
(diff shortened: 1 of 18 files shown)


marimo/_server/file_manager.py
```diff
@@ -8,28 +8,35 @@
...
 
 LOGGER = _loggers.marimo_logger()
 
 
 class AppFileManager:
-    def __init__(self, filename: Optional[str]) -> None:
+    def __init__(
+        self,
+        filename: Optional[str],
+        user_config: MarimoConfig | None = None,
```
akshayka, line 28: Not sure if this should be a required argument. Pyodide `save_file()` (`bootsrap.py`) for example uses `AppFileManager` but doesn't have access to the user config (and doesn't need it to save the file).

marimo/_config/config.py
```diff
@@ -100,6 +100,8 @@ class DisplayConfig(TypedDict):
     theme: Literal["light", "dark", "system"]
     code_editor_font_size: int
     cell_output: Literal["above", "below"]
+    # TODO(akshayka): rename normal -> compact
+    width: Literal["normal", "medium", "full"]
```
mscolnick, line 104: should this be `default_width`?

marimo/_ast/app.py
```diff
@@ -54,6 +54,7 @@ class _AppConfig:
     a single marimo program.
     """
 
+    # TODO(akshayka): rename normal -> compact
```
mscolnick, line 57: could make the breaking change later, but to stop the bleeding, could add `compact` to the list, but still allow `normal` as a valid value

marimo/_server/file_manager.py
mscolnick: it probably doesn't need to take in the full `user_config` right? maybe just `initial_width`?

marimo/_server/file_manager.py
```diff
@@ -8,28 +8,35 @@
...
         self.filename = filename
+        # If provided, user_config overrides default app config for new apps
+        self._user_config = user_config
         self.app = self._load_app(self.path)
 
     @staticmethod
-    def from_app(app: InternalApp) -> AppFileManager:
-        manager = AppFileManager(None)
+    def from_app(
+        app: InternalApp, user_config: MarimoConfig | None = None
```
mscolnick, line 37: if you change the constructor to just take in `initial_width`, then you dont need to pass `user_config: MarimoConfig | None = None` since it will have the width on it

`from_app` should already have `width`

akshayka: Will change from staticmethod to a regular method

mscolnick: i dont think you need to do that - i was just thinking the app already have the `width` on it, which you can pass into the constructor

mscolnick (dismissed):

PR #1571, opened 2024-06-07
fix: `try` blocks delete context variable afterwards, but marimo still considers them refs or defs if used

dmadisetti:
## 📝 Summary

previously

```python
try:
   1 / 0
except Exception as e:
   f = e
   e = 2
refs(), defs()
```

Would give `ref=e`, `def=e,f`. At first glance, this seems fine- but "e" is scoped to the try block and is deleted afterwards, so the correct solution is `ref=[]`, `def=[f]`

Note in the case

```python
e = 7
try:
   1 / 0
except Exception as e:
   f = e
   e = 2
   print("woops")
```

e is no longer set if the except block triggers, but because it was set outside of the block, `ref=`, `def=e,f` is correct

## 🔍 Description of Changes

Added an `ObscuredScope` context to remove Name references if they are obscured by the try block + added some unit tests

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -57,6 +57,7 @@ class Block:
 @dataclass
 class ObscuredScope:
     """The scope in which a name is hidden."""
+
     # Variable id if this block hides a name
     obscured: Optional[str] = None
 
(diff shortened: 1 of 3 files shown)


mscolnick (approved):
nice find and fix!

PR #1572, opened 2024-06-07
couple more tweaks to config width

mscolnick:


marimo/_ast/app.py
```diff
@@ -54,7 +54,7 @@ class _AppConfig:
     a single marimo program.
     """
 
-    width: WidthType = "normal"
+    width: WidthType = "medium"
```
mscolnick, line 57: i change this default since the user config's default is medium? this wouldnt change much but i figured worth being consistent?

akshayka: The only thing is that it would change everyone's existing apps, for those who relied on default being normal (since we didn't serialize the default config). That's why I left it as is

akshayka (commented):
Just one comment, otherwise lgtm!

Issue #1509, opened 2024-05-28
mo.stop propagation breaks


dmadisetti:
### Describe the bug

I'm not able to replicate this without sharing my notebook.

It's a pretty large notebook. I have a stop that should effectively stop the whole notebook, but I start getting name errors at some point since the stop doesn't propagate as far as it should.

![image](https://github.com/marimo-team/marimo/assets/2689338/d0ad1906-37b6-4ca5-8d46-91c5dcf84604)
Name error for projections

![image](https://github.com/marimo-team/marimo/assets/2689338/3daeed76-98fb-4b02-b74a-6bdc152d28e5)
The cell that defines projections

Rerunning the stop cell then works! It feels like maybe a race condition somewhere

### Environment

0.6.11

### Code to reproduce

Large notebook with many parts, since stop condition that should halt everything. Restarting my notebook kernel reliably triggers the incorrect name errors

akshayka:
I think a fix could be to not run a cell if any of its ancestors are stopped.

We do this with `disabled` cells, but not for stopped cells.

dmadisetti:
I think "stop" itself is fine, but something is up with resolving the dependency graph. In the example I gave, marimo should have realized that `a` is not accessible and not tried to run the cells. Also, `b` is not accessible because `a` is not accessible.

I think more aggressive checking of refs could solve this. Here's a silly example that shows how the dep, ref checking is a little broken (beyond stop):

https://marimo.app/l/tq20qa

akshayka:
> I think more aggressive checking of refs could solve this. Here's a silly example that shows how the dep, ref checking is a little broken (beyond stop):

What would you expect `refs` and `defs` to be in your example?

dmadisetti:
I would expect it not to run. Maybe throw an "Unsatisfable context" or even "This cell is in a cycle: cell-0
-> x -> cell-0".

because even in this case (similar to the original example): https://marimo.app/l/pdym22

marimo doesn't check to see if `x` exists, just that the cell that declares the def has run. It should check:
 - the variable exists
 - the cell has run 
 - the cell is not "stopped."
 
Right? I don't know—these are weird edge cases, but I think they might surface in cases like we're seeing with `stop`. A sloppy "except all" could also make this pop up.

Or maybe this is getting unrelated to the stop issue (I can open another issue)

mscolnick:
@dmadisetti is this good to close now with this https://github.com/marimo-team/marimo/pull/1580 (again, awesome work!)

[...]

Issue #1142, opened 2024-04-16
Optionally deep clone run outputs


dmadisetti:
### Describe the bug

I noticed when I did the qmd export for quarto that plots don't properly work:

* [https://dmadisetti.github.io/quarto-marimo/plots.html](https://dmadisetti.github.io/quarto-marimo/plots.html)
* [https://dmadisetti.github.io/quarto-marimo/plots-pdf.pdf](https://dmadisetti.github.io/quarto-marimo/plots-pdf.pdf)
  It looks like the plots all refer to the global `plt.gca` object rather than its snapshot at run.

This doesn't affect the reactive notebook since the plot object is rendered to HTML, essentially creating a snapshot, and then the next cell is executed. However, in script mode the output is just a pointer to the current global state

Suggestion: I think deep copies can be expensive, so maybe an optional deep copy?

### Environment

3.12-dev (I haven't pulled in a few days but I don't think you've made any changed that would affect this)

### Code to reproduce

Relevant export test:

```python
 
import matplotlib.pyplot as plt
 
if __name__ == "__main__":
    objs, _ = app.run()
    for i, o in enumerate(objs):
        # Unpack the Line2d cases
        if isinstance(o, list):
          o, = o
        # Quick hack for testing
        if "matplotlib" in str(type(o)):
          o.get_figure().savefig(f"{i}.png")
```

<details><summary>for completion the [plots.py](http://plots.py) tutorial</summary>

```python
# Copyright 2024 Marimo. All rights reserved.
import marimo

__generated_with = "0.1.69"
app = marimo.App()


@app.cell(hide_code=True)
def __(mo):
    mo.md("# Plotting")
    return


@app.cell(hide_code=True)
def __(check_dependencies):
    check_dependencies()
    return


@app.cell(hide_code=True)
def __(mo):
    mo.md(
        """
        marimo supports several popular plotting libraries, including matplotlib,
        plotly, seaborn, and altair. 

        This tutorial gives examples using matplotlib; other libraries are
        used similarly.
        """
    )
    return


@app.cell(hide_code=True)
def __(mo):
    mo.md("## Matplotlib")
    return


@app.cell(hide_code=True)
def __(mo):
    mo.md(
        """
        To show a plot, include it in the last expression of a cell (just
        like any other output).

        ```python3
        # create the plot in the last line of the cell
        import matplotlib.pyplot as plt
        plt.plot([1, 2])
        ```
        """
    )
    return
 
 
@app.cell
def __(plt):
[...]

PR #1580, opened 2024-06-08
Strict Execution

dmadisetti:
## 📝 Summary

Closes out #1142 and #1509 

This makes the kernel more deterministic by deep copying between cell passes, and ensuring reference existence before execution. It also restricts the scope of execution variables. Additionally, it introduces `mo.zero_copy` and `mo.shallow_copy`, which indicate to the kernel not to copy values (e.g. for very expensive variables).

Previous behavior:

image

New Behavior:

image

With zero_copy

image

Missing:
  - [x] More unit tests (specifically ast / executor)
  - [ ] UI love for Error on missing references
  - [x] Documentation

## 🔍 Description of Changes

This ended up being a much bigger change than expected. I think it still needs some more unit tests, but wanted to get some eyes on it / see if it passes  CI before doing anything more.

It's a pretty invasive change adding an "Executor" class and a copy module. Also had to get block level refs, so it touches AST. Also had to get state working, so had to tinker with that. Took some license with the structuring.

Execution is under "experimental" since I think this needs to be used a little before stability guarantees. currently it does not support script or direct cell runs.

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/__init__.py
+++ b/marimo/__init__.py
@@ -27,7 +27,6 @@ __all__ = [
     "capture_stderr",
     "center",
     "cli_args",
-    "copy",
     "defs",
     "doc",
     "download",
@@ -95,7 +94,7 @@ from marimo._plugins.stateless.style import style
 from marimo._plugins.stateless.tabs import tabs
 from marimo._plugins.stateless.tree import tree
 from marimo._plugins.stateless.video import video
-from marimo._runtime import copy, output
+from marimo._runtime import output
 from marimo._runtime.capture import (
     capture_stderr,
     capture_stdout,
--- a/marimo/__init__.py
+++ b/marimo/__init__.py
@@ -27,6 +27,7 @@ __all__ = [
     "capture_stderr",
     "center",
     "cli_args",
+    "copy",
     "defs",
     "doc",
     "download",
@@ -94,7 +95,7 @@ from marimo._plugins.stateless.style import style
 from marimo._plugins.stateless.tabs import tabs
 from marimo._plugins.stateless.tree import tree
 from marimo._plugins.stateless.video import video
-from marimo._runtime import output
+from marimo._runtime import copy, output
 from marimo._runtime.capture import (
     capture_stderr,
     capture_stdout,
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -31,6 +31,14 @@ class VariableData:
 
     # If kind == function or class, it may be dependent on externally defined
     # variables.
+    #
+    # NB: This is populated by `ScopedVisitor.ref_stack`. Ref stack holds the
+    # references required for the current context, it's more general than a
+    # "block", since it covers all variable level interactions.
+    # e.g.
+    # >> x = foo + bar
+    # x has the required refs foo and bar, and ref_stack holds that context
+    # while traversing the tree.
     required_refs: set[Name] = field(default_factory=set)
 
     # For kind == import
@@ -68,7 +76,7 @@ class ObscuredScope:
 
 @dataclass
 class RefData:
-    """Metadata about variables referenced but not defined by a cell"""
+    """Metadata about variables referenced but not defined by a cell."""
 
     # Whether the ref was deleted
     deleted: bool
@@ -79,6 +87,7 @@ class RefData:
 class ScopedVisitor(ast.NodeVisitor):
     def __init__(self, mangle_prefix: Optional[str] = None) -> None:
         self.block_stack: list[Block] = [Block()]
+        # Names to be loaded into a variable required_refs
         self.ref_stack: list[set[Name]] = [set()]
         self.obscured_scope_stack: list[ObscuredScope] = []
         # Mapping from referenced names to their metadata
@@ -288,9 +297,20 @@ class ScopedVisitor(ast.NodeVisitor):
             super().generic_visit(node)
 
     def _visit_and_get_refs(self, node: ast.AST) -> set[Name]:
+        """Create a ref scope for the variable to be declared (e.g. function,
+        class), visit the children the node, propagate the refs to the higher
+        scope and then return the refs."""
         self.ref_stack.append(set())
         self.generic_visit(node)
         refs = self.ref_stack.pop()
+        # The scope a level up from the one just investigated also is dependent
+        # on these refs. Consider the case:
+        # >> def foo():
+        # >>   def bar(): <- current scope
+        # >>     print(x)
+        #
+        # the variable `foo` needs to be aware that it may require the ref `x`
+        # during execution.
         self.ref_stack[-1].update(refs)
         return refs
 
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -79,7 +79,7 @@ class RefData:
 class ScopedVisitor(ast.NodeVisitor):
     def __init__(self, mangle_prefix: Optional[str] = None) -> None:
         self.block_stack: list[Block] = [Block()]
-        self.ref_stack: list[dict[Name, RefData]] = [{}]
+        self.ref_stack: list[set[Name]] = [set()]
         self.obscured_scope_stack: list[ObscuredScope] = []
         # Mapping from referenced names to their metadata
         self._refs: dict[Name, RefData] = {}
@@ -157,7 +157,7 @@ class ScopedVisitor(ast.NodeVisitor):
             deleted=deleted,
             parent_blocks=self.block_stack[:-1],
         )
-        self.ref_stack[-1][name] = self._refs[name]
+        self.ref_stack[-1].add(name)
 
     def _remove_ref(self, name: Name) -> None:
         """Remove a referenced name."""
@@ -287,8 +287,8 @@ class ScopedVisitor(ast.NodeVisitor):
             # Other nodes that don't introduce a new scope
             super().generic_visit(node)
 
-    def _visit_and_get_refs(self, node: ast.AST) -> dict[Name, RefData]:
-        self.ref_stack.append({})
+    def _visit_and_get_refs(self, node: ast.AST) -> set[Name]:
+        self.ref_stack.append(set())
         self.generic_visit(node)
         refs = self.ref_stack.pop()
         self.ref_stack[-1].update(refs)
@@ -300,7 +300,7 @@ class ScopedVisitor(ast.NodeVisitor):
         refs = self._visit_and_get_refs(node)
         self._define(
             node.name,
-            VariableData(kind="class", required_refs=set(refs.keys())),
+            VariableData(kind="class", required_refs=refs),
         )
 
     def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
@@ -308,7 +308,7 @@ class ScopedVisitor(ast.NodeVisitor):
         refs = self._visit_and_get_refs(node)
         self._define(
             node.name,
-            VariableData(kind="function", required_refs=set(refs.keys())),
+            VariableData(kind="function", required_refs=refs),
         )
 
     def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
@@ -316,7 +316,7 @@ class ScopedVisitor(ast.NodeVisitor):
         refs = self._visit_and_get_refs(node)
         self._define(
             node.name,
-            VariableData(kind="function", required_refs=set(refs.keys())),
+            VariableData(kind="function", required_refs=refs),
         )
 
     def visit_arg(self, node: ast.arg) -> None:
@@ -355,19 +355,32 @@ class ScopedVisitor(ast.NodeVisitor):
         #   x = x
         #
         # Handling value first is required to register `x` as a ref.
+        self.ref_stack.append(set())
         self.visit(node.value)
         for target in node.targets:
             self.visit(target)
+        refs = self.ref_stack.pop()
+        self.ref_stack[-1].update(refs)
 
     def visit_AugAssign(self, node: ast.AugAssign) -> None:
+        # Augmented assign (has op)
+        # e.g., x += 1
+        self.ref_stack.append(set())
         self.visit(node.value)
         self.visit(node.target)
+        refs = self.ref_stack.pop()
+        self.ref_stack[-1].update(refs)
 
     def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
+        # Annotated assign
+        # e.g., x: int = 0
+        self.ref_stack.append(set())
         if node.value is not None:
             self.visit(node.value)
         self.visit(node.annotation)
         self.visit(node.target)
+        refs = self.ref_stack.pop()
+        self.ref_stack[-1].update(refs)
 
     def visit_NamedExpr(self, node: ast.NamedExpr) -> None:
         self.visit(node.value)
@@ -419,7 +432,12 @@ class ScopedVisitor(ast.NodeVisitor):
 
         if isinstance(node.ctx, ast.Store):
             node.id = self._if_local_then_mangle(node.id)
-            self._define(node.id, VariableData(kind="variable"))
+            self._define(
+                node.id,
+                VariableData(
+                    kind="variable", required_refs=self.ref_stack[-1]
+                ),
+            )
         elif (
             isinstance(node.ctx, ast.Load)
             and not self._is_defined(node.id)
@@ -443,6 +461,19 @@ class ScopedVisitor(ast.NodeVisitor):
                     node.id = mangled_name
                 elif block.is_defined(node.id):
                     break
+
+        # Handle refs on the block scope level, or capture cell level
+        # references.
+        if (
+            isinstance(node.ctx, ast.Load)
+            and self._is_defined(node.id)
+            and node.id not in self.ref_stack[-1]
+            and (
+                node.id not in self.block_stack[-1].defs
+                or len(self.block_stack) == 1
+            )
+        ):
+            self.ref_stack[-1].add(node.id)
 
         self.generic_visit(node)
 
@@ -498,7 +529,10 @@ class ScopedVisitor(ast.NodeVisitor):
         def visit_MatchAs(self, node: ast.MatchAs) -> None:
             if node.name is not None:
                 node.name = self._if_local_then_mangle(node.name)
-                self._define(node.name, VariableData(kind="variable"))
+                self._define(
+                    node.name,
+                    VariableData(kind="variable"),
+                )
             if node.pattern is not None:
                 # pattern may contain additional MatchAs statements in it
                 self.visit(node.pattern)
@@ -506,7 +540,10 @@ class ScopedVisitor(ast.NodeVisitor):
         def visit_MatchMapping(self, node: ast.MatchMapping) -> None:
             if node.rest is not None:
                 node.rest = self._if_local_then_mangle(node.rest)
-                self._define(node.rest, VariableData(kind="variable"))
+                self._define(
+                    node.rest,
+                    VariableData(kind="variable"),
+                )
             for key in node.keys:
                 self.visit(key)
             for pattern in node.patterns:
@@ -515,13 +552,21 @@ class ScopedVisitor(ast.NodeVisitor):
         def visit_MatchStar(self, node: ast.MatchStar) -> None:
             if node.name is not None:
                 node.name = self._if_local_then_mangle(node.name)
-                self._define(node.name, VariableData(kind="variable"))
+                self._define(
+                    node.name,
+                    VariableData(kind="variable"),
+                )
 
     if sys.version_info >= (3, 12):
 
         def visit_TypeVar(self, node: ast.TypeVar) -> None:
             # node.name is a str, not an ast.Name node
-            self._define(node.name, VariableData(kind="variable"))
+            self._define(
+                node.name,
+                VariableData(
+                    kind="variable", required_refs=self.ref_stack[-1]
+                ),
+            )
             if isinstance(node.bound, tuple):
                 for name in node.bound:
                     self.visit(name)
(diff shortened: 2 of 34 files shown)


marimo/_runtime/executor.py
```diff
@@ -0,0 +1,274 @@
...
+                            "repository to never deepcopy."
+                        ) from e
+            elif ref not in glbls["__builtins__"]:
+                if ref in cell.defs:
+                    raise NameError(
+                        f"name `{ref}` is referenced before definition."
+                    )
+                raise MarimoMissingRefError(ref)
+
+        # NOTE: Execution expects the globals dictionaty by memory reference,
```
dmadisetti, line 221: "dictionaty"

dmadisetti (commented):
Will require some clean-up iterations- but I'm going to go enjoy my Saturday!
Sorry for the huge PR

dmadisetti:
??? Unsure why __future__ isn't working for 3.8, will investigate later

akshayka:
@dmadisetti This is a very interesting PR, and the implementation is quite clean. 

I have one concern, though — this change will put marimo in a situation where some notebooks only work when execution mode is strict, whereas others will only work when execution mode is relaxed. This would make sharing marimo notebooks more difficult, since users would need to somehow know which runtime configuration to use. 

For this reason I'm not fully convinced that the benefits are large enough to warrant the complexity. But it's possible I am missing something ... Beyond the matplotlib example, how do you envision this feature being used? 

An alternative we could consider: perhaps this feature could be a good motivating candidate for the extension system we talked about.

I know we discussed this feature in #1142, but I was under the impression that that discussion was for `app.run()` and exports, not for the core notebook experience.

dmadisetti:
All relaxed notebooks should be able to run strict notebooks, but not all strict notebooks will be able to run relaxed ones (maybe there's an example I'm not seeing?).

In the cases where notebook behavior is different, in the "relaxed" case- it's indeterminant (For instance the plot example, I'm not sure why $x^3$ ran before $x^2$). Strict mode can still force "relaxed" behavior with zero_copy, and a strict mode user should take that into account if writing a notebook for a general audience. Being opt-in, comes with the understanding that the notebook must be written more carefully

I think this has come from some of my frustrations that marimo seems so close- and then I'll get an unexpected mutation- at which point it's just easier to restart the kernel. This solves that issue.

I've been trying to write some of my code with this kernel, and it's been interesting. I'm catching bugs I don't think I normally would with marimo- I even found a marimo bug with regard to exceptions #1571. It's just a step up wrt reproducibility

But yeah, it is a little invasive, but I don't think there's much overhead to normal operations. If it's too different, it could be the basis for tokarip

akshayka:
Thanks for explaining; the step up  in reproducibility makes sense, and it's cool you were able to find a bug in marimo using this.

There are cases in which I've used mutations when making UIs. There was no indeterminism there (though of course the notebook was stateful), but those notebooks were written before we had mo.state.

The RAM overhead could be very large when working with numerical data, but I suppose since it's opt-in the user knows what they're signing up for.

This is definitely useful ... Let me think a little bit more about whether this goes in marimo core (perhaps it could), or if it's the basis for tokarip?

dmadisetti:
For UI and state, there's a check that makes them automatically "no copy". but it won't catch cases like:

```python
states = [State(), State()]
```

since these are nested in a data structure.

So **1)** the strict kernel could reimplement deepcopy to do the type check _or_ **2)** stay as is, and much more simply, we can make State, UI Elements, and virtual files not deepcopyable to catch these edge cases, which seems correct beyond this PR (I don't think copy falls under their intended behavior as is—for example, I found that the ref counter with virtual files breaks if the file variable is copied).

However, you can still get the same behavior with:

```python
states = mo.copy.zero_copy([State(), State()])
# or even states = mo.copy.shallow_copy([State(), State()]) in this case
```

`marimo._ast` objects are also whitelisted. LMK if you can think of any others.

As is, I would not recommend making this as anything but experimental until it has had some actual use. I keep finding weird cases like "calling a lambda defined in another cell but using a private variable in scope and another variable in defined in a 3rd cell" will fail (this actually passes now, but just as an example).

Tests are passing after adding `strict_kernel` to `all_kernels`, and I have a local commit that adds the runtime/ state tests to be run with `strict_kernel` (there's one failing case based on `required_refs` resolution that I want to fix before I push)

Re memory overhead: This will never be more than x2 (the reference copy and the working copy), with the idle state being the same amount of memory. But this is also why I added `zero_copy`, since some data is too big to reasonably be copied every reference.

But yes, this could feasibly be extracted to something like tokarip, if the visitor changes and execution register stay

---

I don't know this is obviously a little opinionated, maybe we should find some willing beta testers

dmadisetti:
Hmm. Weird. Failing with timeout on 3.11 which is also suspiciously slow with trivial changes too: https://github.com/marimo-team/marimo/actions/runs/9437984531/job/25994902883?pr=1583

akshayka:
> Failing with timeout on 3.11 which is also suspiciously slow with trivial changes too:

Timeout is okay, 3.11 runs code coverage, so sometimes times out.

dmadisetti:
Quickly ran through the tutorials. They all work! Except this by design

image

marimo/_ast/visitor.py
```diff
@@ -29,6 +29,10 @@ def __post_init__(self) -> None:
 class VariableData:
     kind: Literal["function", "class", "import", "variable"] = "variable"
 
+    # If kind == function or class, it may be dependent on externally defined
+    # variables.
+    required_refs: set[Name] = field(default_factory=set)
```
akshayka, line 34: Oh, this is interesting -- I was just thinking how it would be useful to have this information for functions and classes. Is it correct to say that if `required_refs` is the empty set, then the function or class is pure?

marimo/_ast/visitor.py
```diff
@@ -75,6 +79,7 @@ class RefData:
 class ScopedVisitor(ast.NodeVisitor):
     def __init__(self, mangle_prefix: Optional[str] = None) -> None:
         self.block_stack: list[Block] = [Block()]
+        self.ref_stack: list[set[Name]] = [set()]
```
akshayka, line 82: Can you add a comment describing the semantics of `ref_stack`?

marimo/_ast/visitor.py
```diff
@@ -281,20 +287,42 @@ def generic_visit(self, node: ast.AST) -> None:
             # Other nodes that don't introduce a new scope
             super().generic_visit(node)
 
+    def _visit_and_get_refs(self, node: ast.AST) -> set[Name]:
+        self.ref_stack.append(set())
+        self.generic_visit(node)
+        refs = self.ref_stack.pop()
+        self.ref_stack[-1].update(refs)
+        return refs
```
akshayka, line 295: I kind of understand what's happening here. But I don't understand the point of 

```python
self.ref_stack[-1].update(refs)
```

I know I'm just misunderstanding something -- can you add a comment explaining what's happening?

marimo/__init__.py
```diff
@@ -94,7 +95,7 @@
 from marimo._plugins.stateless.tabs import tabs
 from marimo._plugins.stateless.tree import tree
 from marimo._plugins.stateless.video import video
-from marimo._runtime import output
+from marimo._runtime import copy, output
```
akshayka, line 98: Let's keep `copy` private for now, until we have more confidence in the executor and its associated APIs

tests/_ast/test_visitor.py
```diff
@@ -350,8 +370,9 @@ def foo(x=y, y=x, z=a):
     v.visit(mod)
     assert v.defs == set(["foo"])
     assert v.refs == set(["x", "y", "a"])
+    # TODO: Are these required refs?
```
akshayka, line 373: Yes, I think they are required refs -- can't define `foo` without them, right?

akshayka (commented):
Not a full code review, but left a few comments. I will continue to review later.

Overall, I think this is a great PR with several useful new features. I'm still not sure what the best way to expose the strict runtime to users will be, but like you've mentioned, we can punt on that and keep it experimental for now.

I actually think the way you've architected things, this PR isn't invasive. The executor abstraction is clean and not entangled with the rest of the codebase; easy to swap in and out.

marimo/_ast/visitor.py
dmadisetti: It's not guaranteed to be pure since you might pass in a file descriptor or do something like `__import__("requests")`, but it's a good indication that the function _might_ be pure.

tests/_ast/test_visitor.py
dmadisetti: Just checked, they are not: https://marimo.app/l/7swcsf

I'll take a look - but not sure it's worth trying to tease out this edgecase

marimo/_ast/visitor.py
dmadisetti: Added a comment here, but even more detail under variabledata.required_refs

tests/_ast/test_visitor.py
akshayka: `del` just decrements the ref count though — `foo` still refers to the `Namespace` object, for example:

image

But I guess the symbols aren't required to be in globals, but instead are bound at function construction time.

Anyway I think it's fine as is

dmadisetti:
Cool! Good comments. I hope my documentation isn't too overwrought. Today is the first day I used this branch without actively tinkering with it so

Notes from usage:
 - [ ] `NameError` catching is a little too aggressive right now. I was developing from a module and the way the error is currently thrown obfuscates the stack trace in a way that was not ideal.
 - [ ] `StopError` wasn't producing the correct UI behavior? Same area of code that needs refactor / checking
 
Misc todos
 - [ ] Clean up type checking in execution—currently, there are a bunch of random imports. I think this can be done more elegantly.
 - [ ] investigate `required_refs` from function args
 - [ ] (maybe) expand required ref precedent to ignore `np.ndarray[number]`, and make it easily extensible for more types.

 A different PR/ needs discussion:
  - [ ] prevent unintentional clone of state, files, and UI objects. These objects aren't meant to be cloned and might lead to unexpected behavior if they are cloned- so by either throwing an error in `__copy__` or `__reduce__` (or just removing the relevant magic functions), we can stop inadvertent improper usage.

I think unit tests are OK- and I'm going to hold off for external documentation until/ if this is exposed.

marimo/_runtime/executor.py
```diff
@@ -0,0 +1,389 @@
...
+        # Private UI elements should still be in memory.
+        # Or a required ref.
+        required = set()
+        for variable in cell.variable_data.values():
+            required.update(variable.required_refs)
+
+        for df in lcls:
+            if df.startswith(f"_cell_{cell.cell_id}_"):
+                if df in required or isinstance(lcls[df], UIElement):
+                    glbls[df] = lcls[df]
```
akshayka, line 309: Can you explain what's happening here with a comment? I don't follow.

marimo/_runtime/executor.py
```diff
@@ -0,0 +1,389 @@
...
+        glbls.update(backup)
+
+        defs = cell.defs
+        for df in defs:
+            if df in lcls:
+                # Overwrite will delete the reference.
+                # Weak copy holds on with references.
+                glbls[df] = lcls[df]
+            elif df in glbls:
+                del glbls[df]
```
akshayka, line 292: When would this happen? My understanding is `lcls` are the variables the cell defined after running, and `glbls` are the variables that were defined before it ran.

marimo/_runtime/executor.py
```diff
@@ -0,0 +1,389 @@
...
+
+        for df in backup:
+            if df.startswith(f"_cell_{cell.cell_id}_"):
```
akshayka, line 295: Should this be moved to a utility function? There's an `is_local` function in `visitor`, for example.

marimo/_runtime/executor.py
```diff
@@ -0,0 +1,389 @@
...
+
+    >>> x = foo()
+
+    if `x` is the only reference, should `foo` be included in the graph? It
+    depends on the context, so we defer to the type of `x` which has already
+    been computed at this point. If `x` is a known 'primitive' type, and thus
+    does not have an executable body, we can exclude `foo` from the graph.
+    However, `foo` may return a object or another function, which in turn may
+    have references; so if x doesn't match the very low bar 'primitive', its
+    `required_refs` are included in the graph.
```
akshayka, line 339: Wow, very thorough — nice.

marimo/_runtime/executor.py
```diff
@@ -0,0 +1,389 @@
...
+
+        for df in lcls:
+            if df.startswith(f"_cell_{cell.cell_id}_"):
+                if df in required or isinstance(lcls[df], UIElement):
+                    glbls[df] = lcls[df]
+
+
+def build_ref_predicate(
+    glbls: dict[str, Any],
+) -> Callable[[Name, VariableData], bool]:
```
akshayka, line 314: Can you add one sentence somewhere that describes what the returned callable does?

Something like: Returns a callable that takes a variable name and its associated data, and returns `True`  if ...

marimo/_runtime/executor.py
```diff
@@ -0,0 +1,389 @@
...
+        )
+
+    return only_scoped_refs
+
+
+def is_instance_by_name(obj: object, name: str) -> bool:
+    if not (hasattr(obj, "__module__") and hasattr(obj, "__class__")):
+        return False
+    obj_name = f"{obj.__module__}.{obj.__class__.__name__}"
+    return obj_name == name
```
akshayka, line 379: Fancy — are you doing this so the module doesn't have to be installed?

akshayka (commented):
Small comments/questions

akshayka:
All tests passing! 👀 🎉

marimo/_runtime/executor.py
dmadisetti: That's the intent; not certain if it works right now

I think the last change I want to make is to test/get this working since it seemed like it wasn't catching some things. That will probably have to wait for tomorrow evening.

marimo/_runtime/executor.py
dmadisetti: Moved, but in retrospect maybe `marimo.ast.variables` might have been better and moved over the data classes too.

marimo/_runtime/executor.py
dmadisetti: Actually, wasn't that bad. pushed this too

marimo/_runtime/executor.py
dmadisetti: Added a comment, `glbs` still may have a stale variable in place- and cells may depend on that stale variable

marimo/_runtime/executor.py
dmadisetti: The intent was that a private variable that was never used, would not get put into globals. But an unused private variable can still be an eval output, and potentially some other unexpected cases- good catch.

dmadisetti:
More misc:
 - [ ] revisit exceptions

image

---

Maybe over the weekend, just noticed this while working

marimo/_runtime/copy.py
```diff
@@ -0,0 +1,174 @@
...
+
+def _ro_fail() -> None:
+    raise RoError(
+        "Weakly copied objects are directly read only."
+        " Modification is not encouraged, but is"
+        " possible by utilizing the marimo.unwrap_copy() function."
+    )
+
+
+def shadow_wrap(ref_cls: Type[_Copy[T]], base: T) -> T:
```
akshayka, line 56: Is the point of this to try to prevent mutations of attributes? My understanding is that re-assignment will fail, but mutations by calling methods on attributes will still work.

marimo/_runtime/copy.py
```diff
@@ -0,0 +1,174 @@
...
+    will work, but:
+
+    >>> b + a
+
+    will not work, as the `__add__` method is not defined for custom objects in
+    the right operand. However, The original wrapped object can still be
+    accessed with unwrap_copy(). The internal class is named as a hint for this
+    reason, although most times this will be invisible.
+    """
+    slots = set(dir(base)) - set(
```
akshayka, line 81: Can you add a comment explaining why some dunder attrs are being excluded from slots?

marimo/_runtime/copy.py
```diff
@@ -0,0 +1,174 @@
...
+    the base object without inspect.
+
+    However, there are some limitations, e.g. some operations of the wrapped
+    object may not be associative if they are not defined for both the left
+    and right operations. For instance:
+
+    >>> a = shadow_wrap(ZeroCopy, [1, 2, 3])
+    >>> b = [4, 5, 6]
+    >>> a + b
+    [1, 2, 3, 4, 5, 6]
```
akshayka, line 70: Does that mean a zero-copied NumPy array could fail if involved in a binary operation with a vanillia NumPy array?

akshayka (commented):
Few more questions

marimo/_runtime/copy.py
dmadisetti: This is the closest I could get to a "move"

Idea is we check the variable type for ZeroCopy or ShallowCopy to determine clone behavior. Preventing direct mutation is a needed consequence of exposing all the attributes.

We could also not expose the attributes and always require an unwrap or call, and not do this "shadowing"

dmadisetti: Zero copy could just put the ID in a special dictionary. Much less complicated, but does require some trickery 

I think this is a better idea, just type checking was my first idea. Thoughts?

marimo/_runtime/copy.py
dmadisetti: So in numpy

'B + A' works but not 'A + B'

marimo/_runtime/copy.py
dmadisetti: Type checking does have the consequence of being conscious of your global variables though, and the RO forces you to unwrap and makes it clear state change is happening.

Particular to numpy, there is `.view()` which can called.

marimo/_runtime/copy.py
```diff
@@ -0,0 +1,174 @@
...
+    Ref = Union["weakref.ReferenceType[T]", Callable[[], T]]
+else:
+    Ref = Union[weakref.ReferenceType[T], Callable[[], T]]
+
+
+class CloneError(Exception):
+    """Thrown when strict execution fail to deep copy or clone."""
+
+
+class RoError(Exception):
```
akshayka, line 30: `ReadOnlyError`, for readability?

marimo/_runtime/copy.py
akshayka: > This is the closest I could get to a "move"

As in `std::move` from C++? I'm not sure I see the connection, maybe I'm missing something obvious

> Zero copy could just put the ID in a special dictionary. Much less complicated, but does require some trickery

And then get the original object in subsequent cells? That might be simpler, and would fix the issue with associative operators that you mentioned. It would also be compatible with any code that relied on checking the `type` of a zero copied object, whereas wrapping doesn't appear to be:

image

Could a special dictionary be used for `ShallowCopy`, too?

marimo/_runtime/copy.py
```diff
@@ -0,0 +1,174 @@
...
+
+    __ref__: Ref[T]
+
+
+class ZeroCopy(_Copy[T]):
+    """Wrapper class for strict execution (stops deepcopy)."""
+
+
+class ShallowCopy(_Copy[T]):
+    """Wrapper class for strict execution (does copy over deepcopy)."""
```
akshayka, line 45: I suppose this would go in the docs you'll write in a future PR (no need right now). But out of curiosity, when would someone choose `shallow_copy` over `zero_copy`?

marimo/_runtime/copy.py
akshayka: > Type checking does have the consequence of being conscious of your global variables though, and the RO forces you to unwrap and makes it clear state change is happening.

By the time you're using strict and opting into things like `zero_copy` — I imagine you're already being pretty conscious of your global variable usage. 

> the RO forces you to unwrap and makes it clear state change is happening

That's true.

I would opt for a special dictionary because it's simpler and unlike type checking doesn't have gotchas when interacting with the "wrapped" object. But I'll leave it up to you to determine the benefits of typing outweigh its costs.

marimo/_runtime/copy.py
dmadisetti: A container of object that's shouldnt be copied, like `[state, state, state]`

marimo/_runtime/copy.py
dmadisetti: Implementation wise, there would just be a special dictionary for zero and shallow copy, that the helper functions would just interact with, and then a check would be made from the global objects. Invalidating IDs on rerun might be tricky. Also, Python reuses IDs. But if not going down the id route, then there might be some manual gc that would have to happen 

I don't think it would be hard to put in but maybe a few edge cases- so I'm inclined to leave it as is for now

akshayka: > Implementation wise, there would just be a special dictionary for zero and shallow copy, that the helper functions would just interact with, and then a check would be made from the global objects. Invalidating IDs on rerun might be tricky. Also, Python reuses IDs. But if not going down the id route, then there might be some manual gc that would have to happen

Yea. That makes sense to me. I had to solve related edge cases for our registry of UI elements, during which process I learned that Python reuses IDs.

> I don't think it would be hard to put in but maybe a few edge cases- so I'm inclined to leave it as is for now

Ok sounds good.

marimo/_runtime/copy.py
akshayka: > A container of object that's shouldnt be copied, like [state, state, state]

If you don't want something to be copied, wouldn't you use `zero_copy`, not `shallow_copy`?

dmadisetti: The container can mutate w/ ZeroCopy but not shallow copy

`states.append(state)`

will be reset w/ shallow copy

dmadisetti:
Exceptions now look good 
image

dmadisetti:
Sorry for the massive PR. Even though most of it is comments and tests, I think I could have broken it up.

Will try to provide quicker turn around for additional comments since I know this is now blocking

mscolnick:
(its not blocking, I can still work off my branch) thanks for the effort on this one! definitely a tricky problem to solve

akshayka (approved):
LGTM! At this point, since it's experimental -- I think it's good to merge. I'm sure we'll learn more as you and others try it out.

Thanks for the PR!

Issue #1573, opened 2024-06-07
ui.table: full-table sorting


akshayka:
### Description

Enhance `ui.table()` to sort on the entire table contents.

Right now, mo.ui.table truncates tables to 10,000 rows. When you sort a column, eg sort descending, it only sorts on the truncated data -- and this sorting happens in the frontend. This has been confusing for our users. Instead, we would like sorting to happen on the backend. 

This feature fits with our priority of making marimo the best place to work with data.

### Suggested solution

This could be implemented by adding a "sort" Function to ui.table -- UIElements have a way of exposing RPC/Functions to the frontend.

Example of `Function` registration: https://github.com/marimo-team/marimo/blob/main/marimo/_plugins/ui/_impl/table.py#L211-L222

akshayka:
cc @wasimsandhu

akshayka:
The same enhancement could also be made to `mo.ui.dataframe()`

wasimsandhu:
Picking this up!

wasimsandhu:
Completed in #1581 :)

PR #1581, opened 2024-06-09
improvement: Server-side sorting for mo.ui.table

wasimsandhu:
## 📝 Summary

Moves column sorting to the backend for `mo.ui.table`, enabling full table sort per #1573.

## 🔍 Description of Changes

- Register `sort_values` function in `mo.ui.table` with arguments for single-column sort.
- Implement sorting for list of values, list of dicts, dict of lists, and pd/pl/pa dataframes.
- Refactor tanstack react-table in `data-table.tsx` for server-side sorting (see [this SO thread](https://stackoverflow.com/questions/74321216/tanstack-v8-react-table-how-to-sort-server-side-with-manualsorting)).
- Conditionally render `DataTableComponent` from `LoadingDataTableComponent` instead of `DataTablePlugin`.
- Call `sort_values` function in `LoadingDataTableComponent` and pass sort state to `DataTableComponent`, which passes it to `DataTable`.

## ⏰ To Do

- [x] Fix uncaught exception on frontend when clearing sort
- [x] Update sorting state on data change
- [x] Test and debug sort for list of values and dict of lists
- [x] Sorting for dataframes
- [x] Address useEffect warnings
- [x] Fix backend type errors
- [x] Extend to `mo.ui.dataframe` (?)

## 📋 Checklist

## 📜 Reviewers

@akshayka AND @mscolnick


frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -211,13 +221,39 @@ const DataTableComponent = ({
...
   );
 
   const rowSelection = Object.fromEntries((value || []).map((v) => [v, true]));
 
+  const [tableData, setTableData] = useState(data);
+  const [sorting, setSorting] = useState<SortingState>([]);
+
+  useEffect(() => {
+    setSorting([]);
+  }, [data]);
```
mscolnick, line 237: when we filter from upstream, the data may change but do we want to lose sorting? i wonder if sorting should take place in `LoadingDataTableComponent`'s first `useAsyncData`.

if sorting is enabled, we load by RPC, otherwise we load by the csv URL

marimo/_plugins/ui/_impl/table.py
```diff
@@ -270,3 +281,6 @@ def get_column_summaries(self, args: EmptyArgs) -> ColumnSummaries:
             )
 
         return ColumnSummaries(summaries)
+
+    def sort_values(self, args: SortValuesArgs) -> List[JSONType]:
+        return self._manager.sort_values(args.by, args.descending).to_data()
```
mscolnick, line 286: i think we will want to do:
```
self._unfiltered_manager.sort_values(args.by, args.descending).limit(TableManager.DEFAULT_LIMIT).to_data()
```

frontend/src/plugins/impl/DataTablePlugin.tsx
wasimsandhu: I think that's a good idea. How should we retain sorting though? I added this effect because an exception is thrown when sorting is `[{ by: "Column1", descending: true }]` for example, but the new data does not contain a column named `Column1`.

marimo/_plugins/ui/_impl/table.py
wasimsandhu: Thanks, you're right. I didn't look into the init closely enough

marimo/_plugins/ui/_impl/tables/default_table.py
```diff
@@ -140,6 +141,27 @@ def get_column_names(self) -> List[str]:
             return list(first.keys())
         return ["value"]
 
+    def sort_values(
```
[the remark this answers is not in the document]
wasimsandhu: Ah I forgot to implement sort for dataframes!

wasimsandhu:
Screenshot 2024-06-10 at 6 40 09 PM

wasimsandhu:
Summarizing how I'm blocked here after moving the sort state and function call to `LoadingDataTableComponent`.

Currently, `LoadingDataTableComponent` is rendered when the given `data` prop is of type `string` (CSV urls for large data). It then renders `DataTableComponent`. The `sorting` state is passed like:

```
LoadingDataTableComponent -> DataTableComponent -> DataTable
```

The problem is that `DataTableComponent` can be conditionally rendered when `data` is not a file url but contains the actual data. I think the solution here is to manage the loading state within `DataTableComponent`, but I don't want to perform that refactoring in case there was a specific reason to have two components instead of one.

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -50,6 +51,10 @@ type Functions = {
   get_column_summaries: (opts: {}) => Promise<{
     summaries: ColumnHeaderSummary[];
   }>;
+  sort_values: (req: {
+    by: string | null;
+    descending: boolean;
+  }) => Promise<any>;
```
mscolnick, line 57: you can do `object[]`. if that yells at you, you can just eslint ignore

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -70,10 +75,10 @@ export const DataTablePlugin = createPlugin<S>("marimo-table")
       rowHeaders: z.array(z.tuple([z.string(), z.array(z.any())])),
       fieldTypes: z
         .record(
-          z.enum(["boolean", "integer", "number", "date", "string", "unknown"]),
+          z.enum(["boolean", "integer", "number", "date", "string", "unknown"])
         )
         .nullish(),
-    }),
+    })
```
mscolnick, line 81: formatting seems off, you might want to run `cd frontend` ; `pnpm format` . also `pnpm lint` will show you the lint errors

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -90,20 +95,26 @@ export const DataTablePlugin = createPlugin<S>("marimo-table")
...
-      }),
+      })
     ),
+    sort_values: rpc
+      .input(z.object({ by: z.string().nullable(), descending: z.boolean() }))
+      .output(z.union([z.string(), z.array(z.object({}).passthrough())])),
   })
   .renderer((props) => {
     if (typeof props.data.data === "string") {
+      const [sorting, setSorting] = useState<SortingState>([]);
```
mscolnick, line 108: does this fail if you put this in `LoadingDataTableComponent`? `useState` shouldnt be in a conditional, and should be in a React component

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -122,20 +133,43 @@ interface DataTableProps extends Data<unknown>, Functions {
...
       }
       return vegaLoadData(
-        props.data,
+        tableData,
         { type: "csv", parse: getVegaFieldTypes(props.fieldTypes) },
-        { handleBigInt: true },
+        { handleBigInt: true }
       );
-    }, [props.data, props.fieldTypes, props.totalRows]);
+    }, [tableData, props.fieldTypes, props.totalRows]);
```
mscolnick, line 171: Does this work? (combining the two)

```ts
    const { data, loading, error } = useAsyncData<unknown[]>(() => {
      const getSortedData = async () => {
        if (!props.sorting || props.sorting.length === 0) {
          return props.data;
        }

        const sortedData = await props.sort_values({
          by: props.sorting[0].id,
          descending: props.sorting[0].desc,
        });

        return sortedData;
      };

      getSortedData();

      if (props.totalRows === 0) {
        return Promise.resolve([]);
      }

      return vegaLoadData(
        await getSortedData(),
        { type: "csv", parse: getVegaFieldTypes(props.fieldTypes) },
        { handleBigInt: true }
      );
    }, [props.sorting, props.sort_values, props.fieldTypes, props.totalRows]);

File: marimo/_plugins/ui/_impl/table.py [hunk 8]
```diff
@@ -61,6 +61,12 @@ class ColumnSummaries:
     summaries: List[ColumnSummary]
 
 
+@dataclass
+class SortValuesArgs:
+    by: Optional[str]
```
mscolnick, line 66: what does optional mean? should this be ` by: str`?

marimo/_plugins/ui/_impl/table.py
```diff
@@ -273,3 +284,10 @@ def get_column_summaries(self, args: EmptyArgs) -> ColumnSummaries:
             )
 
         return ColumnSummaries(summaries)
+
+    def sort_values(self, args: SortValuesArgs) -> JSONType:
+        return (
+            self._unfiltered_manager.sort_values(args.by, args.descending)
```
mscolnick, line 290: if `by` is allowed to be optional? can we pull the optional check to this function - so its required downstream (which should clean up the code)

marimo/_plugins/ui/_impl/tables/utils.py
```diff
@@ -50,3 +50,31 @@ def get_table_manager_or_none(data: Any) -> TableManager[Any] | None:
         return DataFrameProtocolTableManager(data)
 
     return None
+
+# Copied from Altair
+# https://github.com/vega/altair/blob/18a2c3c237014591d172284560546a2f0ac1a883/altair/utils/data.py#L343
+def arrow_table_from_dataframe_protocol(
```
mscolnick, line 56: we can actually delete this - this got moved elsewhere

frontend/src/plugins/impl/DataTablePlugin.tsx
wasimsandhu: I put this outside of `Loading` so that I could pass it as a prop, since `memo` requires props to change in order for the component to re-render

frontend/src/plugins/impl/DataTablePlugin.tsx
wasimsandhu: Can't call `await` inside the useAsyncData hook:

```
'await' expressions are only allowed within async functions and at the top levels of modules.ts(1308)
DataTablePlugin.tsx(145, 62): Did you mean to mark this function as 'async'?
```

marimo/_plugins/ui/_impl/table.py
wasimsandhu: For a list of values (i.e. 1D list with no column name), I meant to set `by` to `null` so that the `sort_values` function would just sort the list.

Another approach could be to pass a custom `by` value, but there's the one in million chance that someone could use that as a column name.

A third arg seemed like overkill

marimo/_plugins/ui/_impl/table.py
wasimsandhu: good idea

frontend/src/plugins/impl/DataTablePlugin.tsx
mscolnick: Sorry it should be `const { data, loading, error } = useAsyncData<unknown[]>(async () => {`

marimo/_plugins/ui/_impl/table.py
mscolnick: when we pass a list of single values, we actually add a column `"value"`. that can be the sortable column. 

so you could do something like `return sorted(self._normalize_data(self.data), ...)` and you will always have a column

wasimsandhu: Worked like a charm! Thank you

frontend/src/plugins/impl/DataTablePlugin.tsx
wasimsandhu: can't remember why I had it there

wasimsandhu:
Adding some unit tests, brb.

mscolnick (approved):
nice work! this will be super valuable

wasimsandhu:
Woohoo!

PR #1589, opened 2024-06-10
improvement: make ibis and dataframe-like tables lazy evaluate

### LINKED ISSUE #1586: Notebooks with large dataframes hang when opening
Hi there,cool for implementing that new feature! 

Could it be, that the datasources panel automaticly retrieves informations from every datasource? As an Ibis-table can be a connection to a very large databasetable this could take a while and it would be nice to surpress this on demand.

With the change from #1522, my old notebook hangs when opening

_Originally posted by @szst11 in https://github.com/marimo-team/marimo/issues/1522#issuecomment-2158671559_

mscolnick:
Fixes #1586

- make ibis and dataframe-like tables lazy evaluate (specifically for num rows)
- add `DataFrameProtocolTableManager` which can be lazier and more performant when displaying in the datasources panel

### DIFF
--- a/frontend/src/components/editor/chrome/panels/datasources-panel.tsx
+++ b/frontend/src/components/editor/chrome/panels/datasources-panel.tsx
@@ -222,14 +222,16 @@ const DatasetTableItem: React.FC<{
   isExpanded: boolean;
 }> = ({ table, onExpand, onAddTable, isExpanded }) => {
   const renderRowsByColumns = () => {
-    if (table.num_rows == null || table.num_columns == null) {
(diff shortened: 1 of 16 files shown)


akshayka (approved):

szst11:
I expect, that the num rows operation could already be an intense operation, if the lazy dataframe e.g. contains some JOINs on a large table.

mscolnick:
@szst11, that makes sense. We now should only be fetching the rows when displaying in a table or chart (we need to resolve the DF anyways to display it). We no longer read the rows on startup / to display it in the sidebar.

PR #1596, opened 2024-06-11
chore: move frontend to generated api types and fetcher

mscolnick:
- Migrate to code-generated openapi types
- Use fetcher which is typesafe, and rely on typescript inference more
- Some dataclass shape changes for consitency
- Some dataclass renaming for clarify


frontend/src/core/network/requests-network.ts
```diff
@@ -1,241 +1,239 @@
...
-      return API.post<SetComponentValuesRequest>(
-        "/kernel/set_ui_element_value",
-        {
-          objectIds: objectIds,
-          values: values,
-        },
-      );
+    sendComponentValues: (request) => {
+      return marimoClient
+        .POST("/api/kernel/set_ui_element_value", {
```
mscolnick, line 11: these are typesafe, although it looks similar to the previous urls

akshayka (approved):
LGTM conditional on CI

PR #1603, opened 2024-06-12
fix: false positive dataframe-like check

### LINKED ISSUE #1602: DataFrameLike False Positive When __getattr__ Presents
### Describe the bug

When a class implemented `__getattr__` or `__getattribute__`, marimo will treat this instance as a `DataFrameLike` object, trying to render a table from it and raise exceptions.

### Environment

{
  "marimo": "0.6.17",
  "OS": "Darwin",
  "OS Version": "23.5.0",
  "Processor": "arm",
  "Python Version": "3.11.5",
  "Binaries": {
    "Browser": "125.0.6422.142",
    "Node": "v22.1.0"
  },
  "Requirements": {
    "click": "8.1.7",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pygments": "2.18.0",
    "tomlkit": "0.12.5",
    "uvicorn": "0.30.1",
    "starlette": "0.37.2",
    "websockets": "12.0",
    "typing-extensions": "4.7.1",
    "black": "24.4.2"
  }
}

### Code to reproduce

```python
class Boom:
    def __getattr__(self, _):
        return ...
```

```python
b = Boom()
```

This will result an error:
`TypeError: 'ellipsis' object is not callable`

mscolnick:
Fixes #1602

This is a bit more robust checking if its "dataframe-like" and has the dataframe protocol

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/types.py
+++ b/marimo/_plugins/ui/_impl/tables/types.py
@@ -11,3 +11,11 @@ class DataFrameLike(Protocol):
     def __dataframe__(
         self, nan_as_null: bool = False, allow_copy: bool = True
     ) -> DataFrame: ...
+
+
+def is_dataframe_like(value: object) -> bool:
+    return (
+        isinstance(value, DataFrameLike)
+        and hasattr(value, "__dataframe__")
+        and callable(value.__dataframe__)
+    )
--- a/marimo/_plugins/ui/_impl/tables/types.py
+++ b/marimo/_plugins/ui/_impl/tables/types.py
@@ -18,4 +18,8 @@ def is_dataframe_like(value: object) -> bool:
         isinstance(value, DataFrameLike)
         and hasattr(value, "__dataframe__")
         and callable(value.__dataframe__)
+        # By checking these are equal, we likely guarding against
+        # __getattr__ implementations that may different objects
+        and value.__dataframe__ == value.__dataframe__
+        # We don't want to call __dataframe__, in case it has side effects
     )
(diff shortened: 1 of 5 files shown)


marimo/_plugins/ui/_impl/tables/types.py
```diff
@@ -11,3 +11,11 @@ class DataFrameLike(Protocol):
     def __dataframe__(
         self, nan_as_null: bool = False, allow_copy: bool = True
     ) -> DataFrame: ...
+
+
+def is_dataframe_like(value: object) -> bool:
+    return (
+        isinstance(value, DataFrameLike)
+        and hasattr(value, "__dataframe__")
+        and callable(value.__dataframe__)
```
Somainer, line 20: What if `__getattr__` itself returns callable?

marimo/_plugins/ui/_impl/tables/utils.py
```diff
@@ -46,7 +48,7 @@ def get_table_manager_or_none(data: Any) -> TableManager[Any] | None:
                 return manager(data)
 
     # If we have a DataFrameLike object, use the DataFrameProtocolTableManager
-    if isinstance(data, DataFrameLike):
+    if is_dataframe_like(data):
         return DataFrameProtocolTableManager(data)
```
Somainer, line 52: Maybe we can add a fall back to normal rendering approach if `__dataframe__` method itself returns a non-dataframe object or raises an exception.

marimo/_plugins/ui/_impl/tables/types.py
mscolnick: added more defensiveness: checking `value.__dataframe__ is value.__dataframe__` which could guard against some (but not all) of `__getattr__` returning a callable.

also added a check inside `DataFrameProtocolTableManager` after constructing the dataframe to check its shape further and falling back to `None`.

marimo/_plugins/ui/_impl/tables/utils.py
mscolnick: done, thanks for the suggestion!

PR #1604, opened 2024-06-12
improvement: better errors when mutated value or on_change

mscolnick:
We can educate when people try to mutate

### DIFF
--- a/marimo/_plugins/ui/_core/ui_element.py
+++ b/marimo/_plugins/ui/_core/ui_element.py
@@ -286,17 +286,13 @@ class UIElement(Html, Generic[S, T], metaclass=abc.ABCMeta):
             "consider using mo.state()."
         )
 
-    @property
-    def on_change(self) -> Optional[Callable[[T], None]]:
-        return self._on_change
-
-    @on_change.setter
-    def on_change(self, on_change: Optional[Callable[[T], None]]) -> None:
-        del on_change
-        raise RuntimeError(
-            "Setting the on_change handler of a UIElement is not allowed. "
-            "You must set the on_change in the constructor."
-        )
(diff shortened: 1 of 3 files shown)


dmadisetti:
Nice. Since we're here, does copying a UI element break things? It looked like it when I was playing around

mscolnick:
@dmadisetti i'm not sure. @akshayka would know better

marimo/_plugins/ui/_core/ui_element.py
```diff
@@ -269,6 +277,27 @@ def value(self) -> T:
...
+        return self._on_change
+
+    @on_change.setter
+    def on_change(self, on_change: Optional[Callable[[T], None]]) -> None:
+        del on_change
+        raise RuntimeError(
+            "Setting the on_change handler of a UIElement is not allowed. "
+            "You must set the on_change in the constructor."
+        )
+
```
akshayka, line 300: I feel like we don't need these, since .on_change doesn't exist UIElement.

akshayka:
I think copying might break things, but I need to double check. There is special logic for cloning.

akshayka (commented):
Nice! One comment about on_change.

marimo/_plugins/ui/_core/ui_element.py
akshayka: Instead, this could be:

- remove the property
- add

```python
def __setattr__(self, name: str, value: Any) -> None:
  if name == "on_change":
        raise RuntimeError(
          "Setting the on_change handler of a UIElement is not allowed. "
          "You must set the on_change in the constructor."
      )
  super().__setattr__(name, value)
```

akshayka (approved):

PR #1605, opened 2024-06-12
fix: dataframe be sorting follow-ups

mscolnick:
- remove useEffect usage and consolidate loading logic in dataframe plugin
- better generic types in dataframe plugin
- handle sorting when types are not comparable in the `DefaultTableManager`

### DIFF
--- a/frontend/src/components/data-table/column-header.tsx
+++ b/frontend/src/components/data-table/column-header.tsx
@@ -33,11 +33,8 @@ export const DataTableColumnHeader = <TData, TValue>({
     return <div className={cn(className)}>{header}</div>;
   }
 
-  const sortFn = column.getSortingFn();
-  const AscIcon =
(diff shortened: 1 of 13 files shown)


frontend/src/components/data-table/column-header.tsx
```diff
@@ -33,11 +33,8 @@ export const DataTableColumnHeader = <TData, TValue>({
     return <div className={cn(className)}>{header}</div>;
   }
 
-  const sortFn = column.getSortingFn();
-  const AscIcon =
-    sortFn === sortingFns.basic ? ArrowDown01Icon : ArrowDownNarrowWideIcon;
-  const DescIcon =
-    sortFn === sortingFns.basic ? ArrowDown10Icon : ArrowDownWideNarrowIcon;
```
mscolnick, line 40: not sure if it was a regression from BE sorting. or another change, but this logic doesnt work well inferring the icon to show, so instead just always default to `ArrowDownNarrowWideIcon/vice-versa`

PR #1618, opened 2024-06-13
fix:  marimo home directory shouldn't be in sys.path

akshayka:
This change fixes a bug in which the directory in which marimo home was started (when using marimo edit) was incorrectly added to sys.path even when working in a notebook in a subdirectory. This led to modules being importable that shouldn't have been importable.


mscolnick (approved):
could this technically be breaking for some users?

akshayka:
> could this technically be breaking for some users?

maybe ... but I count it as a bug not a feature. no one should have been relying on it because the behavior was bizarre

PR #1619, opened 2024-06-13
fix: Save cell execution times for session resume

### LINKED ISSUE #1330: On resume session (browser refresh), execution time is lost
### Describe the bug

From @delennc: reconnecting to an active notebook, you lose the info on how long it took to run each cell.

### Environment

{
  "marimo": "0.4.10",
  "OS": "Linux",
  "OS Version": "6.5.0-27-generic",
  "Processor": "x86_64",
  "Python Version": "3.10.12",
  "Binaries": {
    "Browser": "123.0.6312.122",
    "Node": "v18.13.0"
  },
  "Requirements": {
    "click": "8.1.3",
    "importlib-resources": "6.1.0",
    "jedi": "0.19.0",
    "markdown": "3.4.4",
    "pymdown-extensions": "10.3",
    "pygments": "2.16.1",
    "tomlkit": "0.12.0",
    "uvicorn": "0.24.0.post1",
    "starlette": "0.27.0",
    "websocket": "missing",
    "typing-extensions": "4.8.0",
    "black": "23.3.0"
  }
}

### Code to reproduce

_No response_

wasimsandhu:
## 📝 Summary
During cell execution, execution times are now saved in `SessionView`. On session resume (browser refresh), these execution times are displayed on initial load. Otherwise, they are calculated by the frontend as currently designed.

Fixes #1330.

## 🔍 Description of Changes

Cell execution time is tracked by the cell operation status. If cell goes from "queued" to "running", a start time is saved. Once it goes from "running" to "idle", the difference is calculated and saved in milliseconds.

Gaining a better understanding of how sessions and cell operations are managed, but it is still incomplete, so it's likely this is a naive approach. As always, please let me know what I'm missing. 🙂

## 📋 Checklist

## 📜 Reviewers

@mscolnick


marimo/_server/session/session_view.py
```diff
@@ -98,6 +101,13 @@ def add_operation(self, operation: MessageOperation) -> None:
             self.cell_operations[operation.cell_id] = merge_cell_operation(
                 previous, operation
             )
+            if not previous:
+                return
+            if previous.status == "queued" and operation.status == "running":
+                self.set_execution_time(operation.cell_id, "start")
+            if previous.status == "running" and operation.status == "idle":
```
mscolnick, line 108: I think it can be interrupted to also end the time (not sure what we actually show on the frontend today if interrupted)

marimo/_server/session/session_view.py
```diff
@@ -156,6 +166,19 @@ def get_cell_console_outputs(
                 outputs[cell_id] = as_list(cell_op.console)
         return outputs
 
+    def set_execution_time(
+        self, cell_id: CellId_t, event: Literal["start", "end"]
+    ) -> None:
+        """Updates execution time for given cell."""
+        if event == "start":
+            time_elapsed = time.time()
```
mscolnick, line 174: We have a timestamp on the operation. Probably the same - but better to use that one. It will work better in tests since we could mock out those times instead of these ones.

marimo/_server/session/session_view.py
wasimsandhu: When cell execution is interrupted, the time elapsed before interruption is stickied to the cell. With these changes, the behavior is the same after session resume:

Screenshot 2024-06-13 at 7 50 00 PM

wasimsandhu:
Backend timestamp calculation seems to be 0.01s slower, should we just subtract the difference for accuracy?

Screenshot 2024-06-13 at 7 57 08 PM

mscolnick (approved):

PR #1621, opened 2024-06-15
improvement: add more types to the openapi schema

mscolnick:
## 📝 Summary

Add more open-api generated datatypes


mscolnick:
will merge after #1580  to avoid the merge conflict

marimo/_cli/development/commands.py
```diff
@@ -14,21 +17,82 @@
...
+    python_type_name,
+    python_type_to_openapi_type,
+)
 
 
 def _generate_schema() -> dict[str, Any]:
-    # dataclass components that we can generate schemas for
-    COMPONENTS = [
+    # dataclass components used in websocket messages
+    # these are always snake_case
```
akshayka, line 35: Aren't these CamelCase?

akshayka (dismissed):

marimo/_cli/development/commands.py
mscolnick: they are not, they are snake case. and not sure we can migrate to cameCase since the `data` attribute on CellOutput would be converted too. 

i think we could if we make `deep_camel_case` only operate on dataclass fields

akshayka: Oh, I was just looking at the class names, but the field names are snake case, which is what you must be referring too. Got it, makes sense

Issue #1039, opened 2024-03-31
Feedback when computation finishes


Eloitor:
### Description

When running long computations I'm often waiting for it to finish. I'd like to get some feedback when computation finishes.

### Suggested solution

- Configurable audio feedback
- Configurable browser notifications

dmadisetti:
colab changes the favicon for running, idle and failed- great visual cue from different tab

wasimsandhu:
Grabbing this next 🥸 cc @akshayka

akshayka:
Sounds good! Let's do the favicon solution suggested in this thread?

wasimsandhu:
Sure! Favicon alert by default, do we want optional settings to allow audio alerts and browser notifications too?

mscolnick:
@wasimsandhu i think those notifications could be valuable. 

Probably behind a setting and I think maybe only when the page is not active (there a DOM attribute to check if the document is active/visible)

wasimsandhu:
Picking this up this week 🫡

mscolnick:
Closed by https://github.com/marimo-team/marimo/pull/1634

PR #1634, opened 2024-06-18
feat: Feedback on notebook run completion

wasimsandhu:
## 📝 Summary
Gives user feedback on notebook run completion via dynamic favicon and optional system notifications.

Addresses feature request in #1039.

### Dynamic favicon
Favicon changes to indicate whether a notebook is running, and whether the run completed successfully or encountered errors.

This favicon persists until the user navigates back to their notebook, or resets back to the marimo favicon after 3 seconds if the notebook is in focus.

Favicon when notebook is running:

blue play

Favicon when a run completed successfully:

green check

Favicon when a run produced errors:

red x

### Notifications
Implemented by @mscolnick, more info in [this comment](https://github.com/marimo-team/marimo/pull/1634#issuecomment-2176296333).

Screenshot 2024-06-18 at 7 08 46 PM
Screenshot 2024-06-18 at 7 09 16 PM

## 📋 Checklist

## 📜 Reviewers

@mscolnick


wasimsandhu:
@mscolnick My changes work in development mode, but I can't get the favicons to appear when running in prod. Tried clearing my cache and testing in incognito, but no luck.

akshayka:
Thanks for this, @wasimsandhu!

> Add optional app (user?) settings for audio and system notifications on run completion.

I'd say to do just the favicon for now. I'd prefer not to add another app setting (unless people really want it).

frontend/src/components/editor/app-container.tsx
```diff
@@ -20,6 +27,52 @@ export const AppContainer: React.FC<PropsWithChildren<Props>> = ({
...
+    favicon.href = errors.length > 0 ? favicons.error : favicons.success;
+    resetFaviconIfComplete();
+    return () => {
+      favicon.href = favicons.idle;
+    };
+  }, [isRunning, errors, favicon, favicons, resetFaviconIfComplete]);
+
+  useEventListener(document, "visibilitychange", (_) => {
+    resetFaviconIfComplete();
+  });
```
mscolnick, line 74: can you we move this logic to its own react component 
`<DynamicFavicon/>`

it could be in the same file, but just nicer to encapsulate the logic

frontend/src/components/editor/app-container.tsx
```diff
@@ -20,6 +27,52 @@ export const AppContainer: React.FC<PropsWithChildren<Props>> = ({
...
+  let favicon: HTMLLinkElement | null =
+    document.querySelector("link[rel~='icon']");
+
+  if (!favicon) {
+    favicon = document.createElement("link");
+    favicon.rel = "icon";
+    document.getElementsByTagName("head")[0].append(favicon);
+  }
+
+  const favicons: Record<string, string> = useMemo(
```
mscolnick, line 42: this doesnt need live in the react component. 

it could be at the module level:
```ts
const FAVICON_ROUTES = {
    idle: "./favicon.ico",
    success: "./circle-check.ico",
    running: "./circle-play.ico",
    error: "./circle-x.ico",
}
```

frontend/src/components/editor/app-container.tsx
```diff
@@ -20,6 +27,52 @@ export const AppContainer: React.FC<PropsWithChildren<Props>> = ({
...
+      return;
+    }
+    favicon.href = errors.length > 0 ? favicons.error : favicons.success;
+    resetFaviconIfComplete();
+    return () => {
+      favicon.href = favicons.idle;
+    };
+  }, [isRunning, errors, favicon, favicons, resetFaviconIfComplete]);
+
+  useEventListener(document, "visibilitychange", (_) => {
```
mscolnick, line 72: can you document how you want this to work?

i think i mentioned we could use `document.visibilityState/visibilitychange`, but that was so when the editor is active, we arent flickering the favicon and only change it when the editor is out of focus. that was only a suggestion, but up-to-you what you feel feels right.

frontend/src/components/editor/app-container.tsx
```diff
@@ -20,6 +27,52 @@ export const AppContainer: React.FC<PropsWithChildren<Props>> = ({
...
+    if (isRunning) {
+      favicon.href = favicons.running;
+      return;
+    }
+    favicon.href = errors.length > 0 ? favicons.error : favicons.success;
+    resetFaviconIfComplete();
+    return () => {
+      favicon.href = favicons.idle;
+    };
+  }, [isRunning, errors, favicon, favicons, resetFaviconIfComplete]);
```
mscolnick, line 70: we can probably inline `resetFaviconIfComplete`, because we will want to clear the timeout too on the returned `unsubscribe`

```ts
  useEffect(() => {
    if (isRunning) {
      FAVICONS.href = favicons.running;
      return;
    }
    favicon.href = errors.length > 0 ? FAVICONS.error : FAVICONS.success;
    // Reset back to idle after 3 seconds
    const timeout = setTimeout(() => {
        favicon.href = FAVICONS.idle;
      }, 3000);
    return () => {
      favicon.href = FAVICONS.idle;
      clearTimeout();
    };
  }, [isRunning, errors.length]);
```

frontend/src/components/editor/app-container.tsx
mscolnick: also - im not sure this actually solves the ticket yet. i think what might happen with the above is:
1. goes to running
2. shows the success
3. then clears after 3 seconds

If you miss that 3 second window, you don't really get the feedback. Maybe the logic could be:

1. goes to running
2. shows the success
3. then clears after 3 seconds ONLY if the document is focused, otherwise keep it there
4. clear when the user comes back into focus (could be after 3 seconds or right away)

also worth documenting this logic in a comment on the `DynamicFavicon` component

mscolnick:
if this feels right, you could also use the native browser notification. this way you dont need to add anything to user settings, and they can configure it on the first try (and if they deny access, it wont happen again)

```ts
function maybeSendNotification() {
  // If in focus, skip
  if (document.visibilityState === "visible") {
    return;
  }

  const sendNotification = () => {
    if (succeeded) {
      const notification = new Notification("Execution completed", {icon: icon});
    } else {
      const notification = new Notification("Execution failed", {icon: icon}););
    }
  }

  if (!("Notification" in window)) {
    // Skip no support
  } else if (Notification.permission === "granted") {
    // Permission already granted
    sendNotification();
  } else if (Notification.permission !== "denied") {
    // We need to ask the user for permission
    Notification.requestPermission().then((permission) => {
      // If the user accepts, let's create a notification
      if (permission === "granted") {
        sendNotification();
      }
    });
  }
}
```

mscolnick:
@wasimsandhu for it working in production - you might need to add those assets to here:
https://github.com/marimo-team/marimo/blob/main/marimo/_server/api/endpoints/assets.py#L93-L98

wasimsandhu:
> @wasimsandhu for it working in production - you might need to add those assets to here: https://github.com/marimo-team/marimo/blob/main/marimo/_server/api/endpoints/assets.py#L93-L98

Already tried this, see [here](https://github.com/marimo-team/marimo/pull/1634/files#diff-78885c9877079be682b42199747571c55f176538e9bd71f1f5d99906719eb799R950).

frontend/src/components/editor/app-container.tsx
wasimsandhu: Sure, I'll add some comments in the file.

You're right, the favicon is reset after 3 seconds if the page is visible, but remains if it is not. I agree that it should persist if the document is not in focus; I'll add that in.

frontend/src/components/editor/app-container.tsx
wasimsandhu: Good catch thanks!

mscolnick (approved):

mscolnick:
nice! looks great!

PR #1639, opened 2024-06-18
feat: kiosk consumers

mscolnick:
A "kiosk" consumer has a "kiosk view" when running marimo. This is some middleground before going full multi-player. It allows you to see and the outputs and interactive with the widgets, but not right any code.

This view is the editor view without the ability to actually edit any code or run cells, but you get the sidebar and other useful editor features.

A kiosk consumer doesn't have less permission than an editor from the API layer, just that we hide potentially conflicting operations. (that we do not keep in sync).

Over time, we will move more state to the backend and we can loosen what is shown, however, the use-case for this right now is for vscode/vim to be able to write code in vscode/vim but see rich outputs in marimo.


marimo/_runtime/requests.py
```diff
@@ -114,7 +114,7 @@ class CreationRequest:
 
 
 @dataclass
-class DeleteRequest:
```
akshayka, line 117: ha thanks

marimo/_server/sessions.py
```diff
@@ -264,6 +272,58 @@ def kernel_connection(self) -> TypedConnection[KernelMessage]:
         return self._read_conn
 
 
+class Room:
```
akshayka, line 275: nice!

akshayka (approved):

PR #1641, opened 2024-06-18
feat: global table search

mscolnick:
Add global search to tables. Supports:
* polars
* pandas
* pyarrow
* lists/dicts

Bugfix:
* clear selection when filter or sort changes

Other:
* i removed the mb-1 on buttons. This caused lots of issues with composition, but if this was useful somewhere, we can add this back per-case.


mscolnick:
image

akshayka (approved):

PR #1652, opened 2024-06-19
feat: column filters

### LINKED ISSUE #950: Feature request: filter tables by value in column
### Discussed in https://github.com/marimo-team/marimo/discussions/948

<sup>Originally posted by **mwchase** March 13, 2024</sup>
I'm pretty sure this isn't currently possible, but perhaps I missed part of the documentation.

I have some tabular data that includes a freeform text field; I'd like to be able to only look at cells that have values "similar to" some search string in that column. (Substring checking should be good enough, but some kind of fuzzy find would be excellent.)</div>

mscolnick:
## 📝 Summary

Closes #950

This adds "contains" and "range" filters to `mo.ui.table`. Specs:
* support text (free text that `contains` and is case-sensitive)
* range min and max. you can achieve `<= X`, `>= X`, `== X`, `X - Y` 
* boolean to set true/false

Not support types:
* dates - may wait to see if this is needed. its much hard to put this in a UI that _always_ works. for example, we would need to be smart if a range by year, month, day, seconds, etc

This works for:
* Polars
* Pandas

But does not work for (yet)
* pyarrow (easy to add once we add a TransformHandler)
* lists/dicts (could check for polars/pandas and convert, but this creates other issues with selection being a different type)

Screenshot 2024-06-20 at 2 00 34 PM
Screenshot 2024-06-20 at 2 00 22 PM
Screenshot 2024-06-20 at 2 00 09 PM
Screenshot 2024-06-20 at 1 59 34 PM


marimo/_plugins/ui/_impl/table.py
```diff
@@ -303,10 +313,22 @@ def search(self, args: SearchTableArgs) -> Union[JSONType, str]:
 
         # If no query or sort, return nothing
         # The frontend will just show the original data
-        if not args.query and not args.sort:
+        if not args.query and not args.sort and not args.filters:
             self._filtered_manager = self._manager
             return []
 
+        if args.filters:
+            # TODO: don't allow no polars and pandas to get here
```
akshayka, line 321: Just checking if this needs to be addressed now or if it's fine to get to later

akshayka (approved):

marimo/_plugins/ui/_impl/table.py
mscolnick: oh woops - i did this already, forgot to remove the todo

PR #1665, opened 2024-06-24
feat: app comp

akshayka:
Embed notebooks in other notebooks.

This change makes it possible to embed a notebook as an output in another notebook. The embedded notebook is rendered as a `vstack` of its non-None outputs, and is reactive with respect to UI element value updates.

This lets you make a more complex notebook out of smaller building-block notebooks.

Enhancements left for the future:

1. Multiple levels of composition. Right now, reactivity wrt to UI elements only allows for 1 level of nesting.
2. API for passing references to the app, or for otherwise injecting/overriding args into the embedded app.

image

image

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -21,7 +21,6 @@ from marimo._ast.cell import (
     Cell,
     CellConfig,
     CellId_t,
-    CellImpl,
 )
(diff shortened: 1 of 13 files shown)


mscolnick (approved):
nice!

PR #1686, opened 2024-06-27
fix: strict execution single cell reference error

dmadisetti:
## 📝 Summary

fix: If a cell references a variable that no other cell potentially produces, strict mode is supposed to create an error.

## 🔍 Description of Changes

A small bug in `transitive_references` (dataflow.py) stopped the expected behavior from happening, but then I realized the error UI was slightly wrong (MarimoErrorOutput.tsx, cell_runner.py, errors.py).

So, I added a unit test (test_runtime.py), and realized that Strict Errors should probably be promoted to kernel level errors (runtime.py).

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/editor/output/MarimoErrorOutput.tsx
+++ b/frontend/src/components/editor/output/MarimoErrorOutput.tsx
@@ -142,7 +142,7 @@ export const MarimoErrorOutput = ({
           </div>
         );
       case "strict-exception":
-        return error.blamed_cell === null ? (
+        return error.blamed_cell == null ? (
           <Fragment key={idx}>
             <p>{error.msg}</p>
             <Tip>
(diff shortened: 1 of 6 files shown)


frontend/src/components/editor/output/MarimoErrorOutput.tsx
```diff
@@ -142,7 +142,7 @@ export const MarimoErrorOutput = ({
           </div>
         );
       case "strict-exception":
-        return error.blamed_cell == null ? (
+        return error.blamed_cell === null ? (
```
mscolnick, line 145: `== null` can be nice since it covers `undefined` too. You may want to keep it that way.

dmadisetti: Agreed

mscolnick (approved):
nice fix!

PR #1700, opened 2024-06-30
feat: codium copilot

mscolnick:
Add supports for [Codium](https://codeium.com/). The onboarding to get an API key is not super easy, so it requires a bit of instructions. Hopefully this will get easier in future.

image

### DIFF
--- a/docs/guides/ai_completion.md
+++ b/docs/guides/ai_completion.md
@@ -37,6 +37,7 @@ marimo using `pip` if you need Copilot.
 
 ```toml
 [completion]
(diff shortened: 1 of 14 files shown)


akshayka (commented):
Is it valid to have both GH Copilot and Codium enabled? 

If it's not valid to enable both, I would suggest having a dropdown in the user config menu to first choose the copilot — GH or Codium — and then reveal additional configuration if needed (such as the Codium API key) in the form.

akshayka (approved):

PR #1709, opened 2024-07-02
feat: sql support

mscolnick:
This PR provides SQL support to each cell.

* This uses (and cleans up) some existing logic of `LanguageAdapters` which parse to and from particular python to render richer UIs. We already do this for markdown, now we do it for SQL
	* I didnt add a hotkey, like we have for markdown since it is probably less common
* We rely on duckdb for this - so we only enable it through KernelCapabilities
* We use duckdb to parse the tables from sql statements to add it to the defs of the cell during the ast visitor

image


mscolnick:
It's ready for review, but still needs extensive edge-case testing

PR #1764, opened 2024-07-10
fix: treat special variable "__" as local variable

### LINKED ISSUE #1675: Special variable name "__" not treated as local
### Describe the bug

I don't think this is really a bug / big issue, but opening an issue just to document it.

"__" is considered a global variable, and is not hidden from other cells.

I just caught it when very lazily assigning var names

### Environment

head

### Code to reproduce

https://marimo.app/l/1uelf3

ontowhee:
## 📝 Summary

Fixes #1675 

## 🔍 Description of Changes

This change has the method `is_local` returning true for the special variable name `__`.

Here is a screenshot showing that `__` is treated as a local variable. It can be redefined across multiple cells, but it cannot be referenced as if it were a global variable.

Screenshot from 2024-07-09 23-12-06

## 📋 Checklist

- [ ] ~For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://discord.gg/JE7nhX6mD8), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).~

## 📜 Reviewers

@dmadisetti

### DIFF
--- a/marimo/_utils/variables.py
+++ b/marimo/_utils/variables.py
@@ -12,4 +12,4 @@ def is_mangled_local(name: str, cell_id: CellId_t) -> bool:
 
 
 def is_local(name: str) -> bool:
-    return name.startswith("_") and not name.startswith("__")
+    return name == "__" or (name.startswith("_") and not name.startswith("__"))
--- a/tests/_ast/test_app.py
+++ b/tests/_ast/test_app.py
@@ -276,6 +276,25 @@ class TestApp:
         app.run()
 
     @staticmethod
+    def test_dunder_rewritten_as_local() -> None:
+        app = App()
+
+        @app.cell
+        def _() -> None:
+            __ = 1  # noqa: F841
+            return
+
+        @app.cell
+        def _() -> None:
+            __  # type: ignore
+            return
+
+        with pytest.raises(NameError) as e:
+            app.run()
+
+        assert "'__' is not defined" in str(e.value)
+
+    @staticmethod
     def test_app_width_config() -> None:
         app = App(width="full")
         assert app._config.width == "full"


ontowhee:

dmadisetti:
Hmm, I guess it seems intentional. I wonder if it was only meant to capture `__dunder__` functions

dmadisetti:
Thanks for the contribution though! I'm not certain why `__` is ignored. Git blame (this was initially in visitor.py) dates this logic to the initial marimo commit

@akshayka any insight?

mscolnick (dismissed):
thanks for the fix!

akshayka:
This fix is fine. The special case against double underscore `__` was meant to prevent dunder variables like `__file__` from being made local / mangled. But it's okay if `__` is local, since it's not a dunder.

Thanks!

akshayka (dismissed):

marimo/_utils/variables.py
```diff
@@ -12,4 +12,4 @@ def is_mangled_local(name: str, cell_id: CellId_t) -> bool:
 
 
 def is_local(name: str) -> bool:
-    return name.startswith("_") and not name.startswith("__")
+    return name == "__" or (name.startswith("_") and not name.startswith("__"))
```
dmadisetti, line 15: `name.startswith("_") and not (name.startswith("__") and name.endswith("__"))` ?

akshayka: but  `__` satisfies `name.startswith("__") and name.endswith("__")`

mscolnick: I think you probably want to just special case when every character is `_`

ontowhee: I was contemplating how to simplify this logic as well, but it was just a special case!

Thanks for the review!

ontowhee:
Looks like it is failing a test. I'll look into it soon and get it fixed.

tests/_ast/test_compiler.py
```diff
@@ -21,17 +21,17 @@ def test_parse_simple() -> None:
 
     @staticmethod
     def test_local_variables() -> None:
-        code = "_, y = f(x)\ndef _foo():\n  _bar = 0\nimport _secret_module as module"  # noqa: E501
+        code = "__ = 10\n_, y = f(x)\ndef _foo():\n  _bar = 0\nimport _secret_module as module"  # noqa: E501
```
ontowhee, line 24: I added `__` to this test.

tests/_ast/test_compiler.py
```diff
@@ -21,17 +21,17 @@ def test_parse_simple() -> None:
...
         cell = compile_cell(code)
         assert cell.defs == {"module", "y"}
         assert cell.refs == {"f", "x"}
         assert cell.imported_namespaces == {"_secret_module"}
 
     @staticmethod
     def test_dunder_dunder_excluded() -> None:
-        code = "__ = 10; __name__ = 20"
+        code = "__name__ = 20"
```
ontowhee, line 32: I removed `__` from the expected values for `cell.defs`.

ontowhee:
There's another test that is failing locally for me, and I am not too sure what it is about. This is the output I see:

```
        # Wait for the message
        while True:
            line = p.stdout.readline().decode()
            if line:
                assert f"Watching {temp_marimo_file}" in line
                break
    
>       assert not path.exists(temp_out_file)
E       AssertionError: assert not True
E        +  where True = <function exists at 0x7f43f4d13560>('/tmp/tmpfzp9reov/notebook.py')
E        +    where <function exists at 0x7f43f4d13560> = path.exists

tests/_cli/test_cli_export.py:391: AssertionError
```

PR #1784, opened 2024-07-12
fix: more robust app composition

akshayka:
More robust app composition when running as a notebook / in a kernel.

1. Kernel maintains a tree of runtime contexts, one for each app that is alive.
2. Introduce an async function `app.embed()` that returns an object with the apps `Html` output and defs.
3. Each nested app is executed with a runner that instantiates and manages a `Kernel` for that app.
4. The `Kernel`'s `set_ui_element_value` method is updated to recursively search child contexts for UI elements that aren't found in its context.

This enables:

1. Arbitrary nesting of apps, including with UI elements.
2. Retrieving the definitions of embedded apps.
3. Handling `mo.state` updates in child apps.
4. Accumulating imperatively appended/replaced output with `mo.output.*`

The last remaining limitation has to do with imperative output in child apps: such shows up "all at once", once the app is done executing, instead of while an app is executing. This becomes tricky because an app consists of multiple cells, so if say the second cell has imperative outputs, then the first one needs to be rendered before the second's imperative outputs are shown, otherwise the output will be jumbled.  But right now all of an app's outputs appear together, not incrementally.  If we chose to incrementally show every cell output (regardless of whether `mo.output.append` was used), that would break rendering apps in `tabs` or other elements, because our `output.*` machinery has no concept of slots.

NB: This is a breaking change because it removes the `_mime_` method on `App` that was previously added for composition. App rendering needs to be async, so we introduced `embed` instead. Hopefully not an issue -- given the limitations with composition today, its usage is likely very low.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -271,8 +271,18 @@ class App:
             runner: cell_runner.Runner,
             run_result: cell_runner.RunResult,
         ):
+            from marimo._plugins.stateless.flex import vstack
+
             del runner
-            self._cached_outputs[cell.cell_id] = run_result.output
+            if (
+                run_result.output is None
+                and run_result.accumulated_output is not None
+            ):
+                self._cached_outputs[cell.cell_id] = vstack(
+                    run_result.accumulated_output
+                )
+            else:
+                self._cached_outputs[cell.cell_id] = run_result.output
 
         filename = "<unknown>"
         ctx = get_context()
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -2,10 +2,10 @@
 from __future__ import annotations
 
 import asyncio
-from collections.abc import Awaitable
 import random
 import string
-import sys
+import weakref
+from collections.abc import Awaitable
 from dataclasses import asdict, dataclass
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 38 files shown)


marimo/_ast/app.py
```diff
@@ -143,18 +143,23 @@ def __init__(self, **kwargs: Any) -> None:
         # a TypeError.
         self._config = _AppConfig.from_untrusted_dict(kwargs)
 
-        self._cell_manager = CellManager()
+        try:
+            # nested applications get a unique cell prefix to disambiguate
+            # their graph from other graphs
+            get_context()
+            cell_prefix = str(uuid4())
+        except ContextNotInitializedError:
```
mscolnick, line 151: i've seen `except ContextNotInitializedError:` in a lot of places. maybe we could just create a util `if has_notebook_context()`, so its a bit more human readable

akshayka: done

marimo/_ast/app.py
```diff
@@ -459,43 +295,45 @@ def _run_cell_sync(
...
-        if value not in self._pending_ui_element_updates:
-            self._pending_ui_element_updates.append(value)
+    async def _set_ui_element_value(
+        self, request: SetUIElementValueRequest
+    ) -> bool:
+        app_kernel_runner = self._get_kernel_runner()
+        return await app_kernel_runner.set_ui_element_value(request)
 
-    def _mime_(self) -> tuple[KnownMimeType, str]:
+    async def embed(self) -> AppEmbedResult:
```
mscolnick, line 304: is this public? worth adding docs? im not sure they show up in the docs yet, but maybe we should add it?

mscolnick (dismissed):

PR #1799, opened 2024-07-15
fix: add cell source code line number to traceback when running as script

### LINKED ISSUE #1620: Better tracebacks when running as a script
The tracebacks when running as a script are inscrutable. This is particularly frustrating when a user runs a script that takes several hours, only to fail with a traceback that cannot be understood.

Ideally, the traceback would point to the exact line in the original script that raised the error.

Instead, tracebacks look like this today:

**Script**:

```python
import marimo

__generated_with = "0.6.19"
app = marimo.App()


@app.cell
def __(y):
    x = 0
    y = y/x
    return x, y


if __name__ == "__main__":
    app.run()
```

**Traceback**:

```bash
  File "/home/akshay/src/marimo-team/marimo/tmp/script.py", line 15, in <module>
    app.run()
  File "/home/akshay/src/marimo-team/marimo/marimo/_ast/app.py", line 360, in run
    return self._run_sync(post_execute_hooks=post_execute_hooks)
  File "/home/akshay/src/marimo-team/marimo/marimo/_ast/app.py", line 275, in _run_sync
    outputs[cid] = execute_cell(cell._cell, glbls)
  File "/home/akshay/src/marimo-team/marimo/marimo/_ast/cell.py", line 459, in execute_cell
    exec(cell.body, glbls)
  File "/tmp/marimo_467203/__marimo__cell_Hbol_.py", line 2, in <module>
NameError: name 'y' is not defined
```

Notice how the user has no idea what `__celll__Hbol__` is, the line number is relative to the cell and not the whole script (line 2 of what?), and the offending line is not even shown.

cc @delennc

ontowhee:
## 📝 Summary

Fixes #1620  

## 🔍 Description of Changes

This is a WIP, a first attempt at improving the traceback. I'm not sure if this is a good approach, and would love guidance.

- It prints out the source file path, line number, and the offending line.
- In `cell_factory`, the line number of the first line of the cell body is calculated and stored in `CellImpl._start_line`.
- The source file path is pulled from `glbls['__file__']`.
- When the cell is executed, the exception is caught, and the last traceback is updated with useful information to include the file path, line number, and offending line.
- Right now it only adds the information for the last frame if the last frame is coming from a `__marimo_<cell_id>_.py` file (otherwise the traceback will already have useful information). We can make it work for all frames, but I just wanted to get it working with the last frame as a starting point. If the frame is not coming from a `__marimo_<cell_id>_.py` file, it should already have useful information.

Example script (see `tests/_runtime/script_data/script_exception.py`):
```python
import marimo

__generated_with = "0.6.19"
app = marimo.App()


@app.cell
def __(y):
    x = 0
    y = y / x
    return x, y


if __name__ == "__main__":
    app.run()
```

Example traceback:
```
  File "/marimo/tests/_runtime/script_data/script_exception.py", line 15, in <module>
    app.run()
  File "/marimo/marimo/_ast/app.py", line 444, in run
    return self._run_internal()
           ^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_ast/app.py", line 434, in _run_internal
    return self._run_sync(
           ^^^^^^^^^^^^^^^
  File "/marimo/marimo/_ast/app.py", line 319, in _run_sync
    output = execute_cell(cell._cell, glbls, self._graph)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/executor.py", line 75, in execute_cell
    return EXECUTION_TYPES[execution_type].execute_cell(cell, glbls, graph)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/executor.py", line 163, in execute_cell
    handle_execute_cell_exception(e, cell, glbls)
  File "/marimo/marimo/_runtime/executor.py", line 126, in handle_execute_cell_exception
    raise e
  File "/marimo/marimo/_runtime/executor.py", line 161, in execute_cell
    exec(cell.body, glbls)
  File "/tmp/marimo_231999/__marimo__cell_Hbol_.py", line 2, in <module>
NameError: name 'y' is not defined
  File "/marimo/tests/_runtime/script_data/script_exception.py", line 10
    y = y / x
    ^^^^^^^^^
```

## 📋 Checklist

## 📜 Reviewers

@akshayka


akshayka:
Thanks for picking this up! Just writing to acknowledge that I've seen it, and will leave feedback as soon as I can.

marimo/_runtime/executor.py
```diff
@@ -95,6 +95,37 @@ async def execute_cell_async(
...
+
+        file_path = glbls["__file__"]
+        source_lineno = cell._start_line + cell_lineno - 1
+        underline_str = "^" * len(offending_code)
+
+        e.add_note(
+            f"""  File "{file_path}", line {source_lineno}"""
+            + f"\n    {offending_code}"
+            + f"\n    {underline_str}"
+        )
```
akshayka, line 124: We need to be compatible back to Python 3.8, but unfortunately `add_note` was added in 3.11, so we can't use it.

marimo/_runtime/executor.py
```diff
@@ -126,7 +157,11 @@ def execute_cell(
         if cell.body is None:
             return None
         assert cell.last_expr is not None
-        exec(cell.body, glbls)
+        try:
+            exec(cell.body, glbls)
+        except Exception as e:
+            handle_execute_cell_exception(e, cell, glbls)
+
         return eval(cell.last_expr, glbls)
```
akshayka, line 165: Do you need a try/catch around `eval` too?

marimo/_runtime/executor.py
```diff
@@ -95,6 +95,37 @@ async def execute_cell_async(
...
+    tb = e.__traceback__
+    while tb and tb.tb_next:
+        tb = tb.tb_next
+
+    if tb is None:
+        raise e
+
+    filename = f"__marimo__cell_{cell.cell_id}_.py"
+
+    if filename in tb.tb_frame.f_code.co_filename and cell._start_line:
```
akshayka, line 110: Is this satisfied only for executing as a script, because `cell._start_line` is nonnegative only in that case?

akshayka (commented):
The general approach looks great! Thanks for taking this on.

Left a comment and a couple of questions. One additional comment below.

Regarding frames, if a marimo cell is somewhere on the stack (even if it's not the last frame), it might help to update that frame to reference the containing file and line number? Otherwise the frame's file name with be the cell name, and the line number will be relative to the start of the cell — which won't help the user find the error in the script. I agree that the last frame is a good place to start. Example below:

```python
  File "/Users/akshay/src/marimo-team/marimo/tmp/error/main.py", line 20, in <module>
    app.run()
  File "/Users/akshay/envs/playground/lib/python3.9/site-packages/marimo/_ast/app.py", line 444, in run
    return self._run_internal()
  File "/Users/akshay/envs/playground/lib/python3.9/site-packages/marimo/_ast/app.py", line 434, in _run_internal
    return self._run_sync(
  File "/Users/akshay/envs/playground/lib/python3.9/site-packages/marimo/_ast/app.py", line 319, in _run_sync
    output = execute_cell(cell._cell, glbls, self._graph)
  File "/Users/akshay/envs/playground/lib/python3.9/site-packages/marimo/_runtime/executor.py", line 75, in execute_cell
    return EXECUTION_TYPES[execution_type].execute_cell(cell, glbls, graph)
  File "/Users/akshay/envs/playground/lib/python3.9/site-packages/marimo/_runtime/executor.py", line 130, in execute_cell
  File "/var/folders/gr/c2cb32s92zgc9vnhprvhj3nw0000gn/T/marimo_44333/__marimo__cell_MJUe__output.py", line 1, in <module>
  File "/Users/akshay/src/marimo-team/marimo/tmp/error/bug.py", line 3, in foo
    x = x / x
```

Notice how the second-to-last frame doesn't tell me where in the notebook file the error occurred.

dmadisetti:
I don't think you need to modify the stack trace at all

Untested, but in this area:

https://github.com/marimo-team/marimo/blob/82d8b74fc5acba0523d3dae2812ebc849eb9bac5/marimo/_ast/compiler.py#L98

```python
    if is_script:
        ast.increment_lineno(module, start_line - 1) # possible off by 1 error here
        body_filename = real_filename
        output_filename = real_filename
        # You could put source of the file here, but python should be able to read it directly
    else:
       # Do linecache tricks and normal usage here
```

https://docs.python.org/3.8/library/ast.html#ast.increment_lineno

mscolnick:
@ontowhee, thanks for the help on this! let us know if you think you can make the changes per the comments, or if you would prefer we take it over. either way works for us

ontowhee:
> I don't think you need to modify the stack trace at all
> 
> Untested, but in this area:
> 
> https://github.com/marimo-team/marimo/blob/82d8b74fc5acba0523d3dae2812ebc849eb9bac5/marimo/_ast/compiler.py#L98
> 
> ```python
>     if is_script:
>         ast.increment_lineno(module, start_line - 1) # possible off by 1 error here
>         body_filename = real_filename
>         output_filename = real_filename
>         # You could put source of the file here, but python should be able to read it directly
>     else:
>        # Do linecache tricks and normal usage here
> ```
> 
> https://docs.python.org/3.8/library/ast.html#ast.increment_lineno

This is interesting. I don't quite understand it yet, but let me take a look in the next few hours and see if I can get something working. If not, I'll post a comment to let you know to take over.

marimo/_ast/compiler.py
```diff
@@ -91,17 +97,22 @@ def compile_cell(code: str, cell_id: CellId_t) -> CellImpl:
...
-    if sys.version_info >= (3, 9):
-        # ast.unparse only available >= 3.9
-        cache(
-            last_expr_filename,
-            ast.unparse(expr) if not isinstance(expr, str) else "None",
-        )
+    if is_script:
+        ast.increment_lineno(module, start_line - 1)
+        body_filename = filename
+        last_expr_filename = filename
```
ontowhee, line 103: The `ast.increment_lineno` gets the correct line when running the script and also when running with the debugger. The line column is off by 4 spaces. I haven't looked into updating it yet.

```
  File "/marimo/tests/_runtime/script_data/script_exception.py", line 15, in <module>
    app.run()
  File "/marimo/marimo/_ast/app.py", line 282, in run
    outputs, glbls = AppScriptRunner(InternalApp(self)).run()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/app/script_runner.py", line 82, in run
    outputs, defs = self._run_synchronous(
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/app/script_runner.py", line 111, in _run_synchronous
    output = execute_cell(cell, glbls, self.app.graph)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/executor.py", line 83, in execute_cell
    return EXECUTION_TYPES[execution_type].execute_cell(cell, glbls, graph)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/executor.py", line 137, in execute_cell
    exec(cell.body, glbls)
  File "/marimo/tests/_runtime/script_data/script_exception.py", line 10, in <module>
    y = y / x
    ^
NameError: name 'y' is not defined
```

I haven't looked into adding the source code. I'm not entirely sure what that looks like yet. I also wasn't sure if I correctly moved the caching code into the else-block.

marimo/_ast/compiler.py
```diff
@@ -136,7 +147,10 @@ def cell_factory(
     to generate refs and defs. If the user introduced errors to the
     signature, marimo will autofix them on save.
     """
-    function_code = textwrap.dedent(inspect.getsource(f))
+    filename = f.__code__.co_filename
+    is_script = f.__globals__["__name__"] == "__main__"
```
ontowhee, line 151: I initially thought about initializing the `is_script` flag when the ScriptRuntimeContext gets initialized [here](https://github.com/marimo-team/marimo/blob/c67a2cb44f66a3c827071050e7f5e627c0829482/marimo/_runtime/app/script_runner.py#L56), and saving the value to the cell or passing it to `compile_cell()`, but it seems like the cell is compiled before the ScriptRuntimeContext is created, so I resorted to checking `__name__`.

This seems to work. I was able to run in edit mode and it did not print the name of the original script file.

ontowhee:
@dmadisetti I was able to try out `ast.increment_lineno` like you suggested, but I left comments for a few items that I wasn't sure about.

I'm happy to continue working on this, but at the same time, it will probably be faster for someone to take over. I'm glad I was able to dig into the code and learn something new! Thanks for the review!

marimo/_ast/compiler.py
dmadisetti: Hmm, not sure what you mean by the line column is off? This looks good to me

ontowhee: I think the traceback is still referencing the column (4) from `__marimo__cell_Hbol_.py`
```
    y = y / x
    ^
```

I think it should be 4 chars to the right (column 8) for the original script.
```
    y = y / x
        ^
```

dmadisetti: Oh I see. Increment lineno is actually pretty naive

```python
def increment_lineno(node, n=1):
    """
    Increment the line number of each node in the tree starting at *node* by *n*.
    This is useful to "move code" to a different location in a file.
    """
    for child in walk(node):
        if 'lineno' in child._attributes:
            child.lineno = getattr(child, 'lineno', 0) + n
    return node
```

See https://github.com/python/cpython/blob/97248204a1cc10bd3a66ce857535d133f6bc86a1/Lib/ast.py#L260
for python's latest implementation.

I think:

```python
def fix_source_position(node, n=1, t=4):
    for child in ast.walk(node):
        # TypeIgnore is a special case where lineno is not an attribute
        # but rather a field of the node itself.
        # Note, TypeIgnore does not have a "col_offset"
        if isinstance(child, TypeIgnore):
            child.lineno = getattr(child, 'lineno', 0) + n
            continue

        if 'lineno' in child._attributes:
            child.lineno = getattr(child, 'lineno', 0) + n

        if 'col_offset' in child._attributes:
            child.col_offset = getattr(child, 'col_offset', 0) + t

        if (
            "end_lineno" in child._attributes
            and (end_lineno := getattr(child, "end_lineno", 0)) is not None
        ):
            child.end_lineno = end_lineno + n
            
        if (
            "end_col_offset" in child._attributes
            and (end_col_offset := getattr(child, "end_col_offset", 0)) is not None
        ):
            child.end_col_offset = end_col_offset + n
    return node
```

marimo/_ast/compiler.py
```diff
@@ -91,17 +97,22 @@ def compile_cell(code: str, cell_id: CellId_t) -> CellImpl:
...
+        body_filename = get_filename(cell_id)
+        last_expr_filename = get_filename(cell_id, suffix="_output")
+        # cache the entire cell's code
+        cache(body_filename, code)
+        if sys.version_info >= (3, 9):
+            # ast.unparse only available >= 3.9
+            cache(
+                last_expr_filename,
+                ast.unparse(expr) if not isinstance(expr, str) else "None",
+            )
```
dmadisetti, line 115: So there's actually a current bug where `_output` doesn't display a line number because it's not put in cache:

image

https://marimo.app/l/64pjvh

 @akshayka is there a reason it has a different file? what happens if body_filename == last_expr_filename?

dmadisetti (commented):
Cool! Glad to see it works

marimo/_ast/compiler.py
ontowhee: Thanks for this code! I was able to add it in `compile.py`. However, I'm wondering if it should be more similar to [fix_missing_locations](https://github.com/python/cpython/blob/5505b91a684b0fc7ffcb3a5b325302671d74fb15/Lib/ast.py#L158), which uses the parent lineno if the child does not have a lineno attribute, rather than defaulting to 0?

marimo/_ast/compiler.py
ontowhee: Good idea about using a dataclass! I've updated it and it looks much cleaner now.

tests/_runtime/test_script.py
```diff
@@ -0,0 +1,30 @@
...
+    p = subprocess.run(
+        [
+            "python",
+            "tests/_runtime/script_data/script_exception_with_imported_function.py",
+        ],
+        capture_output=True,
+    )
+    result = p.stderr.decode()
+    assert "ZeroDivisionError: division by zero" in result
+    assert '/tests/_runtime/script_data/func.py", line 3' in result
```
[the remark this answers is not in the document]
ontowhee: Good catch! Updated!

marimo/_ast/compiler.py
ontowhee: This is interesting. I added another test and script, `tests/_runtime/script_data/script_exception_with_output.py`, and it isn't picking up the original source filename.

```
  File "/marimo/tests/_runtime/script_data/script_exception_with_output.py", line 15, in <module>
    app.run()
  File "/marimo/marimo/_ast/app.py", line 282, in run
    outputs, glbls = AppScriptRunner(InternalApp(self)).run()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/app/script_runner.py", line 82, in run
    outputs, defs = self._run_synchronous(
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/app/script_runner.py", line 111, in _run_synchronous
    output = execute_cell(cell, glbls, self.app.graph)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/executor.py", line 83, in execute_cell
    return EXECUTION_TYPES[execution_type].execute_cell(cell, glbls, graph)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/marimo/marimo/_runtime/executor.py", line 138, in execute_cell
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/marimo_223245/__marimo__cell_Hbol__output.py", line 2, in <module>
NameError: name 'y' is not defined
```

Looking at the traceback, this script runs through `eval(cell.last_expr, glbls)` whereas `tests/_runtime/script_data/script_exception.py` runs through `exec(cell.body, glbls)`. I think the `cell.last_expr.body` attributes need to be updated for the new lineno and col_offset as well to point to the original script. I looked at their values in the debugger and they seemed to correspond to generated files.

I still don't understand the compile_cell() block of code well, but I can come back to this to try a couple of things. It might take me some time to figure this out:
- See if last_expr.body needs to be updated
- See if the original script code can be cached along with the new filenames

marimo/_ast/compiler.py
ontowhee: Or does it default to 0 because of these are cells?

dmadisetti:
Thanks for your work @ontowhee ! I'm going to clean this up so we can get it in. I think it's mostly there

ontowhee:
@dmadisetti Sounds good! Thank you for your help!

dmadisetti:
Closing out in favor of https://github.com/marimo-team/marimo/pull/1893 (I just built on your commits)

PR #1803, opened 2024-07-15
Iterated on async implementation

dmadisetti:
I didn't want to spent too long on this, since I didn't think it was a priority.

Not super happy with the implementation- I think there's room for improvement- but it works!


mscolnick (approved):
this looks way better than what i had. i guess pty wasn't low-level enough? i trust that this is the right approach

PR #1821, opened 2024-07-17
feat: scratchpad

mscolnick:
This enables the scratchpad and cleans up some of the rough edges. 

- [x] There is still one blocking issue that the scratchpad does pollute the globals.

image

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -14,6 +14,7 @@ import sys
 import threading
 import time
 import traceback
+from copy import copy
 from multiprocessing import connection
 from typing import (
     TYPE_CHECKING,
@@ -1113,7 +1114,7 @@ class Kernel:
         runner = cell_runner.Runner(
             roots=roots,
             graph=graph,
-            glbls=self.globals,
+            glbls=copy(self.globals),
             excluded_cells=set(),
             debugger=self.debugger,
             execution_mode=self.reactive_execution_mode,
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -939,6 +939,38 @@ class Kernel:
                     break
             LOGGER.debug("Finished run.")
 
+    def _broadcast_missing_packages(self, runner: cell_runner.Runner) -> None:
+        if (
+            any(
+                isinstance(e, ModuleNotFoundError)
+                for e in runner.exceptions.values()
+            )
+            and self.package_manager is not None
+        ):
+            missing_packages = [
+                self.package_manager.module_to_package(mod)
+                for mod in self.module_registry.missing_modules()
+            ]
+
+            if missing_packages:
+                if self.package_manager.should_auto_install():
+                    self._execute_install_missing_packages_callback(
+                        self.package_manager.name
+                    )
+                else:
+                    MissingPackageAlert(
+                        packages=list(sorted(missing_packages)),
+                        isolated=is_python_isolated(),
+                    ).broadcast()
+
+    def _propagate_kernel_errors(
+        self,
+        runner: cell_runner.Runner,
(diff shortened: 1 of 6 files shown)


tests/_runtime/test_runtime.py
```diff
@@ -868,6 +869,82 @@ async def test_exception_scope_not_captured(
...
+        assert all(m[1]["cell_id"] == SCRATCH_CELL_ID for m in messages)
+        assert m1[1]["status"] == "queued"
+        assert m2[1]["status"] == "running"
+        assert m3[1]["status"] == "idle"
+        assert m4[1]["status"] is None
+        assert (
+            m4[1]["output"]["data"] == "<pre style='font-size: 12px'>1</pre>"
+        )
+        # Does not pollute globals
+        assert "x" not in k.globals
```
mscolnick, line 890: this currently fails

marimo/_runtime/runtime.py
```diff
@@ -1088,40 +1092,38 @@ async def run(
...
+
+        # Create graph of just the scratchpad cell
+        graph = dataflow.DirectedGraph()
+        graph.register_cell(SCRATCH_CELL_ID, cell)
+
         runner = cell_runner.Runner(
             roots=roots,
-            graph=self.graph,
+            graph=graph,
             glbls=self.globals,
```
akshayka, line 1116: @mscolnick — If you replace this with `copy.copy(self.globals)` (i.e. shallow copy), then the kernel globals won't get polluted. The scratch pad will still be able to mutate globals but that's fine, a deepcopy would be too expensive.

marimo/_runtime/runtime.py
```diff
@@ -1088,40 +1092,38 @@ async def run(
...
-            deletion_requests=[],
-        )
+
+        # If cannot compile, don't run
+        try:
+            cell = compile_cell(code, SCRATCH_CELL_ID)
+            cell.defs.clear()  # remove definitions
+            cell.refs.clear()  # remove definitions
+        except SyntaxError:
+            return
```
dmadisetti, line 1108: Shouldn't this broadcast broadcast error before return?
For dry, you can breakout the compilation and error handling from `_try_registering_cell` then just `CellOp.broadcast_error`

dmadisetti (commented):
This looks fun

marimo/_runtime/runtime.py
akshayka: I think it's preferable to be lax here, especially given that it's a scratchpad. The deepcopy will be too expensive in the common case of working with large dataframes, large arrays, etc, and it will also fail for non-pickleable objects.

dmadisetti: scratch won't be responsive, but that might be OK too. It could lead to other issues like an undefined ref (in the case a parent cell is compiled but hasn't run, like lazy) will throw a NameError, but strict mode would point out which cell has not run yet and prevent run.

An intermediate, but more verbose solution could be create `ScratchExecutor` that borrows some of the behavior from strict mode, but doesn't do the deep copying and can simplify the logic. More expensive than just a copy, but graph resolution for notebook sized graphs is still pretty cheap.

Then, for responsive runs, this function could register the cell here, only clear `defs`, and normally call `_run_cells(scratch)` (no need for another Runner). This would take maybe an override parameter

```python
Runner(...
   cell_execution_overrides : dict[Cell_Id: Execution_Type] = {SCRATCH : ScratchExecutor}
)
```

---

Anyway, just devils advocate, this is fine- and this comment can be revisited if scratch does cause reported issues

mscolnick: Yea we can revisit. I dont think we want it to be responsive. and its intended to break outside the marimo reactive runtime. if you do want that, then just create a cell in notebook.

Issue #1568, opened 2024-06-06
BlockingIO Error


andrewhill157:
### Describe the bug

I apologize in advance that this might be a bit annoying to reproduce.

When loading an app like the example given below, some fraction of the time it will fail to run and yield the following type of error (after which point the user has to refresh the app to get things going again):

```
Exception in callback <bound method Distributor._on_change of <marimo._utils.distributor.Distributor object at 0x7fbf4d755c60>>
handle: <Handle Distributor._on_change>
Traceback (most recent call last):
  File "uvloop/cbhandles.pyx", line 61, in uvloop.loop.Handle._run
  File "/ceph/home/andrew/temp/absel/env/lib/python3.10/site-packages/marimo/_utils/distributor.py", line 54, in _on_change
    response = self.input_connection.recv()
  File "/ceph/home/andrew/miniconda3/envs/test_env/lib/python3.10/multiprocessing/connection.py", line 255, in recv
    buf = self._recv_bytes()
  File "/ceph/home/andrew/miniconda3/envs/test_env/lib/python3.10/multiprocessing/connection.py", line 426, in _recv_bytes
    return self._recv(size)
  File "/ceph/home/andrew/miniconda3/envs/test_env/lib/python3.10/multiprocessing/connection.py", line 384, in _recv
    chunk = read(handle, remaining)
BlockingIOError: [Errno 11] Resource temporarily unavailable
Exception in callback <bound method Distributor._on_change of <marimo._utils.distributor.Distributor object at 0x7fbf4d755c60>>

handle: <Handle Distributor._on_change>
Traceback (most recent call last):
  File "uvloop/cbhandles.pyx", line 61, in uvloop.loop.Handle._run
  File "/ceph/home/andrew/temp/absel/env/lib/python3.10/site-packages/marimo/_utils/distributor.py", line 54, in _on_change
    response = self.input_connection.recv()
  File "/ceph/home/andrew/miniconda3/envs/test_env/lib/python3.10/multiprocessing/connection.py", line 256, in recv
    return _ForkingPickler.loads(buf.getbuffer())
_pickle.UnpicklingError: invalid load key, '\x07'.
Exception in callback <bound method Distributor._on_change of <marimo._utils.distributor.Distributor object at 0x7fbf4d755c60>>
handle: <Handle Distributor._on_change>
Traceback (most recent call last):
  File "uvloop/cbhandles.pyx", line 61, in uvloop.loop.Handle._run
[...]

PR #1822, opened 2024-07-18
fix: make recv robust to EAGAIN, EDEADLK

akshayka:
Socket reads in the server sporadically fail. If a read fails with a BlockingIOError, retry it after a short wait (if the connection still polls() true).

This is an attempt to mitigate the issue discussed in #1568.

### DIFF
--- a/marimo/_utils/distributor.py
+++ b/marimo/_utils/distributor.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+import time
 from typing import TYPE_CHECKING, Callable, Generic, TypeVar
 
 from marimo import _loggers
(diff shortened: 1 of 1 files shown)


marimo/_utils/distributor.py
```diff
@@ -49,9 +50,17 @@ def _remove() -> None:
...
+        retry_sleep_seconds = 0.001
         while self.input_connection.poll():
             try:
                 response = self.input_connection.recv()
+            except BlockingIOError as e:
+                # recv() sporadically fails with EAGAIN, EDEADLK ...
+                LOGGER.warning(
+                    "BlockingIOError in distributor receive: %s", str(e)
+                )
+                time.sleep(retry_sleep_seconds)
```
dmadisetti, line 62: Aren't hard sleeps blocking? Any reason not to use asyncio?

akshayka: They are blocking, but `asyncio.add_reader()` doesn't accept an async callback. So we can't use `asyncio.sleep()` here.

dmadisetti: It might even be possible to remove the sleep? 1000ns is just an order magnitude larger than a function call, and might spend that on Logging

You can still invoke asyncio.run(asyncio sleep(time)) from a sync call

akshayka: > You can still invoke asyncio.run(asyncio sleep(time)) from a sync call

No that's not possible here; an event loop is already running. That would also be blocking, even if it were possible :)

dmadisetti (approved):

PR #1828, opened 2024-07-19
fix: dont autoinstall pkgs we already tried installing

### LINKED ISSUE #1736: keep installing packages already installed
### Describe the bug

When running cell, mo poped up an installation window suggesting package notinstalled.
After installation (within the editor, or external), the window keeps show up.
image
Installing package as suggested:
image
the window shows again.
image

### Environment

ubuntu 22.04. Running in a conda virtualenv.

### Code to reproduce

# Import necessary libraries for the project 
import torch 
import torch.nn as nn 
import torch.nn.functional as F 
import numpy as np 
import random 
import gc 

# Import PyTorch Geometric libraries 
import torch_geometric.transforms as T 
import torch_geometric.utils
from torch_geometric.datasets import Planetoid
from torch_geometric.utils import spmm

akshayka:
Fixes an issue that is related to #1736 (infinite loop when installing a package that doesn't have an importable module — found this when importing an empty package).  I am unsure if this fixes the problem in #1736, since that involved packages that should be importable.

### DIFF
--- a/marimo/_runtime/packages/conda_package_manager.py
+++ b/marimo/_runtime/packages/conda_package_manager.py
@@ -17,5 +17,5 @@ class CondaPackageManager(CanonicalizingPackageManager):
 class PixiPackageManager(CondaPackageManager):
     name = "pixi"
 
-    async def install(self, package: str) -> bool:
+    async def _install(self, package: str) -> bool:
         return self.run(["pixi", "add", package])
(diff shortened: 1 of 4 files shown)


mscolnick (approved):
I think you mentioned the same ticket twice in the readme

akshayka:
> I think you mentioned the same ticket twice in the readme

Yea that was intentional. Updated the PR description to be more clear.

PR #1829, opened 2024-07-19
improvement: strict mode NameError fix + cleaner exceptions

dmadisetti:
## 📝 Summary

Previous Exceptions:

image

vs

image

Also clean up the name error for strict execution.

@akshayka OR @mscolnick


mscolnick (dismissed):
awesome 😎

marimo/_runtime/executor.py
```diff
@@ -114,28 +162,63 @@ async def execute_cell_async(
...
+                )
+                # fmt: on
+            else:
+                # fmt: off
+                return (
+                    marimo_output_evaluation
+                    (cell.last_expr, glbls)
+                )
+                # fmt: on
+        except Exception as e:
```
akshayka, line 195: This should be `except BaseException as e:`. Some libraries throw `BaseExceptions` (polars is one), and we need to communicate that exception info back to the user instead of hiding it in an "Unknown Error".

marimo/_runtime/executor.py
```diff
@@ -114,28 +162,63 @@ async def execute_cell_async(
...
+                    (cell.last_expr, glbls)
+                )
+                # fmt: on
+            else:
+                # fmt: off
+                return (
+                    marimo_output_evaluation
+                    (cell.last_expr, glbls)
+                )
+                # fmt: on
```
akshayka, line 194: Instead of formatting our source code to look good in a traceback, would it be possible to just modify the traceback in `cell_runner` to hide marimo internals? This seems difficult to maintain.

marimo/_runtime/runner/cell_runner.py
```diff
@@ -338,36 +346,80 @@ async def run(self, cell_id: CellId_t) -> RunResult:
...
                 self.missing_packages = True
 
             self.cancel(cell_id)
-            run_result = RunResult(output=None, exception=e)
+            run_result = RunResult(output=None, exception=unwrapped_exception)
             tmpio = io.StringIO()
-            traceback.print_exc(file=tmpio)
+            # The executors explicitly raise cell exceptions from base
+            # exceptions such that the stack trace is cleaner.
+            traceback.print_exception(unwrapped_exception, file=tmpio)
```
akshayka, line 415: This is failing with

```
  File "/Users/akshay/src/marimo-team/marimo/marimo/_runtime/runner/cell_runner.py", line 415, in run
    traceback.print_exception(unwrapped_exception, file=tmpio)
TypeError: print_exception() missing 2 required positional arguments: 'value' and 'tb'
```

marimo/_runtime/executor.py
akshayka: Taking a step back, related to my comment above — instead of handling a special `MarimoBaseException`, would it be possible to just rewrite the traceback and hide marimo internals?

akshayka: Could just modify the traceback by throwing out all frames preceding the marimo cell's execution (ie lines before `File  ...__marimo__cell...`)

marimo/_runtime/executor.py
dmadisetti: I think modifying the stack trace, especially with different traces between script mode and notebook mode might introduce more code debt than this.

However, I do agree, that this is a little gimmicky (esp, since lint has to be manually disabled) and not needed.

The current trace of
```python
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Is fine, and I think it'd be better to revert to this than introduce stack trace manipulation.
An intermediate is without the formatting

```python
return marimo_output_evaluation(cell.last_expr, glbls)
```

But at that point, the `eval` makes more sense. LMK what you think

akshayka: Okay, makes sense. Reverting to `eval`/`exec` without special formatting sounds good to me.

dmadisetti:
Also added the name error catching for normal mode:

image

With other regular NameErrors passing through:

image

image

MarimoRef NameErrors in Strict Mode are a bit more verbose, since they should not occur:

image


with a potentially detected NameError preventing execution:

image


& edgecase for self definition

image

marimo/_runtime/executor.py
dmadisetti: Good catch! This is the source of the failing test. Fixed. Re-stack trace manipulation, see other comment.

I think that `MarimoBaseException` is probably not correctly named. It's really the `MarimoExecutionException` or `CellRuntimeException`, and provides the context that it was a evaluation cell that failed. Being able to unwrap the exception for a cleaner stack trace is definitely a plus, but not the only positive since it's an indication the failure came from a cell failure.

LMK if this context changes your thoughts/ if you have thoughts for a better name

tests/_ast/test_app.py
```diff
@@ -235,10 +236,12 @@ def _() -> None:
             _x  # type: ignore  # noqa: F821
             return
 
-        with pytest.raises(NameError) as e:
+        with pytest.raises(MarimoBaseException) as e:
             app.run()
```
akshayka, line 240: I think `app.run()` should still raise the true exception, not a wrapped one.

akshayka (commented):
Few more comments. I do think that wrapping exceptions in `MarimoBaseException` complicates things.

I've seen approaches for directly modifying tracebacks, e.g. discussed here: https://discuss.python.org/t/dropping-frames-from-a-traceback/10385. I know I asked before, but couldn't it be feasible to just omit frames that precede the first file with name `*__marimo__cell_*` in `CellRunner`?  I guess I don't see how that introduces tech debt.

marimo/_runtime/executor.py
```diff
@@ -103,39 +150,54 @@ async def execute_cell_async(
...
+            if _is_coroutine(cell.last_expr):
+                return await eval(cell.last_expr, glbls)
+            else:
+                return eval(cell.last_expr, glbls)
+        except NameError as e:
+            raise_name_error(graph, e)
+        except (BaseException, Exception) as e:
+            # Raising from a BaseException will fold in the stactrace prior
+            # to execution
+            raise MarimoBaseException from e
```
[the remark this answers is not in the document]
dmadisetti: The discussion thread posted doesn't make a great case for manipulating the stacktrace as a practice- but the examples posted are pertinent

The **Jinja** example the trace is changed to make it look as if it comes from the template file.

In **IPython/ jupyter** the raw trace is not carried but the relevant stack info and formatted into their own stacktrace format (also useful for different kernel types):

image

This is also needed a bit more since Jupyter can point to code that no longers exists:

image

Stack traces from files can also give a bit more context:

image

I believe the relevant code is the aptly named [`ultratb`](https://github.com/ipython/ipython/blob/main/IPython/core/ultratb.py).

---

ideally, I think marimo still wants to provide python like stack traces since it's actually python, but maybe it can have a more managed approach? I see this as a follow up PR done in FE, but how's this look for a goal with links? Or is the ultratrace closer to what we'd want? Is the formatting consistent in script mode?

image

If the trace is going to be totally manipulated front-end like jupyter, I do think the complexity +  this is a compelling reason not to use MarimoBaseException/ MarimoRuntimeException

I'm good to revert this, but just on a thought regarding this comment:

The error handling could be moved to run

```python
def run():
    try:
       # ...
    except MarimoRuntimeError as e:
        raise e.__cause__
    except Exception as e:
        # Known for certain to come from marimo itself
        raise MarimoScriptException("Congrats! You definitely found a marimo bug. Please report") from e
```

Which is still useful if an app is used from another script, and raises the raw error unless something is actually wrong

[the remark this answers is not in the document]
akshayka: @dmadisetti , thanks for sharing the context and the mock-up for the future design. The screenshot LGTM; I think  the user doesn't need to see the first frame (`executor.py`), but it's okay if they do.

`run()` logic with error handling looks good. 

No need to revert! If the error handling for running as as script is updated as per your example, that's good with me. I just didn't want to push more complexity onto the user.

tests/_ast/test_app.py
dmadisetti: Cool! I think I'm going to rename it MarimoRuntimeException though, MarimoBaseException is a little meaningless

akshayka: @dmadisetti , need any help getting this over the finish line?

dmadisetti: No, sorry just been busy this week. Just sat down to get this out

dmadisetti:
Not the most dramatic change, but I think it will pair very nicely with #1799

akshayka (approved):
This looks great!! I like the distinction between runtime errors and marimo internal errors you've introduced.

I noticed one test is failing. Do you think that's unrelated? I would guess so ...

akshayka:
Reran and it's passing. Maybe just flaky.

PR #1833, opened 2024-07-19
feat: better traceback, help with google or AI

mscolnick:
Improve traceback readability on the FE. Also add help with Google or AI

## After
Screenshot 2024-07-19 at 4 18 09 PM
Screenshot 2024-07-19 at 4 17 51 PM

## Before
Screenshot 2024-07-19 at 2 32 21 PM


akshayka (commented):
I like this! It's more legible, and search on google/fix with AI are nice touches 👌

Only suggestion -- let's have the traceback expanded by default? It looks better collapsed, but it's more useful expanded, since when you hit an error in real use cases typically you need to see the traceback. The option to hide is nice though.

PR #1849, opened 2024-07-23
ci: check for timed out tests

dmadisetti:
Creating PR to test for slow test


akshayka:
Clever debugging trick; didn't know it was possible to dump the trace like you did.

Looks like this line is hanging?

```python
                # Assert kiosk session received the sync cell_ids
>               data = other_websocket.receive_json()
```

akshayka:
@dmadisetti we're skipping the offending test for now, so I think we can close this PR; low priority to actually fix the offending test. Thanks for debugging!

PR #1868, opened 2024-07-24
Set status to idle after formatters run, not before

akshayka:
This PR modifies the cell state machine to count the cell as "running" when output formatters are running.

Sometimes formatters can take a very long time — today, this happens when outputting a very large list or dict. Before this change, formatters that took a long time would lock the kernel without any UI indication in the frontend, making the notebook appear to be broken. 

This has the side-effect of making a formatter interruptible, which is helpful if the formatter is taking a very long time to run.

Re: lists/dicts formatting, we should probably lists/dicts/tuples when outputting very large ones, to avoid locking the kernel, but can handle that in another PR.

Draft because haven't tested thoroughly.

### DIFF
--- a/tests/_runtime/test_runtime.py
+++ b/tests/_runtime/test_runtime.py
@@ -881,7 +881,7 @@ class TestExecution:
         assert all(m[1]["cell_id"] == SCRATCH_CELL_ID for m in messages)
         assert m1[1]["status"] == "queued"
         assert m2[1]["status"] == "running"
-        assert m3[1]["status"] == None
+        assert m3[1]["status"] is None
         assert (
             m3[1]["output"]["data"] == "<pre style='font-size: 12px'>1</pre>"
         )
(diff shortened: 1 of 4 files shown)


tests/_runtime/test_runtime.py
```diff
@@ -881,11 +881,11 @@ async def test_run_scratch(mocked_kernel: MockedKernel) -> None:
...
         assert m1[1]["status"] == "queued"
         assert m2[1]["status"] == "running"
-        assert m3[1]["status"] == "idle"
-        assert m4[1]["status"] is None
+        assert m3[1]["status"] is None
         assert (
-            m4[1]["output"]["data"] == "<pre style='font-size: 12px'>1</pre>"
+            m3[1]["output"]["data"] == "<pre style='font-size: 12px'>1</pre>"
         )
+        assert m4[1]["status"] == "idle"
```
mscolnick, line 888: oh this fixes also what i thought was a bug which is "why are there more idle messages, if its already idle". its clearer/cleaner that "running" is repeated

mscolnick (approved):

PR #1881, opened 2024-07-25
Fix #775: Ability to style / format mo.ui.table()

### LINKED ISSUE #775: Ability to style / format mo.ui.table()
### Description

Would be great to be able to format the output of mo.ui.table(). 
As an example to be able to add decimals, format numbers as percentages, datetime formating etc... 
image
This is the current output. 

### Suggested solution

A suggested solution would be to add the ability to handle pd options display like pd.options.display.float_format = "{:,.2f}".format.

### Alternative

_No response_

### Additional context

_No response_

metaboulie:
## 📝 Summary

Fix #775: Ability to style / format mo.ui.table()

## 🔍 Description of Changes

1. Table formatting, via GUI based / application mode
Screenshot 2024-07-25 at 20 24 38
  
  - Different Format options based on the datatype of the column
  - Able to clear the format
  - Able to handle `null`, `undefined` and `""`

2. Table formatting, programmatically via python code
  - add a `format_mapping` arg to create a mapping between column names and formats
  - `download_as` will download raw data (no formatting)
  - handle all types of `TableData` well
  - tests for each table type are written

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


marimo/_plugins/ui/_impl/table.py
```diff
@@ -181,6 +182,7 @@ def __init__(
         selection: Optional[Literal["single", "multi"]] = "multi",
         page_size: int = 10,
         show_column_summaries: bool = True,
+        format_mapping: Optional[Dict[str, str | Callable]] = None,
```
mscolnick, line 185: this is great. i like the API, do we need to type `Callable`? `Callable[Any, str]`

marimo/_plugins/ui/_impl/table.py
```diff
@@ -223,19 +226,26 @@ def __init__(
...
+        # field_types = self._manager.get_field_types()
+        field_types = self._formatted_manager.get_field_types()
 
         super().__init__(
             component_name=table._name,
             label=label,
             initial_value=[],
             args={
-                "data": self._filtered_manager.to_data(),
+                "data": self._formatted_manager.to_data(),
```
mscolnick, line 248: I think we can pass the `format_mapping` directly into the table `self._filtered_manager.to_data(format_mapping),`.

And then each `TableManager` implementation can handle the formatting themselves, which should reduce the size of `_apply_formatting`

marimo/_plugins/ui/_impl/table.py
```diff
@@ -346,3 +360,87 @@ def search(self, args: SearchTableArgs) -> Union[JSONType, str]:
...
+                    )
+                    # Replace the column in the copy of the table
+                    data_copy = data_copy.set_column(
+                        col_index, col, transformed_column
+                    )
+            return data_copy
+
+        return data
+
+    def _format_row(self, row: Dict[str, JSONType]) -> Dict[str, JSONType]:
```
mscolnick, line 426: can we bring these utils a file `/marimo/_plugins/ui/_impl/tables/format.py`.

it should be much easier to unit test without having to go through `mo.ui.table`.

marimo/_plugins/ui/_impl/table.py
```diff
@@ -346,3 +360,87 @@ def search(self, args: SearchTableArgs) -> Union[JSONType, str]:
         # Save the manager to be used for selection
         self._filtered_manager = result
         return result.limit(TableManager.DEFAULT_ROW_LIMIT).to_data()
+
+    def _apply_formatting(self, data: TableData) -> TableData:
+        # Apply formatting to the given data based on the format mapping.
+        import pandas as pd
+        import polars as pl
+        import pyarrow as pa
```
mscolnick, line 368: the user may not have all these installed so this would break otherwise. we can move the logic for each dataframe time into their respective `TableManager`

marimo/_plugins/ui/_impl/table.py
metaboulie: like `format_mapping: Optional[Dict[str, str | Callable[Any, Any]]] = None,`?

marimo/_plugins/ui/_impl/table.py
metaboulie: Can we do something like `self._filtered_manager.apply_formatting(self.format_mapping).to_data()` here, and implement `apply_formatting` for each `TableManager` instead?

One benefit of this implementation is in `_convert_value`, where we can also return formatted data:

```python
def _convert_value(
    self, value: list[str]
) -> Union[List[JSONType], "pd.DataFrame", "pl.DataFrame"]:
    indices = [int(v) for v in value]
    self._selected_manager = self._filtered_manager.select_rows(indices)
    self._has_any_selection = len(indices) > 0
    return self._selected_manager.apply_formatting(self.format_mapping).data  # type: ignore[no-any-return]
```
Original `_apply_formatting`, `_format_row`, … can be removed.

mscolnick: I actually don't think we want the formatted value in the conversion / result. we really only want it for displaying purposes

marimo/_plugins/ui/_impl/table.py
mscolnick: yetp that looks good

mscolnick:
wow @metaboulie , this code looks really great! API and implementation 👌 

I think since things look good - the last thing would be Python unit tests for the formatting `to_csv`.
they can go in the files:
`tests/_plugins/ui/_impl/tables/test_pandas_table.py`
`tests/_plugins/ui/_impl/tables/test_*_table.py

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -186,6 +186,7 @@ export const LoadingDataTableComponent = memo(
   <T extends {}>(
     props: Omit<DataTableProps<T>, "sorting"> & { data: TableData<T> },
   ) => {
+    // sourcery skip: use-object-destructuring
```
mscolnick, line 189: nit, i thikn you can remove this

marimo/_plugins/ui/_impl/tables/pandas_table.py
```diff
@@ -25,15 +29,33 @@ def create() -> type[TableManager[Any]]:
         class PandasTableManager(TableManager[pd.DataFrame]):
             type = "pandas"
 
-            def to_csv(self) -> bytes:
+            def to_csv(
+                self, format_mapping: Optional[FormatMapping] = None
+            ) -> bytes:
                 has_headers = len(self.get_row_headers()) > 0
-                return self.data.to_csv(
+                _data = self.data.copy()
```
mscolnick, line 36: this might be an unnecessary copy, you could just do `_data = self.data`

marimo/_plugins/ui/_impl/tables/polars_table.py
```diff
@@ -29,13 +33,19 @@ def create() -> type[TableManager[Any]]:
         class PolarsTableManager(TableManager[pl.DataFrame]):
             type = "polars"
 
-            def to_csv(self) -> bytes:
+            def to_csv(
+                self,
+                format_mapping: Optional[FormatMapping] = None,
+            ) -> bytes:
+                _data = self.data.clone()
```
mscolnick, line 40: same, might be unnecessary. polars df are immutable

marimo/_plugins/ui/_impl/tables/pandas_table.py
mscolnick: you could always write a test too to make sure the incoming df does not get modified

tests/_plugins/ui/_impl/tables/test_polars_table.py
```diff
@@ -306,3 +306,268 @@ def test_search(self) -> None:
...
+                "E": ["2021-01-01", "2021-01-02", "2021-01-03"],
+                "F": [
+                    [1, 2, 3],
+                    [4, 5, 6],
+                    [7, 8, 9],
+                ],  # No formatting applied
+                "G": [2.23606797749979, 5.0, 7.810249675906654],
+            }
+        )
+        assert_frame_equal(formatted_data, expected_data)
```
mscolnick, line 573: these tests are great!

tests/_plugins/ui/_impl/tables/test_table_utils.py
```diff
@@ -39,3 +44,112 @@ def test_is_dataframe_like():
...
+    format_mapping = {"col1": "{:.2f}"}
+    values = [123.456, None]
+    expected = ["123.46", None]
+    assert format_column("col1", values, format_mapping) == expected
+
+    # Test with empty list
+    format_mapping = {"col1": "{:.2f}"}
+    values = []
+    expected = []
+    assert format_column("col1", values, format_mapping) == expected
```
mscolnick, line 155: :chef-kiss:

mscolnick:
couple lines to delete, but otherwise looks great!

frontend/src/plugins/impl/DataTablePlugin.tsx
metaboulie: forget to remove this

mscolnick (approved):

mscolnick:
🚢

PR #1892, opened 2024-07-26
improvement: cleanup markdown, use prose

mscolnick:
Cleanup markdown to allow for dynamic/responsive sizing across marging/padding/font-size/line-height by using `tailwind typography`.

Now we can easily add `size` to `mo.md()`.

In future we can size the whole document (by app config or user config)

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -68,6 +68,7 @@
     "@react-aria/focus": "^3.17.1",
     "@replit/codemirror-vim": "^6.2.1",
     "@storybook/addon-styling": "^1.3.7",
+    "@tailwindcss/typography": "^0.5.13",
     "@tanstack/react-table": "^8.19.2",
     "@textea/json-viewer": "^3.4.1",
     "@types/humanize-duration": "^3.27.4",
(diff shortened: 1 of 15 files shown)


akshayka (commented):
On the whole, looks good! Some comments.

I like the responsiveness that this provides. But can we hold off adding `size` to the public API? I'd like to experiment more and see whether we can get away with only user config or app config.

A couple of things I noticed that would be good to fix before merging:

**JSON Output size.**

JSON Output embedded in markdown is now unnaturally big. It should have the same formatting as code, since it's monospace. I tried fixing this myself, but wasn't sure how to do it cleanly through `tailwind.config.js` alone and spent too much time trying to figure it out. Would appreciate help.

This is what I would like it to look like (how it is today, on main):

image

But in this PR, `marimo-json-output` is unnaturally big. If `marimo-json-output` could pick up the same styles that tailwind puts on `.prose pre`, that would be okay. Here is what it looks like in this PR:

image


**Bullet legibility.**

The bullet color is very light now, and hard to see. I'm all for minimalist, unobtrusive design, but this is too minimalist even for me. Can we darken it?

akshayka (approved):

PR #1897, opened 2024-07-26
fix: marimo edit with absolute path to directory

### LINKED ISSUE #1895: marimo edit C:\Users\user\Desktop\python\pkg\examples finds notebooks, but can't open them
### Describe the bug

I ran `marimo edit C:\Users\user\Desktop\python\pkg\examples` with a fully qualified path. It opened the page with running notebooks, recent notebooks, workspace. It lists all the notebooks in `examples`. But when I click on one I get a error screenshotted below. When I use a relative path, starting in `C:\Users\user\Desktop\python\pkg\` and running `marimo edit examples` then it works expected.

Error:
image

### Environment

{
  "marimo": "0.7.12",
  "OS": "Windows",
  "OS Version": "10",
  "Processor": "Intel64 Family 6 Model 85 Stepping 4, GenuineIntel",
  "Python Version": "3.10.11",
  "Binaries": {
    "Browser": "126.0.6478.127",
    "Node": "--"
  },
  "Requirements": {
    "click": "8.1.7",
    "jedi": "0.19.0",
    "markdown": "3.6",
    "pygments": "2.16.1",
    "tomlkit": "0.12.5",
    "uvicorn": "0.30.1",
    "starlette": "0.37.2",
    "websockets": "12.0",
    "typing-extensions": "4.7.1",
    "ruff": "0.5.5"
  }
}

### Code to reproduce

_No response_

akshayka:
Fixes a bug in which marimo edit /abs/path/ would fail to open notebooks due to incorrect slicing.

Fixes a bug in which marimo edit /abs/path/ or /abs/path would fail when run in a directory other than /abs/path

Fixes #1895.

### DIFF
--- a/frontend/src/components/pages/home-page.tsx
+++ b/frontend/src/components/pages/home-page.tsx
@@ -62,6 +62,7 @@ import {
 } from "../home/state";
 import { Maps } from "@/utils/maps";
 import { Input } from "../ui/input";
+import { Paths } from "@/utils/paths";
 
 function tabTarget(path: string) {
   // Consistent tab target so we open in the same tab when clicking on the same notebook
@@ -262,6 +263,7 @@ const Node = ({ node, style }: NodeRendererProps<FileInfo>) => {
 
   const Icon = FILE_TYPE_ICONS[fileType];
   const iconEl = <Icon className="w-5 h-5 flex-shrink-0" strokeWidth={1.5} />;
+  const root = useContext(RunningNotebooksContext).root;
 
   const renderItem = () => {
     const itemClassName =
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/home-page.tsx
```diff
@@ -277,25 +275,21 @@ const Node = ({ node, style }: NodeRendererProps<FileInfo>) => {
       );
     }
 
-    const relativePath =
-      node.data.path.startsWith(root) && Paths.isAbsolute(node.data.path)
-        ? node.data.path.slice(root.length + 1)
```
akshayka, line 282: Why was this slicing added? Is it okay to remove?

mscolnick: This may just be masking a real issue. Relative paths should be ok to pass - this may happen from the sidebar as well. 

Why is passing a relative path returning a bad file?

mscolnick: I think the real fix is on the backend that the relative path should use the directory from marimo edit, not the cwd

akshayka: Yea makes some sense.

Two reasons it returns a bad file.

1 is that the slicing logic can be off-by-one (which we can fix).

Other reason is that `AppFileRouter` doesn't use the directory, which I guess we can fix (can check if there is a directory?): https://github.com/marimo-team/marimo/blob/aa19fe6212e2bb8039e81b021ae26024d572e5df/marimo/_server/file_router.py#L73-L99

mscolnick: Yea that's right. It's probably not wired down - I'm happy to do this Monday (I initially added the 'marimo edit dir' functionally). 

Also it's nice having this on the backend since it's much easier to test this

akshayka: Okay sounds good, I'll hold off. Thanks!

akshayka: Had spare time, so went ahead and did the change. Feel free to take over or revert if you had something else in mind.

mscolnick (commented):
Looks good - probably worth adding tests for Paths and the Python code

akshayka:
@mscolnick added unit tests

mscolnick (approved):

PR #1898, opened 2024-07-27
fix: cell spacing in run, html export

akshayka:
The vertical layout was missing spacing when code was shown.

### DIFF
--- a/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
+++ b/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
@@ -76,35 +76,37 @@ const VerticalLayoutRenderer: React.FC<VerticalLayoutProps> = ({
 
   return (
     <VerticalLayoutWrapper invisible={invisible} appConfig={appConfig}>
-      {cells.map((cell) => (
-        <VerticalCell
-          key={cell.id}
-          cellId={cell.id}
-          output={cell.output}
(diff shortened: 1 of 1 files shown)


frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
```diff
@@ -76,35 +76,37 @@ const VerticalLayoutRenderer: React.FC<VerticalLayoutProps> = ({
...
-      {mode === "read" && (
-        <ActionButtons
-          canShowCode={canShowCode}
-          showCode={showCode}
-          onToggleShowCode={() => setShowCode((v) => !v)}
-        />
-      )}
-      {/* Shown in read and present */}
-      <FloatingOutline />
+      <div className={showCode ? "flex flex-col gap-5" : undefined}>
```
mscolnick, line 79: Can you just pass this class to the vertical layout? Avoid a bunch of nesting

akshayka: That doesn't quite work because vertical layout has its own nesting. Plus I realized we don't want to put everything (TOC, action buttons) in the flex, just the cell array.

Updated to only wrap in div when needed. So now the DOM is analogous to the `CellArray` component, which also wraps cells in a flexed div under a `VerticalLayoutWrapper`.

mscolnick (approved):

PR #1904, opened 2024-07-28
feat: add linenumber links in tracebacks

dmadisetti:
## 📝 Summary

I think this gets tracebacks in a good place over all. Looks awesome with the front end changes @mscolnick earlier
I looked at doing this with pygments, but it looked like frontend was going to have to hook in anyway, so I think a pure FE solution is best. FE is not my strength, so feedback welcome

image

I played around with a few filename replacements, but settled on this since `marimo://` hints that it's a link, the filename makes it clear that it's the notebook being worked on, and it seemed a bit more balanced against other trace errors than just the cell name

TODO:
 - [x] Add a test or two

## 🔍 Description of Changes

- made renderHTML more modular
- added goToCellLine
- Replaced traceback tokens with relevant cell link

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/editor/links/cell-link.tsx
+++ b/frontend/src/components/editor/links/cell-link.tsx
@@ -4,6 +4,8 @@ import { Logger } from "../../../utils/Logger";
 import { cn } from "@/utils/cn";
 import { displayCellName } from "@/core/cells/names";
 import { useCellActions, useCellIds, useCellNames } from "@/core/cells/cells";
+import { useFilename } from "@/core/saving/filename";
+import { goToCellLine } from "@/core/codemirror/go-to-definition/utils";
 
 interface Props {
   cellId: CellId;
@@ -11,15 +13,18 @@ interface Props {
   shouldScroll?: boolean;
   skipScroll?: boolean;
   onClick?: () => void;
+  formatCellName?: (name: string) => string;
   variant?: "destructive" | "focus";
 }
 
 /* Component that adds a link to a cell, with styling. */
 export const CellLink = (props: Props): JSX.Element => {
-  const { className, cellId, variant, onClick, skipScroll } = props;
+  const { className, cellId, variant, onClick, formatCellName, skipScroll } =
+    props;
   const cellName = useCellNames()[cellId] ?? "";
   const cellIndex = useCellIds().inOrderIds.indexOf(cellId);
   const { showCellIfHidden } = useCellActions();
+  const formatName = formatCellName ?? ((name: string) => name);
 
   return (
     <div
@@ -39,7 +44,7 @@ export const CellLink = (props: Props): JSX.Element => {
         });
       }}
     >
-      {displayCellName(cellName, cellIndex)}
+      {formatName(displayCellName(cellName, cellIndex))}
     </div>
   );
 };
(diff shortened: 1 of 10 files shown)


frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -94,3 +102,85 @@ function lastLine(text: string): string {
...
+const elementContainsMarimoCellFile = (domNode: Element) => {
+  return (
+    domNode &&
+    matchesSelector(domNode, "span.nb") &&
+    domNode.firstChild instanceof Text &&
+    domNode.firstChild.nodeValue?.includes("__marimo__")
+  );
+};
+
+const replaceTracebackFilenames = (domNode: DOMNode) => {
```
mscolnick, line 124: if not too much trouble, i think we should add a unit test for this function

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -94,3 +102,85 @@ function lastLine(text: string): string {
   const lines = el.textContent?.split("\n").filter(Boolean);
   return lines?.at(-1) || "";
 }
+
+const matchesSelector = (domNode: Element, selector: string) => {
+  const [tagName, ...classes] = selector.split(".");
+  const classList = (domNode.attribs.class || "").split(" ");
```
mscolnick, line 108: nit: i think this can be `const classList = [...domNode.classList];`

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -94,3 +102,85 @@ function lastLine(text: string): string {
...
+              />
+            </span>
+          );
+        }
+      }
+    }
+  }
+};
+
+const replaceTracebackPrefix = (domNode: DOMNode) => {
```
mscolnick, line 177: and maybe a test for this one too

mscolnick (dismissed):
this is sick! thanks for adding this. such a qol improvement

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
mscolnick: (also happy to do any of these unit tests if you want)

frontend/src/components/editor/links/cell-link.tsx
```diff
@@ -4,22 +4,27 @@ import { Logger } from "../../../utils/Logger";
 import { cn } from "@/utils/cn";
 import { displayCellName } from "@/core/cells/names";
 import { useCellActions, useCellIds, useCellNames } from "@/core/cells/cells";
+import { useFilename } from "@/core/saving/filename";
```
mscolnick, line 7: this might be the reason for the failing tests. 

in `src/core/saving/filename.ts:6:42` we could add `if (import.meta.vitest)` and not check the dom

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
dmadisetti: left a comment because this tripped me up too, `.classList` isn't available since it's a [@domhandler object](https://github.com/fb55/domhandler/blob/1b2db0f3de85bc3285c7d0cc3814a78fbca1dcd1/src/node.ts#L274)

frontend/src/components/editor/links/cell-link.tsx
dmadisetti: done! import.meta.env.MODE == "test" seemed the most reliable
but the assert also seems a bit strong in this context?

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
dmadisetti: Done!

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
mscolnick: oh, thank you

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
dmadisetti: Added some tests, but maybe not as targeted as they should be, feel free to add more if some edges cases might be lost

frontend/src/components/editor/links/cell-link.tsx
mscolnick: Yes that is better - there is no reliable fallback. I'd rather fail hard so it's obvious than silently and let things sneak through

mscolnick (approved):
looks great!

PR #1955, opened 2024-08-05
feat: import cells

akshayka:
This PR modifies the runtime's calculation of descendants of a cell to exclude edges that are based on modules that have already been imported into the notebook.

**Example:**

A user executes the cell:

```python
import marimo as mo
```

then modifies this cell  by adding another import and and reexecuting it:

```python
import marimo as mo
```

With this change, after the second run, only cells referencing `plt` will run; cells referencing `mo` will not be re-run (unless they also reference `plt`).

This lets users incrementally add imports to a single cell, as they are used to doing in traditional notebooks, without triggering unnecessary computation.

**Implementation:**

- `CellImpl` has a new field, `import_workspace`, that tracks whether a cell is an "import cell" — whether it only includes import statements. This field is also used as workspace by the runtime to keep track of the definitions that this module has imported.
- After an import cell is executed, a post execution hook populates its imported defs in its workspace.
- When an import cell is replaced with another import cell, its imported defs are carried over to the new cell provided that the defs originated from matching import statements.
- The transitive closure computation is updated to take a function that computes the children of a cell
- The runtime passes a function to transitive closure that prunes imported defs when appropriate
- The module watcher is updated to prune imported defs whenever a cell is made stale due to a module change; only defs corresponding to modified modules will be pruned
- When cells are invalidated by the runtime, their imported defs are cleared too.

**TODO:**
- [x] Handle edge case when an import cell raises an exception:

A:
```python
import random
import raise_an_import_error
```

B:
```python
random.randint(0, 10000)
```

Comment out the second import, re-run A — B should run.


tests/_runtime/test_runtime.py
```diff
@@ -1214,32 +1216,22 @@ async def test_runtime_resolution_failure(strict_kernel: Kernel) -> None:
                     cell_id="1",
                     code=textwrap.dedent(
                         """
-                    try:
-                      x = L(1)
-                    except NameError as e:
-                      E = e
-                    raise E
+                    x = L(1)
```
dmadisetti, line 1219: Oh cool!

dmadisetti (commented):
Maybe I need to digest the graphs changes a bit more, but what happens if you have 2 import only cells that share an import:

```python
import numpy as np
import pandas as pd
```

```python
import jax numpy as np
```

does carried outputs catch the double def?

tests/_runtime/test_runtime.py
dmadisetti: I put in this unit test more to document it as a known bug- so the fact it passes correctly now is great !

akshayka:
> Maybe I need to digest the graphs changes a bit more, but what happens if you have 2 import only cells that share an import:

It will still count as a double definition and the runtime will complain, as it does today.

Imports are only carried over from a previous cell to the one replacing it if the import statements are identical (not just the definitions). That's here: https://github.com/marimo-team/marimo/pull/1955/files#diff-ad27e82003280db515fa305a115cd98eae450e2220e027d097c61ac1434890c3R186-R187

So if you replace `import numpy as np` with `import jax.numpy as np`, `np` won't be carried over.

I think carried imports don't come into play across multiple cells (because we check to make sure the full import path matches), but maybe I'm missing an edge case. In an earlier version I neglected to check the full path and had the bug that you're perhaps talking about.

dmadisetti:
I think non-obvious edge cases can be addressed on a one off case, as long as explicitly rerunning the import cell works as expected.

The non-ideal case is that some bug gets the notebook into a bad state, and then the user can't escape it unless they restart the kernel, or intentionally invalidate the cell

dmadisetti:
Reminds me, you should remove this comment with this PR: https://github.com/marimo-team/marimo/blob/da22d015ee0d406ed53c01c3321a20d129507480/marimo/_runtime/executor.py#L71-L92

mscolnick (approved):

PR #1981, opened 2024-08-08
Fix #1403: Redirecting Error Messages

### LINKED ISSUE #1403: Possibility of redirecting error messages to browser console after opting in
### Description

I'm hosting an application on a remote server, that I don't have direct access to. For whatever reason my application behaves differently remotely than locally, as some of the UI components aren't rendering. To find out why, I would very much like to see the error messages, but these are hidden away when the application runs.

### Suggested solution

I propose a feature in which the user can opt in by setting an env variable like MARIMO_DEBUG=True, which causes stdout/stderr to be routed to the browser console. All output should somehow contain metadata about the cell, from which they are generated, in order to help debug components.

### Alternative

The alternative is to have direct access to a server on which the application is hosted. This is not always possible.

### Additional context

_No response_

metaboulie:
## 📝 Summary

Fix #1403: Redirecting Error Messages

## 🔍 Description of Changes

1. Add a command-line option to `marimo run` that forwards console logs to the frontend. Something like `marimo run notebook.py --redirect-console-to-browser`
2. If `is_edit_mode` is `False` and if `redirect_console_to_browser` is `True`, `stdout` and `stderr` should be constructed in `runtime.py`
3. `launch_kernel` in `runtime.py` is updated to accept a boolean argument `redirect_console_to_browser` , and the value of the CLI option is wired down to this function.

Screenshot 2024-08-08 at 19 54 25

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


marimo/_cli/cli.py
```diff
@@ -399,6 +400,7 @@ def new(
         cli_args={},
         auth_token=_resolve_token(token, token_password),
         base_url=base_url,
+        redirect_console_to_browser=False,
```
akshayka, line 403: Should this be `True`?

marimo/_cli/cli.py
```diff
@@ -490,6 +492,14 @@ def new(
     multiple=True,
     help="Allowed origins for CORS. Can be repeated.",
 )
+@click.option(
+    "--redirect-console-to-browser",
+    is_flag=True,
+    default=False,
+    show_default=True,
+    type=bool,
+    help="Redirect console logs to the browser.",
```
akshayka, line 501:
```suggestion
    help="Redirect console logs to the browser console.",
```

akshayka (commented):
This looks great! 🎉 I just have two small comments. Thank you!

marimo/_cli/cli.py
metaboulie: Done 🙂, also updated this in `apps.md`

akshayka (approved):
Nice work!

PR #1993, opened 2024-08-10
feat: Account for state and variable type in save

dmadisetti:
## 📝 Summary

Flesh out of `mo.persistent_cache()`

 - Adds StateRegistry
 - Adds numpy/ tensor hashing
 - Cleans up ContentAddress hashes
 - Introduces pure function detection
 
I read over Mandala, and took a few ideas: https://github.com/amakelov/mandala/discussions/15

TODO:
 - [x] Tests
 - [x] Other easily serializable types like numpy arrays (good tips from joblib: https://github.com/joblib/joblib/blob/bca1f4216a38cff82a85371c45dde79bed977d0e/joblib/hashing.py#L245


tests/_save/test_cache.py
```diff
@@ -171,3 +174,142 @@ async def test_cache_comment_line(self, any_kernel: Kernel) -> None:
         assert k.errors == {}
         assert k.globals["X"] == 1
         assert k.globals["Y"] == 2
+
+
+class TestStateCache:
```
dmadisetti, line 179: TODO:
- [ ] Transitive state refs
- [ ] UI tests
- [ ] Directly test pure_*

marimo/_save/hash.py
```diff
@@ -4,106 +4,173 @@
...
-        | visitor.refs
-    ):
-        ref = is_local_then_mangle(ref, cell_id)
+    fn_cache: FN_CACHE_TYPE = {}
+    for local_ref in sorted(refs):
+        ref = if_local_then_mangle(local_ref, cell_id)
+        if ref in sys.modules:
+            # TODO: ask Akshay about module watching
+            version = ""
+            if pin_modules:
```
dmadisetti, line 117: TODO:
 - [ ] Should invalidate if watched module refreshes

marimo/_runtime/primitives.py
```diff
@@ -0,0 +1,254 @@
...
+    return any(issubclass(value, primitive) for primitive in PRIMITIVES)
+
+
+def is_clone_primitive(value: Any) -> bool:
+    return isinstance(value, CLONE_PRIMITIVES)
+
+
+def is_data_primitive(_value: Any) -> bool:
+    # TODO
+    # check for __array__ and __array_interface__ to determine if it's a numpy
```
dmadisetti, line 47: Handle tuples too. refer to joblib

dmadisetti (commented):
Really need to expand tests
I can put documentation in with this PR, or a follow up.

I have an idea for native save, where cache checks are done right before execute, once this is in a good state- I'll follow up with that PR under an experimental flag- and then probably dog-food it for a bit

dmadisetti:
I think final TODOs:
 - [ ] state registry should be attached to context instead of global
 - [ ] ui registry `lookup` not working as intended

marimo/_runtime/primitives.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def is_data_primitive(value: Any) -> bool:
+    if is_primitive(value):
+        return True
+
+    # If a numpy array, ensure that it's not an object array.
+    if hasattr(value, "__array_interface__"):
+        dtype = getattr(value, "dtype", None)
+        if dtype is not None and dtype.hasobject:
```
akshayka, line 54: Since this is duck-typed, we should have a hasattr:

`if dtype is not None and hasattr(dtype, 'hasobject') and ...`

marimo/_runtime/primitives.py
```diff
@@ -0,0 +1,205 @@
...
+
+def build_ref_predicate_for_primitives(
+    glbls: dict[str, Any],
+    primitives: Optional[tuple[type, ...]] = None,
+) -> Callable[[Name, VariableData], bool]:
+    """
+    Builds a predicate function to determine if a reference should be included
+
+    Args:
+        glbls: The global variables dictionary to base the predicate on
```
akshayka, line 140: Can you add an arg and description for `primitives`?

marimo/_runtime/state.py
```diff
@@ -2,22 +2,108 @@
...
 
+@dataclass
+class StateItem(Generic[T]):
+    id: int
+    ref: weakref.ref[State[T]]
+
+
+class StateRegistry:
+    _states: dict[str, StateItem[Any]] = {}
+    _inv_states: dict[int, set[str]] = {}
```
akshayka, line 24: comment on what `inv_states` is?

marimo/_runtime/state.py
```diff
@@ -2,22 +2,108 @@
...
+
+    def register(self, state: State[T], name: Optional[str] = None) -> None:
+        if name is None:
+            name = str(uuid4())
+        if id(state) in self._inv_states:
```
akshayka, line 29: FYI I have had problems with using Python id as a unique key, because Python reuses ids. Not sure if this would be a problem here.

marimo/_runtime/runner/hooks_post_execution.py
```diff
@@ -169,6 +169,17 @@ def _store_reference_to_output(
...
 
+def _store_state_reference(
+    cell: CellImpl,
+    runner: cell_runner.Runner,
+    run_result: cell_runner.RunResult,
+) -> None:
+    del run_result
+    ctx = get_context()
+    ctx.state_registry.register_scope(runner.glbls, defs=cell.defs)
+    ctx.state_registry.retain_active_states(set(runner.glbls.keys()))
```
akshayka, line 180: Can you explain why this is necessary? For UI elements we only register on creation, and that I think is sufficient. I noticed state objects are already registered on creation in `State.__init__`.

marimo/_save/hash.py
```diff
@@ -4,106 +4,246 @@
...
 
 
-def build_content_hash(
-    graph: DirectedGraph,
+def standardize_tensor(tensor: Tensor) -> Optional[Tensor]:
+    # TODO: Consider moving to a more general utility module.
+    if hasattr(tensor, "__array__") or hasattr(tensor, "toarray"):
+        if not hasattr(tensor, "__array_interface__"):
+            DependencyManager.numpy.require(
+                "to render images from generic arrays in `mo.image`"
```
akshayka, line 122: Stale comment?

tests/_save/test_hash.py
```diff
@@ -146,14 +160,244 @@ def two(persistent_cache, MockLoader, shared) -> tuple[int]:
...
+
+class TestDataHash:
+    @staticmethod
+    @pytest.mark.skipif(
+        not DependencyManager.numpy.has(),
+        reason="optional dependencies not installed",
+    )
+    @pytest.mark.skipif(
+        "sys.version_info < (3, 11) or sys.version_info >= (3, 12)"
+    )
```
akshayka, line 356: Tests for tensors?

akshayka (commented):
Huge change! 💪

I'm still digesting this and revisit to review the larger diffs under `_save`. In the meantime some comments and questions.

marimo/_save/cache.py
```diff
@@ -6,36 +6,102 @@
...
+            if isinstance(value, SetFunctor):
+                self.defs[ref] = value._state()
+            elif isinstance(value, UIElement):
+                self.defs[ref] = value.value
+            else:
+                raise CacheException(
+                    "Failure while saving cached values. "
+                    "Unexpected stateful reference type "
+                    f"({type(ref)}:{ref})."
+                )
```
akshayka, line 96: Are `restore` and `load` the same function? Should one be deleted?

marimo/_save/cache.py
```diff
@@ -6,36 +6,102 @@
...
+            frame.f_locals[var] = self.defs[lookup]
+
+        defs = {**globals(), **frame.f_locals}
+        for ref in self.stateful_refs:
+            if ref not in defs:
+                raise CacheException(
+                    "Failure while restoring cached values. "
+                    "Cache expected a reference to a "
+                    f"variable that is not present ({ref})."
+                )
```
akshayka, line 61: Does a Cached exception leave the frame in a weird state, since variables have already been restored? Do they need to be un-restored, or is it fine? My guess is fine but just double checking ...

marimo/_save/hash.py
```diff
@@ -4,106 +4,246 @@
...
 
-def build_execution_hash(
-    graph: DirectedGraph, cell_id: CellId_t
-) -> Optional[ValidCacheSha]:
+def hash_and_dequeue_execution_refs(
+    hash_alg: Hash,
+    cell_id: CellId_t,
+    graph: DirectedGraph,
+    refs: set[str],
+) -> None:
```
akshayka, line 96: Can you add a function docstring? It looks like this mutates `refs` and `hash_alg`?

marimo/_save/hash.py
```diff
@@ -4,106 +4,246 @@
...
+def hash_and_dequeue_content_refs(
+    hash_alg: Hash,
     cell_id: CellId_t,
-    visitor: ScopedVisitor,
     defs: dict[str, Any],
-) -> Optional[ValidCacheSha]:
+    refs: set[str],
+    graph: DirectedGraph,
+    pin_modules: bool = False,
+) -> None:
```
akshayka, line 162: Function docstring?

tests/_save/test_hash.py
akshayka: Such as torch/tensorflow/scipy sparse

marimo/_save/hash.py
```diff
@@ -4,106 +4,246 @@
...
+            refs.remove(local_ref)
+
+
+def normalize_and_extract_ref_state(
+    visitor_refs: set[str],
+    refs: set[str],
+    defs: dict[str, Any],
+    cell_id: CellId_t,
+    ctx: RuntimeContext,
+) -> set[str]:
```
akshayka, line 208: Same thing, function docstring please.

marimo/_save/hash.py
```diff
@@ -154,20 +294,101 @@ def cache_attempt_from_hash(
...
+        )
+        # Need to run extract again for the expanded ref set.
+        stateful_refs |= normalize_and_extract_ref_state(
+            visitor.refs, refs, defs, cell_id, ctx
+        )
+
+        # Attempt content hash
+        hash_and_dequeue_content_refs(
+            hash_alg, cell_id, defs, refs, graph, pin_modules=pin_modules
+        )
```
akshayka, line 343: Can you leave a comment explaining this algorithm, and why it requires multiple passes?

marimo/_save/cache.py
```diff
@@ -6,36 +6,102 @@
...
+            value = defs[ref]
+            if isinstance(value, SetFunctor):
+                value(self.defs[ref])
+            elif isinstance(value, UIElement):
+                value._update(self.defs[ref])
```
akshayka, line 66: nice, very clean

marimo/_save/hash.py
```diff
@@ -4,106 +4,246 @@
...
+        if not hasattr(tensor, "__array_interface__"):
+            DependencyManager.numpy.require(
+                "to render images from generic arrays in `mo.image`"
+            )
+            import numpy
+
+            # Capture those sparse cases
+            if hasattr(tensor, "toarray"):
+                tensor = tensor.toarray()
+            tensor = numpy.array(tensor)
```
akshayka, line 129: For tensors on GPU, do you need to move them to CPU? I think torch raises when trying to convert to numpy without first moving to CPU. Just checking ...

akshayka (commented):
Another round of comments/questions

marimo/_save/cache.py
dmadisetti: Will put in doc-strings too, but I think `update` might be a better name for cache

`update`: globals -> backfill object to be saved
`restore`: saved object -> globals

lmk if you have better name ideas!

marimo/_save/hash.py
dmadisetti: Good point. Carrying around a lot of state, probably better just to make a `BlockHash` class

marimo/_runtime/runner/hooks_post_execution.py
akshayka: > I already recognize this is an enormous PR, so LMK if I should just try to address the documentation and cleanup here, and tackle these issues in another PR- or try to get it right from the beginning. 

Up to you. You can tackle in another PR if you like -- just leave a comment or TODO in the source if you do.

marimo/_save/cache.py
akshayka: Ah I see when loads from defs -> frame, the other loads from frame -> defs, I missed that. Yea `update` seems like a good name.

akshayka:
TODOs sound good.

> Maybe registry changes? [#1993 (comment)](https://github.com/marimo-team/marimo/pull/1993#discussion_r1752489900)

Registry changes not required in this PR.

marimo/_save/cache.py
dmadisetti: It will be the same as if the cell failed. The `with` statement cannot be nested under a try catch- so it will kick back out to runner which will clean up the defs

marimo/_runtime/state.py
dmadisetti: Left a comment, I think there's aggressive enough cleanup

akshayka:
> Are we OK adding something as large as torch or scipy to optional deps? 

Yea, I'm okay with it. Hopefully with `uv` it's magically fast. If not we can find a way to make the CI more efficient, for example we could split up the optional deps into multiple groups, and run just one workflow with the tensors deps. But that's likely premature.

> Thoughts on what to add? 

Whatever we'd like to support should be added, I believe.

akshayka (dismissed):
Looks great! I've fine with leaving torch/scipy/etc tensor testing to a future PR

dmadisetti:
Not sure why the FE build was stuck on pending. rebased from main to trigger CI again

dmadisetti:
Hmm. Maybe a github bug? Looks like the action path was changed after this PR.
Weird, because it's rebased. Closing out an opening new PR #2316

PR #1998, opened 2024-08-11
fix: duckdb lowerbound

akshayka:
Add an optional [sql] section to the pyproject.toml which enumerates our SQL dependencies with lower bounds, currently only `duckdb >= 1.0`.

Tell users to obtain SQL dependencies with pip install marimo[sql] instead of pip install duckdb, to ensure they get the correct versions.

Update the kernel capabilities check for SQL to check duckdb verison.

Fixes a bug in which we'd enable SQL support even though users were using incompatible versions of duckdb, causing the kernel to crash.

### DIFF
--- a/marimo/_messaging/ops.py
+++ b/marimo/_messaging/ops.py
@@ -6,6 +6,7 @@ Messages that the kernel sends to the frontend.
 
 from __future__ import annotations
 
+import importlib.metadata
 import json
 import sys
 import time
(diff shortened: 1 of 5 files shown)


marimo/_messaging/ops.py
```diff
@@ -339,7 +342,12 @@ class KernelCapabilities:
     terminal: bool = False
 
     def __post_init__(self) -> None:
-        self.sql = DependencyManager.has_duckdb()
+        if DependencyManager.has_duckdb():
```
mscolnick, line 345: I wonder if we could pass the version as an argument to this call. It could provide a better warning too or suggestion to upgrade.

PR #1999, opened 2024-08-11
fix: don't autorun on rename when runtime is lazy

akshayka:
Account for lazy runtime on filename change.


dmadisetti (dismissed):
Looks like schema and snapshot failures. Thanks!

mscolnick (dismissed):
worth adding a small test for this?

mscolnick (approved):

PR #2010, opened 2024-08-13
improvement: add Presentation toggling to menu dropdown, allow forcing slides by URL query param

mscolnick:
* add Presentation toggling to menu dropdown
   * if we are currently in 'edit' mode, it will toggle to 'present' mode before changing the layout type
   * by adding it to the notebook dropdown, it also gets added to the Command Palette
* allow forcing slides by URL query param. e.g. https://marimo.app/?slug=14ovyr8&mode=run&view-as=slides (only works once this lands)
* DRY up some query param strings through the codebase, with docs

image
image


akshayka (approved):
👌

Issue #970, opened 2024-03-20
Freeze Table Column


madystricker:
### Description

It would be great to be able to mark the first column of a table static, similar to freezing the first column in an Excel table.

### Suggested solution

Create a True/False initialization arg for the mo.ui.table() component that freezes the first column if True.

akshayka:
I like that idea! We could allow for freezing the first row too.

hhk998402:
I can take a shot at this

mscolnick:
hey @hhk998402 - that would be awesome! let me know if you need help or want to pair?

initial thoughts on how we can implement this:
- add `frozen_columns` to `mo.ui.table(frozen_columns=)`
  - for an initial version it could just be a number (e.g. 1, 2), but for something more complex, we could make it a list of strings as well which maps to the column names to pin
- Checkout `frontend/src/components/data-table/column-wrapping` for adding a custom table plugin
  - we can do something similar for `frozen-columns`
- then somewhere in `frontend/src/components/data-table/data-table.tsx` we can use this new plugin to check `column.isFrozen() && "sticky"` to make the header sticky

wasimsandhu:
@hhk998402 Were you still working on this? If not, I'll grab this one next.

metaboulie:
> @hhk998402 Were you still working on this? If not, I'll grab this one next.

Hey, I'm currently working on this and have almost fixed it, just some style issues TBD with Myles. I hope this won't bother you. 
🙂

wasimsandhu:
@metaboulie Oh nice! Looking forward to seeing it 🙂

metaboulie:
@wasimsandhu Hey, I think I won't work on this issue, please grab it if you want 🙂

mscolnick:
Closed by https://github.com/marimo-team/marimo/issues/970

PR #2012, opened 2024-08-13
feat: column pinning in mo.ui.table UI

metaboulie:
## 📝 Summary

Column Pinning for tables

## 🔍 Description of Changes

Add column pinning (both left pinning and right pinning) feature for tables

Screenshot 2024-08-14 at 00 36 15

- Implement column pinning as a tanstack table feature
- Pin the selection checkbox column when any column is left pinned
- Unpin the selection checkbox column when no column is left pinned
- Custom styles for pinned columns 

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


frontend/src/components/data-table/column-header.tsx
```diff
@@ -162,6 +166,52 @@ export const DataTableColumnHeader = <TData, TValue>({
...
+  // render pinning options
+  const renderPinning = () => {
+    // if the column cannot be pinned, return null
+    if (!column.getCanPin()) {
+      return null;
+    }
+    return (
+      <DropdownMenuSub>
+        <DropdownMenuSubTrigger>
+          <PinIcon className="mr-2 h-3.5 w-3.5 text-muted-foreground/70" />
```
mscolnick, line 178: if not too much trouble, i think maybe we can extract this className (since its used in other icons). 

something like:
```
const iconClassName = "mr-2 h-3.5 w-3.5 text-muted-foreground/70";
```

frontend/src/components/data-table/column-pinning/feature.ts
```diff
@@ -0,0 +1,194 @@
...
+
+      // get if the column is the first pinned column on the right
+      const isFirstRightPinnedColumn =
+        isPinned === "right" && column.getIsFirstColumn("right");
+
+      return {
+        // apply a box shadow to the column if
+        // it is the last pinned column on the left
+        // or the first pinned column on the right
+        boxShadow: isLastLeftPinnedColumn
```
mscolnick, line 165: it could be nice if we can use classnames instead of styles - so we can leverage our tailwind css.

could we do `shadow-inner` or `shadow-sm` here?

maybe we could add `column.getCommonPinningClassNames` as well.

mscolnick: or maybe this is better:
```ts
getCommonPinningClassNameAndStyles() => {
  return {
   className: "...",
   styles: {},
  }
}
```

mscolnick:
this is great! it will be super useful. how would this fit in this request (https://github.com/marimo-team/marimo/issues/970)? this request specifically wanted a way to do it programmatically (so its saved across sessions and for all users). do you think we can add `pin_columns: int = 0` to `mo.ui.table`. which defaults sets / initializes the the first `pin_columns` as left-pinned

frontend/src/components/data-table/column-pinning/feature.ts
```diff
@@ -0,0 +1,194 @@
...
+        width: column.getSize(),
+
+        // set the column's left position
+        left:
+          isPinned === "left"
+            ? `${
+                column.getStart("left") -
+                (column.id === "select" ? 0 : (column.getSize() * 4) / 5)
+              }px`
+            : undefined,
```
mscolnick, line 181: there seems to be a small gap between columns with this method:

image

frontend/src/components/data-table/column-pinning/feature.ts
mscolnick: just tried this and it may not work. at least instead of `gray`, could we do `var(--base-shadow)`

frontend/src/components/data-table/column-pinning/feature.ts
mscolnick: you could use the column ordering API in here instead. 

https://tanstack.com/table/v8/docs/api/features/column-ordering

still keep the same left/right state, but after each pin/unpin, you can update the ordering:

```
const allColumnIds = table.getColumnIds()
const unsetColumnIds = allColumnIds.filter(id => `!left.includes(id) && !right.includes(id)`)
return table.updateColumnOrder([..left, ...unsetColumnIds, ...right]);
```

then you wont need to mess with `width` and `sizing` and you can just use `sticky`

frontend/src/components/data-table/column-pinning/feature.ts
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        // If only "select" column is pinned to the left, unpin it
+        if (
+          position !== "left" &&
+          newState.left?.length === 1 &&
+          newState.left[0] === "select"
+        ) {
+          // unpin the 'select' column
+          table.getColumn("select")?.pin(false);
```
mscolnick, line 110: can we move `"select"` to a constant to be shared between here and the `SelectColumn`. maybe we can rename to "__select__" so its more unique

frontend/src/components/data-table/column-pinning/types.ts
```diff
@@ -0,0 +1,51 @@
+/* Copyright 2024 Marimo. All rights reserved. */
```
mscolnick, line 1: wait - there is actually already a column pinning feature: https://tanstack.com/table/v8/docs/api/features/column-pinning

so we can just re-use that instead of having our own logic

frontend/src/components/data-table/column-pinning/feature.ts
mscolnick: actually we can disregard this last message, i think what you have is close enough if we can re-use the existing table feature instead of building our own

frontend/src/components/data-table/column-pinning/types.ts
mscolnick: there are two examples to do this:

* https://tanstack.com/table/v8/docs/framework/react/examples/column-pinning
* https://tanstack.com/table/v8/docs/framework/react/examples/column-pinning-sticky

it seems like you did something similar to the second (sticky and absolute positioning). i find the first may be a bit better / easier. 

you can do something like
```
const renderHeaderItem = (column) => {
   ...we have this logic
}

<THead>
 {table.getLeftHeaders().map(renderHeaderItem)}
 {table.getCenterHeaders().map(renderHeaderItem)}
 {table.getRightHeaders().map(renderHeaderItem)}
</THead>
```

thoughts?

metaboulie: yeah this is way better, will change to this

frontend/src/components/data-table/column-pinning/feature.ts
metaboulie: Do you mean to rename `"select"` to `"__select__"` here

https://github.com/marimo-team/marimo/blob/9a7304dd569a7b307db56308129edf29a9c129cc/frontend/src/components/data-table/columns.tsx#L151

mscolnick: yea i am realizing that would be better since a column could be named `"select"` coming from a dataframe

PR #2017, opened 2024-08-14
fix(mo.ui.table): implement row count limit on column summaries; various bug fixes

### LINKED ISSUE #1979: Automatic column summaries of the table component should have a upper row count limit
### Describe the bug

I'm using marimo to work on large datasets with 100 million rows, and found the "run cell" button unresponsive sometimes. After some investigation, I found during the unresponsive time, the kernel is busy running the `get_column_summaries` function, which is issued by the table component in the frontend.

Calculating summaries for 100m rows will take huge amounts of time, and the unresponsive behavior will cause confusions to users. I think marimo should impose a hard limit on column summary feature. Maybe 1 million is a good number?

Or maybe we can try to run the column summary in a background thread. But this will be a huge change I think, and thread safety issue will be hard to deal with.

### Environment

```
{
  "marimo": "0.7.14",
  "OS": "Linux",
  "OS Version": "5.15.0-52-generic",
  "Processor": "x86_64",
  "Python Version": "3.10.12",
  "Binaries": {
    "Browser": "--",
    "Node": "--"
  },
  "Requirements": {
    "click": "8.1.7",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pymdown-extensions": "10.9",
    "pygments": "2.18.0",
    "tomlkit": "0.13.0",
    "uvicorn": "0.30.3",
    "starlette": "0.38.2",
    "websockets": "12.0",
    "ruff": "0.5.5"
  }
}
```

### Code to reproduce

_No response_

### LINKED ISSUE #2016: mo.ui.table()'s column summaries are not updated after changing search/filter criteria
### Describe the bug

When I'm trying to fix #1979, I see the implementation of `get_column_summaries` uses `self._filtered_manager`:

https://github.com/marimo-team/marimo/blob/3969f95cde6146780d1fc3e856965d5675cd152f/marimo/_plugins/ui/_impl/table.py#L340-L344

So I thought a smooth user experience after applying row limit on column summaries would be: user shows a large table -> marimo complains about too many rows -> user applies filter/search -> marimo happily displays column summaries.

But after I implemented the backend logic, I found the column summaries are not updated after changing the filter (no new `get_column_summaries` function call in DevTools' network tab). I think it might be related to the `columnSummaries` are not put inside a `useEffect` in the following code (I'm completely new to React so correct me if I'm wrong):

https://github.com/marimo-team/marimo/blob/3969f95cde6146780d1fc3e856965d5675cd152f/frontend/src/plugins/impl/DataTablePlugin.tsx#L252-L265

I'm not sure whether it is a bug or working as intended. Posting here so maintainers are aware of it.

### Environment

```
{
  "marimo": "0.7.19",
  "OS": "Linux",
  "OS Version": "5.15.0-52-generic",
  "Processor": "x86_64",
  "Python Version": "3.10.12",
  "Binaries": {
    "Browser": "--",
    "Node": "--"
  },
  "Requirements": {
    "click": "8.1.7",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pymdown-extensions": "10.9",
    "pygments": "2.18.0",
    "tomlkit": "0.13.0",
    "uvicorn": "0.30.3",
    "starlette": "0.38.2",
    "websockets": "12.0",
    "ruff": "0.5.5"
  }
}
```

### Code to reproduce

_No response_

mutongx:
## 📝 Summary

1. Add a hard-coded row count upper limit on table's column summaries feature.
2. Update column summaries after changing search/filter criteria
3. Fix the inconsistency of table filtering and row count limiting
4. Some code reorganization

Fixes #1979.
Fixes #2016.

## 🔍 Description of Changes

- Add a `is_disabled` field to `ColumnSummaries` so it can indicate itself being disabled
    - I tried to use exceptions, but I found Function objects only propagates human-readable error messages. So it's unfeasible.
- Add a hard-coded 1,000,000 row count limit to the column summaries feature
    - marimo already imposes a column count limit, so I think limiting only rows here should be OK for most situations
- Make column summaries react to filter/search criteria changes, so the summaries gets updated after such operations
- Separate the `table._filtered_manager` to `table._searched_manager` and `table._limited_manager`
    - The `_searched_manager` holds the full data after search (query, filter and sort), and the `_limited_manager` holds the first 20000 rows of `_searched_manager`
- Reorganized the code in `table.__init__()` for better clarity

## 📋 Checklist

- [ ] (Not applicable) For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://discord.gg/JE7nhX6mD8), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/plugins/impl/DataTablePlugin.tsx
+++ b/frontend/src/plugins/impl/DataTablePlugin.tsx
@@ -40,6 +40,10 @@ import { Objects } from "@/utils/objects";
 
 type CsvURL = string;
 type TableData<T> = T[] | CsvURL;
+interface ColumnSummaries {
+  summaries: ColumnHeaderSummary[];
+  is_disabled?: boolean;
+}
 
 /**
  * Arguments for a data table
@@ -65,10 +69,7 @@ interface Data<T> {
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
 type Functions = {
   download_as: (req: { format: "csv" | "json" }) => Promise<string>;
-  get_column_summaries: (opts: {}) => Promise<{
-    summaries: ColumnHeaderSummary[];
-    is_disabled?: boolean;
-  }>;
+  get_column_summaries: (opts: {}) => Promise<ColumnSummaries>;
   search: <T>(req: {
     sort?: {
       by: string;
@@ -193,6 +194,10 @@ export const LoadingDataTableComponent = memo(
     const [sorting, setSorting] = useState<SortingState>([]);
     const [searchQuery, setSearchQuery] = useState<string>("");
     const [filters, setFilters] = useState<ColumnFiltersState>([]);
+    // Column summaries state
+    const [summaries, setSummaries] = useState<ColumnSummaries>({
+      summaries: [],
+    });
 
     // Data loading
     const { data, loading, error } = useAsyncData<T[]>(async () => {
@@ -252,13 +257,27 @@ export const LoadingDataTableComponent = memo(
     }, [sorting, search, filters, searchQuery, props.fieldTypes, props.data]);
 
     // Column summaries
-    const { data: columnSummaries, error: columnSummariesError } =
-      useAsyncData(() => {
+    const { data: columnSummariesData, error: columnSummariesError } =
+      useAsyncData<ColumnSummaries>(() => {
         if (props.totalRows === 0) {
           return Promise.resolve({ summaries: [] });
         }
+        // Clear the summary first to avoid inconsistency with table
+        setSummaries({ summaries: [] });
         return props.get_column_summaries({});
-      }, [props.get_column_summaries, props.totalRows, props.data]);
+      }, [
+        props.get_column_summaries,
+        filters,
+        searchQuery,
+        props.totalRows,
+        props.data,
+      ]);
+
+    useEffect(() => {
+      if (columnSummariesData) {
+        setSummaries(columnSummariesData);
+      }
+    }, [columnSummariesData]);
 
     useEffect(() => {
       if (columnSummariesError) {
@@ -292,8 +311,7 @@ export const LoadingDataTableComponent = memo(
         <DataTableComponent
           {...props}
           data={data || Arrays.EMPTY}
-          columnSummaries={columnSummaries?.summaries}
-          columnSummariesDisabled={columnSummaries?.is_disabled}
+          columnSummaries={summaries}
           sorting={sorting}
           setSorting={setSorting}
           searchQuery={searchQuery}
@@ -324,7 +342,6 @@ const DataTableComponent = ({
   fieldTypes,
   download_as: downloadAs,
   columnSummaries,
-  columnSummariesDisabled,
   className,
   setValue,
   sorting,
@@ -338,14 +355,13 @@ const DataTableComponent = ({
 }: DataTableProps<unknown> &
   DataTableSearchProps & {
     data: unknown[];
-    columnSummaries?: ColumnHeaderSummary[];
-    columnSummariesDisabled?: boolean;
+    columnSummaries: ColumnSummaries;
   }): JSX.Element => {
   const resultsAreClipped =
     hasMore && (totalRows === "too_many" || totalRows > 0);
 
   const chartSpecModel = useMemo(() => {
-    if (!fieldTypes || !data || !columnSummaries) {
+    if (!fieldTypes || !data || !columnSummaries.summaries) {
       return ColumnChartSpecModel.EMPTY;
     }
     const fieldTypesWithoutExternalTypes = Objects.mapValues(
@@ -355,12 +371,12 @@ const DataTableComponent = ({
     return new ColumnChartSpecModel(
       data,
       fieldTypesWithoutExternalTypes,
-      columnSummaries,
+      columnSummaries.summaries,
       {
         includeCharts: !resultsAreClipped,
       },
     );
-  }, [data, fieldTypes, columnSummaries, resultsAreClipped]);
+  }, [data, fieldTypes, columnSummaries.summaries, resultsAreClipped]);
 
   const columns = useMemo(
     () =>
@@ -414,7 +430,7 @@ const DataTableComponent = ({
           Result clipped. If no LIMIT is given, we only show the first 300 rows.
         </Banner>
       )}
-      {columnSummariesDisabled && (
+      {columnSummaries.is_disabled && (
         <Banner className="mb-2 rounded">
           Column summaries are unavailable. Filter your data to fewer than
           1,000,000 rows.
--- a/frontend/src/plugins/impl/DataTablePlugin.tsx
+++ b/frontend/src/plugins/impl/DataTablePlugin.tsx
@@ -194,10 +194,6 @@ export const LoadingDataTableComponent = memo(
     const [sorting, setSorting] = useState<SortingState>([]);
     const [searchQuery, setSearchQuery] = useState<string>("");
     const [filters, setFilters] = useState<ColumnFiltersState>([]);
-    // Column summaries state
-    const [summaries, setSummaries] = useState<ColumnSummaries>({
-      summaries: [],
-    });
 
     // Data loading
     const { data, loading, error } = useAsyncData<T[]>(async () => {
@@ -257,13 +253,11 @@ export const LoadingDataTableComponent = memo(
     }, [sorting, search, filters, searchQuery, props.fieldTypes, props.data]);
 
     // Column summaries
-    const { data: columnSummariesData, error: columnSummariesError } =
+    const { data: columnSummaries, error: columnSummariesError } =
       useAsyncData<ColumnSummaries>(() => {
         if (props.totalRows === 0) {
           return Promise.resolve({ summaries: [] });
         }
-        // Clear the summary first to avoid inconsistency with table
-        setSummaries({ summaries: [] });
         return props.get_column_summaries({});
       }, [
         props.get_column_summaries,
@@ -272,12 +266,6 @@ export const LoadingDataTableComponent = memo(
         props.totalRows,
         props.data,
       ]);
-
-    useEffect(() => {
-      if (columnSummariesData) {
-        setSummaries(columnSummariesData);
-      }
-    }, [columnSummariesData]);
 
     useEffect(() => {
       if (columnSummariesError) {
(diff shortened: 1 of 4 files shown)


akshayka:
Looking at this, I'm reminded that getting the number rows can be expensive for some tables; @mscolnick, is that a concern here, or was that unrelated?

marimo/_plugins/ui/_impl/table.py
```diff
@@ -68,6 +68,7 @@ class ColumnSummary:
 @dataclass
 class ColumnSummaries:
     summaries: List[ColumnSummary]
+    is_disabled: Optional[bool] = None
```
akshayka, line 71: Doesn't `summaries` being the empty list imply that column summaries are disabled? In which case we could remove the `is_disabled` parameter.

akshayka (commented):
Thanks for the PR! One drive-by comment on the Python code. I'll let @mscolnick review more thoroughly, he is the owner of tables.

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -250,13 +257,27 @@ export const LoadingDataTableComponent = memo(
...
           return Promise.resolve({ summaries: [] });
         }
+        // Clear the summary first to avoid inconsistency with table
+        setSummaries({ summaries: [] });
         return props.get_column_summaries({});
-      }, [props.get_column_summaries, props.totalRows, props.data]);
+      }, [
+        props.get_column_summaries,
+        filters,
+        searchQuery,
```
mscolnick, line 271: you are right - it looks like we were missing these deps so they were not re-rendering on filter changes. thanks for fixing

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -191,6 +194,10 @@ export const LoadingDataTableComponent = memo(
     const [sorting, setSorting] = useState<SortingState>([]);
     const [searchQuery, setSearchQuery] = useState<string>("");
     const [filters, setFilters] = useState<ColumnFiltersState>([]);
+    // Column summaries state
+    const [summaries, setSummaries] = useState<ColumnSummaries>({
```
mscolnick, line 198: i dont think we need to keep track of this state. this is the same as what is inside `useAsyncData`

marimo/_plugins/ui/_impl/table.py
mscolnick: it is helpful to know if there are no summaries vs it actually being disabled. 

it could have no summaries if they failed for some reason or summaries are not supported in the table type

marimo/_plugins/ui/_impl/tables/table_manager.py
```diff
@@ -26,6 +26,8 @@
 class TableManager(abc.ABC, Generic[T]):
     DEFAULT_ROW_LIMIT = 20_000
     DEFAULT_COL_LIMIT = 100
+    DEFAULT_SUMMARY_ROW_LIMIT = 1_000_000
```
mscolnick, line 29: could we add a comment here or link to the ticket, just so no one changes this without investigating.

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -250,13 +257,27 @@ export const LoadingDataTableComponent = memo(
...
     // Column summaries
-    const { data: columnSummaries, error: columnSummariesError } =
-      useAsyncData(() => {
+    const { data: columnSummariesData, error: columnSummariesError } =
+      useAsyncData<ColumnSummaries>(() => {
         if (props.totalRows === 0) {
           return Promise.resolve({ summaries: [] });
         }
+        // Clear the summary first to avoid inconsistency with table
+        setSummaries({ summaries: [] });
```
mscolnick, line 266: i think it is fine to show the previous summaries while the new ones are loading, otherwise it might flicker a bunch

marimo/_plugins/ui/_impl/table.py
mutongx: Thanks @mscolnick for the explanation!

A better way is to implement a generic error handling for Function objects, so we can pass structural error information from the backend. Actually I even wanted to add the upper limit number to the `ColumnSummaries` object so there will be no hardcoded text in frontend (I didn't because it's even uglier than adding `is_disabled`).

frontend/src/plugins/impl/DataTablePlugin.tsx
mutongx: Should be fixed. I remember my first try of simply adding the dependency to `useAsyncData` failed but this time it worked. Not sure why though.

marimo/_plugins/ui/_impl/tables/table_manager.py
mutongx: Comments are added. I also added a comment for `DEFAULT_ROW_LIMIT` and `DEFAULT_COL_LIMIT`, and please take a look. Thanks.

frontend/src/plugins/impl/DataTablePlugin.tsx
mutongx: Fixed. Thanks!

marimo/_plugins/ui/_impl/table.py
mscolnick: I agree structured errors would be great. we don't have great e2e type-safety yet for structured errors, so its fine to put the details inside the 200 response. 

i think you can add the upper limit as well. that seems like a good idea in case its configurable later.

mscolnick: keeping in sync via comments is also totally great, this looks great!

mscolnick (approved):
looks great! thanks for the contribution!

mutongx:
@mscolnick 

Sorry but please don't merge this PR for now. I think I found some bug in the table component's `search()` function.

When I'm writing unit tests, the following test code fails the assert:

```python
def test_table_with_too_many_rows_column_summaries_disabled():
    data = {"a": list(range(TableManager.DEFAULT_SUMMARY_ROW_LIMIT + 1))}
    table = ui.table(data)
    summaries = table.get_column_summaries(EmptyArgs())
    assert summaries.is_disabled is True
```

It's because `table._filtered_manager` actually gets limited to 20000 rows in the following line, so `_filtered_manager` is expected to never reach the 1m limit.

https://github.com/marimo-team/marimo/blob/60b3e5e5c3159a7bfa48fb563cc69f038dd5ff79/marimo/_plugins/ui/_impl/table.py#L262

But `_filtered_manager` is reset to the original `self._manager` in `search()` function:

https://github.com/marimo-team/marimo/blob/60b3e5e5c3159a7bfa48fb563cc69f038dd5ff79/marimo/_plugins/ui/_impl/table.py#L359-L361

and doesn't get limited:

https://github.com/marimo-team/marimo/blob/60b3e5e5c3159a7bfa48fb563cc69f038dd5ff79/marimo/_plugins/ui/_impl/table.py#L384-L386

So there might be some inconsistency in the semantic of `_filtered_manager` here. If we expect `_filtered_manager` to be always less than 20000 rows, then this PR will be invalid. However, a column summary of only 20000 rows is not useful at all.

What do you think?

mutongx:
This should be easy to fix though. Remove the `self._filtered_manager = self._filtered_manager.limit(row_limit)` in table constuctor, and instead only apply the 20000 row limit when sending data to the frontend. I can do this in the PR but I think I need to sort out the logic so there are no potential problems.

mscolnick:
@mutongx , yep i agree with that. I think this might just be an existing bug/quirk. 

You could do something like this:
```diff
        has_more = total_rows == "too_many" or total_rows > row_limit
+        limited_manager = self._filtered_manager
        if has_more:
+            limited_manager = self._filtered_manager.limit(row_limit)

        # pagination defaults to True if there are more than 10 rows
        if pagination is None:
            pagination = total_rows == "too_many" or total_rows > 10

        field_types = self._manager.get_field_types()

        super().__init__(
            component_name=table._name,
            label=label,
            initial_value=[],
            args={
                "data": limited_manager.to_data(format_mapping),
```

mscolnick:
thanks for investigating this and fixing these issues

tests/_plugins/ui/_impl/test_table.py
```diff
@@ -307,3 +308,10 @@ def test_table_with_too_many_rows_unknown_total():
     assert table._component_args["has-more"] is True
     assert table._component_args["total-rows"] == "too_many"
     assert len(table._component_args["data"]) == 30
+
+
+def test_table_with_too_many_rows_column_summaries_disabled():
+    data = {"a": list(range(20))}
+    table = ui.table(data, _internal_summary_row_limit=10)
+    summaries = table.get_column_summaries(EmptyArgs())
+    assert summaries.is_disabled is True
```
mscolnick, line 317: can you add another test that when we search/filter that brings it down to less than 10, summaries is re-enabled?

marimo/_plugins/ui/_impl/table.py
```diff
@@ -246,20 +246,33 @@ def __init__(
...
+        # Holds the data after user selecting from the component
+        self._selected_manager: Optional[TableManager[Any]] = None
+
+        # We will need this when calling table manager's to_data()
+        self._format_mapping = format_mapping
+
+        if _internal_total_rows is not None:
+            total_rows = _internal_total_rows
+        else:
+            total_rows = self._manager.get_num_rows(force=True) or 0
```
mutongx, line 273: The original logic is quite complex. Simply checking for `is None` should be fine, as the `_internal_total_rows` are only used in the following situations:

1. `marimo._sql` ([link](https://github.com/marimo-team/marimo/blob/60b3e5e5c3159a7bfa48fb563cc69f038dd5ff79/marimo/_sql/sql.py#L93)), where the value will be `"too_many"` or `None`
2. `test_table.py` ([link](https://github.com/marimo-team/marimo/blob/main/tests/_plugins/ui/_impl/test_table.py)), where the value should always be non-zero

marimo/_plugins/ui/_impl/table.py
```diff
@@ -246,20 +246,33 @@ def __init__(
                 "https://github.com/marimo-team/marimo/issues"
             )
 
-        row_limit = _internal_row_limit or TableManager.DEFAULT_ROW_LIMIT
+        if _internal_row_limit is not None:
+            self._row_limit = _internal_row_limit
+        else:
+            self._row_limit = TableManager.DEFAULT_ROW_LIMIT
```
mutongx, line 252: This rewrite is to keep the code style consistent with `if _internal_total_rows is None` below.

tests/_plugins/ui/_impl/test_table.py
mutongx: Added. Thanks for reminding!

mscolnick (approved):
thanks again for the addition!

PR #2040, opened 2024-08-17
experimetanl: release marimo docker containers to gcr.io

mscolnick:


.github/workflows/release.yml
```diff
@@ -102,3 +111,84 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npm publish --access public
+
+  publish_docker:
```
mscolnick, line 115: i haven't tested this e2e (since it requires a release). once this does release successfully once, i will write up some docs on what docker containers we publish

akshayka (approved):

PR #2095, opened 2024-08-23
fix: macOS logging

akshayka:
can't rely on fork() process start, reinitialize log level

### DIFF
--- a/marimo/_loggers.py
+++ b/marimo/_loggers.py
@@ -16,22 +16,26 @@ _LOG_FORMATTER = LogFormatter()
 _LOGGERS: dict[str, logging.Logger] = {}
 
 
+def log_level_string_to_int(level: str) -> int:
+    level = level.upper()
+    if level == "DEBUG":
+        return logging.DEBUG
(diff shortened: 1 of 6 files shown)


marimo/_loggers.py
```diff
@@ -16,22 +16,26 @@
 _LOGGERS: dict[str, logging.Logger] = {}
 
 
+def log_level_string_to_int(level: str) -> int:
+    level = level.upper()
+    if level == "DEBUG":
+        return logging.DEBUG
+    elif level == "INFO":
+        return logging.INFO
+    elif level == "WARN":
```
mscolnick, line 25: should we support `WARNING` too?

mscolnick: i realize this just moving code, not new code, so no need to do this now

mscolnick (dismissed):

marimo/_loggers.py
akshayka: Added WARNING. In a follow-up PR I will clean up what we're logging in the kernel

PR #2098, opened 2024-08-23
improvement: support dict in mo.ui.table

mscolnick:
Add support for dictionaries in the table - by converting to key/value columns

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/default_table.py
+++ b/marimo/_plugins/ui/_impl/tables/default_table.py
@@ -105,7 +105,7 @@ class DefaultTableManager(TableManager[JsonTableData]):
             # Column major data
             if self.is_column_oriented:
                 new_data: Dict[Any, Any] = {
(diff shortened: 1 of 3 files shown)


marimo/_plugins/ui/_impl/tables/default_table.py
```diff
@@ -183,7 +216,12 @@ def get_summary(self, column: str) -> ColumnSummary:
     def get_num_rows(self, force: bool = True) -> int:
         del force
         if isinstance(self.data, dict):
-            return len(next(iter(self.data.values()), []))
+            if self.is_column_oriented:
+                first = next(iter(self.data.values()), None)
+                assert isinstance(first, list)
```
akshayka, line 221: I think it can be a tuple, no? I'd remove this assert in general, don't want to crash the kernel if can help it

akshayka (approved):
nice touch

PR #2102, opened 2024-08-23
feat: configure auto-add script metadata to notebook file, when using uv

mscolnick:


marimo/_config/config.py
```diff
@@ -159,9 +159,12 @@ class PackageManagementConfig(TypedDict):
     **Keys.**
 
     - `manager`: the package manager to use
+    - `add_script_metadata`: if true, add script metadata to the notebook
+        Currently only supports `uv`
     """
 
     manager: Literal["pip", "rye", "uv", "poetry", "pixi"]
+    add_script_metadata: bool
```
mscolnick, line 167: open to a different name

akshayka (commented):
Instead of tying this to package management, how about tying it to import statements? Either statically or, if we want precise versions, dynamically. That way the user's script metadata will reflect what's actually in the notebook, and could be pruned when imports were removed.

mscolnick:
the reason i did it in package management was because we know the package name (upon successful installation). at import runtime, we know the module name, but not the package name, and our heuristic for converting that could return false positives (e.g. `ibis` and `ibis-framework`)

since we expect the module to be installed, maybe there is some more information we can find?

akshayka:
> the reason i did it in package management was because we know the package name (upon successful installation). at import runtime, we know the module name, but not the package name, and our heuristic for converting that could return false positives (e.g. ibis and ibis-framework)

The package we install is based on our heuristic, though, so I think a package management based solution has the same possibility for false positives? And it will cause the set of packages in the metadata to monotonically grow, instead of actually reflecting the packages used by the notebook.

mscolnick:
> The package we install is based on our heuristic, though, so I think a package management based solution has the same possibility for false positives?

It would be slightly better since the install would have needed to succeed. But still can contain false positives.

dmadisetti:
Hmm, I have something locally for nix provisioning that is a little similar to this but particular to the markdown format:

```markdown
---                                                                                                                                                                                                                                           
title: KANNs                                                                                                                                                                                                                                  
marimo-version: 0.7.2                                                                                                                                                                                                                         
width: medium                                                                                                                                                                                                                                 
nix: |                                                                                                                                                                                                                                        
    with import <nixpkgs> {};                                                                                                                                                                                                                 
    python311.withPackages (                                                                                                                                                                                                                  
        ps: with ps; [ matplotlib numpy ]                                                                                                                                                                                              
    )                                                                                                                                                                                                                                         
filters:                                                                                                                                                                                                                                      
 - /home/dylan/src/marimo-quarto/extensions/marimo/marimo-execute.lua                                                                                                                                                                         
---  
```

I think having metadata could go a long way. @akshayka's concern of decoupling make sense, but `marimo run` could create a sandbox in the case that metadata is provided, forcing all deps to be declared

mscolnick:
@dmadisetti the sandbox is a really cool idea. I was playing with this earlier:

`uv run --isolated marimo edit marimo/_smoke_tests/quak-demo.py` which could be aliased to `marimo edit --sandbox`

dmadisetti:
Hmm, I think my only concern is that the `script` metadata notation doesn't label the metadata block. I think (I guess selfishly), it would be nice to have other provisioning methods 

```python
# /// nix
# with import <nixpkgs> {};
# python311.withPackages (
#     ps: with ps; [ matplotlib numpy ]
# )
# ///
```

still falls under the broad spec and Dockerfile etc.. could also be supported.

Alternatively, this information could also live in a separate file:

`App(sandbox_bootstrap="my_sandbox.nix")`, `App(sandbox_bootstrap="DOCKERFILE")` or `App(sandbox_bootstrap="uv_sandbox.toml")`

akshayka:
Supporting other sandboxes does seem cool. But perhaps that could be an additional/future feature? At minimum we do need to support PEP 723 script metadata.

marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -48,6 +50,42 @@ class UvPackageManager(PypiPackageManager):
...
+        if packages_to_add:
+            self.run(
+                ["uv", "--quiet", "add", "--script", filepath]
+                + packages_to_add
+            )
+        if packages_to_remove:
+            self.run(
+                ["uv", "--quiet", "remove", "--script", filepath]
+                + packages_to_remove
+            )
```
akshayka, line 84: @mscolnick `uv remove` errors out if you try to remove a package that isn't installed, which is why I also filter `packages_to_remove`. An alternative would be to try to remove the packages one by one instead of all at once

marimo/_runtime/runtime.py
```diff
@@ -639,6 +639,27 @@ def _maybe_register_cell(
                 cell_id, code, carried_imports=carried_imports
             )
 
+            # For any newly imported namespaces, add them to the metadata
+            #
+            # TODO(akshayka): Consider using the module watcher to discover
+            # packages used by a notebook; that would have the benefit of
+            # discovering transitive dependencies, ie if a notebook used a
+            # local module that in turn used packages available on PyPI.
```
akshayka, line 647: @mscolnick FYI, this would make package reproducibility more complete

PR #2103, opened 2024-08-23
fix: don't import tomlkit top-level

akshayka:
We don't install tomlkit when running under wasm (and don't need to). This was breaking marimo wasm.

### DIFF
--- a/marimo/_cli/config/commands.py
+++ b/marimo/_cli/config/commands.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 import click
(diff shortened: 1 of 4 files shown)


mscolnick (dismissed):

mscolnick:
@akshayka , you should add this to the pyproject toml

```
# Ban certain modules from being imported at module level, instead requiring
# that they're imported lazily (e.g., within a function definition).
banned-module-level-imports = ["numpy", "pandas"]
```

PR #2116, opened 2024-08-26
chore: add biome-lint to remove unused imports

mscolnick:
Add some biome linting rules that are used in precommit. This helps speed up contributors since now pre-commit can autofix / remove unused imports and commit to the a PR.

Also this forces type-only imports, which helps speed up bundling and development (since tooling can easily strip these).


mscolnick:
prone to merge conflicts, so merging now

PR #2122, opened 2024-08-27
feat(mo.ui.table): Freeze table columns

wasimsandhu:
## 📝 Summary

Users can now freeze columns to the left or right with `mo.ui.table`. Completes #970.

result

## 🔍 Description of Changes

- Create TanStack table feature for column pinning state management (see [API](https://tanstack.com/table/v8/docs/api/features/column-pinning#table-options)).
- Refactor table and use `getLeftHeaderGroups`, etc. and `getLeftVisibleCells`, etc. to render left- and right-frozen cells on either side of regular table cells.
- TanStack table does not store or manage the actual rendered column sizes in `column.getSize()`. Implemented [this workaround](https://github.com/TanStack/table/discussions/3947#discussioncomment-9564867) to set column sizes manually and then use these to calculate offsets for pinned column styles.
- Used default TanStack pinning styles from [this example](https://tanstack.com/table/v8/docs/framework/react/examples/column-pinning-sticky) (there is room for customization / improvement here):
```javascript
const getCommonPinningStyles = (column: Column<Person>): CSSProperties => {
  const isPinned = column.getIsPinned()
  const isLastLeftPinnedColumn =
    isPinned === 'left' && column.getIsLastColumn('left')
  const isFirstRightPinnedColumn =
    isPinned === 'right' && column.getIsFirstColumn('right')

  return {
    boxShadow: isLastLeftPinnedColumn
      ? '-4px 0 4px -4px gray inset'
      : isFirstRightPinnedColumn
        ? '4px 0 4px -4px gray inset'
        : undefined,
    left: isPinned === 'left' ? `${column.getStart('left')}px` : undefined,
    right: isPinned === 'right' ? `${column.getAfter('right')}px` : undefined,
    opacity: isPinned ? 0.95 : 1,
    position: isPinned ? 'sticky' : 'relative',
    width: column.getSize(),
    zIndex: isPinned ? 1 : 0,
  }
}
```

## 📋 Checklist

## 📜 Reviewers

@mscolnick


wasimsandhu:
@mscolnick One small issue I'm unsure about: I pass the given Python values for left- and right-frozen columns to the table's `initialState`, but this means that the table does not update if the Python code is changed.

Might just be tunnel vision, but I'm not seeing how to manage the table's column pinning state right now. I'll take another look later.

frontend/src/components/data-table/column-pinning/feature.ts
```diff
@@ -0,0 +1,85 @@
...
+  ColumnPinningState,
+  ColumnPinningTableState,
+  makeStateUpdater,
+  RowData,
+  Table,
+  TableFeature,
+  Updater,
+} from "@tanstack/react-table";
+
+export const ColumnPinningFeature: TableFeature = {
```
mscolnick, line 15: i think this is already built into `tanstack`. do you think it would be possible to re-use that?

1. either `__select__` is always pinned (setting it in the initial state)
2. keeping this table feature, but overriding the existing functions if necessary. not sure if its possible, but something like:
```ts
const originaPin = column.pin;
if (originaPin) {
   column.pin = () => {
      // do some custom things
       return originaPin();
   }
}
```


if its not possible, then this is ok, but ideally we can use the internals as much as possible

marimo/_plugins/ui/_impl/table.py
```diff
@@ -280,6 +285,32 @@ def __init__(
 
         field_types = self._manager.get_field_types()
 
+        # Validate frozen columns
+        if (
```
mscolnick, line 289: nice!

frontend/src/components/data-table/data-table.tsx
```diff
@@ -122,32 +141,111 @@ const DataTableInternal = <TData,>({
...
+      position: isPinned ? "sticky" : "relative",
+      zIndex: isPinned ? 1 : 0,
+      background: isPinned ? "white" : "transparent",
+      width: column.getSize(),
+    };
+  };
+
+  // Update column sizes in table state for column pinning offsets
+  // https://github.com/TanStack/table/discussions/3947#discussioncomment-9564867
+  const columnSizingHandler = (
```
mscolnick, line 170: nice find

frontend/src/components/data-table/column-pinning/feature.ts
wasimsandhu: You know, now that I'm looking at this, I'm not really sure why I implemented it myself. Lol

frontend/src/components/data-table/data-table.tsx
```diff
@@ -165,7 +263,11 @@ const DataTableInternal = <TData,>({
           />
         )}
         <Table>
-          <TableHeader>{renderHeader()}</TableHeader>
+          <TableHeader>
```
mscolnick, line 266: can you update (a few lines above

`<div className={cn(className || "rounded-md border")}>`
to 
`<div className={cn(className || 'rounded-md border overflow-hidden')}>`

otherwise the white background doesn't get clipped in the corners

frontend/src/components/data-table/column-pinning/feature.ts
wasimsandhu: removed the "custom" feature -- `ColumnPinning` from tanstack works like a charm

frontend/src/components/data-table/data-table.tsx
```diff
@@ -112,6 +121,15 @@ const DataTableInternal = <TData,>({
     onColumnFiltersChange: onFiltersChange,
     // selection
     onRowSelectionChange: onRowSelectionChange,
+    initialState: {
+      columnPinning: {
+        left:
+          freezeColumnsLeft === undefined
+            ? ["__select__"]
```
mscolnick, line 128: could we share this const `const SELECT_COLUMN_ID = "__select__"` somewhere?

frontend/src/components/data-table/data-table.tsx
```diff
@@ -112,6 +121,15 @@ const DataTableInternal = <TData,>({
     onColumnFiltersChange: onFiltersChange,
     // selection
     onRowSelectionChange: onRowSelectionChange,
+    initialState: {
```
mscolnick, line 124: regarding your question about shared state, 

i think you could do:

```
useEffect(() => {
    table.setState({columnPinning: {left: freezeColumnsLeft, freezeColumnsRight}})
}, [useDeepCompareMemoize([freezeColumnsLeft, freezeColumnsRight])]
```

im doing `useDeepCompareMemoize` since freezeColumnsLeft/freezeColumnsRight may change array instance, but the contents stay the same. its probably not that much more performant

wasimsandhu: Looks like it was tunnel vision -- I did something similar and it worked. Thanks!

frontend/src/components/data-table/data-table.tsx
wasimsandhu: I put it in `types.ts`, let me know if there's a better spot

mscolnick (dismissed):

mscolnick:
🚢

PR #2133, opened 2024-08-27
Update uv instructions

telenieko:
With recent `uv` you can use `uvx marimo ...` directly. No install step.

https://docs.astral.sh/uv/guides/tools/

This PR simply updates index.md to use those new features.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -28,7 +28,7 @@ pip install marimo && marimo tutorial intro
 :::{tab-item} install with uv
 
 ```bash
-uv pip install marimo && marimo tutorial intro
+uvx marimo tutorial intro
 ```
 
 :::


akshayka:
Thanks for the PR! Since marimo is both a CLI tool but also a library, it may be helpful to install marimo in each environment. At least that's how I've been thinking about it. Let me look at `uvx` a bit more closely.

telenieko:

akshayka:
Yea, unfortunately I'm not sure if `uvx marimo ...` would be useful since marimo needs to run in the environment in which it is installed.  I can see `uvx --with-requirements ... marimo ...` being useful however. I'm happy to be corrected or educated further.

telenieko:
> Thanks for the PR! Since marimo is both a CLI tool but also a library, it may be helpful to install marimo in each environment. At least that's how I've been thinking about it. Let me look at `uvx` a bit more closely.

`uvx` will install the tool in an isolated environment for the purpose of running it without a python project. `uv` will do so inside a project.

The advantge of `uvx`, specially on a Getting Started or someone just using the web interface is that there is zero artifacts: only your notebook .py files in the folder!

telenieko:
recheck

mscolnick:
could we make that clearer and support both? e.g.?

```python
# when in an existing virtual environment
uv pip install marimo && marimo tutorial intro
# running marimo in an isolated environment 
uvx marimo tutorial intro
```

mscolnick:
actually `uvx marimo tutorial intro` won't let you install packages (maybe a bug not setting the venv, or maybe intentional) 
so it's likely not a "getting started" instruction we should encourage

telenieko:
I think this is more complicated than I initially thout.
I have opened #2143 with a much broader scope for further discussion.

Issue #2137, opened 2024-08-27
Realtime Markdown Cell Result?


atticus-lv:
### Description

Lots of editors can render markdown directly, they may have sider panel to show the result ( vscode, Pycharm) or just make a toggle mode between the markdown code and the render ( Typora, ext)
So I think when the markdown cell is only use for description, it should have this feature for a quick edit feedback

### Suggested solution

When a markdown cell is pure text, the markdown cell should render the result when there is any changes. 
When detect python vars in it,  user should run it manually

wasimsandhu:
I like this idea, will pick it up if the team approves.

mscolnick:
hey @wasimsandhu, go for it!

mscolnick:
@wasimsandhu - we don't render f'strings using with the markdown editor anymore, those stay as python editors, so i think as long as the `languageAdapter` is the `MarkdownLanguage` adapter, we can just automatically call "run".

I think the place to look is `MarkdownLanguageAdapter.getExtension` and use `movementCallbacks: MovementCallbacks`, which is part of the `LanguageAdapter` interface.

and you can create a codemirror extension that onChange, just calls `movementCallbacks.onRun` 

(i think)

akshayka:
@atticus-lv , this will be available in the next release. Thanks for the feature request!

PR #2138, opened 2024-08-28
improvement: Realtime Markdown rendering

wasimsandhu:
## 📝 Summary

Completes #2137.

## 🔍 Description of Changes

realtime markdown

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


frontend/src/core/codemirror/cm.ts
```diff
@@ -87,6 +88,8 @@ export const setupCodeMirror = (opts: CodeMirrorSetupOpts): Extension[] => {
     // Cell editing
     cellMovementBundle(cellId, cellMovementCallbacks, hotkeys),
     cellCodeEditingBundle(cellId, cellCodeCallbacks, hotkeys),
+    // Markdown autorun
+    markdownAutoRunExtension(cellMovementCallbacks),
```
mscolnick, line 92: Would this be better in the MarkdownLanguageAdapter class? Better separation of concerns and no need to do to the if check

dmadisetti:
Nice! Might be worth seeing how this interacts with the VSCode plugin which already does this

mscolnick:
@dmadisetti - its essentially the exact same logic, but they won't overlap or effect each other.

akshayka:
@wasimsandhu , this is awesome!!

akshayka (commented):
I don't remember if we "debounce"/deduplicate run requests in the backend. I think we might not, in which case if the kernel were busy while you were typing in a markdown cell, it's possible that hundreds of run requests would get queued and then run serially.

Might be something to test -- do time.sleep(10) in one cell, then start typing a bunch of text in a markdown cell and see what happens. I can help improve the BE if this is indeed an issue.

wasimsandhu:
Ah that's a good point. Let me check it out and report back

mscolnick:
@wasimsandhu - its ok if the markdown is queued and doesn't update. but the worry is if after the sleep is done, will it run 30 requests (one for each char) through the kernel or 1. ideally its 1, if that is not the case, I will dedupe RunRequests in another PR.

wasimsandhu:
@mscolnick Ahhh gotcha. If it's alright, I'd love to take a stab at it

mscolnick:
Totally - go for it! Thanks

wasimsandhu:
Just finished implementing @akshayka's de-duplication logic only to realize run requests were not getting duplicated to begin with... the markdown cell is stale while other cells are running, and it does not autorun after the `time.sleep` cell finishes executing.

I guess I must have hallucinated it the first time I observed it? I was positive I saw tons of duplicated requests, but now I'm unable to reproduce it lol. Can someone test on their machine and confirm that requests are not being duplicated as the branch currently stands?

wasimsandhu:
EDIT: Please ignore my previous message... I just restarted the server and reproduced the request duplication. Feels like I just got gaslighted

wasimsandhu:
Looks like I misled myself. After trying some different things, I realized that:

1) multiple run requests are not being sent to the server after blocking cell runs
2) run requests were being sent to the server on *every* interaction with the cell (click, drag, etc.)

Updated extensions.ts so that a run only occurs if the cell's code was changed 😅

akshayka:
Sounds good. I will review tomorrow

mscolnick:
@wasimsandhu i just tested it. you are correct, it works as expected and we aren't sending additional requests

mscolnick (approved):
amazing! such a nice change

PR #1214, opened 2024-04-24
Tweaks for islands off Vite

dmadisetti:
Just a working branch of trying to get islands off vite, corresponding marimo-quarto branch is here

I haven't had time to solidly work on this, but hope that I can address/ fix-up some of the issues tonight:

todos:
 - [x] isIsland
 - [x] clean-up

Bugs:
 - [ ] Islands with multiple outputs currently do not work
 - [x] json outputs still a little weird
 - [x] Dedent breaks functions
example: 
```python
def fn(x):
    pass
```
Could handle quarto side with easy hack of prepending "#\n", but asking render not to dedent is probably a good idea too

Beyond my control:
 - [ ] the island cdn is wishful thinking? https://cdn.jsdelivr.net/npm/@marimo-team/islands@0.4.3/dist/main.js

Things I would like to see:
 - [ ] intermediate flags (cache / pickle definitions for use in non-intermediate cells)
 - [ ] Quarto side, implement execution-options: https://quarto.org/docs/computations/execution-options.html

PR #2156, opened 2024-08-29
fix: islands render_html() head needs marimo tags

akshayka:
HTML generated with `MarimoIslandGenerator.from_file().render_html()` was failing to render because the frontend expects marimo-filename and marimo-mode tags to be present.

### DIFF
--- a/marimo/_islands/island_generator.py
+++ b/marimo/_islands/island_generator.py
@@ -284,7 +284,7 @@ class MarimoIslandGenerator:
         self,
         *,
         version_override: str = __version__,
-        _development_url: Union[str | bool] = False,
+        _development_url: Union[str, bool] = False,
     ) -> str:
         """
         Render the header for the app.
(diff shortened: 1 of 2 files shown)


dmadisetti:
So, this is actually due to a failing islands check. I put this in as a specific tag and we decided to revert it. 

Relevant: https://github.com/marimo-team/marimo/pull/1214

Revert: https://github.com/marimo-team/marimo/pull/1250

mscolnick (approved):

mscolnick:
@dmadisetti i think your original solution works as well, and it is kinda nice adding this on the frontend than the backend so island consumers don't need to manually add these header tags if they arent using the python builder/geneartor

cc @akshayka

dmadisetti (approved):
Makes sense!

akshayka:
I see two approvals, so I'll merge this :)

PR #2161, opened 2024-08-29
fix: handle escaped table names in SQL statements

mscolnick:
The code changes in `sql_visitor.py` handle the case where SQL statements have escaped table names. This is done by checking if the statement starts with `e'` and adjusting the start and end indices accordingly. This ensures that the correct table name is extracted.


akshayka (approved):

dmadisetti:
Woops, didn't see this. Won't handle the case where a table name contains the quote, but that's a pretty far out edge case. Can always refer to my previous comment if it comes up

mscolnick:
@dmadisetti i tried your comment and it broke some of the more common edge-cases currently in the tests

mscolnick:
I could've copied it wrong - I'll definitely refer back to it if this comes up.

PR #2204, opened 2024-09-03
improvement: show mo.ui.table() first results from initial data

mscolnick:


pyproject.toml
```diff
@@ -136,7 +136,7 @@ testoptional = [
     "geopandas~=0.13.0",
     "matplotlib~=3.9.2; python_version > \"3.8\"",
     "pyarrow>=15.0.2,<18",
-    "pyarrow-stubs>=20240830",
```
mscolnick, line 139: some versions got yanked and then they changed their version schema

akshayka (approved):

PR #2209, opened 2024-09-04
fix: allow backslashes in app config (file format)

### LINKED ISSUE #2202: Custom.css not found
### Describe the bug

I added the sample custom.css provided in Theming from the documentation

I used the Application Config option in the settings and pasted the relative path as directed.

C:\vault\ForageReportingAndServices\custom.css

### Environment

marimo theme issue

### Code to reproduce

_No response_

akshayka:
Escape backslashes in app config strings.

Fixes #2202

### DIFF
--- a/marimo/_ast/codegen.py
+++ b/marimo/_ast/codegen.py
@@ -124,7 +124,9 @@ def generate_unparsable_cell(
 def generate_app_constructor(config: Optional[_AppConfig]) -> str:
     def _format_arg(arg: Any) -> str:
         if isinstance(arg, str):
-            return f'"{arg}"'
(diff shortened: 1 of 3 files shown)


mscolnick (approved):

mscolnick (approved):
nice I think escaping is better

mscolnick:
Do we need to do this for layouts? We put them in a subdirectory too, I just haven't see anyone have issues

akshayka:
> Do we need to do this for layouts? We put them in a subdirectory too, I just haven't see anyone have issues

This change escapes backslashes all string arguments, so it will apply to layouts too (if the layout has a backslash).

PR #2219, opened 2024-09-04
fix: pyarrow stubs

akshayka:


mscolnick:
It's been a moving target, they keep yanking new ones every day

mscolnick (dismissed):

PR #2241, opened 2024-09-05
chore: split up test workflow so they only run when changed

mscolnick:
split up test workflow so they only run when changed


akshayka (approved):
nice

PR #2242, opened 2024-09-05
improvement: use light-dark() css function for easier overrides

mscolnick:
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark

This makes it easier for consumers to override css variables and make it easier for devs to manage this. 
However it only reach all 80% of users: https://caniuse.com/?search=light-dark


akshayka (approved):

mscolnick:
going to hold off on this change, given `light-dark()` adoption

PR #2292, opened 2024-09-10
improvement: set maximum bar width with column summaries

mscolnick:
image


akshayka (commented):
I actually liked it before. Skinnier bars are nice but there shouldn't be whitespace in between the bars, otherwise the histogram is confusing at a glance.

akshayka:
> I actually liked it before. Skinnier bars are nice but there shouldn't be whitespace in between the bars, otherwise the histogram is confusing at a glance.

So put another way I'd approve this if there were no whitespace between bars

mscolnick:
@akshayka updated the image with less padding between

akshayka (approved):

PR #2295, opened 2024-09-11
fix: latex mime, exlclude text/plain, depcheck perf

mscolnick:
Some mime fixes:

- ignore `_repr_latex_`, this breaks pandas
- exlclude `text/plain` from mimebundle, since ibis includes this as well as html, and we dont need both
- depcheck: change `has()` to `imported()`
- bugfix for the empty value of `altair_chart`

### DIFF
--- a/marimo/_output/formatting.py
+++ b/marimo/_output/formatting.py
@@ -161,7 +161,9 @@ def get_formatter(
         ("_repr_png_", "image/png"),
         ("_repr_jpeg_", "image/jpeg"),
         ("_repr_markdown_", "text/markdown"),
-        ("_repr_latex_", "text/latex"),
+        # Commenting out _repr_latex_ for now, since pandas pd.Series
+        # uses it for some reason, but returns empty
+        # ("_repr_latex_", "text/latex"),
         ("_repr_text_", "text/plain"),  # last
     ]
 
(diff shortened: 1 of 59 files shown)


marimo/_output/formatting.py
```diff
@@ -161,7 +161,9 @@ def f_mime(obj: T) -> tuple[KnownMimeType, str]:
         ("_repr_png_", "image/png"),
         ("_repr_jpeg_", "image/jpeg"),
         ("_repr_markdown_", "text/markdown"),
-        ("_repr_latex_", "text/latex"),
+        # Commenting out _repr_latex_ for now, since pandas pd.Series
+        # uses it for some reason, but returns empty
+        # ("_repr_latex_", "text/latex"),
```
akshayka, line 166: @mscolnick Shouldn't our registered formatter be returned for `pd.Series`? Line 127 of this method? That is, I would have thought that `get_formatter()` would have already returned a registered formatter for `pd.Series`

mscolnick: yea they _should_, but i dont think they do. its possible they were not registered

mscolnick: ive narrowed it down to module reloading. i think the formatter gets lots or a new module of pandas fails an equality check

akshayka (approved):
LGTM, we can debug/fix the reloading issue in another PR

PR #2300, opened 2024-09-12
fix: pandas filtering for string n/a

### LINKED ISSUE #2299: Column-specific filtering includes null values
### Describe the bug

Filtering for a value in a column also includes the rows where the column value is null

### Environment

0.8.11

### Code to reproduce

Screenshot 2024-09-11 at 5 09 54 PM

mscolnick:
Fixes #2299 

Pandas decides to not filter null items, so this fixes the defaults


akshayka (approved):
Oh Pandas ... thanks for the fix

PR #2312, opened 2024-09-13
feat: support ibis native in mo.ui.table()

### LINKED ISSUE #2269: Limit number of rows in mo.ui.dataframe as preview for Ibis
### Description

I am having a big dataset and want to filter it with mo.ui.dataframe, then all *illions of datapoints are loaded into the table.

### Suggested solution

would it be possible to add a limit(head in pandas) feature to the class marimo.ui.dataframe, which loads only, if a value is given, number of rows as a preview. But the resulting dataframe should be without the limit.

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
This adds an `IbisTableManager` instead of using converting it to a a `PyArrowTableManager`, otherwise the ibis table is no longer lazy.

This also has a few fixes for `str(ibis)` which when interactive mode was enabled, would be very expensive

Also adds a `limit=` field to `mo.ui.dataframe`, which fixes #2269

### DIFF
--- a/marimo/_messaging/ops.py
+++ b/marimo/_messaging/ops.py
@@ -472,7 +472,7 @@ class VariableValue:
             # HACK: We pretty-print tables to avoid str(ibis_table)
             # which can be very slow when `ibis.options.interactive = True`
             table_manager = get_table_manager_or_none(value)
-            if table_manager:
+            if table_manager is not None:
                 return str(table_manager)
             else:
                 return str(value)[:50]
--- a/marimo/_messaging/ops.py
+++ b/marimo/_messaging/ops.py
@@ -39,6 +39,7 @@ from marimo._output.hypertext import Html
 from marimo._plugins.core.json_encoder import WebComponentEncoder
 from marimo._plugins.core.web_component import JSONType
 from marimo._plugins.ui._core.ui_element import UIElement
+from marimo._plugins.ui._impl.tables.utils import get_table_manager_or_none
 from marimo._runtime.context import get_context
 from marimo._runtime.layout.layout import LayoutConfig
 from marimo._utils.platform import is_pyodide, is_windows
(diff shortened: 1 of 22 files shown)


marimo/_messaging/ops.py
```diff
@@ -468,6 +469,14 @@ def __init__(
 
     def _stringify(self, value: object) -> str:
         try:
+            # HACK: We pretty-print tables to avoid str(ibis_table)
+            # which can be very slow when `ibis.options.interactive = True`
+            table_manager = get_table_manager_or_none(value)
+            if table_manager:
```
akshayka, line 475: Good to explicitly check for `None` as a rule. I picked up this habit [at Google](https://google.github.io/styleguide/pyguide.html#2144-decision) and it helps avoid bugs with `False`y objects.

```suggestion
            if table_manager is not None:
```

marimo/_messaging/ops.py
```diff
@@ -468,6 +469,14 @@ def __init__(
...
+                return str(table_manager)
+            else:
+                return str(value)[:50]
```
mscolnick, line 478: this is an important change. and would cause long load times for large ibis tables

marimo/_plugins/ui/_impl/tables/ibis_table.py
```diff
@@ -0,0 +1,219 @@
...
+
+            def get_field_types(self) -> FieldTypes:
+                return {
+                    column: IbisTableManager._get_field_type(self.data[column])
+                    for column in self.data.columns
+                }
+
+            def take(self, count: int, offset: int) -> IbisTableManager:
+                if count < 0:
+                    raise ValueError("Count must be a positive integer")
```
akshayka, line 96: Does count need to be positive (in which case the check should be `if count <=0`) or non-negative (in which case the check is fine)?

akshayka (commented):
Looks good, some tests failing though

PR #2316, opened 2024-09-15
feat: Account for state and variable type in save

dmadisetti:
#1993

CI didn't catch action added since PR creation

hopefully this retriggers CI?


dmadisetti:
Closing out this once since no difference and the other one has an approval

PR #2318, opened 2024-09-15
feat: @marimo.cache + tests and documentation

dmadisetti:
## 📝 Summary

Rebased and squashed my dev branch to add @mo.save.cache

Same as functools.cache but 3 nice differences:

1. Cache persists on cell rerun (by stuffing state into the registry)
2. Cache invalidation occurs on ref changes
3. Comments and formatting won't invalidate the cache

## 🔍 Description of Changes

Adds:
 - A new directory for loaders
 - "Memory Loader"
 - mo.save.cache and mo.save.lru_cache
 - Some light documentation (that can be expounded upon)

## 📜 Reviewers

@akshayka OR @mscolnick


marimo/_save/ast.py
```diff
@@ -140,3 +142,22 @@ def generic_visit(self, node: ast.AST) -> ast.AST:
...
+        expr.lineno = node.lineno
+        expr.col_offset = node.col_offset
+        return expr
+
+
+def strip_function(fn: Callable[..., Any]) -> ast.Module:
+    code, _ = inspect.getsourcelines(fn)
+    function_ast = ast.parse(textwrap.dedent("".join(code)))
+    body = function_ast.body.pop()
+    assert isinstance(body, ast.FunctionDef), "Expected a function definition"
```
dmadisetti, line 159: Might break on async, will check

marimo/_save/hash.py
```diff
@@ -604,8 +623,12 @@ def cache_attempt_from_hash(
         context=context,
         pin_modules=pin_modules,
         hash_type=hash_type,
+        scoped_refs=scoped_refs,
     )
 
+    if as_fn:
```
dmadisetti, line 629: Leave comment, that defs are local to the function, so should not save them

```
def fn()
    global x
```

Could be an issue, but is already an anti-pattern

marimo/_save/loaders/memory.py
```diff
@@ -0,0 +1,61 @@
...
+
+from marimo._save.cache import Cache, CacheType
+from marimo._save.loaders.loader import INCONSISTENT_CACHE_BOILER_PLATE, Loader
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+class MemoryLoader(Loader):
+    """In memory loader for serializable objects."""
```
dmadisetti, line 15: doesn't have to be serializable

marimo/_save/save.py
```diff
@@ -37,34 +40,235 @@
...
+            "Either @cache is not called from a top level cell or "
+            f"{UNEXPECTED_FAILURE_BOILERPLATE}"
+        )
+        self.graph = ctx.graph
+        self.cell_id = ctx.cell_id or ctx.execution_context.cell_id
+        self._args = list(self.fn.__code__.co_varnames)
+
+        self.module = strip_function(self.fn)
+        # frame is _set_context -> __call__ (or init) -> fn wrap
+        f_locals = inspect.stack()[2][0].f_locals
```
dmadisetti, line 143: technically, should get frames from 2 -> module

dmadisetti (commented):
Unsure why that test is failing - looks potentially flaky from my end. Will look into it tomorrow

mscolnick:
@dmadisetti We do have some flaky tests - I can take a look as well

docs/recipes.md
```diff
@@ -589,4 +589,31 @@ it has not seen, it will compute the predictions and store them in a cache. The
 next time it is called with the same parameters, instead of recomputing the
 predictions, it will return the previously computed value from the cache.
 
+Unlike `functools.cache`, `mo.cache` is notebook aware; and will become
+invalidated if a relevant global variable is changed (including State and UI
+Elements), or the function definition is changed.
```
akshayka, line 594: "or the function definition is changed" -- won't this be true for `functools.cache` too?

docs/recipes.md
```diff
@@ -589,4 +589,31 @@ it has not seen, it will compute the predictions and store them in a cache. The
...
+
 See our [best practices guide](guides/best_practices) to learn more.
+
+### Persistent caching for very expensivve computations
+
+**Use case.** If you are using marimo to capture very compute intensive
+results, you may want to save the state of your computations to disk. Ideally,
+if you update your code, then this save should be invalidated. It may also be
+advantageous to add UI elements to explore your results, without having to
+recompute expensive computations.
```
akshayka, line 604: Where does the cache get stored on disk? So users can clear it if needed.

We have a `.marimo` directory now, in which HTML snapshots get saved. Maybe it should be under there, if it isn't already?

docs/guides/best_practices/performance.md
```diff
@@ -1,15 +1,16 @@
 # Performance
 
-## Cache computations with `@functools.cache`
+## Cache computations with `@mo.save.cache`
 
-Use Python's builtin `functools` library to cache expensive computations.
-
-For example,
+You may use `mo.save.cache` to cache expensive computations, in a notebook aware manner.
```
akshayka, line 5: > You may use `mo.save.cache` to cache expensive computations, in a notebook aware manner.

What is "notebook aware"? I might rephrase as "You may use the decorator `mo.save.cache` to cache expensive computations. This works similarly to `functools.cache`, but also <invalidates the cache when UI elements, etc change>. 

Additionally, I have a few questions.

1. If my function didn't close over `State` or `UIElement` objects, but instead took their values as arguments:

```
def foo(state_value, ui_element_value):
   ...
```

would using `functools.cache` and `mo.save.cache` yield equivalent results?

2. When the arguments are the UI Element / State objects themselves, what happens?

```
@mo.save.cache
def foo(state_getter, slider):
  ...
```

Does the cache key on the state/slider values, or on their object identities?

3. Does `cache` work when running as a script.

docs/guides/best_practices/performance.md
```diff
@@ -19,6 +20,11 @@ it has not seen, it will compute the predictions and store them in a cache. The
 next time it is called with the same parameters, instead of recomputing the
 predictions, it will return the previously computed value from the cache.
 
+Alternatively, you may use Python's builtin `functools` library to cache
+expensive computations. It is important to note that `mo.save.cache` is more
+expensive than `functools.cache`, and there are certain instances where
+`functools.cache` is more appropriate.
```
akshayka, line 26: It's not clear to me as a user how to decide between using `mo.save.cache` and using `functools.cache`.

akshayka (commented):
I haven't reviewed in any detail yet, but I skimmed the documentation and had a few questions; left comments.

In general, I think it needs to be clearer to the user what `mo.save.cache`/`mo.save.persistent_cache` can safely cache, and what their limitations are, so that there's no guesswork involved or surprises incurred on the user's part.

It's a cool feature. The promise of notebooks that start instantly, not just outputs but with values already resident in memory, is exciting!

mscolnick:
wow this is going to be mind blowing 🤯

docs/guides/best_practices/performance.md
dmadisetti: Good suggestion, will try to fold it back into. For your Qs:

1. Taking in their values would be exactly the same
2. State is hashable so functools does work for this, but UI elements are not
To your point though
```python
fib(1)
s, _ = state(1)
fib(s) # This is a cache hit as is even though fib(s) would fail by itself
```
This fixable. But not, object identity is not utilized for the hash. This matters when restoring values though

3. Yep!

docs/recipes.md
dmadisetti: Good place to save, just in `.outputs` right now

docs/recipes.md
dmadisetti: "but not if the cell is rerun with no changes"

But will workshop this sentence

docs/recipes.md
akshayka: Actually playing with this, maybe it's good for the cache to not be in a hidden directory? Ie maybe something like your default is fine. Since some of the objects you'll be caching will be rather large. Or, maybe we take a page out of Python's book and choose `__mocache__` or `__marimo__`..

akshayka: Should the directory structure be

- `__marimo__` (or whatever it's called)
  - `<notebook_name>`
    - `<cache_name>`
    
? That is should we group caches belong to the same notebook in an intermediate directory?

akshayka:
@dmadisetti Is there at all a possibility for `persistent_cache` to work in the absence of a user-specified name for the cache? I can see a use case for using `persistent_cache` very liberally, in which case maybe I don't want to think about the names for each of my cell caches. Perhaps this is related to your branch in which you're experimenting with wrapping every cell in `persistent_cache`.

It's okay if this isn't readily done ... just curious.

dmadisetti:
@akshayka name is just used for file path / grouping. I found in terms of manual invalidation, it was nice to group all the blocks in a single place. I think it's a good frictional cost, since it's a chance to explain why the block is expensive

But yes! I think that a cache could be an optional cell-level default, and passed down to the executor (not great for a full execution mode since not everything can be pickled) . In those cases it can be done by the cell-id/name. I think in my experimental branch, I just hardcode these, with pickle-loader hard-coded as well, but maybe this should be opened up in settings:

 - cache_method: enum (Memory, Pickle, TODO: Dill, Numpy, network / remote)
 - cache_by_default: bool

If done on a cell level, I was thinking that maybe a floppy disk icon to denote a saved block, with a toggle in the cell options, and maybe a key binding to activate/ toggle.

docs/recipes.md
dmadisetti: Yeah, there's not a great mechanism for clearing out the cache, so a hidden folder might not be the move. I was thinking of putting in an LRU, but diskspace is relatively cheap now. However, there are still cases where you just want to clean up. No folder potentially needed for remote caching (or maybe optional/limited local LRU)

In the "execution mode" (or native save?) extension to this, I also save code content. If the graph is saved too, then this could be leveraged into a sophisticated check-pointing mechanism/audit (restore your notebook, output, state, and all the variables). Jupyter's `.ipynb_checkpoints` folder makes it clear that it's used for checkpoints, so I worry that `__mocache__` could be misleading if caching expands, and `__marimo__` is maybe not descriptive enough.

Haven't gotten back to review edits (maybe this evening), so plenty of time to think about this. I think all of them work, but once decided, ideally shouldn't be changed for backwards compat.

marimo/_save/save.py
dmadisetti: Adds additional complexity, and is an edge case. Just left comment.

docs/recipes.md
dmadisetti: How's

 - .marimo/
   - cache/
       - `<cache name>` / `<notebook name>`

akshayka: That sounds good to me.

Did you land on storing in the hidden `.marimo`? I'll leave it up to you; whatever you think is best.

akshayka (commented):
@dmadisetti I see you're still pushing fixes — let me know when you think the PR is ready to go, at which point I'll do an editing pass on the docs and merge it. It's a very cool feature.

One question — I have a slight preference for `mo.cache` over `mo.save.cache`, and similarly for the other functions. Less typing and leaves the name `mo.save()` open in case we ever want it. Do you agree, and if so want to rename them?

dmadisetti:
@akshayka Changes were primarily just to get tests passing. Looks like the cache invalidation might be a little nondeterministic on rerun? Was only seeing this when running the tests in bulk like CI does. I don't think it's a major issue, since still an improvement over functools.

I pushed my optimization change as well. I think that the complexity it adds might not be worth it- but there is a definite speed up. Optimization comes from reusing the execution path hash base, and applying the content hash after.
This means you need to segment out the content hash refs, run execution hash on the cells that content hash cannot account for, and then apply the content hash.

A few minor doc updates, but feel free to have a pass

akshayka (commented):
@dmadisetti is this a bug or expected?

image

```python
import marimo

__generated_with = "0.8.17"
app = marimo.App()


@app.cell
def __():
    import marimo as mo
    return (mo,)


@app.cell
def __():
    import functools
    return (functools,)


@app.cell
def __():
    class Namespace:
        ...

    ns = Namespace()
    ns.x = 0
    return Namespace, ns


@app.cell
def __(mo, ns):
    @mo.cache
    def f():
        return ns.x
    return (f,)


@app.cell
def __(f):
    f()
    return


if __name__ == "__main__":
    app.run()
```

akshayka:
@dmadisetti one more

image

dmadisetti:
Thanks! Will take a look and add unit tests. Think I may have introduced this with my "optimization"

dmadisetti:
@akshayka woops. it _was_ part of the spec, since objects are mutable and don't pass the "primitive" signatures that hashing currently does. The thrown errors were wrong though

Bit of feature creep, but added a fallback to pickle.
Also found a scope bug in script mode, that's why the change to `script_context`

The example you gave would work in functools but mutation to the object wouldn't bust the cache. With the push, `@mo.cache` will now cache bust on object mutation (there's a flag called `sensitive` on an the internal lookup function to toggle this behavior), and fail meaningfully if it can't pickle.

akshayka:
> The example you gave would work in functools but mutation to the object wouldn't bust the cache. With the push, `@mo.cache` will now cache bust on object mutation (there's a flag called `sensitive` on an the internal lookup function to toggle this behavior), and fail meaningfully if it can't pickle.

That means in comparison to functools.cache, mo.cache has the additional requirement that closed over objects must be pickleable.

Are there any other requirements that mo.cache has that functools.cache does not have? Asking because it's important to communicate these to the user.

dmadisetti:
functools does require arguments to be pickleable

`@mo.cache` only requires objects to be pickable when they are defined in the same cell as the function, or when they are an argument. The marimo thrown error suggests breaking up the code

OK case: test_execution_hash_diff_block

Bad case: test_execution_hash_same_block_fail

dmadisetti:
Correction. functools requires arguments to be hashable, not just picklable. so this fails:

```python
from functools import cache

@cache
def fn(l):
    return len(l)

fn([1,2,3]) # no good with functools, fine in mo
```

https://marimo.app/l/oyue1e

but would be fine with `@mo.cache`. Differing behavior comes in when there is implicit state. For instance:

```python
from functools import cache

@cache
def fn():
    l.append(1)
    return len(l)

l = [1,2,3]

print(fn(), fn(), fn(), fn())
# functools: 4 4 4 4
# mo.cache: 4 5 6 7 # More "correct" since l mutates
```

```python
from weakref import ref
@cache
def fn():
    l.append(1)
    return len(l)

l = [ref(1), ref(2), ref(3)]

print(fn(), fn(), fn(), fn())
# functools: 4 4 4 4
# mo.cache: error since l in same block and unpickable
```


```python
from weakref import ref
l = [ref(1), ref(2), ref(3)]

# Different Cell
@cache
def fn():
    l.append(1)
    return len(l)

print(fn(), fn(), fn(), fn())
# functools: 4 4 4 4
# mo.cache: 4 4 4 4
```

---

These cases all fall into anti-patterns, and may cause unexpected state changes in both cases

marimo/_save/hash.py
```diff
@@ -269,76 +321,209 @@ def __init__(
...
+            failed = []
+            exceptions = []
+            # By rights, could just fail here - but this final attempt should
+            # provide better user experience.
+            for ref in unhashable:
+                try:
+                    _hashed = pickle.dumps(scope[ref])
+                    content_serialization[ref] = type_sign(_hashed, "pickle")
+                    refs.remove(ref)
+                except TypeError as e:
```
akshayka, line 462:
```suggestion
                except (pickle.picklingError, TypeError) as e:
```

`pickle.dumps` can also raise `pickle.PicklingError`

image

marimo/_save/hash.py
```diff
@@ -269,76 +321,209 @@ def __init__(
...
+            failed = []
+            exceptions = []
+            # By rights, could just fail here - but this final attempt should
+            # provide better user experience.
+            for ref in unhashable:
+                try:
+                    _hashed = pickle.dumps(scope[ref])
+                    content_serialization[ref] = type_sign(_hashed, "pickle")
+                    refs.remove(ref)
+                except (pickle.picklingError, TypeError) as e:
```
akshayka, line 462:
```suggestion
                except (pickle.PicklingError, TypeError) as e:
```

my b, had a typo in my original suggestion

marimo/_save/save.py
```diff
@@ -37,35 +45,309 @@
...
+        attempt.update(scope, meta={"return": response})
+        self._loader().save_cache(attempt)
+        return response
+
+
+def cache(
+    _fn: Optional[Callable[..., Any]] = None,
+    *,
+    pin_modules: bool = False,
+    hash_type: str = "sha256",
```
akshayka, line 203: @dmadisetti okay if we omit `hash_type` from the public API for now? or is there a compelling reason to include?

akshayka (commented):
I am confused by the behavior in this example:

image

`s` is a string, `f` is a `mo.cache`-d function returning `len(x)`, `g` is a cached function returning a random int.

I call `f` and `g` on some value of `s`, and `g` caches the return value `586`.

Then, I change the value of `s` to a new value and call `f` and `g`. `f` registers a cache miss, as I would expect having been trained by `functools.cache`, where the cache is keyed on args. But `g` registers a cache hit and returns `586`. No matter how many times I change the value of `s`, `g` always registers a cache hit.

Is this a bug? Working as intended? Is there a simple way to explain the discrepancy in this behavior?

```python
import marimo

__generated_with = "0.8.22"


@app.cell
def __():
    import marimo as mo
    return (mo,)


@app.cell
def __():
    s = "hello"
    return (s,)


@app.cell
def __(mo):
    @mo.cache
    def f(x):
        return len(x)
    return (f,)


@app.cell
def __(mo):
    @mo.cache
    def g(x):
        import random
        return random.randint(0, 1000)
    return (g,)


@app.cell
def __(f, s):
    f(s)
    return


@app.cell
def __(g, s):
    g(s)
    return


if __name__ == "__main__":
    app.run()
```

marimo/_save/save.py
dmadisetti: Yep. I was checking for speedup against md5 etc. no noticable difference in my basic testing though, I don't think there's a meaningfully different choice

akshayka:
> I am confused by the behavior in this example:

Ah -- @dmadisetti , is this because the function body  of `g` doesn't read `x`?

dmadisetti:
Yes! That is correct, and probably not desirable. Quick fix, good catch

marimo/_save/save.py
```diff
@@ -30,42 +39,313 @@
...
         name: str,
+        *,
+        save_path: str = ".marimo/cache",
         pin_modules: bool = False,
         _loader: Optional[Loader] = None,
     ) -> None:
+        # For an implementation sibling regarding the block skipping, see
+        # `withhacks` in pypi.
         self.name = name
+        save_path = save_path
```
akshayka, line 348: @dmadisetti Did you want to update the `save_path` to append the notebook name? Mainly to make it easier for users to manage their caches.

If so, then the signature could be updated to:

`save_path: str | None = None`

and then `if save_path is None`, `save_path = ".marimo/cache/{notebook-name}/name"`

?

dmadisetti: Hmm. I was thinking about this, but:

1. There should never be a cache collision + the user can add the specificity if they want
2. The directory is still local to the notebook 
3. It allows speedup between notebooks

But 3. Might be hard to intentionally trigger, and the notebook name makes things easy to manually clean up. I think adding notebook name seems reasonable but:

 - What happens on notebook rename
 - is the only mechanism for cleanup manually deleting files?

docs/api/caching.md
```diff
@@ -7,16 +7,15 @@ and caching the values of variables to disk.
...
-2. `mo.cache` is invalidated when closed-over values change, while
-   `functools.cache` returns stale values;
-3. `mo.cache` does not require its arguments to be hashable;
-4. `mo.cache` is not invalidated by comments and code formatting.
+The resulting cache is similar to `functools.cache`, but with the benefit that
+[`mo.cache`](#marimo.cache) won't return stale values (because it keys on
+closed-over values) and isn't invalidated when the cell defining the decorated
+function is simply re-run (because it keys on notebook code). This means that
+like marimo notebooks, [`mo.cache`](#marimo.cache) has no hidden state
+associated with and make you more more productive while developing iteratively.
```
dmadisetti, line 18:
```suggested
associated with the memorized function, which makes you more productive while developing iteratively.
```

docs/api/caching.md
```diff
@@ -31,6 +30,11 @@ For a cache with bounded size, use [`mo.lru_cache`](#marimo.lru_cache).
 
 ## Caching variables to disk
 
+Use `mo.persistent_cache` to cache variables computed in an expensive block of
+code to disk. If marimo detects a cache on disk, the block of code will be
```
dmadisetti, line 34: * cache hit

marimo/_save/save.py
akshayka: > What happens on notebook rename

I think the user would have to manually rename the cache :/ 

> is the only mechanism for cleanup manually deleting files?

Did you have another idea in mind?

dmadisetti: > I think the user would have to manually rename the cache :/

Not specifying notebook is a quick way around that

> Did you have another idea in mind?

I think a remote LRU cache would be cool, but nothing locally

akshayka: > Not specifying notebook is a quick way around that

Yea -- i think it ends up being what do we value more:

1. Not having to manually rename the cache on notebook rename?
2. Making it easy to manually clear the cache for a specific notebook?

I was leaning toward 2 as being more valuable, since I'm assuming the user might be caching big objects. But I'm unsure. Thoughts?

dmadisetti: I think cell level caching should be under `<dir>`/`cache`/`<notebook name>` for sure.

`name` is required for persistent_cache, so `<dir>`/`cache`/`my_large_cache_object` or `<dir>`/`persistent_cache`/`my_large_cache_object` (to separate the 2 caches) seems reasonable to me. The user could always specify `name=notebook_name/large_cache`, which will create the sub directory.

akshayka: Fair enough! Keep as is then.

docs/guides/best_practices/performance.md
```diff
@@ -1,23 +1,76 @@
...
+:::
+
+## Save/load from disk with `mo.persistent_cache`
+
+Use `mo.persistent_cache` to cache variables to disk. The next time your
+run your notebook, the cached variables will be loaded from disk instead of
+being recomputed, letting you pick up where you left off.
+
+Reserve this for expensive computations that you would like to persist across
+notebook restarts. Cached outputs are automatically saved to `.marimo/cache`.
```
akshayka, line 55:
```suggestion
notebook restarts. Cached outputs are automatically saved to `__marimo__/cache`.
```

docs/api/caching.md
```diff
@@ -0,0 +1,40 @@
...
+Use [`mo.cache`](#marimo.cache) to cache the return values of functions in
+memory, based on the function arguments, closed-over values, and the notebook
+code defining the function.
+
+The resulting cache is similar to `functools.cache`, but with the benefit that
+[`mo.cache`](#marimo.cache) won't return stale values (because it keys on
+closed-over values) and isn't invalidated when the cell defining the decorated
+function is simply re-run (because it keys on notebook code). This means that
+like marimo notebooks, [`mo.cache`](#marimo.cache) has no hidden state
+associated with the memorized function, which makes you more productive while developing iteratively.
```
akshayka, line 18:
```suggestion
associated with the cached function, which makes you more productive while developing iteratively.
```

akshayka (approved):
# 💾 🚀🚀

PR #2322, opened 2024-09-15
tests: check schema drift from openapi schema

mscolnick:
Tests to check that marimo's backend is up-to-date with the open-api schema (minus the version)

### DIFF
--- a/openapi/api.yaml
+++ b/openapi/api.yaml
@@ -861,6 +861,13 @@ components:
             app_title:
               nullable: true
               type: string
+            auto_download:
+              items:
+                enum:
+                - html
+                - markdown
+                type: string
+              type: array
             css_file:
               nullable: true
               type: string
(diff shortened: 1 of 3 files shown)


openapi/api.yaml
```diff
@@ -876,6 +883,7 @@ components:
               type: string
           required:
           - width
+          - auto_download
```
dmadisetti, line 886: Oh cool, did it catch something?

dmadisetti (approved):
Nice! Are these going to go into the FE CI?

dmadisetti:
Oh ok. Saw the other PR. Kind of annoying to have to rely on a 3rd part action

openapi/api.yaml
mscolnick: yep it did! wasnt a bug, but was nice to validate it worked on the first try

mscolnick:
> Are these going to go into the FE CI?

These don't need to go in the frontend CI. If there are changes to this file, then the FE ci will kick off. and this forces the backend to make changes if they exist. So the contract, as expected, is this yaml file.

PR #2339, opened 2024-09-17
improvement: handle more output mime types from islands

mscolnick:


marimo/_islands/island_generator.py
```diff
@@ -97,7 +96,11 @@ def render(
         if not (display_code or display_output or is_reactive):
             raise ValueError("You must include either code or output")
 
-        output = handle_mimetypes(self.output) if self.output else None
+        output = (
+            mime_to_html(self.output.mimetype, self.output.data)
+            if self.output
```
akshayka, line 101: @mscolnick I assume this is supposed to be an explicit None check `if self.output is not None`

akshayka (approved):

marimo/_islands/island_generator.py
mscolnick: it doesn't need to be since this is an object (not str or int) - we probably could push this into the typechecker to avoid bike-shedding. 
our eslint rules can handle this properly (when type-checking is included).

PR #2342, opened 2024-09-18
feat: mo.show_code(output)

### LINKED ISSUE #2038: App configuration: show code
### Description

Following @odanoburu's suggestion in [#1975](https://github.com/marimo-team/marimo/issues/1975) that slides would be significantly improved if code (or some code) were allowed to be displayed/visible in the presentation, I would make a similar point for the App version itself. 

Having the code visible in the Notebook version but not in the App version is almost indifferent to the one who wrote the notebook. However, if the notebook is written with someone else in mind who will interact with it during a particular task, whether the code is visible or not in the App version becomes a critical point. One crucial characteristic of computational notebooks that led to their monumental success over the last decade is, I believe, the option available to hide/display specific parts of the code. 

To the best of my knowledge, all notebooks available allow you to hide/display specific parts of the code, even in a truly app view. I am providing a good example. Consider Hex notebooks, which target a segment of users for whom the App version seems more important than the notebook version itself.  However, in Hex notebooks, the App version allows the inclusion of visible code, as well as code that is not visible.  The first image below displays the Notebook version of an example on their website, while the second one shows the App version of the same notebook. Notice that in the App version, one code cell is hidden while another is displayed. 

---
Notebook view:

a

---
App view:

aa

---

If this option to hide/display code in the app version is implemented in marimo, the scope of adopters of this excellent piece of software will drastically increase. 

### Suggested solution

Allow the option to hide/display code in the app version is implemented in marimo

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #2313: Markdown cells, code cells, and smart cells: taking inspiration from Livebook
### Description

Right now, Marimo has essentially two kinds of cell: code cells and Markdown cells. (Although I do appreciate that Markdown cells are secretly just code cells featuring only a `mo.md` call.) Markdown cells are pretty straightforward, but code cells live a kind of double life: sometimes they're about building a UI application via `marimo.ui.*` and friends. Sometimes they're about displaying the actual code itself!

Right now we have to make a global choice about which of these two things is most important, across the entire notebook, based on whether we `--include-code` or not.

I'm proposing to take inspiration from [Livebook](https://livebook.dev) (from the Elixir/Erlang community), which acknowledges this difference by explicitly having three kinds of cell: Markdown cells, code cells, and 'smart' cells, the last of which are mini programmatic UIs wrapping a code cell -- that is, exactly like what Marimo provides. And most importantly... all three of these cells types can be mix-and-match'd in any way.

i'd love to have that approach in Marimo as well.

(In fact Livebook seems to have quite a lot of nice ideas floating around in it, so I suspect there may be other points worth taking inspiration from it as well.)

### Suggested solution

-

### Alternative

_No response_

### Additional context

_No response_

akshayka:
This PR adds `mo.show_code()`, which can be used to embed the cell's code into the output, making it visible in all app views.

image

This is an alternative to source visibility as a cell config, and works in vertical layout, slides, grid.

The output position is configurable (below or above code) via a keyword argument.

Closes #2038
Closes #2313

### DIFF
--- a/marimo/_output/with_code.py
+++ b/marimo/_output/with_code.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
+
 import re
 
 from marimo._output.formatting import as_html
 from marimo._output.hypertext import Html
 from marimo._output.md import md
+from marimo._plugins.ui._impl.input import code_editor
 from marimo._runtime.context import get_context
 from marimo._runtime.context.types import ContextNotInitializedError
-from marimo._plugins.ui._impl.input import code_editor
 
 
 def with_code(output: object = None) -> Html:
--- a/marimo/_output/with_code.py
+++ b/marimo/_output/with_code.py
@@ -1,28 +1,55 @@
 from __future__ import annotations
-import ast
+import re
 
 from marimo._output.formatting import as_html
 from marimo._output.hypertext import Html
 from marimo._output.md import md
 from marimo._runtime.context import get_context
 from marimo._runtime.context.types import ContextNotInitializedError
+from marimo._plugins.ui._impl.input import code_editor
 
 
-class ReplaceWithCodeCalls(ast.NodeTransformer):
+def with_code(output: object = None) -> Html:
+    """Display an output above the code of the current cell.
 
-    def visit_Call(self, node: ast.Call) -> ast.Expression | ast.Call:
-        if (
-            isinstance(node.func, ast.Attribute)
-            and isinstance(node.func.value, ast.Name)
-            and node.func.value.id == "mo"
-            and node.func.attr == "with_code"
-            and len(node.args) > 0
-        ):
-            return ast.Expression(node.args[0])
-        return node
+    Use `mo.with_code` to show the code of the current cell below
+    the cell's output. This is useful if you want a cell's code to
+    appear in the app preview or when running the notebook as an app
+    with `marimo run`.
 
+    In the displayed code, all occurrences of mo.with_code(...) will be
+    replaced with ...
 
-def with_code(output: object) -> Html:
+    Show code that produces the output `factorial(5)`:
+
+    ```python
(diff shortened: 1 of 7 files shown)


marimo/_output/with_code.py
```diff
@@ -0,0 +1,76 @@
...
+    cell = context.graph.cells[cell_id]
+    pattern = r"mo\.with_code\((.*?)\)"
+    code = re.sub(pattern, r"\1", cell.code, flags=re.DOTALL).strip()
+
+    if output is not None:
+        return md(
+            f"""
+{as_html(output)}
+
+{code_editor(value=code, disabled=True, min_height=1)}
```
mscolnick, line 72: can we put the html below the code?

mscolnick (approved):

marimo/_output/with_code.py
```diff
@@ -0,0 +1,76 @@
...
+    ```
+
+    Show code of a cell, without an output:
+
+    ```python
+    def factorial(n: int) -> int:
+        if n == 0:
+            return 1
+        return n * factorial(n - 1)
+    mo.show_code()
```
mscolnick, line 42:
```suggestion
    mo.with_code()
```

marimo/_output/with_code.py
```diff
@@ -0,0 +1,76 @@
...
+
+from marimo._output.formatting import as_html
+from marimo._output.hypertext import Html
+from marimo._output.md import md
+from marimo._plugins.ui._impl.input import code_editor
+from marimo._runtime.context import get_context
+from marimo._runtime.context.types import ContextNotInitializedError
+
+
+def with_code(output: object = None) -> Html:
```
mscolnick, line 13: add to `docs/`

mscolnick (approved):

PR #2351, opened 2024-09-18
chore: format tuple in return to match ruff/black

akshayka:
formats single-element tuples in return statements to match ruff/black to prevent unnecessary git diffs for ruff/black users


mscolnick (dismissed):

mscolnick:
this is a good idea - being more easily compatible with linter/formatter. and i am sure we can do more of this

PR #2358, opened 2024-09-19
fix: encoding table elements when passed through RPC calls

mscolnick:
`asdict` does not encode our UI elements properly and should be used sparingly


akshayka (dismissed):
good catch

PR #2361, opened 2024-09-20
poc: cell divider

mscolnick:


akshayka:
Played around, and I prefer the buttons as they are. It's intuitive as it is and out of the way of the main array body.

PR #2367, opened 2024-09-20
feat: xdg config file

pakelley:
## 🔍 Description of Changes
This allows a user to use a config file that follows the [XDG](https://xdgbasedirectoryspecification.com/) convention (i.e. the config file is located at `~/.config/marimo/marimo.toml`).

This ended up a bit clunky so that I could make a clean test. Not my favorite to messy up the code for the sake of a test though, so if it seems too messy lmk and I can put a bit more time into cleaning that up.

Also, I looked to see if there was any discussion about not wanting to support something like this any didn't find anything, but apologies if I missed something.

Thanks for your work on marimo 🙂 

## 📋 Checklist

(I'm assuming this isn't big enough to warrant this, but lmk)
(I tested that `get_config_path` can find a config file at `~/.config/marimo/marimo.toml` now, but can do some more thorough testing later today.)

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -19,6 +19,7 @@ User Configuration applies globally across all marimo notebooks and is stored in
 1. Current directory
 2. Parent directories (moving up the tree)
 3. Home directory (`~/.marimo.toml`)
+4. [XDG](https://xdgbasedirectoryspecification.com/) directory (`~/.config/marimo/marimo.toml` or `$XDG_CONFIG_HOME/marimo/marimo.toml`)
 
 If no `.marimo.toml` file is found, marimo creates a default one in your home directory.
 
--- a/docs/guides/editor_features/ai_completion.md
+++ b/docs/guides/editor_features/ai_completion.md
@@ -27,7 +27,7 @@ marimo using `pip` if you need Copilot.
 4. Right-click on the extension window and select "Inspect" to open the developer tools for the extension. Then click on "Network"
 5. Copy the token and paste it into the input area, and then press "Enter Token"
 6. This action will log a new API request in the **Network** tab. Click on "Preview" to get the API key.
-7. Paste the API key in the marimo settings in the UI, or add it to your `~/.marimo.toml` file as follows:
+7. Paste the API key in the marimo settings in the UI, or add it to your `marimo.toml` file as follows:
 
 ```toml
 [completion]
@@ -57,7 +57,7 @@ the bottom of your notebook. You can also refactor existing cells by inputting
 
 1. Install openai: `pip install openai`
 
-2. Add the following to your `~/.marimo.toml`:
+2. Add the following to your `marimo.toml`:
 
 ```toml
 [ai.open_ai]
@@ -74,7 +74,7 @@ base_url = "https://api.openai.com/v1"
 To use Anthropic with marimo:
 
 1. Sign up for an account at [Anthropic](https://console.anthropic.com/) and grab your [Anthropic Key](https://console.anthropic.com/settings/keys).
-2. Add the following to your `~/.marimo.toml`:
+2. Add the following to your `marimo.toml`:
 
 ```toml
 [ai.open_ai]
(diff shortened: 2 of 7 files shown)


mscolnick:
hey @pakelley, thank you! yes we have wanted this for a while and there were some discussions in Discord. 

i think we would like to move way from the home dir if possible. i wonder if we can include in this PR:

* If none are found, we create a new one in the home dir. maybe we can instead create a new one in `~/.config/`?
* Could we deprecate the home dir as well? maybe keep it readable in the same priority, but i wonder if we can mv it from there to `~/.config/`? probably a few edge cases, so maybe not worth it.

FYI we store other things in `~/.marimo` as well i think last the last time we checked for an update. im not sure that is config, so that can state there. 

@baggiponte had some thoughts on this too

pakelley:
Hey yeah, happy to play around with your suggestions. 
And yeah, I noticed `~/.marimo`, I considered moving it inside of `~/.local/state` but got a bit shy about it haha. Also, seems like the recent_files.toml` there might be better categorized as "cache" entries for `~/.cache`? 
But splitting that out is definitely starting to feel over my head... happy to take a swing at it though.

Lmk what sounds best for the `.marimo` dir and I'm happy to see what I can do there 👍

mscolnick:
Oh i didn't know of the `.local` and `.cache` folders. I kinda prefer `~/.marimo` for locality, but I understand the standards in place (although still many of my programs dont seem to follow them). 

I think we can leave `~/.marimo` for now, and consider moving to `.local` and `.cache` later.

pakelley:
Ok, I made the process for creating a new config use XDG, but was thinking it might actually be pretty confusing to people if we move their existing config? Idk, maybe this is still what you want, but as I started implementing it it felt sort of weird.
Lmk how everything looks!

pakelley:

mscolnick:
yea we don’t need to move the config. I always think about migrations in two parts: “stop the bleeding” and “heal the wounds”. I think this is great in stopping the bleeding and we can think of a better migration path - maybe when we do a 1.0 and we migrate other “breaks”

mscolnick (approved):
Thanks for the contribution!

pakelley:
Also, fwiw, I've tested (using `marimo -d edit --no-token`) that this:
1. saves to `~/.config/marimo/marimo.toml` when no config exists
2. saves to `~/.marimo.toml` when it already exists
3. saves to `~/.config/marimo/marimo.toml` when it exists (and `~/.marimo.toml` doesn't exist)
4. loads from `~/.marimo.toml` when it exists
5. loads from `~/.config/marimo/marimo.toml` when it exists (and `~/.marimo.toml` doesn't exist)

I think my last question would be whether it makes sense for the XDG paths to take precedence over the home directory ones?
Other than that, as long as this looks good to you, I think it's ready 🙂

mscolnick:
I think it will regardless since we will crawl up the file tree.

I guess in that case- we can take precedence in the .config directory. 

Also seems like the renamed import didn't happen everywhere

baggiponte:
Ops sorry for the late reply and thanks for the RFC.

I never truly recall the differences between the various XDG_DATA/STATE/CACHE envs.

If marimo does not store a lot of stuff, I guess everything can be kept in one directory and the end user can just add the other files to their gitignore (people usually version their .config folder, or use more complex stuff like stow/chezmoi to manage dotfiles).

As per the implementation, I guess the flow could've been:

1. Check if XDG_CONFIG_HOME exists
2. Check if ~/.marimo exists
3. If neither exists, use XDG_CONFIG_HOME

But I think it's a minor detail.

For a future improvement, there could be a small migration script (or just emitting a logging message). I am still a bit tight but I think I'll have time from the week after the next one to work on implementations. Feel free to ping me whenever for reviews! 😊

mscolnick:
Thanks @baggiponte! that's what we have for now and punting the migration until a larger marimo migration (possible other things).

pakelley:
Ah, yep! I missed adding a file in my last changes, my bad 😅

pakelley:
Updated some of the docs/dialogs to say just `marimo.toml` instead of `~/.marimo.toml`... lmk if you have a preference here, but I was poking around and noticed them, and figured this is the least-confusing phrasing. Happy to change those to whatever seems most intuitive to you all though.

mscolnick:
Those updates are perfect

mscolnick:
Going to merge this since looks like a different dep is causing typing issues.

PR #2385, opened 2024-09-23
feat: support installing packages at particular versions in the UI

mscolnick:
Screenshot 2024-09-23 at 10 13 50 AM


akshayka (approved):
Awesome! Pixi gets packages from anaconda I think, which differ from PyPI, so perhaps we don't show package versions for pixi. Otherwise LGTM

akshayka (approved):

PR #2390, opened 2024-09-23
fix inserting tabs

akshayka:
insert a tab if the previous character is a whitespace.

not working reliably -- sometimes only a single space is inserted

### DIFF
--- a/frontend/src/core/codemirror/cm.ts
+++ b/frontend/src/core/codemirror/cm.ts
@@ -10,6 +10,7 @@ import {
   historyKeymap,
   indentWithTab,
   indentMore,
+  insertTab,
 } from "@codemirror/commands";
 import {
   bracketMatching,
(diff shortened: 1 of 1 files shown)


frontend/src/core/codemirror/cm.ts
```diff
@@ -94,17 +95,20 @@ export const setupCodeMirror = (opts: CodeMirrorSetupOpts): Extension[] => {
   ];
 };
 
-const startCompletionAtEndOfLine = (cm: EditorView): boolean => {
+const insertTabOrComplete = (cm: EditorView): boolean => {
```
akshayka, line 98: This starts a completion only if the previous character is not whitespace. Otherwise it inserts a tab.

However, when testing, it seems like sometimes one or two spaces are inserted instead of the full indentation unit, which we set to four spaces. Not sure why ...

PR #2392, opened 2024-09-23
feat: panel for packagemanagement

mscolnick:
image


akshayka:
- [ ] use no confirm options for removing packages when available
- [ ] make reactive on backend — mark cells using removed module as stale

akshayka (approved):

PR #2429, opened 2024-09-27
improvement: remove bounds on number, add warnings and validation

### LINKED ISSUE #2426: mo.ui.number with infinite bounds set values to None when changed
### Describe the bug

I want to create a number component, but I dont want to set an upper or lower limit, so I simply used numpys infinity representation. Unfortunately, if I do that and I change the values using the up or down arrows, the value of the number field is set to None. I would expect that I could set open limits like this, where I don't want to limit the user.

### Environment

{
  "marimo": "0.8.20",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.2",
  "Binaries": {
    "Browser": "125.0.6422.142",
    "Node": "v20.14.0"
  },
  "Dependencies": {
    "click": "8.1.7",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "pygments": "2.17.2",
    "pymdown-extensions": "10.7.1",
    "ruff": "0.5.2",
    "starlette": "0.37.2",
    "tomlkit": "0.12.4",
    "typing-extensions": "4.9.0",
    "uvicorn": "0.29.0",
    "websockets": "11.0.3"
  },
  "Optional Dependencies": {
    "pandas": "2.2.2"
  }
}

### Code to reproduce

Here's a code that creates two number objects. The first uses very large values (integer) manually set (+/- 1e255) to set the boundaries. This works if you change the value using the arrows. The second uses numpy s "inf" value is the floating point representation of infinity as pr IEEE754.

```python

import marimo

__generated_with = "0.8.20"


@app.cell
def __():
    import marimo as mo
    import numpy as np
    return mo, np


@app.cell
def __(mo):
    def _on_change(new_value):
        print(new_value)
    number = mo.ui.number(start=-1e255, stop=1e255, value=5, on_change=_on_change)
    number
    return (number,)


@app.cell
def __(mo, np):
    def _on_change(new_value):
        print(new_value)
    number2 = mo.ui.number(start=-np.inf, stop=np.inf, value=5, on_change=_on_change)
    number2
    return (number2,)


if __name__ == "__main__":
    app.run()

mscolnick:
Fixes #2426

### DIFF
--- a/frontend/src/plugins/impl/NumberPlugin.tsx
+++ b/frontend/src/plugins/impl/NumberPlugin.tsx
@@ -11,28 +11,28 @@ import { NumberField } from "@/components/ui/number-field";
 type T = number;
 
 interface Data {
-  start: T;
-  stop: T;
+  start?: T | null;
+  stop?: T | null;
   step?: T;
   label: string | null;
   debounce: boolean;
   fullWidth: boolean;
 }
 
-export class NumberPlugin implements IPlugin<T, Data> {
+export class NumberPlugin implements IPlugin<T | null, Data> {
   tagName = "marimo-number";
 
   validator = z.object({
-    initialValue: z.number(),
+    initialValue: z.number().nullish(),
     label: z.string().nullable(),
-    start: z.number(),
-    stop: z.number(),
+    start: z.number().nullish(),
+    stop: z.number().nullish(),
     step: z.number().optional(),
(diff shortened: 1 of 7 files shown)


mscolnick:
merging to fix main

akshayka (approved):
v nice

PR #2445, opened 2024-09-28
fix: in sandbox, skip sandbox prompt if possible

akshayka:
This PR adds a new option, `--skip-sandbox-prompt`, to prevent running a sandboxed notebook to forever prompt the user to run with sandbox.

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -69,6 +69,9 @@ def _read_pyproject(script: str) -> Dict[str, Any] | None:
 
 
 def prompt_run_in_sandbox(name: str | None) -> bool:
+    if GLOBAL_SETTINGS.MANAGE_SCRIPT_METADATA:
+        return False
+
     if name is None:
         return False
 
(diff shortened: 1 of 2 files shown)


marimo/_cli/sandbox.py
```diff
@@ -140,7 +146,14 @@ def run_in_sandbox(
...
         temp_file_path,
-    ] + cmd
+    ]
+
+    marimo_help_cmd = uv_run + ["marimo", "edit", "--help"]
+    p = subprocess.run(marimo_help_cmd, capture_output=True)
+    if "--skip-sandbox-prompt" in p.stdout.decode():
+        cmd += ["--skip-sandbox-prompt"]
+
+    cmd = uv_run + cmd
```
akshayka, line 156: This is the best I could come up with to check if the version of marimo that uv will load has the `--skip-sandbox-prompt`. It does mean that we create an isolated environment twice, however ... @mscolnick let me know if you have other ideas

mscolnick (commented):
I think we should use the env that is already set to skip the prompt. We don't need to prompt if we are already in a sandbox

mscolnick (commented):
We should also fix the marimo version in requirements.txt to be the current. Not the future, since our args may not always be backwards or forwards compatible

marimo/_cli/sandbox.py
```diff
@@ -69,6 +69,9 @@ def _read_pyproject(script: str) -> Dict[str, Any] | None:
 
 
 def prompt_run_in_sandbox(name: str | None) -> bool:
+    if GLOBAL_SETTINGS.MANAGE_SCRIPT_METADATA:
```
mscolnick, line 72: We should make sure people install the version of marimo they are running and not the latest version

mscolnick (approved):

PR #2446, opened 2024-09-29
sandbox mode supports named and editable local packages, related dev …

anthonywu:
## 📝 Summary

These scenarios are common in work settings where private packages are running off of `git clone`s of teammates' branches, or packages are distributed as-is via private CI to prod without publishing to PyPi.

1. Support local named packages via `pyproject.toml`'s `package @ some/local/path` spec
1. Support installation of local editable packages (libraries in development, or otherwise not available in PyPi index for any reason) via the use of `uv run`'s `--with-editable` flag.

```sh
Usage: uv run [OPTIONS] <COMMAND>

Options:
      --with-editable <WITH_EDITABLE>          Run with the given packages installed as
```

3. dev ux improvement: when running `make` targets, if expected tools are missing because user skipped reading `CONTRIBUTING.md`, print helpful message about expectations and remediation commands.

## 🔍 Description of Changes

- `marimo/_cli/sandbox.py` is edited to parse for local paths
- `tests/_cli/sandbox.py` adds test to cover all the path parsing and alterations
- `Makefile` gets new `expect-tools` phony target for `pnpm`, `pytest`, `typos`, and added as a target dependency to `make fe` and `make py-test`
- `CONTRIBUTING.md` gets mentions of dev requirements and `typos-cli` install

## 📋 Checklist

### known issues?

There are 4 tests far away from my change that are not passing, and I think either I didn't set the right flags or these are known issues.

```
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_install - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_uninstall - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_list_packages - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_server/api/endpoints/test_ws.py::test_query_params - OSError: [Errno 24] Too many open files
```

## 📜 Reviewers

for review by @akshayka - who implemented the sandbox just days ago

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,24 @@ help:
 	@# https://stackoverflow.com/a/35730928
 	@awk '/^#/{c=substr($$0,3);next}c&&/^[[:alpha:]][[:alnum:]_-]+:/{print substr($$1,1,index($$1,":")),c}1{c=0}' Makefile | column -s: -t
 
+.PHONY: required-build-tools
+expect-tools:
+	@if ! /usr/bin/which -s pnpm typos pytest2; then \
+		echo "📦 pnpm, pytest, typos-cli required to build and test."; \
+		if ! /usr/bin/which -s pnpm; then \
+		    echo "💡 Please perform 'npm install -g pnpm@8' or use your preferred package manager."; \
+        fi; \
+		if ! /usr/bin/which -s pytest; then \
+		    echo "💡 Please perform 'pip install -e .[dev]' to add pytest and other dev requirements."; \
+        fi; \
+		if ! /usr/bin/which -s typos; then \
+		  echo "💡 Please perform 'brew install typos-cli', 'cargo install typos-cli' or use your preferred package manager."; \
+	    fi; \
+	fi
+
 # package frontend into marimo/
 .PHONY: fe
-fe: marimo/_static marimo/_lsp
+fe: expect-tools marimo/_static marimo/_lsp
 
 # install/build frontend if anything under frontend/src or (top-level)
 # frontend/ has changed
@@ -68,7 +83,7 @@ py-check:
 
 .PHONY: py-test
 # test python
-py-test:
+py-test: expect-tools
 	cd marimo && typos && cd - && pytest;
 
 .PHONY: py-snapshots
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -29,7 +29,10 @@ def _get_dependencies(script: str) -> List[str] | None:
     """Get dependencies from string representation of script"""
     try:
         pyproject = _read_pyproject(script) or {}
-        return cast(List[str], pyproject.get("dependencies", []))
+        return [
+            _.strip()
+            for _ in cast(List[str], pyproject.get("dependencies", []))
+        ]
     except Exception as e:
         LOGGER.warning(f"Failed to parse dependencies: {e}")
         return None
@@ -44,6 +47,14 @@ def get_dependencies_from_filename(name: str) -> List[str]:
     except Exception:
         LOGGER.warning(f"Failed to read dependencies from {name}")
         return []
+
+
+def expand_user_home(match) -> str:
+    """
+    Replaces dependencies that contain ~ that represents the home dir
+    returns a replacement depending on the dep prefix
+    """
+    return f"{match.group(1)} {os.path.expanduser(match.group(2))}"
 
 
 def _read_pyproject(script: str) -> Dict[str, Any] | None:
@@ -63,6 +74,9 @@ def _read_pyproject(script: str) -> Dict[str, Any] | None:
             line[2:] if line.startswith("# ") else line[1:]
             for line in matches[0].group("content").splitlines(keepends=True)
         )
+        # gracefully handle paths with '@ $path' to abspaths
+        # to meet pyproject expectation
+        content = re.sub(r"(@|--editable|-e)\s+(~)", expand_user_home, content)
         import tomlkit
 
         return tomlkit.parse(content)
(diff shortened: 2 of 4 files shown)


anthonywu:

anthonywu:
this needs more work, got excited too soon, after startup and the browser launches, there is some later step in `pypi_package_manager.py` and downstream `uv add|remove` that tries to parse the unmodified script header again, and runs into expectation issues. 

```
        Running local_editable_demo.py ⚡

        URL: http://localhost:2730

error: TOML parse error at line 2, column 16
  |
2 | dependencies = [
  |                ^
Expected package name starting with an alphanumeric character, found `-`
```

I plan to cut scope and remove the auto `os.path.expanduser` part, I think worthwhile trying to get `--editable` to work with abs paths.

Makefile
```diff
@@ -5,9 +5,24 @@ help:
 	@# https://stackoverflow.com/a/35730928
 	@awk '/^#/{c=substr($$0,3);next}c&&/^[[:alpha:]][[:alnum:]_-]+:/{print substr($$1,1,index($$1,":")),c}1{c=0}' Makefile | column -s: -t
 
+.PHONY: required-build-tools
+expect-tools:
```
mscolnick, line 9: nice!

is this compatible with windows? or could we skip if windows

marimo/_cli/sandbox.py
```diff
@@ -63,6 +74,9 @@ def _read_pyproject(script: str) -> Dict[str, Any] | None:
             line[2:] if line.startswith("# ") else line[1:]
             for line in matches[0].group("content").splitlines(keepends=True)
         )
+        # gracefully handle paths with '@ $path' to abspaths
+        # to meet pyproject expectation
+        content = re.sub(r"(@|--editable|-e)\s+(~)", expand_user_home, content)
```
mscolnick, line 79: does `uv` not handle the home expansion?

tests/_cli/test_sandbox.py
```diff
@@ -37,3 +39,35 @@ def test_no_dependencies():
...
+    SCRIPT = """
+# Copyright 2024 Marimo. All rights reserved.
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "pandas @ ~/local/path1",
+#     "numpy @ /local/abs/path2",
+#     "polars",
+#     "-e ~/foo/package1",
+#     "--editable ~/bar/package2",
```
mscolnick, line 54: i realized trying to install `pandas @ ~/local/path1` and `--editable ~/bar/package2` in the package management UI may break since adding: https://github.com/marimo-team/marimo/pull/2431

marimo/_cli/sandbox.py
```diff
@@ -144,8 +182,14 @@ def run_in_sandbox(
         # which may conflict with the sandbox requirements
         "--no-project",
         "--with-requirements",
-        temp_file_path,
-    ] + cmd
+        dep_temp_file_path,
+    ]
+
+    if local_editable_packages:
+        cmd.append("--with-editable")
```
mscolnick, line 189: whats the different between passing these to `--with-editable` vs putting them in the requirements.txt as `-e /path/to/editable`

tests/_cli/test_sandbox.py
mscolnick: Fixed in https://github.com/marimo-team/marimo/pull/2448

Makefile
anthonywu: You bring up a good point, do you expect contributors to dev/test outside of Darwin/Linux? Windows devs can always use WSL. The make helper can probably won't work outside of Mac/Linux given "which" is a requirement for checking existence of a tool. We can gate the check for Mac/Linux host and leave windows users to figure it out themselves.

marimo/_cli/sandbox.py
anthonywu: It's pyproject toml schema that doesn't allow home expansion. Try it on main and you'll see the path being rejected.

marimo/_cli/sandbox.py
anthonywu: I ran into an error that said literally `-` is not expected to begin the dependency name. Maybe that's worth exploring an unblock natively in one requirements file, will attempt later. Could be a uv inconsistency with real pip.

tests/_cli/test_sandbox.py
anthonywu: I'll rebase and see if your newest fix solves some of the blockers I was overcoming in my PR, can potentially delete chunks of my code.

mscolnick: Depending how you were installing those deps, my change may have fixed some of this

anthonywu:
@mscolnick here's a good data point:

if I have a requirements.txt as simple as this, going to a real path:

```
-e /Users/anthonywu/workspace/rich-cli
```

```sh
uv run --isolated --no-project --with-requirements /tmp/mixed-install.txt rich --version
1.8.0
```

so `uv run` *does* support `-e` and `--editable` if just added as entries to the tempfile that represents the req list.

So the missing piece might be somewhere else in the repo, and my `sandbox.py` mod here is a workaround rather than the best implementation. I'll close the PR and re-approach at a later time.

mscolnick:
@anthonywu yea that could be an inconsistency. 

`uv run script.py` with inline metadata didn't seem to work for `--editable` paths , but maybe it does if in `--with-requirements`

anthonywu:
@mscolnick perhaps this is a issue to file with `uv` and we just wait for the upstream support. For now, the code bloat probably isn't worth it.

PR #2448, opened 2024-09-29
improvement: allow more package installation rules in the pkg management UI

mscolnick:
Allow fancier package installation rules like editable (`-e`) or `@`

### DIFF
--- a/marimo/_runtime/packages/utils.py
+++ b/marimo/_runtime/packages/utils.py
@@ -40,4 +40,51 @@ def append_version(pkg_name: str, version: Optional[str]) -> str:
 
 
 def split_packages(package: str) -> List[str]:
-    return [pkg.strip() for pkg in package.split()]
+    """
+    Splits a package string into a list of packages.
+
+    This can handle editable packages (i.e. local directories)
+
+    e.g.
+    "package1 package2" -> ["package1", "package2"]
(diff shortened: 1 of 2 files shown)


anthonywu:
observation about default pyproject+uv+marimo behavior in combination:

if I have script metadata:

```toml
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "marimo",
#     "foo @ ~/workspace/foo",
#     "-e ~/workspace/bar",
# ]
# ///
```

and I run it in sandbox mode, the ~ path is assumed to be a relative to PWD, and the whole path is expected to be `file:///` scheme

```
Running in a sandbox: uv run --isolated --no-project --with-requirements /var/folders/53/vkqj14gn2wsc1xp5dbx1wszh0000gn/T/tmp6xxqh9my.txt marimo run /tmp/local_editable_demo.py
  × Invalid `--with` requirement
  ╰─▶ Distribution not found at: file:///Users/anthonywu/workspace/marimo/~/workspace/bar
```

mscolnick:
@anthonywu, i can reproduce the first issue: `~` does not get expanded. i wonder if this can be handled by `uv` upstream. because even if we support expanding this, this would break if you didn't go through marimo, e.g. `python notebook.py` or `uv run notebook.py`. thoughts?

for the second issue: I don't get any errors. can you try upgrading your uv? can you reproduce this with `uv run` as well

PR #2449, opened 2024-09-29
chore: migrate to hatch for better local development testing

mscolnick:
This pull request primarily focuses on transitioning the project to use Hatch for Python environment and build management, replacing previous tools like `setuptools`. It also includes updates to the documentation and configuration files to reflect this change.

Hatch will use `uv` under the hood for fast downloading, while hatch manages the orchestration of creating environment and running tasks.

This makes running tests locally much more reproducible to CI. 
Docs will fail until this is merged and will updates docs deployment (otherwise updating now will make other branches fail).


anthonywu (commented):
Hatch is awesome, great choice!

mscolnick:
@anthonywu, I'm not all that familiar with it, it does seem like the best choice to run different python versions and dep matrices locally. 

Do you think you could be one of the reviewers? Would appreciate your expertise if you know it well.

akshayka (commented):
I'm down to switch to `hatchling` and `hatch`.

It looks like the only limitation of hatch is that it doesn't support extension modules (C, C++, Rust, etc). But we don't have any and won't for a long time if ever.

akshayka (commented):
Played around with this and I like it. Testing seamlessly with different python versions is a big improvement.

I noticed `contributing.md` says that `make py` still supports an editable install (which is nice), but the `dev` optional group has been removed from `pyproject.toml`, making it not work.

As described in this issue

https://github.com/pypa/hatch/issues/1165#issuecomment-1863117556

could we keep the optional groups and have the `hatch` sections refer to them?

mscolnick:
@akshayka, yes that is much better. i will keep the docs and testing deps out of "[dev]" though, to encourage running them in the isolated envs

akshayka (approved):
LGTM once CI issues are ironed out

PR #2453, opened 2024-09-30
feat: (experimental) inline edits

### LINKED ISSUE #3085: Highlight text for AI inline edits
### Description

We want users to be able to select text in a cell and ask an AI-assist to change that code. It should be similar logic to how AI-assist edits full cells but instead the line(s) selection for the cell

### Suggested solution

This should be a codemirror plugin paired with the existing AI /completions endpoint.

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
Fixes #3085

This adds continue.dev and cursor-like inline editing using https://github.com/marimo-team/codemirror-ai
This is activated with `Cmd+L`

https://github.com/user-attachments/assets/170790d4-9e31-435e-af79-bbbf0e9fa643

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -38,6 +38,7 @@
     "@lezer/highlight": "^1.2.1",
     "@lezer/lr": "^1.4.2",
     "@lezer/python": "^1.1.15",
+    "@marimo-team/codemirror-ai": "^0.1.3",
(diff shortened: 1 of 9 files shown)


akshayka (dismissed):
Awesome!

Based on the video I have feedback on styling (the button's design looks out of place), but that's minor and approving since it's behind a feature flag.

PR #2471, opened 2024-10-02
feat: _display_ protocol

akshayka:
If an object implements `_display_() -> Any`, we render it using marimo's built-in renderer applied to `obj._display_()`.

Example:

image

### DIFF
--- a/docs/guides/integrating_with_marimo/displaying_objects.md
+++ b/docs/guides/integrating_with_marimo/displaying_objects.md
@@ -7,9 +7,50 @@ seaborn, Plotly, altair pandas, and more. These rich representations are
 displayed for the last expression of a cell, or when using
 [`mo.output.append`](#marimo.output.append).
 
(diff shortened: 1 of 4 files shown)


mscolnick:
Is this a documented protocol? Should we include it in our documentation and write a test

akshayka:
May revisit later ...

mscolnick:
I found a good reason for this. These are equivalent, so I don't think it is worth the syntactic sugar, but having `_display_` could benefit libraries that don't/can't depend on `marimo`

```
def _display_():
  return obj

def _mime_():
  return ("text/html", mo.as_html(obj))
```

mscolnick (dismissed):

mscolnick:
Docs?

akshayka:
Yea needs docs and tests. Will finish this up later this week

mscolnick (approved):

PR #2480, opened 2024-10-02
fix: handle shadowed variables in mo.cache, update tutorials

akshayka:
This PR updates tutorials to use `mo.cache`, and fixes `mo.cache` to handle shadowed variables.

### DIFF
--- a/marimo/_save/save.py
+++ b/marimo/_save/save.py
@@ -116,6 +116,11 @@ class _cache_base(object):
         f_locals = inspect.stack()[2 + self._frame_offset][0].f_locals
         self.scope = {**ctx.globals, **f_locals}
         # In case scope shadows variables
+        #
+        # TODO(akshayka, dmadisetti): rewrite function args with an AST pass
+        # to make them unique, deterministically based on function body; this
+        # will allow for lifting the error when a ShadowedRef is also used
+        # as a regular ref.
         for arg in self._args:
             self.scope[arg] = ShadowedRef()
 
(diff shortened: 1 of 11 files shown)


dmadisetti:
Is this on the cache hit?

marimo/_save/save.py
```diff
@@ -114,6 +114,9 @@ def _set_context(self, fn: Callable[..., Any]) -> None:
         # checking a single frame- should be good enough.
         f_locals = inspect.stack()[2 + self._frame_offset][0].f_locals
         self.scope = {**ctx.globals, **f_locals}
+        # In case scope shadoes variables
+        for arg in self._args:
+            self.scope.pop(arg, None)
```
akshayka, line 119: ah this is what I was looking for

marimo/_tutorials/plots.py
```diff
@@ -173,15 +173,12 @@ def __(mo):
...
-    return exponent,
+    return (exponent,)
 
 
 @app.cell
 def __(exponent, mo, plt, x):
-    import functools
-
-
-    @functools.cache
```
dmadisetti, line 184: You can check with `_plots.cache_info().hits`, and as was, there would never be a cache hit with functools, since the function would get redefined and clear the cache

I did find that for, mo.cache, private functions did not have a persistent cache- but solved that

tests/_save/test_cache.py
```diff
@@ -934,3 +934,173 @@ def __(g, random):
...
+            # This is solved by throwing an exception when state
+            # shadowing occurs.
+            def f():
+                return state()
+
+            @mo.cache
+            def g(state):
+                return state() + f()
+
+            raise RuntimeError("Should not reach here")
```
dmadisetti, line 1031: nit: change the error here since RuntimeError is expected

dmadisetti: Or just remove

Issue #2474, opened 2024-10-02
[UX] Cmd+Shift+P/O should yank selected text into new cell


mholub:
### Description

It feels like it improves UX slightly

most common case when I want to do it - I write code in one cell and then I want to split piece of it into another cell, happens a lot with marimo.ui elements cause they want to be in separate cells anyway to work - you can't use value in the same cell

But I think it is generally useful and follows principle of least surprise

### Suggested solution

So if you have text selected inside cell source code and press Cmd+Shift+P - it appends new cell (like now) and automatically fills it with selected text (and edits current cell)

wasimsandhu:
Should highlighting text and creating a new cell with the keyboard shortcuts cut and paste that code into the new cell? I implemented this fairly quickly but think this deserves more discussion because the UX feels a little funny to me. I'll open a PR so we can play around with it.

wasimsandhu:
Never mind I like it 👍🏽

PR #2484, opened 2024-10-03
improve: Move highlighted code to newly created cells

wasimsandhu:
## 📝 Summary

Implements #2474.

## 🔍 Description of Changes

When a new cell is created, any highlighted code is cut from the original cell and pasted into the new cell.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/core/cells/__tests__/cells.test.ts
+++ b/frontend/src/core/cells/__tests__/cells.test.ts
@@ -1279,129 +1279,129 @@ describe("cell reducer", () => {
     expect(cell.consoleOutputs).toEqual([STDOUT1, STDOUT2]);
   });
 
-  it("can create a cell using selected code", () => {
+  describe("can create a cell using selected code", () => {
     const firstLine = "import marimo as mo";
     const secondLine = "import pandas as pd";
     const thirdLine = "import numpy as np";
 
-    actions.createNewCell({
-      cellId: firstCellId,
-      before: false,
-      code: `${firstLine}\n${secondLine}\n${thirdLine}`,
-    });
-    const secondCellId = state.cellIds.atOrThrow(1);
+    let editor: EditorView | undefined;
+    let secondCellId: CellId;
 
-    const editor = state.cellHandles[secondCellId].current?.editorView;
-    if (!editor) {
-      throw new Error("Editor not found");
-    }
-    editor.dispatch({
-      selection: {
-        anchor: editor.state.doc.line(2).from,
-        head: editor.state.doc.line(2).to,
-      },
+    beforeEach(() => {
+      actions.createNewCell({
+        cellId: firstCellId,
+        before: false,
+        code: `${firstLine}\n${secondLine}\n${thirdLine}`,
+      });
+      secondCellId = state.cellIds.atOrThrow(1);
+      editor = state.cellHandles[secondCellId].current?.editorView;
     });
 
-    const { from, to } = editor.state.selection.main;
-    const highlighted = editor.state.doc.toString().slice(from, to);
-    expect(highlighted).toBe(secondLine);
+    it("includeSelectionAsInitialCode is true and there is selection", () => {
+      if (!editor) {
+        throw new Error("Editor not found");
+      }
+      editor.dispatch({
+        selection: {
+          anchor: editor.state.doc.line(2).from,
+          head: editor.state.doc.line(2).to,
+        },
+      });
 
-    // includeSelectionAsInitialCode is true and there is selection
-    actions.createNewCell({
-      cellId: secondCellId,
-      before: false,
-      includeSelectionAsInitialCode: true,
+      const { from, to } = editor.state.selection.main;
+      const highlighted = editor.state.doc.toString().slice(from, to);
+      expect(highlighted).toBe(secondLine);
+
+      actions.createNewCell({
+        cellId: secondCellId,
+        before: false,
+        includeSelectionAsInitialCode: true,
+      });
+      expect(formatCells(state)).toMatchInlineSnapshot(`
+        "
+        key: 0
+        code: ''
+
+        key: 1
+        code: '${firstLine}\n${thirdLine}'
+
+        key: 2
+        code: '${secondLine}'"
+      `);
     });
-    expect(formatCells(state)).toMatchInlineSnapshot(`
-      "
-      key: 0
-      code: ''
 
-      key: 1
-      code: '${firstLine}\n${thirdLine}'
+    it("includeSelectionAsInitialCode is false and there is selection", () => {
+      if (!editor) {
+        throw new Error("Editor not found");
+      }
+      editor.dispatch({
+        selection: {
+          anchor: editor.state.doc.line(2).from,
+          head: editor.state.doc.line(2).to,
+        },
+      });
+      actions.createNewCell({
+        cellId: secondCellId,
+        before: false,
+        includeSelectionAsInitialCode: false,
+      });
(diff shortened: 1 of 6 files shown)


akshayka (commented):
Nice QOL improvement! One thing — after splitting, both cells should be marked as stale

wasimsandhu:
@akshayka Good catch thanks!

mscolnick:
@wasimsandhu - can you verify this works with SQL and markdown? We probably have to do something similar to `splitEditor`. could we also write a test similar to `splitEditor`(`utils.test.ts`) and `splitCell` (`cells.test.ts`)

frontend/src/core/cells/cells.ts
```diff
@@ -214,17 +216,24 @@ const {
         : state.cellIds.topLevelIds.indexOf(cellId);
     const insertionIndex = before ? index : index + 1;
 
+    let cellContents = code;
+    const cellEditorView = getCellEditorView(cellId as CellId);
```
mscolnick, line 220: i think we should pass a param to `createNewCell` whether we should `includeSelectionAsInitialCode`. 

we run `createNewCell` in many places and this would be weird for most cases (e.g. snippets, data-source panel, AI cell, etc).

could we also assert `code` and `includeSelectionAsInitialCode` are never both true?

wasimsandhu: Little confused here: do we need `includeSelectionAsInitialCode` if we are already redefining the `code` prop?

wasimsandhu:
> can you verify this works with SQL and markdown?

When editing a SQL/markdown cell and triggering a create new cell above/below (via button or shortcut), a Python cell is created. Do we want to change this behavior so that a cell of the same type is created? (This makes sense to me.)

mscolnick:
> When editing a SQL/markdown cell and triggering a create new cell above/below (via button or shortcut), a Python cell is created. Do we want to change this behavior so that a cell of the same type is created? (This makes sense to me.)

This might take some getting used to. I would keep it python for new cells, but keep the same language for "yanked" cells. thoughts? is that weird ux?

frontend/src/core/cells/cells.ts
```diff
@@ -217,10 +219,16 @@ const {
...
-    if (cellEditorView) {
-      cellContents = extractHighlightedCode(cellEditorView);
-      state.cellData[cellId as CellId].edited = true;
+
+    if (includeSelectionAsInitialCode) {
+      const id = cellId as CellId;
+      const cellEditorView = getCellEditorView(id);
+      if (cellEditorView) {
+        const [highlighted, leftover] = extractHighlightedCode(cellEditorView);
+        cellContents = highlighted;
```
mscolnick, line 228: we should only do this this if `highlighted` is not empty

mscolnick: i think it will catch small bugs like `state.cellData[id].edited=true` being set incorrectly. 

also, doing `state.cellData[id].edited` is mutable, we will need to be immutable when using `react`. 

```python
state.cellData = {
   ...state.cellData,
   [id] : {
     ...state.cellData[id],
     code: ...,
     edited: ...
   }
}
```

wasimsandhu:
> This might take some getting used to. I would keep it python for new cells, but keep the same language for "yanked" cells.

Works for me!

frontend/src/core/cells/__tests__/cells.test.ts
```diff
@@ -1276,4 +1278,130 @@ describe("cell reducer", () => {
     const cell = cells[0];
     expect(cell.consoleOutputs).toEqual([STDOUT1, STDOUT2]);
   });
+
+  it("can create a cell using selected code", () => {
```
mscolnick, line 1282: can you put each of these in there own tests and share setup code?
```
describe("can create a cell using selected code"

  beforeEach(() => {
   // setup code

  })

  it("can create when includeSelectionAsInitialCode is true and there is selection", () => {
    ...test one
   })
   ...
});
```

this makes it easier to follow given the same initial setup.

could you also:
1. verify edited state
2. do a test with markdown `includeSelectionAsInitialCode` - i noticed a bug when i add a cell above/below a markdown cell with `includeSelectionAsInitialCode=true`, it says it is edited

frontend/src/core/cells/cells.ts
```diff
@@ -214,17 +218,39 @@ const {
...
+        const [highlighted, leftover] = extractHighlightedCode(editorView);
+        if (highlighted.length > 0) {
+          newCellCode = highlighted === "" ? code : highlighted;
+          updateEditorCodeFromPython(editorView, leftover);
+          state.cellData = {
+            ...state.cellData,
+            [id]: {
+              ...state.cellData[id],
+              code: leftover,
+              edited: true,
```
mscolnick, line 236: this might need to be `edited: Boolean(leftover) && state.cellData[id].code !== state.cellData[id].code`,

i noticed a bug with markdown with no selection being made stale, it would be great to prove this in a test: get the test to fail first and then fix this code so the tests passes

frontend/src/core/codemirror/language/__tests__/utils.test.ts
```diff
@@ -147,3 +148,54 @@ describe("splitEditor", () => {
...
+      selection: {
+        anchor: mockEditor.state.doc.line(3).from,
+        head: mockEditor.state.doc.line(3).to,
+      },
+    });
+    const [highlighted, leftover] = extractHighlightedCode(mockEditor);
+    expect(highlighted).toEqual(thirdLine);
+    expect(leftover).toEqual(`${firstLine}\n${secondLine}`);
+  });
+});
```
mscolnick, line 201: can you write some tests for markdown and sql?
1. sql x no selection
2. sql x with selection
3. markdown x no selection
4. markdown x with selection

frontend/src/core/cells/cells.ts
mscolnick: made a suggestion above that you may not need this.

frontend/src/core/cells/cells.ts
```diff
@@ -214,17 +218,39 @@ const {
...
     const insertionIndex = before ? index : index + 1;
 
+    let newCellCode = code;
+
+    if (includeSelectionAsInitialCode) {
+      const id = cellId as CellId;
+      const editorView = state.cellHandles[id].current?.editorView;
+      if (editorView) {
+        const [highlighted, leftover] = extractHighlightedCode(editorView);
+        if (highlighted.length > 0) {
```
mscolnick, line 228: i think the issue is that for markdown/sql, highlighted > 0. if you write a test for `extractHighlightedCode`, can we get verify that no selection -> no highlighted

frontend/src/core/codemirror/language/utils.ts
```diff
@@ -69,3 +69,18 @@ export function splitEditor(editor: EditorView) {
     afterCursorCode,
   };
 }
+
+export function extractHighlightedCode(editor: EditorView) {
+  const code = editor.state.doc.toString();
+  let { from, to } = editor.state.selection.main;
```
mscolnick, line 75: you could for `if (from === to) {return ["", code]`

mscolnick: or could do `if (from === to) {return null` to signify nothing was highlighted. it will force consumers of this downstream to check `null`.

i think that is better / clearer

wasimsandhu: Implemented this and it seems to have fixed the stale issue

mscolnick (approved):
awesome work!

Issue #2488, opened 2024-10-03
Refresh connection on App Mode


peterwilli:
### Description

When running as `marimo run <file>`, and doing something intensive with a lot of RAM, and then refreshing the page, it will spawn a completely new session, risking human error, and loading things twice..

### Suggested solution

Is it possible to get the same behaviour as in edit mode, but without showing code?

Like this:
<img width="1021" alt="Showing reconnection panel" src="https://github.com/user-attachments/assets/298a35ad-b095-4b0f-8ce2-3ab460a6715f">

peterwilli:
Wow that's super fast! Thanks! I was just checking the source while waiting for an answer!

Uhm, the thing is the App Mode runs locally in my case. It is launched through an shell script and the user can then use the notebook as application.

In the application, I load an LLM (GGUF) and one user has refreshed the browser, causing the LLM to load twice. I was not aware of this because I assumed the edit mode was similar to run mode, minus the source code.

I'm not aware how I can persist this loading across sessions, except when I launch a HTTP API or something, which I rather avoid (I really like the simplicity it has now)

mscolnick:
Your use-case makes a lot of sense. We can consider adding `--resumable-sessions` (or `--persist-sessions`), with some sort of timeout, to the `marimo run` CLI. 

or `--single-session` (which might be easier to implement, since then we don't need to keep track of which session belongs to who).

Implementation details aside: we may not get to this for a bit, so open to contributions; I'll add some GH labels if someone wants to pick this up.

peterwilli:
I'm actually making it right now, I wasn't sure if you'd merge it but I'll post it as PR if done!

mscolnick:
hell yea! thanks @peterwilli, i think it would be a great addition 🙌

peterwilli:
Here you go! https://github.com/marimo-team/marimo/pull/2489

peterwilli:
Closing this issue as the PR is now in progress

PR #2489, opened 2024-10-03
allow --global-session for marimo run

peterwilli:
## 📝 Summary

I encountered this use-case here: https://github.com/marimo-team/marimo/issues/2488

When doing large computations running entirely locally, especially with non-coder users, it is often important to be able to resume the session when the browser refreshes, or tabs wake up from sleep, especially when loading large files into memory.

In my case, a customer has refreshed their browser, causing a large model to load in memory twice. We don't run marimo on a server but locally, so there's no need for unique sessions.

This feature lets you add `--global-session` to your `marimo run` command, sharing the session even when refreshing, asserting the same session behaviour as `marimo edit`.

For the rest, default behaviour stays, so after this PR, everyone still gets unique sessions in `marimo run`, except if they provide the argument.

This is a living PR, and certainly not perfect yet - I assume that we have to add some e2e tests, but I'm not familiar enough to make this work yet, and need some guidance.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_server/sessions.py
+++ b/marimo/_server/sessions.py
@@ -661,6 +661,7 @@ class SessionManager:
         cli_args: SerializedCLIArgs,
         auth_token: Optional[AuthToken],
         redirect_console_to_browser: bool = False,
+        global_session: bool = False,
     ) -> None:
         self.file_router = file_router
         self.mode = mode
@@ -674,6 +675,7 @@ class SessionManager:
         self.user_config_manager = user_config_manager
         self.cli_args = cli_args
         self.redirect_console_to_browser = redirect_console_to_browser
+        self.global_session = global_session
 
         # Auth token and Skew-protection token
         if auth_token is not None:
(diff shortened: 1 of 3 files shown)


mscolnick:
this is great! the direction looks right - could we verify with some e2e tests? some examples:

1. only one connection is allowed when in `--global-session`. otherwise we will likely get some bugs that there is inconsistent state for different users. (state would exist on the client and server that may not match)
2. we can resume the session while in run mode
3. we probably already have a test for this, but just checking that 2 run modes can be active at once. `test_ws.py` is the place to find these

could you also help update `docs/guides/app.md`? we could add a section `## Global session`

marimo/_server/sessions.py
```diff
@@ -661,6 +661,7 @@ def __init__(
         cli_args: SerializedCLIArgs,
         auth_token: Optional[AuthToken],
         redirect_console_to_browser: bool = False,
+        global_session: bool = False,
```
mscolnick, line 664: maybe we can rename to `global_session_in_run_mode` or something similar. it is verbose, but this could be confused that edit mode also as global session configuration

peterwilli: Yeah certainly, any name that makes sense is good for me :)

mscolnick:
sorry i mispoke, you can just add the tests to the python side. our e2e tests with playright are a bit finicky and flakey so we  don't keep those updated.

peterwilli:
> sorry i mispoke, you can just add the tests to the python side. our e2e tests with playright are a bit finicky and flakey so we don't keep those updated.

Ok, I'm currently checking out the Python tests, how they work etc

peterwilli:
@mscolnick I looked into the test_ws.py, I see how my test would fit there, but I'm struggling to understand where the server is launched, is there some place where the tests are initated?

mscolnick:
@peterwilli, can we help unblock you?

PR #2501, opened 2024-10-05
docs: editing pass

akshayka:
Editing pass on docs


docs/guides/index.md
```diff
@@ -22,7 +22,7 @@
   state
   wasm
   exporting
-  configuration
+  configuration/index
```
mscolnick, line 25: @akshayka we should make sure there are no existing links to configuration.html otherwise they will become dead

akshayka: I did check and didn't find any

PR #2504, opened 2024-10-05
improve: temporarily show code

akshayka:
This PR improves the UX of working with hidden cells. When code is hidden, clicking on the eye icon will now temporarily show the code as long as the editor is focused, and hide it again once the editor is blurred.

In this way users can, for example, hide markdown code and later edit it without permanently unhiding it.

To permanently unhide code, users can use the cell action menu or the keyboard shortcut, as before.

This change is based on user feedback, and it also happens to match the behavior in Observable notebooks.

Interactions with changing the cell config to `show code` while temporarily focused still works, since the user must first click the cell action menu which blurs the editor.

### DIFF
--- a/frontend/src/components/editor/cell/code/cell-editor.tsx
+++ b/frontend/src/components/editor/cell/code/cell-editor.tsx
@@ -108,27 +108,27 @@ const CellEditorInternal = ({
 
   const createBelow = useCallback(
     () => createNewCell({ cellId, before: false }),
-    [cellId, createNewCell]
+    [cellId, createNewCell],
   );
   const createAbove = useCallback(
     () => createNewCell({ cellId, before: true }),
-    [cellId, createNewCell]
+    [cellId, createNewCell],
   );
   const moveDown = useCallback(
     () => moveCell({ cellId, before: false }),
-    [cellId, moveCell]
+    [cellId, moveCell],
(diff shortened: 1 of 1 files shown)


akshayka:
In a follow-up we could consider automatically setting markdown cells to `hide_code=True` on creation

frontend/src/components/editor/cell/code/cell-editor.tsx
```diff
@@ -325,12 +325,15 @@ const CellEditorInternal = ({
...
-  const showCode = async () => {
+  const temporarilyShowCode = async () => {
     if (hidden) {
-      await saveCellConfig({ configs: { [cellId]: { hide_code: false } } });
       updateCellConfig({ cellId, config: { hide_code: false } });
-      // Focus on the editor view
       editorViewRef.current?.focus();
+      editorViewRef.current?.contentDOM.addEventListener(
+        "blur",
+        () => updateCellConfig({ cellId, config: { hide_code: true } }),
```
mscolnick, line 334: We should remove the event listener in this callback after

mscolnick: oh sorry I see the once

mscolnick (approved):

PR #2508, opened 2024-10-06
fix: hidden cell blur on parent, not editorview

akshayka:
When re-hiding a cell, process blur on the parent of the editor view, not the editorview itself, so things like the delete button, language toggle, and codemirror folds are still clickable.

cc @wasimsandhu


mscolnick (approved):

wasimsandhu:
This seems to have broken the temporary hide/unhide. It's not catching a blur on the parent ref

PR #2510, opened 2024-10-06
docs: mo.ui.plotly only supports selecting scatter plots

akshayka:
Address confusion from users, who suggested we improve our docs.


mscolnick:
We also support treemap and sunburst too

PR #2527, opened 2024-10-08
improve: SQL examples

akshayka:
Create a collection of bite-sized examples showing how to use SQL:

- querying dataframes
- parametrizing queries
- read from CSV
- read from JSON
- read from parquet
- connect to sqlite
- connect to postgres


examples/sql/connect_to_postgres.py
```diff
@@ -0,0 +1,184 @@
...
+        ## Copy data from Postgres to duckdb
+
+        To prevent duckdb from continuously re-reading tables from PostgresSQL, you can copy the PostgresSQL databases into DuckDB. Note that this will consume your system's RAM.
+        """
+    )
+    return
+
+
+@app.cell
+def __(db, duckdb_table, mo, test_table):
```
mscolnick, line 150: unrelated, but this import `duckdb_table` seems incorrect.

examples/sql/read_json.py
```diff
@@ -0,0 +1,156 @@
...
+        This notebook shows how to read a JSON file from a local file or a URL into an in-memory table.
+        """
+    )
+    return
+
+
+@app.cell(hide_code=True)
+def __():
+    import marimo as mo
+    import pandas as pd
```
mscolnick, line 33: super nit; but i feel like we should use polars (instead of pandas) when possible

akshayka: yea i agree in general but i'd like to host these on WASM too

examples/sql/connect_to_postgres.py
akshayka: yea weird that it's a def and a ref ...

akshayka: The Python code has logic for not registering something as a ref if it's a def, SQL could be missing that

examples/sql/read_json.py
mscolnick: oh yes good call - we did discuss that

mscolnick (approved):

PR #2542, opened 2024-10-08
example: datasette/llm chatbot

mscolnick:


examples/ai/chat/llm_datasette.py
```diff
@@ -0,0 +1,44 @@
...
+    return llm, mo
+
+
+@app.cell
+def __(llm, mo):
+    model = llm.get_model("gpt-4o-mini")
+    conversation = model.conversation()
+
+    mo.ui.chat(lambda messages: conversation.prompt(messages[-1].content))
+    return conversation, model
```
akshayka, line 40: small nit: should this have one more cell that just shows that chat component's value? Like the other examples

mscolnick: i dont think it needs to since i am not using the value. i think examples should be _examples_ and not tutorials

akshayka: I think we have two flavors. Illustrative examples where people can just look at them and be like, oh, cool.

But the primary point is to help people get started on their on work. And if you believe folks don't read documentation, there's no other place they'll learn things than from examples.

To me, this example you've added is more educational than a standalone interesting thing to look at.

PR #2567, opened 2024-10-09
experimental: mo.ui.data_editor

mscolnick:
This adds an experimental ui element for editing a dataframe, using `mo.ui.data_editor`.

* Add `mo.ui.data_editor` (BE plugin and FE plugin)
* This adds a dependency on `narwhals`. It's a lightweight library (no deps itself) and does a lot of what we try to roll ourselves, by creating a single dataframe interface. There is a lot of other code to clean up by depending on this. This library will also be better tested again more versions of polars/pandas/other-df-library
* We use `ag-grid` for the editing. It is missing key features that only exist in the enterprise that we will likely want and force us to move to something else. This includes: cell range selection, copy/paste ranges. I don't expect this future migration to change the API, only implementation details. I also don't expect the migration to be any more expensive since it was quite easy to add ag-grid and is limited to one file.

Current features:
* edit cells (text, long text, boolean, number)
* add rows

Later features to add:
* delete rows
* copy/paste
* range selection
* date cell

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -84,6 +84,8 @@
     "@xterm/addon-attach": "^0.11.0",
     "@xterm/addon-fit": "^0.10.0",
     "@xterm/xterm": "^5.5.0",
+    "ag-grid-community": "^32.2.2",
+    "ag-grid-react": "^32.2.2",
     "ai": "^3.3.12",
     "ansi_up": "^6.0.2",
     "class-variance-authority": "^0.7.0",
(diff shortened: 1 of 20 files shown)


akshayka (commented):
- API looks good.

- How experimental is this? Given the `EXPERIMENTAL` warning, wondering if we should release under `mo.experimental_data_editor` like we did for islands in the past.

- I am okay with the narwhals dependency. As per [their docs](https://narwhals-dev.github.io/narwhals/backcompat/) on perfect backward compatibility, can we replace

```
import narwhals as nw
```

with

```
import narwhals.stable.v1 as nw
```

?

- Frontend looks fine to me too.

PR #2569, opened 2024-10-09
feat: add column summaries for sql tables in the datasources panel

mscolnick:
image

### DIFF
--- a/marimo/_data/sql_summaries.py
+++ b/marimo/_data/sql_summaries.py
@@ -141,7 +141,6 @@ def get_column_type(table_name: str, column_name: str) -> DataType:
         )
 
     db_column_type = column_info_result[0].lower()
-    print(db_column_type)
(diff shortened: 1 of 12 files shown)


marimo/_data/sql_summaries.py
```diff
@@ -0,0 +1,156 @@
+from __future__ import annotations
+
+from typing import List, Tuple
+
+from marimo._data.get_datasets import _db_type_to_data_type
+from marimo._data.models import ColumnSummary, DataType
+from marimo._sql.sql import _wrapped_sql
+
+
+def get_sql_summary(
```
akshayka, line 10: Is there a way to check if the table is already loaded into `duckdb`, or if it's e.g. from Postgres? I think this could be very expensive for large Postgres/row-oriented tables

mscolnick: Yes we could limit to in-memory

PR #2595, opened 2024-10-11
fix: don't show errors in run mode when re-running

mscolnick:
* Don't show errors in run mode when re-running. This can happen when a cell errors (not shown yet) and then is re-running (is in a queued/running state) so we end up showing the error briefly.
* Since we do send these errors to the frontend anyways, I now log it to the console for developers.


akshayka (approved):
> I now log it to the console for developers.

nice touch

PR #2602, opened 2024-10-11
release marimo-base to test pypi

akshayka:
This change adds a workflow that releases a marimo-base package to TestPyPI.

marimo-base is marimo without static artifacts, which are not needed when running on WASM.

Because Python build tools only allow one project per pyproject.toml, it is not possible to do this by adding a build rule to hatch. Instead, marimo-base is built by adapting the pyproject.toml in CI (this is what CVXPY does).

With this change, the wheel size is reduced from `15M` to `614k`.

### DIFF
--- a/.github/workflows/dev-release-marimo-base.yml
+++ b/.github/workflows/dev-release-marimo-base.yml
@@ -1,4 +1,4 @@
-name: Publish dev release
+name: Publish marimo-base dev release
 
 # Publish development release to test pypi on pushes to main
 on:
--- a/.github/workflows/dev-release-marimo-base.yml
+++ b/.github/workflows/dev-release-marimo-base.yml
@@ -1,0 +1,71 @@
+name: Publish dev release
+
+# Publish development release to test pypi on pushes to main
+on:
+  push:
+    branches:
+      - main
+
+env:
+  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+  TURBO_TEAM: marimo
+
+jobs:
+  publish_dev_release:
+    name: 📤 Publish dev release for marimo-base
+    runs-on: ubuntu-latest
+    defaults:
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-release-marimo-base.yml
```diff
@@ -0,0 +1,71 @@
...
+      # to build marimo-base, a slimmed down marimo distribution with
+      # no static artifacts.
+      # 
+      # Adapted from https://github.com/cvxpy/cvxpy/blob/297278e2a88db3c0084750052a16e60672074da3/.github/workflows/build.yml#L169C1-L180C1
+      - name: Adapt pyproject.toml to build marimo-base
+        run: |
+          # Mac has a different syntax for sed -i, this works across oses
+          sed -i.bak -e 's/name = "marimo"/name = "marimo-base"/g' pyproject.toml
+          # Remove static artifacts
+          sed -i.bak '/artifacts = /d' pyproject.toml
```
mscolnick, line 45: we could put this in `pyproject.toml` under a different target name: https://hatch.pypa.io/1.9/config/build/#pyprojecttoml_18

`wheel-slim`

mscolnick (approved):

.github/workflows/dev-release-marimo-base.yml
akshayka: It's still going to be released as `marimo` on the package indices though. Whereas want to create a separate package called `marimo-base`.

akshayka: Oh I see what you're saying. Additionally I couldn't figure out how to get a custom target to build a wheel.

mscolnick: Oh maybe hatch doesn't allow you to specify a target

PR #2604, opened 2024-10-11
perf: lazy load completion deps (for WASM)

akshayka:
Lazy load some dependencies so they don't block the notebook code from rendering.

### DIFF
--- a/frontend/src/core/wasm/worker/bootstrap.ts
+++ b/frontend/src/core/wasm/worker/bootstrap.ts
@@ -63,7 +63,7 @@ export class DefaultWasmController implements WasmController {
     const pyodide = await loadPyodide({
       // Perf: These get loaded while pyodide is being bootstrapped
       // The packages can be found here: https://pyodide.org/en/stable/usage/packages-in-pyodide.html
-      packages: ["micropip", "docutils", "Pygments", "jedi", "pyodide-http"],
+      packages: ["micropip"],
       // Without this, this fails in Firefox with
       // `Could not extract indexURL path from pyodide module`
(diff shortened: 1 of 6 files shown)


frontend/src/core/wasm/worker/bootstrap.ts
```diff
@@ -206,6 +206,10 @@ export class DefaultWasmController implements WasmController {
 
   private async loadNotebookDeps(code: string, foundImports: Set<string>) {
     const pyodide = this.requirePyodide;
+    // Additional dependencies of marimo that are lazily loaded.
+    //
+    // pyodide-http is a patch to make basic http requests work in pyodide
+    await pyodide.loadPackage(["docutils", "Pygments", "jedi", "pyodide-http"]);
```
mscolnick, line 212: Does this actually make it faster? This is now serial instead of parallel

mscolnick: Oh I see it shows the notebook code faster but not started. You can still push this down to be parallel with the users deps

akshayka: Yea, it makes the code show faster. Will do, thanks for the tip

mscolnick: You can promise.all this below at line 224

mscolnick:
ill make the updates. pinning will help with perf

mscolnick (approved):

mscolnick:
@akshayka - i was wrong about pinning versions. it doesn't matter since they are downloaded from pyodide custom wheels which are pinned by default

PR #2613, opened 2024-10-12
example: seam carving

akshayka:
adapt vincent warmerdam's seam carving example, add to misc/

### DIFF
--- a/examples/misc/seam_carving.py
+++ b/examples/misc/seam_carving.py
@@ -1,0 +1,190 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#     "marimo",
+#     "numba==0.60.0",
+#     "numpy==2.0.2",
+#     "requests==2.32.3",
+#     "scikit-image==0.24.0",
+# ]
+# ///
+
+import marimo
+
+__generated_with = "0.9.6"
+app = marimo.App(width="medium")
+
+
+@app.cell(hide_code=True)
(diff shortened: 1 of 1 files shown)


examples/misc/seam_carving.py
```diff
@@ -0,0 +1,190 @@
...
+        label="Amount of resizing to perform:",
+        show_value=True,
+    )
+    slider
+    return mo, slider
+
+
+@app.cell
+def __(efficient_seam_carve, input_image, mo, slider):
+    with mo.persistent_cache("seam_carves"):
```
dmadisetti, line 85: Doesn't make much difference right now, but with remote caching this would be better:

```python
steps = np.linspace(0.7, 1, 60)
slider = mo.ui.slider(
        steps=steps,
        value=1.0,
        label="Amount of resizing to perform:",
        show_value=True,
    )
    
# Compute them all, so cache hits never happen
with mo.persistent_cache("seam_carves"):
     # Or even do in parallel
     results = list(map(lambda x: efficient_seam_carve(input_image, x), steps))

result = results[np.argmin((steps - slider.value)**2)]
```

Which would allow embedding into a blog post with 0 latency.
I think this is still ideal if you are still developing / messing around though

akshayka: oh my, i really like that idea ✨

mscolnick (approved):

PR #2616, opened 2024-10-13
0.9.8

akshayka:


dmadisetti:
@akshayka marimo is not dependent on marimo-base though right? They are both independent? Just checking from a nix packaging perspective

PR #2631, opened 2024-10-15
fix: consistent value for no column in config

akshayka:
Previously, the value for column in a single-column notebook was sometimes null, sometimes undefined. This led to false positives on notebookNeedsSave. This PR changes it so that when a column is not set, the corresponding config value is `null` (which corresponds to `None` in Python).

Without this fix, if you opened a notebook and saved it with or without making any changes (experimental multi-column was disabled), the notebook was marked as needing to be save for the remainder of the session, no matter how many times you clicked "save".

I found that I also had to change arrayShallowEquals to lodash isEquals for the equality check on config objects to work.

### DIFF
--- a/frontend/src/core/cells/utils.ts
+++ b/frontend/src/core/cells/utils.ts
@@ -37,8 +37,8 @@ export function notebookNeedsSave(
   const names = data.map((d) => d.name);
   return (
     !arrayShallowEquals(codes, lastSavedNotebook.codes) ||
-    !arrayShallowEquals(configs, lastSavedNotebook.configs) ||
     !arrayShallowEquals(names, lastSavedNotebook.names) ||
+    !isEqual(configs, lastSavedNotebook.configs) ||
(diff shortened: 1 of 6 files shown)


frontend/src/core/cells/utils.ts
```diff
@@ -37,8 +37,8 @@ export function notebookNeedsSave(
   const names = data.map((d) => d.name);
   return (
     !arrayShallowEquals(codes, lastSavedNotebook.codes) ||
-    !arrayShallowEquals(configs, lastSavedNotebook.configs) ||
     !arrayShallowEquals(names, lastSavedNotebook.names) ||
+    !isEqual(configs, lastSavedNotebook.configs) ||
```
akshayka, line 41: Not sure why this was working with `arrayShallowEquals` before ...

mscolnick: possibly mutating state instead of keeping it immutable

mscolnick (approved):

PR #2643, opened 2024-10-15
improvement: use narwhals for more logic, get rid of dataprotocol and pyarrow logic

mscolnick:
Move over more logic to narwhals. This will benefit us long term as narwhals continues to get more coverage


marimo/_plugins/ui/_impl/tables/narwhals_table.py
```diff
@@ -0,0 +1,276 @@
...
+            )
+        if is_narwhals_temporal_type(col.dtype):
+            return ColumnSummary(
+                total=total,
+                nulls=col.null_count(),
+                min=col.min(),
+                max=col.max(),
+                mean=col.mean(),
+                # TODO: Implement median not in narwhals
+                # median=col.median(),
```
akshayka, line 199: For median, `col.quantile(0.50, interpolation="nearest")`?

marimo/_utils/narwhals_utils.py
```diff
@@ -41,14 +41,14 @@ def assert_narwhals_series(series: nw.Series) -> None:
...
-def can_narwhalify(obj: Any) -> TypeGuard[IntoFrame]:
+def can_narwhalify(obj: Any, eager_only: bool = False) -> TypeGuard[IntoFrame]:
     """
     Check if the given object can be narwhalified.
     """
     if obj is None:
         return False
     try:
-        nw.from_native(obj, strict=True)
+        nw.from_native(obj, strict=True, eager_only=eager_only)  # type: ignore[call-overload]
```
akshayka, line 51: I was curious about the performance of this method, and saw it uses the same trick as we do -- getting modules from `sys.modules` instead of importing them ✅

Intentional to not allow series?

akshayka (approved):

marimo/_plugins/ui/_impl/tables/narwhals_table.py
mscolnick: sorry already merged, will do this in my followup, if they don't release median by that time

marimo/_utils/narwhals_utils.py
mscolnick: series is true by default. i can make that explicit (in case defaults change and readability)

mscolnick: i think i saw a narwhals PR too to add this utility without the try/catch

PR #2645, opened 2024-10-16
feat: add run in docker for remote urls

mscolnick:
This prompts the user to run the marimo file in a dockerized container when running remote marimo notebooks.

```
marimo edit https://raw.githubusercontent.com/marimo-team/marimo/main/examples/third_party/polars/polars_example.py
```
Screenshot 2024-10-16 at 9 02 42 AM

Screenshot 2024-10-16 at 9 03 01 AM

Couple details on implementation:
* I haven't exposed this as feature-flag yet, since I don't know if we want people to do this for local notebooks. Also most of the other CLI flags won't work, so this can create more issues/bugs than help.
* We prompt on all URL based notebooks (unless `--yes`) is passed.
* We run notebooks using `uvx marimo` and using the uv docker image. (we could use our own later which doesn't buy much at the moment)
* I didn't use the `docker` python library which would be cleaner, so we can avoid the additional dependency

Error state:
Screenshot 2024-10-16 at 9 02 29 AM

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -21,6 +21,9 @@ from marimo._cli.export.commands import export
 from marimo._cli.file_path import validate_name
 from marimo._cli.parse_args import parse_args
 from marimo._cli.print import red
+from marimo._cli.run_docker import (
+    prompt_run_in_docker_container,
+)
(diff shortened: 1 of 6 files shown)


akshayka (approved):
Love this. Agree with the decision to not expose as a CLI flag.

PR #2682, opened 2024-10-22
improvement: improve loading states while islands are initializing

mscolnick:
Various improvements for islands:
* improve loading states while islands are initializing
* fix some flickering CSS when loading
* fix newlines in markdown rendering on first load
* stylings fixes for islands height/padding
* create a namespace `islands`


frontend/src/core/islands/bridge.ts
```diff
@@ -169,3 +171,11 @@ export class IslandsPyodideBridge implements RunRequests, EditRequests {
     });
   }
 }
+
+function makeRelativeWorkerUrl(url: string) {
+  return url.startsWith("./")
```
dmadisetti, line 176: Normally I test islands behind a proxy, and this isn't always consistent. Will try

marimo/_islands/island_generator.py
```diff
@@ -397,12 +398,12 @@ def render_init_island(self) -> str:
             <marimo-island
                 data-app-id="{self._app_id}"
                 data-cell-id="{init_cell_id}"
-                data-reactive="{json.dumps(True)}"
+                data-reactive="{json.dumps(False)}"
             >
                 <marimo-cell-output>
                 {init_output}
                 </marimo-cell-output>
-                {init_input}
```
dmadisetti, line 405: Was this just a NameError before? Maybe a unit test to catch this branch

dmadisetti: But also, is init_island even needed with the toast?

dmadisetti (commented):
Reads well to me! I haven't tried it out, but eager for the toast !

Is the spinning loader still needed with the toast?

frontend/src/core/islands/bridge.ts
mscolnick: depends what i am testing - but mostly ill just run:
`pnpm dev:islands` and then access the server with `<script type="module" src="/src/core/islands/main.ts"></script>`

marimo/_islands/island_generator.py
mscolnick: > Was this just a NameError before? Maybe a unit test to catch this branch

It wasn't a name error. there was no error, I just inlined the variable. ill create a snapshot test

> is init_island even needed with the toast

they somewhat have different use-cases. `init_island` shows a spinner while HTML loads (seems like you may want to show before any static HTML appears). the toast happens only when you interact with static HTML before it "rehydrates"

PR #2711, opened 2024-10-24
improvement: more narwhals for table logic

mscolnick:
Re-use some logic for narhwals - PolarsTableManager and PandasTableManager will extends the NarwhalsTableManager


marimo/_plugins/ui/_impl/tables/narwhals_table.py
```diff
@@ -253,9 +251,15 @@ def get_unique_column_values(self, column: str) -> list[str | int | float]:
...
-        return NarwhalsTableManager(sorted_data)
+    ) -> NarwhalsTableManager[Any]:
+        if isinstance(self.data, nw.LazyFrame):
+            return NarwhalsTableManager(
+                self.data.sort(by, descending=descending)
+            )
+        else:
+            return NarwhalsTableManager(
+                self.data.sort(by, descending=descending)
+            )
```
mscolnick, line 262: this is the exact same, but makes function that use overloading happier

PR #2716, opened 2024-10-24
fix: remove any need for tornado in matplotlib.interactive

### LINKED ISSUE #2713: Sandbox doesn't install mo.mpl.interactive dependency tornado
### Describe the bug

Running `marimo edit --sandbox foo.py` and pasting and running the code below results in `ModuleNotFoundError: No module named 'tornado'`.

Manually adding `import tornado` to get the sandbox to install the dependency then requires restarting the kernel before anything will show up in the plot.

### Environment

```
{
  "marimo": "0.9.14",
  "OS": "Darwin",
  "OS Version": "23.6.0",
  "Processor": "arm",
  "Python Version": "3.11.2",
  "Binaries": {
    "Browser": "130.0.6723.59",
    "Node": "v22.9.0"
  },
  "Dependencies": {
    "click": "8.1.3",
    "docutils": "0.19",
    "itsdangerous": "2.2.0",
    "jedi": "0.18.2",
    "markdown": "3.5.2",
    "narwhals": "1.9.3",
    "packaging": null,
    "psutil": "5.9.4",
    "pygments": "2.14.0",
    "pymdown-extensions": "10.7",
    "pyyaml": "6.0",
    "ruff": "0.0.292",
    "starlette": "0.36.1",
    "tomlkit": "0.12.3",
    "typing-extensions": "4.9.0",
    "uvicorn": "0.27.0.post1",
    "websockets": "12.0"
  },
  "Optional Dependencies": {
    "pandas": "2.0.2",
    "polars": "0.19.12"
  }
}

```

### Code to reproduce

```
import marimo as mo

plt.plot([1, 2])
# plt.gcf() gets the current figure
mo.mpl.interactive(plt.gcf())
```

mscolnick:
Fixes #2713

We don't actually use tornado (we moved it to Starlette at some point), but an import still uses tornado

### DIFF
--- a/marimo/_output/mpl.py
+++ b/marimo/_output/mpl.py
@@ -16,13 +16,13 @@ import base64
 import io
 from typing import Optional
 
-import matplotlib.pyplot as plt  # type: ignore
-from matplotlib._pylab_helpers import Gcf  # type: ignore
-from matplotlib.backend_bases import (  # type: ignore
+import matplotlib.pyplot as plt
+from matplotlib._pylab_helpers import Gcf
+from matplotlib.backend_bases import (
     FigureCanvasBase,
     FigureManagerBase,
 )
-from matplotlib.backends.backend_agg import FigureCanvasAgg  # type: ignore
+from matplotlib.backends.backend_agg import FigureCanvasAgg
 
 from marimo._messaging.cell_output import CellChannel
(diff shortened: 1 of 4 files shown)


mscolnick:
matplotlib interactive won't be compatible in webassembly since at the moment it requires a running server.

PR #2740, opened 2024-10-29
doc: add period to Plotly docstring

ndrezn:
Adds a period to the `plotly` docstring. [In the docs](https://docs.marimo.io/guides/working_with_data/plotting.html#plotly) it ends up looking like a typo:
Screenshot 2024-10-29 at 9 52 33 AM

Looks like convention is to end each param definition without a period, though, so I haven't added periods where the description is only one sentence.


ndrezn:
recheck

PR #2792, opened 2024-11-05
examples: add huggingface examples

mscolnick:
HF examples, as they appear on https://huggingface.co/marimo-team

### DIFF
--- a/marimo/_output/formatters/tqdm_formatters.py
+++ b/marimo/_output/formatters/tqdm_formatters.py
@@ -35,7 +35,9 @@ class TqdmFormatter(FormatterFactory):
                     total = args[2]
 
                 return progress_bar(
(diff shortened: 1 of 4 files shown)


marimo/_output/formatters/tqdm_formatters.py
```diff
@@ -35,7 +35,9 @@ def tqdm_to_progress_bar(
                     total = args[2]
 
                 return progress_bar(
-                    collection=iterable, title=desc, total=total
+                    collection=iterable,
+                    title=desc or "Loading...",
```
mscolnick, line 39: Since this is optional, it's somewhat ugly when nothing is passed and taking a while to load, it looks like marimo is stuck (noticed in the hf example)

akshayka (approved):

PR #2800, opened 2024-11-05
Add pygwalker example

akshayka:
Very simple example showing usage in marimo.

### DIFF
--- a/examples/third_party/pygwalker/example.py
+++ b/examples/third_party/pygwalker/example.py
@@ -1,0 +1,31 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#     "pygwalker==0.4.9.13",
+#     "vega-datasets==0.9.0",
+# ]
+# ///
+import marimo
+
+__generated_with = "0.9.14"
+app = marimo.App(width="medium")
+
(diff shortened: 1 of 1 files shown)


examples/third_party/pygwalker/example.py
```diff
@@ -0,0 +1,31 @@
...
+# ///
+import marimo
+
+__generated_with = "0.9.14"
+app = marimo.App(width="medium")
+
+
+@app.cell
+def __():
+    import pygwalker as pyg
```
Haleshot, line 16: Are you able to achieve rendering the Pygwalker interface using this import? I tried it with datasets hosted through cloud links (S3), tried running this example as is too and it results in the class instance being printed:
`<pygwalker.api.pygwalker.PygWalker object at 0x00000124E362B470>`


I have found using these imports useful:
```
from pygwalker.api.marimo import walk # Usage -> walk(_df, spec="spec.json")
import pygwalker.api.marimo as pyg # Usage -> pyg.walk(_df)
```
this is with regards to the way it was implemented in the [Pygwalker PR](https://github.com/Kanaries/pygwalker/pull/647).

PR #2814, opened 2024-11-07
fix: handle empty or null dates in table

mscolnick:
Safely handle when dates are null or empty


mscolnick:
wait i screwed this up

akshayka (approved):

PR #2817, opened 2024-11-08
feat(image): add caption support to mo.image

### LINKED ISSUE #2732: [mo.image] caption
### Description

add a `caption` arg to `mo.image` as an optional interpretation of the image

### Suggested solution

currently i'm using a syntax like:
```python
mo.vstack(
    [
        mo.image(),
        mo.md().center(),
    ]
)
```
adding a `caption` arg would be cool

### Alternative

_No response_

### Additional context

_No response_

metaboulie:
## 📝 Summary

fixes #2732 

Add optional caption parameter to `mo.image()` that displays descriptive text below the image using the muted foreground color from the theme. Captions are center-aligned and maintain proper spacing.

Screenshot 2024-11-08 at 13 23 16

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


marimo/_plugins/stateless/image.py
```diff
@@ -158,4 +161,18 @@ def image(
         }
     )
     img = h.img(src=resolved_src, alt=alt, style=styles)
+    if caption:
```
akshayka, line 164: This should be

```suggestion
    if caption is not None:
```

based on https://google.github.io/styleguide/pyguide.html#2144-decision

marimo/_plugins/stateless/image.py
```diff
@@ -89,6 +89,7 @@ def _normalize_image(src: ImageLike) -> Image:
 def image(
```
akshayka, line 89: Adding `caption` in the middle of the arguments is a breaking change, because users may be supplying arguments as positional arguments.

`caption` should instead be added as the last argument:

```python
def image(
    src: ImageLike,
    alt: Optional[str] = None,
    width: Optional[int] = None,
    height: Optional[int] = None,
    rounded: bool = False,
    style: Optional[dict[str, Any]] = None,
    caption: Optional[str] = None,
) -> Html
```

akshayka (commented):
Thanks for the PR! A couple of comments.

marimo/_plugins/stateless/image.py
metaboulie: learned, and fixed, thank you :)

akshayka: thank you!

akshayka (approved):
Thanks for the PR! This looks good to me, but I'll let @mscolnick give the final approval tomorrow morning in case he has thoughts on the HTML building.

mscolnick (dismissed):

hkrutzer:
Could also consider [figcaption](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption)

mscolnick:
@metaboulie, i think @hkrutzer's recommendation is a good one, and we can change the div's to figure/figurecaption

metaboulie:
@mscolnick To use figure/figurecaption, I added two methods in `_HTMLBuilder`, hope this is correct

mscolnick (approved):

PR #2825, opened 2024-11-08
fix: handle some sql edge cases for catalogs

mscolnick:
This handles some SQL edges cases (but not all). this handles catalogs/schemas, but still missing multiple CTEs and nested joins.

### DIFF
--- a/tests/_ast/test_sql_visitor.py
+++ b/tests/_ast/test_sql_visitor.py
@@ -9,6 +9,7 @@ from marimo._ast.sql_visitor import (
     SQLDefs,
     SQLVisitor,
     find_sql_defs,
+    find_sql_refs,
 )
 from marimo._dependencies.dependencies import DependencyManager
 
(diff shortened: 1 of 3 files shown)


tests/_ast/test_sql_visitor.py
```diff
@@ -421,3 +422,133 @@ def test_find_created_database() -> None:
...
+        sql = """
+        SELECT * FROM table1;
+        SELECT * FROM table2;
+        """
+        assert find_sql_refs(sql) == ["table1", "table2"]
+
+    @staticmethod
+    def test_find_sql_refs_with_schema() -> None:
+        sql = "SELECT * FROM my_schema.my_table;"
+        assert find_sql_refs(sql) == ["my_schema", "my_table"]
```
akshayka, line 445: Why does `my_table` need to be a ref here?  Couldn't the ref be just the schema (or catalog)?

akshayka: Couldn't you have `my_table` belonging to two different schemas though?

`my_first_schema.my_table`

and

`my_second_schema.my_table`

Creating `my_table` in the second schema shouldn't cause cells referring to `my_table` in the first schema to run, right?

mscolnick: yea ideally not, but right now we don't have any sort of inheritance/scope with variables (just global or not)

mscolnick: to me, the lesser of two evils is the extra dependency rather than no dependency (the app/notebook would fail on startup in this case.)

akshayka (approved):

PR #2826, opened 2024-11-08
fix: check isatty for sandboxing

mscolnick:


marimo/_cli/sandbox.py
```diff
@@ -156,6 +156,11 @@ def prompt_run_in_sandbox(name: str | None) -> bool:
         if GLOBAL_SETTINGS.YES:
             return True
 
+        # Check if not in an interactive terminal (i.e. Docker)
+        # default to False
+        if not sys.stdin.isatty():
+            return False
```
mscolnick, line 162: i think `False` is better for the case for Docker, but not sure about others

akshayka: maybe true for sandbox?

mscolnick: I think `False` is the better default for sandbox IN when running Docker (which is not an interactive termainl)

mscolnick: otherwise you pay installation time on cold-starts which is a bad pattern, and doing `True` would just hide that

akshayka (approved):

PR #2831, opened 2024-11-09
docs: light-theme sidebar color

akshayka:
Use primary background color for sidebar, based on looking at docs for other popular and modern projects.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -112,10 +112,8 @@ html_theme_options = {
     "source_branch": "main",
     "source_directory": "docs/",
     "light_css_variables": {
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -112,10 +112,8 @@
     "source_branch": "main",
     "source_directory": "docs/",
     "light_css_variables": {
-        "color-brand-primary": "#1C7362",
-        "color-brand-content": "#1C7362",
-    },
```
akshayka, line 117: These weren't used since `light_css_variables` had two entries in the dictionary definition.

mscolnick (approved):

PR #2855, opened 2024-11-13
fix: package installation for conflicting names, fix for names with brackets

### LINKED ISSUE #2844: devtools dependency not added to embedded deps when running in sandbox
### Describe the bug

I have a notebook that has embedded deps:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "anyio==4.6.2.post1",
#     "diskcache==5.6.3",
#     "google-api-python-client==2.151.0",
#     "google-cloud-storage==2.18.2",
#     "marimo",
#     "openai==1.52.2",
#     "pandas==2.2.3",
#     "pillow==11.0.0",
#     "psycopg2-binary==2.9.10",
#     "pydantic==2.9.2",
#     "record-type==2023.1.post1",
#     "requests==2.32.3",
# ]
# ///

import marimo

__generated_with = "0.9.17"
app = marimo.App(
    width="medium",
    app_title="...
)
```

Starts fine, and installs the deps listed.

```
 marimo edit --sandbox my_notebook.py
Running in a sandbox: uv run --isolated --no-project --with-requirements /tmp/tmpgna8fd1z.txt marimo edit my_notebook.py
Installed 64 packages in 60ms

        Edit my_notebook.py in your browser 📝

        URL: http://localhost:2718?access_token=XXXXX
```

 In the notebook, I'm using python-devtools (the `devtools` package) to pretty-print a Pydantic model, though, and this dep doesn't seem to be captured. marimo offers to install the wrong package (a turbogears devtools package):

Image

I cancel that and adding the proper `devtools` package manually through the marimo packages interface:

Image

this seems to work:

Image

Image

but it still isn't added to the `dependencies` list in the file header, so have to go through the process again next time the notebook is opened.

1. Why is the `devtools` import matching to the wrong package?
2. Why does the manually-added package show up in the interface's package list but not get added to the file's dep list?

FWIW there seem to be a lot more deps in the marimo interface than in the file. Is that expected? Where does the list in the interface come from? It doesn't seem to match up with the list returned from `pip freeze` either (neither the list returned by my global python/pip, nor with that from the virtualenv in which i have marimo installed)

### Environment

```shell
$ marimo env
{
  "marimo": "0.9.17",
  "OS": "Linux",
  "OS Version": "5.10.102.1-microsoft-standard-WSL2",
  "Processor": "x86_64",
  "Python Version": "3.11.5",
  "Binaries": {
    "Browser": "--",
    "Node": "v18.18.0"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "narwhals": "1.13.3",
    "packaging": "24.0",
    "psutil": "5.9.8",
    "pygments": "2.18.0",
    "pyyaml": "6.0.1",
    "ruff": "0.7.1",
    "starlette": "0.37.2",
    "tomlkit": "0.12.5",
    "uvicorn": "0.29.0",
    "websockets": "12.0"
  },
  "Optional Dependencies": {}
}
```

### Code to reproduce

Start a fresh notebook:

`$ marimo edit --sandbox marimo-devtools-test.py`

Add a single cell:

`import devtools`

Image

mscolnick:
Fixes #2844 

Handles a couple edge cases, which should be clear in the tests

### DIFF
--- a/tests/_runtime/packages/test_package_managers.py
+++ b/tests/_runtime/packages/test_package_managers.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import List
-
 import pytest
 
 from marimo._runtime.packages.package_managers import create_package_manager
@@ -27,7 +25,7 @@ def test_create_package_managers() -> None:
 
 
 def test_update_script_metadata() -> None:
-    runs_calls: List[List[str]] = []
+    runs_calls: list[list[str]] = []
 
     class MockUvPackageManager(UvPackageManager):
         def run(self, command: list[str]) -> bool:
@@ -50,7 +48,7 @@ def test_update_script_metadata() -> None:
 
 
 def test_update_script_metadata_with_version_map() -> None:
-    runs_calls: List[List[str]] = []
+    runs_calls: list[list[str]] = []
 
     class MockUvPackageManager(UvPackageManager):
         def run(self, command: list[str]) -> bool:
@@ -74,7 +72,7 @@ def test_update_script_metadata_with_version_map() -> None:
 
 
 def test_update_script_metadata_with_mapping() -> None:
-    runs_calls: List[List[str]] = []
+    runs_calls: list[list[str]] = []
 
     class MockUvPackageManager(UvPackageManager):
         def run(self, command: list[str]) -> bool:
@@ -102,7 +100,7 @@ def test_update_script_metadata_with_mapping() -> None:
     ]
     runs_calls.clear()
 
-    # It should canonicalize when passed as an import name
+    # It should not canonicalize when passed as an import name
     # and works with brackets
     pm.update_notebook_script_metadata(
         "nb.py", import_namespaces_to_add=["ibis"]
--- a/tests/_runtime/packages/test_package_managers.py
+++ b/tests/_runtime/packages/test_package_managers.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+from typing import List
+
 import pytest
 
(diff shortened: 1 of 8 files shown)


tests/_runtime/packages/test_package_managers.py
```diff
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+from typing import List
```
akshayka, line 3: Since we dropped Python 3.8, we can just use `list` now — `typing.List` is deprecated since Python 3.9.

https://docs.python.org/3/library/typing.html#typing.List

tests/_runtime/packages/test_package_managers.py
```diff
@@ -20,3 +24,96 @@ def test_create_package_managers() -> None:
...
+
+    pm = MockUvPackageManager()
+    # It should not canonicalize when passed explicitly
+    pm.update_notebook_script_metadata("nb.py", packages_to_add=["ibis"])
+    assert runs_calls == [
+        ["uv", "--quiet", "add", "--script", "nb.py", "ibis==2.0"],
+    ]
+    runs_calls.clear()
+
+    # It should not canonicalize when passed as an import name
```
akshayka, line 95:
```suggestion
    # It should canonicalize when passed as an import name
```

?

akshayka (approved):
Just a nit that `typing.List` is deprecated

PR #2878, opened 2024-11-16
improvement: make it clearer/possible to convert to SQL and markdown

mscolnick:
* Allow converting to SQL on an empty cell
* Update the cell action to "Convert to SQL" and "Convert to Markdown", which is clearer and doesn't carry over the `mo.md` and `mo.sql`. Python is still "toggle"


Light2Dark:
Thanks! I think this is really nice. Working as expected from me.

akshayka (approved):

PR #2881, opened 2024-11-16
improvement: if mimebundle only contains markdown/latex, convert to html

mscolnick:
This is what jupyter will do i guess.


marimo/_output/formatters/repr_formatters.py
```diff
@@ -56,6 +56,18 @@ def f_repr(obj: Any) -> tuple[KnownMimeType, str]:
...
+                    for md_mime_type in md_mime_types:
+                        if (
+                            "text/html" not in contents
+                            and md_mime_type in contents
+                        ):
+                            from marimo._output.md import md
+
+                            contents["text/html"] = md(
+                                str(contents[md_mime_type])
+                            ).text
```
mscolnick, line 70: should we then remove text/markdown from the mimebundle?

PR #2883, opened 2024-11-17
[wip] improvement: autocomplete functions

### LINKED ISSUE #2877: Autocomplete not always working
### Describe the bug

Kinda 2 issues I've seen, I think there are some improvements we can make to have the autocomplete powers of a Jupyter notebook.
There are some other issues but I saw on Discord there's some work on a [native vscode plugin](https://discord.com/channels/1059888774789730424/1248732986187448331/1250561528885674014) so I wonder what's the best way forward. Thanks!

1. Doesn't work all the time
Image
This is likely due to `frontend/src/core/codemirror/completion/Autocompleter.ts`
```.ts
  asCompletionResult(
    position: number,
    message: CompletionResultMessage,
  ): CompletionResult {
    console.log("message", message)
    return {
      from: position - message.prefix_length,
      options: message.options.map((option) => {
        return {
          label: option.name,
          type: option.type,
          info: () => constructCompletionInfoNode(option.completion_info),
        };
      }),
      // validFor: /^\w*$/,
    };
  },
```
commenting out the validFor will give proper autocomplete.

2. Not showing docstrings upon '.'
Image
This is due to docstrings_limit at `marimo/_runtime/complete.py#L316`
```.py
def complete(
    request: CodeCompletionRequest,
    graph: dataflow.DirectedGraph,
    glbls: dict[str, Any],
    glbls_lock: threading.RLock,
    stream: Stream,
    docstrings_limit: int = 80, # Here, configuring it to something higher will help
    timeout: float | None = None,
    prefer_interpreter_completion: bool = False,
) -> None:
```

Similar issue: #2080 

### Environment

```
{
  "marimo": "0.9.20",
  "OS": "Darwin",
  "OS Version": "23.6.0",
  "Processor": "arm",
  "Python Version": "3.12.7",
  "Binaries": {
    "Browser": "131.0.6778.69",
    "Node": "v20.11.0"
  },
  "Dependencies": {
    "click": "8.1.3",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.1",
    "markdown": "3.7",
    "narwhals": "1.12.1",
    "packaging": "24.1",
    "psutil": "6.1.0",
    "pygments": "2.18.0",
    "pyyaml": "6.0.2",
    "ruff": "0.6.9",
    "starlette": "0.41.2",
    "tomlkit": "0.13.2",
    "uvicorn": "0.32.0",
    "websockets": "12.0"
  },
  "Optional Dependencies": {
    "duckdb": "1.1.2",
    "pandas": "2.2.3",
    "polars": "1.12.0",
    "pyarrow": "18.0.0"
  }
}
```

### Code to reproduce

_No response_

Light2Dark:
## 📝 Summary

Fixes #2877 
- increase docstring limit so complete docstring hints would show up more often
- remove regex as it doesn't recognize certain inputs. (I'm worried about responsiveness but it's okay locally)
- allow hints to appear in a bracket `def abc(# <---- when the cursor is  here`

## 🔍 Description of Changes

There are still some odd / inconsistent behaviours, so this fix isn't foolproof.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/core/codemirror/completion/Autocompleter.ts
+++ b/frontend/src/core/codemirror/completion/Autocompleter.ts
@@ -60,6 +60,7 @@ export const Autocompleter = {
           info: () => constructCompletionInfoNode(option.completion_info),
         };
       }),
+      validFor: /^\w+$/,
     };
   },
 
--- a/frontend/src/core/codemirror/completion/Autocompleter.ts
+++ b/frontend/src/core/codemirror/completion/Autocompleter.ts
@@ -60,7 +60,6 @@ export const Autocompleter = {
           info: () => constructCompletionInfoNode(option.completion_info),
         };
       }),
-      validFor: /^[\w.]*$/,
     };
   },
 
--- a/frontend/src/core/codemirror/completion/Autocompleter.ts
+++ b/frontend/src/core/codemirror/completion/Autocompleter.ts
@@ -60,7 +60,7 @@ export const Autocompleter = {
           info: () => constructCompletionInfoNode(option.completion_info),
         };
       }),
-      validFor: /^\w*$/,
+      validFor: /^[\w.]*$/,
     };
   },
 
--- a/frontend/src/core/codemirror/completion/completer.ts
+++ b/frontend/src/core/codemirror/completion/completer.ts
@@ -47,9 +47,6 @@ export async function completer(
       documentation: tooltip.html ?? null,
     });
   }
-  if (tooltip) {
-    return null;
-  }
 
   return Autocompleter.asCompletionResult(context.pos, result);
 }
--- a/frontend/src/core/codemirror/completion/completer.ts
+++ b/frontend/src/core/codemirror/completion/completer.ts
@@ -36,7 +36,7 @@ export async function completer(
     return null;
   }
 
-  // If it is a tooltip, show it as a Tooltip instead of a completion
+  // If it is a tooltip, set it in the documentation panel
   const tooltip = Autocompleter.asHoverTooltip({
     position: context.pos,
     message: result,
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -313,7 +313,7 @@ def complete(
     glbls: dict[str, Any],
     glbls_lock: threading.RLock,
     stream: Stream,
-    docstrings_limit: int = 80,
+    docstrings_limit: int = 100,
     timeout: float | None = None,
     prefer_interpreter_completion: bool = False,
 ) -> None:


Light2Dark:
I see there is an issue with some code_completion requests not resolving. It ends up that the autocomplete feels flaky.

`frontend/src/core/codemirror/completion/Autocompleter.ts`
```.ts
export const AUTOCOMPLETER = new DeferredRequestRegistry<
  Omit<CodeCompletionRequest, "id">,
  CompletionResultMessage | null
>(
  "autocomplete-result",
  async (requestId, req) => {
    await sendCodeCompletionRequest({
      id: requestId,
      ...req,
    });
  },
  // We don't care about previous requests
  // so we just resolve them with an empty response.
  // { resolveExistingRequests: () => null }, // for example, commenting this out improves it. Existing reqs are handled
);
```

Playing around with this code, you can see the issues with the autocomplete not working 100% of the time. Still looking into why.
```.py
fig = px.bar(x=["a", "b", "c"], y=[1, 3, 2])
fig.upda
```

frontend/src/core/codemirror/completion/Autocompleter.ts
```diff
@@ -60,7 +60,6 @@ export const Autocompleter = {
           info: () => constructCompletionInfoNode(option.completion_info),
         };
       }),
-      validFor: /^\w*$/,
```
akshayka, line 63: I am concerned about the performance hit from this.

Instead of deleting, can we fix this regex? From looking at the video in your issue, I wonder if ` /^\w+$/` would be better.

akshayka (commented):
Thanks for exploring this — could definitely use improvements. Left a comment.

frontend/src/core/codemirror/completion/Autocompleter.ts
Light2Dark: thanks @akshayka , you are right about the performance hit.
I've been a little stuck because any regex doesn't work well with other libraries, eg polars df
```
df = pl.DataFrame({
    "a": [1,2]
})
df.clo# <---- cursor here doesn't autocomplete
```

I don't know why I'm so stuck unfortunately 🥲

frontend/src/core/codemirror/completion/completer.ts
```diff
@@ -47,9 +47,6 @@ export async function completer(
       documentation: tooltip.html ?? null,
     });
   }
-  if (tooltip) {
-    return null;
-  }
```
Light2Dark, line 52: another thing @akshayka , this removal introduces some side effects. It's not ideal.
```
df = pl.DataFrame({
  "a": [1,2],# <--- cursor here will pop up an autocomplete for Dataframe which overrides Enter key
})
```

I am keen to revert this, an ideal solution is to display as just a tooltip, it would take me too long to implement imo :/

Issue #2435, opened 2024-09-27
Invalid LSP port assignment when custom port is set above 6553


ecornell:
When starting Marimo with a custom port, like 8080, the current code sets the LSP port value to the Marimo port multiplied by 10, potentially setting it to invalid port ranges.

`marimo edit --headless --host 0.0.0.0 --port 8080`

Sep 27 12:27:21 marimo uv[616]: [D 240927 12:27:21 lifespans:65] Setup: lsp
Sep 27 12:27:21 marimo uv[616]: [D 240927 12:27:21 lifespans:82] GitHub Copilot is enabled
Sep 27 12:27:21 marimo uv[616]: [D 240927 12:27:21 sessions:938] Starting LSP server at port **80800**...
Sep 27 12:27:21 marimo uv[616]: [D 240927 12:27:21 sessions:944] ... running command: node /root/.venv/lib/python3.11/site-packages/marimo/_lsp>
Sep 27 12:27:21 marimo uv[616]: [D 240927 12:27:21 sessions:951] ... node process return code (`None` means success): None
Sep 27 12:27:21 marimo uv[616]: [D 240927 12:27:21 sessions:955] Started LSP server at port **80800**

https://github.com/marimo-team/marimo/blob/4b9b293166a5b5667e2c55045fd46ffc27bf0681/marimo/_server/start.py#L100

akshayka:
Thanks. We can fix this

PR #2911, opened 2024-11-20
improvement: proxy lsp through main server

mscolnick:
Closes https://github.com/marimo-team/marimo/issues/2435

This will also fix running in codespaces and other deployed editors.


mscolnick:
need some more testing if this blocks main requests

akshayka (approved):
looks generally good ~

akshayka (commented):
For the type error, in `middleware.py`, if you replace `from websockets.client import connect` with `from websockets import connect`, I think it should work

PR #2912, opened 2024-11-20
example: leafmap

akshayka:
Use leafmap.foliumap, since leafmap.Map is not supported


mscolnick:
I fixed this in leafmap if you upgrade, maybe try upgrading to the latest and it should automatically pick folium

akshayka:
> I fixed this in leafmap if you upgrade, maybe try upgrading to the latest and it should automatically pick folium

Oh thanks. Updated the dependency in the metadata.

Issue #2738, opened 2024-10-29
state example insufficiently explained


gitgithan:
### Documentation is

- [ ] Missing
- [ ] Outdated
- [x] Confusing
- [ ] Not sure?

### Explain in Detail

I can't understand the dataflow of below example from https://docs.marimo.io/guides/state.html and also `marimo tutorial ui` 

I have many questions in bold below.

```
def add_task():
    if task_entry_box.value:
        set_tasks(lambda v: v + [Task(task_entry_box.value)])
        set_task_added(True)

def clear_tasks():
    set_tasks(lambda v: [task for task in v if not task.done])

add_task_button = mo.ui.button(
    label="add task",
    on_change=lambda _: add_task(),
)

clear_tasks_button = mo.ui.button(
    label="clear completed tasks",
    on_change=lambda _: clear_tasks()
)
```
```
task_list = mo.ui.array(
    [mo.ui.checkbox(value=task.done, label=task.name) for task in get_tasks()],
    label="tasks",
    on_change=lambda v: set_tasks(
        lambda tasks: [Task(task.name, done=v[i]) for i, task in enumerate(tasks)]
    ),
)
```

It was hard to understand because for some examples on on_change, it's `lambda v:` while some other examples use `lambda _:`.
**When do we use `lambda v:`, when do we use `lambda _:`, and when do we use both?**

**What does v in  `on_change=lambda v: set_tasks` represent? Is it a list of booleans from the array of checkboxes?**

I tried to debug by adding prints into the `on_change`
```
def debug_type(v):
    print(type(v[0]))
    lambda v: set_tasks(
        lambda tasks: [
            Task(task.name, done=v[i]) for i, task in enumerate(tasks)
        ]
    )
    
task_list = mo.ui.array(
    [
        mo.ui.checkbox(value=task.done, label=task.name)
        for task in get_tasks()
    ],
    on_change=debug_type
)
```
**Why does this make the array only add tasks but can't clear anymore? Even removing the print (but still wrapping in `debug_type` does not clear when clicking clear.**

To understand what `i` is doing, i tried enumerating from 1 instead of default 0, and expected it to break which it did because i guess `v` is limited by the number of checkboxes in the array.
`Task(task.name, done=v[i]) for i, task in enumerate(tasks,1)`.

**However I don't understand why it only broke when I clicked the checkbox, and nothing happens when I click clear?  (i thought clear should break it too since clear is somehow part of this system)** 

[...]

Issue #2921, opened 2024-11-21
mo.ui.code_editor UI improvements


liquidcarbon:
This is useful for using Code Editor cells in `marimo run` (app) mode.

```
sql_editor = mo.ui.code_editor(
    language="sql",
    value=placeholder,
    # min_height=5,         <- 1. this does not do anything
    # debounce=True,        <- 2. new feature requested
    # tab_moves_focus=True  <- 3. new feature requested
)
sql_editor = sql_editor.form(label="### SQL Editor")  # <- 4. on hitting submit (Ctrl+Enter), new line should not be created in editor
```

1. height option doesn't seem to have an affect
2. debounce requested, like in `mo.ui.text`
3. currently Tab changes indentation; with this option, Tab would move focus out of the editor; when `debounce=True` this would effectively run a query
4. If using code editor in a form, Submit can be activated by Ctrl+Enter, but it also creates a new line in the editor.

liquidcarbon:
Illustration of app-mode SQL editor

![Image](https://github.com/user-attachments/assets/d6e159a3-9649-486b-8c84-587547907aad)

mscolnick:
@liquidcarbon - `min_height` is "minimum height of the code editor in pixels". it works fine for me. 5 is too small for a single line

for the other feature requests, would you be interested in contributing those? i think they are great ideas

liquidcarbon:
Aaaah, I didn't even consider the unit of measure - I assumed it height means number of rows, which is how they have it in Gradio.

Alas I have zero experience with TS, so I'll defer to the experts :)

PR #2928, opened 2024-11-22
Add recipe for using lists for state

liquidcarbon:
## 📝 Summary

Example of using lists to hold initial state for triggering computation with a button

## 🔍 Description of Changes

Ideas from https://github.com/marimo-team/marimo/issues/2738#issuecomment-2491924652

## 📋 Checklist

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -426,6 +426,43 @@ import random
 random.randint(0, 1000)
 ```
 
+### Triggering computation from multiple UI elements with initial state
+
+**Use cases.** UI for constructing [complex API requests or database queries](https://gist.github.com/liquidcarbon/8574249103648c4306c22a2e60c486e1). Alternative to `.form()`.
+
+**Recipe.**
+
+1. Import packages
+
+```python
+import marimo as mo
+```
+
+2. Define UI elements
+
+```python
+button = mo.ui.run_button(label="CAPITALIZE")
+text = mo.ui.text()
+```
+
+3. Define initial values
+
+```python
+# this cell won't re-run because Marimo does not mutate objects
+# would not work with TEXT = ""
+TEXT = [""]
+```
+
+4. Run on clicking the button.
+
+```python
+if button.value:
+    TEXT[0] = text.value.upper()
+    
+text, button, TEXT[0]
+```
+
+
 ### Create a counter button
 
 **Use cases.** A counter button, i.e. a button that counts the number of times


dmadisetti:
Hmm, read the thread- but I've always thought this was more of an antipattern. Thought on this?

```python
value = "No joke"

mo.stop(button.value is not None)
value = text.value.upper()

text, button, value
```

liquidcarbon:
That way you'll never see the output, because button's initial value is False

dmadisetti:
Woops! Yes. Edited for correctness.

But this gets around carrying state outside of `mo.State`

There's an experimental "Strict Execution" mode, where this will intentionally fail

mscolnick:
i, too, think this is an anti-pattern because you are modifying state, which we are trying avoid mutations (similar to how [React works](https://react.dev/learn/updating-objects-in-state))

akshayka:
I actually think stashing state in a non-reactive object (such as a Python list) can be simpler than using reactive state. I have noticed when people use reactive state, they have a difficult time reasoning about why cells are running and end up creating bugs that are hard to catch.

With non-reactive state, execution is still determined by the DAG. For this reason I personally prefer this pattern. I hadn't documented it because it does feel "advanced".

akshayka:
The ideal recipe would be something like @dmadisetti's suggestion, though I think it doesn't do quite what @liquidcarbon is trying to do.

liquidcarbon:
In @dmadisetti 's example the form will always reset to this initial value; it's fine in some cases, but I'd like to have a default, pre-populated UI with either 1) examples of how to use it, or 2) from query params

```
value = "No joke" 

mo.stop(not button.value)
value = text.value.upper()

text, button, value  # this is in the next cell, or UI will never show up because initial button.value is False
```

akshayka:
@liquidcarbon, after reflecting I think this is too advanced or niche to add to the recipes — marimo already introduces so many new concepts, that I'm worried our users will get confused. Sorry to lead you astray.

Here is a simpler recipe that I believe does what you need, using `form()` and without using any kind of state.

If it does do what you need, want to change the recipe to this?

https://marimo.app/l/cdiigo

```python
import marimo

__generated_with = "0.9.21"


@app.cell
def __():
    import marimo as mo
    return (mo,)


@app.cell
def __(mo):
    form = mo.ui.text(label="Enter text").form(submit_button_label="capitalize", bordered=False)
    form
    return (form,)


@app.cell
def __(form):
    form.value.upper() if form.value is not None else "Here is my initial value"
    return


if __name__ == "__main__":
    app.run()
```

liquidcarbon:
Sure, your call!  The code is findable in this PR.

I prefer the run_button over the form because it is more customizable (button label, hotkey), also: https://github.com/marimo-team/marimo/issues/2921

akshayka:
@liquidcarbon, sorry I missed your edits. That recipe looks good to me, if you'd like to update the PR.

liquidcarbon:
I'll open a new one.  This latest suggestion is a lot like [gradio examples](https://www.gradio.app/guides/more-on-examples); the original idea was quite different.

PR #2931, opened 2024-11-22
improvement: allow <script src/> in html

mscolnick:
This allows `<script src/>` tags to be properly created. This doesn't open up a new attack vector since this is already possible today with anywidget (running javascript)

cc @koaning

### DIFF
--- a/frontend/src/plugins/core/RenderHTML.tsx
+++ b/frontend/src/plugins/core/RenderHTML.tsx
@@ -39,10 +39,29 @@ const replaceValidIframes = (domNode: DOMNode) => {
   }
 };
 
+const replaceSrcScripts = (domNode: DOMNode): JSX.Element | undefined => {
+  if (domNode instanceof Element && domNode.name === "script") {
+    // Missing src, we don't handle inline scripts
+    const src = domNode.attribs.src;
+    if (!src) {
+      return;
+    }
+    // Check if script already exists
+    if (!document.querySelector(`script[src="${src}"]`)) {
+      const script = document.createElement("script");
+      script.src = src;
+      document.head.append(script);
+    }
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    return <></>;
+  }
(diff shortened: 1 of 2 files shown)


frontend/src/plugins/core/RenderHTML.tsx
```diff
@@ -39,10 +39,29 @@ const replaceValidIframes = (domNode: DOMNode) => {
   }
 };
 
+const replaceSrcScripts = (domNode: DOMNode): JSX.Element | undefined => {
+  if (domNode instanceof Element && domNode.name === "script") {
+    // Missing src, we don't handle inline scripts
```
koaning, line 44: Technically, I guess I might want to in `mohtml` at some point? Maybe via alpine.js or surreal.js.

mscolnick: Maybe, its not hard to add, but tempted to just hold off and see why people need it, if people do, this is the change

```ts
const replaceSrcScripts = (domNode: DOMNode): JSX.Element | undefined => {
  if (domNode instanceof Element && domNode.name === "script") {
    const { src, ...attrs } = domNode.attribs;

    // Handle inline scripts
    if (!src) {
      const script = document.createElement("script");
      // Set any additional attributes
      Object.entries(attrs).forEach(([key, value]) => {
        script.setAttribute(key, value);
      });
      // Set content if any exists
      if (domNode.children?.[0]?.data) {
        script.textContent = domNode.children[0].data;
      }
      document.head.append(script);
      return <></>;
    }

    // Handle src scripts
    if (!document.querySelector(`script[src="${src}"]`)) {
      const script = document.createElement("script");
      script.src = src;
      // Set any additional attributes
      Object.entries(attrs).forEach(([key, value]) => {
        script.setAttribute(key, value);
      });
      document.head.append(script);
    }
    return <></>;
  }
};
```

akshayka (approved):

PR #2939, opened 2024-11-23
fix: add ref handling for useEventListener

Light2Dark:
## 📝 Summary

useEventListener wasn't working properly when you passed in ref.current. For example, the `ArrowUp` & `ArrowDown` keys to move between cells. This is because changes to ref.current will not trigger a re-render.

## 🔍 Description of Changes

useEventListener lets you pass in a ref object instead of ref.current and ensure it re-renders correctly when ref.current is no longer null.

Stumbled upon this as I was trying to add new eventListeners. Shoutout to Claude for helping me understand & write 😅.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/hooks/useEventListener.ts
+++ b/frontend/src/hooks/useEventListener.ts
@@ -17,7 +17,7 @@ function isRefObject<T>(target: T | RefObject<T>): target is RefObject<T> {
 }
 
 export function useEventListener<T extends Target, K extends keyof EventMap<T>>(
-  targetValue: Target | TargetRef,
+  targetValue: T | TargetRef,
   type: K & string,
   listener: (ev: EventMap<T>[K]) => unknown,
   options?: boolean | AddEventListenerOptions,
--- a/frontend/src/hooks/useEventListener.ts
+++ b/frontend/src/hooks/useEventListener.ts
@@ -12,7 +12,9 @@ type EventMap<T extends Target> = T extends Document
       ? WindowEventMap
       : never;
 
-function isRefObject<T>(target: T | RefObject<T>): target is RefObject<T> {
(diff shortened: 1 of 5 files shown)


frontend/src/hooks/useEventListener.ts
```diff
@@ -34,5 +45,5 @@ export function useEventListener<T extends Target, K extends keyof EventMap<T>>(
     return () => {
       target.removeEventListener(type, eventListener, options);
     };
-  }, [type, target, options]);
+  }, [type, targetValue, options]);
```
mscolnick, line 48: If we are passing the RefObject, wont this not trigger re-renders when it changes?

Although if this fixes the underlying issues, I trust this is the right change.

Light2Dark: mutating the ref.current value won't trigger re-renders. But the ref object itself does. I'm basing this because you would normally get a warning when using ref.current in useEffect

`Mutable values like 'ref.current' aren't valid dependencies because mutating them doesn't re-render the component. (react-hooks/exhaustive-deps)`

But I've searched online and see that a common solution is to use `useCallback`, but I think that would require a lot of code modification. Do you think I should look more in this direction?

mscolnick: if this fixes it, i am totally ok with it! i think this was a bug that was already there, but I added ReactCompiler which memoizes a lot such that subtle bugs like this could crop up.

also, you might have fixed this https://github.com/marimo-team/marimo/issues/2940 (thank you!)

mscolnick:
going to merge this! i can use this for another bugfix, thanks @Light2Dark!

PR #2943, opened 2024-11-23
chore: add stricter types for useEventListener

mscolnick:
This does't change any runtime code but adds stricter types so that we can enforce the correct practice that @Light2Dark pointed out (pass the Ref, not the HTMLElement)

cc @Light2Dark


akshayka (approved):

Light2Dark:
tq for completing the original PR! adding typesafety is really helpful

PR #2947, opened 2024-11-24
fix: Altair 5.5 compat

MarcoGorelli:
## 📝 Summary

Currently, `tests/_plugins/ui/_impl/test_altair_chart.py` is failing for DuckDB input, because:
- previously `_data` has already been converted to PyArrow Table
- now, `_data` is still a DuckDB PyRelation, which for now only has "interchange" level support in Narwhals. So, in that case, we convert to PyArrow. We plan to greatly improve DuckDB support next year: https://github.com/narwhals-dev/narwhals/discussions/1370

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/narwhals_table.py
+++ b/marimo/_plugins/ui/_impl/tables/narwhals_table.py
@@ -54,6 +54,10 @@ class NarwhalsTableManager(
         format_mapping: Optional[FormatMapping] = None,
     ) -> bytes:
         _data = self.apply_formatting(format_mapping).as_frame()
+        if nw.get_level(_data) == "interchange":
+            # `write_csv` isn't supported by interchange-level-only
+            # DataFrames, so we convert to PyArrow in this case
+            _data = _data.to_arrow()
         csv_str = _data.write_csv()
         if isinstance(csv_str, str):
             return csv_str.encode("utf-8")
--- a/marimo/_plugins/ui/_impl/tables/narwhals_table.py
+++ b/marimo/_plugins/ui/_impl/tables/narwhals_table.py
@@ -57,8 +57,11 @@ class NarwhalsTableManager(
         if nw.get_level(_data) == "interchange":
             # `write_csv` isn't supported by interchange-level-only
             # DataFrames, so we convert to PyArrow in this case
-            _data = _data.to_arrow()
-        csv_str = _data.write_csv()
+            csv_str = nw.from_native(
+                _data.to_arrow(), eager_only=True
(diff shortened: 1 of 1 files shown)


MarcoGorelli:

marimo/_plugins/ui/_impl/tables/narwhals_table.py
```diff
@@ -54,6 +54,10 @@ def to_csv(
         format_mapping: Optional[FormatMapping] = None,
     ) -> bytes:
         _data = self.apply_formatting(format_mapping).as_frame()
+        if nw.get_level(_data) == "interchange":
+            # `write_csv` isn't supported by interchange-level-only
+            # DataFrames, so we convert to PyArrow in this case
+            _data = _data.to_arrow()
```
mscolnick, line 60: thank you! i assume libraries that are interchange-only force you to install pyarrow?

MarcoGorelli: We don't enforce it (all dependencies are optional in Narwhals), but Altair was previously already converting to PyArrow on the user's behalf, and PyArrow isn't a required dependency in Altair, so anyone upgrading would already have had PyArrow installed

Proper DuckDB support is pretty high on our priority list though, hopefully this step won't be needed for too long

In fact what we could do is:
- already just add `write_csv` for DuckDB and Ibis, even though we don't have a general way to do it for other libraries which are only known to implement the dataframe interchange protocol
- do a try-except here: try just calling `write_csv``, and if `AttributeError` is raised, then we call `to_arrow().write_csv` instead

mscolnick: i was thinking maybe ibis or duckdb would have pyarrow installed. 

the logic you suggested makes sense to me, happy to do that. once that merges, I am happy to add tests for it in `tables/test_narwhals.py`

MarcoGorelli: Ah - Ibis requires PyArrow for all its backends. DuckDB doesn't, although Python DuckDB users probably end up having to install PyArrow at some point anyway

A little complication here is that DuckDB's `to_csv` doesn't accept `None` as an argument (like Polars / pandas / PyArrow do), it only accepts a file name as argument, so I don't think it would be so easy to return a string efficiently, nor do they accept a buffer object. We could write to a temporary file and then read it in Python, but I'd like to explore the perf implications of that

OK with the current solution for now?

PR #2971, opened 2024-11-26
fix: ordered field types

### LINKED ISSUE #2968: The order of the columns in Marino's uitable is very confusing for me
### Describe the bug

I hope Marino's uitable can be displayed in the order of the columns I set.

for example:  I want the order of the columns is ["-1", "0", "1", "2"]
Image

### Environment

Null

### Code to reproduce

import marimo as mo
import pandas as pd

data = {"row_1": [3, 2, 1, 0], "row_2": ["a", "b", "c", "d"]}

pd.DataFrame.from_dict(data, orient="index", columns=["-1", "0", "1", "2"])

mscolnick:
Fixes #2968
dictionaries are ordered by default in python and typescript, but not guaranteed across the network. 

this changes the data-structure of a fieldTypes (column info), so a list instead of a dictionary to preserve order


marimo/_plugins/ui/_impl/tables/default_table.py
```diff
@@ -7,11 +7,12 @@
     List,
     Optional,
     Sequence,
+    Tuple,
```
akshayka, line 10: `typing.Tuple` is deprecated, can just use `tuple` in this file and elsewhere

akshayka (approved):

PR #2478, opened 2024-10-02
added python executable to pip and uv

liquidcarbon:
## 📝 Summary

fixes #2465

## 🔍 Description of Changes

Pointers to runtime python executable were added:
1) pip:
- `pip --python {PY_EXE} install ...`
- `pip --python {PY_EXE} uninstall ...`
- `pip --python {PY_EXE} list`

For pip, the --python option must be placed before the pip subcommand name

2) uv:
- `uv pip install ... -p {PY_EXE} `
- `uv pip uninstall ... -p {PY_EXE} `
- `uv pip list -p {PY_EXE} `

UV allows -p/--python at the end.

PR #2980, opened 2024-11-26
Find ruff when it's not on PATH

liquidcarbon:
## 📝 Summary

Ruff doesn't always go on PATH; if marimo is launched directly via `.venv/bin/marimo`, ruff that sits at `.venv/bin/ruff` will not be found.

## 🔍 Description of Changes

Call ruff using `f"{sys.prefix}/ruff"`

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_utils/formatter.py
+++ b/marimo/_utils/formatter.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import subprocess
+import sys
 from typing import Dict
 
 from marimo import _loggers
@@ -43,7 +44,7 @@ class DefaultFormatter(Formatter):
 class RuffFormatter(Formatter):
     def format(self, codes: CellCodes) -> CellCodes:
         try:
-            process = subprocess.run("ruff", capture_output=True)
+            process = subprocess.run(f"{sys.prefix}/ruff", capture_output=True)
         except FileNotFoundError:
             LOGGER.warning(
                 "To enable code formatting, install ruff (pip install ruff)"
@@ -55,7 +56,7 @@ class RuffFormatter(Formatter):
             try:
                 process = subprocess.run(
                     [
-                        "ruff",
+                        f"{sys.prefix}/ruff",
                         "format",
                         "--line-length",
                         str(self.line_length),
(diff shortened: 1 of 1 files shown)


liquidcarbon:
Not sure if this need more nuance, but given that ruff is marimo's dependency, it should always land next to marimo's executable, and it should be undesirable to run some outside ruff that may be configured differently.

Similar to pip / uv idea in #2478

marimo/_utils/formatter.py
```diff
@@ -43,7 +44,7 @@ def format(self, codes: CellCodes) -> CellCodes:
 class RuffFormatter(Formatter):
     def format(self, codes: CellCodes) -> CellCodes:
         try:
-            process = subprocess.run("ruff", capture_output=True)
+            process = subprocess.run(f"{sys.prefix}/ruff", capture_output=True)
```
akshayka, line 47: I am not confident this will work reliably. 

More reliable might be `subprocess.run(["python", "-m", "ruff"], capture_output=True)`

liquidcarbon: it looks like, the way tests are written, there is in fact no ruff right next to marimo; therefore I'm changing to first check if it exists there, if not, fall back to what it was before (just "ruff")

liquidcarbon:
Not sure what win+py3.9 is doing, the rest seems to be working

marimo/_utils/formatter.py
```diff
@@ -42,8 +45,17 @@ def format(self, codes: CellCodes) -> CellCodes:
 
 class RuffFormatter(Formatter):
     def format(self, codes: CellCodes) -> CellCodes:
+        # first look for ruff next to sys.executable
+        ruff_path = Path(sys.prefix) / "ruff"
```
akshayka, line 49: Shouldn't this be `Path(sys.prefix) / "bin" / "ruff"`? And why is this preferable over ="python -m ruff"`?

akshayka: Okay I understand why not the latter. But how about the former, isn't `bin` necessary?

liquidcarbon: I think that leaves the same ambiguity as to where python is running from as I'm trying to address for ruff.
However, I incorrectly assumed that `Path(sys.prefix) = Path(sys.executable).parent` - you're right that "bin" is missing. My bad.  However, on Windows it's in `Scripts`.

So your suggestion is really optimal and simple here.

akshayka (approved):

akshayka:
Thanks!

Issue #2977, opened 2024-11-26
Wrap list and dictionary output


wasimsandhu:
### Description

The values for my dictionary keys are really long and it's awkward to scroll a string horizontally to read it.

### Suggested solution

Can this view be wrapped by default?

### Alternative

A toggle in the context menu to wrap outputs.

mscolnick:
Without viewing long lists all that much, i feel like not wrapping as the default seems better. It is more often you are aren't look at all the values and i think it would clutter up the output area in every case, instead of just when you want to read it. Open to change my mind if you can share some examples.

wasimsandhu:
I agree that it would make scanning all key-values harder if wrapped by default. A toggle in the right hand corner would be better then.

I can't share my actual data, but I am dealing with tables with 100+ columns, and each column has lots of text. This is what it looks like when I try to view a selected row as a dict:

![Image](https://github.com/user-attachments/assets/384e0145-d4b5-4901-8ca0-107c70483c43)

It would be really helpful if it was wrapped, similar to what viewing a large JSON document in a web page looks like.

mscolnick:
Yea, this can be hard to parse, maybe we can support different views. You can check out https://jsonhero.io/ for some inspiration

wasimsandhu:
I was thinking a "Wrap text" toggle on the right hand corner would suffice for my use case. Can open a PR later this week if so, otherwise we can scope it out some more

PR #2995, opened 2024-11-28
improvement: Wrap output for lists and dictionaries

wasimsandhu:
## 📝 Summary

Completes #2977.

## 🔍 Description of Changes

Add switch to wrap text for list and dictionary (JSON) outputs.

Before:

Screenshot 2024-11-27 at 10 25 32 PM

After:

Screenshot 2024-11-27 at 10 25 16 PM

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/editor/output/JsonOutput.tsx
+++ b/frontend/src/components/editor/output/JsonOutput.tsx
@@ -1,5 +1,5 @@
 /* Copyright 2024 Marimo. All rights reserved. */
-import { memo } from "react";
+import { memo, useState } from "react";
 import {
   type DataType,
   JsonViewer,
@@ -14,6 +14,8 @@ import { TextOutput } from "./TextOutput";
 import { VideoOutput } from "./VideoOutput";
 import { logNever } from "../../../utils/assertNever";
 import { useTheme } from "../../../theme/useTheme";
+import { Labeled } from "@/plugins/impl/common/labeled";
+import { Switch } from "@/components/ui/switch";
 
 interface Props {
   /**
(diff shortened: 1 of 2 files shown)


akshayka (changes requested):
Thanks for the PR. The UI is a bit obtrusive and distracting, and I can imagine it interfering with many people's existing apps.

Some thoughts.

Could the same functionality be done in a much less obtrusive way? Not using the switch component, but making it appear as if it were a native part of the JSON viewer. In fact, does the JSON component we use have this capability?

wasimsandhu:
I'm open to different options! Switch was just the first that came to mind. And we're using [texteainc/json-viewer](https://github.com/TexteaInc/json-viewer), which doesn't seem to have wrapping capability.

wasimsandhu:
Would a right click toggle or context menu option be preferable? A global keyboard shortcut would be even more hidden.

mscolnick (approved):

akshayka:
@wasimsandhu Thanks for the flexibility. Here is what I suggest.

1. Remove the switch component.
2. Wrap strings by default but also ... 
3. Set `collapseStringsAfterLength` to some reasonable number: https://viewer.textea.io/apis

In this way, when a long string is displayed, it won't eat up a ton of vertical space, but when the user expands it, they can still see the full text.

wasimsandhu:
That sounds great! Made the changes. The only problem is that `collapseStringsAfterLength` doesn't work. 🙁 I can investigate later this weekend, but until then I'll change this PR to a draft.

wasimsandhu:
So `collapseStringsAfterLength` won't work with our custom leaf renderers. Got it working by implementing the collapsibility myself. Let me know if you'd like me to move it into its own file, I just left it in the same file.

Screenshot 2024-11-28 at 11 43 19 PM

Screenshot 2024-11-28 at 11 43 29 PM

frontend/src/components/editor/output/JsonOutput.tsx
```diff
@@ -75,6 +75,22 @@ function inferBestFormat(data: unknown): "tree" | "raw" {
...
 }
 
+const CollapsibleTextOutput = (props: { text: string }) => {
+  const [isCollapsed, setIsCollapsed] = useState(true);
+  return (
+    <span className="cursor-pointer">
+      {isCollapsed ? (
+        <span onClick={() => setIsCollapsed(false)}>
+          {props.text.slice(0, 500)}
+          {props.text.length > 500 && "..."}
```
mscolnick, line 85: this changes the default case such that all text is hidden at 500. that is ok w/ me, but just point that out. 

so if you have a lot of keys you want to expand, you'd have to click all of them (maybe not a common case)

wasimsandhu: okay I'll add a comment there. this is making me think that having a dedicated JSON viewer / editor plugin would be nice to have...

akshayka: Yep that's what `mo.tree` is for.

akshayka (approved):
Haven't played with it, but from the screenshot LGTM

ndc33:
hi, this is not working for defaultdict

mscolnick:
@ndc33 thanks for the callout, ive added it here: https://github.com/marimo-team/marimo/pull/4417

ndc33:
that was fast! 
An additional issue with all output is the fixed font size 
On a big monitor with large editor font, the output font stays tiny
I set CSS to scale marimo-json-output as a ratio of --marimo-code-editor-font-size

PR #3000, opened 2024-11-28
improvement: clamp tables at 50 columnns by default

### LINKED ISSUE #2991: Displaying dataframes with many columns can crash the frontend
### Describe the bug

Rendering a dataframe with on the order of 1,000 columns makes my browser tab sluggish, and 10,000 columns crashes the browser tab, regardless of how many rows the dataframe has.

It may make sense to reintroduce a limit on the number of columns that `ui.table` can display.

Related to: https://github.com/marimo-team/marimo/discussions/2899#discussioncomment-11371867

### Environment

0.9.27, Chrome

### Code to reproduce

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "marimo",
#     "numpy==2.1.3",
#     "pandas==2.2.3",
# ]
# ///

import marimo

__generated_with = "0.9.27"


@app.cell
def __():
    import pandas as pd
    return (pd,)


@app.cell
def __():
    import numpy as np
    return (np,)


@app.cell
def __():
    import marimo as mo
    return (mo,)


@app.cell(hide_code=True)
def __(mo):
    mo.md("""Set cols to 10000 to crash the frontend. Number of rows appears to have no effect.""")
    return


@app.cell
def __(mo):
    rows = mo.ui.number(start=1, value=10, label="rows")
    columns = mo.ui.number(start=1, value=10, label="cols")

    mo.hstack([rows, columns], justify="start")
    return columns, rows


@app.cell
def __(columns, np, rows):
    data = np.zeros((rows.value, columns.value))
    return (data,)


@app.cell
def __(data, pd):
    df = pd.DataFrame(data, columns=[str(i) for i in range(data.shape[1])])
    return (df,)


@app.cell
def __(df):
    df
    return


if __name__ == "__main__":
    app.run()
```

### LINKED ISSUE #2996: Rendering dataframes with column summaries can crash browser
_Originally posted by @alexscordellis in https://github.com/marimo-team/marimo/issues/2991#issuecomment-2505031306_

I noticed marimo already (0.9.27) chooses not to send the full dataframe and do the histograms if > 20k rows or > 40 columns so I tried hitting the worst case of 20k rows and 40 columns. This gives about 15MB of csv on the wire, which seems pretty reasonable, but rendering the histograms is very sluggish and has been occasionally enough to give me the "Not enough memory to open this page" chrome error. It seems to be OK the first time I render one such dataframe in a tab, but sometimes crashes later. I haven't been able to pin down exactly when it does. One possible trigger is switching tabs (in Chrome) away and back again. I speculate that the problematic notebooks that I and colleagues have seen at work are ones with this kind of size dataframe, perhaps with multiple such rendered in one notebook. Sorry this is a bit vague and I can't yet give you a deterministic repro.

mscolnick:
Fixes #2996
Fixes #2991

- clamp tables at 50 columnns by default
- delay render of charts until they are visible, this also delays parsing/aggregating the data for the chart

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -205,6 +205,8 @@ class table(UIElement[List[str], Union[List[JSONType], IntoDataFrame]]):
     - `wrapped_columns`: list of column names to wrap
     - `label`: markdown label for the element
     - `on_change`: optional callback to run when this element's value changes
+    - `max_columns`: maximum number of columns to display, defaults to 50.
+      Set to None to show all columns.
     """
 
     _name: Final[str] = "marimo-table"
@@ -220,6 +222,7 @@ class table(UIElement[List[str], Union[List[JSONType], IntoDataFrame]]):
         pagination: Optional[bool] = None,
         selection: Optional[Literal["single", "multi"]] = "multi",
         page_size: int = 10,
+        max_columns: Optional[int] = 50,
         show_column_summaries: Optional[bool] = None,
         format_mapping: Optional[
             Dict[str, Union[str, Callable[..., Any]]]
@@ -255,6 +258,7 @@ class table(UIElement[List[str], Union[List[JSONType], IntoDataFrame]]):
         self._data = data
         # Holds the original data
         self._manager = get_table_manager(data)
+        self._max_columns = max_columns
 
         # Set the default value for show_column_summaries,
         # if it is not set by the user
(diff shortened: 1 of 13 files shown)


marimo/_plugins/ui/_impl/table.py
```diff
@@ -220,6 +222,7 @@ def __init__(
         pagination: Optional[bool] = None,
         selection: Optional[Literal["single", "multi"]] = "multi",
         page_size: int = 10,
+        max_columns: Optional[int] = 50,
```
akshayka, line 225: This should be inserted after the `*` otherwise it'd a breaking change.

mscolnick: Ah yes technically is. I'll fix that in a follow up. For 1.0, we should put most args behind the *

PR #3005, opened 2024-11-29
improvement: warn on integer columns names

mscolnick:


marimo/_plugins/validators.py
```diff
@@ -57,3 +59,21 @@ def warn_js_safe_number(*values: int | float | None) -> None:
                 "integers in JavaScript. This may cause precision issues.",
                 stacklevel=3,
             )
+
+
+def validate_no_integer_columns(df: Any) -> None:
+    if not DependencyManager.pandas.has():
+        return
```
akshayka, line 66: For performance this should be `not DependencyManager.pandas.imported()`, since `df` can't be a Pandas dataframe if Pandas is not imported. Importing pandas can take a long time.

marimo/_plugins/validators.py
```diff
@@ -57,3 +59,21 @@ def warn_js_safe_number(*values: int | float | None) -> None:
                 "integers in JavaScript. This may cause precision issues.",
                 stacklevel=3,
             )
+
+
+def validate_no_integer_columns(df: Any) -> None:
+    if not DependencyManager.pandas.has():
```
mscolnick, line 65:
```suggestion
    if not DependencyManager.pandas.imported():
```

PR #3045, opened 2024-12-03
improvement: force expand markdown content

### LINKED ISSUE #2804: Make the output of markdown cells expanded by default
### Description

Specially for markdown cells, it is undesirable to have the output be collapsed by default, it should be expanded by default. When I write something, I don't want the reader to have to scroll the cell to read it.

### Suggested solution

Make the output of the markdown cells expanded by default. For other kinds of cells, it might be useful to have a way of specifying that we want the output of the cell to be expanded or collapsed by default.

### Alternative

_No response_

### Additional context

Related discussion: https://github.com/marimo-team/marimo/discussions/2765.

mscolnick:
Fixes #2804

When markdown is hidden, we should force expand the output so its not in a scroll bar

### DIFF
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -409,7 +409,10 @@ const CellComponent = (
         />
       </div>
       <OutputArea
+        // Only allow expanding in edit mode
         allowExpand={editing}
+        // Force expand when markdown is hidden
+        forceExpand={isMarkdownCodeHidden}
         output={output}
         className="output-area"
         cellId={cellId}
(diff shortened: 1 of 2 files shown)


frontend/src/components/editor/Cell.tsx
```diff
@@ -409,7 +409,10 @@ const CellComponent = (
         />
       </div>
       <OutputArea
+        // Only allow expanding in edit mode
         allowExpand={editing}
+        // Force expand when markdown is hidden
+        forceExpand={isMarkdownCodeHidden}
```
mscolnick, line 415: this only forces it to be expand when its `markdown + hidden`, instead of just `markdown`. we could easily change this to just markdown if that turns out to be more preferred

akshayka (approved):

PR #3047, opened 2024-12-03
improvement: don't convert classes that extend tuple/dict/list to structure when formatting

mscolnick:
For top-level classes that extend list/dict/tuple, it seems preferable to use the custom repr over converting it to json

image


akshayka (commented):
I think we used to not convert subclasses, but I vaguely recall someone hitting an issue that made us convert subclasses? Maybe I am misremembering

mscolnick:
That was handled here: https://github.com/marimo-team/marimo/pull/1221. and this was just that we _at least_ handle subtypes when flattening. But we can opt-out if a better repr is provided.

PR #3049, opened 2024-12-03
fix: don't unhide when markdown is empty

mscolnick:
Don't unhide the markdown if empty since it may start empty (before any rendering)

### DIFF
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -378,21 +378,6 @@ const CellComponent = (
 
   const hasOutput = !isOutputEmpty(output);
 
-  const unhideCode = useEvent(() => {
-    // Fire-and-forget save
-    void saveCellConfig({ configs: { [cellId]: { hide_code: false } } });
-    updateCellConfig({ cellId, config: { hide_code: false } });
-
-    // focus the editor
-    editorView.current?.focus();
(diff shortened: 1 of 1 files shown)


frontend/src/components/editor/Cell.tsx
```diff
@@ -378,21 +378,6 @@ const CellComponent = (
...
-    // focus the editor
-    editorView.current?.focus();
-    return false;
-  });
-
-  // When markdown cell is hidden & output is cleared, show the code editor
-  if (isMarkdownCodeHidden && !hasOutput) {
-    unhideCode();
-  }
-
```
mscolnick, line 395: cc @Light2Dark, this run's `saveCellConfig` for all markdown that starts hidden (since it has no output). we can add some logic like this back, but for now, users can see the empty editor on hover

Light2Dark: my bad, thanks for fixing this!

Light2Dark: hey @mscolnick , actually what do you mean?
 
If it starts empty, don't we want to show the code instead? Since the output & code will be hidden now. 

https://github.com/user-attachments/assets/3bc76218-b971-461c-bb54-9128ff65edc0

I believe the code I added would fix this

PR #3050, opened 2024-12-04
improvement: add copy/paste to json viewer, support urls

mscolnick:
This adds copy/paste support for floats/sets/lists/dictionaries. This also adds a URL renderer.

**It does not yet support tuples** since those get converted to lists on `json.dump` and not sure I want to make that change yet.

### DIFF
--- a/marimo/_output/formatters/structures.py
+++ b/marimo/_output/formatters/structures.py
@@ -29,6 +29,10 @@ def _leaf_formatter(value: object) -> bool | None | str | int:
         return f"text/plain+float:{value}"
     if value is None:
         return value
+    if isinstance(value, set):
+        return f"text/plain+set:{str(value)}"
+    if isinstance(value, tuple):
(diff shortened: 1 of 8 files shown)


marimo/_output/formatters/structures.py
```diff
@@ -29,6 +29,10 @@ def _leaf_formatter(value: object) -> bool | None | str | int:
         return f"text/plain+float:{value}"
     if value is None:
         return value
+    if isinstance(value, set):
+        return f"text/plain+set:{str(value)}"
+    if isinstance(value, tuple):
+        return f"text/plain+tuple:{json.dumps(value)}"
```
mscolnick, line 35: `isinstance(value, tuple)` does not actually get hit because we don't treat tuples as a leaf (which is the right thing probably at the moment). however, we do treat sets as a leaf

akshayka (approved):
Sweet!

PR #3060, opened 2024-12-04
improvement: tie-break topological sort by top-down order

mscolnick:
This is useful to keep markdown elements (which may not have a dependency) in the same order that they are created, when exporting to a script or ipynb.

This also renders markdown from top to bottom which is less jerky on startup.

### DIFF
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -489,32 +489,36 @@ def get_cycles(
 def topological_sort(
     graph: DirectedGraph, cell_ids: Collection[CellId_t]
 ) -> list[CellId_t]:
-    """Sort `cell_ids` in a topological order.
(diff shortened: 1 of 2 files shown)


marimo/_runtime/dataflow.py
```diff
@@ -489,11 +489,25 @@ def get_cycles(
...
+
+    # Group cells by their parents and sort within each group by registration order
+    roots = sorted(
+        [cid for cid in cell_ids if not parents[cid]],
+        key=lambda x: top_down_keys[x],
+    )
+    sorted_cell_ids: list[CellId_t] = []
     while roots:
+        # Within each level, process cells in registration order
+        roots.sort(key=lambda x: top_down_keys[x])
```
akshayka, line 510: Doing a sort in each iteration could be pretty expensive. Might want to make this a heap: https://docs.python.org/3/library/heapq.html

PR #3080, opened 2024-12-06
improvement: don't send variables if they dont change

mscolnick:
This improves perf a bit from avoiding re-renders when the data is the same.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -14,7 +14,7 @@ import sys
 import threading
 import time
 import traceback
-from copy import copy
+from copy import copy, deepcopy
 from multiprocessing import connection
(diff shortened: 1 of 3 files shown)


marimo/_runtime/runtime.py
```diff
@@ -935,6 +935,8 @@ def mutate_graph(
         LOGGER.debug("Current set of errors: %s", self.errors)
         cells_before_mutation = set(self.graph.cells.keys())
         cells_with_errors_before_mutation = set(self.errors.keys())
+        edges_before = deepcopy(self.graph.children)
+        definitions_before = deepcopy(set(self.graph.definitions.keys()))
```
akshayka, line 939: This deepcopy isn't needed because `set` already copies the `DictKeys` datastructure.

akshayka (dismissed):
nice optimization, small comment

Issue #2325, opened 2024-09-16
Make generated Python files `mypy` error-free


shunichironomura:
### Description

When running `mypy` on marimo notebooks, `mypy` reports `error: Name "__" already defined on line X  [no-redef]` errors because cells in marimo notebooks are all defined as functions with the same name: `__`. This makes it difficult to properly type-check marimo notebooks.

### Suggested solution

There are at least three approaches to resolve this:

1. Define cells as functions with unique names. This approach eradicates the root cause, but possibly has side effects causing other issues.
2. Add `# type: ignore[no-redef]` comments to lines containing `@app.cell` to suppress errors. This approach has probably the fewest side effects but makes the files a bit noisy.
3. Add a top-level `#mypy: disable-error-code="no-redef"` comment. This approach is probably the least noisy but possibly suppresses other `no-redef` errors in the file.

### Additional context

As far as I know, there is no way to configure `mypy` to ignore a certain error in specific files (like `per-file-ignores` config in Ruff) via configuration files such as `pyproject.toml`. So it's impossible to suppress the `no-redef` errors in marimo notebooks without modifying the notebook files.

So my current workaround is to take approach 3 and manually adding the top-level `#mypy: disable-error-code="no-redef"` comment to every marimo notebook.

dmadisetti:
Related is the idea of a marimo specific lint: https://github.com/marimo-team/marimo/issues/1543

Let us know if you come across any other type checking/ static analysis issues particular to mypy and marimo

mscolnick:
@shunichironomura - I dont think this is an alternative or solution, but you can rename cells (in the cell dropdown menu), which is the name given in the generated python file

shunichironomura:
> Related is the idea of a marimo specific lint: #1543
> 
> Let us know if you come across any other type checking/ static analysis issues particular to mypy and marimo

Another mypy/marimo issue that I encountered is that mypy cannot infer the module imported in a different cell. Here's a minimal example:


```python
import marimo

__generated_with = "0.8.15"
app = marimo.App(width="medium")


@app.cell
def __():
    import pathlib
    return pathlib,


@app.cell
def __(pathlib):
    def read_file(file: pathlib.Path) -> str:
        return file.read_text()
    return read_file,


if __name__ == "__main__":
    app.run()
```

Running `mypy` against this file gives:

```
notebooks/mypy_issue.py:13: error: Name "__" already defined on line 7  [no-redef]
[...]

PR #3129, opened 2024-12-11
improvement: generate unique function names to work with linting

### LINKED ISSUE #1954: Can Marimo use unique default names for cells to serialize to a nicer Python file?
### Description

Right now, when I work on a Marimo notebook, I'm delighted that the serialized format is a plain python file. However, by default, each cell is serialized as a function with name "__". This makes for rather funny-looking python files, and if you have any sort of linter in your codebase, it will certainly not be happy.

I can see that if I create a new name for my cell, then it will be used in the serialized python file. I think it would be great if each cell had a unique name, so that the rendered python file inherits those names for the autogenerated functions.

### Suggested solution

When we create a new cell, we should assign a unique name. I've not been able to look into the codebase to see if it would be practical to do something like `cell_0`, `cell_1`, etc, or if a more stateless identifier should be used. Since this is pretty transparent to the user, I would not be too concerned about readability of the name, but if it can be achieved, that would be nice!

I _think_ this behavior is triggered by the default here: https://github.com/marimo-team/marimo/blob/b99ca956cada7aea1e874d4a35d82a7054926b21/marimo/_ast/app.py#L468

I'm not quite sure though!

### Alternative

_No response_

### Additional context

I would be happy to contribute this change myself! I could use some pointers on whether the functionality I'm talking about is coming from where I linked above. If so, I can read up on contributing guidelines, and try to put it together!

mscolnick:
This change makes the code-generation output the cell `def` names as their ~~positions~~ code cell hashes, in order to be unique. 

It doesnt actually matter what gets read, they only update on write.

Fixes #1954, #2325

### DIFF
--- a/marimo/_ast/codegen.py
+++ b/marimo/_ast/codegen.py
@@ -12,7 +12,10 @@ from marimo import __version__
 from marimo._ast.app import App, _AppConfig
 from marimo._ast.cell import CellConfig, CellImpl
 from marimo._ast.compiler import compile_cell
-from marimo._ast.names import default_cell_name_from_code, is_default_cell_name
+from marimo._ast.names import (
+    default_cell_name_from_code,
+    is_internal_cell_name,
+)
 from marimo._ast.visitor import Name
 
 if TYPE_CHECKING:
@@ -176,9 +179,9 @@ def generate_filecontents(
     unshadowed_builtins = set(builtins.__dict__.keys()) - defs
     fndefs: list[str] = []
 
-    # Update default names to be __{4char_hash}
+    # Canonicalize internal names to __{some_hash}
     for idx, name in enumerate(names):
-        if is_default_cell_name(name):
+        if is_internal_cell_name(name):
             names[idx] = default_cell_name_from_code(codes[idx])
 
     for data, name in zip(cell_data, names):
--- a/marimo/_ast/codegen.py
+++ b/marimo/_ast/codegen.py
@@ -12,6 +12,7 @@ from marimo import __version__
 from marimo._ast.app import App, _AppConfig
 from marimo._ast.cell import CellConfig, CellImpl
 from marimo._ast.compiler import compile_cell
+from marimo._ast.names import is_default_cell_name
 from marimo._ast.visitor import Name
 
 if TYPE_CHECKING:
@@ -174,6 +175,12 @@ def generate_filecontents(
 
     unshadowed_builtins = set(builtins.__dict__.keys()) - defs
     fndefs: list[str] = []
+
+    # Update default names to be __{idx}
+    for idx, name in enumerate(names):
+        if is_default_cell_name(name):
+            names[idx] = f"__{idx + 1}"
+
     for data, name in zip(cell_data, names):
         if isinstance(data, CellImpl):
             fndefs.append(to_functiondef(data, name, unshadowed_builtins))
--- a/marimo/_ast/codegen.py
+++ b/marimo/_ast/codegen.py
@@ -12,7 +12,7 @@ from marimo import __version__
 from marimo._ast.app import App, _AppConfig
 from marimo._ast.cell import CellConfig, CellImpl
 from marimo._ast.compiler import compile_cell
-from marimo._ast.names import is_default_cell_name
+from marimo._ast.names import default_cell_name_from_code, is_default_cell_name
 from marimo._ast.visitor import Name
 
 if TYPE_CHECKING:
@@ -176,10 +176,10 @@ def generate_filecontents(
     unshadowed_builtins = set(builtins.__dict__.keys()) - defs
     fndefs: list[str] = []
 
-    # Update default names to be __{idx}
+    # Update default names to be __{4char_hash}
     for idx, name in enumerate(names):
         if is_default_cell_name(name):
-            names[idx] = f"__{idx + 1}"
+            names[idx] = default_cell_name_from_code(codes[idx])
 
     for data, name in zip(cell_data, names):
         if isinstance(data, CellImpl):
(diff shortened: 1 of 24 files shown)


akshayka (commented):
There was a discussion about this on Discord. @dmadisetti landed on position/ordinal naming being not great, which I concurred with. I think this is worth a larger discussion and design.

https://discord.com/channels/1059888774789730424/1059891311190229082/1306680017119477772

mscolnick:
if they are private (prefixed with `__`) then what do they collide with? are you saying people may import these names as `from nb import __3`?

mscolnick:
i see, it could be adding a new cell at the top then increments everything below, causing potential merge conflicts. i will do the cell hash then

marimo/_ast/names.py
```diff
@@ -1,5 +1,13 @@
...
 
 def is_default_cell_name(name: str) -> bool:
-    return name.startswith("__") and (name == "__" or name[2:].isdigit())
+    return name.startswith("__") and (name == "__" or len(name) == 6)
+
+
+def default_cell_name_from_code(code: str) -> str:
+    return (
+        f"{DEFAULT_CELL_NAME}{hashlib.sha256(code.encode()).hexdigest()[:4]}"
+    )
```
akshayka, line 13: I think @dmadisetti's proposal was to use the hashing mechanism developed for caching, but not sure -- will let him comment.

mscolnick: yea that would have the benefit of knowing which cells could change transitively, but that also increases the noise per commit (e.g. adding a new import could change each function name)

mscolnick: it could also be configurable (in future) or part of strict-mode

marimo/_ast/names.py
```diff
@@ -0,0 +1,13 @@
+import hashlib
+
+DEFAULT_CELL_NAME = "__"
+
+
+def is_default_cell_name(name: str) -> bool:
+    return name.startswith("__") and (name == "__" or len(name) == 6)
```
akshayka, line 7: Instead of `is_default_cell_name`, should this be `is_internal_cell_name` with implementation

```python
return name.startswith(DEFAULT_CELL_NAME)
```

That is, let's just treat cell names starting with `__` as "reserved" cell names. I doubt people are using custom names starting with `__` today anyway.

I think that will make migrations easier should we decide to change the representation of the default cell name in the future.

marimo/_ast/names.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def default_cell_name_from_code(code: str) -> str:
+    return (
+        f"{DEFAULT_CELL_NAME}{hashlib.sha256(code.encode()).hexdigest()[:4]}"
```
akshayka, line 12: Should this be 6 characters to decrease the chance of a spurious collision? Birthday paradox is sneaky.

marimo/_ast/codegen.py
```diff
@@ -174,6 +175,12 @@ def generate_filecontents(
 
     unshadowed_builtins = set(builtins.__dict__.keys()) - defs
     fndefs: list[str] = []
+
+    # Update default names to be __{4char_hash}
```
akshayka, line 179: A comment like this could go stale. Instead:

```suggestion
    # Canonicalize internal names to the default cell name based on code
```

tests/_ast/test_codegen.py
```diff
@@ -620,3 +633,10 @@ def test_sqls() -> None:
     cell = compile_cell(code)
     sqls = cell.sqls
     assert sqls == ["SELECT * FROM foo", "ATTACH TABLE bar"]
+
+
+def test_is_default_cell_name() -> None:
+    assert is_default_cell_name("__")
+    assert is_default_cell_name("__1213")
+    assert not is_default_cell_name("__foo")
```
akshayka, line 641: I think this should be considered an internal or "default" cell name too

marimo/_ast/names.py
mscolnick: sounds good

marimo/_ast/names.py
mscolnick: I can increase to six, https://marimo.app/l/d4zs5k

image

akshayka (commented):
I looked into the root issue. For the linters I checked — ruff, mypy, and pyright — simply using `_` (a single underscore) as the cell name gets rid of linter errors. (ruff does not complain on redefinition of double underscore named methods, but mypy and pyright do).

Given this finding, I think we can just use `_` as the default cell name instead of `__`, no need to include hashes. This has the benefit of keeping diffs local to the lines of code that were edited. This also means that `_` becomes a reserved name in marimo notebook files, which I believe to be okay.

marimo/_ast/codegen.py
```diff
@@ -174,6 +178,12 @@ def generate_filecontents(
 
     unshadowed_builtins = set(builtins.__dict__.keys()) - defs
     fndefs: list[str] = []
+
+    # Canonicalize internal names to __{some_hash}
+    for idx, name in enumerate(names):
+        if is_internal_cell_name(name):
+            names[idx] = default_cell_name_from_code(codes[idx])
```
dmadisetti, line 185: Could use

```python
if isinstance(cell_data[idx], CellImpl):
    marimo._save.hash.hash_cell_impl(cell_data[idx])
```

which would be robust to comment / whitespace changes

dmadisetti (commented):
Collisions are not likely, but not unimaginable:

image

Think it's fine. 3% chance of collision at 1000 cells

dmadisetti:
Just seeing @akshayka's comment, that's a great idea
I think that is more in line with being source control friendly

mscolnick:
superseded by #3143

PR #3143, opened 2024-12-12
improvement: single '_' function names to work with linting

### LINKED ISSUE #1954: Can Marimo use unique default names for cells to serialize to a nicer Python file?

### LINKED ISSUE #2325: Make generated Python files `mypy` error-free

mscolnick:
This change makes the code-generation output mypy/pyright compliant

Fixes #1954
Fixes #2325


dmadisetti (commented):
Changes look pretty straight forward, 2 thoughts:

1. Is it worth preventing the user from naming a cell _ or __? Or does that restriction already exist
2. If mypy is a standard that marimo wants to hold itself to, it might be worth adding tests to check some of the generated files: https://mypy.readthedocs.io/en/stable/extending_mypy.html these would also be a starting point for adding any other top level file changes

Otherwise +1 !

akshayka (approved):

PR #3150, opened 2024-12-12
improvement: errors should stay internal in marimo run and logged to console

mscolnick:
This hides stack traces when in run mode, and instead propagates up an `error_id` that can be used to grep the server logs.

Screenshot 2024-12-12 at 4 39 47 PM

Screenshot 2024-12-12 at 4 39 39 PM

Screenshot 2024-12-12 at 5 06 01 PM

### DIFF
--- a/marimo/_messaging/errors.py
+++ b/marimo/_messaging/errors.py
@@ -107,6 +107,50 @@ class MarimoStrictExecutionError:
         return self.msg
 
 
+@dataclass
+class MarimoInternalError:
+    """
+    An internal error that should be hidden from the user.
+    The error is logged to the console and then a new error is broadcasted
+    such that the data is hidden.
+
+    They can be linked back to the original error by the error_id.
+    """
+
+    error_id: str
+    type: Literal["internal"] = "internal"
+    msg: str = ""
+
+    def __post_init__(self) -> None:
+        self.msg = f"An internal error occurred: {self.error_id}"
+
+    def describe(self) -> str:
+        return self.msg
+
+
+def is_unexpected_error(error: Error) -> bool:
+    """
+    These errors are unexpected, in that they are not intentional.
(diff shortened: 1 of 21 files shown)


marimo/_messaging/errors.py
```diff
@@ -107,6 +107,50 @@ def describe(self) -> str:
...
+    mo.stop and interrupt are intentional.
+    """
+    return error.type not in [
+        "ancestor-prevented",
+        "ancestor-stopped",
+        "interruption",
+    ]
+
+
+def is_sensitive_error(error: Error) -> bool:
```
akshayka, line 143: This seems to be the same as `is_unexpected_error`?

marimo/_messaging/errors.py
```diff
@@ -107,6 +107,50 @@ def describe(self) -> str:
...
+    """
+    These errors are sensitive, in that they are intentional.
```
akshayka, line 145: Do you mean "not intentional"?

marimo/_messaging/ops.py
```diff
@@ -248,12 +254,32 @@ def broadcast_error(
         cell_id: CellId_t,
     ) -> None:
         console: Optional[list[CellOutput]] = [] if clear_console else None
+
+        # In run mode, we don't want to broadcast the error. Instead we want to print the error to the console
+        # and then broadcast a new error such that the data is hidden.
+        safe_errors: list[Error] = []
```
akshayka, line 260: I can imagine a use case for which broadcasting still makes sense. But I guess this is okay as a default, and we can make it configurable in the future if people ask.

akshayka (approved):

marimo/_messaging/errors.py
mscolnick: They are slightly different. One is interrupt and the other is internal

PR #3152, opened 2024-12-12
feat: markdown caption support, tabs, critic, emoji, magic links

mscolnick:
Enable more markdown features. 

base64 is not enabled (but the code is there as a reference) until we come up with a better approach for it. e.g only allow files in a public/ or assets/ directory

markdown_pymdownx


akshayka (approved):
Occurred to me that at some point we should document the enabled extensions (unless we don't want to commit to supporting them) 

https://docs.marimo.io/guides/outputs.html#markdown

PR #3154, opened 2024-12-13
fix: higher-order components should call `on_update_completion()`

### LINKED ISSUE #3141: multiple run_buttons triggered simulateneously when in mo.ui.dictionary
### Describe the bug

I have a an marimo dictionary that holds several run_buttons. I want to use the dictionary because I want to trigger different logic based on which run_button I press, and I want to generate the run_buttons at runtime (as the number can vary, depending on user input).

I found that when I trigger the first button ("run button 1", in the example below), the cell that depends on that button is run twice. Next, if I trigger the second run button ("run button 2"), both buttons are registered as pressed (I can see, because their "value" property is set to True). I guess that's not the intended functionality. I was under the impression that the "value" should be reset to "False" once consumed.

### Environment

{
  "marimo": "0.9.34",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.2",
  "Binaries": {
    "Browser": "129.0.6668.60",
    "Node": "v20.14.0"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.20.1",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "narwhals": "1.14.0",
    "packaging": "24.0",
    "psutil": "5.9.8",
    "pygments": "2.17.2",
    "pymdown-extensions": "10.7.1",
    "pyyaml": "6.0.1",
    "ruff": "0.5.2",
    "starlette": "0.37.2",
    "tomlkit": "0.12.4",
    "uvicorn": "0.29.0",
    "websockets": "11.0.3"
  },
  "Optional Dependencies": {
    "pandas": "2.2.2"
  }
}

### Code to reproduce

Here's a simple example: Press the "run button 1" and observe the consuming cell prints output twice. Next press "Run button 2" and observe that both buttons are registered as the cell prints "Clicked..." twice (once for button1 and once for button 2). I included a single run_button, to check how that behaves, and that's as expected.

```
import marimo

__generated_with = "0.9.34"


@app.cell
def __():
    import marimo as mo
    from time import time
    return mo, time


@app.cell
def __(mo):
    d=mo.ui.dictionary({"run_button1": mo.ui.run_button(label="run button 1"), "run_button2": mo.ui.run_button(label="run button 2")})
    return (d,)


@app.cell
def __(d):
    d
    return


@app.cell
def __(d, time):
    _t=time()
    for name, button in d.items():
        if button.value:
            print(f"Clicked: at {_t}", name)
    return button, name


@app.cell
def __(mo):
    run_button = mo.ui.run_button(label="single run button")
    run_button
    return (run_button,)


@app.cell
def __(run_button, time):
    if run_button.value:
        print("Clicked run button at time ", time())
    return


if __name__ == "__main__":
    app.run()

### PR DESCRIPTION
**Pull request by akshayka**
Higher order components like dicts and arrays need to call their children's completion hooks after their value is updated.

Fixes #3141

### DIFF
--- a/marimo/_config/reader.py
+++ b/marimo/_config/reader.py
@@ -1,3 +1,4 @@
+# Copyright 2024 Marimo. All rights reserved.
 from pathlib import Path
 from typing import Any, Dict, Optional, Union, cast
 
--- a/marimo/_config/secrets.py
+++ b/marimo/_config/secrets.py
@@ -1,3 +1,4 @@
+# Copyright 2024 Marimo. All rights reserved.
 from typing import Any, Dict, TypeVar, Union, cast
 
 from marimo._config.config import MarimoConfig, PartialMarimoConfig
--- a/marimo/_plugins/ui/_core/ui_element.py
+++ b/marimo/_plugins/ui/_core/ui_element.py
@@ -461,9 +461,12 @@ class UIElement(Html, Generic[S, T], metaclass=abc.ABCMeta):
         if self._on_change is not None:
             self._on_change(self._value)
 
-    def _on_update_completion(self) -> None:
-        """Callback to run after the kernel has processed a value update."""
-        return
+    def _on_update_completion(self) -> bool:
+        """Callback to run after the kernel has processed a value update.
+
+        Return true if the value of the component has changed, false otherwise
+        """
+        return False
 
     def __deepcopy__(self, memo: dict[int, Any]) -> UIElement[S, T]:
         # Custom deepcopy that excludes elements that can't be deepcopied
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by mscolnick** [comment]
```
marimo/_plugins/ui/_impl/batch.py:99: error: Return type "None" of "_on_update_completion" incompatible with return type "bool" in supertype "UIElement"  [override]
marimo/_plugins/ui/_impl/array.py:127: error: Return type "None" of "_on_update_completion" incompatible with return type "bool" in supertype "UIElement"  [override]
```

mscolnick (approved):
nice!

PR #3166, opened 2024-12-13
improve: marimo export html-wasm CLI message

akshayka:
Tell users how to test the notebook locally after export


liquidcarbon:
I'm able to open and run as a regular HTML file in the browser, without httpserver.  This did not work in 0.10.0

PR #3229, opened 2024-12-19
docs: add CLI to the docs, reformat CLI so renders propperly

mscolnick:
This adds our CLI docs to docs.marimo.io. 

I needed to reformat our CLI so that they get rendered as markdown properly


marimo/_cli/cli.py
```diff
@@ -477,8 +479,7 @@ def new(
 
 Example:
 
-  \b
-  * marimo run notebook.py
+    marimo run notebook.py
```
mscolnick, line 482: this makes is a code block

Issue #2293, opened 2024-09-10
app.function: top-level functions in notebook files


akshayka:
This issue is for making it possible to import functions directly from marimo notebooks:

```python
from notebook import my_function
```

Only "pure" functions will be importable, ie functions that have no refs besides imported modules. Some care will need to be taken to handle modules.

A cell that includes a single function def, and nothing more, will be saved in the notebook file as:

```python
@app.function
def my_function():
  ...
```

Discussed on:
- Discord Discussion: https://discord.com/channels/1059888774789730424/1059891311190229082/1253813629317288108
- GitHub Discussion: https://github.com/marimo-team/marimo/discussions/2288

patrick-kidger:
I worry that privileging *just* single-cell functions is going to be a huge UX footgun. There are many other things I might reasonably want the same behaviour from. (Classes are the obvious example.) Moreover I will almost always want my function to depend upon other values:
```python
SOME_CONSTANT = 42

def the_answer_to_life_the_universe_and_everything() -> int:
    return SOME_CONSTANT
```
Or -- I may simply write code that depends upon being able to import a function like this... and now find that I am unable to adjust this function's behaviour without breaking that import resolution.

---

Can I make an alternate suggestion? Marimo supports adding meaningful names for cells, so it could be arranged for the syntax `import my_notebook; my_notebook.my_cell.my_function` to work.

In particular this would:
- Automatically evaluate all cells that `my_cell` depends upon.
- Set `__module__` and `__qualname__` appropriately, so that even dynamic resolution via `importlib.import_module` works correctly.
- ... and as such, basically support the full generality of things that we may wish to do in Python.

akshayka:
> Or -- I may simply write code that depends upon being able to import a function like this... and now find that I am unable to adjust this function's behaviour without breaking that import resolution.

Oh, that's a very good point; that kind of action at a distance would be sad. I hadn't considered this.

Regarding

> so it could be arranged for the syntax import my_notebook; my_notebook.my_cell.my_function to work

It's definitely an interesting idea, and would be readily done by abstracting over the low-level API we have for programmatically running cells: https://docs.marimo.io/api/cell.html. AFAIK right now I'm the only consumer of that API, it's quite low level.

I've considered what you've suggested before, but the downside is that you won't get code completion, type information, etc. when typing out

```
`my_notebook.my_cell.my_function
```

whereas top-level functions would play well with an LSP and other static analysis tools.

[...]

PR #3238, opened 2024-12-19
tidy: Tweaks to enable pytest on marimo nbs

dmadisetti:
I was really interested in having tests in notebooks (re discussion yesterday), so I played around with it a tiny bit this afternoon. As is, pytest does not work out of the box since:
 1. the wrapped functions become Cell instances
 2. coroutine detection through the `_is_coroutine` spec should be as an attribute

This change enables both of these minor adjustments. Now, pytest will pick up
marimo `test_*` functions, but they will fail since directly invoking a cell is
currently not allowed. I put in a little unit test to fight against a possible
regression until such time when a top level fn spec is enabled (see #2293).

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -1,6 +1,7 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
+import functools
 import inspect
 import random
 import string
@@ -464,6 +465,8 @@ class CellManager:
             )
             cell._cell.configure(cell_config)
             self._register_cell(cell, app=app)
+            # NB. in place metadata update.
+            functools.wraps(func)(cell)
             return cell
 
         if func is None:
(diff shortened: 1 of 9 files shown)


mscolnick (dismissed):
wow this is a cool awesome find. would be amazing to support `test_` cells

akshayka (commented):
Thanks! The cell execution API could be improved. Could consider lifting the restriction on calling cells directly, just found it weird because you can call them without passing any of what appear to be the required positional args.

Noticed a test is failing?

dmadisetti:
Yes, I think the test is related- but not totally obvious how. Digging in now.

> Could consider lifting the restriction on calling cells directly, just found it weird because you can call them without passing any of what appear to be the required positional args.

I think it might be useful to just allow a direct call if the correct args are passed in, but require "run" if used in the context of the notebook.

akshayka:
> I think it might be useful to just allow a direct call if the correct args are passed in,

In the case of `pytest`, wouldn't the framework call the cell without arguments, relying on marimo to compute the ancestors instead?

PR #3247, opened 2024-12-19
improve: parse google-style attributes in docstrings

akshayka:
Also, show examples right after summary, then args, attrs, ...

I noticed that some (but not all) of the attributes sections had been converted to google-style and weren't rendering properly. This change fixes that.

### DIFF
--- a/marimo/_utils/docs.py
+++ b/marimo/_utils/docs.py
@@ -1,3 +1,4 @@
+# Copyright 2024 Marimo. All rights reserved.
 import re
 from textwrap import dedent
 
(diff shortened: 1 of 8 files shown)


marimo/_utils/docs.py
```diff
@@ -25,14 +26,39 @@ def google_docstring_to_markdown(docstring: str) -> str:
...
     returns_table: list[tuple[str, str]] = []
     raises_table: list[tuple[str, str]] = []
     in_args = False
+    in_attributes = False
     in_returns = False
     in_raises = False
     in_examples = False
     examples_lines: list[str] = []
 
+    def _handle_arg_or_attribute(
```
mscolnick, line 39: thank you for DRYing this up!

mscolnick (approved):

PR #3251, opened 2024-12-20
feat: make cells runnable with full arguments

dmadisetti:
> In the case of pytest, wouldn't the framework call the cell without arguments, relying on marimo to compute the ancestors instead?

Not if fixtures are provided. Pretty easy change, but marking as draft because I think it might dramatically expand surface area, and happy to discuss more

## 📜 Reviewers

@akshayka OR @mscolnick


dmadisetti:
Or if top level functions and imports come into play, then a test cell is just like any other test function, just embedded within the notebook.

akshayka (commented):
I'm good with this change, I don't think the surface area is that much higher actually.

That said, here is an alternative proposal to consider. Thoughts on checking if running under `pytest` in `__call__`, and if so just forwarding the call to `cell.run()`? And if not, raising an error as we do today?


I think this would make it possible to run

```
pytest my_notebook.py
```

and the test cells would automatically be executed, without the need for defining fixtures.

mscolnick (approved):

akshayka:
@dmadisetti thoughts on implementing the alternative I suggested instead?

dmadisetti:
Hmm, my previous test workflow seems a little stale- sorry for the churn here.
I think that sounds ideal- since it's what users will most likely want.

I still like the idea of being able to use fixtures- but leaving it in adds a tiny bit more complexity in spoofing the signature. As is, fixtures need a "_fixture" suffix, otherwise the signature is stripped, and run as if it's invoked from `cell.run`.

akshayka (approved):
This is awesome!

image

PR #3253, opened 2024-12-20
Add mapping for quaternion -> numpy-quaternion

moble:
## 📝 Summary

This adds a mapping for the `quaternion` package, which is named `numpy-quaternion` on PyPI.

## 🔍 Description of Changes

Just added the key:value pair to the `module_name_to_pypi_name` dict.  I'm hoping a test is not needed for such a simple change, though I have tried it locally and verified that it works by automatically installing `numpy-quaternion`.

## 📋 Checklist

- [X] I have read the [contributor guidelines]([[https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md]]).

## 📜 Reviewers

@mscolnick

### DIFF
--- a/marimo/_runtime/packages/module_name_to_pypi_name.py
+++ b/marimo/_runtime/packages/module_name_to_pypi_name.py
@@ -639,6 +639,7 @@ def module_name_to_pypi_name() -> dict[str, str]:
         "pyutilib": "PyUtilib",
         "pyximport": "Cython",
         "qs": "qserve",
+        "quaternion": "numpy-quaternion",
         "quickapi": "django-quickapi",
         "quickunit": "nose-quickunit",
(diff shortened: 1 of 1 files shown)


moble:

moble:
recheck

akshayka (approved):
While there is also a package called quaternion, it looks like numpy-quaternion is more commonly used and more actively developed.

So, looks good, thank you!

moble:
> While there is also a package called quaternion, it looks like numpy-quaternion is more commonly used and more actively developed.

And, in fact, importing that other package requires `import Quaternion`, rather than `import quaternion`, so even though the default mapping would `pip install quaternion`, but then error when trying to actually import it.  Gotta love PyPI!

PR #3254, opened 2024-12-20
examples: ui element usage examples

akshayka:
Adds a collection of small self-contained examples for our various UI elements. In a follow-up, it might make sense to have our API docs point to these examples as a the source of truth (instead of inlining `marimo-embed`).

Adds a pointer to these examples and the API docs to `marimo tutorial intro`.


mscolnick:
looks great. 

@akshayka we could easily create a new mkdocs block: `/// marimo-embed-file | examples/ui/array.py`

mscolnick (approved):

PR #3262, opened 2024-12-20
improvement: map jupyter magic to helpful descriptions

mscolnick:
When the user types something that matches a common jupyter command, we potentially toast with helpful message (and maybe update some config or open a panel)

### DIFF
--- a/frontend/src/core/codemirror/compat/jupyter.tsx
+++ b/frontend/src/core/codemirror/compat/jupyter.tsx
@@ -120,7 +120,7 @@ export function jupyterHelpExtension(): Extension {
       ),
     },
     {
-      match: "!ls ",
+      match: "!ls",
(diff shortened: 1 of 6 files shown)


frontend/src/core/codemirror/compat/jupyter.tsx
```diff
@@ -0,0 +1,161 @@
...
+      match: "!ls",
+      onMatch: () => {
+        // noop
+      },
+      title: "Listing files",
+      description: (
+        <>
+          Shell commands are not directly supported.
+          <br />
+          Use <code>sys.subprocess(['ls'])</code> instead.
```
akshayka, line 132: This should be: `subprocess.run([`ls`])`

akshayka (dismissed):

PR #3266, opened 2024-12-21
docs: warn users more strongly against using mo.state() and on_change handlers.

akshayka:
Too many of our users fall into callback driven programming which is very ill-suited to marimo. This PR updates the docs to more strongly warn users against using mo.state()/on_change.

### DIFF
--- a/docs/api/state.md
+++ b/docs/api/state.md
@@ -1,10 +1,13 @@
 # State
 
-!!! warning "Advanced topic!"
+!!! warning "Stop! Read the interactivity guide first!"
+    **Read the guide on [creating interactive
+    elements](../guides/interactivity.md)** before reading this one!
 
-    This API doc covers reactive state (`mo.state`), an advanced topic.
+!!! warning "Advanced and dangerous topic!"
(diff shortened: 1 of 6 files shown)


docs/api/state.md
```diff
@@ -1,36 +1,35 @@
 # State
 
-!!! warning "Advanced topic!"
+!!! warning "Stop! Read the interactivity guide first!"
+    **Read the guide on [creating interactive
+    elements](../guides/interactivity.md)** before reading this one!
 
-    This API doc covers reactive state (`mo.state`), an advanced topic.
+!!! warning "Advanced and dangerous topic!"
```
mscolnick, line 7: "Dangerous" seems a bit aggressive. Maybe like: "you likely don't need this"

akshayka: Yeah, fair enough. I don't know what a "dangerous topic" is. I grew a bit weary of seeing `mo.state()` used in the wild, and really just want to deter all but the most determined from using it. Updated.

mscolnick (approved):

PR #3274, opened 2024-12-21
improvement: handle pl.Duration

### LINKED ISSUE #3267: Support pl.Duration in dataframe viewer
### Description

First of all, most sincere thanks for building this and for sharing your work.
There is no doubt that your work is high quality and benefits way more people than you may have set out to, by simply removing developer friction.

Hopefully this is a minor item that can be picked up at your leisure - 
When viewing a dataframe that has `pl.Duration` columns, the table viewer doesn't know about these columns.  
It shows `Invalid Date-Invalid Date`, as seen below.

Image

### Suggested solution

Make the table code aware of pl.Duration.

### Alternative

_No response_

### Additional context

Unrelated but just adding here as a general comment:  
I'm sure you're already aware of [Great Tables](https://posit-dev.github.io/great-tables/articles/intro.html).  
Please see if it helps in any part of the marimo experience.

mscolnick:
Fixes #3267 

This handles narwhals and polars Duration types, by treating them as `floats` and not temporal. We differ in the that we don't show the units for each entry (just at the top).

For pandas, we treat them as strings (with the same formatting as their plain HTML)

Screenshot 2024-12-21 at 2 02 30 PM

Screenshot 2024-12-21 at 2 02 33 PM

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/narwhals_table.py
+++ b/marimo/_plugins/ui/_impl/tables/narwhals_table.py
@@ -215,18 +215,21 @@ class NarwhalsTableManager(
                 p75=col.quantile(0.75, interpolation="nearest"),
                 p95=col.quantile(0.95, interpolation="nearest"),
             )
-        if col.dtype == nw.Duration:
+        if col.dtype == nw.Duration and isinstance(col.dtype, nw.Duration):
             unit_map = {
-                "ms": col.dt.total_milliseconds,
-                "ns": col.dt.total_nanoseconds,
-                "us": col.dt.total_microseconds,
+                "ms": (col.dt.total_milliseconds, "ms"),
+                "ns": (col.dt.total_nanoseconds, "ns"),
+                "us": (col.dt.total_microseconds, "μs"),
+                "s": (col.dt.total_seconds, "s"),
             }
+            method, unit = unit_map[col.dtype.time_unit]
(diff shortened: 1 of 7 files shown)


marimo/_plugins/ui/_impl/tables/narwhals_table.py
```diff
@@ -212,6 +215,19 @@ def _get_summary_internal(self, column: str) -> ColumnSummary:
...
+                "ns": col.dt.total_nanoseconds,
+                "us": col.dt.total_microseconds,
+            }
+            return ColumnSummary(
+                total=total,
+                nulls=col.null_count(),
+                min=str(unit_map["us"]().min()) + "μs",
+                max=str(unit_map["us"]().max()) + "μs",
+                mean=str(unit_map["us"]().mean()) + "μs",
+            )
```
MarcoGorelli, line 230: it might be worth factoring out the `total_*` call, so it doesn't get recalculated?
```python
            duration = unit_map["us"]()
            return ColumnSummary(
                total=total,
                nulls=col.null_count(),
                min=str(duration.min()) + "μs",
                max=str(duration.max()) + "μs",
                mean=str(duration.mean()) + "μs",
            )
```

akshayka (approved):
Lgtm modulo marco's comment

PR #3289, opened 2024-12-24
chore: use polars instead of pandas in examples

akshayka:
* Migrate the SQL examples to use Polars.
* Can merge when polars in pyodide is supported

### DIFF
--- a/examples/sql/connect_to_motherduck.py
+++ b/examples/sql/connect_to_motherduck.py
@@ -3,8 +3,9 @@
 # dependencies = [
 #     "altair==5.4.1",
 #     "duckdb==1.1.0",
+#     "polars==1.18.0",
+#     "pyarrow==18.1.0",
 #     "marimo",
-#     "pandas==2.2.3",
 # ]
 # ///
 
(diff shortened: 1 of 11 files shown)


examples/sql/connect_to_motherduck.py
```diff
@@ -3,8 +3,9 @@
 # dependencies = [
 #     "altair==5.4.1",
 #     "duckdb==1.1.0",
+#     "polars==1.18.0",
+#     "pyarrow==18.1.0",
```
Light2Dark, line 7: hey @akshayka , would it be better to use "polars[pyarrow]>=1.18.0"
https://duckdb.org/docs/guides/python/polars.html

akshayka: Thanks! Will keep this in mind for the future. I think serializing both accomplishes the same thing?

Light2Dark: yes it does. Maybe just a small thing for future maintenance since we don't need to update the pyarrow version. But yeah it's fine.

PR #3294, opened 2024-12-26
feat: apply pytest rewrites for better testing

dmadisetti:
merry merry 🎄 

---

Slight improvement on pytest errors, by manually running pytest's assertion rewrite magic. So that takes us from this:

image

to this

image

Also a little fun- this rewrite can be applied in general and shown in editor.

image

Not sure the overhead this might have- so not sure about setting it generally, but could set `test_rewrite=True` if the cell name starts with `test_`. Wanted to run this by you all first

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_ast/compiler.py
+++ b/marimo/_ast/compiler.py
@@ -12,6 +12,7 @@ import token as token_types
 from tokenize import TokenInfo, tokenize
 from typing import TYPE_CHECKING, Any, Callable, Optional
 
+from marimo import _loggers
 from marimo._ast.cell import (
     Cell,
     CellId_t,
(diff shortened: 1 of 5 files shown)


marimo/_ast/compiler.py
```diff
@@ -163,6 +167,20 @@ def compile_cell(
         # since there is an actual file to read from.
         cache(filename, code)
 
+    # pytest assertion rewriting, gives more context for assertion failures.
+    if test_rewrite:
+        # pytest is not required, so fail gracefully if needed
+        try:
+            from _pytest.assertion.rewrite import (  # type: ignore
```
akshayka, line 174: This is a private API. If no public API can be used, let's handle a much more generic exception than `ImportError` (just `Exception`)

akshayka (commented):
Nice! 1 comment.

dmadisetti:
Cool!

Thoughts on exposing the interface to this error type?

akshayka (approved):
LGTM! I don't think we need to expose an interface to errors in the pytest rewrite.

Issue #3299, opened 2024-12-27
Cell.run() does not supports dict and dataframe arguments


benbaja:
### Describe the bug

Hello,
I was planning to use the newest testing feature for my notebook, but unfortunately i'm facing an error when trying to run cells from another notebook dedicated to testing.
Using the Cell.run() function throws a TypeError This cell raised an exception: TypeError('unhashable type: 'DataFrame' / 'dict') when using a pandas DataFrame or a dictionary as one of the kwargs.

<img width="828" alt="Image" src="https://github.com/user-attachments/assets/eb09787d-1dff-4a89-8db3-ce624157888f" />

### Environment

{
  "marimo": "0.10.7",
  "OS": "Darwin",
  "OS Version": "23.6.0",
  "Processor": "arm",
  "Python Version": "3.13.0",
  "Binaries": {
    "Browser": "131.0.6778.205",
    "Node": "v23.2.0"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.18.4",
    "packaging": "24.2",
    "psutil": "6.1.0",
    "pygments": "2.18.0",
    "pymdown-extensions": "10.12",
    "pyyaml": "6.0.2",
    "ruff": "0.8.3",
    "starlette": "0.42.0",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.12.2",
    "uvicorn": "0.34.0",
    "websockets": "14.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.1.3",
    "pandas": "2.2.3",
    "polars": "1.17.1",
    "pyarrow": "18.1.0"
  }
}

### Code to reproduce

```python
### notebook.py
@app.cell
def print_col_df(df):
   print(df.columns.tolist())
```

```python
### test_notebook.py
@app.cell
def imports():
   from notebook import print_col_df
   import pandas as pd
   return (print_col_df, pd)

@app.cell
def test_print_col_df(print_col_df, pd):
   df = pd.DataFrame(data={'col1': [1, 2], 'col2': [3, 4]})
   output, defs = print_col_df.run(df=df)
   assert output == ['col1', 'col2']
```

akshayka:
@benbaja , this should now be fixed, marimo version 0.10.8. Thank you for reporting!

[...]

PR #3300, opened 2024-12-27
fix: typo in run subs

dmadisetti:
Solves #3299 

Just making sure it wasn't my recent change.

@akshayka OR @mscolnick


akshayka (approved):
Was just about to sit down to fix this, and noticed your PR. Thank you!

PR #3308, opened 2024-12-29
chore: update env section of bug issue template

akshayka:
nest under details; include backticks


mscolnick:
I wonder if we should do the same ticks for the reproduction step, but not hidden under details.

PR #3310, opened 2024-12-30
feat: add sqlglot to parse sql dataflow

### LINKED ISSUE #3103: More comprehensive parsing of the SQL dataflow graph
### Description

We fail to parse Tables and Common Table Expression (CTE) in certain nested expressions or joins in SQL. 

We would like to be able to cover more edge-case. They can be found in:
* `tests/_ast/test_sql_visitor.py` > "@pytest.mark.xfail"

### Suggested solution

We could continue parsing ourselves using the token/ast that duckdb returns, or use a more comprehensive library like [sqlglot](https://github.com/tobymao/sqlglot)

### Alternative

_No response_

### Additional context

_No response_

Light2Dark:
## 📝 Summary

Fixes #3103. Adds sqlglot as an optional dependency to handle complex sql refs parsing. Also supports DML statements (insert, update and delete).

## 🔍 Description of Changes

- set sqlglot version to lowest possible (likely >23.4, due to ibis lib)
- If we're keen on supporting dml's, then I could add more tests (I added some here alrdy)
- `def find_sql_def()` can also be modified to use sqlglot in the future, not included in this PR.
- small change: updated duckdb version to 1.1.0 to be synchronised across all installations (some were 1.0 and some were 1.1)

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_ast/sql_visitor.py
+++ b/marimo/_ast/sql_visitor.py
@@ -348,19 +348,17 @@ def find_sql_refs(
             for table in expression.find_all(exp.Table):
                 append_refs_from_table(table)
 
-        # this traversal is only available for select statements
-        root = build_scope(expression)
-        if root is None:
-            continue
-        if is_dml:
-            LOGGER.warning(
-                "Scopes should not exist for dml's, may need rework if this occurs"
-            )
+        # build_scope only works for select statements
+        if root := build_scope(expression):
+            if is_dml:
+                LOGGER.warning(
+                    "Scopes should not exist for dml's, may need rework if this occurs"
+                )
 
-        for scope in root.traverse():  # type: ignore
-            for _alias, (_node, source) in scope.selected_sources.items():
-                if isinstance(source, exp.Table):
-                    append_refs_from_table(source)
+            for scope in root.traverse():  # type: ignore
+                for _alias, (_node, source) in scope.selected_sources.items():
+                    if isinstance(source, exp.Table):
+                        append_refs_from_table(source)
 
     # remove duplicates while preserving order
     return list(dict.fromkeys(refs))
--- a/marimo/_ast/sql_visitor.py
+++ b/marimo/_ast/sql_visitor.py
@@ -341,9 +341,7 @@ def find_sql_refs(
 
     expression_list = parse(sql_statement)
     for expression in expression_list:
-        dml_expression = False
-        if expression.find(exp.Update, exp.Insert, exp.Delete):
-            dml_expression = True
+        if is_dml := bool(expression.find(exp.Update, exp.Insert, exp.Delete)):
             for table in expression.find_all(exp.Table):
                 append_refs_from_table(table)
 
@@ -351,7 +349,7 @@ def find_sql_refs(
         root = build_scope(expression)
         if root is None:
             continue
-        if dml_expression:
+        if is_dml:
             LOGGER.warning(
                 "Scopes should not exist for dml's, may need rework if this occurs"
             )
--- a/marimo/_ast/sql_visitor.py
+++ b/marimo/_ast/sql_visitor.py
@@ -322,29 +322,44 @@ def find_sql_refs(
     from sqlglot.optimizer.scope import build_scope
 
     refs: list[str] = []
-    asts = parse(sql_statement)
-    for sql_ast in asts:
-        root = build_scope(sql_ast)
-        if root is None:  # Skip ddl's and comments
+
+    def append_refs_from_table(table_expr: exp.Table) -> None:
+        if table_expr.catalog == "memory":
+            # Default in-memory catalog, only include table name
+            refs.append(table_expr.name)
+        else:
+            # We skip schema if there is a catalog
+            # Because it may be called "public" or "main" across all catalogs
+            # and they aren't referenced in the code
+            if table_expr.catalog:
+                refs.append(table_expr.catalog)
+            elif table_expr.db:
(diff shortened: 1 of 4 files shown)


tests/_ast/test_sql_visitor.py
```diff
@@ -593,7 +601,7 @@ def test_find_sql_refs_with_lateral_join() -> None:
         SELECT * FROM employees,
         LATERAL (SELECT * FROM departments WHERE departments.id = employees.dept_id) dept;
         """
-        assert find_sql_refs(sql) == ["employees", "departments"]
+        assert find_sql_refs(sql) == ["departments", "employees"]
```
Light2Dark, line 604: i need to investigate more on this ordering. Nested tables are referenced first, I'm not sure what impact that has.

pyproject.toml
```diff
@@ -75,16 +75,22 @@ marimo = "marimo._cli.cli:main"
 homepage = "https://github.com/marimo-team/marimo"
 
 [project.optional-dependencies]
-sql = ["duckdb >= 1.0.0", "polars[pyarrow] >= 1.9.0"]
+sql = [
+    "duckdb>=1.1.0",
```
akshayka, line 79: nit: If `duckdb 1.0` is still compatible with this change, we should prefer it over `1.1` so not to break users who haven't upgraded.

marimo/_ast/sql_visitor.py
```diff
@@ -315,109 +313,52 @@ def find_sql_refs(
...
+            # We skip schema if there is a catalog
+            # Because it may be called "public" or "main" across all catalogs
+            # and they aren't referenced in the code
+            if table.catalog:
+                refs.append(table.catalog)
+            elif table.db:
+                refs.append(table.db)  # schema
+
+            if table.name:
+                refs.append(table.name)
```
akshayka, line 340: For `my_catalog.my_schema.my_table`, where `my_catalog` is explicitly given in the code, I wonder if the ref should just be `my_catalog` instead of `my_catalog` and `my_table`? Because two separate catalogs could have overlapping table names. Analogous to how in Python, the ref for `my_instance.my_variable` is just `my_instance`. 

A genuine question. Curious what makes more sense to you.

akshayka (commented):
This is fantastic, thank you so much! I'm very excited by what a migration to sqlglot will unlock in the future, including support for other dialects.

Left one nit and one question.

marimo/_ast/sql_visitor.py
akshayka: I think right now, `my_catalog.my_schema.my_table` generates a def of just `my_table` -- so the logic of adding `my_table` as a ref makes sense in the context of this PR. Perhaps we can revisit this question later.

mscolnick: Yea this was a known limitation / edge case when we added this, but it was the lesser of two evils. If using sqlglot always gives us a valid catalog, we could only use fully qualified/verbose table names. 

But given this does the same logic as before, I think we can look into it later (agreed with @akshayka)

Light2Dark: can revisit this when we modify `find_sql_defs` too

marimo/_ast/sql_visitor.py
```diff
@@ -315,109 +313,52 @@ def find_sql_refs(
...
+
+    expression_list = parse(sql_statement)
```
akshayka, line 342: Since we only support `duckdb` right now, and since duckdb has some special functions/syntax that other dialects don't have, I think this should be:

```suggestion
    expression_list = parse(sql_statement, dialect="duckdb")
```

akshayka (commented):
Just one comment, and then looks good to me. Thanks for this change!

tests/_ast/test_sql_visitor.py
```diff
@@ -626,18 +626,6 @@ def test_find_sql_refs_nested_intersect() -> None:
...
-    def test_find_sql_refs_with_recursive_cte() -> None:
-        sql = """
-        WITH RECURSIVE cte AS (
-            SELECT 1 AS n FROM table1
-            UNION ALL
-            SELECT n + 1 FROM cte WHERE n < 10
-        )
-        SELECT * FROM cte;
-        """
-        assert find_sql_refs(sql) == ["table1"]
```
akshayka: If this is not yet supported, but you'd like to keep the test, you can decorate it with `pytest.mark.xfail`, as done here: https://github.com/marimo-team/marimo/blob/f8dc9f5c210a44375f5a431771efcfc9a6f17ea7/tests/_ast/test_sql_visitor.py#L569

Light2Dark: it does work, I didn't see the need for it since we have tested nested cte's. Also noted on the dialect change. I may add back the has_duckdb dependency too.

tests/_ast/test_sql_visitor.py
```diff
@@ -726,3 +726,8 @@ def test_find_sql_refs_select_in_delete() -> None:
         );
         """
         assert find_sql_refs(sql) == ["table1", "table2"]
+
+    @staticmethod
+    def test_find_sql_refs_invalid_sql() -> None:
+        sql = "SELECT * FROM"
+        assert find_sql_refs(sql) == []
```
akshayka, line 733: Nice test, thanks

akshayka (approved):
Fantastic work! This will have a large impact on marimo's SQL support; thanks so much for getting the ball rolling!

PR #3319, opened 2024-12-31
Allow real-time cell editing by multiple users

davidbrochart:
## 📝 Summary

This PR makes it possible for multiple users to edit the same notebook in real-time. This is using [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) and more precisely [Yjs](https://github.com/yjs/yjs) in the frontend and [pycrdt](https://github.com/jupyter-server/pycrdt) in the backend.
Unlike the "kiosk mode" which only shows a read-only view of the outputs, the collaborative mode allows editing cell inputs. Users can edit the same cell at the same time: CRDTs ensure that all users eventually see the same text, potentially resolving conflicts.

## 🔍 Description of Changes

A new WebSocket endpoint is created at `ws/{cell_id}` for each notebook cell. When connecting to it, a cell input represented as a [shared text](https://docs.yjs.dev/api/shared-types/y.text) in the frontend is synchronized with [its peer](https://jupyter-server.github.io/pycrdt/api_reference/#pycrdt.Text) in the backend. The backend shared cell receives changes from every connected frontend, and broadcast them to every connected frontend.

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/marimo/_server/api/endpoints/ws.py
+++ b/marimo/_server/api/endpoints/ws.py
@@ -3,9 +3,20 @@ from __future__ import annotations
 
 import asyncio
 import json
+from dataclasses import dataclass
 from enum import IntEnum
+from functools import partial
 from typing import Any, Callable, Optional
 
+from pycrdt import (
+    Doc,
+    Text,
+    TransactionEvent,
+    YMessageType,
+    create_sync_message,
+    create_update_message,
+    handle_sync_message,
+)
 from starlette.websockets import WebSocket, WebSocketDisconnect, WebSocketState
 
 from marimo import _loggers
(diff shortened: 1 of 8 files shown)


marimo/_server/api/endpoints/ws.py
```diff
@@ -95,6 +106,109 @@ async def websocket_endpoint(
     ).start()
 
 
+@dataclass
+class Cell:
```
dmadisetti, line 110: Bit of a Namespace conflict with other "Cell" classes. Thoughts on EditingCell or something?

marimo/_server/api/endpoints/ws.py
```diff
@@ -95,6 +106,109 @@ async def websocket_endpoint(
...
+        while True:
+            update = await update_queue.get()
+            message = create_update_message(update)
+            await websocket.send_bytes(message)
+    except Exception:
+        return
+
+
+async def clean_cell(cell_id: str) -> None:
+    await asyncio.sleep(60)
```
dmadisetti, line 140: Can you add a comment for why this is needed/ 60 is chosen?

marimo/_server/api/endpoints/ws.py
```diff
@@ -95,6 +106,109 @@ async def websocket_endpoint(
...
+async def send_updates(
+    websocket: WebSocket, update_queue: asyncio.Queue[bytes]
+) -> None:
+    try:
+        while True:
+            update = await update_queue.get()
+            message = create_update_message(update)
+            await websocket.send_bytes(message)
+    except Exception:
+        return
```
dmadisetti, line 136: Probably worth logging a failure

dmadisetti (commented):
Wow! That's amazing, super cool it's such a small code change too

I wonder how scalable the per cell connections are, but the bones are there to do a full notebook sync. I think this also opens up potential race conditions, since the frontend sends down cell content, but the content isn't synced to the cell impl here. But this is probably still worth being in experimental

Super excited for this!

davidbrochart:
> I wonder how scalable the per cell connections are, but the bones are there to do a full notebook sync. I think this also opens up potential race conditions, since the frontend sends down cell content, but the content isn't synced to the cell impl here. But this is probably still worth being in experimental

Right, Jupyter defines a [shared document for the whole notebook](https://github.com/jupyter-server/jupyter_ydoc/blob/db70566443ffc92e9bf62e2064e9b928ca2e96d7/jupyter_ydoc/ynotebook.py), which consists of an array of cells. This means that there is only one WebSocket connection per notebook, and that cell operations (delete, move) are themselves handled using CRDTs.
I didn't want to do that in this PR because I think it could interfere with the kiosk mode, which already handles cell operations like deletions.
Also, in Jupyter there is a single place in the backend where the notebook content is saved to disk. Here each client will trigger a save operation, since that still goes to the `ws` endpoint. I think that it's a bigger change for another PR.

dmadisetti:
The cell ordering in marimo is superficial, so I think contents can remain disjoint and rely on cell id lookup without needing to explicitly handle move (but maybe needs some logic for delete, just for cleanup)

Great first step, and barring typing and tests- I think my only other reservation is hiding it behind a flag until it's a bit more baked. That way we can get it up, and jam on it a little/ checks it works with the VSCode plugin etc

I think the most naive implementation would be to remove pycrdt as a required dep, and branch on its presence (failed ws endpoint causes default fe behavior)- but maybe someone else has a better idea

akshayka:
> I think the most naive implementation would be to remove pycrdt as a required dep, and branch on its presence (failed ws endpoint causes default fe behavior)- but maybe someone else has a better idea

We could use a feature flag under the experimental key: https://github.com/marimo-team/marimo/blob/main/marimo/_config/config.py#L244

It would be enabled by manually modifying the marimo TOML file to include

```
[experimental]
rtc = True
```

davidbrochart:
I think the experimental RTC flag is better than activating the feature if `pycrdt` is installed, in case `pycrdt` is installed and used for something else.

marimo/_server/api/endpoints/ws.py
```diff
@@ -95,6 +106,118 @@ async def websocket_endpoint(
...
+    # so that it can be synchronized and to prevent content duplication.
+    # This task will be cancelled when any client reconnets.
+    await asyncio.sleep(60)
+    async with ycell_lock:
+        ycell = ycells[cell_id]
+        if not ycell.clients:
+            del ycells[cell_id]
+
+
+@router.websocket("/ws/{cell_id}")
```
mscolnick, line 156: Do you expect there to be scalability concerns if there are potentially 40+ cells/WebSocket connections?

mscolnick:
this is awesome @davidbrochart! going to merge to the ongoing branch

marimo/_server/api/endpoints/ws.py
davidbrochart: Hard to say, if that happens the next step would be to have this top-level shared document for the whole notebook.

davidbrochart:

PR #3337, opened 2025-01-03
experimental: Real-time collabration

mscolnick:
Heavy lifting done by @davidbrochart in #3319

----------

### 📝 Summary
This PR makes it possible for multiple users to edit the same notebook in real-time. This is using [CRDTs](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) and more precisely [Yjs](https://github.com/yjs/yjs) in the frontend and [pycrdt](https://github.com/jupyter-server/pycrdt) in the backend.
Unlike the "kiosk mode" which only shows a read-only view of the outputs, the collaborative mode allows editing cell inputs. Users can edit the same cell at the same time: CRDTs ensure that all users eventually see the same text, potentially resolving conflicts.

### 🔍 Description of Changes
A new WebSocket endpoint is created at ws/{cell_id} for each notebook cell. When connecting to it, a cell input represented as a [shared text](https://docs.yjs.dev/api/shared-types/y.text) in the frontend is synchronized with [its peer](https://jupyter-server.github.io/pycrdt/api_reference/#pycrdt.Text) in the backend. The backend shared cell receives changes from every connected frontend, and broadcast them to every connected frontend.


akshayka (dismissed):

akshayka (dismissed):
LGTM.

I saw the edge case of needing a more robust key than the filename. Are there other edge cases to call out or document?

mscolnick:
@akshayka no other known limitations but needs some stress testing. And won't work in wasm. And I'm pretty sure multi-column won't work.

davidbrochart:
Thanks for all the refactoring and tests @mscolnick, this looks great.
For RTC to work in WASM, an option would be to use WebRTC through [y-webrtc](https://github.com/yjs/y-webrtc). There is experimental support for this in [jupyter-shared-drive](https://github.com/davidbrochart/jupyter-shared-drive). Also maybe worth looking at: https://github.com/yjs/titanic.

Issue #3110, opened 2024-12-10
Improved error messages in the UI with clear call-to-action when there are errors


mscolnick:
### Description

Not all our error messages have clear call to actions. These errors exist in `frontend/src/components/editor/output/MarimoErrorOutput.tsx`

### Suggested solution

1. We should improve the messaging, with a CTA
2. Can we improve deep links (linking to cells, or variables, or opening a panel)
3. Can we have auto-fixes or auto-installation or auto-import
4. Can some of the errors be less prominent (either not red or muted red) when they are dependent on another error

mscolnick:
also fixed by https://github.com/marimo-team/marimo/pull/4384

PR #3343, opened 2025-01-04
feat: add auto-fix for multiple definctions and popular imports

mscolnick:
Part of #3110

This adds two autofixes for different marimo errors. This opens the door for more autofixes as well. Right now this happens on all the frontend because 1) the code transformations are not that difficult and 2) still requires frontend logic like focusing the editor or adding a new cell. We could move some of this to the backend but would prefer to put the auto-fixes into a a proper linting framework + LSP.

The autofixes:
* multiple definitions -> Make this cell's variables local by wrapping the cell in a function.
* missing name -> Add a new cell for the missing import

image

### DIFF
--- a/frontend/src/core/errors/errors.ts
+++ b/frontend/src/core/errors/errors.ts
@@ -44,7 +44,7 @@ export function getAutoFixes(error: MarimoError): AutoFix[] {
       return [];
     }
 
-    const cellCode = IMPORT_MAPPING[name];
+    const cellCode = getImportCode(name);
 
     return [
       {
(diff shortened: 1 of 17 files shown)


frontend/src/core/errors/errors.ts
```diff
@@ -0,0 +1,89 @@
...
+  re: "import re",
+  scipy: "import scipy",
+  sk: "import sklearn as sk",
+  sns: "import seaborn as sns",
+  stats: "import scipy.stats as stats",
+  sys: "import sys",
+  tf: "import tensorflow as tf",
+  torch: "import torch",
+  xr: "import xarray as xr",
+};
```
mscolnick, line 89: could probably consolidate this

frontend/src/core/errors/errors.ts
```diff
@@ -0,0 +1,89 @@
...
+  }) => Promise<void>;
+}
+
+export function getAutoFixes(error: MarimoError): AutoFix[] {
+  if (error.type === "multiple-defs") {
+    return [
+      {
+        title: "Wrap in a function",
+        description:
+          "Wrap the cell contents in a function so they are marked as definitions of the cell.",
```
akshayka, line 24:
```suggestion
          "Make this cell's variables local by wrapping the cell in a function.",
```

akshayka (dismissed):
This is awesome! I played with it, and it works well. Just one comment on formatting.

Issue #3355, opened 2025-01-07
Polars LazyFrame `show_graph` has poor graphic quality in Marimo


kjgoodrick:
### Description

I am submitting a PR for this suggestion.

When using polars LazyFrames it is often desirable to display the query plan prior to collecting the query. There are multiple ways to do this in Marimo. 

1. Leave the LazyFrame object as the last line in a cell.
   - This works well and displays a nice SVG image in the cell (as long as graphviz is installed and on the path). ![Image](https://github.com/user-attachments/assets/0756de45-05ba-4565-8f65-214855821d01)
3. Manually call the `show_graph` function    
   - This is necessary if one wants to show the optimized version of the query plan
   - Unfortunately, this produces a very poor output if matplotlib is installed or an error if it is not![Image](https://github.com/user-attachments/assets/2e39572c-ab10-48f6-a372-5cb237b628fd)
3. Use `explain` to output a text-based query plan
   - This works well and Marimo even formats the text (whereas Jupyter notebooks by default show the raw string) 
   - However, it can be harder to follow than the graph for more complicated queries. ![Image](https://github.com/user-attachments/assets/dd21baf3-6fa7-4228-af85-9ef3c6b89712)

It would be nice if Marimo could display high quality graphics for the query plan even when the user wants to view the optimized plan and/or doesn't have matplotlib / graphviz installed. It would also be nice if the graph followed the theme of the notebook (dark / light) and could nicely display plans that have many terms (e.g. a query that transforms many columns in one step).

### Suggested solution

My suggestion is to register a polars extension that adds a marimo `mo` namespace to LazyFrames and allows users to display high quality query plan graphs in all situations with only a slight change to their code. This added polars extension code will be maintained within the marimo repository so that it will not require adding any code to the polars repository. 

Because marimo already has support for displaying mermaid graphs and polars can return the raw text defining the graph (in dot notation) it makes sense to parse this and convert to mermaid.

Once added this approach would allow for results that:

1. Are natively high quality for all means of showing the graph
2. Wrap the text in long columns
3. Do not require graphviz or matplotlib to be installed
4. Adopt to the user theme by default

| Case | Current | Proposed | Notes |
|---|---|---|---|
|Show Graph|![Image](https://github.com/user-attachments/assets/e1f18cff-33b5-45d7-8d54-98b496010bb2)| ![Image](https://github.com/user-attachments/assets/f9eff464-f5ec-4e3c-85d6-27e20d301c42) |
[...]

PR #3356, opened 2025-01-07
Polars extension

kjgoodrick:
## 📝 Summary

Fixes issue #3355 by adding LazyFrame extension that allows output of mermaid graphs for polars lazyframe query plans. 

## 🔍 Description of Changes

In this pull request I have added a polars extension that adds a `mo` namespace to LazyFrames, this namespace includes one "public" method `show_graph` that accepts the same arguments as polars LazyFrame `show_graph` and returns an `Html` object containing the mermaid chart. 

Issue #3355 shows side by side outputs.

Before merging it would be necessary to add documentation for this new polars namespace, but I wanted to request your feedback before I took that step.

## 📋 Checklist

## 📜 Reviewers

@akshayka

### DIFF
--- a/marimo/__init__.py
+++ b/marimo/__init__.py
@@ -116,6 +116,9 @@ from marimo._plugins.stateless.style import style
 from marimo._plugins.stateless.tabs import tabs
 from marimo._plugins.stateless.tree import tree
 from marimo._plugins.stateless.video import video
+from marimo._polars import (
+    lazyframe,  # noqa: F401  # Import is required for lazyframe registration
+)
 from marimo._runtime import output
 from marimo._runtime.capture import (
     capture_stderr,
--- a/marimo/_polars/lazyframe.py
+++ b/marimo/_polars/lazyframe.py
@@ -1,0 +1,63 @@
+import re
+from typing import TYPE_CHECKING, Any
+
+from marimo import mermaid
+from marimo._dependencies.dependencies import DependencyManager
+from marimo._output.hypertext import Html
+
+if TYPE_CHECKING:
+    import polars as pl
+
+if DependencyManager.polars.has():
+    import polars as pl
+
+    @pl.api.register_lazyframe_namespace("mo")
+    class Marimo:
+        def __init__(self, ldf: pl.LazyFrame):
+            self._ldf: pl.LazyFrame = ldf
+
+        def show_graph(self, **kwargs: Any) -> Html | str:
+            # We are specifying raw_output already, so we need to remove if it is passed
+            raw_output = kwargs.pop("raw_output", False)
+
+            dot = self._ldf.show_graph(raw_output=True, **kwargs)
+
+            if raw_output:
+                return dot
+
+            return self._dot_to_mermaid_html(dot)
+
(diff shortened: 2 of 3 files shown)


marimo/__init__.py
```diff
@@ -116,6 +116,9 @@
 from marimo._plugins.stateless.tabs import tabs
 from marimo._plugins.stateless.tree import tree
 from marimo._plugins.stateless.video import video
+from marimo._polars import (
+    lazyframe,  # noqa: F401  # Import is required for lazyframe registration
+)
```
akshayka, line 121: If registration happens on import of the polars module, then this can be removed.

https://github.com/marimo-team/marimo/blob/main/marimo/_output/formatters/df_formatters.py

As far as I can tell there's no need to add `lazyframe` to the marimo public API?

akshayka (commented):
Instead of adding a new namespace and requiring users to learn about it, I believe we can register custom formatters for Polars, in this file:

https://github.com/marimo-team/marimo/blob/main/marimo/_output/formatters/df_formatters.py

For example, we could monkey-patch `show_graph` to do the right thing in marimo.

I agree that the ideal solution is to get Polars to support marimo natively. They can detect if they are running in a marimo notebook by checking `mo.running_in_notebook()`, and then integrate with marimo in one of various ways (https://docs.marimo.io/guides/integrating_with_marimo/displaying_objects/).

@MarcoGorelli , if you don't mind, I would appreciate hearing your thoughts on whether implementing this upstream in Polars would be possible, or if we should plan to monkey-patch Polars in marimo instead.

MarcoGorelli:
> @MarcoGorelli , if you don't mind, I would appreciate hearing your thoughts on whether implementing this upstream in Polars would be possible, or if we should plan to monkey-patch Polars in marimo instead.

if the solution isn't too complex on the Polars side, then I think this would be welcome!

mscolnick:
Looks like there was some appetite for mermaid formatting in polars: https://github.com/pola-rs/polars/issues/12075

That part seems non-controversial, but could go as far as making the default `'mermaid'` instead of `'dot'` in marimo

kjgoodrick:
Thank you all for the discussion on this. It sounds like there is already interest on the polars side for introducing support for mermaid, which is great to here. If polars does support mermaid and is ok with us inserting a simple snippet for detecting marimo then I think we could forgo the polars extension which would remove the need to add any imports to the marimo init file. 

I think then that the best way forward would be to:
1. Add Mermaid support to polars `show_graph`
2. Add a dataframe formatter to Marimo that will change the behavior of leaving a LazyFrame on the last line
    - Use polars mermaid generation developed in 1 
3. Add support to polars `show_graph` for detecting marimo and calling the display functionality developed in 2. 

If that sounds good to you all, I can close this PR and then make a new one to handle step 2 above.

akshayka:
That sounds good to me, thank you!

Would you like to work with Polars to lead 1 and 3 as well? If you opened an issue or PR with them, please feel free to tag me.

kjgoodrick:
Yes, I will open a PR in Polars for 1 to get started and tag you in it.

mscolnick:
Closing in favor of https://github.com/marimo-team/marimo/pull/5247

PR #3381, opened 2025-01-09
chore: fix smoke test for vegafusion

akshayka:
- some APIs don't exist
- versions stale, missing packages sandbox


mscolnick (approved):
yay package reproducibility

PR #3404, opened 2025-01-12
fix: better chatbot input & bubble

### LINKED ISSUE #3401: support multiline for mo.ui.chat input & bubble
### Describe the bug

Image

1. The input textbox does not support multiple lines.
  Suggestion: make it an autosizing textarea. Some options:
  a) Add a new package - https://github.com/Andarist/react-textarea-autosize
  b) some css, js hack - https://css-tricks.com/auto-growing-inputs-textareas/#aa-other-ideas
  c) same behaviour as chat panel - but I think this isn't an option, that uses codemirror which is overkill.

     I tried and b) should work, avoids useLayoutEffect hooks from a)

2. Shift+enter / cmd+enter to add new line, Enter to submit.
   - This is a bit tricky since we need to override the code editor shortcut (shift+enter will go to the next cell)
   - One suggestion is we keep the default behaviour of textarea (Enter creates a new line), and we set `Ctrl+Enter` to submit the form

3. The chat bubble will auto format everything to 1 line.
    I believe there's a simple solution,`chat-ui.tsx`
    ```.tsx
      <p className={cn(message.role === "user" && "whitespace-pre-wrap")}>
        {renderMessage(message)}
      </p>
    ```

### Environment

<details>

```
{
  "marimo": "0.10.12",
  "OS": "Darwin",
  "OS Version": "23.6.0",
  "Processor": "arm",
  "Python Version": "3.12.8",
  "Binaries": {
    "Browser": "131.0.6778.265",
    "Node": "v23.2.0"
  },
  "Dependencies": {
    "click": "8.1.3",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.20.1",
    "packaging": "24.2",
    "psutil": "6.1.1",
    "pygments": "2.18.0",
    "pymdown-extensions": "10.13",
    "pyyaml": "6.0.2",
    "ruff": "0.6.9",
    "starlette": "0.45.0",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "14.1"
  },
  "Optional Dependencies": {
    "anywidget": "0.9.13",
    "duckdb": "1.1.3",
    "ibis-framework": "9.5.0",
    "pandas": "2.2.3",
    "polars": "1.18.0",
    "pyarrow": "17.0.0"
  }
}
```

</details>

### Code to reproduce

```.py
mo.ui.chat(
  mo.ai.llm.openai(
    "gpt-4o-mini",
    system_message="You are a helpful chemist. Output your answer in markdown or latex.",
  ),
  show_configuration_controls=True,
)
```

Light2Dark:
## 📝 Summary

Screenshot 2025-01-13 at 1 24 50 AM

Fixes #3401 . Use codemirror as chatbot input,  `/` triggers autocomplete.

## 🔍 Description of Changes

- multiline input
- multiline chat bubble
- removed an unnecessary extension (codemirror-mentions) and implemented custom one
note: dataset atom is not finding any tables, hence `@` is not working.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/editor/ai/add-cell-with-ai.tsx
+++ b/frontend/src/components/editor/ai/add-cell-with-ai.tsx
@@ -204,6 +204,7 @@ interface PromptInputProps {
   onSubmit: (e: KeyboardEvent | undefined, value: string) => void;
   additionalCompletions?: AdditionalCompletions;
   theme: ResolvedTheme;
+  maxHeight?: string;
 }
 
 /**
@@ -222,6 +223,7 @@ export const PromptInput = ({
   onClose,
   additionalCompletions,
   theme,
+  maxHeight,
 }: PromptInputProps) => {
   const handleSubmit = onSubmit;
   const handleEscape = onClose;
@@ -372,6 +374,7 @@ export const PromptInput = ({
       className={cn("flex-1 font-sans overflow-auto my-1", className)}
       autoFocus={true}
       width="100%"
+      maxHeight={maxHeight}
       value={value}
       basicSetup={false}
       extensions={extensions}
--- a/frontend/src/components/editor/ai/add-cell-with-ai.tsx
+++ b/frontend/src/components/editor/ai/add-cell-with-ai.tsx
@@ -190,7 +190,7 @@ export const AddCellWithAI: React.FC<{
 };
 
 export interface AdditionalCompletions {
-  triggerSymbol: string; // Symbol that will trigger autocompletion when text begins with it
+  triggerCompletionRegex: RegExp;
   completions: Completion[];
 }
 
@@ -285,17 +285,16 @@ export const PromptInput = ({
       }),
     );
 
-    // Trigger autocompletion for text that begins with @ or
-    // @ + additional symbols specified
-    const matchBeforeRegex = additionalCompletions
-      ? new RegExp(`[@${additionalCompletions.triggerSymbol}](\\w+)?`)
-      : /@(\w+)?/;
+    const matchBeforeRegexes = [/@(\w+)?/]; // Trigger autocompletion for text that begins with @
+    if (additionalCompletions) {
+      matchBeforeRegexes.push(additionalCompletions.triggerCompletionRegex);
+    }
     const allCompletions = additionalCompletions
       ? [...completions, ...additionalCompletions.completions]
       : completions;
 
     return [
-      mentions(matchBeforeRegex, allCompletions),
+      mentions(matchBeforeRegexes, allCompletions),
       EditorView.lineWrapping,
       minimalSetup(),
       Prec.highest(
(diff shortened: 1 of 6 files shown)


frontend/src/plugins/impl/chat/chat-ui.tsx
```diff
@@ -309,22 +334,29 @@ export const Chatbot: React.FC<Props> = (props) => {
...
               });
             }}
           />
         )}
-        <input
-          name="prompt"
-          ref={inputRef}
+        <PromptInput
+          className="rounded-sm mr-2"
+          placeholder="Type your message here, / for prompts"
```
mscolnick, line 345: can we change the placeholder of prompts exist or not

frontend/src/components/editor/ai/add-cell-with-ai.tsx
```diff
@@ -277,8 +283,17 @@ export const PromptInput = ({
       }),
     );
 
+    // Trigger autocompletion for text that begins with @ or
+    // @ + additional symbols specified
+    const matchBeforeRegex = additionalCompletions
```
mscolnick, line 288: Can we do 2 separate mentions plugins? One for "@" data and one for "/" prompts; instead of appending the trigger, replace it.

mscolnick:
I think the mention changes are good / worth keeping. I plan to add support for MCPs so we could eventually replace this with https://github.com/marimo-team/codemirror-mcp

frontend/src/components/editor/ai/add-cell-with-ai.tsx
Light2Dark: do you mean if there are `additionalCompletions`, we only use the "/" regex?

it's also possible to accept multiple regexes, if we want to make it extensible

```.tsx
export function mentions(
  matchBeforeRegexes: RegExp[],
  data: Completion[] = [],
): Extension {
  return autocompletion({
    override: [
      (context: CompletionContext) => {
        const word = matchBeforeRegexes
          .map((regex) => context.matchBefore(regex))
          .find(Boolean);
```

mscolnick: from the code, it looks like `@` would trigger autocomplete for prompts, and I was thinking that `@` only trigger autocomplete for datasets, while `/` triggers for prompts. you could maybe do this in a single `mentions` plugin, but doing two (one for datasets and one for prompts) might be easier

Light2Dark: hmm, you can't do 2 autocompletion extensions, since they will overwrite and cause a conflict from codemirror. Seems a solution will be tricky if we want to have 2 mentions plugin.

```.tsx
return [
      mentions(/@(\w+)?/, completions),
      mentions(/\/(\w+)?/, additionalCompletions?.completions),
      ...
]
```
error: Config merge conflict for field override

but anyways, `@` will indeed trigger autocomplete, but there are no matching prompts beginning with `@`, so nothing is found.

mscolnick: gotcha. i see, that makes sense. 

then what you have is good to me.

mscolnick (approved):
awesome stuff!

PR #3456, opened 2025-01-16
fix: empty notebooks should have empty venvs in sandboxes

akshayka:
If there are no dependencies (specifically, if uv is not running a script with inline metadata), uv may use a cached venv, even though we are passing --isolated; `--refresh` ensures that the venv is actually ephemeral

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -288,16 +288,23 @@ def run_in_sandbox(
         python_version = None
 
     # Construct base UV command
-    uv_cmd = [
-        "uv",
-        "run",
-        "--isolated",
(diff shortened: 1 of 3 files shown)


mscolnick (approved):

marimo/_cli/sandbox.py
```diff
@@ -264,6 +258,12 @@ def run_in_sandbox(
         get_dependencies_from_filename(name) if name is not None else []
     )
 
+    # If there are no dependences, which can happen for marimo new or
```
akshayka, line 261:
```suggestion
    # If there are no dependencies, which can happen for marimo new or
```

PR #3462, opened 2025-01-16
load snippets location from config

bennyweise:
## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -12,7 +12,7 @@ if sys.version_info < (3, 11):
 else:
     from typing import NotRequired
 
-from typing import Any, Dict, Literal, Optional, TypedDict, Union, cast, List
+from typing import Any, Dict, List, Literal, Optional, TypedDict, Union, cast
 
 from marimo._output.rich_help import mddoc
 from marimo._utils.deep_merge import deep_merge
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -227,6 +227,18 @@ class GoogleAiConfig(TypedDict, total=False):
     api_key: str
 
 
+@dataclass
+class SnippetsConfig(TypedDict):
+    """Configuration for snippets.
+
+    **Keys.**
+
+    - `custom_path`: the path to the custom snippets directory
+    """
+
+    custom_path: NotRequired[str]
+
+
 @mddoc
 @dataclass
 class MarimoConfig(TypedDict):
@@ -242,7 +254,7 @@ class MarimoConfig(TypedDict):
     package_management: PackageManagementConfig
     ai: NotRequired[AiConfig]
     experimental: NotRequired[Dict[str, Any]]
-
+    snippets: NotRequired[SnippetsConfig]
 
 @mddoc
 @dataclass
@@ -259,6 +271,7 @@ class PartialMarimoConfig(TypedDict, total=False):
     package_management: PackageManagementConfig
     ai: NotRequired[AiConfig]
     experimental: NotRequired[Dict[str, Any]]
+    snippets: SnippetsConfig
 
 
 DEFAULT_CONFIG: MarimoConfig = {
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -256,6 +256,7 @@ class MarimoConfig(TypedDict):
     experimental: NotRequired[Dict[str, Any]]
     snippets: NotRequired[SnippetsConfig]
 
+
 @mddoc
 @dataclass
 class PartialMarimoConfig(TypedDict, total=False):
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -12,7 +12,7 @@ if sys.version_info < (3, 11):
 else:
     from typing import NotRequired
 
-from typing import Any, Dict, Literal, Optional, TypedDict, Union, cast
+from typing import Any, Dict, Literal, Optional, TypedDict, Union, cast, List
 
 from marimo._output.rich_help import mddoc
 from marimo._utils.deep_merge import deep_merge
@@ -236,7 +236,8 @@ class SnippetsConfig(TypedDict):
     - `custom_path`: the path to the custom snippets directory
     """
 
-    custom_path: NotRequired[str]
+    custom_paths: NotRequired[List[str]]
+    include_default_snippets: NotRequired[bool]
 
 
 @mddoc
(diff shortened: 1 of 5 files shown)


marimo/_config/config.py
```diff
@@ -227,6 +227,18 @@ class GoogleAiConfig(TypedDict, total=False):
...
+@dataclass
+class SnippetsConfig(TypedDict):
+    """Configuration for snippets.
+
+    **Keys.**
+
+    - `custom_path`: the path to the custom snippets directory
+    """
+
+    custom_path: NotRequired[str]
```
mscolnick, line 239: Thoughts about this being a list of strings? I could see someone asking for multiple paths

marimo/_snippets/snippets.py
```diff
@@ -86,10 +90,26 @@ def is_markdown(code: str) -> bool:
...
+                yield os.path.join(default_root, file)
+
+    # Yield files from custom path if configured
+    if custom_path:
+        LOGGER.debug("Using custom snippets path: %s", custom_path)
+        custom_root = os.path.realpath(custom_path)
+        for _root, _dirs, files in os.walk(custom_root):
+            for file in files:
+                if file.endswith(".py"):
+                    yield os.path.join(custom_root, file)
```
mscolnick, line 115: maybe this could be DRYed up a bit:

```python
def _yield_py_files(root_path: str) -> Iterator[str]:
    real_path = os.path.realpath(root_path)
    for _root, _dirs, files in os.walk(real_path):
        for file in files:
            if file.endswith(".py"):
                yield os.path.join(real_path, file)

# Get all paths
default_path = str(import_files("marimo").joinpath("_snippets").joinpath("data"))
config = get_default_config_manager(current_path=None).get_config()
custom_paths: list[str] = config.get("snippets", {}).get("custom_paths", [])
if custom_paths:
    LOGGER.debug("Using custom snippets path: %s", custom_paths)
paths = [default_path] + custom_paths

for path in paths:
    yield from _yield_py_files(path)
```

marimo/_config/config.py
bennyweise: I was planning to ask you the same question - I think a list makes more sense here. Also, should the default path optional as well in the config? Something like

```
[tools.marimo.snippets]
use_default_path: true
custom_paths:
  - foo/bar
  - foo/baz
```

mscolnick: yea thats not a bad idea. maybe `include_default_snippets` with default `true`?

bennyweise: Updated no as a list of strings, with `include_default_snippets` as an option with default `true`

marimo/_snippets/snippets.py
bennyweise: Basically conforms to this style now - thanks for the pointers

bennyweise:
Have updated based on the comments, and tested locally with the standard test cases of real directory  / non-existent directory / recursively symlinked directory, and all look ok (but I may be missing some pathological cases). 

In terms of formal tests, I was wondering if there is a recommended way to test with different configs - I could patch the `get_default_config` method to inject various options, but that doesn't feel like a great solution. Alternatively I can extract the snippet path operations from the configuration, but I don't know that those tests would really achieve much.

bennyweise:

mscolnick:
hey @bennyweise, this looks great! 

I do think a test would be nice. You don't need to test change the config, etc. but maybe we can refactor the `get_snippets` to be more testable

```
def read_snippet_files(include_default_snippets: bool, custom_paths: list[str]):
    paths = []
    if include_default_snippets:
        paths.append(import_files("marimo") / "_snippets" / "data")
    if custom_paths:
        paths.extend(custom_paths)
    for root_path in paths:
        if not root_path.is_dir():
            # Note: currently no handling of permissions errors, but theoretically
            # this shouldn't be required for `is_dir` or `rglob`
            # Other possible errors:
            # - RecursionError: not possible, since by default symlinks are not followed
            # - FileNotFoundError: not possible, `is_dir` checks if the path exists,
            # but also resolve() is not called with strict=True
            LOGGER.warning(
                "Snippets path %s not a directory - ignoring", root_path
            )
            continue
        for file in root_path.resolve().rglob("*.py"):
            yield str(file)
```

and we can write a few different tests for 
```
assert list(read_snippet_files(True, [])) == [..]
assert list(read_snippet_files(False, [])) == [..]
assert list(read_snippet_files(False, [good_path])) == [..]
assert list(read_snippet_files(False, [bad_path])) == [..]
assert list(read_snippet_files(False, [good_path, bad_path])) == [..]
... etc.
```

bennyweise:
> hey @bennyweise, this looks great!
> 
> I do think a test would be nice. You don't need to test change the config, etc. but maybe we can refactor the `get_snippets` to be more testable
> 
> ```
> def read_snippet_files(include_default_snippets: bool, custom_paths: list[str]):
>     paths = []
>     if include_default_snippets:
>         paths.append(import_files("marimo") / "_snippets" / "data")
>     if custom_paths:
>         paths.extend(custom_paths)
>     for root_path in paths:
>         if not root_path.is_dir():
>             # Note: currently no handling of permissions errors, but theoretically
>             # this shouldn't be required for `is_dir` or `rglob`
>             # Other possible errors:
>             # - RecursionError: not possible, since by default symlinks are not followed
>             # - FileNotFoundError: not possible, `is_dir` checks if the path exists,
>             # but also resolve() is not called with strict=True
>             LOGGER.warning(
>                 "Snippets path %s not a directory - ignoring", root_path
>             )
>             continue
>         for file in root_path.resolve().rglob("*.py"):
>             yield str(file)
> ```
> 
> and we can write a few different tests for
> 
> ```
> assert list(read_snippet_files(True, [])) == [..]
> assert list(read_snippet_files(False, [])) == [..]
> assert list(read_snippet_files(False, [good_path])) == [..]
> assert list(read_snippet_files(False, [bad_path])) == [..]
> assert list(read_snippet_files(False, [good_path, bad_path])) == [..]
> ... etc.
> ```

Great, I've added some tests for the standard cases. There are some edge cases that may or may not be useful to resolve - e.g. you will end up with duplicate snippets if the default directory is listed in the custom paths. Is this something I should address with this change?

mscolnick:
> you will end up with duplicate snippets if the default directory is listed in the custom paths.

@bennyweise, I think this is ok to ignore for now. 

code/tests look great to me!

PR #3477, opened 2025-01-17
feat(ui): Add keyboard shortcuts for navigating between sidebar panels

metaboulie:
## 📝 Summary

Add keyboard shortcuts to navigate between panels in the sidebar, making it easier for users to switch between different views without using the mouse. The shortcuts are Ctrl-Shift-{ and Ctrl-Shift-}.

https://github.com/user-attachments/assets/d7b59f55-c367-424e-9af9-5beac036a4d6

## 🔍 Description of Changes

- Added two new hotkey definitions in `hotkeys.ts`:
  - `panel.nextPanel`: Navigate to next panel (Ctrl/Cmd-Shift-})
  - `panel.previousPanel`: Navigate to previous panel (Ctrl/Cmd-Shift-{)
- Added state management actions in `chrome/state.ts` to handle panel navigation:
  - `nextPanel`: Cycles to the next visible panel
  - `previousPanel`: Cycles to the previous visible panel
- Connected hotkeys to panel navigation in `app-chrome.tsx` using `useHotkey` hook
- Navigation only works when the sidebar is open
- Navigation cycles through visible panels in the order they appear in the sidebar

## 📋 Checklist

- [x] The change is self-contained and focused on improving keyboard navigation.
- [x] The code follows existing patterns for hotkey implementation and state management.
- [x] I have run the code and verified that the panel navigation works as expected.

## 📜 Reviewers

@akshayka OR @mscolnick


mscolnick:
`Ctrl-Shift-{ and Ctrl-Shift-}` is used to move tabs in the browser, so this might be a bad default. I would be open to
1. making them `NOT_SET` so people can set them if they'd like. `vscode` doesn't seem to have a way to cycle through.
2. making just move next `F3` (or a similar F-key), and prev is `NOT_SET`. so they can cycle through them, instead of navigate back and forth
3. doing `F3` and `Shift-F3`

vscode uses F3 for a lot of next/prev 
Screenshot 2025-01-17 at 10 37 11 AM

mscolnick (approved):

Issue #3471, opened 2025-01-16
persisted memoization (ie saving @mo.cache values)


gabrielgrant:
### Description

I'd like to have the control offered by the standard memoization interface of the `mo.cache` decorator, but have the values persist across kernel restarts (as with `marimo.persistent_cache`)

There are a few third-party options that do something similar (at least for pure functions):

- [`joblib.Memory`](https://joblib.readthedocs.io/en/latest/generated/joblib.Memory.html)
- [diskcache's @cache.memoize](https://grantjenks.com/docs/diskcache/tutorial.html#fanoutcache)
- klepto's [inf_cache](https://klepto.readthedocs.io/en/latest/#klepto.inf_cache) with `file_archive`

How does marimo's built-in caching compare to these options?

AFAICT the main difference would be that `mo.cache` takes closed-over dependent variables and/or source code changes into account in the same way `mo.persistent_cache` does, right? Are there any differences between the caching rules of the two marimo interfaces? Or just whether values are persisted? What is the thinking in having the style of cache interface (function memoization vs block-based-with-context-manager) also dictate whether the cache is persisted to disk or not?

### Suggested solution

Possibly just adding a "persisted" flag to `mo.cache`?

Not sure all the details of how caching is implemented internally, but might make sense to re-use one of these external libs in order to not have to fully reinvent the wheel?

### Alternative

Recommend using one of the existing third-party libs

gabrielgrant:
One (potential? hopefully?) difference is that these other solutions seem to have issues with consistently/correctly caching Pandas DataFrames:

https://github.com/joblib/joblib/issues/1611
https://github.com/grantjenks/python-diskcache/issues/314

This is obviously a pretty common use-case in a notebook environment. Is this handled correctly in marimo's cache implementation?

dmadisetti:
Just looked at this code again, and the answer is better than joblib, but still not ideal.

marimo has 2 modes of hashing, ContentAddressed (which is the content of the variable) and ExecutionContext (Which is the code run to produce the variable, used for cases where we cannot get a clear hash).

IF a dataframe is all numerical data then the raw, contiguous memory behind the df is used for hashing. If it is not (df contains objects), then the fallback is to ExecutionContext.

More explicit handling could be done, and I don't think column names contribute to the hash in this case.

PR #3550, opened 2025-01-23
feature: allow persistent_cache to be used as a decorator

dmadisetti:
## 📝 Summary

fixes #2653 #3471

## 🔍 Description of Changes

Enables `persistent_cache` to be used as a decorator for functions, and `cache` to be used as a context block. e.g.

```python
@mo.persistent_cache
def expensive_function_written_to_disk():
    ...

# or

with mo.cache("expensive_block_in_memory") as c:
    ...
```

`cache` is also used as the general entry point for custom "Loaders" 

---

The breadth of the API makes the implementation a bit hairy, but I think that if it's a smooth experience for the user then it's worth it.

@akshayka

### DIFF
--- a/marimo/_save/save.py
+++ b/marimo/_save/save.py
@@ -576,6 +576,7 @@ def cache(  # type: ignore[misc]
     ```
 
     **Args**:
+
     - `name`: the name of the cache, used to set saving path- to manually
       invalidate the cache, change the name.
     - `pin_modules`: if True, the cache will be invalidated if module versions
--- a/marimo/_save/save.py
+++ b/marimo/_save/save.py
@@ -2,18 +2,28 @@
 from __future__ import annotations
 
 import ast
+import functools
 import inspect
 import io
 import os
 import sys
 import traceback
 from collections import abc
-from functools import singledispatch
+from pathlib import Path
 
 # NB: maxsize follows functools.cache, but renamed max_size outside of drop-in
 # api.
 from sys import maxsize as MAXINT
-from typing import TYPE_CHECKING, Any, Callable, Optional, Type, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Optional,
+    Type,
+    Union,
+    cast,
+    overload,
+)
 
 from marimo._messaging.tracebacks import write_traceback
 from marimo._runtime.context import get_context
@@ -214,6 +224,7 @@ class _cache_context(object):
         loader: Loader,
         *,
         pin_modules: bool = False,
+        hash_type: str = DEFAULT_HASH,
     ) -> None:
         # For an implementation sibling regarding the block skipping, see
         # `withhacks` in pypi.
@@ -227,6 +238,7 @@ class _cache_context(object):
         self._body_start: int = MAXINT
         # TODO: Consider having a user level setting.
         self.pin_modules = pin_modules
+        self.hash_type = hash_type
         self._loader = loader
 
     def __enter__(self) -> Self:
@@ -293,6 +305,7 @@ class _cache_context(object):
                     loader=self._loader,
                     context=pre_module,
                     pin_modules=self.pin_modules,
+                    hash_type=self.hash_type,
                 )
 
                 self.cache_type = self._cache.cache_type
@@ -366,7 +379,20 @@ class _cache_context(object):
         return False
 
 
-@singledispatch
+# A note on overloading:
+# Single dispatch cues off only the first argument, and expects a similar
+# signature for every overload: https://peps.python.org/pep-0443/
+# However.
+# The context and call APIs are slightly different, so `*` expansions are used
+# to propagate that information down to the actually implementation.
+# As such, we also leverage the `@overload` decorator to provide the correct
+# signature and documentation for the singledispatch entry points.
+
+# mypy also seems to do pretty poorly with this overloading, so there is liberal
+# removal type rules.
+
+
+@functools.singledispatch
 def _cache_invocation(
     arg: Any,
     loader: Union[LoaderPartial, Loader, LoaderType],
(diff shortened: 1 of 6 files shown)


marimo/_save/save.py
```diff
@@ -511,3 +364,289 @@ def __exit__(
...
+    that would otherwise be expensive to compute already materialized in
+    memory.
+
+    **Usage.**
+
+    ```python
+    with persistent_cache(name="my_cache"):
+        variable = expensive_function()  # This will be cached to disk.
+        print("hello, cache")  # this will be skipped on cache hits
+    ```
```
akshayka, line 600: Can you add a decorator example as well?

marimo/_save/save.py
```diff
@@ -511,3 +364,289 @@ def __exit__(
...
+        assert isinstance(loader, LoaderPartial), (
+            UNEXPECTED_FAILURE_BOILERPLATE
+        )
+        loader = loader.create_or_reconfigure(name)()
+    return _cache_context(name, loader, *args, **kwargs)
+
+
+def cache(
+    name: Union[str, Optional[Callable[..., Any]]] = None,
+    *args: Any,
```
akshayka, line 458: Why change the signature to use varargs/varkwargs?

marimo/_save/save.py
```diff
@@ -511,3 +364,289 @@ def __exit__(
...
+        _frame_offset=2,
+        **kwargs,
+    )
+
+
+def persistent_cache(
+    name: Union[str, Optional[Callable[..., Any]]] = None,
+    save_path: str | None = None,
+    *args: Any,
+    **kwargs: Any,
```
akshayka, line 578: Same question, why varargs/kwargs?

marimo/_save/save.py
```diff
@@ -511,3 +364,289 @@ def __exit__(
...
+@singledispatch
+def _cache_invocation(
+    arg: Any,
+    loader: Union[LoaderPartial, Loader, LoaderType],
+    *args: Any,
+    frame_offset: int = 1,
+    **kwargs: Any,
+) -> Union[_cache_call, _cache_context]:
+    del loader, args, kwargs, frame_offset
+    raise TypeError(f"Invalid type for cache: {type(arg)}")
```
akshayka, line 378: Pretty clean!

akshayka (commented):
Clean implementation! Couple questions on signature changes.

marimo/_save/save.py
gabrielgrant: IIUC this can dispatch to either `_cache_call` or `_cache_context`, but they take different args/kwargs, so variability is needed. Rather than using `Any`, tho, I think this could have more specific type annotations using [unpack](https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs) and a union those two signatures' args/kwargs?

dmadisetti:
Tried to throw in some overloading to clean up the type signatures.

---

From a comment I left:

> Single dispatch cues off only the first argument, and expects a similar signature for every overload: https://peps.python.org/pep-0443/
> However.                                                                                                                                                                                                                                                                  
> The context and call APIs are slightly different, so `*` expansions are used to propagate that information down to the actually implementation.
> As such, we also leverage the `@overload` decorator to provide the correct signature and documentation for the > singledispatch entry points.

---

That being said, overloading seems to do interesting things in Jedi. Cursor over persistent cache gives the implementation doc string
![image](https://github.com/user-attachments/assets/5c6bddc0-2349-48ec-97a5-1d3e0b487bcf)

But opening it as a function gives the first overload

![image](https://github.com/user-attachments/assets/8f2a255b-7815-4d92-96ab-f4802ae06643)

Also unsure what will happen when docs tries to deploy this.

Updated docs are a little half baked (but acceptable) until we decide how we want to handle this, or whether the signature hacks are going a bit too far.

marimo/_save/save.py
dmadisetti: Cheers!

marimo/_save/save.py
dmadisetti: Let me know what you think of the overloading? Could walk it back a little too

dmadisetti: (that being said, I added examples in some of the overload docs)

marimo/_save/save.py
dmadisetti: `Any` is pretty consistent with other uses of `*` packing in the codebase, but I did leave a comment, since `*` is more to keep values ambiguous until they're fed into the correct constructors

marimo/_save/save.py
dmadisetti: In addition to matching the signature throughout, extra `kwargs` are now used by Loaders to enable:

```python
with MyLoader.cache("my_cache", constructor_values=123) as c:
    ...
```

or

```python
@MyLoader.cache
def my_fn(args):
    ...
```

a bit of feature creep, but even when testing these changes, I noticed how cumbersome 

```python
with cache(_loader=MyLoader.partial(constructor_values=123)) as c:
    ...
```

was, and seemed like low hanging fruit.

dmadisetti:
Ok, so only the implementation doc strings are rendered:

image

Which means, probably best to have the big doc strings on the implementation, and maybe more minor ones on the `@overloads`. Or is this a jedi bug?

akshayka:
> Which means, probably best to have the big doc strings on the implementation, and maybe more minor ones on the @overloads. Or is this a jedi bug?

Hmm, the docs aren't rendered using `jedi` but using [mkdocs](https://mkdocstrings.github.io/python/).

For completion, I think I wouldn't worry too much about what how `jedi` renders. We are exploring moving to an LSP anyway.

dmadisetti:
I was saying showing both doc strings in the documentation panel.
But true, LSP is likely to change it. I'll remove the little stubs I wrote

akshayka:
> Which means, probably best to have the big doc strings on the implementation

But yes, that makes sense to me.

akshayka (commented):
I'm happy with the overloading approach! LGTM modulo moving main docstrings to the implementation, if that's what you think is best.

marimo/_save/save.py
gabrielgrant: Yea, to be clear, wasn't saying that needed to happen, just that more specific typing was an option if desired -- while a nice addition, `unpack` is pretty recent and not super widely used (in marimo and beyond, ime)

Thanks for your work on this! I'm trying this branch out on a couple long computations rn 🚀

akshayka (approved):

PR #3563, opened 2025-01-24
feat: SQL engines

mscolnick:
_Heaving lifting done by @Light2Dark, I am just the messenger._ 

-------

This PR adds support for different SQL engines. First you define an engine like:

```python
import marimo as mo
from sqlalchemy import create_engine

engine = create_engine("snowflake://")
```

Then in all future SQL cells, you can access this engine. We auto-discover any variables that are SQL engines and populate the dropdown. Engine discovery is similar to how handle dataframe discovery: when we process a global variable to send to the variable panels, we also check if its a supported engine. 

This currently supports `duckdb` and `SQLAlchemy` as engines, with the setup in place to support many others. 

image


marimo/_runtime/runner/hooks_post_execution.py
```diff
@@ -121,6 +126,26 @@ def _broadcast_datasets(
...
 
+@kernel_tracer.start_as_current_span("broadcast_data_source_connection")
+def _broadcast_data_source_connection(
+    cell: CellImpl,
+    runner: cell_runner.Runner,
+    run_result: cell_runner.RunResult,
+) -> None:
+    del run_result
+    engines = get_engines_from_variables(
+        [(variable, runner.glbls[variable]) for variable in cell.defs]
```
akshayka, line 137: Check if `variable` is in `runner.glbls`

marimo/_sql/engines.py
```diff
@@ -0,0 +1,100 @@
...
+        return "sqlalchemy"
+
+    @property
+    def dialect(self) -> str:
+        return str(self._engine.dialect.name)
+
+    def execute(self, query: str) -> Any:
+        # Can't use polars.imported() because this is the first time we
+        # might come across polars.
+        if DependencyManager.polars.has():
```
akshayka, line 79: Also check for pandas here since it's used to read the query

PR #3568, opened 2025-01-24
fix: allow editable installs when devloping locally with sandbox

mscolnick:
cc @gabrielgrant

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -8,6 +8,7 @@ import signal
 import subprocess
 import sys
 import tempfile
+from pathlib import Path
 from typing import Any, Dict, List, Optional, cast
 
 import click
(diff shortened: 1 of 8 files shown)


akshayka (approved):

gabrielgrant:
thanks for the quick response to this! 

(for ref -- based on convo here: https://discord.com/channels/1059888774789730424/1332190342228742224 and #3561 )

flekschas:
I was trying to find out how to install a local package in editable mode and came across this ticket as couldn't find any documentation about it. After some digging, I realized I can get it working by manually setting the [metadata](https://docs.marimo.io/guides/package_reproducibility/#auto-tracking-inline-script-metadata) as follows and then running the notebook in sandbox mode:

```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "jupyter-scatter",
#     "marimo",
#     "numpy==2.2.2",
#     "pandas==2.2.3",
# ]
#
# [tool.uv.sources]
# jupyter-scatter = { path = "../", editable = true }
# ///
```

Hope this is useful for other folks who might come across this ticket and wonder the same thing.

Could be helpful to add a note to the docs somewhere. Maybe here https://docs.marimo.io/guides/editor_features/package_management/ or https://docs.marimo.io/guides/package_reproducibility/#adding-and-removing-packages?

mscolnick:
@flekschas, great call, thanks for the suggestion: https://github.com/marimo-team/marimo/pull/3786

PR #3595, opened 2025-01-28
fix: persistent loader save path (+ feat: test tweaks / json loader for interpretability)

dmadisetti:
Fixes the issue where the first function call had a different save path from subsequent calls (just inconsistent path setting reported by @akshayka )

Added some tests, and json loader that I was using to debug (also a feature suggestion from @leventov)

### DIFF
--- a/marimo/_save/loaders/json.py
+++ b/marimo/_save/loaders/json.py
@@ -1,0 +1,35 @@
+# Copyright 2024 Marimo. All rights reserved.
+from __future__ import annotations
+
+import dataclasses
+import json
+
+from marimo._save.cache import Cache, CacheType
+from marimo._save.loaders.loader import BasePersistenceLoader, LoaderError
+
+
+class JsonLoader(BasePersistenceLoader):
+    """Readable json loader for basic objects."""
+
+    def __init__(self, name: str, save_path: str) -> None:
+        super().__init__(name, "json", save_path)
+
+    def load_persistent_cache(
+        self, hashed_context: str, cache_type: CacheType
+    ) -> Cache:
+        with open(self.build_path(hashed_context, cache_type), "rb") as handle:
+            cache = json.load(handle)
+            # Handle unserializable stateful_refs
+            cache["stateful_refs"] = set(cache["stateful_refs"])
+            try:
+                return Cache(**cache)
+            except TypeError as e:
+                raise LoaderError(
+                    "Invalid json object for cache restoration"
+                ) from e
+
+    def save_cache(self, cache: Cache) -> None:
+        with open(self.build_path(cache.hash, cache.cache_type), "w") as f:
+            dump = dataclasses.asdict(cache)
+            dump["stateful_refs"] = list(dump["stateful_refs"])
+            json.dump(dump, f, indent=4)
--- a/marimo/_save/loaders/loader.py
+++ b/marimo/_save/loaders/loader.py
@@ -1,11 +1,14 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
+import os
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, Any, Type
 
 from marimo._runtime.context import get_context
+from marimo._runtime.context.types import ContextNotInitializedError
+from marimo._runtime.runtime import notebook_dir
 from marimo._runtime.state import State
 from marimo._save.cache import CACHE_PREFIX, Cache, CacheType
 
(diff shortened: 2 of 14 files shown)


marimo/_save/loaders/json.py
```diff
@@ -0,0 +1,35 @@
...
+            except TypeError as e:
+                raise LoaderError(
+                    "Invalid json object for cache restoration"
+                ) from e
+
+    def save_cache(self, cache: Cache) -> None:
+        with open(self.build_path(cache.hash, cache.cache_type), "w") as f:
+            dump = dataclasses.asdict(cache)
+            dump["stateful_refs"] = list(dump["stateful_refs"])
+            json.dump(dump, f, indent=4)
```
mscolnick, line 35: should we slap a version on this json?

marimo/_save/loaders/loader.py
```diff
@@ -22,7 +25,10 @@
 )
 
 
-LoaderType = TypeVar("LoaderType", bound="Loader")
+class LoaderError(BaseException):
```
akshayka, line 28: Can you add a comment why this is subclasses `BaseException` and not `Exception`?

marimo/_save/loaders/loader.py
```diff
@@ -50,14 +56,21 @@ def __call__(self, name: str) -> Loader:
...
+        except ContextNotInitializedError:
+            return State(self(name), _name=name, _context=context)
         if ctx.state_registry is None:
             return State(self(name), _name=name, _context=context)
 
         loader_state: State[Loader] | None = ctx.state_registry.lookup(
             name, context=context
         )
         if loader_state is None:
+            # There's a change it's in the registry, but the reference is None.
```
akshayka, line 70:
```suggestion
            # There's a chance it's in the registry, but the reference is None.
```

marimo/_save/loaders/loader.py
```diff
@@ -113,10 +126,10 @@ def cache_attempt(
             )
         loaded = self.load_cache(hashed_context, cache_type)
         # TODO: Consider more robust verification
-        assert loaded.hash == hashed_context, INCONSISTENT_CACHE_BOILER_PLATE
-        assert set(defs | stateful_refs) == set(loaded.defs), (
-            INCONSISTENT_CACHE_BOILER_PLATE
-        )
+        if loaded.hash != hashed_context:
+            raise LoaderError("Hash mismatch in loaded cache.")
+        if set(defs | stateful_refs) != set(loaded.defs):
```
akshayka, line 131: Might be misunderstanding something, but the `set` is not needed right?

```suggestion
        if (defs | stateful_refs) != set(loaded.defs):
```

marimo/_save/save.py
```diff
@@ -693,6 +698,7 @@ def persistent_cache(
 def persistent_cache(  # type: ignore[misc]
     name: Union[str, Optional[Callable[..., Any]]] = None,
     save_path: str | None = None,
+    method: str = "pickle",
```
akshayka, line 701: I know there was a discussion of configuring a set of loaders to be used together, rather than choosing a single loader method: https://github.com/marimo-team/marimo/discussions/2653#discussioncomment-11464871

What bearing does that discussion have on this proposed API? I would prefer just one way to specify the serialization type — if we plan to do that at a notebook level, per type, then `method` wouldn't be needed as an argument here.

akshayka (commented):
The fix looks good; left a question about the loader `method` API.

marimo/_save/save.py
dmadisetti: Thanks for pointing that out! I did miss that comment. Commenting in thread and pasting here:

---

I do think a default loader on a notebook level makes sense in various contexts, leveraging a general `persistent_cache` seems the best way to make it override-able at the notebook or user setting.

On a per object basis, this definitely increases the complexity for marimo- I think it's OK to rely on out of the box functionality, and let the user handle what should be explicitly saved in what format, e.g.

```python
with persistent_cache("big df", method="arrow"):
     my_df = compute_df()

with persistent_cache("np", method="numpy"):
     my_arr = compute_np()
```

Opposed to something like:

```
with Loader("cache", (np.array, "npy"), (pd.dataframe, "arrow"))
     my_df = compute_df()
     my_arr = compute_np()
```
There are tons of great serializations libraries out there, and being careful and intentional around the blocks you choose to cache and the correct method, will always out perform a general approach. I think that doing this under the hood may make sense, but it should be still over-ridable if desired (e.g. pickle is surprisingly fast, but does use more memory than other serialization methods)

marimo/_save/loaders/json.py
dmadisetti: Yep. Should probably happen at some point. I put this on the Cache level, so it applies to the pickle objects too

marimo/_save/save.py
akshayka: Ah, I saw your comment in the discussion and commented there first.

Like I wrote in the discussion I prefer the second API, I think it's the better developer experience and it's still explicit. That said, your current API could be extended to the second one in the future ...

dmadisetti: `str | dict[Type, str]` but yup, I think that works opposed to a separate `method` vs `methods`.

akshayka: Sounds good. Update the docstring to document `method`?

dmadisetti: Done!

akshayka (approved):

PR #3609, opened 2025-01-28
tidy: make tests stricter to better capture cache state invalidation

### LINKED ISSUE #3583: Adding and removing `@mo.cache` can cause state to improperly be flushed
### Describe the bug

Adding and removing `@mo.cache` from a function can flush the internal state variable and cause continuous state lookup misses.

This seems relatively non deterministic (at least locally) and seems to best be reproduced by:
1. removing the cache decorator
2. Executing the cell and doing a few other changes
3. Adding back the decorator and then executing the function in quick succession (e.g. rapidly changing the slider)

Likely related to the non-determinism in this test (Note how there's a range of accepted values): https://github.com/marimo-team/marimo/blob/c0b9fd11745b620c74406f7189730fed58291794/tests/_save/test_cache.py#L831

Also noticed, that just waiting after the decorator gets in this state, and eventually things seem to go back to normal

Reported by @akshayka 

### Environment

HEAD

### Code to reproduce

```python
# Copyright 2024 Marimo. All rights reserved.

import marimo

__generated_with = "0.10.15"
app = marimo.App(width="columns")


@app.cell(column=0)
def _(foo):
    foo()
    return


@app.cell
def _(get_context, slider):
    from pathlib import Path
    from marimo._save.loaders.json import JsonLoader
    # @mo.cache
    def foo():
        import random
        return slider.value + random.random()

    slider.value, get_context(
    ).state_registry, get_context(
    ).state_registry._states, get_context(
    ).state_registry._inv_states, get_context(

    ).state_registry.lookup("cache_partial:foo"), get_context(
    ).state_registry._states.get("cache_partial:foo").ref()()
    return JsonLoader, Path, foo


@app.cell
def _(slider):
    from marimo._runtime.context import get_context

    (z:=1), slider.value, get_context(
    ).state_registry, get_context(
    ).state_registry._states, get_context(
    ).state_registry._inv_states, get_context(

    ).state_registry.lookup("cache_partial:foo"), _ref.ref()() if (_ref :=get_context(
    ).state_registry._states.get("cache_partial:foo")) else "Nope"
    return get_context, z


@app.cell(column=1, hide_code=True)
def _(mo, slider):
    mo.md(
        f"""
        {slider}.

        ## {slider.value}
        """
    )
    return


@app.cell
def _(foo, threads):
    import threading
    for thread in threading.enumerate():
        print(thread.name, thread.native_id)
        threads.add(thread.native_id)
    foo(), len(threads)
    return thread, threading


@app.cell
def _():
    threads = set({})
    return (threads,)


@app.cell
def _(mo):
    slider = mo.ui.slider(1, 20)
    return (slider,)


@app.cell
def _():
    import marimo as mo
    return (mo,)


if __name__ == "__main__":
    app.run()
```

dmadisetti:
resolves #3583

Ran with 

```bash
for x in (seq 50);  pytest -vv tests/_save/test_cache.py || break; end
```

Also ran the same command @ 755a8558 (commit before #3595) and got failures at about run 5-10.

In the playground, I was unable to replicate the same failing behavior. Came across cases where the cache object was flushed, but it wasn't continually flushed like previously (https://marimo.app/?v=0.10.18-dev16&slug=tmgyxl)

### DIFF
--- a/tests/_save/test_cache.py
+++ b/tests/_save/test_cache.py
@@ -777,7 +777,10 @@ class TestCacheDecorator:
         # the registry. The actual cache hit is less important than caching
         # occurring in the first place.
         # 2 * 9 + 2
-        assert k.globals["fib"].hits in (9, 18, 20)
+        assert k.globals["fib"].hits not in (9, 18), (
+            "Potentially flaky edge case."
+        )
+        assert k.globals["fib"].hits == 20
 
     async def test_cross_cell_cache_with_external_ui(
         self, k: Kernel, exec_req: ExecReqProvider
@@ -828,7 +831,10 @@ class TestCacheDecorator:
         assert k.globals["b"] == 55
         assert k.globals["impure"] == [60, 157, 60]
         # 2 * 9 + 2
-        assert k.globals["fib"].hits in (9, 18, 20)
+        assert k.globals["fib"].hits not in (9, 18), (
+            "Potentially flaky edge case."
+        )
+        assert k.globals["fib"].hits == 20
 
     async def test_rerun_update(
         self, k: Kernel, exec_req: ExecReqProvider
--- a/tests/_save/test_cache.py
+++ b/tests/_save/test_cache.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import textwrap
+import warnings
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


dmadisetti:
Redundant assert more for git blame context in case this starts failing again

dmadisetti:
Oh nice. Still a little flaky. Not a big deal, since I think it's only in this thrashing case. Could throw a warning

mscolnick (approved):

PR #3626, opened 2025-01-30
improvement: use ast for sql parsing

mscolnick:
Use the lezer AST for SQL parsing instead of regex for better reliability.


Light2Dark (approved):
awesome! im guessing it's performant too, which is nice

Issue #3623, opened 2025-01-30
Memory Leak and/or Sessions not Cleaned Up


mixxen:
### Describe the bug

We are using Marimo to provide a user interface dashboard for our users. The app has multiple tables, multiple forms, and a few Altair data visualizations. It also uses Polars to cache data for the application.

We've noticed that after a short period, Marimo consumes a huge amount of memory. After investigation, we found that every time we refresh the page, Marimo allocates more memory and consumes more CPU, eventually saturating the host resources and becoming extremely slow and unusable.

Below you can see Marimo's memory usage after refreshing the page 10 times. In this test, there is no data in the Polars data frames:

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND 
1197755 root      20   0 3647480 659596 172288 S  83.5   1.0   0:47.85 marimo 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND 
1197755 root      20   0 5846168   2.2g 172376 S 115.9   3.6  14:12.20 marimo
```

Here is a Grafana view on our server over a couple of hours (just testing with a couple of users). The data in the Polars data frames is very minimal for this test:

![Image](https://github.com/user-attachments/assets/40709321-0f78-4f4b-9996-9df3118b5b4d)

Memory consumption spikes when someone hits the dashboard. After they exit, memory is not reclaimed, and over time, the app becomes noticeably slower.

We tried running Marimo like this:

```
marimo run src/ui/pages/main.py --host 0.0.0.0 --port 8000
```

```
uvicorn ui.app:make_app --factory --host 0.0.0.0 --port 8000 --timeout-keep-alive 300 --ws-ping-interval 1 --ws-ping-timeout 60 --timeout-graceful-shutdown 1
```

I've searched the documentation and issues for anything regarding session cleanup, etc. I did not find anything. We could use help to determine if this is a bug in Marimo or if we are doing something incorrectly on our end. Marimo is a pretty great idea; thank you in advance.

### Environment

<details>

```
app-1  | {
app-1  |   "marimo": "0.10.17",
app-1  |   "OS": "Linux",
app-1  |   "OS Version": "6.8.0-1021-aws",
app-1  |   "Processor": "x86_64",
app-1  |   "Python Version": "3.10.16",
app-1  |   "Binaries": {
app-1  |     "Browser": "--",
app-1  |     "Node": "--"
app-1  |   },
[...]

PR #3627, opened 2025-01-30
fix: handle periodic cleanup for sessions

mscolnick:
This may help #3623, but I have not tested yet. 

I will test this manually in the meantime, and try to make a reliable test out of it.

### DIFF
--- a/marimo/_server/sessions.py
+++ b/marimo/_server/sessions.py
@@ -68,6 +68,7 @@ from marimo._server.tokens import AuthToken, SkewProtectionToken
 from marimo._server.types import QueueType
 from marimo._server.utils import print_, print_tabbed
 from marimo._tracer import server_tracer
+from marimo._utils.debounce import debounce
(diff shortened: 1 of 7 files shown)


marimo/_server/sessions.py
```diff
@@ -1065,6 +1064,42 @@ def get_active_connection_count(self) -> int:
             ]
         )
 
+    # Check every 1 minute.
+    # We check the TTL later, so the 1 minute check is just to avoid
+    # unnecessary work.
+    @debounce(60)
+    def cleanup_orphaned_sessions(self) -> None:
```
akshayka, line 1071: Perplexing bug; I would have thought that

https://github.com/marimo-team/marimo/blob/0d719dd64756f3eb10c73a62193fb404166fde7a/marimo/_server/api/endpoints/ws.py#L468-L488

handled this case. Unless somehow the websocket's on_disconnect is not called?

akshayka (approved):

PR #3630, opened 2025-01-30
Change OUTPUT_MAX_BYTES to a config option

### LINKED ISSUE #3376: OUTPUT_MAX_BYTES should be a setting
HNY!

Just ran into the OUTPUT_MAX_BYTES warning, and it's quite a nuisance IMO.

1) there's no recovering from this helpful message - not great, especially in App mode because it effectively breaks the page till the app is restarted
2)  why does it `MARIMO_OUTPUT_MAX_BYTES` need to be an env var? it's not possible to change it without restarting everything - couldn't it be a setting? what's the idea of separating these rather random settings out into env vars?

```
marimo supports the following environment variables for advanced configuration:

| Environment Variable          | Description                                                                                                                  | Default Value   |
| ----------------------------- | ---------------------------------------------------------------------------------------------------------------------------- | --------------- |
| `MARIMO_OUTPUT_MAX_BYTES`     | Maximum size of output that marimo will display. Outputs larger than this will be truncated.                                 | 5,000,000 (5MB) |
| `MARIMO_STD_STREAM_MAX_BYTES` | Maximum size of standard stream (stdout/stderr) output that marimo will display. Outputs larger than this will be truncated. | 1,000,000 (1MB) |

```

3) OR could `mo._messaging.streams.OUTPUT_MAX_BYTES = 10_000_000` work similarly to `pd.options.display` (which can be changed  upfront in the notebook)?
4)  I have indeed produced large output, but it might not have been considered that it is spread across many lazy tabs (altair charts)

paddymul:
## 📝 Summary

Fixes #3376  [OUTPUT_MAX_BYTES should be a setting #3376](https://github.com/marimo-team/marimo/issues/3376)
## 📋 Checklist

Working through these

## 📜 Reviewers

@mscolnick

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -91,14 +91,28 @@ class RuntimeConfig(TypedDict):
     - `on_cell_change`: if `lazy`, cells will be marked stale when their
       ancestors run but won't autorun; if `autorun`, cells will automatically
       run when their ancestors run.
-    - `execution_type`: if `relaxed`, marimo will not clone cell declarations;
-      if `strict` marimo will clone cell declarations by default, avoiding
-      hidden potential state build up.
+    - `output_max_byes`:  Byte limits on outputs. Limits exist for two reasons:
+        1. We use a multiprocessing.Connection object to send outputs from
+           the kernel to the server (the server then sends the output to
+           the frontend via a websocket). The Connection object has a limit
+           of ~32MiB that it can send before it chokes
+           (https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Connection.send).
+
+        2. The frontend chokes when we send outputs that are too big, i.e.
+           it freezes and sometimes even crashes. That can lead to lost work.
+           It appears this is the bottleneck right now, compared to 1.
+
+           Usually users only output gigantic things accidentally, so refusing
+           to show large outputs should in most cases not bother the user too much.
+           In any case, it's better than breaking the frontend/kernel.
+
+        Output not shown if larger than OUTPUT_MAX_BYTES=5MB
     """
 
     auto_instantiate: bool
     auto_reload: Literal["off", "lazy", "autorun"]
     on_cell_change: OnCellChangeType
+    output_max_byes: int
 
 
 # TODO(akshayka): remove normal, migrate to compact
@@ -291,6 +305,7 @@ DEFAULT_CONFIG: MarimoConfig = {
         "auto_instantiate": True,
         "auto_reload": "off",
         "on_cell_change": "autorun",
+        "output_max_byes": 5_000_000
     },
     "save": {
         "autosave": "after_delay",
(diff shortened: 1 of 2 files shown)


paddymul:

marimo/_config/config.py
```diff
@@ -91,14 +91,28 @@ class RuntimeConfig(TypedDict):
     - `on_cell_change`: if `lazy`, cells will be marked stale when their
       ancestors run but won't autorun; if `autorun`, cells will automatically
       run when their ancestors run.
-    - `execution_type`: if `relaxed`, marimo will not clone cell declarations;
-      if `strict` marimo will clone cell declarations by default, avoiding
-      hidden potential state build up.
+    - `output_max_byes`:  Byte limits on outputs. Limits exist for two reasons:
```
paddymul, line 94: I copied this over from _messaging/streams.py.  It's probably a bit verbose for a GUI message.

marimo/_messaging/ops.py
```diff
@@ -153,9 +153,11 @@ def __post_init__(self) -> None:
 
     @staticmethod
     def maybe_truncate_output(
-        mimetype: KnownMimeType, data: str
+        mimetype: KnownMimeType,
+        data: str,
+        output_max_bytes: int = OUTPUT_MAX_BYTES,
```
paddymul, line 158: ideally we'd remove the default value.  I will check the other call sites to see if I can retrieve a config object there.

Or is it always possible to call `get_context()` inside of this function, so `output_max_bytes` doesn't even have to be an argument?

marimo/_config/config.py
mscolnick: i think we lost `execution_type` docs here. i think we can keep this (but don't need to add it below). its a hidden config that still might be in use

paddymul: That probably should have just been a comment in the PR because it distracts from this PR.  I removed the docs because I didn't see the config setting, I'll add it back in my PR.

paddymul: Fixed.

marimo/_messaging/ops.py
```diff
@@ -196,7 +198,12 @@ def broadcast_output(
         status: Optional[RuntimeStateType],
         stream: Stream | None = None,
     ) -> None:
-        mimetype, data = CellOp.maybe_truncate_output(mimetype, data)
+        marimo_conf = get_context().marimo_config
```
akshayka, line 201: `get_context` may throw a `ContextNotInitializedError`, which should be handled

marimo/_config/config.py
```diff
@@ -291,6 +308,7 @@ class PartialMarimoConfig(TypedDict, total=False):
         "auto_instantiate": True,
         "auto_reload": "off",
         "on_cell_change": "autorun",
+        "output_max_byes": 5_000_000,
```
akshayka, line 311: The config should be backward compatible with the environment variable, at least for now: so if the config is not set, we should check if MARIMO_OUTPUT_MAX_BYTES is defined.

akshayka (commented):
Thanks for the PR! Main comment is that we should take care to be backward compatible with people who are setting this via an environment variable. The config can take precedence over the environment variable, but we should still check for the presence of the latter.

paddymul:
functionality provided by #3858 closing

PR #3634, opened 2025-01-31
fix: kernel globals memory leak in run mode

akshayka:
This change fixes a memory leak in run mode in which a kernel's globals memory appeared to not be freed on session (thread) exit. However, another (smaller) leak appears to still exist.

This change also makes sure to stop the stream's buffered writer thread when kernel exits.

Also, add a TODO for performance of VariableValue construction -- calling str() on some values, such as a bytearray of size 1GB, can make memory usage spike to 25GB.

Hopefully improves #3623

### DIFF
--- a/marimo/_messaging/ops.py
+++ b/marimo/_messaging/ops.py
@@ -539,9 +539,9 @@ class VariableValue:
             if table_manager is not None:
                 return str(table_manager)
             else:
+                # TODO(akshayka): str(value) can be extremely expensive
+                # for some objects; find a better solution.
                 return str(value)[:50]
-
-            return str(value)[:50]
         except BaseException:
             # Catch-all: some libraries like Polars have bugs and raise
             # BaseExceptions, which shouldn't crash the kernel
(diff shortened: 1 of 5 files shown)


mscolnick (approved):

marimo/_messaging/ops.py
```diff
@@ -539,9 +539,9 @@ def _stringify(self, value: object) -> str:
             if table_manager is not None:
                 return str(table_manager)
             else:
+                # TODO(akshayka): str(value) can be extremely expensive
+                # for some objects; find a better solution.
                 return str(value)[:50]
```
dmadisetti, line 544: Some low hanging fruit could be to see if narwhals can digest the data (implements array interface if that's the case), and return a stringed slice based on that

mscolnick: we do use narwhals for the dataframe stuff. it doesn't handle just arrays. but I've written something custom to handle those: https://github.com/marimo-team/marimo/pull/3635

akshayka:
> Looks great, just a suggestion for `str` that could be a todo
> 
> 
> 
> I'm sure you know about the gc library, any insights into what else is accumulating?

I poked around but I'm not sure

PR #3644, opened 2025-01-31
fix: correctness of formatting

mscolnick:
This uses repr() instead of str() for formatting. 

Context: https://discord.com/channels/1059888774789730424/1334996923261648998

Specifically this is what Jupyter does and helps disambiguate between `True` and `"True"`, `1` and `"1"`.

This also makes numpy results much better 

Screenshot 2025-01-31 at 5 35 39 PM


akshayka (approved):
Looks great, thank!

PR #3664, opened 2025-02-03
fix: md format fix for columns and sql engine

dmadisetti:
SQL engines can be specified with `engine=` and column is now preserved in markdown blocks.
Documented changes in the tutorial file

@mscolnick

### DIFF
--- a/marimo/_tutorials/markdown_format.md
+++ b/marimo/_tutorials/markdown_format.md
@@ -1,7 +1,6 @@
 ---
 title: Markdown
 marimo-version: 0.10.19
-width: columns
 ---
 
 # Markdown file format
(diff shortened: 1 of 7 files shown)


marimo/_tutorials/markdown_format.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Markdown
-marimo-version: 0.10.9
+marimo-version: 0.10.19
+width: columns
```
mscolnick, line 4: Intentional?

tests/_cli/test_markdown_conversion.py
```diff
@@ -193,7 +193,11 @@ def test_markdown_with_sql() -> None:
 
     # SQL notebook
 
-    ```sql {.marimo query="export"}
+    ```python {.marimo}
+    mem_engine = mo.create_engine("sqlite:///:memory:")
```
mscolnick, line 197: nit: could we get rid of `mo.create_engine` in favor of `create_engine` in case people think its a real function when looking through the codebase

mscolnick (approved):

PR #3665, opened 2025-02-03
improvement: show engine's tables in the datasources panel

mscolnick:
This adds the SQLEngine's tables in the datasources panel. Also adds some icons/pretty displays (which we can later re-use in the dropdown).

The engine variable name is clickable to jump to its definition in the notebook

Screenshot 2025-02-02 at 10 37 17 PM


marimo/_sql/engines.py
```diff
@@ -59,16 +67,22 @@ def is_compatible(var: Any) -> bool:
...
+    def __init__(
+        self, engine: Engine, engine_name: Optional[str] = None
+    ) -> None:
         self._engine = engine
+        self._engine_name = engine_name
 
     @property
     def source(self) -> str:
-        return "sqlalchemy"
+        return str(self._engine.dialect.name)
```
Light2Dark, line 85: is this intentional?

mscolnick: yea this helps with a better name / icon in the frontend. if the naming is confusing, we could not call is `source` (although i don't think the `source` is `"sqlalchemy"` either)

Light2Dark: gotcha, its fine then. the test needs to be fixed but yea.

Light2Dark (dismissed):

akshayka (dismissed):

PR #3681, opened 2025-02-04
Add openai-whisper to module name list

koaning:
Fix the issue where `whisper` is not picked up as `openai-whisper`. 

image


akshayka (approved):
Thanks!

Issue #3494, opened 2025-01-18
Unable to interrupt/cancel operations that use Databricks Connect and Ibis


kyrre:
### Describe the bug

We are using Ibis and the Databricks Connect (Spark Connect) backend to run queries over large datasets. 

However, if I mess up a query and try cancelling it using the "Stop (interrupt) execution" button, then marimo will register the interruption request:

```plain
[10:59:06 AM]
STDERR
(cell-12)
[34m[D 250118 10:59:06 handlers:25](B[m Interrupt request received
```

but then just continue waiting forever (even after the query is finished in Databricks).

If I hit push the "Stop (interrupt) execution" button twice the exception will be raised and I get the following trace:

<details>

```plain

Traceback (most recent call last):
  File "/Users/kyrre.wahl.kongsgaard/projects/cdc-dashboards/.venv/lib/python3.12/site-packages/pyspark/sql/connect/client/core.py", line 648, in __iter__
    for response in self._call:
                    ^^^^^^^^^^
  File "/Users/kyrre.wahl.kongsgaard/projects/cdc-dashboards/.venv/lib/python3.12/site-packages/grpc/_channel.py", line 543, in __next__
    return self._next()
           ^^^^^^^^^^^^
  File "/Users/kyrre.wahl.kongsgaard/projects/cdc-dashboards/.venv/lib/python3.12/site-packages/grpc/_channel.py", line 960, in _next
    _common.wait(self._state.condition.wait, _response_ready)
  File "/Users/kyrre.wahl.kongsgaard/projects/cdc-dashboards/.venv/lib/python3.12/site-packages/grpc/_common.py", line 156, in wait
    _wait_once(wait_fn, MAXIMUM_WAIT_TIMEOUT, spin_cb)
  File "/Users/kyrre.wahl.kongsgaard/projects/cdc-dashboards/.venv/lib/python3.12/site-packages/grpc/_common.py", line 116, in _wait_once
    wait_fn(timeout=timeout)
  File "/Users/kyrre.wahl.kongsgaard/.local/share/uv/python/cpython-3.12.7-macos-aarch64-none/lib/python3.12/threading.py", line 359, in wait
    gotit = waiter.acquire(True, timeout)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3703, opened 2025-02-05
compat: interrupt using KeyboardInterrupt

akshayka:
Interrupt the kernel using a keyboard interrupt instead of a custom exception, for compatibility with libraries like databricks-connect that do custom handling of keyboard interrupts.

Fixes https://github.com/marimo-team/marimo/issues/3494

### DIFF
--- a/marimo/_runtime/control_flow.py
+++ b/marimo/_runtime/control_flow.py
@@ -3,15 +3,10 @@ from typing import Optional
 
 from marimo._output.rich_help import mddoc
 
-
-class MarimoInterrupt(BaseException):
-    """Raised when user stops execution of entire program with interrupt.
-
-    Inherits from `BaseException` to prevent accidental capture with
-    `except Exception` (similar to `KeyboardInterrupt`)
-    """
-
-    pass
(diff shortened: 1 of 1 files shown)


mscolnick (approved):

marimo/_runtime/control_flow.py
```diff
@@ -3,15 +3,10 @@
...
-
-    Inherits from `BaseException` to prevent accidental capture with
-    `except Exception` (similar to `KeyboardInterrupt`)
-    """
-
-    pass
+# We used to define a custom MarimoInterrupt BaseException to interrupt the
+# kernel; however, some third-party libraries like databricks-connect have
+# special case handling of KeyboardInterrupt.
+MarimoInterrupt = KeyboardInterrupt
```
dmadisetti, line 9: Does inherit work?

akshayka:
It might. But for compatibility, it might be better to just do what
ipykernel does, ie raise a KeyboardInterrupt?

dmadisetti (approved):
parity with ipykernel seems reasonable and more likely to stay compatible

PR #3721, opened 2025-02-07
fix: anywidget partial updates

mscolnick:
Only send updates of the changed state, otherwise will send the full data when e.g. selection changes

cc @manzt

### DIFF
--- a/frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx
+++ b/frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx
@@ -203,14 +203,16 @@ const LoadedSlot = ({
   return <div ref={ref} />;
 };
 
-class Model<T extends Record<string, any>> implements AnyModel<T> {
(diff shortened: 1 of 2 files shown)


akshayka (dismissed):

frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx
```diff
@@ -257,11 +259,20 @@ class Model<T extends Record<string, any>> implements AnyModel<T> {
...
-    this.onChange(this.data);
+    if (this.dirtyFields.size === 0) {
+      return;
+    }
+    const partialData: Partial<T> = {};
+    this.dirtyFields.forEach((key) => {
+      partialData[key] = this.data[key];
+    });
+    this.dirtyFields.clear();
+    this.onChange(partialData as T);
```
manzt, line 275: think this `as` can be removed

mscolnick: oh yes

Issue #3624, opened 2025-01-30
Propogate column information when cell selected in `mo.ui.Table`


jp-gr:
### Description

In the `mo.ui.Table` component (https://docs.marimo.io/api/inputs/table/) it's possible to select one or more rows when clicking on the table. I propose adding more selection modes to make it possible to select a cell and get both the row and column information associated with that cell.

In our use case we have some fairly wide dataframes and would like our users to be able to trigger some charts on more than one dimension by just clicking one cell.

### Suggested solution

In the `mo.ui.Table` component we have the argument:

```
    selection: Optional[
        Literal["single", "multi"]
    ] = "multi",
```

At the moment this allows two modes:

- `single`: only one row can be selected
- `multi`: many rows can be selected

I propose adding a new option to this:

- `cell`: when a cell is clicked then both the row and column information is sent to the defined `on_change` callback.

At the moment, the `on_change` callback is called with a list of dicts of all the information in the selected rows. I'm not sure what is a clean way to do this, but two suggestions come to mind:

- Keep this first argument the same but add an additional argument which contains the row_idx and the column name which was selected.
- Add the row_idx and column name into the existing argument e.g.:

```
[{'first_name': 'Michael', 'last_name': 'Scott'}]
```

Would become:

```
[{'first_name': 'Michael', 'last_name': 'Scott', 'row_idx': 1, column_selected: 'first_name'}]
```

mscolnick:
Makes sense, I think we can support this. I would see the options as:

```
single -> implies single-row (for back-compat)
multi -> implies multi-row (for back-compat)

single-cell
multi-cell

# maybe layer
single-column
multi-column
```

In call cases, the response will likely be the same shape passed in. So if a dataframe is pass in, the `on_change` and `.value` will be a dataframe. if its a list/dict, itll be a list/dict.

jp-gr:
Thanks for the quick reply!

I agree in terms of the data returned. Our main use case is actually dataframes, so the user would necessarily get the row index in the dataframe case. However in the list/dict case how would you propose giving the user enough metadata to make it useful?

For our dataframe use case we have data like:
```
pd.DataFrame({
        "day": ["2020-01-01", "2020-01-02"],
        "id": [1, 1],
        "metric1": [1.0, 2.0],
        "metric2": [2.0, 4.0],
    })
```  

If I had set this up as a list/dict instead of a dataframe and the user clicked the `4.0` value, would you propose passing in something like the following?
```
[...]

PR #3725, opened 2025-02-08
Add additional selection modes

nojaf:
## 📝 Summary

Fixes https://github.com/marimo-team/marimo/issues/3624

## 🔍 Description of Changes

This is a WIP, but I started with adding the new selection modes to both the JavaScript and Python side. 

I still need to to the actual implementation, but since I'm new to this project I could benefit from some early feedback.

Open questions:
- [ ] Where should the checkbox in the cell be positioned? What do you folks have in mind here UI-wise?
- [ ] Should `initial_selection` be taking into account here? Or would that remain row specific?

## 📋 Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable). (I believe the conversation in https://github.com/marimo-team/marimo/issues/3624 satisfies this)

## 📜 Reviewers

@mscolnick


nojaf:

marimo/_plugins/ui/_impl/tables/narwhals_table.py
```diff
@@ -100,6 +101,11 @@ def select_rows(self, indices: list[int]) -> TableManager[Any]:
     def select_columns(self, columns: list[str]) -> TableManager[Any]:
         return self.with_new_data(self.data.select(columns))
 
+    def select_cells(self, cells: list[Cell]) -> TableManager[Any]:
+        df = self.as_frame()
+        # This does not work and I could use some help
+        return self.with_new_data([df.at[row, col] for row, col in cells])
```
nojaf, line 107: Hello @mscolnick, any ideas what I need to here?
I'm a bit confused what exactly should be returned here.
This code doesn't work because of narwhal stuff

image

But I'm also unclear if I need this.

Because I'm unsure how this would play out if the `on_change` callback?

frontend/src/components/data-table/cell-selection/feature.ts
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    table.getIsAllCellsSelected = () => {
+      const state = table.getState().cellSelection ?? [];
+      return (
+        state.length === table.getRowCount() * table.getAllColumns().length
+      );
+    };
+
+    table.getIsAllPageCellsSelected = () => false; // TODO: I don't quite have a notion of pages yet
```
mscolnick, line 70: i think we can remove this. i think its when you use a checkbox at the top of the table to select all within the page, so i don't think itll make sense

frontend/src/components/data-table/cell-selection/feature.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  createCell: <TData extends RowData>(
+    cell: Cell<TData, unknown>,
+    column: Column<TData>,
+    row: Row<TData>,
+    table: Table<TData>,
+  ) => {
+    cell.getIsSelected = () => {
+      const state: CellSelectionState = table.getState().cellSelection ?? [];
+      return state.some(
```
mscolnick, line 81: this technically can be slow (for each cell, for each selected cell), which is `O(n^2)`. this could be `O(n)` instead, but may make other downstream logic more difficult, so fine to leave as is. maybe a comment though

mscolnick: here is a datastructure you may find helpful to use instead of the array.

feel free to remove some of the functions you don't use


```ts
class OrderedDoubleKeyMap<K1, K2, V> {
    private map: Map<K1, Map<K2, V>> = new Map();
    private orderArray: Array<[K1, K2]> = [];

    set(key1: K1, key2: K2, value: V): this {
        // Create nested map if it doesn't exist
        if (!this.map.has(key1)) {
            this.map.set(key1, new Map());
        }

        const nestedMap = this.map.get(key1)!;
        
        // Only add to order array if key pair doesn't exist
        if (!nestedMap.has(key2)) {
            this.orderArray.push([key1, key2]);
        }

        nestedMap.set(key2, value);
        return this;
    }

    get(key1: K1, key2: K2): V | undefined {
        return this.map.get(key1)?.get(key2);
    }

    has(key1: K1, key2: K2): boolean {
        return this.map.has(key1) && this.map.get(key1)!.has(key2);
    }

    delete(key1: K1, key2: K2): boolean {
        if (!this.has(key1, key2)) return false;

        const nestedMap = this.map.get(key1)!;
        nestedMap.delete(key2);

        // Remove from order array
        const index = this.orderArray.findIndex(
            ([k1, k2]) => k1 === key1 && k2 === key2
        );
        if (index !== -1) {
            this.orderArray.splice(index, 1);
        }

        // Clean up empty nested map
        if (nestedMap.size === 0) {
            this.map.delete(key1);
        }

        return true;
    }

    /**
     * Toggles a value at the specified keys.
     * If the value doesn't exist, it's added with the provided value.
     * If the value exists, it's removed.
     * @returns true if the value was added, false if it was removed
     */
    toggle(key1: K1, key2: K2, value: V): boolean {
        if (this.has(key1, key2)) {
            this.delete(key1, key2);
            return false;
        } else {
            this.set(key1, key2, value);
            return true;
        }
    }

    /**
     * Alternative toggle implementation that returns the current state
     * @returns the current value after toggling, or undefined if removed
     */
    toggleAndGet(key1: K1, key2: K2, value: V): V | undefined {
        if (this.has(key1, key2)) {
            this.delete(key1, key2);
            return undefined;
        } else {
            this.set(key1, key2, value);
            return value;
        }
    }

    clear(): void {
        this.map.clear();
        this.orderArray = [];
    }

    size(): number {
        return this.orderArray.length;
    }

    *entries(): IterableIterator<[K1, K2, V]> {
        for (const [key1, key2] of this.orderArray) {
            const value = this.get(key1, key2);
            if (value !== undefined) {
                yield [key1, key2, value];
            }
        }
    }

    forEach(callback: (value: V, key1: K1, key2: K2) => void): void {
        for (const [key1, key2] of this.orderArray) {
            const value = this.get(key1, key2);
            if (value !== undefined) {
                callback(value, key1, key2);
            }
        }
    }
}
```

frontend/src/components/data-table/cell-selection/types.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface CellSelectionInstance<TData extends RowData> {
+  setCellSelection: (updater: Updater<CellSelectionState>) => void;
+
+  toggleAllRowsSelected: (value?: boolean) => void;
+
+  resetCellSelection(defaultValue?: CellSelectionState): void;
+
+  getIsAllCellsSelected: () => boolean;
+
+  getIsAllPageCellsSelected: () => boolean;
```
mscolnick, line 38: can you make all these fields optional? its a have-to-know footgun of the table plugin system. since plugins are optional. the other tables may not have these methods and they could error (we use the table in the variable sidebar)

frontend/src/components/data-table/columns.tsx
```diff
@@ -307,8 +355,11 @@ function getFilterTypeForFieldType(
 function getCellStyleClass(
   justify: "left" | "center" | "right" | undefined,
   wrapped: boolean | undefined,
+  isSelected: boolean,
 ): string {
   return cn(
+    "cursor-pointer", // TODO: only when select cell?
+    isSelected && "bg-blue-200",
```
mscolnick, line 362: can you use `bg-[var(--blue-2)]` or some number from there. this will be dark-mode compatible

frontend/src/components/data-table/columns.tsx
```diff
@@ -307,8 +355,11 @@ function getFilterTypeForFieldType(
 function getCellStyleClass(
   justify: "left" | "center" | "right" | undefined,
   wrapped: boolean | undefined,
+  isSelected: boolean,
 ): string {
   return cn(
+    "cursor-pointer", // TODO: only when select cell?
```
mscolnick, line 361: yea ideally you can check `canSelectCell` higher up and pass that down

frontend/src/components/data-table/pagination.tsx
```diff
@@ -27,10 +28,19 @@ export const DataTablePagination = <TData,>({
...
     const numRows = table.getRowCount();
-    const isAllSelected = selected === numRows;
+    let isAllSelected = selected === numRows;
+
+    const isCellSelection =
+      selection === "single-cell" || selection === "multi-cell";
+    if (isCellSelection) {
+      selected = cellSelection.length;
+      isAllPageSelected = table.getIsAllPageCellsSelected();
+      isAllSelected = table.getIsAllCellsSelected();
```
mscolnick, line 42: i think we don't need to support `isAllPageSelected` and `isAllSelected`, so this can be `false`.

we also may want to remove the top checkbox from the checkbox column (if not already)

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -99,7 +101,10 @@ type Functions = {
   }>;
 };
 
-type S = Array<string | number>;
+// TODO: this was Array<number | string>
+// I don't quite get this. Does it represent the row-id?
+// Is the row-id the index property in our case?
+type S = Array<{ rowId: string; columnName?: string }>;
```
mscolnick, line 107: yea i think `Array<number | string>` was a mistake. i think `Array<number>` is likely what is shouldve been. but what you have now looks correct

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -508,20 +516,44 @@ const DataTableComponent = ({
...
-        setValue(Object.keys(nextValue));
+        setValue(Object.keys(nextValue).map((r) => ({ rowId: r })));
+      }
+    },
+  );
+
+  const cellSelection = useMemo(
+    () => value.filter((v) => v.columnName !== undefined) as CellSelectionState,
+    [value],
+  );
```
mscolnick, line 544: i would remove the memo and just do:

```
const cellSelection = value.filter((v) => v.columnName !== undefined);
const nextValue = Functions.asUpdater(updater)(cellSelection);
```

frontend/src/plugins/impl/DataTablePlugin.tsx
mscolnick: actually, since the backend does support just an list of `int`, lets do `Array<number | { rowId: string; columnName: string }>`

marimo/_plugins/ui/_impl/table.py
```diff
@@ -440,18 +448,32 @@ def _get_banner_text(self) -> str:
...
+            ]
+            self._has_any_selection = len(coordinates) > 0
+            return self._searched_manager.select_cells(coordinates)  # type: ignore
         else:
-            self._selected_manager = self._searched_manager.select_rows(
-                indices
-            )
-        self._has_any_selection = len(indices) > 0
-        return unwrap_narwhals_dataframe(self._selected_manager.data)  # type: ignore[no-any-return]
+            indices = [int(v["rowId"]) for v in value]
```
mscolnick, line 467: we are doing some unnecessary transformation to dictionary and back. if its a list of ints. lets just keep it that way

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -508,20 +516,44 @@ const DataTableComponent = ({
...
 
   const handleRowSelectionChange: OnChangeFn<RowSelectionState> = useEvent(
     (updater) => {
       if (selection === "single") {
         const nextValue = Functions.asUpdater(updater)({});
-        setValue(Object.keys(nextValue).slice(0, 1));
+        setValue(
+          Object.keys(nextValue)
+            .slice(0, 1)
+            .map((r) => ({ rowId: r })),
```
mscolnick, line 530: i think it would be preferable to keep this as a list instead of paying the cost to transform it to a dictionary and back

tests/_plugins/ui/_impl/test_table.py
```diff
@@ -253,7 +257,7 @@ def test_value_with_sorting_then_selection() -> None:
             page_number=0,
         )
     )
-    assert list(table._convert_value(["0"])) == [
+    assert list(table._convert_value([{"rowId": "0"}])) == [
```
mscolnick, line 260: i don't think we should add more memory pressure if we don't need. a list of numbers or strings is fine for row selection

frontend/src/plugins/impl/DataTablePlugin.tsx
mscolnick: i think its a string because `Object.keys()` can weirdly return strings if i recall, so i think thats why we made it that. probably better to just handle `Array<number | string | { rowId: string; columnName: string }>`

frontend/src/components/data-table/cell-selection/feature.ts
nojaf: I've added a note here, depending on the usage of this within G-Research, I might circle back to this.

mscolnick:
these changes look great!

mscolnick:
@nojaf, was just waiting till it was marked ready for review. ill look through it now

mscolnick:
@nojaf, it looks like backend typechecking is failing

mscolnick:
@nojaf, you can `/* eslint-disable @typescript-eslint/no-empty-interface */` at the top

nojaf:
@mscolnick thanks!

PR #3735, opened 2025-02-09
feat: add databases in datasource

Light2Dark:
## 📝 Summary

https://github.com/user-attachments/assets/1da6aef9-6f69-4886-a648-398e8cd4e518

---

**TODOs:**
- [x] search tables function is not working correctly, silent previewColumn firing may not work
- [X] add tests, fix typing
- [ ] stress test with large db
- [X] test with different db's. The worst thing to do is it hangs when there's an error in the bg. It should fail and return no 
     - [X] snowflake
     - [X] bigquery
     - [x] mysql
     - [X] postgres
     - [X] sqlite
     - [x] mssql
- [ ] test with `session.bind` flow
- [ ] multi-db / not specifying db. Idt sqlalchemy supports db discovery, so 1 way is to execute raw sql queries depending on dialect (i.e `show databases;`)

## 🔍 Description of Changes

Experimenting a little, I find an optimal request solution is something like below. Make an initial req for db, schemas and table list, upon expanding a table, make another request to get column info, PKs, indexes. This avoids too long of an initial request time while fetching sufficient info. With snowflake sample db (7 schemas, 10+ tables each), initial request takes about 3s. If I fetched all data, this would take ~25s

image

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


mscolnick:
@Light2Dark, a user on [Discord](https://discord.com/channels/1059888774789730424/1338959377411084390/1338959377411084390) had 100+ tables. We don't need to lazy load tables too just yet, but could be worth keeping a lookout / design in such a way to support it later. I do think by avoiding loading the columns its only 2 round trips (instead of a fan-out).

frontend/src/components/editor/chrome/panels/datasources-panel.tsx
```diff
@@ -143,76 +152,310 @@ export const DataSourcesPanel: React.FC = () => {
...
+  databaseName,
+  schemaName,
+  tables,
+}) => {
+  const [data, setData] = React.useState<DataTable | null>(null);
+  const [loading, setLoading] = React.useState<boolean>(false);
+  const [error, setError] = React.useState<string | null>(null);
+
+  // TODO: Move all of this to table list view, not here.
+  React.useEffect(() => {
```
mscolnick, line 376: we also have `useAsyncData` which handles the boiler plate

frontend/src/core/datasets/data-source-connections.ts
```diff
@@ -9,11 +10,10 @@ export type ConnectionName = TypedString<"ConnectionName">;
 
 export const DEFAULT_ENGINE = "_marimo_duckdb" as ConnectionName;
 
-export interface DataSourceConnection {
+// Extend the backend type but override the name property with the strongly typed ConnectionName
+export interface DataSourceConnection
+  extends Omit<DataSourceConnectionType, "name"> {
```
mscolnick, line 15: very nice

frontend/src/core/functions/FunctionRegistry.ts
```diff
@@ -15,3 +21,13 @@ export const FUNCTIONS_REGISTRY = new DeferredRequestRegistry<
     ...req,
   });
 });
+
+export const PreviewSQLTable = new DeferredRequestRegistry<
```
mscolnick, line 25: let me know if its worth wrapping `DeferredRequestRegistry` in an LRU cache. happy to write a class for that if you'd want

marimo/_data/models.py
```diff
@@ -66,6 +72,33 @@ class DataTable:
     variable_name: Optional[VariableName]
     columns: List[DataTableColumn]
     engine: Optional[VariableName] = None
+    type: DataTableType = "table"
+    primary_keys: Optional[List[str]] = None
+    indexes: Optional[List[str]] = None
```
mscolnick, line 77: this is amazing!

marimo/_runtime/runner/hooks_post_execution.py
```diff
@@ -165,14 +167,21 @@ def _broadcast_data_source_connection(
...
-            Datasets(tables=tables).broadcast()
+        if isinstance(engine, SQLAlchemyEngine):
+            databases = engine.get_databases(
+                include_schemas=True,
+                include_tables=True,
+                include_table_details=False,
+            )
+        elif isinstance(engine, DuckDBEngine):
+            databases = engine.get_databases()
+        if databases:
```
mscolnick, line 178: `databases` could be unbound if not of the engines exist. so I would declare 
 `databases: Optional[Databases]` on line 170

marimo/_server/api/endpoints/datasources.py
```diff
@@ -48,3 +51,69 @@ async def preview_column(
         from_consumer_id=ConsumerId(app_state.require_current_session_id()),
     )
     return SuccessResponse()
+
+
+@router.get(
+    "/preview_sql_table/{engine:path}/{database:path}/{schema:path}/{table_name:path}"
+)
+@requires("edit")
+async def preview_sql_table(request: Request) -> BaseResponse:
```
mscolnick, line 60: `get` is totally fine, and does feel like it aligns more with REST semantics, but we've been using posts with a body because:

1. the body does validation
2. its an object that can be passed around, e.g. `PreviewSQLTableRequest`
3. its easier to make API changes

openapi/api.yaml
```diff
@@ -472,6 +488,38 @@ components:
       - time
       - unknown
       type: string
+    Database:
```
mscolnick, line 491: adding the new objects to `marimo/_cli/development/commands.py` should result in a cleaner yaml file

marimo/_server/api/endpoints/datasources.py
Light2Dark: I see, one reason I thought of using get is that libraries like react-query would cache the request. `post` makes that a little tricky. But, I'm happy to move to post, the reasons make sense.

mscolnick: we can still cache with react-query, we would just need to serialize the params. since we pass the params to the code-generated client anyways, i don't think we will have access to the http url to cache on

openapi/api.yaml
Light2Dark: should be kinda clean already, the objects show up nicely 🤔

Light2Dark:
hey @mscolnick, the frontend seems a little tricky. Is there a reason you're putting the following state in jotai? Instead of local to a component
```.tsx
function initialState(): DatasetsState {
  return {
    tables: [],
    expandedTables: new Set(),
    expandedColumns: new Set(),
    columnsPreviews: new Map(),
  };
}
```

openapi/api.yaml
mscolnick: oh shoot, sorry i misread this

mscolnick:
@Light2Dark, jotai was originally used because request and responses came from different places. if you want to `useAsyncData` and keep the data local, that is fine. but we would need to handle caching too.

Light2Dark:
hey @mscolnick, if you have the time, would be nice to get some eyes on `datasources.tsx`. I made some liberal changes which may not be in the direction you initially planned. Happy to revert/change the approach though.

marimo/_sql/engines.py
```diff
@@ -15,6 +22,10 @@
 if TYPE_CHECKING:
     import duckdb
     from sqlalchemy import Engine
+    from sqlalchemy.sql.type_api import TypeEngine
+
+# Internal engine name for DuckDB, wwe need to ensure this is unique
```
mscolnick, line 27: `wwe` -> `we`

marimo/_sql/get_engines.py
```diff
@@ -37,9 +41,26 @@ def engine_to_data_source_connection(
...
 ) -> DataSourceConnection:
+    databases: list[Database] = []
+    if isinstance(engine, SQLAlchemyEngine):
+        databases = engine.get_databases(
+            include_schemas=True,
+            include_tables=True,
+            include_table_details=False,
+        )
+    elif isinstance(engine, DuckDBEngine):
+        databases = engine.get_databases()
```
mscolnick, line 52: could we add an `else: LOGGER.warn`

mscolnick:
@Light2Dark i do like the new design. I really like the engines being separate and that you added the "Python" sesion. Couple pieces that might've got regressed or would be nice to support:

1. the top-most items have rounded borders. Maybe this could be removed?
Screenshot 2025-02-17 at 12 59 49 PM

2. I think the third+ row could have left-side rounded borders. Or the background could extend to all the way to the left.

Screenshot 2025-02-17 at 1 02 49 PM

3. The keyboard navigation doesn't feel as smooth as it was before. I think its because we can multiple TableList (and hence CommandList). Could we instead have one? And when we can select a DB, Schema, etc using keyboard navigation and on Enter it will expand.

mscolnick:
@Light2Dark - sorry made a small change to imports after you rebased. hopefully a quick fix. 

happy to get this merged early if its ready and it can soak on main for a bit (i can try to stress test it too)

Light2Dark:
> @Light2Dark - sorry made a small change to imports after you rebased. hopefully a quick fix.
> 
> happy to get this merged early if its ready and it can soak on main for a bit (i can try to stress test it too)

yeap no worries, I'm also testing it with other flavours, but it looks good so far!

frontend/src/components/datasources/datasources.tsx
```diff
@@ -0,0 +1,833 @@
...
+        icon={<DatabaseIcon />}
+      />
+    );
+  }
+
+  const hasSearch = !!searchValue.trim();
+
+  return (
+    <Command
+      className="border-b bg-background rounded-none h-full pb-10 overflow-auto"
```
mscolnick, line 138: i noticed a blue border (maybe always was there). we can remove that with `outline-none`

frontend/src/components/datasources/datasources.tsx
```diff
@@ -0,0 +1,833 @@
...
+}> = ({ connection, children, hasChildren }) => {
+  const hasEngine = connection.databases.length > 0;
+  // If the connection has no engine, it's the internal duckdb engine
+  const engineName = hasEngine
+    ? connection.databases[0]?.engine || "In-Memory"
+    : connection.name;
+
+  return (
+    <>
+      <DatasourceLabel>
```
mscolnick, line 225: maybe in future, we could add a super small refresh icon that only appears on hover (using `hover-action` and `hover-action-parent` class names), that would refresh the tables

mscolnick:
just played around with it, feels and looks great!

Light2Dark:
made a fix for mssql, not going to touch this for now.

mscolnick (approved):

PR #3745, opened 2025-02-10
fix: merge anywidget partial updates

mscolnick:
Keeps the full state on the backend, merges partial updates


manzt (approved):
LGTM, thanks for quick fix and tests.

PR #3754, opened 2025-02-11
example: turn a notebok into api endpoints

mscolnick:
This is a simple example of how to use FastAPI with marimo. This example turns marimo notebooks into an API endpoint, which can be embedded in any FastAPI app.

- Turning functions defined in a notebook into an API endpoint
- Overriding global variables and returning cell outputs


akshayka (approved):
Ha, cool!

PR #3773, opened 2025-02-12
Secret Manager

aditya-arolkar-swe:
## 📝 Summary

Implemented a basic secret manager

## 🔍 Description of Changes

Added`set_secret`, `get_secret` and `delete_secret` methods to the App class 

## 📋 Checklist

## 📜 Reviewers

@mscolnick


aditya-arolkar-swe:

marimo/_ast/app.py
```diff
@@ -171,7 +172,7 @@ class App:
     function.
     """
 
-    def __init__(self, **kwargs: Any) -> None:
+    def __init__(self, allow_secrets: bool = True, **kwargs: Any) -> None:
```
dmadisetti, line 175: If secrets are default on, what are instances where someone would want to turn them off? Do we even need the flag?

dmadisetti (commented):
Cool! I think that secret management is something marimo needs.

Couple thoughts on implementation though:

1. the way marimo serializes means that the app level information will be overwritten every single time, so exposing the api through app would mean changes to codegen
2. Seems to put a lot of burden on the user (manually inputting or/ and requiring a hardcoded input)
3. `app` is not super accessible through the runtime, so it would be hard to actually use in a notebook

I really like how colab manages secrets (https://medium.com/@parthdasawant/how-to-use-secrets-in-google-colab-450c38e3ec75) or kaggle (https://chioujryu.medium.com/kaggle-feature-launch-user-secrets-42b080d705ac). There are a few UX choices that I think makes these work well:

 1. Ephemeral setting though the interface. There's no chance of doing `app.set_secret()` and forgetting it in source. A dialog workflow (like you have with getpass) could replace the panel they have (although the panel seems reasonable)
 
 2. Mechanism to share secrets easily between notebooks (you can just turn the toggle on or off from the panel)
 
 3. Does not require setup every time. Serialization means that this will only work at start up and then re-opening the notebook will require the hardcoded values 
 
 ---
 
 Colab has it easy since they can just utilize a google data store and be responsible for your data.
 
For saving anything locally, it's imperative to use  known/ platform compatible solutions.
For instance after unlocking your GPG keyring, a session is opened to decrypt and encrypt values freely: https://gnupg.readthedocs.io/en/latest/ (maybe `keyring` preferred on Windows/Mac)

---

But can see the work put in to this and thank you!

marimo/_ast/app.py
aditya-arolkar-swe: Was aiming to put a CLI argument that can flag --no-secrets if we want that turned off for any reason. Didn't fully implement that, but left pieces of the logic between functions.

aditya-arolkar-swe:
> Cool! I think that secret management is something marimo needs.
> 
> Couple thoughts on implementation though:
> 
>     1. the way marimo serializes means that the app level information will be overwritten every single time, so exposing the api through app would mean changes to codegen
> 
>     2. Seems to put a lot of burden on the user (manually inputting or/ and requiring a hardcoded input)
> 
>     3. `app` is not super accessible through the runtime, so it would be hard to actually use in a notebook
> 
> 
> I really like how colab manages secrets (https://medium.com/@parthdasawant/how-to-use-secrets-in-google-colab-450c38e3ec75) or kaggle (https://chioujryu.medium.com/kaggle-feature-launch-user-secrets-42b080d705ac). There are a few UX choices that I think makes these work well:
> 
>     1. Ephemeral setting though the interface. There's no chance of doing `app.set_secret()` and forgetting it in source. A dialog workflow (like you have with getpass) could replace the panel they have (although the panel seems reasonable)
> 
>     2. Mechanism to share secrets easily between notebooks (you can just turn the toggle on or off from the panel)
> 
>     3. Does not require setup every time. Serialization means that this will only work at start up and then re-opening the notebook will require the hardcoded values
> 
> 
> Colab has it easy since they can just utilize a google data store and be responsible for your data.
> 
> For saving anything locally, it's imperative to use known/ platform compatible solutions. For instance after unlocking your GPG keyring, a session is opened to decrypt and encrypt values freely: https://gnupg.readthedocs.io/en/latest/ (maybe `keyring` preferred on Windows/Mac)
> 
> But can see the work put in to this and thank you!

Thanks for the review and comments Dylan. I agree that attaching the method to App wasn't the best. I'm only just getting familiar to this repo and wasn't sure where to place it. I have added a commit that attempts to move it to the Marimo class instead. So now you can call it with: 
```
import marimo as mo
sm = mo.secret_manager()
sm.set_secret('test_key', 'test_value') 
```
Also, I initially used `keyring` as the backend package, but tried to do `os.environ` calls instead hoping that would reduce the dependencies. There will be leakage issues with that implementation though. 

>     3. Does not require setup every time. Serialization means that this will only work at start up and then re-opening the notebook will require the hardcoded values
How do you think this can be implemented then? I'm newer to web development, and am curious to know what sort of persistent data store can be implemented between sessions.

PR #3790, opened 2025-02-13
improvement: mo.latex() to load macros

### LINKED ISSUE #3781: KaTeX Macro Support
### Description

As outlined in https://github.com/marimo-team/marimo/discussions/1941, I would love to be able to have some Macro Support for KaTeX.
This would come in handy for repeated use of convoluted symbols and would really help us for teaching and presenting our stuff in notebooks.

Our current workflow for jupyter notebooks is a workaround, where we do
```python
import IPython.display
IPython.display.display_latex(IPython.display.Latex(filename="macros.tex"))
```
where `macros.tex` looks like
```
\newcommand{\rot}[1]{{\rm curl }\left( #1 \right)}
\newcommand{\Grad}[1]{{\rm Grad}\left( #1 \right)}
\newcommand{\Div}[1]{{\rm Div }\left( #1 \right)}
```

A feature like that would be great!

### Suggested solution

In an optimal case, we would love to be able to point to a file populated by KaTeX macro commands, which would then be available to use in all markdown cells without any additional import.

mscolnick:
Fixes #3781

### DIFF
--- a/docs/api/markdown.md
+++ b/docs/api/markdown.md
@@ -6,6 +6,15 @@ elements.
 
 ::: marimo.md
 
+## Loading LaTeX macros
+
+You can load LaTeX macros using `mo.latex(filename=...)`.
+
+::: marimo.latex
+
+!!! warning "Side effects"
+    The `mo.latex()` function has side effects (registering the LaTeX macros) and should be used in the same cell as `import marimo`. Otherwise, the LaTeX macros may not be loaded before the cells that use them.
+
 ## Icons
 
 We support rendering icons from [Iconify](https://icon-sets.iconify.design/).
--- a/marimo/_output/md.py
+++ b/marimo/_output/md.py
@@ -234,7 +234,7 @@ def latex(*, filename: Union[str, Path]) -> None:
     ```python
     import marimo as mo
 
-    mo.latex(url="https://example.com/macros.tex")
+    mo.latex(filename="https://example.com/macros.tex")
     ```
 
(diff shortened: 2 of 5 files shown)


akshayka (dismissed):

marimo/_output/md.py
```diff
@@ -215,3 +218,45 @@ def md(text: str) -> Html:
...
+
+    mo.latex(filename="macros.tex")
+    ```
+
+    or
+
+    ```python
+    import marimo as mo
+
+    mo.latex(url="https://example.com/macros.tex")
```
dmadisetti, line 237: @mscolnick is this functionality in? Or did you mean `filename`?

docs/api/markdown.md
```diff
@@ -6,6 +6,15 @@ elements.
 
 ::: marimo.md
 
+## Loading LaTeX macros
+
+You can load LaTeX macros using `mo.latex(filename=...)`.
+
+::: marimo.latex
+
+!!! warning "Side effects"
```
dmadisetti, line 15: Maybe a runtime error if md case been called before latex- or is that too mean?

marimo/_output/md.py
mscolnick: whoops, i did, thanks

docs/api/markdown.md
mscolnick: probably too mean. i think it is ok

PR #3796, opened 2025-02-14
fix: basic support for bokeh.plotting.show

akshayka:
Patches bokeh.plotting.show() to display its first argument in the output area. Other arguments are ignored. This is similar to our patch of `IPython.display.display`.

Not an elegant fix, but should work for many use cases. The proper fix would be to support `bokeh`'s `output_notebook()` — I believe this uses `IPython.display.publish_display_data`under the hood, which we don't yet support.

cc @bryevdv

image

### DIFF
--- a/marimo/_output/formatters/bokeh_formatters.py
+++ b/marimo/_output/formatters/bokeh_formatters.py
@@ -20,28 +20,15 @@ class BokehFormatter(FormatterFactory):
     def register(self) -> Callable[[], None]:
         import bokeh.models  # type: ignore[import-not-found,import-untyped,unused-ignore] # noqa: E501
         import bokeh.plotting  # type: ignore[import-not-found,import-untyped,unused-ignore] # noqa: E501
-        from bokeh.io import install_notebook_hook  # type: ignore[import-not-found,import-untyped,unused-ignore] # noqa: E501
-        from bokeh.io.state import curstate  # type: ignore[import-not-found,import-untyped,unused-ignore] # noqa: E501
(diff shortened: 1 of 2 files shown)


bryevdv:
For reference the default built-in Jupyter hooks are registered like this:
```python
install_notebook_hook('jupyter', load_notebook, show_doc, show_app)
```
If you used this API you could make your `load_notebook` (which is what `output_notebook` does) be a no-op, and you could make `show_app` emit an exception, since Bokeh server apps can't be run in a WASM notebook. Then `show_doc` would be your customized version of `show`.

akshayka:
Thanks @bryevdv for the feedback.

That mostly works, except it appears `output_notebook()` resets `curstate().notebook_type` to Jupyter, and I don't think I can expect users to manually pass a notebook type to this function (since they'll just copy tutorials like you mentioned).

So it seems like we'd at least need to monkey-patch `output_notebook` to set `curstate.notebook_type` to marimo and do nothing else?

bryevdv:
> That mostly works, except it appears output_notebook() resets curstate().notebook_type to Jupyter

If that's true it sounds like a bug or oversight on our end. Do you have a link to the code that you saw doing that handy?

akshayka:
> So I think there probably out to be a way to override the function parameter default, and it was just an oversight that there is not one.

Thanks, that appears to be it.

For now I think patching `show` and `output_notebook` is easiest. I experimented with using the hooks, but it appeared I still needed to call `output_notebook(notebook_type="jupyter")` for marimo's installed hook to take effect, but that brought in a dependency on IPython that we didn't need.

It's certainly possible I did something wrong, but happy with this simple patch for now.

mscolnick (approved):

PR #3807, opened 2025-02-14
feat: Session Schema and Serialize outputs 

### LINKED ISSUE #3082: Serialize outputs in separate file so app can be viewable when auto-instantiate is off
### Description

Serialize outputs in separate file so app can be viewable when autorun is off. This is different from persistent caching, even if configured on all cells. This is much cheaper and just stores the HTML of the outputs (which can be easily found in the `SessionView`), this does not "hydrate" the runtime, so all outputs will be considered stale.

This helps with users who always use `auto-instantiate=off`

### Suggested solution

SessionView has a serialize/deserialize function on it that can read/write to an outputs file. This will be stored in the shared `__marimo__` folder, similar to persistent cache and auto-export.

Proposed path would be `__marimo__/outputs.json`.
Other options could be `__marimo__/session.json`, if somehow it ends up containing more than outputs (variables, datasets schemas, etc)

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
Fixes #3082

This adds support for serializing/deserializing part of the SessionView so that we can re-hyrdate the outputs on startup. 
While this looks like it may conflict with caching (`mo.cache`), it has a much different use-case. This is a much simpler version for different tradeoffs:
- only storing output (not variables or previous runs)
- everything is stale on startup

This acts like Jupyter's output storage, but storage in a different file.

This adds:
- a code-first schema `NotebookSessionV1`
- we generate the openai spec (for easier code-reviews) and a GitHub Action to detect breaking changes
- serialization/deserialization
- `AsyncBackgroundTask` util that our `SessionCacheWriter` runs on
- `Session` starts this task, only when auto-instantiate if off

Note:
- timings are not handled in the ser/de or tests (yet), will handle in a followup
- If the path is `foo/bar/baz.py`, the cache file is →  `foo/bar/__marimo__/session/baz.py.json`

In a followup:
- Handle timings
- WASM support


akshayka (approved):
Awesome! Generally looks good to me. The design (using the marimo auxiliary folder, snapshotting every couple seconds when autorun is off) makes sense.

I assume that this is robust to cases when the code and cached session become inconsistent (eg, if the user deletes the __marimo__ folder, or manually edits the file).

tests/_server/session/snapshots/mime_bundle_session.json
```diff
@@ -0,0 +1,24 @@
+{
+  "version": "1",
+  "metadata": {
+    "marimo_version": "0.0.0"
+  },
+  "cells": [
```
dmadisetti, line 6: Wow, looks very ipynb

dmadisetti (approved):
Cool! Haven't run it, but is there a visual indicator that something is from saved session?

mscolnick:
@dmadisetti 

 > is there a visual indicator that something is from saved session?

There is an inferred one: when you have auto-instantiate false, you still see the outputs although its marked stale (yellow). 

I think next we can add small banner mentioning this (maybe with a `Run all` button to hydrate all the variables back into memory)

leventov:
@mscolnick why not centralise in a single "workspace" dir? `__marimo__/sessions/foo/bar/baz.py.json`

mscolnick:
@leventov, how are you inferring the workspace dir?

leventov:
@mscolnick I didn't see how the code locates this dir it already e.g. for `__marimo__/cache`. I assumed it already does looks at `./__marimo__`, `../__marimo__`, `../../__marimo__`, etc., for all parent dirs of the notebook's file, and chooses the first one that is present; if none is present, creates `./__marimo__`.

PR #3810, opened 2025-02-16
improvement: add index column for stable selection in table, persist across search/filter/sort

### LINKED ISSUE #3610: Table row selection not correctly preserved when searching
### Describe the bug

When using the search feature to filter a `mo.ui.table()`, and then making a selection after that, the selection is not preserved.

### Environment

<details>

```
{
  "marimo": "0.10.13",
  "OS": "Darwin",
  "OS Version": "23.1.0",
  "Processor": "arm",
  "Python Version": "3.12.0",
  "Binaries": {
    "Browser": "132.0.6834.112",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.22.0",
    "packaging": "24.2",
    "psutil": "6.1.1",
    "pygments": "2.19.1",
    "pymdown-extensions": "10.14",
    "pyyaml": "6.0.2",
    "ruff": "0.9.2",
    "starlette": "0.45.2",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "14.1"
  },
  "Optional Dependencies": {
    "pandas": "2.2.3"
  }
}
```

</details>

### Code to reproduce

```python
import marimo as mo
import pandas as pd

url = "https://raw.githubusercontent.com/mwaskom/seaborn-data/master/penguins.csv"
df = pd.read_csv(url)
mo.ui.table(df, selection='single')
```

mscolnick:
Fixes #3610

Previously we would reset selection on sort/filter/search because we didn't have a stable row id for selection. Now, for some dataframes, we add an internal selection column that is opaque to the user but allows us to keep track of the selection across sorting/filtering/searching.

This only works for eager dataframes since we can give them indexes. This won't work for dict/list or lazy dataframes like ibis - for these, we revert to the old behaviour.

### DIFF
--- a/frontend/src/components/data-table/TableActions.tsx
+++ b/frontend/src/components/data-table/TableActions.tsx
@@ -57,7 +57,7 @@ export const TableActions = <TData,>({
               ? (value: boolean) => {
                   if (value) {
                     const allKeys = Array.from(
-                      { length: table.getRowModel().rows.length },
+                      { length: table.getRowCount() },
                       (_, i) => [i, true] as const,
                     );
                     onRowSelectionChange(Object.fromEntries(allKeys));
--- a/frontend/src/components/data-table/__test__/columns.test.tsx
+++ b/frontend/src/components/data-table/__test__/columns.test.tsx
@@ -229,4 +229,16 @@ describe("generateColumns", () => {
     });
     expect(cell?.props.className).toContain("center");
   });
+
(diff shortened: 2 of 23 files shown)


akshayka (approved):
Smart, nice

PR #3813, opened 2025-02-16
fix: from_panel using private APIs

akshayka:
`from_panel` was using a private panel API that was recently removed; this caused our tests to fail and presumably broke our Panel integration.

This PR updates the plugin to use a public alternative, copying a recent change made in Panel.

Context: https://github.com/holoviz/panel/pull/7716/

(cc @philippjfr)


mscolnick (approved):

philippjfr (approved):

philippjfr:
Looks great, thanks for the fix!

PR #3832, opened 2025-02-18
chore: cleanup utils folder for less dependencies

mscolnick:
move some code / files around for less dependencies


dmadisetti:
Nice! Was thinking of this recently. it might be worth making a `marimo/_utils/types.py` or something

I think `visitor` gets imported a ton for `Name` and `CellId_t` might be defined several places

mscolnick:
yep was thinking the same. probably just `marimo/_types/`. i don't think they belong in util

PR #3852, opened 2025-02-19
fix: add back bigquery logo, move help into select dropdown

mscolnick:
Tiny bug fix for adding back the biquery logo. 

Some tests and better typesafety


Light2Dark:
thank you!

PR #3855, opened 2025-02-19
fix: mo.ui.date popover

mscolnick:


Yasin197:
it looks like the bug is still there in the 0.11.7 release

PR #3858, opened 2025-02-20
improve: configure output max bytes with pyproject.toml

### LINKED ISSUE #3376: OUTPUT_MAX_BYTES should be a setting

akshayka:
This adds runtime configuration to control the max size in bytes for outputs in console outputs. Previously these were only controlled by environment variables. This change lets apply their settings globally via a pyproject.toml, or locally for a file via inline script metadata.

Closes #3376


mscolnick (approved):
should we update the default if it's been requested so often?

PR #3864, opened 2025-02-20
compat: Register webbrowser.open hook for pyodide

dmadisetti:
## 📝 Summary

`import webbrowser.open` seems to be broken in WASM (which is fine, I think opening pages is potentially dangerous)- so this loads in a hook to insert an iframe into output if webbrowser is called instead of failing. As a result, the following code now works

```python
import antigravity
```

but with some marimo tweaks that might make for good social engagement on 25/04/01

image

(PS: this needs a static upload)

Had a spare hour and noticed it when looking through cpython/Lib
Let me know if it's too irreverent and I can walk back the feature- but seems enabling this seems like full python support to me

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_runtime/marimo_browser.py
+++ b/marimo/_runtime/marimo_browser.py
@@ -46,7 +46,7 @@ def browser_open_fallback(
     ):
         mo.output.append(
             mo.image(
-                "https://marimo.app/static/antigravity.png",
+                "https://marimo.app/images/antigravity.png",
                 alt=(
                     "The image shows 2 stick figures in XKCD style. The one "
                     'on the left says:"You\'re Flying! How?", a floating '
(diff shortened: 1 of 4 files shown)


akshayka (dismissed):
I love this! :)

mscolnick:
i can do the static upload

marimo/_runtime/marimo_browser.py
```diff
@@ -0,0 +1,91 @@
...
+    # For other python lore, try:
+    #   import this
+    stack = inspect.stack()
+    if len(stack) > 3 and (
+        stack[2].filename.endswith("antigravity.py")
+        or (stack[1].filename.endswith("antigravity.py"))
+    ):
+        mo.output.append(
+            mo.image(
+                "https://marimo.app/static/antigravity.png",
```
mscolnick, line 49:
```suggestion
                "https://marimo.app/images/antigravity.png",
```

mscolnick: waiting for upload

dmadisetti:
Very cool!

https://marimo.app/?slug=j7w20i&v=0.11.8-dev6

 Thanks! @mscolnick

Issue #3869, opened 2025-02-21
Code Is Repeated after Minimizing Marimo Interface


Nashat90:
### Describe the bug

Whenever Multitasking with marimo and other tools, once i go back to marimo after minimizing it, the code in all cells are repeated multiple times making the usage of the Marimo not very friendly.
I tried to upgrade to newer version. Same issue persists/ 

### Environment

<details>

```

{
  "marimo": "0.11.7",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 186 Stepping 2, GenuineIntel",
  "Python Version": "3.12.4",
  "Binaries": {
    "Browser": "132.0.6834.196",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.18.1",
    "itsdangerous": "2.2.0",
    "jedi": "0.18.1",
    "markdown": "3.7",
    "narwhals": "1.16.0",
    "packaging": "23.2",
    "psutil": "5.9.0",
    "pygments": "2.15.1",
    "pymdown-extensions": "10.12",
    "pyyaml": "6.0.1",
    "ruff": "0.8.2",
    "starlette": "0.41.3",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.11.0",
    "uvicorn": "0.32.1",
    "websockets": "14.1"
  },
  "Optional Dependencies": {
    "altair": "5.0.1",
    "anywidget": "0.9.13",
    "duckdb": "1.1.3",
    "pandas": "2.2.2",
    "polars": "1.22.0",
    "pyarrow": "14.0.2"
  },
  "Experimental Flags": {
    "multi_column": true,
    "tracing": true,
    "rtc": true
  }
}

```

</details>

### Code to reproduce

import pandas as pd
import warnings
warnings.simplefilter(action='ignore', category=FutureWarning)
df = pd.read_excel(r"C:\Users\ASUS\Desktop\...Pigging Aug\TARC Frac Job Analysis\Lone Star All.xlsx")import pandas as pd
import warnings
warnings.simplefilter(action='ignore', category=FutureWarning)
df = pd.read_excel(r"C:\Users\ASUS\Desktop\....Pigging Aug\TARC Frac Job Analysis\Lone Star All.xlsx")import pandas as pd
import warnings
warnings.simplefilter(action='ignore', category=FutureWarning)
[...]

PR #3877, opened 2025-02-21
chore: warn users about RTC code duplication

akshayka:
RTC is still a work in progress, and is known to duplicate user code (#3869). This change gives users a heads up so they can debug the issue on their own if it arises.


mscolnick (approved):

mscolnick:
Probably better to remove to the form (but keep the feature flag). 

Although harder for people to turn it off.

PR #3879, opened 2025-02-22
suggestion: cell id by hash

dmadisetti:
re: https://github.com/marimo-team/marimo/pull/3875

Regarding UIs between saves, this is something I was tinkering with prior to the stable ids in #3061.
I didn't really touch it after, but stable UIs were key for loaded UI cache objects too

Just sharing here, this is an old branch things might have shifted- just did a loose merge
@mscolnick


mscolnick:
let's chat live about this sometime. i have questions easier said in person

PR #3892, opened 2025-02-24
feat: koaning/motest parity (pytest in blocks)

dmadisetti:
## 📝 Summary

@koaning had the great idea of bringing tests editor side.
Part of that means being able to run tests that are runtime dependent.
This statically extracts tests and creates a wrapper that will call run on execution.

A little separate from the top level function push, but complimentary.

Works with a bit of a hack that includes injecting TestClasses into scope for pytest.
This was more of a proof of concept I ended up pushing to completion.

I think it's a useful addition- but not anything we explicitly put on our roadmap.
Happy to mark as draft and rethink this

## 🔍 Description of Changes

pytest on:

```python
@app.cell
async def _(pytest, x, y, Z):
    @pytest.mark.xfail(
        reason=("To ensure this doesn't just eval."),
        raises=AssertionError,
        strict=True,
    )
    def test_independence():
        assert x + y == 0

    def test_another_dependent_cell():
        assert x + y == Z

    class TestClass:
        @pytest.mark.xfail(
            reason=("To ensure this doesn't just eval."),
            raises=AssertionError,
            strict=True,
        )
        def test_method(self):
            assert x + y == 0

        @staticmethod
        @pytest.mark.xfail(
            reason=("To ensure this doesn't just eval."),
            raises=AssertionError,
            strict=True,
        )
        def test_static_method() -> None:
            assert x + y == 0

        def test_normal(self):
            assert x + y == Z

        @staticmethod
        def test_static() -> None:
            assert x + y == Z

    async def test_async_cell():
        assert True
```

will now work. Should be a native parity solution with [koaning/motest](https://github.com/koaning/motest)

## Next?

As for editor side, the cell is now compiled with a `__test__` flag that should denote if it can be "tested".
Here's a UI idea, but I'm not a UX person/ think research into other testing integrations might be a good idea first.

Untitled(4)

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/docs/guides/testing/pytest.md
+++ b/docs/guides/testing/pytest.md
@@ -12,7 +12,8 @@ For example,
 pytest test_notebook.py
 ```
 
-runs and tests all notebook cells whose names start with `test_`.
+runs and tests all notebook cells whose names start with `test_`, or cells that
+contain only `test_` functions and `Test` classes.
 
 !!! tip "Naming cells"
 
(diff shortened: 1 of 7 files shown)


mscolnick (commented):
nice! looks great. could you update `docs/guides/testing/pytest.md` too?

mscolnick (approved):

PR #3898, opened 2025-02-24
improvement: bring back SQL schema completion and AI '@' reference for datasources

mscolnick:
This brings the new datasources schemas to SQL completion in the SQL cells as well as `@` referencing them in the AI inputs

Couple todos, for later:
- [ ] We bring all table names to the top of the schema (so we don't need reference them by their fully qualified name), but we should only do this for the default db/schema
- [ ] AI completion does not handle duplicate table names (i.e. same table name from different DBs) correctly and will only pick the first one

### DIFF
--- a/frontend/src/components/editor/renderers/CellArray.tsx
+++ b/frontend/src/components/editor/renderers/CellArray.tsx
@@ -358,7 +358,7 @@ const AddCellButtons: React.FC<{
             onClick={isAiButtonOpenActions.toggle}
           >
             <SparklesIcon className="mr-2 size-4 flex-shrink-0" />
-            Generate with AI!!!
+            Generate with AI
           </Button>
         </Tooltip>
       </>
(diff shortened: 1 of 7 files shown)


frontend/src/components/editor/renderers/CellArray.tsx
```diff
@@ -358,7 +358,7 @@ const AddCellButtons: React.FC<{
             onClick={isAiButtonOpenActions.toggle}
           >
             <SparklesIcon className="mr-2 size-4 flex-shrink-0" />
-            Generate with AI!!!
```
mscolnick, line 361: i must have accidentally slipped this in here at some points 😬

mscolnick:
@Light2Dark, that sounds great!

>  We bring all table names to the top of the schema (so we don't need reference them by their fully qualified name), but we should only do this for the default db/schema

This is low priority. Oversharing in the dropdown is probably ok. It doesn't create conflict, just will throw an error from the DB (which might give you a hint)

> AI completion does not handle duplicate table names (i.e. same table name from different DBs) correctly and will only pick the first one

I think it is still nice to refer them from their short name (maybe the long name also exists in the completion info). But on accepting, we likely want to `apply` the full table name. There is an `apply` field on the completion. 


It would be great to have a set of utils. Maybe even a wrapper class around `DataSourceConnection` that has all these utility methods

mscolnick:
@Light2Dark, yes your logic is better. if i understand, use the shortest (and correct) name possible without conflicting

PR #3937, opened 2025-02-28
improve: error messages for import *, multiple definition

akshayka:
Provide more helpful error messages for import star and multiple definition errors, explaining the reasoning behind these errors.

This PR also changes error description text to be muted instead of red, matching the accordion text, based on feedback that too much red can be stressful. Headlines (which describe the exception type) are still red and bold.

image


mscolnick (approved):
looks good. I think we can merge many of the errors at some point to just type/msg/details

PR #3938, opened 2025-02-28
experimental: reactive pytest

dmadisetti:
## 📝 Summary

Follow up to: #3892

This allows eligible cells to automatically and reactively run pytest in a notebook:

image

It also hooks into the runtime, so tests don't have to run from scratch

This is currently hidden under an experimental flag for a few reasons:

1. I noticed live editing the cells sometimes puts them in an invalid state requiring another run (pytest loads the module off disk, and I think there's a bit of a lag)
2. similar to top level functions, I think that it seems like a bit of magic for when the functionality turns on. I think some more documentation and UX communication might be needed before an intentional release.
3. My unit tests aren't super robust and I haven't tested out every edge case
4. Would nice to get some feedback before pushing too much further
 
Just a note, this is tightly integrated with pytest, I don't think there are many other meaningful frameworks now nose is dead, but marimo should attempt to be agnostic.

## 📜 Reviewers

@akshayka OR @mscolnick @koaning

### DIFF
--- a/marimo/_runtime/runner/hooks_post_execution.py
+++ b/marimo/_runtime/runner/hooks_post_execution.py
@@ -1,6 +1,7 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
+import sys
 from typing import Callable
 
 from marimo import _loggers
@@ -337,6 +338,26 @@ def _broadcast_outputs(
             clear_console=False,
             cell_id=cell.cell_id,
         )
+
+
+@kernel_tracer.start_as_current_span("run_pytest")
+def _attempt_pytest(
+    cell: CellImpl,
+    runner: cell_runner.Runner,
+    run_result: cell_runner.RunResult,
+) -> None:
+    del run_result
+    if cell._test:
+        try:
+            import marimo._runtime.pytest as marimo_pytest
+
+            if runner.execution_context is not None:
+                with runner.execution_context(cell.cell_id):
+                    result = marimo_pytest.run_pytest(cell.defs, runner.glbls)
+                    if result.output:
+                        sys.stdout.write(result.output)
+        except ImportError:
+            pass
 
 
 @kernel_tracer.start_as_current_span("reset_matplotlib_context")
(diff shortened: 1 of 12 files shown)


akshayka (commented):
Quite interesting! I haven't really reviewed code yet, but I wanted to confirm that the idea is that even when moved out of experimental, this would be opt-in, correct? Otherwise I think it would be too magical

marimo/_runtime/runner/hooks_post_execution.py
```diff
@@ -339,6 +340,26 @@ def format_output() -> formatting.FormattedOutput:
...
 
 
+@kernel_tracer.start_as_current_span("run_pytest")
+def _attempt_pytest(
+    cell: CellImpl,
+    runner: cell_runner.Runner,
+    run_result: cell_runner.RunResult,
+) -> None:
+    del run_result
+    if cell._test:
```
akshayka, line 350: Only run if the runner isn't lazy?

dmadisetti: Implemented as a conditional post-execution hook. With the experimental flag is in place, this would be a no-op with lazy

But definitely for removing the flag. I was thinking maybe a separate run button?

mscolnick:
I'd make it opt-out, otherwise it won't be as discoverable. 

And otherwise, why else would you add pytest to your notebooks

akshayka:
> I'd make it opt-out, otherwise it won't be as discoverable.
> 
> And otherwise, why else would you add pytest to your notebooks

You might just call a function `test_` something, without "adding pytest" to your notebook. That might seem odd, but with a large number of users that will happen. It would be jarring if just defining a function caused it to run. If someone wants to write unit tests in a notebook (a somewhat nice but compelling use case), they will be sufficiently motivated to find it in the docs.

akshayka (approved):
🔧

akshayka:
@dmadisetti mind resolving the merge conflict? (sorry) otherwise good to go, thanks.

PR #3945, opened 2025-02-28
fix: top level function serialization broke in wasm example

dmadisetti:
## 📝 Summary

Explicitly catch IOError for notebook on top level serialization I was tinkering with: https://marimo.app/l/n0s9hn


akshayka (approved):

akshayka:
Failure looks unrelated, thanks for the fix.

Issue #3464, opened 2025-01-16
Named dataframes reset to default **_df** name after kernel restart


ZazenX:
### Describe the bug

Im running marimo On jupterhub using the [marimo proxy package](https://github.com/jyio/jupyter-marimo-proxy)

I've noticed upon kernel restart. Or restarting my singleuser notebook server and refreshing marimo, SQL cell output have a tendency to lose their variable name and reset to the default **_df** variable name. I'm using pandas as the dataframe backend.

### Environment

<details>

```
{
  "marimo": "0.10.13",
  "OS": "Linux",
  "OS Version": "5.15.0-118-generic",
  "Processor": "x86_64",
  "Python Version": "3.11.11",
  "Binaries": {
    "Browser": "--",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.22.0",
    "packaging": "24.2",
    "psutil": "6.1.1",
    "pygments": "2.19.1",
    "pymdown-extensions": "10.14",
    "pyyaml": "6.0.2",
    "ruff": "0.9.1",
    "starlette": "0.45.2",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.12.2",
    "uvicorn": "0.34.0",
    "websockets": "14.1"
  },
  "Optional Dependencies": {
    "duckdb": "1.1.3",
    "pandas": "2.2.3"
  }
}
```

</details>

mscolnick:
@ZazenX, after you change your `_df` name, you need to run the cell (since changing the name is changing the code behind the scenes). Are you sure you were running them and then saving the file? 

If not, that would make sense. Otherwise, I cannot reproduce this issue. If you can still reproduce this, can you include steps?

sswatson:
In my testing, code cells save their content on blur. That is, if I write some code into a cell, then click off of that cell, then restart my browser window, the code that I wrote is preserved.

I haven't thought through all the implications, but my feeling is that the output-variable field would ideally have the same behavior.

mscolnick:
@sswatson, i see, yes that would make sense. it's autosave on delay (instead of on blur). but we are likely not updating the code when you blur the input

sswatson:
That's what I observed in my testing yesterday.

Furthermore, I was also able to repro the behavior indicated in the OP.

[...]

PR #3960, opened 2025-03-02
Fix blur behavior for output-variable-name input field

sswatson:
## 📝 Summary

As discussed in #3464, there is currently a bug in the output-variable input field on SQL cells.

The problem is that if you change the variable name and do shift-enter, the name change doesn't propagate to `notebookState` and therefore doesn't persist to disk. When the same notebook is opened again later, therefore, these variable names are lost.

I believe that the reason for this is that there's some kind of race condition between the cell evaluation that occurs when you do shift-enter, and the state-update operation that is supposed to occur in response to the same event. However, I have not been able to track down the root cause of this behavior.

## 🔍 Description of Changes

This PR makes the smallest possible change to workaround this problem. It circumvents the race condition by making the variable-update happen before the cell execution, specifically on the `keydown` event.

The ideal fix for this problem might be to identify the race condition and adjust the event-handling mechanisms accordingly, but in my estimation this is a reasonable and effective approach in the meantime.

To check that this PR fixes the problem, make a Marimo notebook while on `main` and create a SQL cell that says `select 1`. Then change the output variable name, do shift enter, and then immediately refresh the browser page. You should find that the variable name has reverted. Then switch to this branch and repeat the experiment; you should find that the variable name persists.

## 📋 Checklist

## 📜 Reviewers

@mscolnick

P.S. I have not added tests yet because I'm not sure whether this is the approach the maintainers want to take.

### DIFF
--- a/frontend/src/core/codemirror/language/panel.tsx
+++ b/frontend/src/core/codemirror/language/panel.tsx
@@ -46,6 +46,14 @@ export const LanguagePanelComponent: React.FC<{
 
   if (languageAdapter instanceof SQLLanguageAdapter) {
     showDivider = true;
+    const sanitizeAndTriggerUpdate = (e: React.SyntheticEvent<HTMLInputElement>) => {
+      // Normalize the name to a valid variable name
+      const name = normalizeName(e.currentTarget.value, false);
+      languageAdapter.setDataframeName(name);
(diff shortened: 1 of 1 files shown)


sswatson:

mscolnick (approved):
awesome nice find, thank you

PR #3969, opened 2025-03-04
improvement: add pythonpath confiugration

mscolnick:
By default, marimo will not add any additional directories to the Python path. This keeps the behavior of `marimo edit nb.py` and `python nb.py` consistent.

You can add directories to the Python path by setting the `pythonpath` key in the
runtime configuration. These directories will be added to the head of `sys.path`,
similar to how the `PYTHONPATH` environment variable works. This allows Python to
find and import modules from these directories.

```toml title="pyproject.toml"
[tool.marimo.runtime]
pythonpath = ["project/src"]
```


akshayka (dismissed):
Looks great!

Issue #3015, opened 2024-12-01
export and convert with same filename lead to file overwrite


s-celles:
### Description

Hello,

export and convert marimo command can lead to overwrite existing file if (by mistake) user use same file name for input and output.

Best regards

### Suggested solution

if input and output file are the same marimo should ask confirmation from user y/N (default N)

if ouput file name is an ever existing file which will be overwriten it should also ask confirmation from user y/N

(we are all humans and make mistakes... if software can help to avoid some that will be great)

mscolnick:
@s-celles this would be nice to have. Any interest in contributing this one?

s-celles:
Sorry @mscolnick I'm very busy this month and won't have the time to tackle this.

PR #3987, opened 2025-03-05
Ask for confirmation if file name already exist instead of overriding it 

xeniumcode:
Adding confirmation  during export and covert if the filename  already exist instead direct override. 

Resolves : #3015 

## 📋 Checklist

@mscolnick


xeniumcode:
How to resolve this, remove temp_file after every test or add random file name for each test ?

marimo/_cli/export/commands.py
```diff
@@ -57,6 +57,16 @@ def watch_and_export(
 
     def write_data(data: str) -> None:
         if output:
+            output_path = Path(output)
+            # Check if the file exists
+            if output_path.exists():
```
mscolnick, line 62: we should probably only check once when `watch` is enabled, or not check at all.

could we also check if `GLOBAL_SETTINGS.YES`, is set and skip?

maybe we could pull this out to a util:

```python
def maybe_prompt_to_overwrite(path: Path) -> bool:
    if GLOBAL_SETTINGS.YES:
        return True

    # Check if not in an interactive terminal
    # default to False
    if not sys.stdin.isatty():
        return True
    
    if output_path.exists():
         return True

    confirmed = click.confirm(
        f"Warning: The file '{output}' already exists. Overwrite?",
        default=False,
    )
    
    if not confirmed:
        click.exit()
```

mscolnick:
> How to resolve this, remove temp_file after every test or add random file name for each test ?

@xeniumcode, i think removing the temp dir after each test would be good

marimo/_cli/convert/commands.py
```diff
@@ -62,6 +63,14 @@ def convert(
         notebook = convert_from_md(text)
 
     if output:
+        output_path = Path(output)
+        if prompt_to_overwrite(output_path):
+            confirmed = click.confirm(
+                f"Warning: The file '{output}' already exists. Overwrite?",
+                default=False,
+            )
```
mscolnick, line 71: could we put the click.confirm in the prompt_to_overwrite?

mscolnick:
@xeniumcode implementation looks great! Would it be possible to add a single test? If it is not simple, no worries, I can add one as well after it merges

mscolnick:
Actually I'll merge and follow up

PR #3995, opened 2025-03-06
Update module_name_to_pypi_name.py

koaning:
Add jsscatter dep. It's for this [anywidget](https://github.com/flekschas/jupyter-scatter).


koaning:
Nevermind. That package does not support marimo at the moment. 
CleanShot 2025-03-06 at 15 48 28

PR #4005, opened 2025-03-06
chore: use Path over open

mscolnick:
Prefer `Path().read/write` over `open`

### DIFF
--- a/marimo/_cli/development/commands.py
+++ b/marimo/_cli/development/commands.py
@@ -5,6 +5,7 @@ import ast
 import os
 import subprocess
 import sys
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import click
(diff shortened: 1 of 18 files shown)


marimo/_cli/development/commands.py
```diff
@@ -393,8 +395,7 @@ def inline_packages(name: str) -> None:
     package_names = module_name_to_pypi_name()
 
     def get_pypi_package_names() -> list[str]:
-        with open(name) as file:
-            tree = ast.parse(file.read(), filename=name)
+        tree = ast.parse(name.read_text(), filename=name)
```
akshayka, line 398: encoding utf-8?

akshayka (commented):
Looks good, one encoding missing maybe (already missing before this change)

PR #4022, opened 2025-03-07
Adding a check for file in convert cli. Earliar it used to raise long error

xeniumcode:
Earlier it used to raise a big error , now handling it with click utility
## 📋 Checklist

### DIFF
--- a/marimo/_cli/convert/commands.py
+++ b/marimo/_cli/convert/commands.py
@@ -50,6 +50,9 @@ def convert(
     you may need to fix errors like multiple definition errors or cycle
     errors.
     """
+    if not Path(filename).exists():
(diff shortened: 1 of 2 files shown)


marimo/_cli/convert/commands.py
```diff
@@ -50,6 +50,9 @@ def convert(
     you may need to fix errors like multiple definition errors or cycle
     errors.
     """
+    if not Path(filename).exists():
```
mscolnick, line 53: you can technically convert from a remote URL, so this would make that fail unfortunately. can we do this check elsewhere? i will also add a test so this is more apparent

mscolnick (approved):

PR #4025, opened 2025-03-07
mep/8: frontend function hints

dmadisetti:
## 📝 Summary

@mscolnick this was my very rough draft approach to passing hints upstream:

image

For reference, this is how observable renders their functions:

image

and what was suggested:

image

image

But I'm sure there's probably a more cohesive marimo style (do we have have any brand docs?):


akshayka:
No brand docs yet ... but that's a good idea.

At first glance it feels a bit repetitive to include the function signature in the output area? I like the idea of communicating hints and function-ness though ...

mscolnick:
@dmadisetti is the name configurable on the frontend? is it enforced by the backend?

when happens to the name if it goes from a pure function to non-pure function?

dmadisetti:
> @dmadisetti is the name configurable on the frontend? is it enforced by the backend?

Name will be overwritten as is, which seems reasonable? But the current UI does make it a bit jarring. Might be worth hiding in these cases.

> when happens to the name if it goes from a pure function to non-pure function?

Then the TopLevelStatus "hint" indicates that something has gone wrong- but the name is retained. It could be pushed back to "_"

---

Currently unindented comments after (or before) the function are stripped. I'm think we should retain the comments, and have that break "purity" if someone really wanted to intentionally keep it in cell mode.

dmadisetti:
> At first glance it feels a bit repetitive to include the function signature in the output area? I like the idea of communicating hints and function-ness though ...

Agreed. I think this is an issue shared with markdown too. Many possible UIs but one might be using the markdown style blocks of no border, and expansion on click

akshayka:
> > At first glance it feels a bit repetitive to include the function signature in the output area? I like the idea of communicating hints and function-ness though ...
> 
> Agreed. I think this is an issue shared with markdown too. Many possible UIs but one might be using the markdown style blocks of no border, and expansion on click

I guess — why do top-level functions need an output to begin with? If it's mostly to have a spot to show the status hint, on why a function is not pure, perhaps we can find a different design, such as some kind of tooltip.

dmadisetti:
I like the tooltip styling in setup-cell. Definitely doesn't have to display the function

dmadisetti:
image


Still pretty rough, feel free to commit directly. Icon differences might be a bit too subtle

marimo/_messaging/ops.py
```diff
@@ -321,6 +324,19 @@ def broadcast_stale(
     ) -> None:
         CellOp(cell_id=cell_id, stale_inputs=stale).broadcast(stream)
 
+    @staticmethod
+    def broadcast_serialization(
+        cell_id: CellId_t,
+        serialization: TopLevelStatus,
+        stream: Stream | None = None,
+    ) -> None:
+        status = (
```
mscolnick, line 333: could we make this hint typesafe (e.g `Literal[]`)?

dmadisetti: It's a dynamic value since the hint can be dependent on the refs etc... But I put something in to indicate it's from a static set of values, and mypy didn't seem to care.

dmadisetti:
Feel free to make UI tweaks to this branch.

As I see it, the outstanding issues for mep/8 are:

 - whitespace idempotent (I think this is easy enough, just requires tinkering with spacing)
 - handle invalid syntax in setup block (probably tack this on to above PR)
 - static app loading (a nice security feature, but I don't think it's a blocker)
 - class exports (@app.class_definition or a general `@app.export` that covers `@app.function` too)
 - documentation:
    - File format tutorial: (See work in progress https://marimo.app/l/0u57k9)
    - Update the editor feature page to mention the new `app.function` feature.
    - Create a new page discussing code reuse/using marimo notebooks as function libraries.
    - Document setup cells.
    
And then maybe let it bake in experimental. Release will require:
- tutorial re-serialization

PR #4026, opened 2025-03-07
experimental: setup cell UI

mscolnick:
UI for the setup cell. Doesn't do anything at the moment, and behind an undocumented feature flag / not exposed in the UI.

Screenshot 2025-03-10 at 7 41 34 PM
-------
Screenshot 2025-03-10 at 7 41 05 PM


dmadisetti:
Nice! Good start. The CellId doesn't actually get into SaveNotebookRequest for some reason, but once it does should serialize

Issue #3997, opened 2025-03-06
`show_code` and "Cell output area" above or below


moble:
### Description

I think many people who are used to IPython notebooks really struggle with seeing output above the code that produces it.  The "Cell output area" option is great for this.  But `show_code` appears to only show the code below the output.  Particularly for app view, I think it will be much clearer to users of my app if they see the code, and then the output

I think this function should

  1. Have an option to specify whether code should come first or output.
  2. If the option is not present, go with whatever is in the settings.

### Suggested solution

The first item seems easy enough; it would just be an option to the function that would reverse the order in [this statement](https://github.com/marimo-team/marimo/blob/8d176ca02baaca57aea212edd36b5d2fe5c57af9/marimo/_output/show_code.py#L77-L80).  But I don't understand the settings system enough to figure out item 2.

mscolnick:
Closed by #4039

PR #4039, opened 2025-03-09
Adding a boolean to see the code first, then result in marimo.show_code

xeniumcode:
### 📝 Summary

Added a ``code_first`` boolean to ``marimo.show_code``  to control the order of code and results (default: False, maintaining previous behavior). A new setting for this, as requested in part 2 of the issue, I think is unnecessary.

Related Issue: #3997 

### 📋 Checklist

@mscolnick

### DIFF
--- a/marimo/_output/show_code.py
+++ b/marimo/_output/show_code.py
@@ -16,7 +16,10 @@ def substitute_show_code_with_arg(code: str) -> str:
     modified_code = re.sub(pattern, r"\1", code, flags=re.DOTALL).strip()
     # Remove code_first=True or code_first=False from the end
     modified_code = re.sub(
-        r",\s*code_first\s*=\s*(True|False)$", "", modified_code
+        r",?\s*code_first\s*=\s*(True|False),?\s*\)?$",
+        "",
+        modified_code,
+        flags=re.DOTALL,
     ).strip()
     return modified_code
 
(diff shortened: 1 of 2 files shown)


marimo/_output/show_code.py
```diff
@@ -73,11 +73,19 @@ def factorial(n: int) -> int:
     code = substitute_show_code_with_arg(cell.code)
```
mscolnick, line 73: this `substitute_show_code_with_arg` likely doesn't handle the new argument and will lead an incorrect result. we should adapt `substitute_show_code_with_arg` and write lots of tests (Claude should be pretty good at generate test cases.

tests/_output/test_show_code.py
```diff
@@ -80,3 +80,139 @@ def test_substitute_show_code() -> None:
...
+        == "mo.show_code(1, code_first=False)"
+    )
+
+    code = "mo.show_code(mo.show_code(foo(3) + 1, code_first=False), code_first=True)"
+    assert (
+        substitute_show_code_with_arg(code)
+        == "mo.show_code(foo(3) + 1, code_first=False)"
+    )
+
+    code = "mo.show_code(mo.show_code(42, code_first=True), code_first=False)"
```
mscolnick, line 214: can you try multi-line and with a comma:

```
mo.show_code(
	42,
	code_first=True,
)
```

and


```
mo.show_code(
	42,
	code_first=True
)
```

mscolnick: i guess we also won't support this case?
```
mo.show_code(
	42,
	code_first=mo.app_meta().mode == "app",
)
```

mscolnick (approved):
looks great! going to merge and change it to 'above' and 'below', just in case we eventually want left/right

moble:
@mscolnick Thanks for handling this.

> A new setting for this, as requested in part 2 of the issue, I think is unnecessary.

I just want to clarify that I didn't mean to suggest a new setting; I just meant that the default should be to fallback to the current setting in User settings > Display > Outputs > Cell output area.

Anyway, this will work for me.  Thanks again!

PR #4043, opened 2025-03-09
improvement: add 'auto' as an option for auto_discovery

mscolnick:
This adds 'auto' as an option for `auto_discovery`, which is nice for things like `sqlite` so these can be discovered while Snowflake or Bigquery don't need to be.

Should we also do this for `columns`?

cc @Light2Dark 

Cons:
* this could add confusion with the double `auto`

### DIFF
--- a/docs/guides/working_with_data/sql.md
+++ b/docs/guides/working_with_data/sql.md
@@ -189,13 +189,13 @@ marimo will automatically discover the database connection and display the datab
 
 ???+ note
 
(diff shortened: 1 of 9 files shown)


Light2Dark:
Nice!

Is the value of this option mainly for LLM and code completion? Because the latency to discover tables for sqlite, postgres is pretty low.

But yeah, we should make auto discovery smarter for the future, eg. check number of schemas, tables b4 discovering. Async discovery. All of these optimizations fit under "auto".

Light2Dark (approved):
let me know if you'd like me to test. but lgtm

Issue #4077, opened 2025-03-12
Polars UTC timestamps shown converted to local timezone


adamreeve:
### Describe the bug

When displaying a Polars DataFrame containing a UTC DateTime column, Marimo displays column values converted to the local time zone, which is quite misleading as the column type is shown as `datetime[μs, UTC]`. 

![Image](https://github.com/user-attachments/assets/f4e7be8b-870c-4efc-9ffa-45db9224006c)

Hovering over the value shows it in both UTC and the local timezone, but I would expect the main display to be in UTC to match the column type.

This seems to be the opposite problem to https://github.com/marimo-team/marimo/issues/2871.

### Environment

<details>

```
{
  "marimo": "0.11.18",
  "OS": "Linux",
  "OS Version": "6.12.11-100.fc40.x86_64",
  "Processor": "",
  "Python Version": "3.12.5",
  "Binaries": {
    "Browser": "132.0.6834.159",
    "Node": "v16.3.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.30.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pymdown-extensions": "10.14.3",
    "pyyaml": "6.0.2",
    "ruff": "0.9.10",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.12.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "polars": "1.24.0"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
from datetime import datetime
import pytz
import polars as pl

pl.DataFrame({
'timestamp': [datetime(2010, 10, 7, 13, 15, tzinfo=pytz.UTC)]
})
```

mscolnick:
Thanks for filing, I have a fix to use the existing timezone from polars instead of always local https://github.com/marimo-team/marimo/pull/4078

adamreeve:
Thanks for the quick fix!

PR #4078, opened 2025-03-12
fix: render polars dates in UTC when a timezone is specified

mscolnick:
Fix for #4077

Screenshot 2025-03-12 at 8 06 01 PM

### DIFF
--- a/frontend/src/components/data-table/types.ts
+++ b/frontend/src/components/data-table/types.ts
@@ -38,3 +38,11 @@ export type DataTableSelection =
   | "single-cell"
   | "multi-cell"
   | null;
+
+export function shouldRenderDateInUTC(dtype: string | undefined): boolean {
+  if (!dtype) {
+    return false;
+  }
+  // Check for datetime[X,Y] format
+  return /^datetime\[[^,]+,[^,]+]$/.test(dtype);
+}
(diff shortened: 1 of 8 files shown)


frontend/src/components/data-table/types.ts
```diff
@@ -38,3 +38,11 @@ export type DataTableSelection =
   | "single-cell"
   | "multi-cell"
   | null;
+
+export function shouldRenderDateInUTC(dtype: string | undefined): boolean {
+  if (!dtype) {
+    return false;
+  }
+  // Check for datetime[X,Y] format
+  return /^datetime\[[^,]+,[^,]+]$/.test(dtype);
```
akshayka, line 47: Should one of X or Y be "utc"?

akshayka (approved):
lgtm, one question

frontend/src/components/data-table/types.ts
mscolnick: just posted a screenshot in in the description. whenever there is a timezone (not just UTC), polars at least renders as UTC.

mscolnick: actually this might not be true. polars just converts it to UTC. let me see if I can get the column in polars in another tz

mscolnick: yea i was wrong. i need to extract the timezone

akshayka (approved):

PR #4080, opened 2025-03-13
improvement: better prompt-to-install when missing deps, grouped installed, Optional Features panel

### LINKED ISSUE #3113: Easier installation of optional dependencies for the marimo editor
### Description

Easier installation of optional dependencies that enable: 
- SQL support (duckdb)
- AI features (openai/anthropic)
- Charting (altair)
- Charting w/ performance (vega-fusion)

### Suggested solution

We can re-use the Package installation modal that happens when there are imports that fail. We will also need to make sure the UI is updated appropriately (`capabilities` is refreshed).

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
Fixes #3113 

Screenshot 2025-03-12 at 9 30 17 PM

* Remove 'SQL' from KernelCapabilities. If the user tries to run SQL, we will not prompt to install the necessary deps.
* We can now prompt the user for multiple missing deps
* Update `ai` endpoints to throw missing modules so that we can prompt the user as well.
* `Optional Dependencies` panel in the settings menu to show the enabled features and install. This is both for education and ease of installation.

I didn't want to put everything in `KernelCapabilities` since its another source of truth, and requires checking it on startup, which could be a waste of resources/blocking

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -1313,43 +1313,50 @@ class Kernel:
             for e in runner.exceptions.values()
             if isinstance(e, (ModuleNotFoundError, ManyModulesNotFoundError))
         ]
-        if (
-            len(module_not_found_errors) > 0
-            and self.package_manager is not None
-        ):
-            missing_modules: set[str] = set()
-            for e in module_not_found_errors:
-                if isinstance(e, ManyModulesNotFoundError):
-                    missing_modules.update(e.package_names)
-                elif e.name is not None:
-                    missing_modules.add(e.name)
 
-            # Grab missing modules from module registry and from module not found errors
-            missing_modules = (
-                self.module_registry.missing_modules() | missing_modules
+        if len(module_not_found_errors) == 0:
+            return
+
+        if self.package_manager is None:
+            return
+
+        missing_modules: set[str] = set()
+        missing_packages: set[str] = set()
+
+        # Populate missing_modules and missing_packages
+        # from the errors
+        for e in module_not_found_errors:
+            if isinstance(e, ManyModulesNotFoundError):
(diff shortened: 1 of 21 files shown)


marimo/_runtime/runtime.py
```diff
@@ -1310,15 +1311,22 @@ def _broadcast_missing_packages(self, runner: cell_runner.Runner) -> None:
...
+            if isinstance(e, (ModuleNotFoundError, ManyModulesNotFoundError))
         ]
         if (
             len(module_not_found_errors) > 0
             and self.package_manager is not None
         ):
+            missing_modules: set[str] = set()
+            for e in module_not_found_errors:
+                if isinstance(e, ManyModulesNotFoundError):
+                    missing_modules.update(e.package_names)
```
akshayka, line 1323: Should this be adding module names, not package names?

akshayka (commented):
Just one comment/question

marimo/_runtime/runtime.py
mscolnick: The module names can be lossy, so I need to add package names. We convert it right below though

akshayka: just double checking — any issues with the code below? Will adding package names to `missing_modules` make `module_to_package` unhappy?

image

Like should this be

```python
missing_packages: list[str] = []
missing_modules: set[str] = set()
for e in module_not_found_errors:
    if isinstance(e, ManyModulesNotFoundError):
        missing_packages.extend([p for p in e.package_names if not self.package_manager.attempted_to_install(p)
```

mscolnick: ah i see now, cleaning it up

PR #4090, opened 2025-03-13
improve: allow header comments in SQL cells.

akshayka:
Update the SQL language adapter to accept cells that start with hash comments. These comments are preserved on transform back to Python.

While people don't typically write the Python for SQL cells directly, AIs do; this helps accommodate verbose AIs.


mscolnick (dismissed):
nice!

PR #4092, opened 2025-03-13
improvement: open external document on go-to-definction when LSP is enabled

mscolnick:
Add go-to-definition for external files, upgrade our `@marimo-team/codemirror-languageserver`


akshayka (approved):
awesome!

PR #4097, opened 2025-03-13
[wip] feat: add CLI args to control UI elements when running notebook as script

eckp:
## 📝 Summary

WIP: Add parameter to ui element instantiation, to set default value from command line interface argument.

To give control over ui element values when running notebook as script, solving [this question](https://community.marimo.io/whats-the-expected-behaviour-when-running-notebooks-that-contain-ui-elements-as-scripts-mzv0CqpnmKVv).

## 🔍 Description of Changes

1. Expose `mo.raw_cli_args()` similar to mo.cli_args(), but keep them as `list of str` instead of pre-parsing them into a `dict` to preserve lists as values ([circumventing `parse_args` limitations](https://github.com/marimo-team/marimo/blob/ca6460bf41510b446bf5d2ab482d04fcdd4299c2/marimo/_cli/parse_args.py#L19)).
3. Create a global `argparse.ArgumentParser` object in `mo.ui._impl.input.py`, to which each ui element can add its argument if `cli_name` is given.
4. If the argument has a value, update the default ui element value.

## 📋 Checklist

    - **Applicable to `mo.raw_cli_args()`?**

## 📜 Reviewers

@mscolnick


eckp:

test.py
```diff
@@ -0,0 +1,43 @@
...
+def _():
+    import marimo as mo
+
+    mo.raw_cli_args()
+    return (mo,)
+
+
+@app.cell
+def _(mo):
+    b = mo.ui.number(cli_name="test")
```
mscolnick, line 17: we've actually talked about adding `key`s to UI elements, as it will have other use cases. 

another option that you can do today, why not initialize this with `mo.ui.number(value=mo.cli_args().get("test"))`?

eckp:
I like `key`, will use that. 
Which other use cases does it have? Didn't find any issue or discussion regarding this yet.

The current `cli_args()` parsing does not support lists as values, which I would need to support cli-defaults for `mo.ui.range_slider`. 

Hence my workaround was to have an unparsed version `raw_cli_args()`. However, that does not need to be exposed all the way, could keep it private to `runtime` for example.

eckp:
Example file `test.py`:
```python
import mo
print(mo.cli_args())
print(mo.cli_args().get_all('range_slider'))
```
yields
```shell
$ python test.py --range_slider 5 6 -v
{'range_slider': '5 6', 'v': ''}
['5 6']
```
so `cli_args()` does not support lists as values ([as indicated here](https://github.com/marimo-team/marimo/blob/ca6460bf41510b446bf5d2ab482d04fcdd4299c2/marimo/_cli/parse_args.py#L19)).

While we can parse the string into two separate ints before setting them as default value on the user side:
```python
print(raw_value := mo.cli_args().get_all("range_slider")[0])
print(parsed_value := [int(v) for v in raw_value.split(" ")])
mo.ui.range_slider(0, 10, value=parsed_value)
``` 
```shell
$ python test.py --range_slider 5 6 -v
5 6
[5, 6]
``` 
I thought it would be nice to encapsulate this logic directly inside the UI element, and rely on argparse to return the list of values. 

Additionally, we could use argparse to generate a help message for the cli from the UI element's label, limits, etc.

If this sounds useful, I'll proceed with expanding this to other UI elements and adding tests.

mscolnick:
You need to use the arg twice to the get the list values.

eckp:
> You need to use the arg twice to the get the list values.

Thanks, that works!

I still see value in integrating a cli option in the ui in- and output elements themselves, as it fits the idea of deploying marimo notebooks as a pure python script in a production pipeline, without any boilerplate code on user side.

Do you see a way forward for this idea? If so, feedback on the current draft would be very welcome.
Also, I'd be curious about the other uses of the `key` arg for compatibility considerations.

mscolnick:
@eckp, by doing it in user code, the user can validate or transform the CLI inputs. also it's readable by the end-user what is going on. 

i don't there is us much more boiler plate:

```python
mo.ui.text(key="input")
# vs
mo.ui.text(value=mo.cli_args().get("input"))
```

--------------

Another idea for `key` i thought of was a way to keep values in sync

```
mo.ui.number(key="age")
mo.ui.slider(key="age")
```

PR #4038, opened 2025-03-08
test: xfail flakey test

mscolnick:

PR #4111, opened 2025-03-14
fix: Enable coverage of _save tests

dmadisetti:
@mscolnick re https://github.com/marimo-team/marimo/pull/4038 tests only seem to be failing in coverage. Seems to be because `with_block` does funny things with sys.trace which coverage uses. This mitigates it a bit- but seems to be a bit strange with just the first test which seems to have a messed up code pointer (which may well be a python bug? Seems to always fail on whatever the first test is)

### DIFF
--- a/marimo/_save/save.py
+++ b/marimo/_save/save.py
@@ -295,7 +295,8 @@ class _cache_context(SkipContext):
                     if lineno >= self._body_start:
                         self.skip()
                 return
-            elif i > 2:
+            # <module> -> _trace_wrapper -> _trace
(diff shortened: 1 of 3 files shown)


mscolnick (dismissed):

mscolnick:
wow nice find

mscolnick (approved):

dmadisetti:
Ehh, might revisit at some point, because doesn't catch the coverage it should: https://app.codecov.io/github/marimo-team/marimo/tree/main/marimo%2F_save

PR #4019, opened 2025-03-07
chore: Add support for narwhals `Time` dtype

FBruzzesi:
## 📝 Summary

[Narwhals#2113](https://github.com/narwhals-dev/narwhals/pull/2113) is going to introduce `Time` dtype, this PR prepares for that

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [ ] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

## 📜 Reviewers

@akshayka OR @mscolnick

PR #4122, opened 2025-03-16
improvement: json & list view in datatables

### LINKED ISSUE #4119: improve json & array view in datatables
### Description

We could improve the view for json and arrays in Tables and Dataframes.

Image

### Suggested solution

I have some working code to use `JsonOutput` to display it nicely, I modified the backend to identify dtypes `list` and `object` (right now is `unknown`).

Image

### Alternative

_No response_

### Additional context

Nested lists are quite tricky. From a brief research, csv's aren't ideal for nested data, it's recommended to use parquet/json/other format instead. https://github.com/pola-rs/polars/issues/6038

Light2Dark:
## 📝 Summary

Fixes #4119. Saves dataframe as json instead of csv to support more dtypes during serialization. Polars upgraded to support more types.
image

## 🔍 Description of Changes

- Multi-index rows are not supported yet, can do in follow up. A warning is added to suggest that the user can reset_index()
- Clamped polars version due to a regression in latest: https://github.com/pola-rs/polars/issues/21799
- I still haven't fixed rendering buttons nicely.
image

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


mscolnick (commented):
this is awesome, great improvement. 

There are likely a couple issues that existed in the csv that required a couple edge case handling that we may need to handle for json as well

- BigInt (when we serialize, we can serialize as strings instead of numbers)
- nested structs in polars to_csv and to_json
- non serialize-able types

This probably removes some other hacks too that we had with csv's so that's great

Light2Dark:
polars.exceptions.InvalidOperationError: `quantile` operation not supported for dtype `time`

tests failing likely related to this 🤔: https://github.com/marimo-team/marimo/pull/4019

Light2Dark:
hey @mscolnick, wanted to get your thoughts. `orient=records` causes index information to be lost. We could:
1. reset_index() and do some hacks to get that information
2. use another orient, but will need to update the frontend vega-loader parsing.

mscolnick:
@Light2Dark `drop_index` sounds like it would work, but we might want to ignore the basic range index

we have some logic in Pandas: 
`get_row_headers` to check if there are any custom headers
`has_headers = len(self.get_row_headers()) > 0`

and then if so, you can `drop_index()` write before writing to json

Light2Dark:
`mo.ui.table` doesn't break it
image

PR #3794, opened 2025-02-14
feat: allow marimo configuration in notebook.py as script metadata (per PEP 723)

mscolnick:
Fixes #2885
Fixes #2774

This adds inline script metadata (PEP 723) support for marimo configuration, similar to our `pyproject.toml` config.

e.g. can add:

```
# /// script
# [tool.marimo.runtime]
# auto_instantiate = false
# on_cell_change = "lazy"
# ///
```

to the top of the script to add per-notebook configuration. the precedence is

**Script config > pyproject.toml config > user config**

Issue #3054, opened 2024-12-04
Option to configure persistent caching for all cells within app or directory


leventov:
### Description

When the notebook is used locally always saving all results to disk is cheap. Therefore it makes sense to alleviate the cognitive burden for developers (assessing whether the cell is "expensive" or not) and source code clutter (that is, extra `@persistent_cache` decorators) by providing an option configurable per-app or per-directory (such as in a `.marimo/config` yaml/config file) to automatically save and invalidate the results of all cells whenever they are changed.

### Suggested solution

As described above.

### Alternative

An App` subclass, such as `PersistedApp`? But the problem with this is that it still has decision burden (per app rather than per cell, but still), and it cannot be dropped transparently when the app is deployed on the web.

dmadisetti:
Yes! I have a working branch that does this. There are a few bugs, and will probably go into experimental `.toml` setting for now.

Can you expound more on what you mean by having persistent cache for a directory?

gabrielgrant:
This would be great, so long as it is paired with:

1. Some indication of whether results are being computed or coming from cache
2. An easy way to clear/ignore the cache and recompute

PR #4133, opened 2025-03-17
Persistence subsystem refactoring

leventov:
## 📝 Summary

Implements the first stage of persistence subsystem refactoring, as discussed at length in #3176.

## 🔍 Description of Changes

This PR only has textual detailed design and implementation plan yet. This is the first version of this design and plan that looks consistent to me (doesn't have internal contradictions) and is complete (sufficient for implementation): no high-level hand-weaving and descriptions, just specific implementation details and algorithms.

The hardest part was to get inter-process locking and "new snapshot creation" + "entry writing" machinery correct, especially in the presence of *non-cooperative* synchronization systems like Dropbox. I iterated through a lot of designs internally and was founding race conditions with them until I arrived at the one described in the document. I don't claim it's optimal (likely not), but at least I'm relatively confident that it's concurrency bug-free.

I'll add the actual code implementation of this detailed plan soon (I already starting to write it with Claude Code), and I expect it to take an order of magnitude less time than it took to create the plan itself (more than 1 month), unless new concurrency or other problems are be found in the process.

I suggest reviewers (@mscolnick and @dmadisetti) to also reason and review on the level of the text description because the code is going to be even bigger in volume than this description.

## 📋 Checklist

## 📜 Reviewers

@mscolnick @dmadisetti

### DIFF
--- a/dev/auto_cleanup_detailed_plan.md
+++ b/dev/auto_cleanup_detailed_plan.md
@@ -1,0 +1,1127 @@
+# Detailed Implementation Plan: Marimo's Persistence Subsystem Refactoring
+
+## 1. Introduction and Goals
+
+### 1.1. Overview
+
+This document includes the detailed design and the implementation plan for refactoring of Marimo's persistence subsystem. **The goals of this refactoring are to enable (semi-)automatic cleanup and provide a more robust and flexible foundation different use cases: both *persistent cache* (already) and  *persistent cell execution* (in the future).**
+
+This plan builds upon and synthesizes discussions from [#3176](https://github.com/marimo-team/marimo/issues/3176) and related issues.
+
+The key improvements include:
+
+*   **Content-Addressed Storage:**  Both entries and their contents will be stored using content-addressed filenames to ensure data integrity (and, secondarily, to enable deduplication), similarly to [Git's object database](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects), [git-annex](https://git-annex.branchable.com/internals/key_format/), and many backup systems.
+*   **Separation between _entries_ and _contents_:** Multiple entries can reference the same contents (deduplication). In the initial implementation, both entires and contents are stored in the local file-system, in separate folders. This permits using different mechanisms for their synching/sharing and versioning, such as Git (for entries) and Dropbox (for contents).
+    *   In the future, this separation will also enable **configuring different storage backends for entries and contents**, such as storing the entries in a server database such as PostgreSQL, and contents in object storage such as Amazon S3. This is inspired by [`jj`'s storage-independent backend APIs](https://jj-vcs.github.io/jj/latest/technical/architecture/#storage-independent-apis).
+*   **Cleanup:**  A configurable cleanup mechanism will manage disk space usage, removing outdated or unnecessary persisted entries and contents.
+*   **`Persister` Abstraction:**  A new `Persister` abstraction will encapsulate the storage logic, providing a clean interface for interacting with the persistent cache "facade" (Marimo's use-case; another possible future "facade" for the persistence subsystem would be persistent cell execution) and allowing for future extensions (e.g., SQLite entry storage, remote entry or content storage).
+*   **`Entry` Data Structure:**  The existing `Cache` class will be replaced with a new `Entry` class, representing the contents of a `mo.persistent_cache()` block (or, in the future, a persisted cell) execution at a point in time. What comprises block's contents is discussed below.
+*   **Legacy Support:**  Backward compatibility with the existing `PickleLoader` and `JsonLoader` will be maintained through a `LegacyPersister` implementation.
+*   **Improved Concurrency:** The design incorporates lessons from [`jj`'s concurrency](https://jj-vcs.github.io/jj/latest/technical/concurrency/) to ensure that the persistence system is robust and safe for concurrent modifications even when there are non-cooperative Git checkout operations, Dropbox or other synchronization systems, the file system in which the content files are stored is network-attached and therefore unreliable, etc.
+
+This plan prioritizes incremental implementation, starting with `FsPersister` that uses local file system for storing both entries and contents, stores entries in [TOML](https://toml.io/en/) files, and all content objects as separate files on the file system.
+
+More advanced features, such as entry or content versioning, storing entries and content metadata in SQLite, "consolidating" lots of small content objects in a single file (a la Git's *object packs*), remote database-based `Persister`, and advanced content deduplication are deferred to future iterations.
+
+### 1.2. Scope
+
+This implementation plan focuses on the following core components of the persistence subsystem refactoring (currently used only for the **persistent cache** use-case):
+
+*   **`Persister` Abstraction:** The core `Persister` abstract class is used instead of the legacy `Loader` when in *persistent caching* logic of the Marimo core. However, `MemoryLoader` is **not** replaced for *in-memory caching* (`mo.cache()`), although all loaders are changed to use `Entry` abstraction instead of `Cache`.
+*   **`Entry` Abstraction:** The `Entry` class is defined and integrated into the Marimo core as a replacement of `Cache`.
+*   **Filesystem-Based Storage via `FsPersister`:**  A concrete implementation of `Persister` that uses local filesystem-based storage for both entries and content. `FsPersister` is designed to be highly configurable in the future, but currently implements only a few concrete choices regaring entry and content storage.
+*   **Entries stored in TOML:**  In the initial version of `FsPersister`, entries are stored as plain text, in TOML files.
+*   **Cleanup Mechanisms:**  Cleanup mechanisms are implemented, allowing for the removal of entries (and, later, the contents referenced by these entries) based on a user-provided *cleanup strategy* function.
+*   **Backward Compatibility via `LegacyPersister`:** `LegacyPersister` provides support for *existing* persistent cache "deployments" in `__marimo__/cache/` directories.
+*   **API Changes:**  The necessary changes to the `App()` constructor permit graceful upgrade to the new persistence subsystem. If the `__marimo__/cache/` directory is present, the `App` will use `LegacyPersister` unless another implementation of `Persister` is explicitly passed to the `App()` constructor. If `__marimo__/cache/` is *not* present, `App` uses the new `FsPersister` that in turn creates and uses a different directory: `__marimo__/persist/`.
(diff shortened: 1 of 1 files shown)


dmadisetti:
Thanks @leventov

It's going to take me a second to properly parse through all 16k words, but I'll try to sometime later today.
If you have a guide to reading that would be appreciated.

---

After a quick read just some initial thoughts (feel free to chime in if it seems like I misunderstood something):

## Persister
I can put in the "Persister" (I think our previous discussions I called it an Adapter, but maybe we can compromise somewhere. Store? Storage? Interface?). I've been eager to rig this up with Redis, and also try directly leverage Nix's caching by manually creating `.nar` files.

Just a note, Loader vs BasePersistenceLoader has already been decoupled to some extent, but to disambiguate, loader might be better named `Converter` (though I will keep it loader for now)

```python
@dataclass
class EntryKey:
    execution_hash: str
    hash: str
    cache_type: CacheType
    path: Optional[Path]
# Created is particular to the proposed "Persister/Adapter method"


class Store:
    @abstractmethod
    def get(
        self,
        entry: EntryKey, # feel free to build execution key from this
        **kwargs: Any, # Whatever other args might be needed
    ) -> Optional[bytes]:
        """
        Retrieve an entry from the persistence subsystem.
        """

    @abstractmethod
    def put(
        self,
        cache: Cache,
        BaseConverter: converter
        **kwargs: Any,
    ) -> None:
        """ 
        Store EntryMetadata on self if needed
        """
        blob: bytes = converter.to_blob(cache)
        # Do whatever with the blob, potentially switching on loader.
        ...

    @abstractmethod
    def post_execution_hook(
        self,
        cell: CellImpl,
        runner: cell_runner.Runner,                                                                                                                                                                                                                                             
        run_result: cell_runner.RunResult,      
        ) -> None: ...
        # Do something like cleanup here


    @abstractmethod
    def __del__(self) -> None:
        """
        __del__ over on_exit
        Suggested, since should be singleton- only deleted on shutdown.
        Motivation: no current, easy, shutdown hook.
        """
        ...
```

## Cache
I've not sure what's gained renaming Cache -> Entry. Also, in your description, this removes a fair amount of effort and care required for managing State and UI objects, in the original Cache Spec:

```python
class Cache:
    defs: dict[Name, Any]
    hash: str
    stateful_refs: set[str]
    cache_type: CacheType
    hit: bool
```

What I will do is expose 

```python
execution_hash : str
```

On the cache object (Which is computed prior to the content hash variables and should be a sub for your base seed for your executionkey, which you might want to bake the file-contents/ path into).

### App
> The default behaviour, where the object_id_mode is not explicitly passed to FsPersister (or, FsPersister is not explicitly passed to App()) can be controlled via [.marimo.toml](https://docs.marimo.io/guides/configuration/#user-configuration) (so-called user configuration). So, an option like fs_persister.default_object_id_mode can be added to to .marimo.toml. Extending the SQLite metaphour, this would analogous to SQLite's [compile-time options](https://www.sqlite.org/compile.html).

I don't think this requires a serialization changes so much as just a user (`./config/marimo.toml`) change.
This can still be controlled on a per notebook level with [PEP 723](https://peps.python.org/pep-0723/) (see #3794)

leventov:
> I can put in the "Persister" (I think our previous discussions I called it an Adapter, but maybe we can compromise somewhere. Store? Storage? Interface?). I've been eager to rig this up with Redis, and also try directly leverage Nix's caching by manually creating .nar files.

`Store` sounds fine to me, with `FsPersistentStore` and `LegacyPersistentStore` as the names of the two impls that were called `FsPersister` and `LegacyPersister` in the doc.

```python
@dataclass
class EntryKey:
    execution_hash: str
    hash: str
    cache_type: CacheType
    path: Optional[Path]
```
Execution hash - is this the equivalent for `block_id` (block hash) in the doc? It seems ambiguous as the field name because as you pointed [here](https://github.com/marimo-team/marimo/issues/3176#issuecomment-2619384612) + see my next comment in that thread, this should be the hash of the block code *only*, not a tree hash. Also, for `mo.persistent_cache(name="foo")` this name should be the `block_id` (also, the names of the named cells, when #3054 is implemented).

`created_at` is also important.

If these changes are OK for you, I'll add `EntryKey` in the code that I'm writing at the moment. Or you intended to add it on your own, before my PR?

`path: Optional[Path]` -- is this coming from `save_path` argument of `mo.persistent_cache()`?

```python
    @abstractmethod
    def put(
        self,
        cache: Cache,
        BaseConverter: converter
        **kwargs: Any,
    ) -> None:
```
As part of my wanderings in the design space (software abstraction and interface space, in this case), I considered providing an argument like converter (object) to `[Persister|Store].get()` and `put()`. I then stripped that because conversion logic is going to be too intertwined with `FsPersistentStore.put()` logic: it will create a custom `Pickler` subclass with overridden `Pickler.persistent_id()` whose logic depends on the object_id_mode... A nice and clean converter abstraction that would work equally well for legacy store and new store will not survive. Luckily I don't think it's important because we aren't going to have many of them (or *any* beyond pickle and json) and I don't think users will be eager to add their own. That's why in the design that I proposed, conversion/serialization format is specified via `ContentSpec = Mapping[ContentKind, ContentFormat]`, where `ContentFormat = Literal["pickle", "json", "auto"]`, so, conversion format is specified as `"pickle"` rather than a `converter` object.

> post_execution_hook

Not sure what you mean by this. Why the logic should be different from the proposed in the doc? In the current proposal, cleanup *may* be triggered right from `put()`. Is your concern here that it might make `put()` not "real-time" so even if cleanup trigger conditions are ripe, you would like the imminent cleanup to be delayed until after `post_execution_hook()` and be executed in background? I think it's possible, I can do this, but want to clarify that it's what you meant.

> def __del__(self)

https://stackoverflow.com/a/1481512/ argues against del, any objections to using https://docs.python.org/3/library/atexit.html?

> I've not sure what's gained renaming Cache -> Entry. Also, in your description, this removes a fair amount of effort and care required for managing State and UI objects, in the original Cache Spec:

They main reason why I want to rename is because "Cache" is just a confusing name for this abstraction: "cache" is universally understood as the store as a whole, not a single entry within it. Also, it's a misnomer because "cache" is something ephemeral (even if on-disk), while `FsPersistentStore` can be used not *just* as "ephemeral cache" but also as "persistent execution". This itself may delude LLM coding agents (though less so humans), but what will definitely confuse them more is conceptual "clash" is there is a mash-up of "entries" and "caches". LLM coding agents are very sensitive to such inconsistencies. Or else I should make the naming thoughout the refactored persistence subsystem completely consistent with "cache" (including `cache_log/`, `cache_snapshots/`, `CacheMetadata` class, etc.). Don't you think it will be way too confusing and misnomer-y for a *persistence* subsystem?

Having said that, we cannot actually remove nor even rename `Cache` class itself because the instances of this class are pickled by `PickleLoader` and we need at least to be able to keep unpickling those already stored in `__marimo__/cache`, or perhaps even keep pickling them as `Cache` objects after the refactoring so that `__marimo__/cache` doesn't store a mash-up of pickled `Cache` and `Entry` objects. So, `Cache` will survive, but I intend to keep it to *internals* of `Loader` implementations only (`PickleLoader` and perhaps `JsonLoader`), and strip it as a concept and name from anywhere else.

> in your description, this removes a fair amount of effort and care required for managing State and UI objects

If my description does this it's my ignorance and I should change the description. I didn't intend to affect the existing functionality. I thought I already did this with putting `CacheType` in `Entry.extra` and storing `stateful_refs` in `MemoryLoader` along with `Entry` objects. I will try to understand why it's not enough, specific pointers appreciated.

> I don't think this requires a serialization changes so much as just a user (./config/marimo.toml) change.
This can still be controlled on a per notebook level with [PEP 723](https://peps.python.org/pep-0723/) (see https://github.com/marimo-team/marimo/pull/3794)

All `[fs_persister.defaults]` and `[fs_persister.default_cleanup_strategy]` configs exactly mirror the constructor arguments to `FsPersistentStore`. So, on the notebook level, the "config" is just importing and constructing `FsPersistentStore` explicitly with needed arguments and then passing it into. `[fs_persister]` is meant to be an additional layer of "global defaults" that apply to all notebooks (apps) withing the scope of that `.marimo.toml`, to alleviate the need to repeat configurations in every notebook. Does this sound right or you would change this?

dmadisetti:
I got Store working with Redis yesterday, and was working on a remote Nix cache and tests, but I'll push what I have up- I think having something concrete to review will go a long way.

---

> In the current proposal, cleanup may be triggered right from put()

I'm not sure it should be abstract level the

> atexit

Maybe, we'd have to look over marimo teardown. It's not system exit, it's runtime when this would need to trigger 

> configs exactly mirror the constructor arguments to FsPersistentStore

You'll see the pattern I use for Redis, this isn't a problem with just config. I think it makes sense to not change serialization much. Top level function changes will potentially make notebooks more dangerous, and considering a push to statically load apps for edit mode.

dmadisetti:
@leventov do you think you could build on #4147 (after  merge) in a way that doesn't disrupt current functionality? Ideally starting with a proof of concept, that can be expanded with successive PRs?

I'm still not 100% clear on your implementation plans, and your doc doesn't relate on _why_ it's an improvement (even if it's self evident to you, what are the limitations of the current mechanism, and how does your proposal address those concerns. Cleanup and concurrency are fair, but solutions like a Redis with LRU turn on already address those). I've quickly read through your doc several times, and it needs a bit more organization and concision

Happy to put work where I think there is synergy (like #4147)

leventov:
@dmadisetti I already rebased the code on top of the `dm/cache-store` branch, although it doesn't "integrate" yet because there are interface mismatches, described [here](https://github.com/marimo-team/marimo/pull/4147#issuecomment-2733951219). [`App.save_key`](https://github.com/marimo-team/marimo/pull/4147#discussion_r2001349303) is another important outstanding thing.

> Cleanup and concurrency are fair, but solutions like a Redis with LRU turn on already address those

Focusing on Redis contradicts the desideratum in the first message of #3176:

> Agnostic towards the way cache is shared between machines and/or backed up: Git/Github, git-annex, CI systems cache directories (such as Github Actions's [cache action](https://github.com/actions/cache)), Marimo Cloud, Dropbox, rsync, backup systems, etc.

Redis in specific is not durable (and if there is a durable mode, I don't trust it: I don't trust Redis in general, I avoid it whenever possible), it's an extra stateful runtime that should be kept around where it seems many serverless things can work (or piggy-backing other server things that are already used by the user, such as Git server).

Another big point is **deduplication**. Entry/object separation is needed to enable deduplication basically. For *persistent execution* use case, deduplication should be a big deal because it's expected that there will be lots of cells with differing inputs or codes (and hence different `hash`) but the output stays the same.

Entries (execution keys more specifically) are also designed to be extensible in the future to accommodate *versions*, I discuss them in #3176.

What I'm implementing is not a "new" design, it's exactly the design that you agreed with (seemingly) in the end of #3176 discussion, already with concessions from my perspective. The doc committed in this PR is just an elaboration of that design because what seemed "simple" on the hand-weavy level and could be described in a couple of sentences turned out to require a lot of careful adjustment of certain details about locking, operation sequencing, formats, etc.

PR #4147, opened 2025-03-18
feat: configurable marimo store cache

dmadisetti:
Introduce `stores` to give @leventov surface area to implement their idea.
However, also puts in a proof of concept `RedisStore`, and some relevant tests

Should take a follow up to properly expose RedisStore, an some other stores


dmadisetti (commented):
@leventov

I see your code as just being another "Store", I think with this it doesn't need to hook in any deeper- but let me know if there's something I'm missing

marimo/_save/stores/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from marimo._save.stores.nix import NixRemoteStore
+from marimo._save.stores.redis import RedisStore
+from marimo._save.stores.store import Store, StoreType
+
+StoreKey = Literal["file", "redis"]
+
+CACHE_STORES: dict[StoreKey, StoreType] = {
+    "file": FileStore,
+    "redis": RedisStore,
+    "nix": NixRemoteStore,
```
mscolnick, line 14: how do you pass config to this?

marimo/_save/stores/store.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Singleton
+    def __new__(cls) -> Store:
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+            # Put any initialization here.
+        return cls._instance
+
+    @abstractmethod
+    def get(self, key: HashKey, loader: Loader) -> bytes:
```
mscolnick, line 27: i wonder if `get` should return `Optional[bytes]`. so you don't need to check `hit` before which could be two networks requests instead of one. although hit is likely cheap, it's still a waterfall.

`hit` should still exist for the case where you only need to check existence

marimo/_save/stores/__init__.py
dmadisetti:
```toml
[experimental.cache]
store = "redis"
```

No error handling around failures for right now

tests/_save/loaders/test_json_loader.py
```diff
@@ -64,142 +65,146 @@ def test_cache_hit_miss(self) -> None:
...
-
-        assert loaded_json2["hash"] == "hash2"
-        assert loaded_json2["cache_type"] == "Deferred"
+        assert not loader.cache_hit(key("empty", "Pure"))
+
+
+# These are reasoanble, granular unit tests
+# But should be covered by the general testing
+#
+#    def test_load_persistent_cache(self) -> None:
```
mscolnick, line 86: if these are covered and currently a maintenance burden because of the changes, feel free to remove them

tests/_save/loaders/test_json_loader.py
```diff
@@ -64,142 +65,146 @@ def test_cache_hit_miss(self) -> None:
...
-        # Load the second JSON and verify contents
-        with open(cache2_path) as f:
-            loaded_json2 = json.load(f)
-
-        assert loaded_json2["hash"] == "hash2"
-        assert loaded_json2["cache_type"] == "Deferred"
+        assert not loader.cache_hit(key("empty", "Pure"))
+
+
+# These are reasoanble, granular unit tests
```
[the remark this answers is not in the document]
mscolnick: yea feel free to remove them if they are in the way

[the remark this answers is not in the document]
dmadisetti: Do we have a test coverage goal?

marimo/_save/stores/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+    cache_config = (
+        get_default_config_manager(current_path=None)
+        .get_config()
+        .get("experimental", {})
+        .get("cache", {})
+    )
+    store_type = cache_config.get("store", DEFAULT_STORE_KEY)
+    assert store_type in CACHE_STORES, f"Invalid store type: {store_type}"
+    return CACHE_STORES[store_type](**cache_config.get("args", {}))
```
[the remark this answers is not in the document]
mscolnick: for future: i wonder if the `Store` could include a `dataclass` that we can parse/validate against the `args`

marimo/_save/stores/nix.py
```diff
@@ -0,0 +1,337 @@
...
+
+    This implementation mimics the API of FileStore but uses HTTP requests
+    to put and get serialized NAR files to/from a Cachix binary cache.
+    """
+
+    def __init__(
+        self,
+        cache_name: str = "marimo",
+        auth_token: str = "",
+        base_url: str = "https://app.cachix.org/api/v1",
```
mscolnick, line 44: is this on the free 5gb for open-source?

we also have http://cache.marimo.app/ (i can make the API look similar)

tests/_save/loaders/test_json_loader.py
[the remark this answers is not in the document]
mscolnick: no we don't, happens after merge and non-blocking

marimo/_save/stores/__init__.py
mscolnick: i guess you could always design it as:

```
[cache] 
store = "composite"
args = {
  "redis": {},
  "file": {},
}
```

marimo/_save/stores/nix.py
dmadisetti: Yup, one of the free personal use tokens.

Started playing with this thinking a tiny bit further out for run verification (nix has some interesting artifact signature ideas, and dependency tracking). But I might break it out and shelve it for the time.
`StaticStore`/ `BucketStore` for WASM might be the next best "Store" to explore instead.

I'm not convinced `cache.marimo.app` should leverage nix caching methods- but it might be a nice plug and play.

leventov:
> I see your code as just being another "Store", I think with this it doesn't need to hook in any deeper- but let me know if there's something I'm missing

The big remaining difference from [Persister](https://github.com/marimo-team/marimo/pull/4133/files#diff-600c74c3cc8aa9681d01545e9f4bf41a333b42d057e867a6af24aca890877cd7R125) is the presence of `created_at` in the keys. Other extra params to `get()` and `put()` that I added: `exact_match_created_at` and `content_spec` in principle could be shoveled into `kwargs`, although I think they just make sense (as well as other parts of the behaviour specifications of these methods documented in the link above that are not part of the method signatures themselves), so why don't make them explicit on the `Store` level. A single `loader` argument is not ready for other content kinds (not variables but HTML outputs and stderr strings), hence `content_spec`. Why loader-as-object won't work for `FsPersistentStore` I explained [here](https://github.com/marimo-team/marimo/pull/4133#issuecomment-2732550674). Also there the motivation for Entry vs. Cache:
> They main reason why I want to rename is because "Cache" is just a confusing name for this abstraction: "cache" is universally understood as the store as a whole, not a single entry within it. Also, it's a misnomer because "cache" is something ephemeral (even if on-disk), while FsPersistentStore can be used not just as "ephemeral cache" but also as "persistent execution store". This itself may delude LLM coding agents (though less so humans), but what will definitely confuse them more is conceptual "clash" is there is a mash-up of "entries" and "caches". LLM coding agents are very sensitive to such inconsistencies. Or else I should make the naming thoughout the refactored persistence subsystem completely consistent with "cache" (including cache_log/, cache_snapshots/, CacheMetadata class, etc.). Don't you think it will be way too confusing and misnomer-y for a persistence subsystem?

`on_exit()` *doesn't* need to be part of `Store` interface: whatever exit hooks a particular store needs, it can register itself in the constructor.

`Store.cleanup()` I included for CLI access (also possibly HTTP/UI access later) to `cleanup()` via `marimo persist cleanup`. It's not just a nicety, it's supposed to be called with default configs to actually clean the contents, as described [here](https://github.com/marimo-team/marimo/pull/4133/files#diff-600c74c3cc8aa9681d01545e9f4bf41a333b42d057e867a6af24aca890877cd7R1055), although I doubt the ergonomics of this decision myself and you can tell me this is a bad default. Maybe the default should be so-called "simple approach to object cleanup (see there) and if the user has Dropbox or something else non-cooperative for syncing contents it's on them to enable the "semi-automatic" approach with the need for explicit `marimo perisist cleanup --include_content` calls.

With all that, you may say that `cleanup()` can still be confined to `FsPersistentStore` and then Marimo will handle it manually, `isinstance(store, FsPersistentStore)` and then call this `cleanup()`. However, I think another reason to make this a general method on the `Store` level is to share the `CleanupStrategy` logic among all stores: I think it's too much cognitive burden for users if different stores have completely different "cleanup strategy" abstractions and logics.

marimo/_save/stores/__init__.py
leventov: > is there a use case for multiple caches?

A separate store for cache and "persistent cell execution" use case might make sense. "Persistent cell execution" can be configured as part of #3054 with an extra argument to the `App()` constructor, like `App(..., default_cell_store_results=True, cell_store_type="cache"|"persistent_execution")`.

With this, there could be two sections in `.marimo.toml`, `[cache]` and `[persistent_execution]` configuring *default* stores for these use cases for all notebooks that end up using this `.marimo.toml` file.

Apart from that, I envisioned that a specific store can be configured per app freely, a la `App(cache_store=RedisStore(...))`. Whether these store classes are singletons I think should be class-dependent rather than universal across all store classes.

marimo/_save/hash.py
```diff
@@ -435,7 +447,10 @@ def __init__(
 
         # Finally, utilize the unrun block itself, and clean up.
         self.cache_type = cache_type
-        self.hash_alg.update(hash_raw_module(module, hash_type))
+        module_hash = hash_raw_module(module, hash_type)
+        self.hash_alg.update(module_hash)
+        # Update execution path without content data.
+        self.exe_alg.update(module_hash)
```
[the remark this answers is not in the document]
leventov: `App.save_key` defaults to the notebook path relative to the parent of the `__marimo__` dir used for the given notebook, which itself is determined in the "dir ancestry search" way, similar to `.marimo.toml`. If the `Store` used is such that it doesn't use the local file system in this way (i.e., neither LegacyFileStore nor FsPersistentStore), or if `save_path` doesn't have a `__marimo__` element, the `save_key` defaults to just the notebook file name. It's user's responsibility to add explicit `App(save_key=former_save_key)` before renaming or moving the notebook if they *don't* want to lose the cache or the persistent execution history. If they rename or move file from the Marimo UI, this should be done automatically, though.

`Store`-level (*not* `FsPersistentStore`-level) configuration option `block_ids_include_app_save_key = true` (true is the default) controls whether block_ids are prefixed with `f"{save_key}:{module_hash_or_block_name}". Can be false if the user has "store per notebook" so including these save_keys would be unnecessary, or if their cleanup strategy doesn't make use of block_ids.

marimo/_save/stores/__init__.py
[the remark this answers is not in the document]
dmadisetti: Agreed. I think this change should go in with some docs too

marimo/_save/hash.py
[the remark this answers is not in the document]
dmadisetti: Alright, well removed execution hash on the key since I don't actively use it. I don't think a save key is needed on an App level when you can just use the file name.

marimo/_save/stores/__init__.py
dmadisetti: Failure in store mode shouldn't fallback I don't think. Ideally you want to change the method, or set the method for the particular notebook.

dmadisetti:
Removed most of Devin's tests, but tried to keep similar coverage. Also added backwards compatibility tests for loading old artifacts.

Experimented with adding a nix store, and Myles mentioned expanding cache.marimo.io, but I think a static store export for WASM would be the ideal next step

The Redis store feels pretty magical. I'm a fan.

mscolnick (approved):

PR #4155, opened 2025-03-18
improve: indication when cell output is loading

akshayka:
Decrease opacity of stale loading outputs to make staleness more easily parsed visually.

When cells are queued or running (not edited), it needs to be clear that their output is loading and stale. This is important both for editing notebooks and running long-running cells that have big outputs, as well as for app view.


mscolnick:
reading the comment that it used to be .2. i still think .5 is still too low / i find it hard to read

PR #4162, opened 2025-03-19
feat: add scroll to cell

### LINKED ISSUE #3346: Open a marimo notebook to a specific cell
### Description

You should be able to open a notebook to a specific cell with a url. 

### Suggested solution

I guess this would have to work by adding some kind of id to the @app.cell decorator that should be unique. That could get incorporated into a div or something in the browser for that cell.  I also guess there would have to be some javascript to move the browser to the desired id from a url.

Then a url like http://localhost:2718/?id=some-uuid would open to that cell.

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
Closes #3346

Add a new cell action to copy a link to it, and logic to scroll to the cell on startup.

This does require naming the cell. We could maybe later add using the cell-id, but that does change when the order changes.

Screenshot 2025-03-19 at 11 47 05 AM

### DIFF
--- a/frontend/src/components/editor/actions/useCellActionButton.tsx
+++ b/frontend/src/components/editor/actions/useCellActionButton.tsx
@@ -26,6 +26,7 @@ import {
   ChevronLeftIcon,
   ChevronRightIcon,
   ScissorsIcon,
+  LinkIcon,
 } from "lucide-react";
 import type { ActionButton } from "./types";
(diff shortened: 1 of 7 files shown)


akshayka (approved):

Light2Dark (approved):

mscolnick:
> looks good! do you think we should add a toast? `toast({ description: "Link copied to clipboard" });`

yes, just merged and then read that

PR #4168, opened 2025-03-20
fix: remove dml warning; add test

mscolnick:
This warning was logged with the query: 

```sql
insert into table1 (column1) select distinct column1 from table2 order by random();
```

The result looks correct, so I think it should be ok to remove this warning. cc @Light2Dark

### DIFF
--- a/marimo/_ast/sql_visitor.py
+++ b/marimo/_ast/sql_visitor.py
@@ -361,17 +361,12 @@ def find_sql_refs(
         if expression is None:
             continue
 
-        if is_dml := bool(expression.find(exp.Update, exp.Insert, exp.Delete)):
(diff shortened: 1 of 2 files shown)


Light2Dark:
bad error message lol, I thought it would add duplicate refs, but seems like it doesn't.

Light2Dark (approved):

marimo/_ast/sql_visitor.py
```diff
@@ -361,17 +361,12 @@ def append_refs_from_table(table: exp.Table) -> None:
         if expression is None:
             continue
 
-        if is_dml := bool(expression.find(exp.Update, exp.Insert, exp.Delete)):
```
Light2Dark, line 364: small nit (don't need to change this PR): is_dml is nice to keep for clarity imo

PR #4172, opened 2025-03-20
improvement: allow dropdowns to be any type

### LINKED ISSUE #4171: Dropdown should support non-strings
### Description

`mo.ui.dropdown(options=[1,2,3])` will crash with:

```
Bad Data
options.0: Expected string, received number
```

This can be worked around with dict comprehension.

### Suggested solution

It would be preferable if these `ui` elements just worked with options as any list, not just a list of strings. Same with `multiselect`. This would avoid doing `{str(x): x for x in [1,2,3]}` workarounds. 

### Alternative

_No response_

### Additional context

Image

mscolnick:
Fixes #4171 

Quite a small change for a big benefit. Dropdowns can now support any type

### DIFF
--- a/marimo/_plugins/ui/_impl/input.py
+++ b/marimo/_plugins/ui/_impl/input.py
@@ -879,6 +879,13 @@ class code_editor(UIElement[str, str]):
         return value
 
 
+def _to_option_name(option: Any) -> str:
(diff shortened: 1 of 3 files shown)


astrowonk:
Incredibly fast feature request turnaround!

PR #4193, opened 2025-03-21
feat: python 3.13 support

mscolnick:
3.13! 

Had to skip some of the `strict_mode` stuff, but its experimental anyways, so don't want to block.


akshayka (dismissed):
looks good except snapshot failing. will let @dmadisetti comment on the skipped tests

Light2Dark:
nice! curious, can we remove 3.9, no longer support it?

mscolnick:
@Light2Dark - not till 10/2025 and reaches end of life https://devguide.python.org/versions/

dmadisetti (commented):
Woops. Thought I left a review.

LGTM! Strict mode needs revisiting, and created the respective issue here: #4199

This behavior was reported before in strict mode and I wasn't able to replicate it, so good to know I have a place to start.

PR #4198, opened 2025-03-21
improve: smaller column summary charts

akshayka:
This change makes column summary charts smaller, both width and height, to prioritize fitting more table columns on the screen.

This change also replaces the manual bar width calculation and lets the bin sizes be determined by vega/the chart renderer.

Before: 

image

After: 

image

### DIFF
--- a/frontend/src/components/data-table/__tests__/__snapshots__/chart-spec-model.test.ts.snap
+++ b/frontend/src/components/data-table/__tests__/__snapshots__/chart-spec-model.test.ts.snap
@@ -45,13 +45,6 @@ exports[`ColumnChartSpecModel > snapshot > array 1`] = `
       "axis": null,
       "bin": true,
       "field": "a",
-      "scale": {
-        "align": 0,
-        "paddingInner": 0,
-        "paddingOuter": {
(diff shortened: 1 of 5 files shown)


mscolnick (commented):
Can you try when few bins like one or two

akshayka:
Yea, here's an example:


image

mscolnick:
with one it looks awkward: 

Screenshot 2025-03-21 at 5 50 22 PM

akshayka:
> with one it looks awkward:
> 
> Screenshot 2025-03-21 at 5 50 22 PM

This is similar to what's on main though? It's just a little wider. I updated this branch to not include the max label if it's the same as the min.

---

Somewhat related, there's an issue with using "nominal" for the bar chart type. I know we do it to accommodate nulls, but it interferes with the axis scaling. It put bars where there shouldn't be any sometimes. For example, this column looks uniform — but it actually isn't. There should be empty spaces. When we fix this, we won't be able to use custom bar widths, which is why I thought it made sense to just make the change to use the renderer's computed widths.

image

It should actually look like this:

image

PR #4205, opened 2025-03-22
improvement: default altair charts to full width, add typechecking

mscolnick:
This defaults the altair charts to be full-width when a width is not given. While it is opinionated, i think it looks much better. 

This can be ignore by providing an actual width. 

This also adds typechecking for altair in our codebase


akshayka (commented):
I think this may cause issues for numerical charts, where full width will distort the axes. I sometimes don't use ui.altair_chart() for this reason.

akshayka (approved):

PR #4208, opened 2025-03-23
improve: show column summary histogram axes

akshayka:
For numerical histograms in column summaries, this PR renders the x-axis with Vega. This makes plots more legible, especially when histograms are sparse.

Some examples:

image

image

Large numbers:

image


mscolnick (approved):
wow nice change

PR #4210, opened 2025-03-23
fix: j/k shouldn't make new cells in vim mode

### LINKED ISSUE #4209: vim: j/k creates many new cells
### Describe the bug

When you `j` past the last cell or `k` above the first cell then marimo keeps making new cells as long as you hold the character down. I frequently hold `j/k` down through cells so I frequently accidentally create 10 or 15 new cells which I have to then delete. In a normal text file, a vim user often holds those down to get to the end of the file where the cursor then stops and I suggest marimo would be best to do the same, i.e. `j` stop at the last line of the last cell and `k` stop at the first line of the first cell.

Users use `ctrl+shift+o|p` to create new cells above or below the current cell so I think it is perfectly adequate to just use those keybinds to create a new last or first cell. Or if you really think it necessary that `j` can create new cells at the end then perhaps limit it to only one new empty cell creation.

Arguably this is an enhancement request but accidentally creating a heap of empty cells is such a frequent annoyance that I have raised this as a bug.

### Environment

```
{
  "marimo": "0.11.25",
  "OS": "Linux",
  "Processor": "",
  "Python Version": "3.13.2",
  "Binaries": {
    "Browser": "--",
    "Node": "v23.9.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.31.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "ruff": "0.11.2",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.2.1",
    "pandas": "2.2.3",
    "polars": "1.25.2",
    "pyarrow": "19.0.1"
  },
  "Experimental Flags": {}
}
```

### Code to reproduce

_No response_

akshayka:
Fixes #4209

### DIFF
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -165,8 +165,8 @@ function useCellHotkeys(
         editorView.current?.focus();
       }
     },
-    "cell.focusDown": () => actions.moveToNextCell({ cellId, before: false }),
-    "cell.focusUp": () => actions.moveToNextCell({ cellId, before: true }),
+    "cell.focusDown": () => actions.moveToNextCell({ cellId, before: false, noCreate: true }),
+    "cell.focusUp": () => actions.moveToNextCell({ cellId, before: true, noCreate: true }),
     "cell.sendToBottom": () => actions.sendToBottom({ cellId }),
     "cell.sendToTop": () => actions.sendToTop({ cellId }),
     "cell.aiCompletion": () => {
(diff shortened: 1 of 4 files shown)


mscolnick (approved):
nice find

PR #4247, opened 2025-03-25
improvement : Adding functionality to handle numpy array as src in mo.audio simliar to ipython audio

### LINKED ISSUE #4197: `IPython.display.Audio` is not being updated
### Describe the bug

I have created a slider for start offset of an audio file. Moving it updates the value and re-runs the affected cells, but `Audio` is not updated, i.e. it still plays the same section. If I comment out that cell, run it, uncomment, and run it again, it is updated. I looked around if there is anything special I had to do for using `IPython.display` objects but couldn't find anything.

### Environment

<details>

```
{
  "marimo": "0.11.21",
  "OS": "Linux",
  "Processor": "",
  "Python Version": "3.9.21",
  "Binaries": {
    "Browser": "--",
    "Node": "v23.9.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.31.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "ruff": "0.11.0",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "polars": "1.25.2"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.11.21"


@app.cell
def _():
    import marimo as mo
    import torchaudio
    from IPython.display import Audio
    return Audio, mo, torchaudio


@app.cell
def _(torchaudio):
    SAMPLE_WAV = "/home/kureta/Music/Flute Samples/01. Air.wav"
    waveform, SAMPLE_RATE = torchaudio.load(SAMPLE_WAV)
    return SAMPLE_RATE, SAMPLE_WAV, waveform


@app.cell
def _(mo):
    offset = mo.ui.slider(0.0, 60.0, 0.5, label="Start offset in seconds.")
    offset
    return (offset,)


@app.cell
def _(Audio, SAMPLE_RATE, offset, waveform):
    start, end = (
        int(offset.value * SAMPLE_RATE),
        int((offset.value + 10.0) * SAMPLE_RATE),
    )
    Audio(data=waveform[:, start:end], rate=SAMPLE_RATE)
    return end, start


@app.cell
def _():
    return


if __name__ == "__main__":
    app.run()
```

xeniumcode:
Added functionality to handle `numpy` array in mo.audio 

fixes #4197

@akshayka

### DIFF
--- a/marimo/_plugins/stateless/audio.py
+++ b/marimo/_plugins/stateless/audio.py
@@ -60,18 +60,18 @@ def audio(
     """Render an audio file as HTML.
 
     Args:
-        src: a path or URL to an audio file, bytes,
+        src: a path or URL to an audio `file`, `bytes`,
             or a file-like object opened in binary mode
-            Numpy 1d array containing the desired waveform (mono)
-            Numpy 2d array containing waveforms for each channel.Shape=(NCHAN, NSAMPLES).
-        rate : integer
+            `Numpy 1d array` containing the desired waveform (mono)
+            `Numpy 2d array` containing waveforms for each channel.Shape=(NCHAN, NSAMPLES).
+        rate :
             The sampling rate of the raw data.
-            Only required when data parameter is being used as an array
-        normalize : bool
+            Only required when data parameter is being used as an array.
+        normalize :
             Whether audio should be normalized (rescaled) to the maximum possible
-            range. Default is `True`. When set to `False`, `data` must be between
+            range. Default is `True`. When set to `False`, `src` must be between
             -1 and 1 (inclusive), otherwise an error is raised.
-            Applies only when `data` is a list or array of samples; other types of
+            Applies only when `src` is a numpy array; other types of
             audio are never normalized.
 
     Returns:
--- a/marimo/_plugins/stateless/audio.py
+++ b/marimo/_plugins/stateless/audio.py
@@ -54,7 +54,7 @@ def convert_numpy_to_wav(data, rate: int, normalize: bool) -> bytes:
 
 def get_resolved_src(
     src: Union[str, io.BytesIO], rate: Optional[int], normalize: bool
-) -> str:
+) -> Optional[str]:
     """Determines the correct URL for the given audio source."""
 
     if isinstance(src, (io.BufferedReader, io.BytesIO)):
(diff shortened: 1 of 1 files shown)


marimo/_plugins/stateless/audio.py
```diff
@@ -49,6 +102,13 @@ def audio(
             resolved_src = mo_data.audio(
                 f.read(), ext=os.path.splitext(src)[1]
             ).url
+    elif isinstance(src, np.ndarray):
```
mscolnick, line 105: we should first check if numpy is imported with `DependencyManager.numpy.imported()`. 

i guess we don't have any existing tests. i can follow-up to create some tests for this file afterwards (unless you'd like to add them)

marimo/_plugins/stateless/audio.py
```diff
@@ -49,6 +102,16 @@ def audio(
             resolved_src = mo_data.audio(
                 f.read(), ext=os.path.splitext(src)[1]
             ).url
+    elif DependencyManager.numpy.imported():
```
mscolnick, line 105: i would pull this out into a local function with early returns:

```
def get_resolved_src():
    if ...

    if...
    
    if...
```

because this new `elif` actually makes the last `else` not get hit in same cases when it should

mscolnick (approved):

mscolnick:
@xeniumcode, looks great. just failing on some typechecking

xeniumcode:
for the input type of `def convert_numpy_to_wav(data, rate: int, normalize: bool) -> bytes:` do I also create DependencyManager.numpy.imported():

mscolnick:
@xeniumcode it should be fine not to because that code path already checks `DependencyManager.numpy.imported()`

mscolnick (approved):
awesome stuff!

PR #4269, opened 2025-03-26
poc: marimo edit data.csv

mscolnick:
Seems mildly useful and we can improve the default template for opening a csv/sql table.

usage:
```
marimo edit data.csv
marimo edit db.sqlite
```

### DIFF
--- a/marimo/_cli/file_path.py
+++ b/marimo/_cli/file_path.py
@@ -111,6 +111,10 @@ class DataFileReader(FileReader):
 
     def read(self, name: str) -> tuple[str, str]:
         path = Path(name)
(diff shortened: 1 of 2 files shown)


akshayka (commented):
I think I like this, but I wonder about the maintainability, and work required to improve the templates.

If we change from polars to something else in the future — is that a breaking change? I can also imagine people asking for Pandas, duckdb.

Might be fine but just thought I'd bring it up for discussion

mscolnick:
@akshayka, yea that was my worry too. it can open up a bag of requests / opinions. 

i'll let this sit for a bit and revisit

PR #4284, opened 2025-03-28
improve: cached session invalidation

akshayka:
This change adds cache invalidation to the session view restoration logic when autorun on startup is disabled. 

The cache is keyed on the notebook code and the marimo version. Invalidating on marimo version change may be more aggressive than needed, it is a proxy for plugin API changes; we can relax this in the future if needed.

### DIFF
--- a/marimo/_server/api/endpoints/ws.py
+++ b/marimo/_server/api/endpoints/ws.py
@@ -37,6 +37,7 @@ from marimo._server.model import (
     SessionMode,
 )
 from marimo._server.router import APIRouter
+from marimo._server.session.serialize import SessionCacheKey
 from marimo._server.sessions import Session, SessionManager
 from marimo._types.ids import CellId_t, ConsumerId, SessionId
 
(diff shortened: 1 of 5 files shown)


marimo/_server/api/endpoints/ws.py
```diff
@@ -651,7 +652,16 @@ def get_session() -> Session:
...
             # if auto-instantiate if false, replay the previous session
             if not self.auto_instantiate:
-                new_session.sync_session_view_from_cache()
+                from marimo import __version__
+
+                app = new_session.app_file_manager.app
+                codes = tuple(
+                    cell_data.code
+                    for cell_data in app.cell_manager.cell_data()
+                )
```
mscolnick, line 661: can these lines be pushed into `sync_session_view_from_cache`?

```
                from marimo import __version__

                app = new_session.app_file_manager.app
                codes = tuple(
                    cell_data.code
                    for cell_data in app.cell_manager.cell_data()
                )
```

akshayka: Done, also updated the tests, realized they weren't testing what I wanted

marimo/_server/session/serialize.py
```diff
@@ -285,13 +292,38 @@ def rename_path(self, new_path: str | Path) -> None:
...
+            _hash_code(code) != cell["code_hash"]
+            for code, cell in zip(key.codes, notebook_session["cells"])
+        ):
+            return False
+        if (
+            key.marimo_version
+            != notebook_session["metadata"]["marimo_version"]
+        ):
+            return False
+        return True
```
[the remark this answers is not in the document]
akshayka: I think all or nothing with code is better? We can get into weird states on partial matches, since we aren't hashing the ancestors of the cells. But maybe if we bring some of Dylan's caching machinery into this, we could do partial matches.

This does mean that partially executed notebooks won't have cached outputs but I think that's okay, since this is more of a quality of life feature rather than something people should rely on as a way to save outputs (at least for now).

mscolnick (approved):

PR #4299, opened 2025-03-28
fix: create public folder and then file

Light2Dark:
## 📝 Summary

Creates a public folder and then saves the base64 file inside it. If notebook is in a nested directory, it will save it in a public folder next to it.
```
![](public/screenshot.png)
![](nested/public/screenshot.png)
```

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/frontend/src/core/codemirror/markdown/__tests__/commands.test.ts
+++ b/frontend/src/core/codemirror/markdown/__tests__/commands.test.ts
@@ -220,7 +220,20 @@ describe("insertImage", () => {
     vi.mocked(sendCreateFileOrFolder).mockResolvedValueOnce({
       success: true,
       info: {
-        path: "hello.png",
+        path: "public",
+        name: "public",
+        isMarimoFile: true,
+        isDirectory: true,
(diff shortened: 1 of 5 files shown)


mscolnick:
@Light2Dark , we do have login the backend that will increment the file/folder if it already exists. i did this to match the functionality of vscode.

i think instead, we should just send one request (`./public/<filename>`)

and in `create_file_or_directory`, we should just also create the folders if it doesn't exist:


```python
        if file_type == "directory":
            full_path.mkdir(parents=True, exist_ok=True)
        else:
+          full_path.parent.mkdir(parents=True, exist_ok=True)
            full_path.write_bytes(contents or b"")
```

couple other things:

1. could we add a test for the change to `create_file_or_directory`.
2. you might want to test `marimo edit path/to/nested/notebook.py`. because it might surface a bug that instead of `./public/<filename>` as the filename. it might need to be `{store.get(filenameAtom)}/public/<filename>`

mscolnick (approved):
nice, clean solution!

PR #4301, opened 2025-03-28
improvement: start of secrets handling in the kernel

mscolnick:
This is just the BE implementation for reading secrets (currently just from environment variables).

Will add the frontend UI in another PR


dmadisetti (commented):
We have no https middleware right? Might be worth reminding users http is insecure to MITM. But that could be documentation or FE side

tests/_runtime/test_manage_script_metadata.py
```diff
@@ -302,10 +302,10 @@ async def test_install_missing_packages_micropip_other_modules(
     del sys.modules["already_installed"]
 
 
-async def test_broadcast_missing_packages(
+async def testmissing_packages_hook(
```
akshayka, line 305: `test_missing`

tests/_runtime/test_manage_script_metadata.py
```diff
@@ -434,10 +436,10 @@ def __init__(self) -> None:
     del sys.modules["pyodide"]
 
 
-def test_broadcast_missing_packages_pip(
+def testmissing_packages_hook_pip(
```
akshayka, line 439: `test_missing`

akshayka (dismissed):
Two small typos ow looks good

PR #4302, opened 2025-03-28
feat: utf-8 latex completions

### LINKED ISSUE #4300: UTF-8 latex-style autocompletion support
### Description

UTF-8 latex-style auto-completion in Marimo, allowing compact variable naming like for example λᵢ = 20 instead of "lambd_i = 20" and so on. It makes math-focused code so much more readable and beautiful.

Is this possible to implement in Marimo?

### Suggested solution

Ability to type for example \lambda\<tab\> to create a λ, \Sigma\<tab\> to create a Σ, \_1\<tab\> to create a ᵢ etc. Similar to the functionality available in for example Jupyter notebooks.

### Alternative

An easily accessible menu to auto-generate UTF-8 characters by picking. Similar to the "Symbols" menu in Microsoft Word.

### Additional context

_No response_

mscolnick:
Fixes #4300 

## Current
image

## Alternative 
Screenshot 2025-03-28 at 4 58 51 PM

### DIFF
--- a/frontend/src/core/codemirror/language/markdown.ts
+++ b/frontend/src/core/codemirror/language/markdown.ts
@@ -450,7 +450,9 @@ const getLatexSymbolList = once((): Completion[] => {
     displayLabel: command,
     type: "latex-symbol",
     boost: 10,
-    apply: symbol,
+    // We complete the command, instead of the symbol since
+    // some commands take arguments.
+    apply: `\\${command}`,
     info: () => {
       const div = document.createElement("div");
       div.textContent = `${symbol} ${description}`;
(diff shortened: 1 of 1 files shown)


akshayka (commented):
Huh interesting. I have never seen anyone type unicode into LaTeX directly. I tried it out. It seems to work for some symbols but this kind of completion doesn't make sense for symbols that expect arguments (like sqrt for example).

I tried Typst just now, as an example of a best-in-class editing experience for typesetting math. They have completion but they complete the latex symbol name instead of the unicode symbol — completing `$\sq$` to `$\sqrt$` instead of `$√$`. I know that's not what the original issue asked for but I think that makes more sense, and you can still see the beautiful math symbols in the output area.

johanjans:
I have not been able to try it out yet, but does this work also for python-variables outside of a $$-environment?

Another nice use case for something like this could also be for example π = sp.pi, etc.

PR #4306, opened 2025-03-29
feat: argparse support when running as a notebook

akshayka:
This change makes the kernel set `sys.argv` to the CLI extra arguments, i.e. the arguments after the filename, in edit and run sessions and when exporting. This matches semantics when running as a script. 

This makes it possible to write an argparse or simple-parsing specification that works when:

1.  editing or running a notebook as an app
2. running as a script
3. exporting to HTML / IPYNB

Without this change, argparse worked for 2 but not for 1 and 3.

For example, for

```
marimo edit notebook.py -- foo bar -baz 42
```
`sys.argv` would be set to

```
["notebook.py", "foo", "bar", "-baz", "42"]
```

With this change, `mo.cli_args()` is not really needed anymore, but we can keep it around (it has some [public usage](https://github.com/search?q=mo.cli_args&type=code)).


mscolnick:
It might be worth announcing this as a breaking change. I've see a few notebooks that used sys.args and stripped this themselves

mscolnick:
It might be worth calling out the all patches now in the docs and linking to the places they are mentioned in the docs.

It can be useful to other libraries consuming marimo, in case something goes wrong.

akshayka:
> Also probably should call out the all patches now in the docs and linking to the places they are mentioned in the docs.
> 
> It can be useful to other libraries consuming marimo, in case something goes wrong.

Done

> It might be worth announcing this as a breaking change. I've see a few notebooks that used sys.args and stripped this themselves

We can bump the minor and announce in the release notes on release

docs/guides/troubleshooting.md
```diff
@@ -101,13 +101,43 @@ If you find yourself needing to force execution order often, it might be a sign
...
+
+When running as a notebook, marimo makes the following changes to variables:
+
+- marimo patches `pdb.Pdb` with a custom class to enable interactive debugging
+  with the `breakpoint()` function
+- marimo patches `sys.argv` when running as a notebook to match what you would
+  see when [running as a script](../guides/scripts.md).
+- local variables currently have their names mangled, meaning source code introspection
+  that uses local variables may not work; this behavior may change in the future.
+
```
mscolnick, line 140: this is a great addition, nice

mscolnick (approved):

docs/guides/troubleshooting.md
akshayka: yea, good idea!

PR #4313, opened 2025-03-29
improvement: dataflow perf

mscolnick:
Improve typings and datastructures. 
Prefer inline list/set compression for hot paths as it can be more optimized.

### DIFF
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import threading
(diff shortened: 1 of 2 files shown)


dmadisetti (approved):
Nice- did you run a benchmark suite? Just wondering if the perf is quantified in some way

mscolnick:
@dmadisetti, i did not, and im usually the first one to say pre-optimization is the root of all evil / should measure before improving perf

I would love to create some benchmark suites, though

PR #4329, opened 2025-03-31
feat: show dropdown of environment keys in DB Connection UI

mscolnick:


akshayka (dismissed):
Looks good module failing test.

Want to add to docs on using env variables with databases?

I wonder whether if in the future we should force people to use env variables for database passwords. I can see people using the password field (thinking it will go through a secret manager), not looking at the generated code — and then accidentally sharing a notebook that has plaintext secrets.

PR #4348, opened 2025-04-01
feat: secrets panel

mscolnick:
Secrets panel. Currently only supports **reading** environment variables. 

Next steps will be writing and then different secret providers.

Screenshot 2025-04-01 at 11 46 36 AM

### DIFF
--- a/frontend/src/components/editor/chrome/panels/secrets-panel.tsx
+++ b/frontend/src/components/editor/chrome/panels/secrets-panel.tsx
@@ -1,0 +1,110 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+import React from "react";
+import { KeyIcon } from "lucide-react";
+import { useAsyncData } from "@/hooks/useAsyncData";
+import { Spinner } from "@/components/icons/spinner";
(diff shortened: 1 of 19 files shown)


frontend/src/components/editor/chrome/panels/secrets-panel.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { copyToClipboard } from "@/utils/copy";
+import { cn } from "@/utils/cn";
+import { SECRETS_REGISTRY } from "@/core/secrets/request-registry";
+import { Badge } from "@/components/ui/badge";
+import type { ListSecretKeysResponse } from "@/core/network/types";
+
+// dotenv providers should be at the top
+function sortProviders(providers: ListSecretKeysResponse["keys"]) {
+  return providers.sort((a, b) => {
+    if (a.provider === "env") {
```
Light2Dark, line 26: typo? there are 2 'env'

mscolnick: one is checking `a.provider` and the other is checking `b.provider` and they return the opposite, so i need both

Light2Dark (approved):

PR #4349, opened 2025-04-01
feat: write env vars from secrets panel

mscolnick:
image


Light2Dark:
some minor fixes, there was a duplicate model. The copy button can be reused in other components too imo (I took it from json output), but this is a start

Light2Dark (approved):

PR #4397, opened 2025-04-04
docs: understanding errors

akshayka:
Add docs pages on understanding common marimo errors:

* Multiple definitions
* Star imports
* Cycle errors

A follow-up will link to these from the marimo error output (using our canonical link shortener).


mscolnick (approved):

dmadisetti:
Doesn't have to be done now- but might be nice to give these specific errors (and potentially future ones, or suggestions re-lint) an ID

For instance, this is Ruff's error overview: https://docs.astral.sh/ruff/rules/

With a specific case: https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/

PR #4436, opened 2025-04-08
fix: place nulls at last

### LINKED ISSUE #4381: Table view sort always puts null values first
### Describe the bug

Sorting on a column in a table will always put nulls first, so if there are a lot of null values in the column you have to paginate through in order to see any non-null values (both in asc and desc order).

My preference is to always put nulls last, since if one is sorting on a column value, usually one wants to see rows with non-null values regardless if asc or desc.

### Environment

https://marimo.app/l/mvvscm

### Code to reproduce

```python
import marimo as mo
import pandas as pd

df = pd.DataFrame({"id": range(100), "value": list(range(50)) + [None]*50})
df
```

Light2Dark:
## 📝 Summary

Fixes #4381. For narwhals, pandas, polars, the change is straightforward. But it's trickier for default table.

Use tuple is None technique for default table sorting: [stackoverflow](https://stackoverflow.com/questions/18411560/sort-list-while-pushing-none-values-to-the-end)

There is one issue:
Sorting has never worked on dictionary (row-oriented data)
image

Because it's column names are not 'key' and 'value', so I've made a change to fix this. Not sure if it may lead to other bugs.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/default_table.py
+++ b/marimo/_plugins/ui/_impl/tables/default_table.py
@@ -324,7 +324,16 @@ class DefaultTableManager(TableManager[JsonTableData]):
                 )
             except TypeError:
                 # Handle when values are not comparable
-                def sort_func(i: int) -> tuple[bool, str] | str:
+                def sort_func_str(i: int) -> tuple[bool, str] | str:
+                    # For ascending, generate a tuple of (is_none, value)
+                    # (True, None) will be for None values
+                    # (False, x) will be for other values.
+                    # As False < True, None values will be sorted to the end.
+
+                    # For descending, (is_not_none, value) tuple
+                    # (False, None) will be for None values.
+                    # (True, x) will be for other values.
+                    # As True > False, other values come before None values
                     if descending:
                         return (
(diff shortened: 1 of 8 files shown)


marimo/_plugins/ui/_impl/tables/default_table.py
```diff
@@ -323,9 +324,18 @@ def sort_values(
                 )
             except TypeError:
                 # Handle when values are not comparable
+                def sort_func(i: int) -> tuple[bool, str] | str:
```
Light2Dark, line 327: i can try to refactor this too, repeated 3 times :>

mscolnick (dismissed):
this looks great. the duplication in the DefaultTable is ok if can't be simplified. It's well tested

Light2Dark:
I've marked a test as failing, lmk if that's not intended, we can also not fix the existing issue (dicts can't be sorted)

mscolnick:
@Light2Dark, looks great! this is fine to xfail for now

mscolnick (approved):

PR #4437, opened 2025-04-08
improvement: Configurable SQL output type

mscolnick:
**SQL Output Types**

marimo supports different output types for SQL queries, which is particularly useful when working with large datasets. You can configure this in your application configuration in the top right of the marimo editor.

The available options are:

- `"native"`: Uses DuckDB's native lazy relation (recommended for best performance)
- `"lazy-polars"`: Returns a lazy Polars DataFrame
- `"pandas"`: Returns a Pandas DataFrame
- `"polars"`: Returns an eager Polars DataFrame
- `"auto"`: Automatically chooses based on installed packages (first tries `polars` then `pandas`)

For best performance with large datasets, we recommend using `"native"` to avoid loading the entire result set into memory and to more easily chain SQL cells together. By default, only the first 10 rows are displayed in the UI to prevent memory issues.

### DIFF
--- a/docs/guides/working_with_data/sql.md
+++ b/docs/guides/working_with_data/sql.md
@@ -83,17 +83,17 @@ marimo supports different output types for SQL queries, which is particularly us
 
 The available options are:
 
-- `"native"`: Uses DuckDB's native lazy relation (recommended for best performance)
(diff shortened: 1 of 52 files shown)


Light2Dark (approved):
looks good, I'll see if we can show cursor.result better for sqlalchemy on native mode.

PR #4455, opened 2025-04-10
improvement: map vim :w to save notebook

mscolnick:


Light2Dark (approved):

frontend/src/core/codemirror/keymaps/vim.ts
```diff
@@ -99,6 +99,16 @@ const addCustomVimCommandsOnce = once(() => {
...
+
+  // Save command
+  Vim.defineEx("write", "w", (cm: CodeMirror) => {
+    const view = cm.cm6;
+    if (view) {
+      const actions = view.state.facet(cellActionsState);
+      actions.saveNotebook();
+    }
+  });
+  Vim.noremap("S", ":w<CR>", "normal");
```
andyrift, line 111: `S` is not mapped to `:w` in normal vim, and shouldn't be by default, I am sorry that my example caused confusion

PR #4461, opened 2025-04-10
Add debounce option to marimo-code-editor

nojaf:
## 📝 Summary

This is a part of #2921.

## 🔍 Description of Changes

I'm trying to introduce a `debounce` option similar to `mo.ui.text`.

## 📋 Checklist

## 📜 Reviewers

Hi @mscolnick, this works and is ready for review.

### DIFF
--- a/frontend/src/plugins/impl/CodeEditorPlugin.tsx
+++ b/frontend/src/plugins/impl/CodeEditorPlugin.tsx
@@ -63,22 +63,23 @@ const CodeEditorComponent = (props: CodeEditorComponentProps) => {
   const maxHeight = props.maxHeight ? `${props.maxHeight}px` : undefined;
 
   const [localValue, setLocalValue] = useState(props.value);
-  const { onChange: onChangeDebounce } = useDebounceControlledState<string>({
+  const { onChange: setValueDebounced } = useDebounceControlledState<string>({
     initialValue: props.value,
     delay: Number.isFinite(props.debounce) ? (props.debounce as number) : 0,
     onChange: props.setValue,
+    disabled: !Number.isFinite(props.debounce),
   });
 
   const handleChange = useCallback(
     (newValue: string) => {
       setLocalValue((_) => newValue);
       if (typeof props.debounce === "number") {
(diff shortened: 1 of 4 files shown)


frontend/src/plugins/impl/CodeEditorPlugin.tsx
```diff
@@ -33,6 +37,7 @@ export class CodeEditorPlugin implements IPlugin<T, Data> {
     minHeight: z.number().optional(),
     maxHeight: z.number().optional(),
     showCopyButton: z.boolean().optional(),
+    debounce: z.optional(z.union([z.boolean(), z.number()])),
```
mscolnick, line 40: instead of optoional, could you do `z.union([z.boolean(), z.number()]).default(false),`. this way the types from above would reduce from `debounce?: boolean | number;` to `debounce: boolean | number;` and its one less case to think about

frontend/src/plugins/impl/CodeEditorPlugin.tsx
```diff
@@ -57,6 +62,37 @@ const CodeEditorComponent = (props: CodeEditorComponentProps) => {
...
+        props.setValue(newValue);
+      }
+    },
+    [setValueDebounced, props.debounce, props.setValue],
+  );
+
+  const extensions =
+    props.debounce === true
+      ? [
+          events.content({
```
mscolnick, line 88: could we use: `EditorView.domEventHandlers({` blur: ()...` instead of bringing in another library?

i assume that library exists for a reason, so maybe it does more than `EditorView.domEventHandlers`? if so, can you add a note why?


also could we memoize the extensions:

```
const onBlur = useEvent(() => {
    return props.setValue(localValue);  
});

const extensions = useMemo(() => {
   if (props.debounce === true) {
        return [events.content({ blur: onBlur })]
   }

    return []
}, [props.debounce, onBlur]
```

the reason for `useEvent` is that this callback can stay memorized even if `localValue` changes.

mscolnick (approved):
nice job @nojaf

PR #4472, opened 2025-04-10
fix: setup cell is a 'root' to all cells

dmadisetti:
## 📝 Summary

Always run the setup cell if it is present and stale (making it like a root cell)

@akshayka

### DIFF
--- a/tests/_runtime/test_runtime.py
+++ b/tests/_runtime/test_runtime.py
@@ -880,6 +880,54 @@ class TestExecution:
         assert set(k.errors.keys()) == {"0", "1"}
         assert not k.graph.cells["1"].stale
 
+    async def test_setup_runs(self, any_kernel: Kernel) -> None:
+        k = any_kernel
+        from marimo._ast.names import SETUP_CELL_NAME
+
+        await k.run(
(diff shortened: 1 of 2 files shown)


tests/_runtime/test_runtime.py
```diff
@@ -880,6 +880,54 @@ async def test_child_of_errored_cell_with_error_not_stale(
...
+        if k.lazy():
+            # Wasn't stale previously
+            assert k.graph.get_stale() == set()
+            await k.run([er])
+        assert not k.graph.get_stale()
+        assert k.globals["y"] == 1
+
+        # set setup to stale
+        k.graph.cells[SETUP_CELL_NAME].set_stale(True)
+        print(k.graph.cells[SETUP_CELL_NAME].stale)
```
akshayka, line 926: Stray print to remove

akshayka (dismissed):

tests/_runtime/test_runtime.py
dmadisetti: Thanks!

akshayka (approved):

PR #4473, opened 2025-04-10
feat: support vimrc from the user config

mscolnick:
Support reading from the user's vimrc to create custom vim mappings in the editor.

cc @andyrift

### DIFF
--- a/configs/.vimrc
+++ b/configs/.vimrc
@@ -1,0 +1,2 @@
+" For testing "
+imap kj <Esc>
--- a/docs/guides/editor_features/overview.md
+++ b/docs/guides/editor_features/overview.md
@@ -40,6 +40,29 @@ marimo supports vim keybindings.
 
 - `gd` - go to definition
 - `dd` - when a cell is empty, delete it
+- `:w` - to save the notebook
+
+**Custom vimrc:**
+
+You can customize your vim experience by adding a `.vimrc` configuration in the user settings or pyproject.toml
+
+/// tab | User config
+
+```toml title="marimo.toml"
+[keymap]
+vimrc = /User/absolute/path/to/.vimrc
+```
+
+///
+
+/// tab | pyproject.toml
+
+```toml title="pyproject.toml"
+[tool.marimo.keymap]
+vimrc = relative/path/.vimrc
+```
+
+///
 
 ## Overview panels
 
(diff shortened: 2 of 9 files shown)


akshayka (approved):

andyrift:
That is good
I actually almost finished my own parser and it is a little different, I tried to go more general, but you have some smarter ideas for parsing text

I am not sure about the use of `mapCommand`, I went with separate implementations for `map`, `unmap`, `noremap` and `mapclear`

What should I do with my code?
I could create a separate pull request at least for inspiration

mscolnick:
@andyrift - once we release this (or unless you want to build from source), can you try this out and see if it works? if it doesn't would love if you can share your implementation in a PR as well

andyrift:
I built from source and tried to use this

Current implementation does not seem to work
- `imap jk <Esc>` leaves insert mode but inserts `k`
- `map H ^` does not go to the beginning of the line but instead inserts `H` on press in normal mode
- `map W :w<CR>` also does not work

All this might be due to mapping to `action` type command with `mapCommand`.
The other methods of the `Vim` object don't concern themselves with such details and just map the keys to each other

mscolnick:
ah ok, so i was totally off. thanks for checking @andyrift. does your registration logic do it correctly?

andyrift:
Yes, I just tested the same examples from above, but your code replaced with mine, and they work correctly

jabellcu:
I cannot manage to make it work. I am working on Windows, marimo 0.13.1. This is how my marimo.toml looks:
```
[keymap]
preset = "vim"
vimrc = "C:\Users\usr_name\_vimrc"

[keymap.overrides]

```

I have also tried to put the vimrc line under keymap.overrides. I have also tried to add an "r" before the double quotes. I have restarted marimo after every change. Nothing seems to work. Is there anything I am doing wrong? Thanks

mscolnick:
@jabellcu, probably a windows path issue. is there anything in your browser console if you filter by `vimrc`

andyrift:
@jabellcu, You could try escaping backslashes like `"C:\\Users\\..."` or replacing double quotes with single quotes like `'C:\Users\...'`

jabellcu:
> @jabellcu, You could try escaping backslashes like `"C:\\Users\\..."` or replacing double quotes with single quotes like `'C:\Users\...'`

Thank you guys. I have tried both approaches, and they both work: I no longer see any error in the marimo.toml file on startup, nor any 'Failed to get file details' message.

I still don't get the expected behavior, though. I have `inoremap jk <esc>` in my _vimrc file, but when I press `jk` in insert mode, it does not go back to normal mode. It just deletes the 'j' after pressing 'k'. I can write 'jk' if I wait a bit before pressing 'k'. Do you know what might be going on? Is it the same issue @andyrift mentioned above?

PR #4491, opened 2025-04-11
docs: clarify tests should be exclusive in cells

dmadisetti:
## 📝 Summary

Add some clarification regarding testing cells only comprising of `test_`, `Test`

maybe not aggressive enough?

### DIFF
--- a/docs/guides/testing/pytest.md
+++ b/docs/guides/testing/pytest.md
@@ -23,16 +23,18 @@ def __():
 
 @app.cell
 def __(inc, pytest):
-    def test_fails():
-        assert inc(3) == 5, "This test fails"
-
-    def test_sanity():
-        assert inc(3) == 4, "This test passes"
-
     class TestBlock:
-        @pytest.mark.parametrize(("x", "y"), [(3, 4), (4, 5)])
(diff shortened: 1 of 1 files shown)


docs/guides/testing/pytest.md
```diff
@@ -5,7 +5,8 @@
 By default, marimo detects and runs tests in your notebooks automatically. If
 the optional `pytest` library is installed, marimo will run `pytest` on any cells that
 contain test functions (functions that start with `test_`) or test classes
-(classes that start with `Test`) on a per cell basis.
+(classes that start with `Test`) on a per cell basis. **Note**, these cells
+should _only_ contain test functions and/or test classes.
```
akshayka, line 9: Here's my suggestion.

" marimo will run `pytest` on any cells that
contain test functions (functions that start with `test_`) or test classes
(classes that start with `Test`) — and only test functions or test classes. This means that if a cell contains functions, constants, variables or anything else that are not test functions, pytest will not run against them."

dmadisetti: +1

Incorporated this but with a bit of a deeper rewrite

mscolnick (approved):

PR #4496, opened 2025-04-11
fix: arviz formatter

### LINKED ISSUE #4494: `_repr_html_` not being used
### Describe the bug

I am having issues with `arviz.InferenceData` object rendering HTML

It has the `_repr_html_` implemented but it seems like the `__str__` is being used. 

### Environment

In the notebook

### Code to reproduce

```python
# /// script
# dependencies = [
#   "marimo",
#   "arviz",
# ]
# ///

import marimo

__generated_with = "0.12.8"


@app.cell
def _():
    import arviz as az

    centered_data = az.load_arviz_data("centered_eight")
    assert isinstance(centered_data, az.InferenceData)
    html = centered_data._repr_html_()
    assert isinstance(html, str)

    centered_data
    return az, centered_data, html


@app.cell
def _(centered_data):
    print(str(centered_data))
    return


@app.cell
def _(az):
    dir(az.InferenceData)
    return


if __name__ == "__main__":
    app.run()
```

Top looks like this

Image

But it is an interactive section in jupyter notebook

mscolnick:
Fixes #4494 

cc @Haleshot

### DIFF
--- a/marimo/_output/formatters/arviz_formatters.py
+++ b/marimo/_output/formatters/arviz_formatters.py
@@ -18,17 +18,10 @@ class ArviZFormatter(FormatterFactory):
         return "arviz"
 
     def register(self) -> None:
-        import arviz as az  # type: ignore
         import matplotlib.pyplot as plt  # type: ignore
         import numpy as np  # type: ignore
 
         from marimo._output import formatting
-
-        @formatting.formatter(az.InferenceData)  # type: ignore
-        def _format_inference_data(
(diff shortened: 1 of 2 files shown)


dmadisetti (approved):
_repr_html_ has been implemented for 5 years, so I think we're good on backwards compat too:

https://github.com/arviz-devs/arviz/blame/314b80586ce41f5b4f0c68e84d0f5e544c20c4c4/arviz/data/inference_data.py#L210

PR #4516, opened 2025-04-14
feat: move LSP out of experimental,  keep LSP optional

mscolnick:
- Move LSP out of feature flag
- Keep it optional by adding it to `KernelCapabilities`. Don't show errors when enabled but not installed (since it is enabled by default)

### DIFF
--- a/frontend/src/components/app-config/optional-features.tsx
+++ b/frontend/src/components/app-config/optional-features.tsx
@@ -20,7 +20,6 @@ import { Button } from "@/components/ui/button";
 import { cn } from "@/utils/cn";
 import { SettingSubtitle } from "./common";
 import { isWasm } from "@/core/wasm/utils";
-import { getFeatureFlag } from "@/core/config/feature-flag";
 
 interface Package {
   name: string;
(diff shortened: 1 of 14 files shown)


marimo/_plugins/ui/_core/ui_element.py
```diff
@@ -85,7 +85,7 @@ class MarimoConvertValueException(Exception):
 
 
 @mddoc
-class UIElement(Html, Generic[S, T], metaclass=abc.ABCMeta):
```
mscolnick, line 88: @akshayka @dmadisetti is this ok to remove? it made the LSP completions signature be `ABCMeta` instead of the UI element class name

mscolnick: copilot says: 
> Removing 'metaclass=abc.ABCMeta' from UIElement may inadvertently allow instantiation of an abstract class.

I think this is ok.

frontend/src/components/app-config/optional-features.tsx
```diff
@@ -178,6 +178,8 @@ export const OptionalFeatures: React.FC = () => {
           })}
         </TableBody>
       </Table>
+
+      <p className="text-muted-foreground mt-2">*Requires server restart</p>
```
dmadisetti, line 182: good flag, can you add this to pytest while you are here?

dmadisetti: well I guess kernel restart, not server

marimo/_plugins/ui/_core/ui_element.py
dmadisetti: Generally only used to enforce @abstractmethod which isn't used here- so no downsides

dmadisetti (approved):
No pyproject.toml changes?

mscolnick:
> No pyproject.toml changes?

I didn't want to add to the basic `marimo` yet. but its installable under `marimo[lsp]`

PR #4520, opened 2025-04-14
feat: add default table page size configuration

mscolnick:
Configurable default page size for the table. (defaults to 10)


Light2Dark:
I can help to add this but should we set min to 1 and max to 200 for this config?

Light2Dark (approved):

mscolnick:
@Light2Dark we don't have a limit enforced on the table but we can add one if you think there is a reasonable limit.

PR #4526, opened 2025-04-15
fix: correct pnpm version check

bulletmark:
This commit fixes two errors:

1. pnpm on Arch is version 10.8.0 but was failing the check because it was merely looking for "9.*" but it should look for any version 9 and above.

2. The check was looking for "9." anywhere in the version string but it should only look at major version component.


bulletmark:

dmadisetti (approved):
Nice. Really 3 errors because the "." wasn't released

bulletmark:
> Nice. Really 3 errors because the "." wasn't released

Yes, as you can see I fixed that as well but forgot about it when I wrote the points above.

PR #4528, opened 2025-04-15
fix: fix refresh interval constraint

bulletmark:
`mo.ui.refresh()` allows the user to set a period interval down to 0.1 secs but then artificially constrains it to >=1 sec at run time. There doesn't seem to be any downside to removing this constraint as done by this pr/commit.


bulletmark:

dmadisetti (approved):
LGTM! Thanks;

mscolnick (approved):

PR #4529, opened 2025-04-15
fix: share context with post-hooks to prevent id collisions

### LINKED ISSUE #4515: UI Element Registry Fails on Indirect call
### Describe the bug

Constructing a UI element in a round about way seems to trip up the registry

### Environment

HEAD

### Code to reproduce

see fileformat.py in #4497

dmadisetti:
## 📝 Summary

fixes #4515

Minimal repro: https://marimo.app/l/flusuu

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_runtime/runner/hooks_post_execution.py
+++ b/marimo/_runtime/runner/hooks_post_execution.py
@@ -265,25 +265,14 @@ def _broadcast_outputs(
         run_result.success()
         or isinstance(run_result.exception, MarimoStopError)
     ) and should_send_output:
-
-        def format_output() -> formatting.FormattedOutput:
-            formatted_output = formatting.try_format(run_result.output)
-
-            if formatted_output.exception is not None:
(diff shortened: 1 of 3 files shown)


marimo/_runtime/runner/hooks_post_execution.py
```diff
@@ -246,7 +246,7 @@ def _store_state_reference(
 @kernel_tracer.start_as_current_span("broadcast_outputs")
 def _broadcast_outputs(
     cell: CellImpl,
-    runner: cell_runner.Runner,
+    _runner: cell_runner.Runner,
```
akshayka, line 249: Looks like not used -- should we remove?

dmadisetti: Signature is needed for the call (just a general "hook")

mscolnick (approved):

akshayka (approved):

PR #4531, opened 2025-04-15
feat: add col filtering

### LINKED ISSUE #3973: improve column selection in data tables
### Description

This is a pretty cool feature that's useful for data analysis and investigation. I use DataGrip so I miss this feature in marimo. It also exists in Excel/Google sheets but isn't as convenient (multiple clicks) to use.

**DataGrip**
Image

https://www.jetbrains.com/help/datagrip/tables-filter.html#use_the_local_filter

**Google sheets**
Image

### Suggested solution

Likely need to add more backend functions to perform this search and call them from the frontend. Also need to keep in mind multi column selection.

Image

I think we could simplify the filtering, allow search & selection here.

### Alternative

_No response_

### Additional context

may need to set a limit for large tables / prevent long computation.

Light2Dark:
## 📝 Summary

Fixes #3973 
image
image

Backend will perform main aggregations, frontend can filter based on search query upon those aggregations

- [x] use Table component instead of raw table
- [x] logic for searching
- [x] select all, deselect all logic
- [X] ~new filter logic, apply IN clause~ fix select in logic
- [x] cleanup UI
- [x] caching results
- [x] draggable popover - tried using dnd-kit but it was painful/strange to implement, this seems simpler

## 📋 Checklist


frontend/src/components/data-table/column-header.tsx
```diff
@@ -522,23 +537,185 @@ const TextFilter = <TData, TValue>({
...
+}) => {
+  const [chosenValues, setChosenValues] = useState<unknown[]>([]);
+
+  const { data, loading, error } = useAsyncData(async () => {
+    if (!calculateTopKRows) {
+      return null;
+    }
+    const res = await calculateTopKRows({ column: column.id, k: 10 });
+    return await loadTableData(res.data);
+  }, []);
```
mscolnick, line 574: you can make the popover controlled with `const [isOpen, setIsOpen] = useState(false)`. and then only load if `isOpen` is true

Light2Dark: oh oops, sorry I left that comment in, it doesn't load immediately anymore.

marimo/_plugins/ui/_impl/table.py
```diff
@@ -132,6 +132,17 @@ class GetDataUrlResponse:
...
 
+@dataclass
+class CalculateTopKRowsArgs:
+    column: ColumnName
+    k: int
+
+
+@dataclass
+class CalculateTopKRowsResponse:
+    data: Union[JSONType, str]
```
mscolnick, line 143: i think we should just return the json `{"key": count}` or tuple: `[(key, number)]`. 

tuple might handle None/null better since keys get converted to strings

frontend/src/components/data-table/columns.tsx
```diff
@@ -110,6 +114,9 @@ export function generateColumns<T>({
   textJustifyColumns?: Record<string, "left" | "center" | "right">;
   wrappedColumns?: string[];
   showDataTypes?: boolean;
+  calculateTopKRows?: CalculateTopKRows;
+  setFilters?: SetFilters;
+  filters?: ColumnFiltersState;
```
mscolnick, line 119: passing these down will cause re-render to the columns. instead, these can be grabbed off the `column`  object in `header: ({ column }) => {`

frontend/src/components/data-table/column-header.tsx
```diff
@@ -522,23 +546,214 @@ const TextFilter = <TData, TValue>({
...
+      </Button>
+      <Button
+        variant="linkDestructive"
+        size="sm"
+        className=""
+        onClick={onClear}
+        disabled={clearButtonDisabled}
+      >
+        Clear
+      </Button>
```
mscolnick, line 756: could this be refactored into another file?

also, i think you can use `column.getFilterValue` and `column.setFilterValue`. here are the APIs: https://tanstack.com/table/v8/docs/guide/column-filtering#column-filter-apis

frontend/src/components/data-table/column-header.tsx
```diff
@@ -343,23 +372,180 @@ const TextFilter = <TData, TValue>({
...
+          <Input
+            value={query}
+            onChange={(e) => setQuery(e.target.value)}
+            placeholder="Search"
+            className="my-0 py-1"
+            autoFocus={true}
+          />
+          {dataTable}
+        </div>
+      </PopoverContent>
```
mscolnick, line 548: Small UX suggestion: if we can use a `<Combobox`, we can get keyboard navigation to work correctly.

it would mean you couldn't use the table, but you can still format each `ComboItem` the same way.

frontend/src/components/data-table/column-header.tsx
```diff
@@ -54,37 +80,47 @@ export const DataTableColumnHeader = <TData, TValue>({
...
+        </DropdownMenuTrigger>
+        <DropdownMenuContent align="start">
+          {renderDataType(column)}
+          {renderSorts(column)}
+          {renderCopyColumn(column)}
+          {renderColumnPinning(column)}
+          {renderColumnWrapping(column)}
+          {renderFormatOptions(column)}
+          <DropdownMenuItemFilter column={column} />
+          <RenderSetFilter onClick={() => setIsSetFilterOpen(true)} />
```
mscolnick, line 113: instead of `onClick`, what about as a `<DropdownMenuSub` like we do with the other filter

frontend/src/components/data-table/column-header.tsx
```diff
@@ -343,23 +372,180 @@ const TextFilter = <TData, TValue>({
...
+          <Button
+            variant="link"
+            size="sm"
+            onClick={() => setIsSetFilterOpen(false)}
+          >
+            X
+          </Button>
+        </PopoverClose>
+        <div className="flex flex-col gap-1.5 pt-1">
+          <span className="text-sm font-semibold mx-auto mb-2">Set Filter</span>
```
mscolnick, line 538: I wonder if this is now confusing with the existing full-text search filter for a column.

1. could we combine these?
2. could we name them something different: `:search-icon: Search`, `:filter Filter`

frontend/src/components/data-table/column-header.tsx
```diff
@@ -343,23 +361,220 @@ const TextFilter = <TData, TValue>({
...
+  calculateTopKRows?: CalculateTopKRows;
+  column: Column<TData, TValue>;
+}) => {
+  const [chosenValues, setChosenValues] = useState<unknown[]>([]);
+  const [query, setQuery] = useState<string>("");
+  const [position, setPosition] = useState({ x: 0, y: 0 });
+  const dragStartPos = useRef({ x: 0, y: 0 });
+  const [isDragging, setIsDragging] = useState(false);
+
+  const handleMouseDown = (e: React.MouseEvent) => {
```
mscolnick, line 391: i think it would be nice to pull this out to `components/ui/draggable-popover`

```tsx
const DraggabelPopover = (props: PopoverProps) => ...
```

frontend/src/components/data-table/column-header.tsx
```diff
@@ -343,23 +361,220 @@ const TextFilter = <TData, TValue>({
...
-    </div>
+          <GripHorizontalIcon className="h-5 w-5 mt-1 text-muted-foreground/40" />
+        </div>
+        <PopoverClose className="absolute top-2 right-2">
+          <Button
+            variant="link"
+            size="sm"
+            onClick={() => setIsFilterValueOpen(false)}
+          >
+            X
```
mscolnick, line 573: How about `<XIcon/>`? it might look a bit better

mscolnick (approved):

PR #4556, opened 2025-04-16
feat: release .env secrets out of experimental

mscolnick:
Release dotenv secrets out of experimental.

### DIFF
--- a/frontend/src/components/app-config/user-config-form.tsx
+++ b/frontend/src/components/app-config/user-config-form.tsx
@@ -1294,35 +1294,37 @@ export const UserConfigForm: React.FC = () => {
               )}
             />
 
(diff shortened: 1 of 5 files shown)


dmadisetti (commented):
`KeyRoundIcon` and `KeySquareIcon` seem like better choices

No obvious way (to the user) on how to get the variables. While in env (are all secrets going to be stuffed in env?), might be worth creating mo.secrets.get/set/has as a public api

I think it's still worth displaying `note: setting over http` when setting a secret if detected

I liked the UX of space turning into a _ but other characters are not:

fa8649f16f1df68aafab6c7a600b33ab

Thankfully this didn't crash my computer, but accidentally sourcing .env could

PR #4497, opened 2025-04-11
docs: new file format documentation

dmadisetti:
## 📝 Summary

Refreshes:
 - BYOE (bring your own editor) page
 - Fileformat tutorial

Introduces:
 - Reusable functions page

Issue #1543, opened 2024-06-02
`marimo lint/tidy notebook.py`


dmadisetti:
### Description

Discussion in https://github.com/marimo-team/marimo/discussions/1379 brings up the great point that tools like ruff will struggle with marimo since it cannot prune imports and unused variables (since all defs are returned)

### Suggested solution

marimo **can** suggest values to prune based on the graph.
It can check to see if there are any dangling `defs`.

some mo specific lint hints:
 - If a `def` is utilized in only one cell, suggest turning it private.
 - If a `def` is not utilized, suggest removing it.

Additional more opinionated tidy options
 - Suggestions to split or merge cells based on topology (e.g. 1 liners that are sequentially dependent or very large cells)
 - Suggestions to move cells with only code and no output to bottom of notebook
 - Suggestions to group imports into cells

This can be a wrapper/ preprocessor for a tool like ruff that can then check the notebook on a per cell level.

### Alternative

Use tools like ruff as is, but miss out on marimo-specific clean-up

dmadisetti:
I've been thinking about this a bit more. Notebooks are used to:
1. show stuff
2. or do stuff

lint should be able to determine if all code does either of these things.

It's easy to determine if code is used to "show stuff", as the code can be traced down to outputs.
Determining code that "does stuff" is a little harder, but can be made explicit.
It might be worth introducing the following api:

```python
from marimo.save import side_effect

with side_effect("Pinging site"):
    requests.get("example.com")
```

lint can then determine whether code blocks feed into either an output, OR a side effect. Common side effects like "open" could be aliased, with:

```python
with side_effect("write to file"):
    with open("file", "w") as f:
        f.write("x")
```

Being flagged as not needed.

---

Side effect blocks can also be leveraged for native saving, in which case the block will be forced to run, even if the rest of the cell is cached.

patrick-kidger:
Personally I'm -1 on this. I don't think Marimo should seek to reinvent every orthogonal piece of tooling -- this has always been one of the greatest weaknesses of Jupyter notebooks.

It should be possible to still achieve linting, type-checking etc. just by changing the serialize-as-`.py` format. For example, for the case of unused imports: just don't return them as outputs from the function. Marimo is already tracking downstream usages anyway, so this should be doable. Normal linters will then detect the unused import.

dmadisetti:
Pruning returns would be easy to do, and is a great idea- certainly an elegant way of getting cells to conform to style.

[...]

PR #4560, opened 2025-04-16
feat: release reusable functions

dmadisetti:
## 📝 Summary
Flag removal to turn on toplevel definitions.

Just in queue
Deps:
   #4497 -> #4557 -> this

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_tutorials/dataflow.py
+++ b/marimo/_tutorials/dataflow.py
@@ -10,7 +10,7 @@
 
 import marimo
 
-__generated_with = "0.12.2"
+__generated_with = "0.12.10"
 app = marimo.App()
 
 
@@ -259,14 +259,14 @@ def _(mo):
 def _():
     planet = "Mars"
     planet
-    return (planet,)
+    return
 
 
 @app.cell
 def _():
     planet = "Earth"
     planet
-    return (planet,)
+    return
 
 
 @app.cell(hide_code=True)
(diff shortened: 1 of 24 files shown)


marimo/_tutorials/dataflow.py
```diff
@@ -259,14 +259,14 @@ def _(mo):
 def _():
     planet = "Mars"
     planet
-    return (planet,)
+    return
```
akshayka, line 262: Here and elsewhere — why have the definitions been removed from the returns?

dmadisetti: Because they are no longer referenced by another cell, which lets mypy point the out to the user, but sounds like we need to emphasize that in the docs

akshayka: Oh I see. I may have missed this in an earlier PR.

But there are scenarios in which you define variables that aren't used by other cells — but that isn't necessarily an error, no? In which case I wouldn't want mypy complaining.

dmadisetti: only complains if you create the variable but never use it, so 

Screenshot 2025-04-17 at 3 04 35 PM

akshayka: Okay good point. Discussion from https://github.com/marimo-team/marimo/issues/1543#issuecomment-2358434583 also makes sense.

akshayka (approved):
# 🧩 🚢

PR #4561, opened 2025-04-17
fix: remove auto-scroll from chat

mscolnick:


koaning (approved):
Just ran this locally and it indeed feels a bunch better. Thanks!

PR #4571, opened 2025-04-17
fix: Fix threading to properly deconstruct + use state

### LINKED ISSUE #4564: mo.Threads are not cleaned up
### Describe the bug

On changing a mo.Thread target definition, an noticed that the background thread was still running (I just changed print from 1 to 2 of the smoke test and saw both print statements)

Likewise the plain threading.Thread definition seemed to be running on cell deletions. It might be worth having thread specific deletion handlers if we're going to support mo.Thread

### Environment

<details>

```
```

</details>

### Code to reproduce

```python


import marimo

__generated_with = "0.12.10"
app = marimo.App()


@app.cell
def _():
    import marimo as mo
    from time import sleep
    return mo, sleep


@app.cell
def _(sleep):
    def poll():
        while True:
            sleep(1)
            print(2)
    return (poll,)


@app.cell
def _():
    import threading
    return


@app.cell
def _(mo, poll):
    with mo.redirect_stdout():
        t = mo.Thread(target=poll)
        t.start()
    return (t,)


@app.cell
def _(t):
    t.is_alive()
    return


if __name__ == "__main__":
    app.run()
```

dmadisetti:
## 📝 Summary

fixes #4564

And state seems to work. Now actually going to bed 😷 

@akshayka

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -532,6 +532,7 @@ class Kernel:
             self._post_execution_hooks.append(_render_toplevel_defs)
 
         self._globals_lock = threading.RLock()
+        self._state_lock = threading.RLock()
         self._completion_worker_started = False
 
         self.debugger = debugger_override
@@ -959,6 +960,9 @@ class Kernel:
                     )
             else:
                 if name in self.globals:
+                    if isinstance(self.globals[name], threading.Thread):
+                        # If the variable is a thread, we need to stop it
+                        self.globals[name].join()
                     del self.globals[name]
 
                 if (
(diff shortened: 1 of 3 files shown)


marimo/_runtime/runtime.py
```diff
@@ -959,6 +960,9 @@ def _delete_variables(
                     )
             else:
                 if name in self.globals:
+                    if isinstance(self.globals[name], threading.Thread):
+                        # If the variable is a thread, we need to stop it
+                        self.globals[name].join()
```
akshayka, line 965: I don't think this is right. This will block the kernel as long as the thread is running. I'd rather leak the thread, or find a way to terminate it without rendering the kernel unusable.

You can try by running the following in a `mo.Thread` then deleting the cell:

```python
def foo():
  import time
  while True: time.sleep(1)
```

akshayka (commented):
I don't think the thread cleanup logic is correct, see my comment.

I also don't quite understand what issue the change to state are solving. Using state in threads isn't well-supported right now, since state updates only trigger execution when the kernel is processing a run request. Not sure that these changes will fix the underlying issue.

I'd prefer for this change to only address one thing, either thread destruction or state. But getting state working in threads right now is low priority, so I wouldn't spend cycles on it.

marimo/_runtime/runtime.py
dmadisetti: It's not right, and with the current smoke test won't even trigger since it needs a bound name to even get here.

dmadisetti:
I was tinkering around last night thinking about a dependable backend mo.ui.refresh for essentially cron - but I hear you on how "off the clock" cycles can contribute to working cycles

dmadisetti:
Closed by #4782, but may be good starting point for adding state

PR #4576, opened 2025-04-17
docs: minor function fileformat edits

akshayka:
Minor edits

* explain returned defs
* simplify reusing functions doc

### DIFF
--- a/docs/guides/reusing_functions.md
+++ b/docs/guides/reusing_functions.md
@@ -52,6 +52,7 @@ import numpy as np
 ```
 
 To add a setup cell in the editor, open the notebook menu and select "Add setup cell".
+(The setup cell is guaranteed to run before other cells.)
 
 ### 2. Define your function
 
@@ -119,5 +120,5 @@ print(stats)
 ## Learn more
 
 For more on marimo's file format, check out our [documentation on using your
-own editor](https://docs.marimo.io/guides/editor_features/watching/) or view
-our [file format tutorial](https://marimo.app/?slug=8n55fd).
+own editor](../guides/editor_features/watching.md) or view
+our [file format tutorial](https://links.marimo.app/tutorial-fileformat).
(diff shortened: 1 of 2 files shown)


docs/guides/reusing_functions.md
```diff
@@ -51,9 +51,7 @@ classes will need:
 import numpy as np
 ```
 
-To add a setup cell in the editor, use the General menu and select "Add setup cell" (💠).
-
-Setup cells are guaranteed to run before all other cells.
+To add a setup cell in the editor, open the notebook menu and select "Add setup cell".
```
dmadisetti, line 54: I still don't think the options is super findable by itself, and the latter statement is true

docs/guides/reusing_functions.md
```diff
@@ -137,7 +118,6 @@ class DataProcessor:
 
 ## Learn more
 
-For more detailed information about marimo's file format and features, check
-out our [documentation on using your own
-editor](https://docs.marimo.io/guides/editor_features/watching/) or view our
-[file format tutorial](https://marimo.app/?slug=8n55fd).
+For more on marimo's file format, check out our [documentation on using your
+own editor](https://docs.marimo.io/guides/editor_features/watching/) or view
+our [file format tutorial](https://marimo.app/?slug=8n55fd).
```
dmadisetti, line 123: Probably worth converting the tutorials to link.marimo, I'll take the latter half if you grab the first

marimo/_tutorials/fileformat.py
```diff
@@ -147,7 +147,7 @@ def _(mo):
 
         In the generated code, there is a function for each cell. The arguments
         of  the function are the cell's refs , and its returned variables are
-        its defs.
+        its defs that are referenced by other cells.
```
dmadisetti, line 150: 👍🏾

docs/guides/reusing_functions.md
dmadisetti: Actually, got them all. prefix `links.marimo.app/tutorial-<>`

docs/guides/reusing_functions.md
akshayka: Yea requires some searching. The emoji I found confusing, we could find a way to get the lucide icon in here at some point but didn't want to set that up. Added the second sentence back but in parentheses, since it isn't the point we're trying to get across.

PR #4587, opened 2025-04-18
fix: use context.issue in notify translators

akshayka:
The previous workflow used an API that doesn't exist on merged pull requests (`github.event`)


dmadisetti:
@akshayka Maybe a semi-superifical change to the readme to test it?

akshayka:
> @akshayka Maybe a semi-superifical change to the readme to test it?

@dmadisetti good call, done

PR #4588, opened 2025-04-18
fix: test_save_with_invalid_file assertion

akshayka:
CI is failing because (it seems) the temporary file in this test does not actually start with `import marimo` but instead starts with script metadata.

I'm not sure why it's starting with script metadata (a side effect of another test?), but it's better to check that the file starts with what it originally started with anyway (assuming it was a valid file).

### DIFF
--- a/tests/_server/api/endpoints/test_files.py
+++ b/tests/_server/api/endpoints/test_files.py
@@ -157,8 +157,9 @@ def test_save_with_invalid_file(client: TestClient) -> None:
         + "# Linter ignore\n"
     )
 
-    # Prepend a header to the file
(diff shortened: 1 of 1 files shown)


dmadisetti:
This is due to a race condition with uv. Woops should have synced before looking into it myself. Other tests are flaky because of this too in the same file

I've only seen it on 3.10 though

akshayka:
> This is due to a race condition with uv. Woops should have synced before looking into it myself. Other tests are flaky because of this too in the same file
> 
> I've only seen it on 3.10 though

All good, I should have synced too! We can go with your fix

PR #4589, opened 2025-04-18
stop gap: check header for race condition

dmadisetti:
## 📝 Summary
Stop gap for flaky test

Supersecedes https://github.com/marimo-team/marimo/pull/4588

@akshayka

### DIFF
--- a/tests/_server/api/endpoints/test_files.py
+++ b/tests/_server/api/endpoints/test_files.py
@@ -143,9 +143,7 @@ def test_save_with_header(client: TestClient) -> None:
     # Race condition with uv (seen in python 3.10)
     if file_contents.startswith("# ///"):
         file_contents = file_contents.split("# ///")[2].lstrip()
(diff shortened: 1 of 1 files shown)


dmadisetti:
I could have explicitly checked this works but did

`for i in (seq 100); hatch run +py=3.10 test:test tests/_server/api/endpoints/test_files.py; end`

which was fine locally

akshayka (approved):
I couldn't figure out why script metadata was being added in the first place. But I'm fine with this as a stopgap.

dmadisetti:
I'm pretty sure uv inserts it instead of us? which is why we get the race condition
but yes, not sure how UV is deciding to do that either

dmadisetti:
I'm going to manually retrigger the 3.10 after tests are green just to be sure

akshayka:
Yea, strange. I checked and uv/package manager should be mocked in our tests.

Issue #4581, opened 2025-04-18
Table component losing selection state when searching


sorinbiriescu:
### Describe the bug

Hello,

Using the Table component, the row selection is lost when doing a search, table.value is set to empty list as well. For tables with a lot of variables / columns, this is a problem as I use search often to find the rows of interest and use the selection as a subset for analysis in other marimo components.

### Environment

Unfortunately, I'm getting an error with marimo env, but I have marimo-0.12.10 installed

<details>

```
  File "C:\Users\XXXXXX\AppData\Local\Programs\Python\Python310\lib\site-packages\marimo\_cli\envinfo.py", line 62, in get_system_info
    "Node": get_node_version() or "--",
  File "C:\Users\XXXXXX\AppData\Local\Programs\Python\Python310\lib\site-packages\marimo\_utils\health.py", line 25, in get_node_version
    return stdout.strip().split()[-1]
IndexError: list index out of range
```

</details>

mscolnick:
@sorinbiriescu - this is because not all tables have a stable ID. we do support this for polars/pandas, but cannot support this for interchange tables (duckdb or ibis) or lists/dictionaries.

sorinbiriescu:
The tables are constructed from a list of dict
`[{"col_1": "value A", "col_2": "value B"}, ...]`

To give the full context, I'm manipulating SPSS files via pyreadstat, the files have integrated data + metadata with variable (column) names and labels. The library allows to get the metadata without loading the entire dataset. So the metadata can be extracted to this list of dict that I use to construct the table in marimo and select only the variables (columns) that I need, then load them from the main dataset. It saves a lot of time as normally I'm interested in 10-15 variables at a time from 2000, the performance is notable. The problem, as noted before, is that I need to use the search function of the table to find what I need in the metadata table and it's losing the previous selections. 

I did make a callback function for the table that records all selections to a set + a button to reset the set, but it seems for me not so elegant, so I wanted to check if I was doing something wrong on my side or if this is a limitation of the table component. I checked the docs, but no luck.

manzt:
> So I did a small test and it seems to keep the selection if the list of dict is made into a dataframe and then the table created from the dataframe. So I guess it's a non-issue.

Great. I'm going to close this out!

PR #4590, opened 2025-04-18
fix: get node version edge case

akshayka:
Fix marimo env when node version is not obtainable. Came up in #4581


dmadisetti (approved):
Surprised mypy didn't catch this? Maybe I expect too much from static testing

dmadisetti:
Oh, I guess still a string

PR #4592, opened 2025-04-18
docs: changelog

akshayka:
Link to the GitHub releases for the changelog, using it as the source of truth. (This is what Polars does too.)

In the future we can mirror the release notes in our docs on a dedicated page. But I didn't want to invest the effort in that right now, and this is still an improvement.


dmadisetti (approved):
+1 !

In case this does change want to flag that nixpkgs also expects a change log location: https://github.com/r-ryantm/nixpkgs/blob/bc0c7a7d0984c622dedd103d1f7c3e1ac0ef07bf/pkgs/development/python-modules/marimo/default.nix#L83

PR #4593, opened 2025-04-18
chore: add setup_cell to released experiments

akshayka:
`setup_cell=True` was in my marimo.toml


dmadisetti:
I don't think this was ever actually set. Not sure how it got there

dmadisetti:
But if it's in your config, it might be in others- so worth putting it in finished

dmadisetti (approved):

PR #4596, opened 2025-04-19
fix: improve graceful shutdown when running under uv

### LINKED ISSUE #4224: marimo fails to terminate using double ctrl-c when running under uv
### Describe the bug

If I run marimo, e.g. `marimo tutorial intro`, then I can stop it easily by typing ctrl-c twice quickly, once to initiate termination and another to not bother answering the subsequent `Are you sure you want to quit? (y/n)` prompt. That is how I normally terminate marimo.

However, if I run `uvx marimo tutorial` intro then the second cntl-c just hangs and I have to kill the uv tool process manually.

I raised [this issue](https://github.com/astral-sh/uv/issues/12108#issuecomment-2745933178) on the uv issue-tracker thinking it was a uv problem but zanieb there suggest it is an issue in marimo. I have since discovered that the issue exists natively in marimo when running in a sandbox (which also runs under uv).

E.g. run `marimo edit --sandbox` then create a new notebook and then press ctrl-c twice.

### Environment

```
{
  "marimo": "0.11.26",
  "OS": "Linux",
  "Processor": "",
  "Python Version": "3.13.2",
  "Binaries": {
    "Browser": "--",
    "Node": "v23.9.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.31.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "ruff": "0.11.2",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.2.1",
    "pandas": "2.2.3",
    "polars": "1.25.2",
    "pyarrow": "19.0.1"
  },
  "Experimental Flags": {}
}
```

### Code to reproduce

_No response_

akshayka:
Improves if not completely fixes #4224 

uv sends one more SIGINT than the user does after two Ctrl-Cs, which
was causing the server to die without cleanup. This introduces some
hacks to work around that behavior. It's not perfect but appears
much improved.

Tested by manually trying various combinations of:

```
[uv run] marimo tutorial intro --headless
```

followed by multiple ctrl-c's, and various combinations of `y` and `n` to the shutdown prompt.

Also fixes an issue with graceful shutdown with `--sandbox`.

A better solution would be to make shutdown robust to an arbitrary number of interrupts, but that ended up being a headache, so if this is still a problem after this change is merged we can just disable the double ctrl-c-to-exit behavior and tell people to use `Ctrl-D` or the manual confirm, neither of which leak resources.

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -382,10 +382,14 @@ def run_in_sandbox(
     def handler(sig: int, frame: Any) -> None:
         del sig
         del frame
-        if sys.platform == "win32":
-            os.kill(process.pid, signal.CTRL_C_EVENT)
-        else:
-            os.kill(process.pid, signal.SIGINT)
+        try:
+            if sys.platform == "win32":
+                os.kill(process.pid, signal.CTRL_C_EVENT)
+            else:
+                os.kill(process.pid, signal.SIGINT)
+        except Exception:
+            # Process may have already been terminated.
+            pass
 
     signal.signal(signal.SIGINT, handler)
 
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -387,7 +387,7 @@ def run_in_sandbox(
                 os.kill(process.pid, signal.CTRL_C_EVENT)
             else:
                 os.kill(process.pid, signal.SIGINT)
-        except Exception:
+        except ProcessLookupError:
             # Process may have already been terminated.
             pass
 
--- a/marimo/_server/api/interrupt.py
+++ b/marimo/_server/api/interrupt.py
@@ -15,11 +15,9 @@ from marimo._server.utils import (
 class InterruptHandler:
     def __init__(self, quiet: bool, shutdown: Callable[[], None]) -> None:
         self.quiet = quiet
-        self._shutdown = shutdown
+        self.shutdown = shutdown
         self.loop = asyncio.get_event_loop()
         self.original_handler = signal.getsignal(signal.SIGINT)
-        self._confirming_exit = False
-        self._has_shutdown = False
 
     def _add_interrupt_handler(self) -> None:
         try:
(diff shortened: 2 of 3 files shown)


marimo/_cli/sandbox.py
```diff
@@ -382,10 +382,14 @@ def run_in_sandbox(
...
-        if sys.platform == "win32":
-            os.kill(process.pid, signal.CTRL_C_EVENT)
-        else:
-            os.kill(process.pid, signal.SIGINT)
+        try:
+            if sys.platform == "win32":
+                os.kill(process.pid, signal.CTRL_C_EVENT)
+            else:
+                os.kill(process.pid, signal.SIGINT)
+        except Exception:
```
dmadisetti, line 390: `ProcessLookupError` if the process is already gone

marimo/_server/api/interrupt.py
```diff
@@ -55,18 +55,28 @@ def _interrupt_handler(self) -> None:
...
+                    self.shutdown()
+                    return
 
-            response = input(
-                f"\r{TAB}\033[1;32mAre you sure you want to quit?\033[0m "
-                "\033[1m(y/n)\033[0m: "
-            )
-            if response.lower().strip() == "y":
+                response = input(
+                    f"\r{TAB}Are you sure you want to quit? (y/n): "
```
dmadisetti, line 64: I think we have the colors.py lib if we want to add back parity

marimo/_server/api/interrupt.py
```diff
@@ -55,18 +55,28 @@ def _interrupt_handler(self) -> None:
...
+                if response.lower().strip() == "y":
+                    self.shutdown()
+                    return
+            except (KeyboardInterrupt, EOFError, asyncio.CancelledError):
+                print_()
                 self.shutdown()
                 return
-        except (KeyboardInterrupt, EOFError, asyncio.CancelledError):
+        except KeyboardInterrupt:
+            # This is a hack to workaround the fact that uv can send two SIGINT for
```
dmadisetti, line 75: Thought through this 

```python
def handle(depth=0):
    try:
        if depth < MAX_ATTEMPTS:
            return handle(depth + 1)
        ...
    except:
        ...
```

but **still** a chance for a race condition killing it before recursion.

I also considered it might be best to debounce in the sandbox pass to the process
BUT another hypothesis for the double sigint, is that hitting ctrl-c in your terminal sends sigint to the process group (or maybe sending 1 to uv and uv is sending to the whole process group) so manually sending the sig int in sandbox at all is actually duplicate behavior.

If it's not duplicate behavior (or maybe terminal / uv version specific), then I think this is good enough

dmadisetti (commented):
Might be worth testing the removal of the sandbox forwarding since UV might be forwarding to the whole process group

marimo/_server/api/interrupt.py
akshayka: Yea I was just thinking about trying something like this. Will try it out

marimo/_server/api/interrupt.py
akshayka: For some reason `input` is not colorizing and instead showing the ANSI codes. I see this on my machine and another user saw it too (Discord), so simpler to just not colorize this.

marimo/_server/api/interrupt.py
akshayka: The issue is that `input()` is not reentrant so we can't re-enter `_input_handler.` I tried some other logic, to have a custom handler that raises the KeyboardInterrupt, keeping track of how many times it was called. But that also has a race and ended up being more buggy. Kept the logic as is.

akshayka:
> Might be worth testing the removal of the sandbox forwarding since UV might be forwarding to the whole process group

I tested this but sandbox needs it, otherwise the main program (which is on `process.wait()`) just aborts.

dmadisetti (approved):

akshayka:
Failing tests look unrelated

PR #4611, opened 2025-04-21
fix: Filter by values styling and selection

mscolnick:
- styling: reduce padding and change icon. fix height in loading and error state
- use Set over list for better perf
- fix select all when there are no options from filtering. 
- fix using `Enter` and click to select rows

image


Light2Dark:
thanks!

Light2Dark (approved):

PR #4635, opened 2025-04-23
fix: sandbox alerts to stderr

dmadisetti:
## 📝 Summary

`marimo export md mynotebook.py > file`

creates a bunch of noise at the top of the file. Easy way to get the same functionality is just to log to stderr instead of stdout

@akshayka OR @mscolnick


akshayka (approved):

akshayka:
Looks good except the CLI test needs to be updated

mscolnick (approved):

PR #4638, opened 2025-04-23
marimo-snippets on dataframe page

koaning:
Maybe it starts making sense to add more of these snippets on our docs? This is a WIP to see @akshayka how you feel about this. I think it makes some sense to keep some of the GIFs around because they help highlight what you can do.


akshayka (commented):
We should keep the GIFs to highlight various features of the things we're showing.

As for snippets, the mkdocs directive is better suited to our docs, at least today:

* it can take a file as input
* consistent with the rest of our docs

Snippets is still valuable, since it lets us and others embed outside mkdocs.

koaning:
Added the snippets and removed the marimo-snippets usage.

koaning:
@akshayka spotted this really old PR, close or merge?

docs/guides/working_with_data/dataframes.md
```diff
@@ -43,21 +43,41 @@ df
 /// tab | polars
 
 ```python
-import polars as pl
+import pandas as pd
```
mscolnick: shouldn't this be polars?

docs/guides/working_with_data/dataframes.md
```diff
@@ -134,12 +174,29 @@ transformed_df.value
 
 ///
 
-<div align="center">
-<figure>
-<img src="/_static/docs-dataframe-transform-code.png"/>
-<figcaption>Copy the code of the transformation</figcaption>
-</figure>
-</div>
```
mscolnick, line 142: personally i like the `gif` (they are quick to load and demo what to do).

could we put the demos in another tab? `🐻‍❄️ polars | 🐼 pandas | ⚡ live example`

koaning: Good call, moving stuff into a "live example" tab makes everything look a lot cleaner.

koaning: I'll put the image back, it's not a gif, but I don't really recall why it got removed. I guess I might've imagined that we no longer needed it now that there was a live demo?

akshayka (approved):
This is great! Really love the tabbed live example. It also feels like it loads instantly (because I guess it loads in the background).

mscolnick (approved):

PR #4645, opened 2025-04-23
fix: export to html when using arrow files

mscolnick:


frontend/src/core/static/files.ts
```diff
@@ -79,6 +79,16 @@ export function patchVegaLoader(
 
   function maybeGetVirtualFile(url: string): string | undefined {
     const pathname = new URL(url, document.baseURI).pathname;
+
+    // If if the URL starts with file://, then using the document.baseURI
+    // will not work. In this case, should just chop off from /@file/...
+    if (url.startsWith("file://")) {
```
dmadisetti, line 85: I couldn't see how this change is related? Looks fine, was just wondering

marimo/_server/export/exporter.py
```diff
@@ -84,10 +84,7 @@ def export_as_html(
                     e,
                 )
                 continue
-            mime_type, _ = mimetypes.guess_type(basename) or (
-                "text/plain",
-                None,
-            )
+            mime_type = mimetypes.guess_type(basename)[0] or "text/plain"
```
dmadisetti, line 87: NB
 > Deprecated since version 3.13: Passing a file path instead of URL is [soft deprecated](https://docs.python.org/3/glossary.html#term-soft-deprecated). Use [guess_file_type()](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_file_type) for this.
 
Fine here, but I think we use it in a few places.

dmadisetti (approved):
Good catch! looks like this has been in there for a year

frontend/src/core/static/files.ts
mscolnick: its not related but i noticed it when trying to test out this change. basically the patch doesn't work from opening a local html file. (only when served from http(s))

marimo/_server/export/exporter.py
mscolnick: guess_file_type says added in 3.13

PR #4648, opened 2025-04-24
feat: toggle between markdown formats and set r-string as default

### LINKED ISSUE #3895: Ability to use f-string in the markdown editor
### Discussed in https://github.com/marimo-team/marimo/discussions/3884

<sup>Originally posted by **Warning39** February 22, 2025</sup>
I made the following comments in an issue - https://github.com/marimo-team/marimo/issues/3762#issuecomment-2663958927 that I logged earlier.

There is a bit of extra friction with that limitation. I've noticed that the markdown editor in marimo only works with `r` strings and not `f` or `rf` strings. Is there a reason for that limitation?

So instead of being able to type this into the markdown editor:

```
### Some Heading

<img src="public/SolarEnergyDistribution.png" width="600"/>

Some formula $F = ma$ 
```

I have to edit via the code editor as opposed to the markdown editor, i.e.

```
mo.md(rf"""
### Some Heading

{mo.image("public/SolarEnergyDistribution.png", width=600)}

Some formula $F = ma$
""")
```
@mscolnick replied with:

> Regarding the mo.md() string formatted, there we edge cases such as lack of auto-completion of the python code when parameterizing the markdown editor. We can maybe revisit this to add support for this, but for now this needs to happen in Python.

How about an option to override this in the GUI on a per-cell basis. If I have a cell with a lot of markdown and just need the odd bit of python code in it like the example above to call `mo.image()` then I'd be happy in this case to miss out python auto-completion in this cell.</div>

### LINKED ISSUE #4573: Convert to markdown on empty cell should use r-string
### Describe the bug

When writing markdown cells with backslashes, it doesn't escape the backslashes:

Image

as it is converted to this python code:

Image

Converting it to raw strings would fix this (although I'm not sure how it would interact with other features).

It can also be fixed by adding a backslash, but it can be confusing for beginners to understand why only sometimes two backslashes are needed:

Image

### Environment

<details>

```
{
  "marimo": "0.12.10",
  "OS": "Darwin",
  "OS Version": "24.4.0",
  "Processor": "arm",
  "Python Version": "3.13.2",
  "Binaries": {
    "Browser": "135.0.7049.86",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.6",
    "narwhals": "1.33.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.13.0",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "pandas": "2.2.3",
    "pycrdt": "0.11.1",
    "python-lsp-ruff": "2.2.2",
    "python-lsp-server": "1.12.2",
    "ruff": "0.11.4"
  },
  "Experimental Flags": {
    "lsp": true,
    "multi_column": true
  }
}
```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.12.10"
app = marimo.App()


@app.cell
def _(mo):
    mo.md("""$\frac{1}{2}$""")
    return


if __name__ == "__main__":
    app.run()
```

Light2Dark:
## 📝 Summary

Fixes #4573 . Defaults to using r-string.

Fixes #3895, #3884 . Introduce a toggle in markdown cells that can flip between raw string, f-string and raw + f-string.
image

## 🔍 Description of Changes

- possible to use Select, but I felt this was cleaner. If we want to add a plain option, maybe change to a select.
- small lint fix for expressions in while loop (https://biomejs.dev/linter/rules/no-assign-in-expressions/)

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/core/codemirror/language/__tests__/markdown.test.ts
+++ b/frontend/src/core/codemirror/language/__tests__/markdown.test.ts
@@ -180,6 +180,14 @@ describe("MarkdownLanguageAdapter", () => {
       expect(offset).toBe(9);
     });
 
+    it("defaults to r-string when there is no last quote prefix", () => {
+      const adapter = new MarkdownLanguageAdapter();
+      const code = "Hello world";
+      const [wrappedCode, offset] = adapter.transformOut(code);
+      expect(wrappedCode).toBe(`mo.md(r"""Hello world""")`);
+      expect(offset).toBe(10);
+    });
+
     it("single line", () => {
       const code = "Hello world";
       adapter.lastQuotePrefix = "";
--- a/frontend/src/core/codemirror/language/__tests__/markdown.test.ts
+++ b/frontend/src/core/codemirror/language/__tests__/markdown.test.ts
@@ -1,6 +1,7 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 import { expect, describe, it } from "vitest";
 import { MarkdownLanguageAdapter } from "../markdown";
+import { getQuotePrefix } from "../panel";
 
 const adapter = new MarkdownLanguageAdapter();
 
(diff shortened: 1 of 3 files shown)


akshayka:
- Instead of cycling through the options, which fields a little difficult to navigate, why not two checkboxes, one for `r` and one for `f`?
- I do think `f` strings should continue to use the Python view by default, until we get completion in for Python variables. That might be what you planned but just thought I'd mention.

mscolnick:
bullet two will be mostly solved by #4657

akshayka:
Let me know when this is ready for a closer look; I'd like to play with it before we merge.

mscolnick:
@akshayka - i agree, starting it in Python but showing the toggle would be at the very least. 

right now we do have:

- code formatting
- ~LSP/jedi completion~ custom variable completion (with datatype and value). this has a tradeoff of better completion when thinking about templating, but harder write things like polars transforms inside. I can demo the trade-offs live.
- syntax highlighting

since these are two separate branches, going to merge, but we do either before releasing:

1. for f-string, revert to starting in python and show the toggle
2. use jedi completions instead of variable suggestions
3. keep as-is

PR #4657, opened 2025-04-24
improvement: add python variable completions inside SQL {}

mscolnick:
This adds python variable completion inside SQL `{}` blocks. 
This also adds it markdown blocks as well when using `f-strings` although not exposed yet (until #4648).


frontend/src/core/codemirror/language/embedded-python.ts
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Cheap variable completion source without having to go through the
+ * backend.
+ */
+export const variableCompletionSource = (
+  context: CompletionContext,
+): CompletionResult | null => {
+  // Check if we're inside a {} block by looking for an opening brace
+  const beforeCursor = context.state.doc.sliceString(0, context.pos);
+  const lastOpenBrace = beforeCursor.lastIndexOf("{");
```
akshayka, line 139: For an `f` string, two consecutive braces inserts a literal `{`, which should not trigger completion, eg:

image

Light2Dark (dismissed):

Light2Dark:
the completions weren't getting updated when a local table updated so i made a change for that. 

Do you think we could show this better? It should show value/attribs of the variable after the dot
image

mscolnick:
@Light2Dark i think I can get completions on variable in a follow up. 

I didn't reuse the existing python completions because it returns a lot of garbage like Python globals. I think I can tweak it though.

Light2Dark (approved):

PR #4670, opened 2025-04-25
improvement: better formats for table and charts. avoid additional round trip

mscolnick:
This improves table performance by avoiding additional round trips (sending file URLs and loading when only 10 rows), and uses better formats in different areas.

1. remove `to_data` as this indirection often made us choose the wrong format 
2. pass back json data for each paging, instead of a URL of a json data, to avoid a round trip
3. make the column charts use `arrow` to reduce memory usage

Now
* First table load: json
* Subsequent table load: json
* Table column summaries: arrow
* Table chart tab: arrow


Light2Dark:
looks like it works now!

Light2Dark (approved):

Light2Dark (approved):

akshayka (approved):

PR #4685, opened 2025-04-27
feat: enables sandboxing for markdown files

### LINKED ISSUE #2369: Preserved shebangs / header comments in marimo md documents
### Description

I think with the new sandbox change, editing does not strip the shebangs- which is great. Now this works out of the box, and is very nice:

`cat notebook.py`
```python
#!/usr/bin/env -S marimo export html --no-include-code
import marimo
# ...
```

`chmod +x notebook.py`

`./notebook.py`

---

You can push this further and can now create an environment with nix:

```python
#! /usr/bin/env nix
#! nix shell --impure --expr ``
#! nix with (import (builtins.getFlake ''nixpkgs'') {});
#! nix    let
#! nix        venv = python311.withPackages (
#! nix            ps: with ps; [ marimo matplotlib numpy ]
#! nix        );
#! nix    in pkgs.buildEnv {
#! nix        name = "marimo-env";
#! nix        # NB adding ffmpeg, openmpi and gcc as a nix shell dep.
#! nix        paths=[venv ffmpeg gcc-unwrapped.lib openmpi];
#! nix    } 
#! nix ``
#! nix --command marimo run

import marimo
# ...
```

Note, you can even combine this with sandbox to pin your python and UV to nix:

```python
#! /usr/bin/env nix
#! nix shell --impure --expr ``
#! nix with (import (builtins.getFlake ''nixpkgs'') {});
#! nix    python311.withPackages (
#! nix            ps: with ps; [ marimo uv setuptools ]
#! nix    )
#! nix ``
#! nix --command marimo edit --sandbox
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "numpy",
# ]
# ///

import marimo
# ...
```

In theory, you could even have nix build or launch an actual container

### Suggested solution

This already works, so could just be an undocumented feature. However, I'm thinking of preserving `#! nix` and `script` sandboxes in the markdown format (e.g):

```python
---
title: Notebook
marimo-version: 0.0.0
width: medium
nix: |
    with import <nixpkgs> {};
    python311.withPackages (
        ps: with ps; [ marimo uv jaxlibWithCuda ]
    )
---   
```

(and or `script: |`) So having the documentation for the normal case would be nice too.

I am also very conscious that I am full on with the nix koolaid :grimacing:, so documenting and then potentially having to retain support might not be appealing.

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #4680: Export as qmd
### Description

same as md but should insert the marimo filter
also ensure that other headers are preserved

### Suggested solution

-

### Alternative

_No response_

### Additional context

_No response_

dmadisetti:
## 📝 Summary

 - preserves unknown frontmatter entries (e.g. `description`, `tag`, or `author`)
 - saves header values in `header` entry (fixes #2369)
 - Allows sand boxing from `header` entries OR a barebones `sandbox` entry
 - Conversion to qmd adds the needed quarto filter (fixes #4680)

needs:
 - maybe some cleanup?
 - tests:
   - [x] Conversion logic
   - [x] Sandboxing logic


marimo/_cli/convert/markdown.py
```diff
@@ -232,11 +222,16 @@ def _tree_to_app(root: Element) -> str:
         )
         sources.append(get_source_from_tag(child))
 
+    header = root.get("header", None)
+    sandbox = root.get("sandbox", None)
```
mscolnick, line 226: `pyproject` or `script_metadata` might be a better name, since you can include config in the sandbox as well (that get used outside the sandbox)

dmadisetti: Agreed

dmadisetti:
Hmm, failing tests were from #4710 so feel free to retrigger CI once that gets in

PR #4707, opened 2025-04-28
fix: use proxy scheme instead of server

mscolnick:


dmadisetti (approved):
Nice adding a test, I remember this same block coming up a few times

marimo/_server/api/middleware.py
```diff
@@ -360,9 +360,10 @@ async def __call__(
             if self.path_rewrite:
                 ws_path = self.path_rewrite(ws_path)
             ws_url = urljoin(ws_target_url, ws_path)
-            if scope["scheme"] in ("http", "ws"):
+            scheme = urlparse(ws_url).scheme
+            if scheme in ("http", "ws"):
```
spasche, line 364: If the target scheme is already ws or wss, I don't think it needs to be replaced?

Also, that might lead to an unexpected replacement with a target such as `ws://example.com/http` which would be replaced to `ws://example.com/ws`.

mscolnick: you are right. i will fix this

PR #4720, opened 2025-04-29
feat: Exposes types if they are explicitly declared by the user

dmadisetti:
## 📝 Summary

```python
myvar: int = 0
```

```python
@app.cell
def foo(myvar: int) -> tuple[int]:
    defvar: int = 1
    return (defvar,)
```

on reference or definition, will carry its annotation in the signature

couple other notes:
 - I also tried putting in `-> None` for blank returns, but it looks like a pretty noisy change.
 - I'm not a huge fan of the return type because an actual call to the function doesn't return those values.
 - Return signature only occurs when all the variables are signed
 - types which cannot be properly resolved are coerced to strings in signatures

@akshayka


dmadisetti:
or

Maybe we define the type alias:
```python
Edges: TypeAlias = tuple[Any, ...] | ActualReturnType
```

```python
@app.cell
def foo(myvar: int) -> marimo.Edges:
    ...
```

I think the ActualReturnType is HTML output right now

akshayka (commented):
The signature for arguments looks good! Nice surgical change, too.

If you are unsure about the return signature, how about we only add types for arguments right now?

dmadisetti:
Thanks! So call just returns the value from run which gets processed by mime, but maybe ideally it should be the raw value (I'd argue this for run as well)

the alias would then maybe be better as

```python
marimo.Output[type] # are generics supported in 3.9?
```

But I don't think return type are all that important. For some reason, linters don't ask for return types on decorated pytests- I wonder what their decorators do that suppresses it

akshayka (approved):
Looks great — not sure what's going on with CI though?

PR #4739, opened 2025-04-30
fix: invalid f-string in markdown mode breaks

dmadisetti:
## 📝 Summary

Removed callout code (I never implemented this in the format, and now we have native admonitions)
+ add syntax error catch for parse (export runtime error). Added a unit test because catching the actual behavior is a dynamic thing

@akshayka OR @mscolnick


mscolnick (approved):

dmadisetti:
Woops sorry CI failed. Maybe it is worth having YOLO permissions

mscolnick (approved):

PR #4743, opened 2025-04-30
fix: per-block reference resolution

### LINKED ISSUE #4742: [ast] missing ref with nested scopes
### Describe the bug

The following cell should have `x` as a ref, but doesn't.

```python
def f():
    print(x)
    def g():
        def h():
            x
        # oops, this will evict x as a ref
        x = 0
```

### Environment

<details>

```
0.13.2
```

</details>

### Code to reproduce

_No response_

akshayka:
For each instance of a referenced name, we need to track the blocks it appears in;
previously, references on new blocks (scopes) were evicting
references on other scopes. These newer references could be
resolved but the other ones weren't, leading to us drop references.

Fixes #4742

### DIFF
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -124,6 +124,8 @@ class RefData:
 
     # Whether the ref was deleted
     deleted: bool
+    # Block in which this ref was referenced
+    block: Block
     # Ancestors of the block in which this ref was used
     parent_blocks: list[Block]
 
@@ -157,7 +159,10 @@ class ScopedVisitor(ast.NodeVisitor):
         self.ref_stack: list[set[Name]] = [set()]
         self.obscured_scope_stack: list[ObscuredScope] = []
         # Mapping from referenced names to their metadata
-        self._refs: dict[Name, RefData] = {}
+        # Each referenced name may be defined multiple blocks;
+        # ie, self._refs maps each name to RefData for each block it was
+        # defined in
+        self._refs: dict[Name, list[RefData]] = {}
         # Function (node, name, block stack) -> None
         self._on_def = on_def if on_def is not None else lambda *_: None
         # Function (node) -> None
@@ -189,7 +194,11 @@ class ScopedVisitor(ast.NodeVisitor):
     @property
     def deleted_refs(self) -> set[Name]:
         """Referenced names that were deleted with `del`."""
-        return set(name for name in self._refs if self._refs[name].deleted)
+        return set(
+            name
+            for name in self._refs
+            if any(ref.deleted for ref in self._refs[name])
+        )
 
     def _if_local_then_mangle(
         self, name: str, ignore_scope: bool = False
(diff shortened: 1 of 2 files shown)


marimo/_ast/visitor.py
```diff
@@ -189,7 +194,11 @@ def refs(self) -> set[Name]:
     @property
     def deleted_refs(self) -> set[Name]:
         """Referenced names that were deleted with `del`."""
-        return set(name for name in self._refs if self._refs[name].deleted)
+        return set(
+            name
+            for name in self._refs
+            if any(ref.deleted for ref in self._refs[name])
```
dmadisetti, line 200: Hmm, I think maybe only the top block?

```python
x = 0
def fn():
    if False:
        z = x
        del x
x
```

dmadisetti: I think only possible with global keyword

akshayka: I think looking at all the blocks is right, because the `_refs` data structure keeps one `RefData` per name per block.

In your example, `x` is not added as a ref, so there's no issue.

I did however find a related bug, and just pushed up a fix and a couple of unit tests.

tests/_ast/test_visitor.py
```diff
@@ -972,6 +972,44 @@ def h():
...
+    v.visit(mod)
+    assert v.defs == {"x", "fn"}
+    assert not v.refs
+    assert not v.deleted_refs
+
+
+def test_deleted_ref() -> None:
+    code = cleandoc(
+        """
+    def fn():
```
dmadisetti, line 998: I think this is only true when there's a global, otherwise del pushes the block to be in `scope`.

Maybe another bug when playing around but:

```python
x = 1
def f():
    global x
```

refs = {'x'}

but seems like refs = {}

but honestly, I wouldn't be opposed to just banning global

akshayka: > I think this is only true when there's a global, otherwise del pushes the block to be in scope.

I don't understand what you mean. This test passes; is there another test that should pass but doesn't?

akshayka: (Pushed a fix for the global test case.)

dmadisetti: The del case is the same as:

```python
x = 1
```

```python
def g():
    z = x
    x = 0
```

Which if you actually try to run gives you a NameError, because python- detecting the assign without global and scopes x.

We're find all the bugs because currently refs == {"x"} for this case - which is not really how this behaves, it should be refs == {}.

Likewise in the del case, I think refs == {}, deleted_refs == {} (since it's local effectively).

This changes when a global is present though.

```python
def g():
    global x
    z = x
    x = 0
```
refs == {"x"}

```python
def g():
    global x
    z = x
    del
```
refs == {"x"}, deleted_refs == {"x"}

akshayka:
Pushed a couple other `global` fixes. If there are other issues with `global`, let's handle those in a different PR/issue (unless they are caused by the change in this PR).

tests/_ast/test_visitor.py
akshayka: Oh got it! I see what you're saying. Thanks.

akshayka:
Converting to a draft, may not have time to get to this in the immediate future.

mscolnick (approved):

dmadisetti (approved):
Awesome. The bug has probably always existed and no one has noticed, so I think xfail is fine.

Starting to get in the weeds of "what's possible but very messy code"

PR #4778, opened 2025-05-01
improvement: add css underline to traceback links

mscolnick:
- add css underline to traceback links
- remove community forum link (soon to be offline)

### DIFF
--- a/frontend/src/components/editor/output/Outputs.css
+++ b/frontend/src/components/editor/output/Outputs.css
@@ -114,5 +114,5 @@
 }
 
 .traceback-cell-link:hover {
-    font-weight: 500;
+  font-weight: 500;
 }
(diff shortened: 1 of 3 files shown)


frontend/src/components/editor/output/Outputs.css
```diff
@@ -110,4 +110,5 @@
 
 .traceback-cell-link {
   color: inherit;
+  text-decoration: underline;
```
dmadisetti, line 113:
```python
.traceback-cell-link:hover {
    font-weight: bold;
}
```

Or maybe just 500, I think bold is a little aggressive

frontend/src/components/editor/output/Outputs.css
```diff
@@ -110,4 +110,5 @@
...
 }
```
dmadisetti, line 114:
```suggestion
  text-decoration: underline;
}

.traceback-cell-link:hover {
    font-weight: 500;
}
```

mscolnick: This feels weird, does is not change the width of everything?

mscolnick: 500 does look good actually

PR #4779, opened 2025-05-01
feat: highlight lines when cell has a traceback

mscolnick:
image

### DIFF
--- a/frontend/src/core/cells/cells.ts
+++ b/frontend/src/core/cells/cells.ts
@@ -14,11 +14,7 @@ import {
   focusAndScrollCellIntoView,
 } from "./scrollCellIntoView";
 import { CellId } from "./ids";
-import {
-  outputToTraceback,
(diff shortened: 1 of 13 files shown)


frontend/src/core/cells/cells.ts
```diff
@@ -1495,6 +1496,46 @@ export function flattenTopLevelNotebookCells(
   );
 }
 
+export function createTacebackInfoAtom(
```
dmadisetti, line 1499: Nice catch from Copilot
```suggestion
export function createTracebackInfoAtom(
```

Light2Dark (approved):

dmadisetti (commented):
Super cool, couple notes:
 - Text selection in red lines is not visible
 - modifying the cell and inserting lines above keeps shifts the red lines to blanks (Maybe we just get rid of them on change, or shift the lines and get rid of them on line change?)

PR #4782, opened 2025-05-02
feat: mo.thread lifecycle

### LINKED ISSUE #4564: mo.Threads are not cleaned up

akshayka:
Closes #4564 

This change adds a `should_exit` property to `mo.Thread` which evaluates to `True`
when the thread's spawning cell lifecycle has ended. This makes it
possible for developers to clean up their threads.

This change also adds a utility method, mo.current_thread(), that
returns the current thread of control if it is a marimo.Thread object,
and raises otherwise. Developers could technically use
threading.current_thread() instead, but this method provides
type safety and validation.

### DIFF
--- a/marimo/_runtime/threads.py
+++ b/marimo/_runtime/threads.py
@@ -1,7 +1,6 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
-import functools
 import threading
 from typing import Any
 
--- a/marimo/_runtime/threads.py
+++ b/marimo/_runtime/threads.py
@@ -1,10 +1,12 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
+import functools
 import threading
 from typing import Any
 
 from marimo._messaging.streams import ThreadSafeStream
+from marimo._runtime.cell_lifecycle_item import CellLifecycleItem
 from marimo._runtime.context.kernel_context import KernelRuntimeContext
 from marimo._runtime.context.script_context import ScriptRuntimeContext
 from marimo._runtime.context.types import (
@@ -21,7 +23,7 @@ THREADS = set()
 
 
 class Thread(threading.Thread):
-    """A Thread subclass that is aware of marimo internals.
+    """A Thread subclass that can communicate with the frontend.
 
     `mo.Thread` has the same API as threading.Thread,
     but `mo.Thread`s are able to communicate with the marimo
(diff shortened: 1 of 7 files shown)


mscolnick (approved):

marimo/_runtime/threads.py
```diff
@@ -33,17 +35,56 @@ class Thread(threading.Thread):
...
+        exit_event = threading.Event()
+        self._exit_event = exit_event
 
         if not runtime_context_installed():
             return
 
+        class ThreadLifecycle(CellLifecycleItem):
+            def create(self, context: RuntimeContext) -> None:
+                del context
+                pass
```
dmadisetti, line 77: nit pass

marimo/_runtime/threads.py
```diff
@@ -33,17 +35,56 @@ class Thread(threading.Thread):
...
+
+            def dispose(self, context: RuntimeContext, deletion: bool) -> bool:
+                del context
+                del deletion
+                exit_event.set()
+                return True
+
         ctx = get_context()
+        ctx.cell_lifecycle_registry.add(ThreadLifecycle())
```
dmadisetti, line 86: Nice 👍🏾
Much better than the manual incorrect clean up in #4571

marimo/_runtime/threads.py
dmadisetti, line 77:
```suggestion
```

dmadisetti (approved):
Nice! Docs look good too

marimo/_runtime/threads.py
dmadisetti: see below

dmadisetti (approved):
Great change.

I wonder if we are under utilizing the Lifecycle mechanism. UI Registry and state registry could both probably have less boiler plate by registering themselves

PR #4783, opened 2025-05-02
chore: cleanup executor, prefer composition

mscolnick:
Prefer composition. This will make re-using/mixing executors (e.g. CachingExecutor) much easier.


dmadisetti (approved):
Nice bit of cleanup! Doesn't get rid of the async/ sync duplication - but good start

PR #4785, opened 2025-05-02
fix: Uses SQL tables if available for script calls

### LINKED ISSUE #4675: Fileformat creates lint errors with sql
### Describe the bug

Image

### Environment

Vim, pyright

Rendering from 0.13.2

### Code to reproduce

Maybe a viable solution is:

```python
@app.cell
def _(graphemes):
    emoji_freq = mo.sql(
        f"""
        select
            count(grapheme) as total,
            grapheme
        from
            graphemes
        group by
            grapheme
        order by
            total
        """,
        used_tables=[graphemes]
    )
    return emoji_freq  # regardless
```

dmadisetti:
Also fixes some lint errors
fixes #4675


tests/_sql/test_sql.py
```diff
@@ -360,3 +360,18 @@ def test_sql_with_ibis_expression_result():
     with patch.object(IbisEngine, "sql_output_format", return_value="native"):
         result = sql("SELECT * FROM test", engine=duckdb_backend)
         assert isinstance(result, Expr)
+
+
+@pytest.mark.skipif(not HAS_PANDAS and HAS_DUCKDB, reason="pandas is required")
```
mscolnick, line 365:
```suggestion
@pytest.mark.skipif(not HAS_PANDAS and not HAS_DUCKDB, reason="pandas is required")
```

marimo/_sql/utils.py
```diff
@@ -31,16 +32,28 @@ def wrapped_sql(
...
     except ContextNotInitializedError:
-        relation = connection.sql(query=query)
-    else:
+        pass
+
+    try:
         relation = eval(
             "connection.sql(query=query)",
-            ctx.globals,
+            tables,
```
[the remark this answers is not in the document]
dmadisetti: Woops. Good catch

Light2Dark:
just tested and it feels good!

dmadisetti:
Actually copilot just created a bug

frontend/src/core/codemirror/language/sql.ts
```diff
@@ -131,8 +131,29 @@ export class SQLLanguageAdapter implements LanguageAdapter {
...
+      .join(",\n        ");
+    // Table params only valid in DuckDB
+    const tablesParam =
+      this.engine === this.defaultEngine && tablesString
+        ? `,\n    tables={\n        ${tablesString}\n    }`
+        : "";
+
+    const end = `\n    """${showOutputParam}${engineParam}${tablesParam}\n)`;
+
+    // TODO: Ruff-wasm is now more main stream (adopted by jupyter-ruff)
```
mscolnick, line 154: this manual formatting is quite easy (and gets reformatted with black in the backend). im not sure this is a TODO we want to do for the extra dependency

dmadisetti: Noticed that the line length varies the behavior (ruff rather fit everything on one line if it can). I think the manual formatting may eventually be more effort than it is worth

marimo/_sql/utils.py
```diff
@@ -31,19 +32,46 @@ def wrapped_sql(
...
+        tables = {}
+
+    previous_globals = {}
     try:
         ctx = get_context()
+        previous_globals = ctx.globals.copy()
+        ctx.globals.update(tables)
+        tables = ctx.globals
     except ContextNotInitializedError:
-        relation = connection.sql(query=query)
```
Light2Dark, line 37: Small thing, looks like it defaults to `connection.sql` when there is no context.  The suggested change would only allow manual eval?

Light2Dark (approved):

dmadisetti:
@Light2Dark thoughts on

```python
@app.cell
def _(graphemes):
    emoji_freq = mo.sql(
        f"""
        select
            count(grapheme) as total,
            grapheme
        from
            graphemes
        group by
            grapheme
        order by
            total
        """
    )
    return (emoji_freq, graphemes) # return the table passed in to make it valid 
```

Light2Dark:
> @Light2Dark thoughts on
> 
> ```python
> @app.cell
> def _(graphemes):
>     emoji_freq = mo.sql(
>         f"""
>         select
>             count(grapheme) as total,
>             grapheme
>         from
>             graphemes
>         group by
>             grapheme
>         order by
>             total
>         """
>     )
>     return (emoji_freq, graphemes) # return the table passed in to make it valid 
> ```

looks like a bug no? to return graphemes

dmadisetti:
Merge conflict, potential revisit, but unlikely

PR #4787, opened 2025-05-02
feat: row expanded view for table data

Light2Dark:
## 📝 Summary

Popover that displays single row data vertically. By default, it will show all rows in current page. If there is selection, it will show those rows instead.

image

image

Some limitations:
- The data is limited to whatever page the table is on. Does not persist selections across pages.
- When you navigate through the page and select/deselect, the index on the page stays the same, meaning you could view different data as you select more rows. We don't keep track of the selected row id as the page number.

Todo:
- [ ] add tanstack table feature (focus row)
- [ ] tests

## 📋 Checklist

- [X] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).


frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -358,6 +364,9 @@ export const LoadingDataTableComponent = memo(
       }
     }, [props.pageSize, paginationState.pageSize]);
 
+    const containerRef = useRef<HTMLDivElement>(null);
+    const cellId = useFindCellId(containerRef);
```
mscolnick, line 368: the plugin gets a prop called `hostElement`. i wonder if passing that down would be easier, and remove the `setTimeout`. (the `setTimeout` feels like something is wrong)

frontend/src/components/editor/chrome/panels/data-selection.tsx
```diff
@@ -0,0 +1,9 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+
+import { DataSelectionPanel as SelectionPanel } from "@/components/data-table/selection-panel/data-selection";
+
+export const DataSelectionPanel: React.FC<{
+  handleDragging: (isDragging: boolean) => void;
+}> = ({ handleDragging }) => {
+  return <SelectionPanel handleDragging={handleDragging} />;
```
mscolnick, line 8: on thing we could look at is making this a "Slot". i created a library `@marimo-team/react-slotz` which we use in the codebase already - you can look for some existing usage.

this would allow the plugins to do:

<PortalCellContext>
 {some_custom_ui}
</PortalCellContext>

and then ` {some_custom_ui}` can be picked up in this sidebar

frontend/src/components/data-table/selection-panel/data-selection.tsx
```diff
@@ -0,0 +1,239 @@
...
+  TableHeader,
+  TableRow,
+} from "@/components/ui/table";
+import type { Cell } from "@tanstack/react-table";
+import { DATA_TYPE_ICON } from "@/components/datasets/icons";
+import { Input } from "@/components/ui/input";
+import { Tooltip } from "@/components/ui/tooltip";
+import { renderCellValue } from "../columns";
+
+export const DataSelectionPanel: React.FC<{
```
mscolnick, line 38: i think we can split this up between the content and the resize logic

frontend/src/components/editor/chrome/panels/data-selection.tsx
Light2Dark: really cool approach, thanks for the rec!

frontend/src/components/editor/chrome/panels/selection-panel.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export const DataSelectionPanelSlot: React.FC = () => {
+  return <Slot name={SlotNames.DATA_SELECTION} />;
+};
+
+export const DataSelectionItem: React.FC<PropsWithChildren> = ({
+  children,
+}) => {
+  return <Fill name={SlotNames.DATA_SELECTION}>{children}</Fill>;
+};
```
mscolnick, line 15: this could probaly be a generic: `<CellAwarePanel/> or <ContextAwarePanel/>`

frontend/src/components/editor/chrome/panels/selection-panel.tsx
```diff
@@ -0,0 +1,15 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+
+import { SlotNames } from "@/core/slots/slots";
+import { Fill, Slot } from "@marimo-team/react-slotz";
+import type { PropsWithChildren } from "react";
+
+export const DataSelectionPanelSlot: React.FC = () => {
+  return <Slot name={SlotNames.DATA_SELECTION} />;
```
mscolnick, line 8: if we want to render the pinning and sidebar logic here, instead can do `useSlot()`. and check if the list is empty to just render `null`

frontend/src/components/data-table/TableActions.tsx
```diff
@@ -42,6 +45,8 @@ export const TableActions = <TData,>({
   getRowIds,
   toggleDisplayHeader,
   chartsFeatureEnabled,
+  toggleSelectionPanel,
+  isOwnerOfContextAwarePanel,
```
mscolnick, line 49: This is leaky of a name. Maybe just isSelectionPanelToggled

frontend/src/components/data-table/data-table.tsx
```diff
@@ -206,7 +219,7 @@ const DataTableInternal = <TData,>({
         )}
         <Table>
           {renderTableHeader(table)}
-          {renderTableBody(table, columns)}
+          {renderTableBody(table, columns, isSelectionPanelOpen)}
```
Light2Dark, line 222: not sure if this is ideal vs adding some state with the new feature.

mscolnick: I think this is fine

frontend/src/components/editor/chrome/panels/context-aware-panel.tsx
```diff
@@ -0,0 +1,124 @@
...
+      </Panel>
+    </>
+  );
+};
+
+export const ContextAwarePanelItem: React.FC<PropsWithChildren> = ({
+  children,
+}) => {
+  return (
+    <ErrorBoundary>
```
Light2Dark, line 97: added an error boundary because there's a strange bug when navigating certain rich elements (although it's not consistent). Discovered when navigating rich_elements.py
```
Uncaught Error: [handleFillUpdated] component was expected to be defined
    at Manager.handleFillUpdated (index.mjs:89:13)
    at index.ts:109:7
    at Array.map (<anonymous>)
    at Object.emit (index.ts:108:7)
    at SlotzController.update (index.mjs:181:14)
```

Light2Dark: adding an error boundary still causes the panel to close, but at least doesn't affect the table.

frontend/src/components/data-table/focus-row/types.ts
```diff
@@ -0,0 +1,32 @@
...
+// define types for feature's table options
+export interface FocusRowOptions {
+  enableFocusRow?: boolean;
+  onFocusRowChange?: OnChangeFn<FocusRowState>;
+}
+
+// Define types for feature's table APIs
+export interface FocusRowInstance {
+  focusRow: (rowIdx: FocusRowState) => void;
+  getFocusedRowIdx: () => FocusRowState;
```
mscolnick, line 21: These need to be optional since the feature is optional. Otherwise it might break in the variable sidebar

frontend/src/components/data-table/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This is the row ID that is used to identify a row in the table.
+ * It is stable across renders and pagination. It may not exist.
+ *
+ */
+export function getStableRowId<TData>(row: TData): string | undefined {
+  if (row && typeof row === "object" && INDEX_COLUMN_NAME in row) {
+    return String(row[INDEX_COLUMN_NAME]);
+  }
+}
```
mscolnick, line 46: These are great bits of cleanup!

frontend/src/components/editor/chrome/panels/context-aware-panel.tsx
mscolnick: This is a good call to add the boundary anyways. Happy to leave for now. I can look at the bug - might be in the slotz library.

frontend/src/hooks/useResizeHandle.ts
```diff
@@ -80,7 +91,7 @@ export const useResizeHandle = ({
       document.removeEventListener("mousemove", handleMouseMove);
       document.removeEventListener("mouseup", handleMouseUp);
     };
-  }, [onResize]);
+  }, [maxWidth, minWidth, onResize]);
```
mscolnick, line 94: I don't think this effect should rerun each time on the width changes, since they are adding and removing listeners. Instead maybe you could do 

`useEvent` to pull out the handlers, and then pass the handlers as deps here

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -664,6 +692,25 @@ const DataTableComponent = ({
...
+      // Close if panel is open
+      setSelectionPanelOwner(null);
+    } else {
+      // Set the owner to current id, which will cause the panel to open
+      setSelectionPanelOwner(id);
+    }
+  }
+
+  const handleFocusRowChange: OnChangeFn<number> = useEvent((updater) => {
+    setFocusedRowIdx(updater);
```
mscolnick, line 711: Can likely kill this function and inline setFocusedRowIdx

Light2Dark: do you mean just do this? Since we would need to pass the state down otherwise
```typescript
<DataTable
    ....
    onFocusRowChange={(rowIdx) => setFocusedRowIdx(rowIdx)}
```

mscolnick: just `onFocusRowChange={setFocusedRowIdx(rowIdx}`

mscolnick (approved):

mscolnick (approved):

PR #4776, opened 2025-05-01
fix: cleanup ipython display updates

mscolnick:
Cleanup always expanding dictionary

PR #4788, opened 2025-05-02
Fix weakvaluedict issue

brenoca:
## 📝 Summary

@mscolnick, unfortunately, the WeakValueDictionary broke the iPython update implementation
Now it only outputs the initial empty bar, as it can be seen in the video below

https://github.com/user-attachments/assets/2d591719-119f-44fd-9a25-e788540b3faf

## 🔍 Description of Changes

Reverting WeakValueDictionary change while keeping the cleanup functions.

Analysis and issues
- noticed the dictionary adds one record per cell display
- known issue, the iPython is not unpatching, therefore the dictionary is never cleared

 Here is a short video where I analyse the current behaviour

https://www.loom.com/share/16a282de56724584ac0c49bc81a68539

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_output/formatters/ipython_formatters.py
+++ b/marimo/_output/formatters/ipython_formatters.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 import functools
-import weakref
 from dataclasses import dataclass
 from typing import Any, Callable, Optional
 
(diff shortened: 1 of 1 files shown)


dmadisetti:
I think not cleaning up is fine until we actually see performance issues.
I did propose a potential fix here, since there might be a race condition with GC: https://github.com/marimo-team/marimo/pull/4776#discussion_r2070709118 if we need to revisit

dmadisetti:
Thanks @brenoca ! You also only have to do the CLA thing once

PR #4789, opened 2025-05-02
improvement: refactor cache Store, add RESTStore

mscolnick:


marimo/_save/stores/redis.py
```diff
@@ -22,13 +15,14 @@ def __init__(self, **kwargs: Any) -> None:
         # See list of options here:
         self.redis = redis.Redis(**kwargs)
 
-    def get(self, key: HashKey, loader: Loader) -> Optional[bytes]:
-        return self.redis.get(str(loader.build_path(key))) or None
+    def get(self, key: str) -> Optional[bytes]:
+        result = self.redis.get(key)
+        if result is None:
+            return None
+        return result.encode("utf-8")  # type: ignore[no-any-return]
```
dmadisetti, line 22: is encode right here? Not sure what typing is suggesting, I thought `.get` returned bytes already

dmadisetti (commented):
Awesome cleanup. Weird before- I think code debt of Store abstraction coming after Loader. Really nice `path` api usage too

Want to try this out, but quick note is that HashKey was replaced for string. Should be back online at 1PST if you can wait

marimo/_save/stores/redis.py
mscolnick: get returns a string. i wasn't sure what the right encoding was. do you know?

pyproject.toml
```diff
@@ -527,6 +527,10 @@ manager = "uv"
 [tool.marimo.keymap]
 vimrc = "configs/.vimrc"
 
+[tool.marimo.experimental.cache]
+store = "rest"
```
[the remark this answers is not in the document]
mscolnick: this is not the default, and may change in the API, so i would exclude from the docs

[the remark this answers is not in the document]
mscolnick: actually this breaks tests too i think, so going to remove for us

dmadisetti (approved):
Barring tests, LGTM- I think complexity was added prematurely and this is a great simplification

tests/_save/store/test_store.py
```diff
@@ -22,4 +22,6 @@ def _(mo, store):
                 a = 1
                 b = 2
 
-            assert store._cache[cache._cache.key.hash]
+            key = f"mock/P_{cache._cache.key.hash}.pickle"
```
dmadisetti, line 25: It might just be this

PR #4806, opened 2025-05-04
improvement: use script metadata for config and deps in pyodide

mscolnick:
This adds support for inline-script metadata for pyodide.

If there are dependencies declared inline, then we use those. Otherwise we fallback to the pyodide parsing of imports (which was what we did previously).

This also parses inline script metadata for config.

--------------

This looks like more lines of code, but mostly:
1. lots of tests written and backfilled
2. some utils moved around


marimo/_pyodide/pyodide_session.py
```diff
@@ -158,14 +160,44 @@ def find_packages(self, code: str) -> list[str]:
...
-            return []
+        # Prefer dependencies from script metadata
+        try:
+            reader = PyProjectReader.from_script(code)
+            script_deps = reader.dependencies
+
+            def strip_version(dep: str) -> str:
+                try:
+                    return (
+                        dep.split("==")[0]
```
dmadisetti, line 172: re.split(r"[><=~]=", dep)[0]

dmadisetti (commented):
Nice cleanup. I think we can remove tomlkit from pyproject too then?

mscolnick:
@dmadisetti tomlkit is needed for writing toml, which we use quite a bit. tomllib just supports reading

tests/_pyodide/test_pyodide_session.py
```diff
@@ -92,44 +92,27 @@ def mock_pyodide() -> Generator[MagicMock, None, None]:
...
 @pytest.fixture
 def pyodide_session(
-    app_file_manager: AppFileManager,
     mock_on_write: Callable[[KernelMessage], None],
-    app_metadata: AppMetadata,
+    pyodide_app_file: Path,
 ) -> PyodideSession:
     return PyodideSession(
-        app=app_file_manager,
+        app=AppFileManager(filename=str(pyodide_app_file)),
```
dmadisetti, line 101: Oh tricky, may worth leaving a comment somewhere? I guess we both know now, but definitely not an immediately obvious one

marimo/_output/formatters/formatters.py
```diff
@@ -149,7 +152,13 @@ def find_spec(  # type:ignore[no-untyped-def]
             original_find_spec=original_find_spec,
         ) -> Any:
             del self
-            spec = original_find_spec(fullname, path, target)
+
+            try:
+                spec = original_find_spec(fullname, path, target)
+            except Exception as e:
+                LOGGER.warning(f"Error finding spec for {fullname}: {e}")
```
dmadisetti, line 159: I see, I guess the question is why is this getting patched in in the first place?

PR #4816, opened 2025-05-05
feat: Post mortem for debugging failures

dmadisetti:
## 📝 Summary

Opposed to inserting a breakpoint, drop into "post mortem mode"
Played around a bit before deciding this was. the best way. LMK if you have feedback on the api namespace

Adding tests:
 - [x] exceptions correctly set on failure in cell_runner

@Light2Dark @mscolnick


frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -115,14 +117,11 @@ export const MarimoTracebackOutput = ({
             size="xs"
```
mscolnick, line 117: we could add `{lastTracebackInfo && !isWasm() &&`, since i don't think pdb is well supported. i will need to do this on the traceback line-number feature too

dmadisetti: Good call. Done! Captured on the line num too

Light2Dark:
Do you think we could have both the breakpoint and hook debugger? 🤔.. I can see why we want to prefer one, if so, we could add to the docs! I think breakpoints are useful for debugging issues which do not raise exceptions.

I think it'd be helpful to have a tooltip (that maybe points to a link in docs) explaining how debugging works. I was quite confused initially. Or we could put in the Get Help dropdown.

dmadisetti:
@Light2Dark we do have both because you can click the little bug icons- the button just mapped to the last one previously.

But I do think a tooltip is useful

Light2Dark:
I meant one button for 'Hook Debugger' and another for 'Add Breakpoint'.

mscolnick:
im happy with just the "add breakpoint" being at the line number. the larger button i added was more of a placeholder. and postmortem is better in most cases

frontend/package.json
```diff
@@ -44,6 +44,7 @@
     "@lezer/python": "^1.1.17",
     "@marimo-team/codemirror-ai": "^0.1.11",
     "@marimo-team/codemirror-languageserver": "^1.15.14",
+    "@marimo-team/frontend": "link:",
```
mscolnick, line 47: This seems incorrect. This is itself

marimo/_runtime/runtime.py
```diff
@@ -1512,6 +1513,20 @@ async def _run_with_uninstantiated_requests(
...
+            self.debugger is None
+            or cell_id not in self.debugger._last_tracebacks
+            or cell_id not in self.graph.cells
+        ):
+            return
+
+        with self._install_execution_context(cell_id):
+            self.debugger.post_mortem_by_cell_id(cell_id)
+            # entry = compile('__import__("pbd").pm()', filename, 'single')
+            # eval(entry, self.globals)
```
mscolnick, line 1528: Remove?

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -110,19 +113,16 @@ export const MarimoTracebackOutput = ({
...
-              const view = getCellEditorView(lastTracebackInfo.cellId);
-              if (view) {
-                insertDebuggerAtLine(view, lastTracebackInfo.lineNumber);
-              }
+              sendPdb({ cellId: tracebackInfo.cellId });
             }}
           >
             <BugPlayIcon className="h-3 w-3 mr-2" />
-            Insert breakpoint
+            Hook debugger
```
mscolnick, line 125: Maybe "Re-run with debugger"

mscolnick (approved):
looks great!

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
dmadisetti: Hmm, it it's actually rerunning though, just putting back where the failure happened.

maybe: "Launch debugger" ?

mscolnick: Yea maybe Launch debugger is better then

dmadisetti:
Couple things I noticed:
 -  When debugger is open, the notebook is locked. I see this as a trap that users might fall into.
 - Starting interactive mode (`interact`) (creates python repl), and then exiting it, locks the kernel

juanmpga:
Love this! To me post mortem is much more convenient than setting breakpoints. Usually I'm not writing functionality myself but using someone else's and the bug is due to inputs that look fine but are actually incorrect, then it's often much easier to see what's failing when looking at how variables look like at the line where the code crashed.

dmadisetti:
@mscolnick Cool if I merge before the next release?

PR #4820, opened 2025-05-06
feat: add both left and right resize handlers to a column.

Light2Dark:
## 📝 Summary

I think resizing a column from the left also solves some issues with viewport so it's preferred imo.
The left line is only displayed when hovered. Not sure what the right styles are

image

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/hooks/useResizeHandle.ts
+++ b/frontend/src/hooks/useResizeHandle.ts
@@ -23,7 +23,7 @@ export const useResizeHandle = ({
   useEffect(() => {
     const resizableDiv = resizableDivRef.current;
 
-    if (!resizableDiv) {
+    if (!resizableDiv || (!leftHandle && !rightHandle)) {
       return;
     }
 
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/useResizeHandle.ts
```diff
@@ -4,69 +4,90 @@ import { useEffect, useRef } from "react";
 interface UseResizeHandleProps {
   onResize?: (width: number) => void;
   startingWidth: number | "contentWidth";
-  direction: "left" | "right";
 }
 
+// Currently supports left and right resizing
 export const useResizeHandle = ({
```
mscolnick, line 10: i wonder if this could have been left as-is, can you just create 2 when you use it:

```
const left = useResizeHandle({direction: "left"})
const right = useResizeHandle({direction: "right"})
```

mscolnick (approved):

frontend/src/hooks/useResizeHandle.ts
Light2Dark: Because useResizeHandle returns a resizableDivRef and style class
We would need to merge them in the child components I think, so instead I opted for the other approach.
```
const { resizableDivRef, handleRef, style } = useResizeHandle({...})
```

But i haven't explored merging the refs/styles approach.

mscolnick: Ah that makes sense

PR #4832, opened 2025-05-06
fix: File level cache settings are not passed to store

dmadisetti:
## 📝 Summary

@mscolnick Still compat with pyodide right?


mscolnick (approved):
yep this should be find for pyodide.

PR #4840, opened 2025-05-07
fix: support mairmo edit on marimo.app

mscolnick:
Suport for `marimo edit "https://marimo.app/e?slug=oitgic"`

### DIFF
--- a/marimo/_cli/file_path.py
+++ b/marimo/_cli/file_path.py
@@ -11,7 +11,7 @@ from tempfile import TemporaryDirectory
 from typing import Optional
 from urllib.error import HTTPError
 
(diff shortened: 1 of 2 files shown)


marimo/_cli/file_path.py
```diff
@@ -126,24 +126,35 @@ def download(url: str) -> tuple[bool, str]:
         if url.startswith("https://static.marimo.app/"):
             return download(os.path.join(url, "download"))
```
dmadisetti, line 127:
```suggestion
            return download(url)
```

static.marimo.app/e?slug=wxyz

not sure what the /download endpoint is

PR #4845, opened 2025-05-08
chore: make LanguageAdapters stateless, store metadata in EditorState

mscolnick:
This makes the `LanguageAdapter` stateless and instead adds `LanguageAdapterMetadata` which is a map of state that is easier to access/update. 

This change makes real-time syncing much easier. 

This also adds some new test converge and cleans up some code (moving to to new files)

### DIFF
--- a/frontend/src/core/codemirror/language/utils/quotes.ts
+++ b/frontend/src/core/codemirror/language/utils/quotes.ts
@@ -1,7 +1,5 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
-import { logNever } from "@/utils/assertNever";
-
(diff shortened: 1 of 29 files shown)


frontend/src/core/codemirror/language/utils/quotes.ts
```diff
@@ -18,30 +16,3 @@ export function splitQuotePrefix(quote: string): [QuotePrefixKind, string] {
   }
   return ["", quote];
 }
-
-export function upgradePrefixKind(
```
mscolnick, line 22: unused/undesired since we have explicit toggles now

PR #4847, opened 2025-05-08
fix: ibis data types, from upgrade

mscolnick:
Newer version of ibis has changed the expected datatype response. Also handles some more cases in type conversion.


dmadisetti (dismissed):
LGTM, surprised there isn't a more general check. Lots of names for strigns

PR #4855, opened 2025-05-08
fix: Restart cell in pdb

dmadisetti:
## 📝 Summary

Still doing a bit of a deep dive into PDB. Worth actually trying to use it and making sure we aren't serving up a broken experience. restart/run failed dramatically- but now actually restart

I think pretty formatting for PDB can be done FE- but like we said probably not worth it just yet.

@mscolnick

### DIFF
--- a/marimo/_server/sessions.py
+++ b/marimo/_server/sessions.py
@@ -920,7 +920,7 @@ class SessionManager:
                     UpdateCellCodes(
                         cell_ids=cell_ids,
                         codes=codes,
-                        code_is_stale=not should_autorun,
+                        code_is_stale=True,
                     ),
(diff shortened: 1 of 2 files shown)


marimo/_server/sessions.py
```diff
@@ -920,7 +920,7 @@ async def on_file_changed(path: Path) -> None:
                     UpdateCellCodes(
                         cell_ids=cell_ids,
                         codes=codes,
-                        code_is_stale=not should_autorun,
+                        code_is_stale=True,
```
mscolnick, line 923: Ah it's always true

mscolnick (approved):

PR #4858, opened 2025-05-09
fix: don't block on background tasks created in cells

### LINKED ISSUE #3780: datashader rasterize not supported?
### Describe the bug

Would Marimo be able to support DataShader's rasterise()?

In Jupyter, plots with millions of points can be rasterized before being sent to the browser .

E.g. https://holoviews.org/user_guide/Large_Data.html when converted to Marimo, the rasterized plots are not re-rendered when zoomed in.

### Environment

<details>

```
```

</details>

### Code to reproduce

https://holoviews.org/user_guide/Large_Data.html
https://raw.githubusercontent.com/holoviz/holoviews/main/examples/user_guide/15-Large_Data.ipynb

mscolnick:
Fixes #3780 

This gives other async tasks a chance to execute while the control request is empty. For example, we create our own background tasks (e.g. writing session metadata) and sometimes a cell can as well (holoviews does, for example in the datashader)

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -2716,9 +2716,22 @@ def launch_kernel(
     ui_element_request_mgr = SetUIElementRequestManager(set_ui_element_queue)
 
     async def control_loop(kernel: Kernel) -> None:
+        from queue import Empty
+
         while True:
             try:
-                request: ControlRequest | None = control_queue.get()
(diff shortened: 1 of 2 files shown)


dmadisetti (approved):
I think this is OK

Sorry, was taking a second to think through alternatives. But yes, blocking for task based async

PR #4863, opened 2025-05-09
improvement: allow configuration the location of the web static assets

mscolnick:


marimo/_utils/requests.py
```diff
@@ -0,0 +1,144 @@
+import http.client
```
mscolnick, line 1: this might be useful over time with the other fetching we do without `requests` or `urlib3`

mscolnick:
possible with `marimo edit --asset-url=`

PR #4870, opened 2025-05-11
feat: added input field to ui slider

### LINKED ISSUE #4747: UI slider with optional text input field for value
### Description

It would be useful to have a slider which also allows the user to directly type in a precise value (e.g. like `mo.ui.number` does).

A slider is very convenient to quickly explore a certain range of values, but using it to input a precise value, especially when the step is small, is impractical. 
On the other hand, a  `mo.ui.number` allows the user to directly type in a precise number, but lacks the convenience of the slider when trying to rapidly span a wide range of values without any need for precision.

There are many cases in which it would be convenient to have both ways of providing a numerical input value combined in the same UI element. This way the slider could still be used for a preliminary exploration of the parameter range, while also allowing a direct, exact input when required.

### Suggested solution

Provide a way to allow the slider's value to be directly typed in by the user.

For example, add the optional argument `editable_value` next to `show_value` in the `marimo.ui.slider` initializer:
```
slider(
    start: Optional[Numeric] = None,
    stop: Optional[Numeric] = None,
    [...]
    show_value: bool = False,
    editable_value: bool = False,
    [...]
)
```
When a `slider` is constructed with `show_value=True` and `editable_value=True`, the value shown to the right of the slider is a text input, allowing the user to directly type in a value, in addition to using the slider.
(in this solution, I wouldn't know how to best handle the case of `editable_value=True` with  `show_value=False`: you could either decide that `editable_value=True` implies that the value is shown regardless of `show_value`, or raise an exception if `show_value!=editable_value`

I've just started out using marimo and I do not know enough about its internals to assess the feasibility and complexity of implementing this proposed feature.

### Alternative

A somewhat similar result can be obtained by having both a  `mo.ui.slider` and a `mo.ui.number` and synchronizing them using `mo.state`. 
However this approach is a bit messy and cumbersome, especially with many inputs in the same notebook or when multiple such inputs are programmatically created in a `mo.ui.array`. It also wastes UI screen space with duplicated input elements. 

### Additional context

_No response_

bjoaquinc:
## 📝 Summary

Integrated number-field into the slider and added a include_input argument to manage visibility of input. Resolves #4747 by adding the requested feature.

Screenshot 2025-05-11 at 12 32 38 PM

## 🔍 Description of Changes

- Added includeInput to SliderPlugin.tsx with validation
- Added include_input to input.py along with a description
- Integrated NumberField into SliderPlugin
- Added a two way data binding between the slider and the integrated number-field

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/frontend/src/plugins/impl/SliderPlugin.tsx
+++ b/frontend/src/plugins/impl/SliderPlugin.tsx
@@ -7,6 +7,7 @@ import { Slider } from "../../components/ui/slider";
 import { Labeled } from "./common/labeled";
 import { cn } from "@/utils/cn";
 import { prettyScientificNumber } from "@/utils/numbers";
+import { NumberField } from "@/components/ui/number-field";
 
 type T = number;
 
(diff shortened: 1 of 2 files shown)


mscolnick (approved):
Clean implementation!

PR #4888, opened 2025-05-12
improvement: generate altair code from chart spec

mscolnick:
Generate Altair code from a Vega-lite spec. And python-poet utilities


Light2Dark (approved):

Light2Dark:
would you want to merge this, since it's not wired in yet. I can hook up some UI and hide it under dev mode until it's in a more ready state.

image

frontend/src/utils/python-poet/poet.ts
```diff
@@ -33,15 +33,69 @@ export class Variable implements PythonCode {
 }
 
 export class Literal implements PythonCode {
-  constructor(public value: unknown) {}
+  constructor(
+    public value: unknown,
+    public objectAsFieldNames = false,
+    public removeNull = false,
+    public removeUndefined = true,
```
mscolnick, line 40: we should change this to:

```
    public value: unknown,
    public opts: {objectAsFieldNames?: boolean, removeNull?: boolean, removeUndefined?: boolean}
```

it improves readability and otherwise it will be very easy to mix them up

PR #4889, opened 2025-05-12
improvement: pass df name if exists to table plugin

mscolnick:
not sure if this is a good or correct approach. 

this helps with code-generating the altair code on the frontend with the correct datasource

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import functools
+import inspect
 from dataclasses import dataclass
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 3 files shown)


marimo/_plugins/ui/_impl/table.py
```diff
@@ -421,6 +422,21 @@ def __init__(
         self._lazy = _internal_lazy
         self._page_size = page_size
 
+        # HACK: this is a hack to get the name of the variable that was passed in
+        dataframe_name = '"<data>"'
```
dmadisetti, line 426: Have this issue with UI and state, but use a post execution hook in case there is reassignment etc...

This does work though, so

mscolnick:
not going with this approach

PR #4891, opened 2025-05-12
improvement: add runtime.default_sql_output

mscolnick:


dmadisetti (commented):
Are there docs for what native means? / What auto behavior is?

mscolnick:
@dmadisetti, yes there are https://docs.marimo.io/guides/working_with_data/sql/?h=sql+output+types#sql-output-types

PR #4895, opened 2025-05-13
fix: patch polars write_json in wasm with write_csv

mscolnick:
Patch polars to fallback to `write_csv` when `write_json` fails in wasm.

### DIFF
--- a/tests/_runtime/test_patches.py
+++ b/tests/_runtime/test_patches.py
@@ -1,10 +1,22 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
+import io
+from types import ModuleType
+from typing import TYPE_CHECKING
+from unittest.mock import patch
+
+import pytest
+
+from marimo._dependencies.dependencies import DependencyManager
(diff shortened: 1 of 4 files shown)


tests/_runtime/test_patches.py
```diff
@@ -175,3 +187,59 @@ async def test_webbrowser_easter_egg(
...
+)
+def test_polars_write_json_patch(tmp_path: Path):
+    import sys
+
+    import polars as pl
+
+    file_path = tmp_path / "test.json"
+
+    # Fake put pyodide in sys.modules
+    sys.modules["pyodide"] = ModuleType("pyodide")
```
akshayka, line 204: Should this be in a `try`, with the unpatch in a `finally`, so that this test failing doesn't side-effect to other tests?

akshayka (approved):

tests/_runtime/test_patches.py
mscolnick: i can put this in a decorator or fixture. there are a few places this exists

PR #4899, opened 2025-05-13
fix: row viewer panel for many columns

Light2Dark:
## 📝 Summary

Previously the number of cols were clamped hence not all columns showed up in the panel.

The workaround is to unclamp during search, and also pass all field types and clamp the field types on the frontend.

Alternative: Create a new function to get all field types, so don't need to slice in the frontend.

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


Light2Dark:
> Does this clamp column summaries (min/max/etc) and column chart data?

yeah, `table.py` still clamps chart_data. If I don't clamp on the frontend, this shows up. So, the data & summaries are clamped but field types are unclamped.

image

PR #4913, opened 2025-05-13
CI build improvements

schlich:
## 📝 Summary

- Adds linux-64 to pixi config (was mac only)
- Adds a GH Actions workflow to build frontend & backend using pixi
- Extends max memory allocation for frontend to 4 GB

## 🔍 Description of Changes

Pixi is a good idea for providing a more consistent development environment.  Adding linux to the pixi config worked right off the bat.  If the code ever gets to a point where mac and linux need different build dependencies, that is configurable through pixi.

The Github Action required a pixi.lock file. I think it's a good idea to add this anyway so that devs have a reproducible environment. Sometimes lockfiles require a little extra attention to git workflows and maintenance to avoid merge conflicts, but it's worth the tradeoff in my experience (usually just stashing or deleting them when pulling changes works fine).

All of the above paragraph also applies for the pnpm lockfile which was also added.

The mac image hit an out-of-memory error; this is also an error i have consistently run into on my local linux environment FWIW. I extended the max memory allocated to node by setting an environment variable in the package.json build script configuration.  Worked like a charm.  Maybe the frontend devs might want to see what the largest memory consumers are here?

Regarding the new workflow: i'd be happy to merge the build into existing workflows (such as test or release workflows) but figured i'd get some feedback first; but in general i think a there's a lot of room consolidation of development environment configurations both in github actions and in the local configuration setup... happy to put in more work on that with the go-ahead from the marimo team, and also happy to put my thoughts down in more detail on a separate issue.

## 📋 Checklist

## 📜 Reviewers

@dmadisetti @mscolnick

### DIFF
--- a/.github/workflows/dev_build.yaml
+++ b/.github/workflows/dev_build.yaml
@@ -6,8 +6,18 @@ on:
       - test/*
 
 jobs:
-  build:
+  build-linux:
     runs-on: ubuntu-latest
+    steps:
+     - uses: actions/checkout@v4
+     - uses: prefix-dev/setup-pixi@v0.8.3
+       with:
+         pixi-version: v0.47.0
+         cache: true
+     - run: pixi run make fe 
+     - run: pixi run make py
+  build-mac:
+    runs-on: macos-latest
     steps:
      - uses: actions/checkout@v4
      - uses: prefix-dev/setup-pixi@v0.8.3
--- a/.github/workflows/dev_build.yaml
+++ b/.github/workflows/dev_build.yaml
@@ -1,0 +1,16 @@
+name: Dev Build
+
+on:
+  push:
+    branches:
+      - test/*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+     - uses: prefix-dev/setup-pixi@v0.8.3
+       with:
+         pixi-version: v0.47.0
+         cache: true
+     - run: pixi install
--- a/.github/workflows/dev_build.yaml
+++ b/.github/workflows/dev_build.yaml
@@ -1,9 +1,9 @@
 name: Dev Build
 
 on:
-  push:
+  pull_request:
     branches:
-      - test/*
+      - main
 
 jobs:
   build:
--- a/.github/workflows/dev_build.yaml
+++ b/.github/workflows/dev_build.yaml
@@ -9,7 +9,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-     - uses: action/checkout@v4
+     - uses: actions/checkout@v4
      - uses: prefix-dev/setup-pixi@v0.8.3
        with:
          pixi-version: v0.47.0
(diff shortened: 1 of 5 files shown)


schlich:
It looks like there are already synchronization issues with the lock file, and it might be prudent to expand the usage of pixi for CI builds.  I'll go ahead and do that for this PR but happy to try a different approach if this is undesirable.

.github/workflows/dev_build.yaml
```diff
@@ -0,0 +1,28 @@
+name: Dev Build
+
+on:
+  push:
+    branches:
+      - test/*
+
+jobs:
+  build-linux:
```
mscolnick, line 9: you could do this as a `matrix` https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

frontend/pnpm-lock.yaml
```diff
@@ -4,11 +4,11 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-pnpmfileChecksum: sha256-QrioRE0rVa7k37ba9jgFOwNQflzbUTpdxTezotowvIE=
+pnpmfileChecksum: eqzj7c6t3bs5v7xcjer2wfwkp4
```
mscolnick, line 7: this change might be b/c of a different pnpm version. you could do `make check-prereqs`

schlich: yep, i'm already on it.  updating to pnpm 10 and node 22 (as the release workflow specifies) fixes the issue.

.github/workflows/dev_build.yaml
```diff
@@ -6,18 +6,11 @@ on:
       - test/*
```
mscolnick, line 6: does this run when we want? maybe?

```
on:
  push:
    paths:
      - 'tests/'
```

mscolnick: maybe since this doesn't need to run a lot, it could be:

```
on:
  push:
    paths:
      - pyproject.toml
      - Makefile
```

schlich: It was so i could run the actions on my personal repo before opening the pull request, was planning on switching before opening the PR but forgot, thanks for the reminder.

dmadisetti: Hmm, maybe just `pyproject` or `package.json`

schlich:
@mscolnick do you know if this is a known issue?  seems unrelated to these changes

https://github.com/marimo-team/marimo/actions/runs/15006759600/job/42167333687

mscolnick:
@schlich that should be fixed on main if you merge `main` back into your branch

dmadisetti:
CI is breakage is different (though I thought the fix got merged earlier)
Cool! I'll have to use pixi more. Hatch seems to a good job for standardizing tests, but build versions for everything can't hurt!

dmadisetti (dismissed):

dmadisetti (approved):
main is fixed so lgtm
I do think targeting specifically pyproject, Make, and package.json make sense- but the extra cycles probably aren't that more expensive

PR #4915, opened 2025-05-13
fix: put Inspector in typecheck

mscolnick:
A small patch that would help with: https://github.com/marimo-team/marimo/discussions/4849

### DIFF
--- a/marimo/_utils/health.py
+++ b/marimo/_utils/health.py
@@ -177,4 +177,4 @@ def communicate_with_timeout(
         return process.communicate(timeout=timeout)
     except subprocess.TimeoutExpired:
         process.kill()
-        return "Error: Process timed out", ""
+        return "", "Error: Process timed out"
--- a/marimo/_utils/health.py
+++ b/marimo/_utils/health.py
@@ -10,6 +10,8 @@ from marimo import _loggers
 
 LOGGER = _loggers.marimo_logger()
 
+TIMEOUT = 10  # seconds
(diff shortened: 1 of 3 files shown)


marimo/_utils/health.py
```diff
@@ -19,7 +21,7 @@ def get_node_version() -> Optional[str]:
             stderr=subprocess.PIPE,
             text=True,
         )
-        stdout, stderr = process.communicate()
+        stdout, stderr = process.communicate(timeout=TIMEOUT)
```
dmadisetti, line 24: Should probably catch the error:

proc = subprocess.Popen(...)
try:
    outs, errs = proc.communicate(timeout=15)
except TimeoutExpired:
    proc.kill()
    return None

marimo/_utils/health.py
```diff
@@ -166,3 +168,13 @@ def get_chrome_version_linux() -> Optional[str]:
...
+
+
+def communicate_with_timeout(
+    process: subprocess.Popen[str], timeout: float = TIMEOUT
+) -> tuple[str, str]:
+    try:
+        return process.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        process.kill()
+        return "Error: Process timed out", ""
```
dmadisetti, line 180: On stderr

```suggestion
        return "", "Error: Process timed out"
```

dmadisetti (dismissed):

PR #4920, opened 2025-05-14
fix: Fix execution hash regression 

dmadisetti:
@akshayka

should traverse all parents starting from ref definition.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -120,11 +120,17 @@ class _SetupContext:
             )
         self._frame = with_frame
         self._previous = {**with_frame.f_locals}
+        # Python 3.13+ has different behavior for f_locals
+        if not hasattr(self._frame.f_locals, "clear"):
+            # Since we statically parse this is fine, but may cause issues in
+            # script runs.
+            return
         # A reference to the key app must be maintained in the frame.
         # This may be a python quirk, so just remove refs to explicit defs
-        for var in self._cell.defs:
-            if var in self._previous:
-                del self._frame.f_locals[var]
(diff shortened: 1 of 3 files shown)


mscolnick (dismissed):

dmadisetti:
Moving back to draft to indicate transitive state isn't working 100%, and the test should be redone, because the line number hack is silly

dmadisetti:
This does not fix stateful execution path hashes. A more substantial fix cache busts in a big way- so I think this is a good stop gap for now

akshayka:
@dmadisetti does that mean this would still yield a false positive cache hit on the embedding example I shared?

dmadisetti:
It'll break if it uses an execution path that requires state / ui- writing up a little plan rn

marimo/_ast/app.py
```diff
@@ -120,11 +120,17 @@ def __enter__(self) -> None:
...
+            return
         # A reference to the key app must be maintained in the frame.
         # This may be a python quirk, so just remove refs to explicit defs
-        for var in self._cell.defs:
-            if var in self._previous:
-                del self._frame.f_locals[var]
+        self._frame.f_locals.clear()
+        for p in self._previous:
+            if p not in self._cell.defs:
+                self._previous[p] = self._previous[p]
```
akshayka, line 133: Sorry, not sure if I am reading this wrong — why assign `self._previous[p]` to itself?

dmadisetti: You are right. Reverting and breaking this out into a separate issue .

akshayka (approved):

PR #4928, opened 2025-05-14
tests: Capture invalid setup cell behavior

dmadisetti:


akshayka:
Discussed with @dmadisetti — no need to guard against referencing variables outside the setup block, it's a tiny discrepancy that is only hit when running as a script.

dmadisetti:
@akshayka Also removes the other `sys._getframe` reference

marimo/_ast/app.py
```diff
@@ -135,24 +134,15 @@ def __exit__(
         if exception is not None:
             # Always should fail, since static loading still allows bad apps to
             # load.
+            # But don't record the variables.
             return False
 
-        # Manually hold on to defs for injection into script mode.
+        # Use exec to update the locals properly
```
dmadisetti, line 140:
```suggestion
```

dmadisetti:
Failing test is valid, and maybe a good reason to keep _some_ of the old behavior.

mscolnick (approved):

PR #4947, opened 2025-05-15
improvement: python poet cleanup

mscolnick:
Some cleanup just branching off the main branch

- removes `objectAsFieldNames` since this does actually work for nested object. but we have this logic, just needed to re-use as `args` in the `FunctionCall`
-  removes `objectsToPythonCode`. i added `Literal` to handling `toCode` so we could compose this a bit better.

### DIFF
--- a/frontend/src/utils/python-poet/poet.ts
+++ b/frontend/src/utils/python-poet/poet.ts
@@ -33,24 +33,23 @@ export class Variable implements PythonCode {
 }
 
 interface LiteralOptions {
-  objectAsFieldNames?: boolean;
   removeNull?: boolean;
   removeUndefined?: boolean;
 }
 
 export class Literal implements PythonCode {
   constructor(
-    public value: unknown,
-    public opts: LiteralOptions = {
-      objectAsFieldNames: false,
-      removeNull: false,
-      removeUndefined: true,
-    },
+    public readonly value: unknown,
+    public readonly opts: LiteralOptions = {},
   ) {}
+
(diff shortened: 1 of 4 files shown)


frontend/src/utils/python-poet/poet.ts
```diff
@@ -71,13 +70,18 @@ export class Literal implements PythonCode {
...
+    }
+
     if (Array.isArray(this.value)) {
       if (this.value.length === 0) {
         return "[]";
       }
       return `[\n${indent(
         this.value
-          .map((item) => new Literal(item).toCode())
+          .map((item) => new Literal(item, this.opts).toCode())
```
mscolnick, line 84: needed to pass down options if we recurse

frontend/src/utils/python-poet/poet.ts
```diff
@@ -33,24 +33,23 @@ export class Variable implements PythonCode {
...
   ) {}
 
+  static from(value: unknown, opts: LiteralOptions = {}): Literal {
+    return new Literal(value, opts);
+  }
+
   toCode(): string {
     const EMPTY_VALUE = "";
-    const { objectAsFieldNames, removeNull, removeUndefined } = this.opts;
+    const { removeNull = false, removeUndefined = true } = this.opts;
```
mscolnick, line 52: this way properly using adds the defaults. previously we would overwrite other defaults

Light2Dark (approved):

PR #4948, opened 2025-05-15
fix: ignore click 8.2.0

mscolnick:
There is a fix foward but requires more LOC. This ignores the bad version and `8.2.1` should have the fix

Related: https://github.com/pallets/click/issues/2897


mscolnick:
It is odd our CI did not catch this. Maybe missing a few tests.

dmadisetti (approved):

PR #4970, opened 2025-05-18
feat: Add unique transform to `mo.ui.dataframe`

### LINKED ISSUE #3738: New dataframe transform: Unique
### Description

I would like to be able to get the unique values of a column

### Suggested solution

Using the distinct operator in sql or the unique operators in both polars and pandas should do the trick

### Alternative

_No response_

### Additional context

_No response_

wasimsandhu:
## 📝 Summary

Closes #3738.

### Frontend

- Add icon to `mo.ui.dataframe` transform panel.
- Define `UniqueTransformSchema` with four options for `keep` parameter (first, last, any, none) in `frontend/src/plugins/impl/data-frames/schema.ts`.
  - The only odd one out is `any`, which is only for Polars dataframes.

### Backend

- Create handler and types for `UniqueTransform` in `marimo/_plugins/ui/_impl/dataframes/transforms`.
- Implement unique transform for pandas, polars, and ibis dataframes.
- Write Python statement for each dataframe type in `print_code.py`.

## Driveby Changes

- Removed `frontend/src/plugins/impl/data-frames/python` as it appears to be unused, but please let me know if I'm mistaken.
- Fix Pylint error about using `if` instead of `elif` for all `if` blocks that return in `dataframes/apply.py`.
  - I disabled Pylint somewhere along the way but left this one in anyway. Happy to revert.

## Tests

- Wrote small unit tests for unique transform handler using all 4 `keep` options.
- Updated snapshots.
- Caveat:
  - Since the result of a `unique` transform appears to be non-deterministic for polars and ibis dataframes, it seems that I couldn't easily extend the tests for `print_code.py`. However, I did include a notebook in the smoke tests in which I manually verified that the python print code can be run and appears to return the same number of rows as the plugin.

## Notes

- I can't figure out why the frontend build job is failing. All frontend checks and tests are passing in my env, and `cd frontend; pnpm install --no-frozen-lockfile` makes no changes to the lockfile.
- The unique transform result appears to be non-deterministic for polars and ibis. There is a `maintain_order` argument that we could include for polars dataframes, but not sure if there is one for ibis.

## 📋 Checklist

## 📜 Reviewers

Not sure who to tag here so I'll put @mscolnick.


frontend/src/plugins/impl/data-frames/python/python-print.ts
```diff
@@ -1,194 +0,0 @@
-/* Copyright 2024 Marimo. All rights reserved. */
```
mscolnick, line 1: i guess this is on the backend now?

frontend/pnpm-lock.yaml
```diff
@@ -4,11 +4,11 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-pnpmfileChecksum: sha256-QrioRE0rVa7k37ba9jgFOwNQflzbUTpdxTezotowvIE=
```
mscolnick, line 7: you will need to upgrade pnpm to the latest version (v10, I think)

mscolnick (approved):
looks great! i think just need to revert the pnpm change

mscolnick (approved):

PR #4972, opened 2025-05-18
fix: support multiple anywidget widget models

### LINKED ISSUE #3909: Lonboard (anywidget) map not rendering layers
### Describe the bug

[Lonboard](https://developmentseed.org/lonboard/latest/) map visualization not working as expected.
Layers aren't rendered.

### Environment

<details>

```json
{
  "marimo": "0.11.9",
  "OS": "Darwin",
  "OS Version": "22.6.0",
  "Processor": "arm",
  "Python Version": "3.11.7",
  "Binaries": {
    "Browser": "133.0.6943.127",
    "Node": "v20.17.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.28.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "ruff": "0.9.7",
    "starlette": "0.46.0",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0"
  },
  "Optional Dependencies": {
    "anywidget": "0.9.14",
    "duckdb": "1.2.0",
    "pandas": "2.1.4",
    "polars": "1.23.0",
    "pyarrow": "19.0.1"
  },
  "Experimental Flags": {
    "chat_sidebar": true,
    "multi_column": true
  }
}
```

</details>

### Code to reproduce

More details in #3908 

The expected map should look like:

Image

But I am getting:

Image

mscolnick:
Fixes #3909 

This implements additional sub-models that anywidget supports, that may not necessarily be tied to a widget.


frontend/src/plugins/impl/anywidget/model.ts
```diff
@@ -8,14 +14,60 @@ import { z } from "zod";
...
+
+    // Add timeout to prevent hanging
+    const timeout = new Promise<Model<any>>((_, reject) => {
+      setTimeout(() => {
+        reject(new Error(`Model not found for key: ${key}`));
+        this.models.delete(key);
+      }, this.timeout);
+    });
+
+    return Promise.race([deferred.promise, timeout]);
```
manzt, line 40: Thinking this could just reject the `Deferred` directly, unless I'm missing something?

```ts
setTimeout(() => {
  deferred.reject(new Error(`Model not found for key: ${key}`));
  this.models.delete(key);
}, this.timeout);

return deferred.promise;
```

As an aside, I've been using `AbortSignal` for stuff like this recently since there is a dedicated "timeout" and signals themselves compose. Not suggesting we use in this PR, but something to be aware of potentially for making "abortable" APIs.

```ts
AbortSignal.timeout(this.timeout).addEventListener("abort", () => {
  deferred.reject(new Error(`Model not found for key: ${key}`));
  this.models.delete(key); 
});
```

marimo/_plugins/ui/_impl/anywidget/types.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
```
manzt, line 1: This is great to have, thanks

marimo/_smoke_tests/anywidget_examples/lonboard_example.py
```diff
@@ -0,0 +1,100 @@
...
+
+        _tile_geometries = shapely.from_wkt(_df["tile"])
+        _tile_centroids = shapely.centroid(_tile_geometries)
+        gdf = gpd.GeoDataFrame(
+            _df[["avg_d_kbps"]], geometry=_tile_centroids, crs="EPSG:4326"
+        )
+        gdf.to_parquet(_local_path)
+
+    gdf
+    return (gdf,)
```
manzt, line 66: This is a pretty large file. Not sure how often this runs, but we could just inline a small subset of the dataframe to avoid downloading the large dataset.

marimo/_plugins/ui/_impl/anywidget/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    DependencyManager.ipywidgets.require("for anywidget support.")
+    import ipywidgets  # type: ignore
+
+    _remove_buffers = ipywidgets.widgets.widget._remove_buffers  # type: ignore
+
+    # Get the initial state of the widget
+    state, buffer_paths, buffers = _remove_buffers(message)  # type: ignore
+
+    return state, buffer_paths, buffers  # type: ignore
```
manzt, line 31: Not necessarily something to change in this PR, but in general I prefer avoiding blanket `# type: ignore` comments. If possible, I prefer ignoring specific error codes (e.g. `# type: ignore[attr-defined]`) so it’s clearer what mypy is skipping. However, as we've discussed, adding some type of lint rule to disallow this usage (across the codebase) would be the right approach and something outside the scope of this PR.

manzt (dismissed):
Really impressive — tested on a few complex widgets and it's working well from what I can see.

One note: (I believe) this introduces a distinction between widgets with and without views. That _seems_ okay, but it's a shift from how standard anywidget in Jupyter works and may add some complexity/edge cases. Might be worth aligning the model/view lifecycle more closely down the line.

For context, typical `ipywidgets.Widget` (from which `anywidget.AnyWidget` inherits) behavior (regardless of view) looks like:

- Instantiating a widget class immediately creates a model on the frontend (even if not displayed).
- Displaying the widget creates a new view (calls the `render` function) on top of the existing model.
- Multiple views can be created from the same model.
- The model is always headless; rendering happens only in the view layer.

This pattern helps support both viewed and headless widgets consistently, and also using the same model for different views.

marimo/_smoke_tests/anywidget_examples/lonboard_example.py
mscolnick: this never runs. its just a smoke test so only a paper trail to bug fixes

frontend/src/plugins/impl/anywidget/model.ts
mscolnick: good suggestions. i can look into making these changes

mscolnick:
@manzt - yes this could use another refactor to get us closer to every model is headless and any UI can use any model. that was my original misunderstanding of anywidget/comms when I first implemented this.

can definitely do that in a followup

PR #4973, opened 2025-05-19
feat: added disabled prop to switch

bjoaquinc:
## 📝 Summary

Added disabled to switch

Screenshot 2025-05-18 at 10 08 38 PM

## 🔍 Description of Changes

- Added disabled property and cleaned code in SwitchPlugin.tsx
- Mapped disabled in python class with definition

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/frontend/src/plugins/impl/SwitchPlugin.tsx
+++ b/frontend/src/plugins/impl/SwitchPlugin.tsx
@@ -6,17 +6,22 @@ import { Switch } from "@/components/ui/switch";
 import type { IPlugin, IPluginProps } from "@/plugins/types";
 import { Labeled } from "./common/labeled";
 
-export class SwitchPlugin
-  implements IPlugin<boolean, { label: string | null }>
-{
+type T = boolean;
+interface Data {
+  label: string | null;
+  disabled?: boolean;
+}
+
+export class SwitchPlugin implements IPlugin<T, Data> {
   tagName = "marimo-switch";
 
   validator = z.object({
     initialValue: z.boolean(),
     label: z.string().nullable(),
+    disabled: z.boolean().optional(),
   });
 
-  render(props: IPluginProps<boolean, { label: string | null }>): JSX.Element {
+  render(props: IPluginProps<T, Data>): JSX.Element {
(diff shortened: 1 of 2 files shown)


frontend/src/plugins/impl/SwitchPlugin.tsx
```diff
@@ -6,17 +6,22 @@ import { Switch } from "@/components/ui/switch";
 import type { IPlugin, IPluginProps } from "@/plugins/types";
 import { Labeled } from "./common/labeled";
 
-export class SwitchPlugin
-  implements IPlugin<boolean, { label: string | null }>
-{
+type T = boolean;
```
dmadisetti, line 9: Data is great, but I feel like T obfuscates unnecessarily

bjoaquinc: True, and it's unlikely that the value of T will change. Want me to change it back?

frontend/src/plugins/impl/SwitchPlugin.tsx
```diff
@@ -6,17 +6,22 @@ import { Switch } from "@/components/ui/switch";
...
+interface Data {
+  label: string | null;
+  disabled?: boolean;
```
dmadisetti, line 12: Is bool optional? Looks like it is always set in python

dmadisetti (commented):
Thanks! Either or on T

frontend/src/plugins/impl/SwitchPlugin.tsx
bjoaquinc: I based my implementation of this on the existing implementation in TextInputPlugin.tsx which defines it as optional. Technically the value is always set because it has a default value of false.

frontend/src/plugins/impl/SwitchPlugin.tsx
bjoaquinc: I'll keep the T so that it's structurally similar to several of the other Plugin files

mscolnick (approved):

PR #4989, opened 2025-05-20
wip: use narwhals for ibis

mscolnick:


MarcoGorelli:
thanks for trying this out!

i think within 1.5 - 2 months we should have `stable.v2` ready, it may not be a bad idea to wait til then?

(and if there's any other API changes you may have suggestions for, please do raise them so we can make the necessary changes in time)

mscolnick:
@MarcoGorelli yea i think we will wait, but it will definitely clean up some code

mscolnick:
closed by https://github.com/marimo-team/marimo/pull/6612

PR #4994, opened 2025-05-20
improvement: narwhals perf and arg name changes

mscolnick:
Perf: This improves some query logic for narwhals for better batching of expressions.
Migration: This migrates some name changes that are deprecated, but have been out for a while
Wider support: in some cases, we upgrade narwhals (`nw.v1` to `nw_main`) for better support for newer features like writing to csv. This code can be cleaned up with `nw.v2` is put up.


marimo/_plugins/ui/_impl/tables/narwhals_table.py
```diff
@@ -302,120 +284,129 @@ def _get_summary_internal(self, column: str) -> ColumnSummary:
...
-                nulls=col.null_count(),
-                unique=col.n_unique(),
+
+        frame = self.data.lazy()
+        col = nw.col(column)
+        dtype = self.nw_schema[column]
+        units: dict[str, str] = {}
+
+        # Base expressions for all types
+        exprs: dict[str, nw.Expr] = {
```
Light2Dark, line 294: I'm wondering if there's a better data struct to have this be typesafe?

I was thinking this
```
ColumnSummaryKeys = Literal[
    "total",
    "nulls",
    "unique",
    "min",
    "max",
    "mean",
    "median",
    "std",
    "true",
    "false",
    "p5",
    "p25",
    "p75",
    "p95",
]

exprs: dict[ColumnSummaryKeys, nw.Expr]
```

but if it's too redundant, maybe we don't need

mscolnick: this is not a bad idea. i think its ok for now

mscolnick: it should fail a runtime check at least with `ColumnSummary` (and we have test coverage for it)

marimo/_utils/narwhals_utils.py
```diff
@@ -177,3 +173,38 @@ def can_narwhalify_lazyframe(df: Any) -> TypeGuard[Any]:
...
+@overload
+def upgrade_narwhals_df(df: nw.DataFrame[Any]) -> nw.DataFrame[Any]: ...
+
+
+@overload
+def upgrade_narwhals_df(df: nw.LazyFrame[Any]) -> nw.LazyFrame[Any]: ...
+
+
+def upgrade_narwhals_df(
+    df: Union[nw.DataFrame[Any], nw.LazyFrame[Any]],
```
Light2Dark, line 187: just for my understanding, why is this needed? If I'm using the same narwhals version, does it do anything?

Light2Dark, line 187: just for my understanding, why is this needed? If I'm using the same narwhals version, does it do anything?

mscolnick: narwhals aims to be fully backwards compat - https://narwhals-dev.github.io/narwhals/backcompat/
but they put some valuable features we would like on the "main" narwhals API. so there are a few cases i want to upgrade the narwhals API (such as `write_csv`)

once we move to v2 (after they make it, it currently is not there), then we can remove this and use that version.

PR #4995, opened 2025-05-20
feat: copy table values to clipboard, for Excel/Sheets

### LINKED ISSUE #4991: Copy table to Clipboard
### Description

I'd like an option to copy an entire table (not just the visible rows) to the clipboard, for easier pasting into e.g. Google Sheets. The current alternative is to download and upload a CSV, but that is more time-consuming and cumbersome.

### Suggested solution

Add "Copy to Clipboard" in the Download dropdown in the bottom right of a table alongside the current CSV, JSON, Parquet options.

### Will you submit a PR?

- [ ] Yes

### Alternative

_No response_

### Additional context

_No response_

mscolnick:
Closes #4991 

Copy table in various way, mainly to copy into excel or google sheets

Screenshot 2025-05-20 at 10 52 41 AM

### DIFF
--- a/frontend/src/components/data-table/download-actions.tsx
+++ b/frontend/src/components/data-table/download-actions.tsx
@@ -5,6 +5,10 @@ import {
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
+  DropdownMenuSub,
+  DropdownMenuSubContent,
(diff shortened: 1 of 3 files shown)


Light2Dark (approved):
🚀

PR #5019, opened 2025-05-21
Fix duplicate `import marimo as mo` in .ipynb conversion.

asford:
## 📝 Summary

Fixes #5017

## 🔍 Description of Changes

See #5017

## 📋 Checklist

Test written but not executed

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] I have added tests for the changes made.
- [ ] I have run the code and verified that it works as expected.

## 📜 Reviewers

@akshayka OR @mscolnick

PR #5022, opened 2025-05-21
fix: couple more edge cases for Fix duplicate `import marimo as mo`

mscolnick:
Another edge case for #5019: 
- inside comments
- inside def


dmadisetti (approved):
Captures the tests, but probably worth a rewrite at some point

PR #5043, opened 2025-05-22
improvement: marimo.log file and set default log-level

mscolnick:


marimo/_loggers.py
```diff
@@ -51,28 +72,62 @@ def set_level(level: str | int = logging.WARNING) -> None:
...
+                # Don't increase the log level of a file handler
+                handler.setLevel(min(_LOG_LEVEL, handler.level))
+            elif isinstance(handler, logging.StreamHandler):
+                handler.setLevel(_LOG_LEVEL)
 
 
 def get_logger(name: str, level: Optional[int] = None) -> logging.Logger:
+    """
+    Get a logger with a given name and level.
+    If the logger is alreadted created, we return the cached logger.
```
akshayka, line 86:
```suggestion
    If the logger is already created, we return the cached logger.
```

akshayka (dismissed):

PR #5045, opened 2025-05-22
feat(sandbox): Prefer `uv export` for dumping requirements.txt

manzt:
## 📝 Summary

Sandbox environments currently have trouble reusing cached virtual environments from `uv`, resulting in long startup times for packages like `pandas`.

From my testing, this appears to be caused by the way requirements are parsed and exported via `PyProjectReader`. The new `uv export --script` functionality allows us to defer requirements generation to uv, which seems to lead to better cache hits and improved reuse of environments.

https://github.com/user-attachments/assets/d3de1b57-b9b8-4b1a-845c-f7672457a4e9

## 🔍 Description of Changes

Rather than removing `PyProjectReader`, which is still useful in non-uv contexts (pyodide), this PR adds a shortcut: when `uv` is available, we use `uv export --script` to generate requirements.txt directly. If that fails for any reason, we fall back to the existing logic that extracts dependencies manually.

This change better aligns with how uv expects requirements to be structured and should reduce cold-start latency in sandboxed environments.

Note: This PR does not yet handle relative paths for local dependencies. I think we should address that separately by introducing a method to canonicalize relative paths during export.

~Putting this in draft for now. Some tests are failing and I’ll investigate when I get a chance.~

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -188,12 +188,13 @@ def construct_uv_flags(
         "--no-project",
         # trade installation time for faster start time
         "--compile-bytecode",
-        # layer additional deps on top of the requirements
-        "--with",
-        ",".join(additional_deps),
         "--with-requirements",
         temp_file.name,
     ]
+
+    # Layer additional deps on top of the requirements
+    if len(additional_deps) > 0:
+        uv_flags.extend(["--with", ",".join(additional_deps)])
 
     # Add refresh
     if uv_needs_refresh:
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -147,6 +147,15 @@ def _uv_export_script_requirements_txt(
     return result.stdout.split("\n")
 
 
+def _resolve_requirements_txt_lines(pyproject: PyProjectReader) -> list[str]:
+    if pyproject.name and pyproject.name.endswith(".py"):
+        try:
+            return _uv_export_script_requirements_txt(pyproject.name)
+        except subprocess.CalledProcessError:
+            pass  # Fall back if uv fails
+    return pyproject.requirements_txt_lines
+
+
 def get_marimo_dir() -> Path:
     return Path(__file__).parent.parent.parent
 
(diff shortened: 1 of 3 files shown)


marimo/_cli/sandbox.py
```diff
@@ -163,6 +188,9 @@ def construct_uv_flags(
         "--no-project",
         # trade installation time for faster start time
         "--compile-bytecode",
+        # layer additional deps on top of the requirements
+        "--with",
+        ",".join(additional_deps),
```
mscolnick, line 193: possibly double check if `additional_deps` is empty?

marimo/_cli/sandbox.py
```diff
@@ -137,7 +160,12 @@ def construct_uv_flags(
     # NB. Used in quarto plugin
 
     # If name if a filepath, parse the dependencies from the file
-    dependencies = pyproject.requirements_txt_lines
+    try:
+        # Try exporting env with uv
+        dependencies = _uv_export_script_requirements_txt(pyproject.name)
```
mscolnick, line 165: this could be a pyproject.toml (i think), so we could do a light check that it ends with `.py`

marimo/_cli/sandbox.py
manzt: Good idea.

marimo/_cli/sandbox.py
manzt: When would that happen?

```sh
marimo edit . --sandbox
``` 

?

manzt:
A nice side effect of this change is that `uv export` already captures the full resolution semantics (including `uv.tool` metadata and PEP 723) so we don't have to worry about imperfectly replicating that logic in `PyProjectReader.requirements_txt_lines`, or keeping up with changes to uv's behavior.

marimo/_cli/sandbox.py
mscolnick: yea exactly. you can start the sandbox without a script/file in mind. still ephemeral, but starts empty

tests/_cli/test_sandbox.py
```diff
@@ -294,15 +294,12 @@ def test_construct_uv_cmd_with_additional_deps() -> None:
...
+    with_dependencies_index = uv_cmd.index("--with") + 1
+    with_dependencies = uv_cmd[with_dependencies_index]
 
-    # Read the requirements file to verify additional deps were added
-    with open(req_file_path) as f:
-        requirements = f.read()
-        assert "numpy>=1.20.0" in requirements
-        assert "pandas" in requirements
+    assert "pandas" in with_dependencies
+    assert "numpy>=1.20.0" in with_dependencies
```
manzt, line 303: These additional dependencies are no longer in the requirements file.

mscolnick: awesome, this is a great change.

mscolnick (approved):
this is great, glad we don't need to handle the `uv.sources` and whatever other new semantics get added in future. 

thanks for adding the fallback in case people are still on older versions of `uv` still

PR #5047, opened 2025-05-22
fix: typing with default cells

dmadisetti:
## 📝 Summary

Reported from discord

Type carrying only seemed to work with toy examples, since ToplevelExtraction only retained named cells- this change retains all cell statuses allowing for correct variable extraction.

@akshayka OR @mscolnick


akshayka (commented):
Should there be an end-to-end test that calls `codegen::generate_filecontents`?

dmadisetti:
Yeah, I'll add a snapshot test

mscolnick (approved):

PR #5056, opened 2025-05-23
fix: recursive cell hashing

dmadisetti:
## 📝 Summary

WIP playing with implementation

### DIFF
--- a/marimo/_save/cache.py
+++ b/marimo/_save/cache.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import re
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


dmadisetti:
I have a more up to date version. But the final product should cache bust, so been holding off

mscolnick (dismissed):

dmadisetti:
I think this is OK for review- but the release it goes in should be a minor change since this is cache breaking. Corresponding changes don't require any more upstream alterations

mscolnick (approved):

PR #5066, opened 2025-05-23
fix: respect user log level

akshayka:
This fixes an issue in which the user specified log level had no effect (debug logs were not shown in console nor file);
in addition to setting the handlers' levels, we need to set the logger's level as well.

This also removes the call to logging.basicConfig(): doing anything
with the root logger is considered poor practice for libraries as
it interferes with clients' ability to configure logging for their
application. This is somewhat complicated by the fact that marimo
is both an application and a library but I think still stands.


mscolnick (commented):
I think this will make no logs below warning go to the file logger by default, no?

PR #5081, opened 2025-05-25
fix: github lsp build script, language server path

akshayka:
The build script was calling mkdir on a directory that already existed, and failing silently.

This change fixes that and also updates the script to terminate on error so we don't get any more silent failures.

Additionally, this updates the path to the language server to not include dist -- when generating with pnpm 10.11 on macOS, at least, no such directory is present.

### DIFF
--- a/scripts/buildlsp.sh
+++ b/scripts/buildlsp.sh
@@ -1,17 +1,24 @@
 #!/bin/sh
+#
+# Builds the copilot lsp, then copies the generated dist/ folder to
+# marimo/_lsp/copilot
+
+# This script should fail if any intermediate step fails
+set -e
 
 cd lsp
 if pnpm build; then
   echo "Removing old lsp files..."
   rm -rf ../marimo/_lsp/
   echo "Copying new lsp files..."
-  mkdir ../marimo/_lsp
+  mkdir -p ../marimo/_lsp
   cp dist/index.cjs ../marimo/_lsp/
   # There seems to be a discrepancy between CI and mac for pnpm builds
   if [ -d dist/copilot ]; then
+    echo "branch 1"
     cp -R dist/copilot ../marimo/_lsp/copilot
   else
-    mkdir ../marimo/_lsp
+    echo "branch 2"
     cp -R dist ../marimo/_lsp/copilot
   fi
   echo "Compilation succeeded.\n"
--- a/lsp/index.ts
+++ b/lsp/index.ts
@@ -47,7 +47,6 @@ const Logger = {
 const COPILOT_LSP_PATH = path.join(
   __dirname,
   "copilot",
-  "dist",
   "language-server.js",
 );
 if (!fsSync.existsSync(COPILOT_LSP_PATH)) {


akshayka:
Converted to draft because appears to be not quite right. But what's on main is not correct.

mscolnick (approved):

scripts/buildlsp.sh
```diff
@@ -1,18 +1,30 @@
...
   echo "Removing old lsp files..."
   rm -rf ../marimo/_lsp/
   echo "Copying new lsp files..."
-  mkdir ../marimo/_lsp
+  mkdir -p ../marimo/_lsp
   cp dist/index.cjs ../marimo/_lsp/
   # There seems to be a discrepancy between CI and mac for pnpm builds
   if [ -d dist/copilot ]; then
+    echo "Copying dist/copilot to _lsp/copilot"
     cp -R dist/copilot ../marimo/_lsp/copilot
```
dmadisetti, line 19: I think it's `dist/copilot/dist/language-server.js` -> `../marimo/copilot/dist/language-server.js`, which means

```js
const COPILOT_LSP_PATH = path.join(__dirname, "copilot", "language-server.js");
```

fails

scripts/buildlsp.sh
```diff
@@ -1,18 +1,30 @@
...
+# This script should fail if any intermediate step fails
+set -e
 
 cd lsp
 if pnpm build; then
   echo "Removing old lsp files..."
   rm -rf ../marimo/_lsp/
   echo "Copying new lsp files..."
-  mkdir ../marimo/_lsp
+  mkdir -p ../marimo/_lsp
```
dmadisetti, line 14: `set -e` is awesome idea. But why is this broken? Given `rm -rf ../marimo/_lsp/`? If this was a local issue, there's a
chance that some file with bad permissions could not be deleted

dmadisetti:
@akshayka I think there's a difference between yours, @mscolnick  and mine and CI. I think this works for CI (`dist/dist/lanuage-server.js`) and yours (`dist/lanuage-server.js` I'm guessing), but breaks from the tree structure I see (`dist/copilot/dist/lanuage-server.js`).

dmadisetti:
Ok, seems fine with 0.10.11, I get the 3rd branch

Checked the CI and it gets the second branch:
`Copying contents of dist/dist/. to _lsp/copilot`

The first branch is not right, but should be fine with the suggestion. We can honestly just remove it if all our pnpms are up to date

scripts/buildlsp.sh
dmadisetti, line 19:
```suggestion
  if [ -d dist/copilot/dist ]; then
    echo "Copying dist/copilot to _lsp/copilot"
    cp -R dist/copilot/dist ../marimo/_lsp/copilot
```

scripts/buildlsp.sh
```diff
@@ -1,18 +1,30 @@
...
+  elif [ -d dist/dist ]; then
+    echo "Copying contents of dist/dist/. to _lsp/copilot"
+    cp -R dist/dist/. ../marimo/_lsp/copilot/
   else
-    mkdir ../marimo/_lsp
-    cp -R dist ../marimo/_lsp/copilot
+    echo "Copying the contents of dist/. to _lsp/copilot"
+    # Do NOT place dist as a subdirectory in copilot/, just copy its contents;
+    # the period after dist/ is important here.
+    cp -R dist/. ../marimo/_lsp/copilot/
```
dmadisetti, line 27: Hmm, this almost confirms `rm -rf ../marimo/_lsp/` didn't clean up everything. This is a good catch all, but the folder should not already exist due to the` rm`

PR #5097, opened 2025-05-27
fix: Clear diagnostics for other cells

mscolnick:
Clear diagnostics for other cells

Styles:
- smaller scrollbars in the table and tooltip
- fix padding in tooltip hover


frontend/src/core/codemirror/language/languages/python.ts
```diff
@@ -70,6 +70,8 @@ const lspClient = once((lspConfig: LSPConfig) => {
   const ignoredRuffRules = [
     // Even ruff documentation of this rule explains it is not useful in notebooks
     "B018", // Useless expression
+    // isort
+    "I001", // Import block is un-sorted or un-formatted
```
Light2Dark, line 74: I notice imports don't get sorted atm, is this something to implement in the future?

Issue #4318, opened 2025-03-30
Allow user to set AI assist provider


hongbo-miao:
### Description

We have `claude-3-7-sonnet` model behind LiteLLM https://github.com/BerriAI/litellm which exposes OpenAI format API.

However, currently based on

![Image](https://github.com/user-attachments/assets/8a9b1668-fcf1-4c0e-a4ea-6ff9b08226c7)

> If the model starts with "claude-", we will use your Anthropic API key. If the model starts with "gemini-", we will use your Google Al API key. Otherwise, we will use your OpenAI API key.

it will use Anthropic API instead of OpenAI API.

### Suggested solution

It would be great to allow user to set AI assist provider, thanks! ☺️

hongbo-miao:
Hi @mscolnick I think guessing based on `base_url` or `model` are kind of tricky. And the logic may become more and more complex.

I saw Cline has a good way to set up:

![Image](https://github.com/user-attachments/assets/97ec6d10-dcbb-4645-aa9a-d23a09fc5c0a)

![Image](https://github.com/user-attachments/assets/277ba545-a0c8-44de-ae0e-e6b967822405)

![Image](https://github.com/user-attachments/assets/d28b4e4d-617c-4607-a324-3cc3077c3ec7)

If some UI allows us to select provider like Cline, that would be awesome, thanks!

mscolnick:
Sounds good, this looks like a much better approach.

sanzoghenzo:
Hi, I started to implement the changes for #4460, and as suggested I'm introducing an `ai.provider_type` configuration option (focusing on the back-end for now).

Should we totally get rid of the model based selection and the require the `provider_type`, or should we keep the current logic as a fallback?

should we also move the `model` option up to the `ai` section (`AiConfig` class)?
Obviously I would still check for `ai.open_ai.model` to avoid breaking changes, maybe displaying a deprecation warning for a while.

mscolnick:
@sanzoghenzo - maybe we can keep the logic for now as a fallback. 

the default `provider_type` could be an empty string `""`, and if its empty, then we use the old logic. the frontend can allow you to set a specific provider, but won't let you set back to an empty string.

sanzoghenzo:
> for moving `ai.open_ai.model` to `ai.model`. I think that makes sense...just thinking out loud, i am now wondering if each config should have their own model in case you want to switch providers easily and keep one model per provider. (i personally don't do that)

[...]

PR #5100, opened 2025-05-27
feat: add ai.provider_type option and handle azure with custom domains

### LINKED ISSUE #4460: Support for Azure OpenAI with custom domain (Azure API / corporate environment)
### Description

I'm not sure if you prefer to have a new issue or if I should add this to as a comment of a closed issue, so forgive me if I choose the wrong way 😅 ...

#3117 added Azure OpenAI support for AI Assistant based on the domain of the `base_url` option, which has to be `.openai.azure.com`.

My organization gave me (and my colleagues) access to an Azure OpenAI instance via an url with domain `azure-api.net`, so I cannot use it with marimo.

### Suggested solution

reconsider the introduction of the `api_type` parameter to handle Azure OpenAI custom domains.

The rest of the url is the same, so the other parameters can still be parsed with the current logic

### Alternative

- add a list of possible azure domains for the url parsing, but who knows which are they?
- Add an `[ai.azure]` section to handle the Azure OpenAI separately

### Additional context

I'm also using Aider, so it would be great to be able to reuse the `AZURE_API_BASE`, `AZURE_API_KEY` and `AZURE_API_VERSION` envvars, but I understand this would be out of scope

sanzoghenzo:
Sorry for the delay!

## 📝 Summary

Adds a `ai.provider_type` option to the settings in order to explicitly state the provider to use.

Plus, it handles Azure OpenAI models with custom domains.

**BEWARE**: as I mentioned in https://github.com/marimo-team/marimo/issues/4318#issuecomment-2821597569 the frontend part is just drafted, I don't really know what I did 😅 

closes #4460 #4318

## 🔍 Description of Changes

- added the `provider_type` and `azure`  arguments to the `AiConfig` class
- extracted the Azure part of the `openai` class to an `azure` `ChatModel` subclass, that does the parsing of the model from the url
- added a `_require_api_key` function to DRY the code
- added a `for_azure` classmethod to `AnyProviderConfig`, plus a `_for_openai_like` classmethod to share the common code with `for_openai`
- added the `from_provider_type` method to `AnyProviderConfig` to get the provider from the `provider_type` setting.
- turned `AnyProviderConfig` `staticmethod`s to `classmethod`s and introduced the `_get_ai_config` function to DRY the code

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/marimo/_server/ai/providers.py
+++ b/marimo/_server/ai/providers.py
@@ -75,79 +75,101 @@ class AnyProviderConfig:
     ca_bundle_path: Optional[str] = None
     client_pem: Optional[str] = None
 
-    @staticmethod
-    def for_openai(config: AiConfig) -> AnyProviderConfig:
-        if "open_ai" not in config:
-            raise HTTPException(
-                status_code=HTTPStatus.BAD_REQUEST,
-                detail="OpenAI config not found",
-            )
-        key = _get_key(config["open_ai"], "OpenAI")
-        return AnyProviderConfig(
-            base_url=_get_base_url(config["open_ai"]),
+    @classmethod
+    def for_openai(cls, config: AiConfig) -> AnyProviderConfig:
+        return cls._for_openai_like(config, "open_ai", "OpenAI")
+
+    @classmethod
+    def for_azure(cls, config: AiConfig) -> AnyProviderConfig:
+        return cls._for_openai_like(config, "azure", "Azure OpenAI")
+
+    @classmethod
+    def _for_openai_like(
+        cls, config: AiConfig, key: str, name: str
+    ) -> AnyProviderConfig:
+        ai_config = _get_ai_config(config, key, name)
+        key = _get_key(ai_config, name)
+        return cls(
+            base_url=_get_base_url(ai_config),
             api_key=key,
-            ssl_verify=config["open_ai"].get("ssl_verify", True),
-            ca_bundle_path=config["open_ai"].get("ca_bundle_path", None),
-            client_pem=config["open_ai"].get("client_pem", None),
+            ssl_verify=ai_config.get("ssl_verify", True),
+            ca_bundle_path=ai_config.get("ca_bundle_path", None),
+            client_pem=ai_config.get("client_pem", None),
         )
 
-    @staticmethod
-    def for_anthropic(config: AiConfig) -> AnyProviderConfig:
-        if "anthropic" not in config:
-            raise HTTPException(
-                status_code=HTTPStatus.BAD_REQUEST,
-                detail="Anthropic config not found",
-            )
-        key = _get_key(config["anthropic"], "Anthropic")
-        return AnyProviderConfig(
-            base_url=_get_base_url(config["anthropic"]),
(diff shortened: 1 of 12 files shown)


marimo/_server/ai/providers.py
```diff
@@ -75,79 +75,101 @@ class AnyProviderConfig:
...
+            return cls.for_google(config)
         elif _model_is_bedrock(model):
-            return AnyProviderConfig.for_bedrock(config)
+            return cls.for_bedrock(config)
         else:
-            return AnyProviderConfig.for_openai(config)
+            return cls.for_openai(config)
+
+    @classmethod
+    def from_provider_type(cls, config: AiConfig) -> AnyProviderConfig | None:
```
mscolnick, line 147: i guess for backwards compat we can check if `provider_type` is defined and if so, use `from_provider_type` else use `for_model`?

mscolnick:
@sanzoghenzo i just pushed up some changes to the frontend. i had some thoughts/questions:

1. this is likely a breaking change for folks using `azure`. this is ok, but we will need to call this out in the release notes or make have some sort of migration
2. the `azure` block re-uses the `OpenAIConfig`. are all these settings relevant? would they diverge? I think duplication could be preferred over re-use →  so instead create a `AzureAiConfig`
3. right now `model` is hoisted to the top. do you think we should have a model on each config? since we are no longer inferring it? this may also need some small migration.

marimo/_server/ai/providers.py
sanzoghenzo: Yes! I've added it for the `/chat` endpoint.

sanzoghenzo:
>     1. this is likely a breaking change for folks using `azure`. this is ok, but we will need to call this out in the release notes or make have some sort of migration

Definitely! Is there a designated module/function already in place to add  this migration?

>     2. the `azure` block re-uses the `OpenAIConfig`. are all these settings relevant? would they diverge? I think duplication could be preferred over re-use →  so instead create a `AzureAiConfig`

Honestly I don't really know, I'll check to see if they (mainly are the ssl and certificate options for the httpx client) are needed.
 
>     3. right now `model` is hoisted to the top. do you think we should have a model on each config? since we are no longer inferring it? this may also need some small migration.

Well, having the model tied to each provider enables us to configure multiple providers and quickly switch providers, so it would be a good addition!

mscolnick:
Yep @sanzoghenzo that sounds exactly right to me. 

We can keep the `from_model` as a fallback in case provider_type gets deleted or cleared, in the meantime. I can follow up with simplifying that code path later.

mscolnick:
@sanzoghenzo I'm going to get this feature in another PR. We are going with fully qualified IDs (`azure/gpt-4o`) instead of a dropdown for the provider. This is because it will make having different models and providers for other parts of the UI easier.

sanzoghenzo:
Ok, sorry for not being active here, other projects got precedence, and I didn't use the azure model that much.

mscolnick:
Superseded by https://github.com/marimo-team/marimo/pull/6018

PR #5106, opened 2025-05-28
improvement: add Upgrade action to package manager UI

### LINKED ISSUE #4047: Updating sandboxed versions
Is there any recommended way to just update all the versions stored in the dependencies?  Do I need to just go through by hand and find the current version of each package?

I like the idea of locking the versions, so that my web interface can continue working, but I'd also like to periodically go in and try it to see if it can use new versions of everything, possibly manually downgrading things that don't work.

mscolnick:
Fixes #4047

Add a way to upgrade deps in the package manager UI.

This allows you to upgrade all deps (including transitive), but in a followup, i plan to split up main deps and transitive deps (if it is supported). e.g `uv tree`.

TODO
- [x] test uv scripts
- [x] test uv pip
- [x] test uv projects
- [x] test pip
- [ ] test rye
- [x] test pixi

Screenshot 2025-05-27 at 8 44 54 PM


marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -86,10 +88,18 @@ def should_auto_install(self) -> bool:
     def is_manager_installed(self) -> bool:
         return is_pyodide()
 
-    async def _install(self, package: str) -> bool:
+    async def _install(self, package: str, *, upgrade: bool) -> bool:
         assert is_pyodide()
         import micropip  # type: ignore
 
+        # If we're upgrading, we need to uninstall the package first
```
Light2Dark, line 95: Is there a chance where we uninstall, try to install new packages which fails, resulting in just uninstalling the original package.

Light2Dark (approved):

Light2Dark (approved):

PR #5109, opened 2025-05-28
fix: convert to markdown & sql

### LINKED ISSUE #5088: Mark down cells dont save their type "mark down" properly
### Describe the bug

In my script i have cells which I have applied "Convert to Markdown" to. This works fine while working with it. The text is shown as mark down text without errors. 

But after saving, closing and restarting the kernel and the notebook, these cells are plain text and as they dont obey to python syntax, I get syntax error on all of these. So after restart, they need to be converted to markdown again for the notebook to properly run. 

### Will you submit a PR?

- [ ] Yes

### Environment

app._unparsable_cell(
    r"""
    extract the data
    """,
    column=None, disabled=False, hide_code=True, name="_"
)

This cell wasn't run because it has errors

line 1 plot the situation ^^^ SyntaxError: invalid syntax 

### Code to reproduce

_No response_

Light2Dark:
## 📝 Summary

Fixes #5088. If we keep code as is, then it doesn't get wrapped in `mo.md` or `mo.sql`.
I may be missing something obvious or something regressed recently..

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/editor/actions/useCellActionButton.tsx
+++ b/frontend/src/components/editor/actions/useCellActionButton.tsx
@@ -273,7 +273,7 @@ export function useCellActionButtons({ cell }: Props) {
             return;
           }
           maybeAddMarimoImport(autoInstantiate, createCell);
-          switchLanguage(editorView, "markdown", { keepCodeAsIs: true });
+          switchLanguage(editorView, "markdown", { keepCodeAsIs: false });
         },
       },
       {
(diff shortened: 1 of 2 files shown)


mscolnick (approved):
likely a regression from the RTC changes (or always existed, but has now been surfaced from the RTC changes)

PR #5120, opened 2025-05-29
chore: cleanup config passing to HTML from server

mscolnick:
This is a start of removing some of the complexity taken on by passing config via HTML dom elements. This also paves the way to clean up some global state that has leaked over time.

There are some DOM elements still left around, but will handle those in a followup


dmadisetti:
Looks good at first glance and I see the utility of being able to pass this down. Is it just filename that's left?

AFK but my thought is to:
 - try local islands dev
 - quarto
 - also the `marimo edit html` functionality

akshayka:
Looks good at a high level. Noticed the backward compatibility, nice. Some tests failing, haven't kicked the tires, but looks good otherwise

PR #5131, opened 2025-05-30
improvement: notebook schema serialization

mscolnick:


dmadisetti (dismissed):
LGTM, before you merge can you leave a comment on the parts you touched in case we ever need to bump to NotebookV2? Just something for git blame to pick up

PR #5132, opened 2025-05-30
feat: add redshift engine

Light2Dark:
## 📝 Summary

Will clean up some stuff like converting to polars and errors, in follow up PRs.

## 📋 Checklist

- [X] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).

### DIFF
--- a/marimo/_sql/engines/types.py
+++ b/marimo/_sql/engines/types.py
@@ -13,9 +13,9 @@ from marimo._runtime.context.types import (
     get_context,
     runtime_context_installed,
 )
+from marimo._sql.utils import raise_df_import_error
 from marimo._types.ids import VariableName
 
-ENGINE_REGISTRY: list[type[BaseEngine[Any]]] = []
 NO_SCHEMA_NAME = ""
 
 
@@ -72,11 +72,6 @@ class BaseEngine(ABC, Generic[CONN]):
 T = TypeVar("T", bound=BaseEngine[Any])
 
 
-def register_engine(cls: type[T]) -> type[T]:
-    ENGINE_REGISTRY.append(cls)
-    return cls
-
-
 class EngineCatalog(BaseEngine[CONN], ABC):
     """Protocol for querying the catalog of an engine."""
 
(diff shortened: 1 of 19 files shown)


marimo/_sql/engines/types.py
```diff
@@ -130,6 +124,7 @@ def execute(self, query: str) -> Any:
         """Execute a SQL query and return a dataframe."""
         pass
 
+    # TODO: Maybe this should be called during init of db's
```
mscolnick, line 127: Users could change this afterwards. Are you worried about the performance of it?

marimo/_sql/sql.py
```diff
@@ -42,6 +41,7 @@ def sql(
         | ClickhouseClient
         | ChdbConnection
         | IbisEngine
+        | RedshiftEngine
```
mscolnick: I think ibis and redshift are actually not correct here or needed, because they pass the ibis or redshift connector, not the actual engine  

I wonder if this is valuable though or if this should be some opaque type like the DBAPI (or Any)

marimo/_sql/utils.py
```diff
@@ -12,6 +12,8 @@
 
 if TYPE_CHECKING:
     import duckdb
+    import polars as pl
+    from polars._typing import ConnectionOrCursor
```
mscolnick, line 16: Probably ok given its just typing

mscolnick (approved):
Awesome work on this! Few comments, but approving pending the bit of cleanup

marimo/_sql/sql.py
Light2Dark: Yeah, i think we can use dbapiconnection, the docs should help

marimo/_sql/engines/types.py
Light2Dark: mm hmm yeah, since perf hasn't been an issue, ig we can keep it.

I'll clean up in a following

PR #5143, opened 2025-05-30
fix: Throw reasonable error for state set in the same cell

### LINKED ISSUE #5139: Calling setter in `mo.state` definition cell throws unexpected error
### Describe the bug

I am attempting to store pre-computed data for interactive visualisations. As a result, I needed to recompute data on command.

My issue arises with the following minimal case:
```py
import marimo as mo
import numpy as np

gettr, settr = mo.state(np.array([1, 2]))
settr(np.array([2, 1]))
```
This gives the error:
```
  File "/Users/filup/Documents/Programming/Python/marimo/math-notes/.venv/lib/python3.12/site-packages/marimo/_runtime/runtime.py", line 1854, in update_stateful_values
    ) - self.graph.get_defining_cells(name)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/filup/Documents/Programming/Python/marimo/math-notes/.venv/lib/python3.12/site-packages/marimo/_runtime/dataflow.py", line 83, in get_defining_cells
    return self.definitions[name]
           ~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'b149e675-4a98-42d2-866d-59198af92fb1'

```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.13.15",
  "OS": "Darwin",
  "OS Version": "24.5.0",
  "Processor": "arm",
  "Python Version": "3.12.5",
  "Binaries": {
    "Browser": "--",
    "Node": "v21.6.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8",
    "narwhals": "1.41.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "starlette": "0.46.2",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.2",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "anywidget": "0.9.18",
    "openai": "1.82.1",
    "pandas": "2.2.3",
    "pyarrow": "20.0.0",
    "loro": "1.5.1",
    "python-lsp-ruff": "2.2.2",
    "python-lsp-server": "1.12.2",
    "ruff": "0.11.12"
  },
  "Experimental Flags": {
    "lsp": true
  }
}
```

</details>

### Code to reproduce

_No response_

dmadisetti:
## 📝 Summary

fixes #5139
It does not make sense to set the value in the same cell right? Just making sure that my understanding of the behavior is consistent

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_runtime/state.py
+++ b/marimo/_runtime/state.py
@@ -193,7 +193,7 @@ class SetFunctor(Generic[T]):
 
         # Need to explicitly check that we are not in the defining cell.
         if (
-            self._state._context == None
+            self._state._context is None
             and self._state._defining_cell == ctx.execution_context.cell_id
         ):
             return
--- a/marimo/_runtime/state.py
+++ b/marimo/_runtime/state.py
@@ -153,6 +153,7 @@ class State(Generic[T]):
         self.allow_self_loops = allow_self_loops
         self._set_value = SetFunctor(self)
         self._defining_cell: Optional[CellId_t] = None
+        self._context = _context
 
         try:
             ctx = get_context()
@@ -191,10 +192,11 @@ class SetFunctor(Generic[T]):
             return
 
         # Need to explicitly check that we are not in the defining cell.
-        if self._state._defining_cell == ctx.execution_context.cell_id:
-            raise RuntimeError(
-                "State setter cannot be called in the defining cell. "
-            )
+        if (
+            self._state._context == None
+            and self._state._defining_cell == ctx.execution_context.cell_id
+        ):
+            return
         # Note we could allow this with:
         # >>> ctx.state_registry.register_scope(ctx.glbls, defs=cell.defs)
         # But might have unintended consequences.
--- a/marimo/_runtime/state.py
+++ b/marimo/_runtime/state.py
@@ -9,6 +9,7 @@ from uuid import uuid4
 
 from marimo._output.rich_help import mddoc
 from marimo._runtime.context import ContextNotInitializedError, get_context
+from marimo._types.ids import CellId_t
 
 T = TypeVar("T")
 Id = int
@@ -151,11 +152,15 @@ class State(Generic[T]):
         self._value = value
         self.allow_self_loops = allow_self_loops
         self._set_value = SetFunctor(self)
+        self._defining_cell: Optional[CellId_t] = None
 
         try:
+            ctx = get_context()
             if _registry is None:
-                _registry = get_context().state_registry
+                _registry = ctx.state_registry
             _registry.register(self, _name, _context)
+            if ctx.execution_context:
+                self._defining_cell = ctx.execution_context.cell_id
         except ContextNotInitializedError:
             # Registration may be picked up later, but there is nothing to do
             # at this point.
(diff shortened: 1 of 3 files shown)


mscolnick (dismissed):

akshayka (dismissed):

dmadisetti:
Hmm, maybe we do silently allow? Failing test case:

```python
    async def test_set_ui_element_value_lensed_with_state(                                                                                                                  
        self, any_kernel: Kernel, exec_req: ExecReqProvider                                                                                                                 
    ) -> None:                                                                                                                                                              
        """Test setting value of a lensed element with on_change set_state"""                                                                                               
        k = any_kernel                                                                                                                                                      
        await k.run([exec_req.get(code="import marimo as mo")])                                                                                                             
                                                                                                                                                                            
        # Create an array and output it ...                                                                                                                                 
        cell_one_code = """                                                                                                                                                 
        get_state, set_state = mo.state(None)                                                                                                                               
                                                                                                                                                                            
        array = mo.ui.array(                                                                                                                                                
            [mo.ui.slider(0, 10, value=1, on_change=set_state),                                                                                                             
            mo.ui.slider(0, 10, value=1, on_change=set_state)                                                                                                               
        ]);                                                                                                                                                                 
        """                                                                                                                                                                 
        await k.run([exec_req.get(code=cell_one_code)])                                                                                                                     
        await k.run([er := exec_req.get(code="state = get_state()")])    
```

marimo/_runtime/state.py
```diff
@@ -181,6 +187,19 @@ def __call__(self, update: T | Callable[[T], T]) -> None:
...
+
+        if not ctx.execution_context:
+            return
+
+        # Need to explicitly check that we are not in the defining cell.
+        if (
+            self._state._context is None
+            and self._state._defining_cell == ctx.execution_context.cell_id
+        ):
+            return
```
akshayka, line 199: This isn't quite right (see how `test_set_ui_element_value_lensed_with_state` is now failing), it prevents the on change handler from running.

I don't think we should change how state updates are registered here. Looking at original issue, I am more confused about why the name being looked up is not a name at all.

```
  File "/Users/filup/Documents/Programming/Python/marimo/math-notes/.venv/lib/python3.12/site-packages/marimo/_runtime/runtime.py", line 1854, in update_stateful_values
    ) - self.graph.get_defining_cells(name)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/filup/Documents/Programming/Python/marimo/math-notes/.venv/lib/python3.12/site-packages/marimo/_runtime/dataflow.py", line 83, in get_defining_cells
    return self.definitions[name]
           ~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'b149e675-4a98-42d2-866d-59198af92fb1'
```

`b149e675-4a98-42d2-866d-59198af92fb1` is certainly not a bound variable name.

dmadisetti: State is associated to variable name on a post execution hook by examining the def results. So when state_set is set in the same definition- the registry doesn't have an associated bound name.

We could run inspect to get the name though- or explicitly handle in runtime.

akshayka: I don't follow why state needs to be associated with a variable name in this way. Looking into it, it looks like it was for handling file watching.

In the original implementation of state there was no need. The cells to run are after a state update are determined at runtime here, by just finding which cells have the state object among their refs:

https://github.com/marimo-team/marimo/blob/main/marimo/_runtime/runner/cell_runner.py#L314-L319

Looking at the code it now looks like there are two mechanisms for running cells that are made stale by state updates: `the state_updates` dict which is consumed in `run_cells_internal` and the new method added for file watching which eagerly enqueues an execute stale cells request.

Original mechanism:

https://github.com/marimo-team/marimo/blob/main/marimo/_runtime/runtime.py#L620-L622

https://github.com/marimo-team/marimo/blob/main/marimo/_runtime/runtime.py#L1336-L1340

https://github.com/marimo-team/marimo/blob/main/marimo/_runtime/runtime.py#L1417-L1420

New/second mechanism:

https://github.com/marimo-team/marimo/blob/main/marimo/_runtime/runtime.py#L1435-L1442

I think we just live with this tech debt since this isn't high priority.

Instead of fixing the original issue perhaps we just silence the console output since it's scary. I'm going to close this PR because it's not high priority and the fix isn't clear.

PR #5146, opened 2025-05-30
improvement: marimo converters to/from various formats

mscolnick:
Passing around source code is better than files (otherwise can force you to create temp files, doesn't work nicely to convert from one to the other).

* Single conversion entrypoint. Everything goes through the Internal Representation (IR).
* Cleans up markdown to use the IR instead of the full app and file-router
* Cleans up and simplifies some save logic. DRYs up a lot of code


marimo/_cli/development/commands.py
```diff
@@ -481,7 +482,152 @@ def print_all_routes(app: Any, base_path: str = "") -> None:
...
+    help="Host to serve the preview on",
+    type=str,
+)
+@click.option(
+    "--headless",
+    is_flag=True,
+    default=False,
+    help="Don't automatically open the browser",
+)
+def preview(file_path: Path, port: int, host: str, headless: bool) -> None:
```
dmadisetti, line 511: Oh nice

marimo/_convert/markdown/markdown.py
```diff
@@ -161,20 +180,23 @@ def get_cell_config_from_tag(tag: Element, **defaults: bool) -> CellConfig:
...
+T = TypeVar("T")
+
+
 # TODO: Consider upstreaming some logic such that this isn't such a terrible
 # hack. At some point rewriting / overriding the markdown parser would be a
 # better idea than all these little work arounds.
 @dataclass
-class SafeWrap:
-    app: App
+class SafeWrap(Generic[T]):
```
dmadisetti, line 190: Been meaning to move this file, so this is good. This might break quarto, but will check- generic typing seems more robust than the previous case

I think the test suite is pretty robust, so if they pass (esp, snapshot) I'm good with this

marimo/_server/export/exporter.py
```diff
@@ -298,42 +309,37 @@ def export_as_md(
...
-                    ).configure(cell_data.config)
-                    cell = Cell(
-                        _cell=cell_impl,
-                        _name=cell_data.name,
-                        _app=file_manager.app,
-                    )
-                    cell_data.cell = cell
-                except SyntaxError:
-                    pass
+            language: Language = "python"
```
dmadisetti, line 334: No reason to recompile if the cell is present

mscolnick: the cell is not present now as far as i can tell. im only passing in the serialized notebook IR, not the rich `App`

marimo/_ast/codegen.py
```diff
@@ -349,6 +350,16 @@ def generate_app_constructor(config: Optional[_AppConfig]) -> str:
...
+    return generate_filecontents(
+        codes=[cell.code for cell in ir.cells],
+        names=[cell.name for cell in ir.cells],
+        cell_configs=[CellConfig.from_dict(cell.options) for cell in ir.cells],
+        config=_AppConfig.from_untrusted_dict(ir.app.options),
+        header_comments=ir.header.value if ir.header else None,
+    )
+
+
 def generate_filecontents(
```
dmadisetti, line 363: Call to this can be cleaned up while you're here. Does serialization carry header info?

marimo/_cli/development/commands.py
dmadisetti: Doesn't work on markdown files

marimo/_ast/codegen.py
mscolnick: the IR does, but not the noteobok.json (will add in a followup)

PR #5161, opened 2025-06-02
Add mechanism for injecting custom file stores

pepijndevos:
## 📝 Summary

There is currently a fixed order of precedence of checking different FileStores as the source for loading and saving a notebook. This works exactly right for current deployments, and even has hardcoded references to the online playground, but doesn't allow custom deployments to have their own sources of data.

## 🔍 Description of Changes

Adds a `file_stores` property to the app config element which takes global class names that are pushed onto the `notebookFileStore` via a new method `CompositeFileStore.insert`.

This allows someone to write code like the following to load notebooks from [PouchDB](https://pouchdb.com/)

```html
<marimo-app-config data-config='{"file_stores": ["PouchDBFileStore"]}' hidden></marimo-app-config>
```

```javascript
        window.PouchDBFileStore = class PouchDBFileStore {
            constructor() { ... }
            async saveFile(contents) { ... }
            async readFile() { ... }
        };
```

I'm open to suggestions for better ways to do this and not very attached to this code

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick


pepijndevos:

frontend/src/core/config/config-schema.ts
```diff
@@ -204,6 +204,7 @@ export const AppConfigSchema = z
     html_head_file: z.string().nullish(),
     auto_download: z.array(z.enum(["html", "markdown", "ipynb"])).default([]),
     sql_output: SqlOutputSchema,
+    file_stores: z.array(z.string()).optional(),
```
mscolnick: rather than putting this in app config here, could we place it in the `mount(` function. this way, you should be able to even pass the full store (not just a string)

mscolnick: this way, you should be able to even pass the full store (not just a string)

mscolnick: just recently i've refactored the entrypoint, so you can soon (in the next release) do something like:

```ts
<script type="module">
      import { mount } from 'https://cdn.jsdelivr.net/npm/@marimo-team/frontend@0.0.0/dist/mount.js';
      mount(
        {
          mode: edit,
          ....
        },
        document.getElementById('root')
      );
    </script>
```

so this should be a better entrypoint for `file_stores`

pepijndevos: I'm trying to test this locally but I can't find any mount.js file that exports mount in a way that I can test this PR. Am I missing something?

pepijndevos: I got it working with `__MARIMO_MOUNT_CONFIG__` which is... fine I guess. I'm just puzzled by the code you shared that I can't get to work at all.

mscolnick: `__MARIMO_MOUNT_CONFIG__` works as well. the code i shared is not shipped yet, but will be in the next release.

i think i need to add a new entrypoint for `mount.js`. but `__MARIMO_MOUNT_CONFIG__` works fine too

pepijndevos: That makes sense. I was thinking maybe I need some bundler magic that turns the paths in https://github.com/marimo-team/marimo/blob/main/frontend/index.html into hashed asset paths. Which I probably still do for maintainability. Does that get generate by vite or something? For my demo script I'm kinda just directly copying and referencing stuff from node_modules.

PR #5169, opened 2025-06-03
Support some JupyterBook cell tags in marimo convert

manzt:
## 📝 Summary

`_transform_sources` is no longer a pure transform (it's wasn't really with the transforms that are source-to-source, followed by some that modify number of cells). It now forms a "pipeline" for converting Jupyter notebook cells to marimo (python-only) format.

This structure lets us handle tags like "hide-code" and cleanly separate source-only and cell-level logic.

## 🔍 Description of Changes

The pipeline is:

- A series of source-to-source transforms
- One source → CodeCell binding step (binds/resolve source with ipynb metadata to create `CellConfig`)
- A series of cell-to-cell transforms (may add/remove cells)

I had to update some tests... I chattted with @mscolnick about maybe updating some of these tests with snapshotting since this is so internal.

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/marimo/_convert/ipynb.py
+++ b/marimo/_convert/ipynb.py
@@ -6,6 +6,7 @@ import json
 import re
 import sys
 from collections import defaultdict
+from dataclasses import dataclass, field
 from typing import Any, Callable, Union
 
 from marimo._ast.cell import CellConfig
@@ -23,8 +24,18 @@ from marimo._schemas.serialization import (
 )
 from marimo._types.ids import CellId_t
 
-# Define a type for our transform functions
+# Define a type for our 1:1 source-only transform functions
 Transform = Callable[[list[str]], list[str]]
+
+
+@dataclass
+class CodeCell:
+    source: str
+    config: CellConfig = field(default_factory=CellConfig)
+
+
+# Define a type for transforms that add/remove cells
+CellsTransform = Callable[[list[CodeCell]], list[CodeCell]]
 
 
 def transform_fixup_multiple_definitions(sources: list[str]) -> list[str]:
@@ -74,7 +85,7 @@ def transform_fixup_multiple_definitions(sources: list[str]) -> list[str]:
     return [transform(source) for source in sources]
 
 
-def transform_add_marimo_import(sources: list[str]) -> list[str]:
+def transform_add_marimo_import(sources: list[CodeCell]) -> list[CodeCell]:
     """
     Add an import statement for marimo if any cell uses
     the `mo.md` or `mo.sql` functions.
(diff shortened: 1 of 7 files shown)


marimo/_convert/ipynb.py
```diff
@@ -608,22 +619,29 @@ def on_ref(
...
+    sources: list[str], metadata: list[dict[str, Any]], hide_flags: list[bool]
+) -> list[CodeCell]:
     """
-    Handle cell metadata, such as tags or cell IDs.
+    One-time transformation that binds sources and (ipynb) metadata into CodeCell objects.
+
+    This marks the boundary between source-only transformations and cell-level transformations.
+
+    - If "hide-cell" is present in the tags, the cell is marked hidden (and removed)
+    - Remaining tags (if any) are inserted as a comment at the top of the source.
```
dmadisetti, line 631: We have a few other options like disable etc..
Are there other analogs?

tests/_convert/test_ipynb.py
```diff
@@ -7,8 +7,8 @@
 
 from marimo._convert.ipynb import (
     _transform_sources,
+    bind_cell_metadata,
```
dmadisetti: Be nice to have a hide cell specific case. Also a case with junk tags

dmadisetti (commented):
Nice cleanup too!

tests/_convert/test_ipynb.py
manzt: I added a snapshot with that use case and moved this out of the unit test. I figure we can add many more tags to that "kitchen sink" example for future tests.

marimo/_convert/ipynb.py
manzt: I can have a closer look. I think for now we could start to enumerate what we anticipate we could handle, then we snapshot the "cell_metadata.ipynb" notebook with a case of the specific one we expect.

tests/_convert/test_ipynb.py
```diff
@@ -518,24 +495,6 @@ def test_transform_duplicate_definitions_with_reference_to_previous():
...
-            "scrolled": False,
-            "custom": {"key": "value"},
-        }
-    ]
-    result = transform_cell_metadata(sources, metadata)
-    assert result == [
-        "# Cell tags: tag1, tag2\nprint('Complex metadata')",
-    ]
-
-
```
manzt, line 538: I moved these tests to `cell_metadata.ipynb` input, with the associated snapshot. I think it will be easier to test the expected outputs that way in the future (just updating the snapshot, potentially for the same tags/metadata as we better adapt).

scripts/generate_ipynb_fixtures.py
```diff
@@ -102,6 +101,61 @@ def main() -> None:
...
+            nb.new_markdown_cell("A markdown cell."),
+            nb.new_markdown_cell(
+                "A markdown cell with tags: ['blah'].",
+                metadata={"tags": ["blah"]},
+            ),
+            nb.new_markdown_cell(
+                "A markdown cell with tags: ['blah', 'hide-cell'].",
+                metadata={"tags": ["blah", "hide-cell"]},
+            ),
+        ],
```
manzt, line 156: Previously the code for markdown cells were displayed if they had tags, subtle bug.

akshayka (commented):
@manzt type check is failing o/w looks good!

manzt:
> @manzt type check is failing o/w looks good!

Good catch, fixed!

akshayka:
CI failure looks unrelated, merging

PR #5172, opened 2025-06-03
feat: allow env level app overrides

dmadisetti:
## 📝 Summary

Much smaller footprint than a flag option. Can also disable FE if an override is present.

@mscolnick

### DIFF
--- a/marimo/_ast/app_config.py
+++ b/marimo/_ast/app_config.py
@@ -72,6 +72,14 @@ class _AppConfig:
 
         return self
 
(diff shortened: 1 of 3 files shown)


dmadisetti:
/marimo create-test-release

so this does not do what we want?

dmadisetti:
Woops. Do not merge. Tests may be correctly broken, not sure why they didn't time out and not fail

mscolnick (approved):

PR #5178, opened 2025-06-03
allow passing an extra script to wasm export

pepijndevos:
## 📝 Summary

in #5161 I introduced a method for setting a custom file store, which works, but using it seems very fragile.

The reason is that marimo uses vite to render the index.html template with content hashes baked into the assets. So when writing my own html, I was faced with the problem of including the right assets which will have a different file name for every minor change.

I have not found a way to load these assets in a maintainable way, so instead I've extended the WASM export to be usable directly.

## 🔍 Description of Changes

The python wasm export command has a new parameter `--extra-script` which is passed through several layers until it eventually injects an extra script tag at the end of the body. The idea is that you can pass a script file to modify and extend Marimo.

I have tested this with a dummy script, but tomorrow morning I'll port the PouchDB file handler over and confirm it works.

## 📋 Checklist

## 📜 Reviewers

@mscolnick

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -51,15 +51,13 @@
         });
         resizeObserver.observe(obj.contentWindow.document.body);
       }
+      window.__MARIMO_MOUNT_CONFIG__ = '{{ mount_config }}';
     </script>
     <marimo-filename hidden>{{ filename }}</marimo-filename>
     <title>{{ title }}</title>
   </head>
   <body>
     <div id="root"></div>
-    <script data-marimo="true">
-      window.__MARIMO_MOUNT_CONFIG__ = '{{ mount_config }}';
-    </script>
     <script type="module" src="/src/main.tsx"></script>
   </body>
 </html>
--- a/marimo/_cli/export/commands.py
+++ b/marimo/_cli/export/commands.py
@@ -511,6 +511,12 @@ and cannot be opened directly from the file system (e.g. file://).
     type=bool,
     help=_sandbox_message,
 )
+@click.option(
+    "--extra-script",
+    type=click.Path(path_type=Path),
+    required=False,
+    help="Extra JavaScript file to include in the exported HTML.",
+)
 @click.argument(
     "name",
     required=True,
@@ -524,6 +530,7 @@ def html_wasm(
     show_code: bool,
     include_cloudflare: bool,
     sandbox: Optional[bool],
+    extra_script: Optional[Path] = None,
 ) -> None:
     """Export a notebook as a WASM-powered standalone HTML file."""
     import sys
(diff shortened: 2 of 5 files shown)


pepijndevos:
Okay I've changed it to take any number of URLs instead, and confirmed that this allows me to configure a PouchDBFileStore and load/save files from it.

mscolnick:
we already support custom head.html files. can you use this instead? https://docs.marimo.io/guides/configuration/html_head/#custom-html-head

pepijndevos:
If I understand correctly that is configured in the notebook rather than the app config, and the whole point is to load the notebook from a database, so I need to inject code _before_ the notebook is loaded but _after_ the initial configuration is set (so I can override configuration settings)

[edit] nvm it can be set in app config, brb

pepijndevos:
Okay the html-head approach works!

mscolnick:
sorry @pepijndevos, my recommendation was just to use the html_head app configuration instead of this PR. 

this just adds a build step but really could exist outside of marimo. im not sure you need to pass this information through marimo just to append it to the HTML file

pepijndevos:
It's not clear to me how to use html head app configuration without this pr before loading the notebook. I can see several ways to modify the html of the wasm notebook:

1. Add it to the notebook. Too late for my use.
2. Write my own html. Vite produces assets with hashes in the name making this hard to maintain.
3. Patch up the html externally. Hacky but my best bet if this pr is rejected.
4. Extend the wasm export to allow custom html. This pr.
5. Build Marino as a well behaved library with stable exported modules. The mount API you're talking about?

All I want to do is have a wasm notebook with saving enabled and a custom file store. I just don't see how to do it currently.

dmadisetti:
Have you looked into islands and export?

Assets in /public are accessible through wasm natively

pepijndevos:
> Have you looked into islands and export?

I have read the docs ten times over but maybe I'm missing something. Islands seems more about static site generators and fixed cells rather than customizing the whole app. Export is exactly the thing I'm trying to fix here so you can customize the exported html.

> Assets in /public are accessible through wasm natively

From within the notebook you mean? I don't think that's relevant.

What i tried initially is to npm install Marino into my app but as i mentioned I can't find a maintainable way to include the assets into my page because vite hashes all the names so it's `index-efcebfa.js`

What I want is exactly what this pr does: export all the assets and html into my app but customize the html. Maaaaybe islands can be made to do that? Maybe a public mount API can be an alternative?

pepijndevos:
@mscolnick so what do you think should happen?

I think this is a reasonable change for adding extra html to a generated wasm notebook.

I looked again at islands and it doesn't do what I want, so to the best of my understanding the alternatives are to export the HTML and then splice in extra html externally, or wait for a fully fledged mount API that I can use as a library to insert all the right javascript, fonts and css.

pepijndevos:
I adjusted my script with a defer and now it needs to be before the marimo module so I guess I'll just do the html splicing myself.

A proper library API would be more robust eventually I think.

PR #5183, opened 2025-06-03
poc: run all banner for on-startup=lazy

mscolnick:


akshayka:
I was thinking a context-sensitive rectangle button that shows to the left of the hamburger menu and settings buttons (top right). Context-sensitive in the sense that:

1. only appears when auto-instantiate is off
2. disappears when you click it (perhaps after a being "notebook started" message), or if you manually run any cell

The button background color could be the yellow-ish color we use throughout the notebook to communicate staleness.

PR #5208, opened 2025-06-05
Use runtime assertions instead of type assertions

manzt:
## 📝 Summary

Replaces some (type only) `as` casts with runtime checks using a new `assert()` helper. 

## 🔍 Description of Changes

Adds a generic `assert` helper, which helps with type narrowing.

## 📋 Checklist

## 📜 Reviewers

@akshayka


frontend/src/mount.tsx
```diff
@@ -223,7 +223,7 @@ function initStore(options: unknown) {
     Logger.error("Invalid marimo mount options", parsedOptions.error);
     throw new Error("Invalid marimo mount options");
   }
-  const mode = parsedOptions.data.mode as AppMode;
```
manzt, line 226: just not necessary

frontend/src/utils/assert.ts
```diff
@@ -0,0 +1,31 @@
...
+ * ```ts
+ * const value: boolean = Math.random() <= 0.5;
+ * assert(value, "value is greater than than 0.5!");
+ * value // true
+ * ```
+ *
+ * @param expression The expression to test.
+ * @param msg The optional message to display if the assertion fails.
+ * @throws an {@link Error} if `expression` is not truthy.
+ */
```
akshayka: Love the documentation

akshayka (dismissed):

mscolnick:
We already have this util called `invariant` (changes are great tho)

manzt:
> We already have this util called invariant (changes are great tho)

I'll switch to that! Thanks.

manzt:
I removed the TypeScript overloads for `invariant` because they don't expand what the base implementation with `unknown` can communicate to the type checker. Removing also makes it slightly more flexible. Also ported over the documentation with an extra example. I removed throwing a specific error (`AssertionError`) for now, since we could always add in the future if we wanted.

akshayka (approved):

PR #5215, opened 2025-06-06
Discover `uv` executable with `UV` environment variable

manzt:
## 📝 Summary

Use the `UV` environment variable set by uv to resolve the binary path, rather than assuming "uv" is available on the system path.

## 🔍 Description of Changes

This PR replaces hardcoded "uv" subprocess calls with a utility function `find_uv_bin()`, which checks the `UV` environment variable. When a process is launched via `uv`, it sets `UV` to the executable path of the binary. The naming is taken from [`uv` python package](https://github.com/astral-sh/uv/blob/0109af1aa510bd4f0759b0d53d842c3716e5cc3f/python/uv/_find_uv.py#L8), which we could replace in the future.

For sandboxing, in cases where `find_uv_bin()` returns the unresolved string "uv" (i.e. `UV` isn't set), we still explicitly check that it's available in the system path before proceeding.

### DIFF
--- a/tests/_cli/test_cli_export.py
+++ b/tests/_cli/test_cli_export.py
@@ -382,7 +382,7 @@ class TestExportHTML:
         output = p.stderr.decode()
         # Check for sandbox message
         assert "Running in a sandbox" in output
-        assert "uv run --isolated" in output
+        assert "run --isolated" in output
         html = normalize_index_html(output)
         # Remove folder path
         dirname = path.dirname(temp_marimo_file)
@@ -921,7 +921,7 @@ class TestExportIpynb:
         output = p.stderr.decode()
         # Check for sandbox message
         assert "Running in a sandbox" in output
-        assert "uv run --isolated" in output
+        assert "run --isolated" in output
 
     @staticmethod
     @pytest.mark.skipif(
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -20,6 +20,7 @@ from marimo._utils.inline_script_metadata import (
     PyProjectReader,
     is_marimo_dependency,
 )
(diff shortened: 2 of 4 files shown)


dmadisetti (dismissed):
Nice, tested with nix

dmadisetti:
Failing tests are relevant though

tests/_cli/test_cli_export.py
```diff
@@ -921,7 +921,7 @@ def test_cli_export_ipynb_sandbox(temp_marimo_file: str) -> None:
         output = p.stderr.decode()
         # Check for sandbox message
         assert "Running in a sandbox" in output
-        assert "uv run --isolated" in output
+        assert "run --isolated" in output
```
manzt, line 924: An alternative would be too look for `uv.exe` or `uv` in this output... the issue is that running the tests with/without uv will determine whether it's uv or a full executable path on windows.

I decided against including the `uv` name because it makes the test flaky based on whether it's run with `uv run hatch` or `hatch`.

manzt: Side note, I do think it's preferable to show the end user the full executed command:

```sh
marimo edit foo.py
# Run in a sandboxed venv containing this notebook's dependencies? [Y/n]: Y
# Running in a sandbox: /Users/manzt/.local/bin/uv run --isolated --no-project --compile-bytecode --with-requirements /var/folders/y1/hqtbfdpd7pgd8lvx65d992vc0000gn/T/tmpqz2r870m.txt --python >=3.13 marimo edit test.py
```

Rather than replacing the uv_cmd part with `uv` to pass the test.

dmadisetti: Yeah agreed. Since it's just a string match, I think it's OK to be a little loose with it too

PR #5216, opened 2025-06-06
fix: Skip PEP 723 check on directories

### LINKED ISSUE #4999: pep 723 check runs on directories
### Describe the bug

`marimo edit .`

gives [W 250520 10:27:33 inline_script_metadata:107]

Maybe an opportunity to use the folder pyproject as a default

### Will you submit a PR?

- [ ] Yes

### Environment

HEAD

### Code to reproduce

_No response_

manzt:
Fixes #4999

I think there could be an opportunity to try to read from `pyproject.toml` ourselves... but honestly I'd rather defer to recommending users use `uv` to bootstrap the environment in that case.

Their project discovery is very robust, and it would ensure alignment. For example:

```py
uv init my_project && cd my_project
uv add polars duckdb
uv run --with=marimo marimo . # will have polars and duckdb installed

mkdir notebooks
cd notebooks
uv run --with=marimo marimo . # will also have polars and duckdb installed
```

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -34,6 +34,9 @@ def maybe_prompt_run_in_sandbox(name: str | None) -> bool:
     if name is None:
         return False
 
+    if Path(name).is_dir():
+        return False
+
     pyproject = PyProjectReader.from_filename(name)
     if not pyproject.dependencies:
         return False


marimo/_cli/sandbox.py
```diff
@@ -34,6 +34,9 @@ def maybe_prompt_run_in_sandbox(name: str | None) -> bool:
     if name is None:
         return False
 
+    if Path(name).is_dir():
```
dmadisetti, line 37: I think we might want to log though to let the user know that sandbox won't work

manzt: Is it the user expectation to enable sandbox if it's not explicitly set with a flag (`--sandbox`)?

Passing `marimo edit foo.py` that contains PEP 723 will prompt the user to opt into bootstrapping with the sandbox env because we discover the metadata. But in this case, there is no metadata that is discovered.

I guess could check for `pyproject.toml` in this directory and just warn the user that we don't check for dependencies, prompting to use `uv run` instead if that's the desired behavior.

dmadisetti: Was thinking along the lines of the explicit sandbox

```bash
$ uv run marimo edit --sandbox .
```

Suggesting `uv run` is good message I think.

but yes I think no log is fine for

```bash
$ uv run marimo edit .
```

manzt: My understanding of the sandbox feature is that it operates on individual scripts/files, not directories. Would we want an error instead of a warning for a directory? 

```sh
$ uv run marimo edit --sandbox .
# error: `--sandbox` expects a Python file, not a directory. Did you mean to pass a script path?
```

manzt: Ah, I see — when I run `uv run marimo edit --sandbox .`, it opens the landing page as expected, but any notebook I create from that directory has sandboxing enabled. So it seems like `--sandbox` applies to the entire directory, and any new files launched from there inherit sandbox mode.

Is that the expected behavior? And is this documented anywhere? It might be worth clarifying that `--sandbox` can operate in both file and directory modes, especially if directory sandboxing implies a broader sandboxing context for all new notebooks. This seems like a recipe for issues if one switches between notebooks with sandboxing...

manzt: I made a longer issue describing some inconsistencies with sandboxing for multiple files: https://github.com/marimo-team/marimo/issues/5219

dmadisetti: Slightly wrong- but your issue captures the correct description.

In theory we could swap out python versions too. Maybe the separate FE stuff I'm tinkering with could allow us to have "kernels"

dmadisetti (approved):
LGTM, let's leave the other sandbox questions to the followup issue

PR #5243, opened 2025-06-09
feat: change page size in tables directly

Light2Dark:
## 📝 Summary

image

- Select input to change page size of tables (this doesn't persist on refresh)
- added a xs variant to number field
- small style refactor for col explorer

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/components/data-table/pagination.tsx
+++ b/frontend/src/components/data-table/pagination.tsx
@@ -130,7 +130,7 @@ export const DataTablePagination = <TData,>({
             </SelectTrigger>
             <SelectContent>
               <SelectGroup>
-                <SelectLabel>Page size</SelectLabel>
+                <SelectLabel>Rows per page</SelectLabel>
                 {[...pageSizes].map((size) => {
                   const sizeStr = size.toString();
                   return (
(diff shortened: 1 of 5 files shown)


frontend/src/components/data-table/pagination.tsx
```diff
@@ -101,9 +102,27 @@ export const DataTablePagination = <TData,>({
...
 
+  const pageSize = table.getState().pagination.pageSize;
+
   return (
     <div className="flex flex-1 items-center justify-between px-2">
-      <div className="text-sm text-muted-foreground">{renderTotal()}</div>
+      <div className="flex items-center gap-2">
+        <div className="text-sm text-muted-foreground">{renderTotal()}</div>
+        <div className="flex items-center gap-1 text-xs">
+          <NumberField
```
mscolnick: i wonder if we should create opinionated "steps" e.g. 5/10/25/100, instead of a free-from input.

we could also do `[props.pageSize, 5, 10, 25, 100].sort()`

Light2Dark: yeah..my thinking was it's better to be more flexible than less.
I'm not super happy with the UI, I think next refactor may be to add a header

mscolnick: we offer the flexibility with `mo.ui.table(page_size)`. as a end-user, you may just want to quickly increase/decrease. 

im not super opinionated that it needs to be a dropdown vs a number, but maybe we can poll some other folks?

PR #5247, opened 2025-06-09
feat: Render mermaid-based (optimized) query plan for `pl.LazyFrame`

### LINKED ISSUE #3355: Polars LazyFrame `show_graph` has poor graphic quality in Marimo

manzt:
Replaces markdown/text-based *repr_html* with mermaid-based diagram. The main difference is that we display the optimized query plan, rather than prompting the user to call `show_graph(optimized=True)` or `explain(optimized=True)`, where `optimized` is already the default (like `collect()`). I think it's a more accurate/opinionated default.

image

Closes marimo-team/marimo#3355

### DIFF
--- a/marimo/_output/formatters/df_formatters.py
+++ b/marimo/_output/formatters/df_formatters.py
@@ -12,7 +12,7 @@ from marimo._output.formatters.formatter_factory import (
     FormatterFactory,
     Unregister,
 )
-from marimo._output.md import md
+from marimo._plugins.stateless.mermaid import mermaid
 from marimo._plugins.ui._impl import tabs
 from marimo._plugins.ui._impl.table import get_default_table_page_size, table
 from marimo._runtime.patches import patch_polars_write_json
@@ -35,9 +35,12 @@ def include_opinionated() -> bool:
     return True
 
 
-# adapted from https://github.com/pola-rs/polars/pull/20607
-def dot_to_mermaid(dot: str) -> str:
-    """Not comprehensive, only handles components of the dot language used by polars."""
+def polars_dot_to_mermaid(dot: str) -> str:
+    """Converts polars DOT query plan renderings to mermaid.
+
+    Adapted from https://github.com/pola-rs/polars/pull/20607
+    Note: Not comprehensive, only handles components of the dot language used by polars.
+    """
 
     edge_regex = r"(?P<node1>\w+) -- (?P<node2>\w+)"
     node_regex = r"(?P<node>\w+)(\s+)?\[label=\"(?P<label>.*)\"]"
@@ -75,12 +78,9 @@ class PolarsFormatter(FormatterFactory):
         return "polars"
 
     def register(self) -> Unregister | None:
-        from unittest.mock import patch
-
         import polars as pl
 
         from marimo._output import formatting
-        from marimo._output.hypertext import Html
 
         if not include_opinionated():
             return None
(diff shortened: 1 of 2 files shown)


manzt:
Lol sorry, experimented with generating the commit with claude code... meh. Curious what you think about testing for this. We could just snapshot some polars plans to mermaid outputs.

mscolnick (dismissed):

mscolnick (approved):

PR #5260, opened 2025-06-10
feat: Allow selecting optional "extras" for missing packages

manzt:
I'm fairly pleased with the UI for this, but parsing to and from package specifiers is a bit of a challenge. I wonder if having a richer datatype we send back and forth from the front end could be more robust.


marimo/_runtime/runtime.py
```diff
@@ -2476,30 +2479,31 @@ def missing_packages_hook(self, runner: cell_runner.Runner) -> None:
     async def install_missing_packages(
         self, request: InstallMissingPackagesRequest
     ) -> None:
-        """Attempts to install packages for modules that cannot be imported
-
-        Runs cells affected by successful installation.
-        """
-        assert self.package_manager is not None
-        if request.manager != self.package_manager.name:
```
dmadisetti, line 2484: Hmm, was this dead logic before?

manzt: Oops, no I moved down the assert, but accidentally removed this line.

frontend/src/components/editor/package-alert.tsx
```diff
@@ -362,10 +410,158 @@ interface PackageVersionSelectProps {
...
+    const response = await fetch(
+      `https://pypi.org/pypi/${cleanPythonModuleName(packageName)}/json`,
+      {
+        method: "GET",
+        signal: AbortSignal.timeout(5000), // 5s timeout
+      },
+    );
+    const pkgInfo = PyPiPackageResponse.parse(await response.json());
+    return pkgInfo.info.provides_extra || [];
+  }, [packageName]);
```
manzt: We could probably extract this into some kind of hook + context (since we duplicate the request to get the versions).

mscolnick: yea could be nice into a utils or hooks file. i generally like splitting up all the logic and dumb-components from each other.

mscolnick: also once it is in a utils file, claude is really good at generating some tests

manzt: good idea. i'll try that out in the morning

mscolnick:
i think removing `/test/` and `/dev/` for now would be ideal

frontend/src/components/editor/package-alert.tsx
```diff
@@ -362,10 +411,182 @@ interface PackageVersionSelectProps {
...
+    }
+  };
+
+  const canSelectExtras = !loading && !error;
+
+  return (
+    <div className="flex items-center max-w-72">
+      <span className="shrink-0">{packageName}</span>
+
+      {selectedExtras.length > 0 ? (
```
mscolnick, line 447: you might be able to simplify either as a `<Dropdown` or `<MultiSelect` with a custom header above. this could give you keyboard navigation for free.

i wouldn't waste too much time moving it though, this is still great otherwise

frontend/src/hooks/usePackageMetadata.ts
```diff
@@ -0,0 +1,118 @@
...
+interface SuccessResponse<T> {
+  loading: false;
+  data: T;
+  error: undefined;
+}
+
+type AsyncDataResponse<T> =
+  | LoadingResponse
+  | ErrorResponse
+  | SuccessResponse<T>;
```
manzt, line 47: I think this type better captures the states of the `useAsyncData` hook. Could update in a follow up PR there and elsewhere.

mscolnick: awesome - yea ive done something like this on other projects. definitely down to update `useAsyncData` to do the same

frontend/src/hooks/usePackageMetadata.ts
```diff
@@ -0,0 +1,116 @@
...
+
+function setCachedData(
+  normalizedPackageName: string,
+  data: PackageMetadata,
+): void {
+  packageCache.set(normalizedPackageName, {
+    data,
+    timestamp: Date.now(),
+  });
+}
```
mscolnick: if you want to make the generic util for this:

```ts
class Cache<T> {
  private cache = new Map<string, { data: T; timestamp: number }>();
  private readonly CACHE_DURATION: number;

  constructor(cacheDurationMs: number = 5 * 60 * 1000) {
    this.CACHE_DURATION = cacheDurationMs;
  }

  private cleanupExpiredEntries(): void {
    const now = Date.now();
    for (const [key, cached] of this.cache.entries()) {
      if (now - cached.timestamp > this.CACHE_DURATION) {
        this.cache.delete(key);
      }
    }
  }

  get(key: string): T | undefined {
    this.cleanupExpiredEntries();

    const cached = this.cache.get(key);
    if (!cached) {
      return undefined;
    }

    const now = Date.now();
    if (now - cached.timestamp > this.CACHE_DURATION) {
      this.cache.delete(key);
      return undefined;
    }

    return cached.data;
  }

  set(key: string, data: T): void {
    this.cache.set(key, {
      data,
      timestamp: Date.now(),
    });
  }

  clear(): void {
    this.cache.clear();
  }
}
```

mscolnick: we also have `LRUCache` which you can do and set to 10 or so.

i don't think a TTL is that needed (not sure the chances of a release while in a session)

manzt: Good idea. I extracted into a helper. Probably YAGNI with refreshing the browser tab and normal cache, but I know of long-live jupyter sessions.

PR #5272, opened 2025-06-11
Make async data states impossible to misrepresent in `useAsyncData`

manzt:
Migrates the async data state management to align with [React Query](https://tanstack.com/query/latest/docs/framework/react/overview) conventions, replacing the previous implementation that allowed ambiguous states.

## Changes

Updated to use React Query-like state patterns with explicit status types:
- **Pending**: `{ status: "pending", isPending: true, isFetching: true, data: undefined, error: undefined }`
- **Loading**: `{ status: "loading", isPending: false, isFetching: true, data: T, error: undefined }`
- **Error**: `{ status: "error", isPending: false, isFetching: false, data: undefined, error: Error }`
- **Success**: `{ status: "success", isPending: false, isFetching: false, data: T, error: undefined }`

## Side Effects

- Updated all components using `useAsyncData` to use new property names
- **file-viewer.tsx**: Implementation updated to work with new discriminated union types

This change shrinks the state space and _should_ make invalid states unrepresentable.

### DIFF
--- a/frontend/src/hooks/useAsyncData.ts
+++ b/frontend/src/hooks/useAsyncData.ts
@@ -25,9 +25,9 @@ interface LoadingResult<T> {
   isFetching: true;
 }
 
-interface ErrorResult {
+interface ErrorResult<T> {
   status: "error";
-  data: undefined;
+  data: undefined | T;
   error: Error;
   isPending: false;
   isFetching: false;
@@ -42,10 +42,10 @@ interface SuccessResult<T> {
 }
 
 const Result = {
-  error(e: Error): ErrorResult {
+  error<T>(e: Error, staleData?: T): ErrorResult<T> {
     return {
       status: "error",
-      data: undefined,
+      data: staleData,
       error: e,
       isPending: false,
       isFetching: false,
@@ -83,7 +83,7 @@ const Result = {
 export type AsyncDataResult<T> =
   | PendingResult
   | LoadingResult<T>
-  | ErrorResult
+  | ErrorResult<T>
   | SuccessResult<T>;
 
 export function combineAsyncData<T extends unknown[]>(
@@ -155,7 +155,7 @@ export function useAsyncData<T>(
 } {
   const [nonce, setNonce] = useState(0);
   const [result, setResult] = useState<
-    PendingResult | LoadingResult<T> | ErrorResult | SuccessResult<T>
+    PendingResult | LoadingResult<T> | ErrorResult<T> | SuccessResult<T>
   >(Result.pending());
 
   const asProps =
@@ -195,7 +195,8 @@ export function useAsyncData<T>(
         if (controller.signal.aborted) {
           return;
         }
-        setResult(Result.error(error));
+        // Carry over the previous data (if any)
+        setResult((prev) => Result.error(error, prev.data));
       });
 
     return () => {
--- a/frontend/src/hooks/useAsyncData.ts
+++ b/frontend/src/hooks/useAsyncData.ts
@@ -11,36 +11,55 @@ import useEvent from "react-use-event-hook";
 
 interface LoadingResponse {
   loading: true;
+  reloading: false;
   data: undefined;
+  error: undefined;
+}
+
+interface ReloadingResponse<T> {
+  loading: true;
+  reloading: true;
+  data: T;
   error: undefined;
 }
 
 interface ErrorResponse {
   loading: false;
+  reloading: false;
   data: undefined;
   error: Error;
 }
 
 interface SuccessResponse<T> {
   loading: false;
+  reloading: false;
   data: T;
   error: undefined;
 }
 
 function isSuccess<T>(
-  x: LoadingResponse | ErrorResponse | SuccessResponse<T>,
+  x:
+    | LoadingResponse
+    | ReloadingResponse<T>
+    | ErrorResponse
+    | SuccessResponse<T>,
 ): x is SuccessResponse<T> {
   return !x.loading && x.error === undefined;
 }
 
 function isError(
-  x: LoadingResponse | ErrorResponse | SuccessResponse<unknown>,
+  x:
+    | LoadingResponse
+    | ReloadingResponse<unknown>
+    | ErrorResponse
+    | SuccessResponse<unknown>,
 ): x is ErrorResponse {
   return !x.loading && x.data === undefined;
 }
 
 export type AsyncDataResponse<T> =
   | LoadingResponse
+  | ReloadingResponse<T>
   | ErrorResponse
   | SuccessResponse<T>;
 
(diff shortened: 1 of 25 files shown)


frontend/src/hooks/useAsyncData.ts
```diff
@@ -1,30 +1,90 @@
...
 
-interface AsyncDataResponse<T> {
-  data: T | undefined;
-  loading: boolean;
-  error: Error | undefined;
-  setData: Dispatch<SetStateAction<T | undefined>>;
-  reload: () => void;
+interface LoadingResponse {
+  loading: true;
+  data: undefined;
```
mscolnick, line 14: i think we used to show the previous data for "reloads". you might be able to see this on `mo.ui.table`. would it be possible to make this `data: T | undefined`;

or we could add a `ReloadingResponse`, but that may be overkill.

we could still add `reloading: boolean`; on `LoadingResponse` if we want. (i think in some places we infer reloading if `data !== undefined`

manzt: I don't think it's overkill. It will stay isolated to this function and the helper, and would be useful to capture the extra variant in usage:

```ts
const { loading, reloading, error, data, reload } = useAsyncData<T>(...);

if (reloading) {
  data; // T
} else if (loading) {
  data; // undefined
} else if (error) {
  error; // Error
} else {
  data; // T
}
```

manzt: You of course don't need to check `reloading`, but it's an option if you care about that state.

manzt: Ok, I might have gone a bit far with the changes... but I looked at React Query for API convention. If we ever wanted to migrate to that (off our own custom utility) it should basically be a drop-in replacement for the API we support in this PR now.

The core idea now is that we can check `isPending` for an initial load and `isFetching` for any time we care about `T | undefined`. I went through the code base, and the pattern I found is that most times were we use the `refetch` (formerly `reload`) is where we care about the `isFetching` nuance. Everywhere else, we align with the React Query convention:

> or most queries, it's usually sufficient to check for the isPending state, then the [error], then finally, assume that the data is available and render the successful state:

frontend/src/hooks/useAsyncData.ts
```diff
@@ -1,31 +1,135 @@
...
+  status: "loading";
+  data: T;
+  error: undefined;
+  isPending: false;
+  isFetching: true;
+}
+
+interface ErrorResult {
+  status: "error";
+  data: undefined;
```
mscolnick: the last regression (and happy to be convinced that it is ok) is that if you get an error, you still may want to show the previous `data`. for example in a table, you may want to show the old data but put the error in the banner (showing both)

frontend/src/hooks/useAsyncData.ts
manzt: Totally fair if this migration doesn’t feel worth it or the benefits aren’t obvious. I mostly went down this rabbit hole to better understand the frontend, and it ended up touching more than I expected. No worries at all if we don’t want to take it in this direction :)

mscolnick: maybe you already looked at this, but this is what react-query does. not that we need all, but could be good to align (you might have, haven't verified)

```

- `status: QueryStatus`
  - Will be:
    - `pending` if there's no cached data and no query attempt was finished yet.
    - `error` if the query attempt resulted in an error. The corresponding `error` property has the error received from the attempted fetch
    - `success` if the query has received a response with no errors and is ready to display its data. The corresponding `data` property on the query is the data received from the successful fetch or if the query's `enabled` property is set to `false` and has not been fetched yet `data` is the first `initialData` supplied to the query on initialization.
- `isPending: boolean`
  - A derived boolean from the `status` variable above, provided for convenience.
- `isSuccess: boolean`
  - A derived boolean from the `status` variable above, provided for convenience.
- `isError: boolean`
  - A derived boolean from the `status` variable above, provided for convenience.
- `isLoadingError: boolean`
  - Will be `true` if the query failed while fetching for the first time.
- `isRefetchError: boolean`
  - Will be `true` if the query failed while refetching.
- `data: TData`
  - Defaults to `undefined`.
  - The last successfully resolved data for the query.
- `dataUpdatedAt: number`
  - The timestamp for when the query most recently returned the `status` as `"success"`.
- `error: null | TError`
  - Defaults to `null`
  - The error object for the query, if an error was thrown.
- `errorUpdatedAt: number`
  - The timestamp for when the query most recently returned the `status` as `"error"`.
- `isStale: boolean`
  - Will be `true` if the data in the cache is invalidated or if the data is older than the given `staleTime`.
- `isPlaceholderData: boolean`
  - Will be `true` if the data shown is the placeholder data.
- `isFetched: boolean`
  - Will be `true` if the query has been fetched.
- `isFetchedAfterMount: boolean`
  - Will be `true` if the query has been fetched after the component mounted.
  - This property can be used to not show any previously cached data.
- `fetchStatus: FetchStatus`
  - `fetching`: Is `true` whenever the queryFn is executing, which includes initial `pending` as well as background refetches.
  - `paused`: The query wanted to fetch, but has been `paused`.
  - `idle`: The query is not fetching.
  - see [Network Mode](../../guides/network-mode.md) for more information.
- `isFetching: boolean`
  - A derived boolean from the `fetchStatus` variable above, provided for convenience.
- `isPaused: boolean`
  - A derived boolean from the `fetchStatus` variable above, provided for convenience.
- `isRefetching: boolean`
  - Is `true` whenever a background refetch is in-flight, which _does not_ include initial `pending`
  - Is the same as `isFetching && !isPending`
- `isLoading: boolean`
  - Is `true` whenever the first fetch for a query is in-flight
  - Is the same as `isFetching && isPending`
```

mscolnick (dismissed):

frontend/src/hooks/useAsyncData.ts
mscolnick: ah sorry, only read the last comment, didn't see your " but I looked at React Query for API convention". 

im definitely down to align with react query, and eventually move off of it (didn't need the complexity at the time)

frontend/src/hooks/useAsyncData.ts
manzt: Oh yeah, you're right. To my best estimation, we don't rely on this right now, but React Query allows for errors to have the `data` filled with `placeholderData` (carried over from some previous state). I updated our `ErrorResult` to carry over a previous data (if any). It allows for this kind of pattern:


```ts
if (isPending) {
   return <Loader />
}

if (error) {
  return <ErrorComponent error={error} maybeData={data} />
}

return <Component data={data} />
```

frontend/src/hooks/useAsyncData.ts
manzt: Cool, I think we could merge this. API is very close to React Query... if there is some feature we think we are missing, then it should be super straightforward to switch over.

PR #5281, opened 2025-06-12
feat: enhance `ImportError` handling for better package suggestions

manzt:
Trying to handle all possible import patterns is a bit pathological—there’s no strong convention across packages. For instance, `fsspec` is [all over the place](https://github.com/fsspec/filesystem_spec/blob/26f1ea75351e39a80b29b27bea792351f3e8da9f/fsspec/registry.py#L62-L230). There’s a loose pattern, but it’s not consistent enough to catch reliably.

That said, I think we now have a solid approach for the common case where an error is explicitly re-raised from a failed import:

```py
except ImportError as err:
    raise ImportError("You can install this another way") from err
```

Moreover, I set up a few "passes" for package extractors (i.e., functions of the form `(error_message: str) -> list[str] | None`) that can eagerly "handle" an error. This makes it easy to add special cases, or even incorporate fuzzier logic down the line (e.g., if the user has AI API keys enabled).

The only other somewhat high-confidence "extractor" I can think of is a regex that catches a quoted word within the import error.

### DIFF
--- a/marimo/_runtime/packages/import_error_extractors.py
+++ b/marimo/_runtime/packages/import_error_extractors.py
@@ -24,6 +24,12 @@ def extract_missing_module_from_cause_chain(
     return None
 
 
+def strip_quotes(s: str) -> str:
+    if len(s) >= 2 and s[0] == s[-1] and s[0] in {"'", '"'}:
+        return s[1:-1]
+    return s
+
+
 def extract_packages_from_pip_install_suggestion(
     message: str,
 ) -> list[str] | None:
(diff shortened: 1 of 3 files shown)


marimo/_runtime/packages/import_error_extractors.py
```diff
@@ -0,0 +1,105 @@
...
+            isinstance(current, ModuleNotFoundError)
+            and hasattr(current, "name")
+            and current.name
+        ):
+            return current.name
+        current = current.__cause__
+    return None
+
+
+def extract_packages_from_pip_install_suggestion(
```
manzt, line 33: One thing I'm not totally sure about here is what if a package decides to suggest pip installing itself with extras. Would we want to suggest that to the end user? or do something more structured like look and see where the error is coming from?

mscolnick: i think that is actually a good feature, since it will fix its own inline script metadata

mscolnick (approved):

manzt:
It seems like the failing tests are unrelated. I'm going to merge.

PR #5284, opened 2025-06-12
improvement: connecting indicator and readonly tooltip

mscolnick:
Screenshot 2025-06-12 at 2 06 13 PM
Screenshot 2025-06-12 at 2 25 02 PM

### DIFF
--- a/frontend/src/core/codemirror/readonly/extension.ts
+++ b/frontend/src/core/codemirror/readonly/extension.ts
@@ -1,9 +1,13 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 import { Compartment, EditorState, StateEffect } from "@codemirror/state";
-import { type EditorView, ViewPlugin } from "@codemirror/view";
+import { ViewPlugin } from "@codemirror/view";
 import { WebSocketState } from "../../websocket/types";
 import { connectionAtom } from "../../network/connection";
 import type { createStore } from "jotai";
+import {
+  isAppConnected,
+  isAppConnecting,
+} from "@/core/websocket/connection-utils";
 
 /**
  * State effect for updating readonly status based on connection
@@ -62,8 +66,10 @@ let hideIndicatorTimeout: number | null = null;
  * When connected, the editor becomes editable.
  */
 export function dynamicReadonly(store: ReturnType<typeof createStore>) {
+  const getConnectionState = () => store.get(connectionAtom).state;
+
   // Initial readonly state - start as readonly until connected
-  let isReadonly = true;
+  let isReadonly = !isAppConnected(getConnectionState());
 
   const connectingIndicatorPlugin = ViewPlugin.define((view) => {
     let listenersAttached = false;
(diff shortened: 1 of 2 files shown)


dmadisetti:
Is typing still allowed?

mscolnick:
@dmadisetti not while it's connecting

frontend/src/core/codemirror/readonly/extension.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    function onUserInput() {
+      showNearCaret();
+    }
+
+    function attachListeners() {
+      if (!listenersAttached) {
+        dom.addEventListener("keydown", onUserInput, true);
+        dom.addEventListener("paste", onUserInput, true);
```
manzt, line 99: I prefer the options object because it's slightly more explicit what the boolean is. 

```suggestion
        dom.addEventListener("keydown", onUserInput, { capture: true });
        dom.addEventListener("paste", onUserInput, { capture: true });
```

frontend/src/core/codemirror/readonly/extension.ts
```diff
@@ -0,0 +1,187 @@
...
+        listenersAttached = true;
+      }
+    }
+    function detachListeners() {
+      if (listenersAttached) {
+        dom.removeEventListener("keydown", onUserInput, true);
+        dom.removeEventListener("paste", onUserInput, true);
```
manzt, line 106: ditto

```suggestion
        dom.removeEventListener("keydown", onUserInput, { capture: true });
        dom.removeEventListener("paste", onUserInput, { capture: true });
```

frontend/src/components/editor/alerts/connecting-alert.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const isConnecting = useAtomValue(isConnectingAtom);
+
+  return (
+    <FloatingAlert title="Connecting" show={isConnecting} delayMs={DELAY_MS}>
+      <div className="flex items-center gap-2">
+        <Spinner className="h-4 w-4" />
+        <p>Establishing a connection to a marimo runtime...</p>
+      </div>
+    </FloatingAlert>
+  );
```
manzt, line 20: Would we want to present different states here, or just connecting?

manzt (approved):
Awesome. Minor nit with some event listener stuff. Seems like we have the machinery now to change the hover text if we want to show other kinds of progress than "connecting..."

manzt:
Oops, sorry at airport and was without connection.

dmadisetti:
@manzt these are live if you are interested, wanted the dev release to check things out. Might follow up with your comments in another PR

frontend/src/components/editor/alerts/connecting-alert.tsx
mscolnick: for now just connecting. what states did you have in mind? it might be nice to show not connected or maybe with a reconnect button (if/when we have that flow, currently we don't)

frontend/src/core/codemirror/readonly/extension.ts
mscolnick: my toxic trait is i ignore nits like this (something lintable) and instead add rules to fix it once and for all 😂 https://github.com/marimo-team/marimo/pull/5293

PR #5300, opened 2025-06-13
perf: cell range perf

mscolnick:
perf: cell range perf


frontend/src/components/data-table/range-focus/use-cell-range-selection.ts
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  const handleCellMouseOver = useEvent(
+    (e: React.MouseEvent, cell: Cell<TData, unknown>) => {
+      if (e.buttons !== 1) {
+        return;
+      }
+
+      // Check if selecting without subscribing to state
+      const isCurrentlySelecting = isSelectingAtom.init;
```
Light2Dark: `const isCurrentlySelecting = store.get(isSelectingAtom);` -> `const isCurrentlySelecting = store.get(isSelectingAtom);`

we need the updated value when dragging over cells

mscolnick: updated using a reducer and added tests

Issue #4726, opened 2025-04-29
`ruff` listed as mandatory dependency, should be optional


falkaer:
### Describe the bug

Hi,

Ruff is listed under dependencies in pyproject.toml, despite being described as an optional dependency in documentation - when using uv sync to install marimo, this forces the installation of a PyPi-distributed ruff build.

Concretely this causes issues because the PyPI ruff will be preferred to a system ruff when the Python environment is activated, which breaks my editor on NixOS because the PyPI ruff is dynamically linked, which is not supported.

### Environment

<details>

```
{
  "marimo": "0.13.2",
  "OS": "Linux",
  "OS Version": "6.13.8-arch1-1",
  "Processor": "",
  "Python Version": "3.11.11",
  "Binaries": {
    "Browser": "--",
    "Node": "v23.9.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.33.0",
    "packaging": "23.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pymdown-extensions": "10.14.3",
    "pyyaml": "6.0.1",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.13.0",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "nbformat": "5.10.4",
    "pandas": "2.1.1",
    "pyarrow": "19.0.1",
    "pycrdt": "0.11.1",
    "pytest": "7.4.4",
    "ruff": "0.11.2"
  },
  "Experimental Flags": {}
}
```

</details>

dmadisetti:
FWIW, marimo is also in nixpkgs- but yes this is annoying

PR #5302, opened 2025-06-13
feat: add `ty` LSP

### LINKED ISSUE #1163: Add pyright type-checking
### Description

Love the inline type-checking in VSCode, and would love to see it here as well. 

### Suggested solution

Add pyright type-checking and underlining.

### Alternative

_No response_

### Additional context

_No response_

winstxnhdw:
## 📝 Summary

This PR integrates Astral's `ty` as an LSP in `marimo`.

image

image

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

// cc: @nojaf


winstxnhdw:
I am not sure, but I think the Marimo frontend LSP client works by creating a websocket client to the Python server. Since `ty` doesn't have websocket support, I will probably need to implement some bridge? There's also [lsp-ws-proxy](https://github.com/qualified/lsp-ws-proxy) we can use to do this for us but I am not sure how are we going to handle that external dependency, or would the team prefer for something like this to be written in-house?

mscolnick:
@winstxnhdw awesome progress so far!

1. we have a node version of lsp-ws-proxy, in the `/lsp` directory. we could run `ty` from there. 
2. we could make `lsp-ws-proxy` an optional dep and require it when `ty` is enabled. that repo does look unmaintained / hasn't had changes
3. we could add our own ws endpoint to talk with ty (and other LSPs) via `stdio`(basically roll our own `lsp-ws-proxy`). it would improve our pylsp integration too probably.

winstxnhdw:
Since we already have one for Copilot, I'll probably try reusing that. Once we verify that everything works, I am happy to submit another PR that will introduce our own WebSocket proxy in Rust. Maybe we can finally get rid of `node` as a dependency (:

winstxnhdw:
Hey @mscolnick, I am thinking now that we can support multiple LSPs, should we have a drop-down menu similar to the AI Code Completion providers for LSPs?

image

mscolnick:
> I am thinking now that we can support multiple LSPs, should we have a drop-down menu similar to the AI Code Completion providers for LSPs?

It may be possible in the future to run multiple at a time (one for completion and another for type-checking). We don't fully support this at the moment (the backend may need to do some aggregation, or multiple connections in the frontend).

I think for now (until we add more than two LSPs), you can just make this another checkbox toggle. If there is a conflict with the pylsp LSP, then we can make sure it is clear which takes precendence (in the docs and editor)

frontend/src/core/runtime/runtime.ts
```diff
@@ -135,13 +135,7 @@ export class RuntimeManager {
   /**
    * The URL of the copilot server.
    */
-  getLSPURL(lsp: "pylsp" | "copilot"): URL {
```
mscolnick, line 138: i think this needs to be reverted. copilot should not include query params

winstxnhdw: But we do not populate the query params anyways, so it seems redundant?

frontend/src/core/runtime/runtime.ts
```diff
@@ -135,13 +135,7 @@ export class RuntimeManager {
...
-    if (lsp === "copilot") {
-      // For copilot, don't include any query parameters
-      const url = this.formatWsURL(`/lsp/${lsp}`);
-      url.search = "";
-      return url;
-    }
+  getLSPURL(lsp: "pylsp" | "copilot" | "ty"): URL {
```
mscolnick, line 138: `formatWsURL` might. and its safer and clearer anyways. this was added because of a bug

winstxnhdw:
I am trying to push the regenerated `openapi` scripts but the pre-commit hook is stopping me because of a lint error in one of them.

frontend/src/core/codemirror/language/languages/python.ts
```diff
@@ -191,12 +192,22 @@ export class PythonLanguageAdapter implements LanguageAdapter<{}> {
...
       if (lspConfig?.pylsp?.enabled && hasCapability("pylsp")) {
-        const client = lspClient(lspConfig);
+        clients.push(pylspClient(lspConfig));
+      }
+      if (lspConfig?.ty?.enabled && hasCapability("ty")) {
+        clients.push(tyLspClient(lspConfig));
+      }
+
+      if (clients.length > 0) {
+        const client = new FederatedLanguageServerClient(clients);
```
mscolnick: i know i wrote this, but maybe in case there are weird edge cases with federations, we could do:

```
const client = clients.length === 0 ? clients[0] : new FederatedLanguageServerClient(clients);
```

winstxnhdw: How would we access an index of `clients` if `clients` is empty?

mscolnick: Sorry I meant when length == 1

winstxnhdw:
CI failure seems unrelated. Also, waiting on this [PR](https://github.com/marimo-team/codemirror-languageserver/pull/14) to be merged.

frontend/src/components/app-config/user-config-form.tsx
```diff
@@ -1341,43 +1383,46 @@ export const UserConfigForm: React.FC = () => {
...
-                      <FormLabel>Model</FormLabel>
-                      <FormControl>
-                        <Input
-                          list="ai-model-datalist"
-                          data-testid="ai-model-input"
-                          className="m-0 inline-flex"
-                          placeholder="gpt-4-turbo"
-                          {...field}
+                render={({ field }) => {
+                  const modelDatalistId = React.useId();
```
mscolnick: there was probably a warning you were trying to fix. its ok to have that warning. we should remove `const modelDatalistId = React.useId()` instead since it might break react renders

mscolnick: this `render={({ field }) => {` may have an `id` in `render={({ field, id }) => {`

marimo/_messaging/ops.py
```diff
@@ -418,11 +418,13 @@ class CompletedRun(Op):
 class KernelCapabilities:
     terminal: bool = False
     pylsp: bool = False
+    ty: bool = False
 
     def __post_init__(self) -> None:
         # Only available in mac/linux
         self.terminal = not is_windows() and not is_pyodide()
         self.pylsp = DependencyManager.pylsp.has()
+        self.ty = DependencyManager.ty.has()
```
dmadisetti, line 427: So this will only work when ty is a python dep? Not an issue, just want to make sure I understand. Any reason we wouldn't want to check ty?

Maybe we want to make a package group to go with `marimo[lsp]`

mscolnick: `ty` is still quite very early, so lets not include it in `marimo[lsp]`

winstxnhdw: If it were up to me, I'd make `ty` the default type checker and ship it with `marimo` ootb. It's fast and it supports incremental type checking. However, here I just followed what `pylsp` did. Happy to change it to whatever the team wants.

mscolnick: we can consider this once `ty` is stable

dmadisetti: Pre-packaged binaries do end up causing issues on some systems (#4726). This is what we do for ruff:

https://github.com/marimo-team/marimo/blob/16d779c2d058cd0bd043bac9834ba94b16614c5f/marimo/_utils/formatter.py#L19-L29

But I think this is fine. It's beta, and we can take the code churn on ourselves. Thanks again for your work

PR #5325, opened 2025-06-17
fix: more Supsense in more places

mscolnick:


frontend/src/components/scratchpad/scratchpad.tsx
```diff
@@ -187,17 +187,19 @@ export const ScratchPad: React.FC = () => {
...
-                basicSetup={{
-                  highlightActiveLine: false,
-                  highlightActiveLineGutter: false,
-                }}
-                value={item.trim()}
-                editable={false}
-                readOnly={true}
-              />
+              <Suspense>
+                <LazyAnyLanguageCodeMirror
```
dmadisetti, line 191: Shouldn't suspense be on the component level instead?

mscolnick: it needs to go above the component in the tree that is being lazily-loaded

PR #5331, opened 2025-06-17
fix: show wasm always always connected

mscolnick:
Wasm doesn't have a health endpoint so its always connected


dmadisetti:
Yeah fair, we could have our own identifier for WASM too though

PR #5335, opened 2025-06-18
fix: reasonable session and notebook merge

dmadisetti:
## 📝 Summary

Using off the shelf edit distance to properly merge session and notebook.

Tests pass, but might have to use a md5 package (that's the right hash right?)

@akshayka OR @mscolnick


mscolnick:
can we maybe start off with something simpler like #5337.

1. I don't think we need edit distance
2. we can just hash on the backend so its consistent with the session.json code hash

mscolnick (approved):

PR #5343, opened 2025-06-18
Add new module name mapping

hannahker:
## 📝 Summary

Adding a mapping for a package that my team uses, which would be great to auto-install. 

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/marimo/_runtime/packages/module_name_to_pypi_name.py
+++ b/marimo/_runtime/packages/module_name_to_pypi_name.py
@@ -580,6 +580,7 @@ def module_name_to_pypi_name() -> dict[str, str]:
         "novaclient": "python-novaclient",
         "oauth2_provider": "alauda-django-oauth",
         "oauth2client": "oauth2client",
+        "ocha_stratus": "ocha-stratus",
         "odf": "odfpy",
         "ometa": "Parsley",
         "openid": "python-openid",


hannahker:

mscolnick:
i think we automatically convert `ocha_stratus` to `ocha-stratus` (replacing `_`). did you try this and it does not work?

hannahker:
I've been able to get it to work locally (ie. `marimo edit --sandbox`), but am struggling to troubleshoot issues when exporting to a WASM HTML notebook that I'd like to deploy via GH pages. 

Once I convert to WASM and then open via running a local server, my notebook is missing most of its functionality and I can see the following in the browser console:

```
Loading from micropip: ['datasources', 'python-dotenv', 'matplotlib.pyplot', 'ocha-stratus', 'utils']
[STDERR] [12:47:07 PM] (Hbol) {"msg":"Expected an image object, but got <class 'marimo._runtime.runtime.URLPath'> instead.","exception_type":"ValueError","raising_cell":null,"type":"exception"} [index-D3XtisvU.js:67:359070](http://[::1]:8000/assets/index-D3XtisvU.js)
[STDERR] [12:47:08 PM] (lEQa) Matplotlib is building the font cache; this may take a moment.
[index-D3XtisvU.js:67:359070](http://[::1]:8000/assets/index-D3XtisvU.js)
[STDERR] [12:47:09 PM] (lEQa) {"msg":"No module named 'ocha_stratus'","exception_type":"ModuleNotFoundError","raising_cell":null,"type":"exception"} [index-D3XtisvU.js:67:359070](http://[::1]:8000/assets/index-D3XtisvU.js)
[STDERR] [12:47:09 PM] (ZBYS) {"msg":"An ancestor raised an exception (ModuleNotFoundError): ","exception_type":"ModuleNotFoundError","raising_cell":"lEQa","type":"exception"} [index-D3XtisvU.js:67:359070](http://[::1]:8000/assets/index-D3XtisvU.js)
[STDERR] [12:47:09 PM] (aLJB) {"msg":"An ancestor raised an exception (ModuleNotFoundError): ","exception_type":"ModuleNotFoundError","raising_cell":"lEQa","type":"exception"} [index-D3XtisvU.js:67:359070](http://[::1]:8000/assets/index-D3XtisvU.js)
```

I figured it might be a mapping issue with the name (hence the PR), but now I'm wondering if it's more likely missing from [the pyodide list](https://pyodide.org/en/stable/usage/packages-in-pyodide.html). This is my first time with both marimo and WASM though, so totally open to any other debugging ideas. 

Happy to shift this to a issue instead.

hannahker:
Ah I think I figured out the source of my issue. Would love confirmation though. 

We use `ocha-stratus` to access data from our database, and so `psycopg2-binary` is a downstream dependency. My understanding is that this is NOT downloadable from `micropip` because of the limitations around threaded applications. So in effect, we are not able to deploy any notebooks that get data from our database via the WASM/GH pages path.

akshayka:
> My understanding is that this is NOT downloadable from micropip because of the limitations around threaded applications. So in effect, we are not able to deploy any notebooks that get data from our database via the WASM/GH pages path.

If `psycopg2-binary` is a dependency of `ocha-stratus`, and if `psycopg2-binary` doesn't work in Pyodide (which you seem to have verified), then yes your conclusion would be correct. Sorry that you're unable to use this in WASM. 

Pyodide has a process for requesting packages be added to their distribution (via their GitHub issues), but if it's due to threading then this would be tricky to add.

hannahker:
Yes after some more research it looks like a no-go. As I'm understanding, WASM will not support database connections on the whole -- makes sense after thinking about it a bit more, as full browser-side exposure to a database is a big security concern. I'd assume this also impacts other publishing pathways that rely on WASM/Pyodide (like quarto). 

It would be great to have some explanation and disclaimer about this in the documentation. Most of my notebooks will have database connections and so this is a big rethink on how I can best share and publish my notebooks. I'd imagine that many others migrating from Jupyter notebooks would also not have the background understanding of WASM to anticipate this limitation.

mscolnick:
@hannahker - we could probably add the disclaimer here: https://docs.marimo.io/guides/wasm/#limitations

hannahker:
Yes that would fit! Although, there's a good chance many users could miss that, so you might want to link to those limitations across the other places that WASM notebooks are references (eg [here](https://docs.marimo.io/guides/exporting/#export-to-wasm-powered-html)).

akshayka:
I'm closing this PR in light of the above discussion. #5439 improves our documentation on WASM's package limitations.

PR #5344, opened 2025-06-18
fix: remove autoscroll

mscolnick:
This autoscroll logic might be old and unused since we still get auto-scroll via the `scrollKey` logic in the cell action/reducer. 

This autoscroll was triggered incorrectly when connection is delayed or RTC, but no longer seems needed.


dmadisetti (approved):
LGTM, does this remove all autoscroll though?

mscolnick:
@dmadisetti nope it does not. 

> we still get auto-scroll via the scrollKey logic in the cell action/reducer.

PR #5349, opened 2025-06-18
improvement: support uploading folders in the sidebar (keep heirarchy)

mscolnick:


frontend/src/components/editor/file-tree/upload.tsx
```diff
@@ -58,3 +68,34 @@ export function useFileExplorerUpload(options: DropzoneOptions = {}) {
...
+    return file.webkitRelativePath;
+  }
+  if ("path" in file && typeof file.path === "string") {
+    return file.path;
+  }
+  if ("relativePath" in file && typeof file.relativePath === "string") {
+    return file.relativePath;
+  }
+  return undefined;
+}
```
manzt: Think we can safely update the signature here, since this utility only can return valid `FilePath`s. That way it's kind of the only "safe" way to get a our branded type, without making one our own.


```suggestion
function getPath(file: File): FilePath | undefined {
  if (file.webkitRelativePath) {
    return file.webkitRelativePath as FilePath;
  }
  if ("path" in file && typeof file.path === "string") {
    return file.path as FilePath;
  }
  if ("relativePath" in file && typeof file.relativePath === "string") {
    return file.relativePath as FilePath;
  }
  return undefined;
}
```

frontend/src/components/editor/file-tree/upload.tsx
```diff
@@ -58,3 +68,34 @@ export function useFileExplorerUpload(options: DropzoneOptions = {}) {
...
+
+/**
+ * Strip leading slashes from a path.
+ *
+ * TODO: this may not support windows paths.
+ */
+function stripLeadingSlash(path: string | undefined): string | undefined {
```
manzt:
```suggestion
function stripLeadingSlash(path: FilePath | undefined): FilePath | undefined {
```

frontend/src/components/editor/file-tree/upload.tsx
```diff
@@ -40,14 +40,24 @@ export function useFileExplorerUpload(options: DropzoneOptions = {}) {
     },
     onDrop: async (acceptedFiles) => {
       for (const file of acceptedFiles) {
+        // We strip the leading slash since File.path can return
+        // `/path/to/file`.
+        const filePath = stripLeadingSlash(getPath(file));
+        let directoryPath = "" as FilePath;
+        if (filePath) {
+          directoryPath = PathBuilder.guessDeliminator(filePath).dirname(
+            filePath as FilePath,
```
manzt: Think we can remove this cast if we update the utility function signatures below.

```suggestion
            filePath,
```

manzt (approved):
Looks great! Just a couple of suggestions with helper functions for branded types.

PR #5350, opened 2025-06-19
improvement: better sort order for mo.ui.file_browser()

mscolnick:


marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -218,33 +218,47 @@ def _list_directory(
                 "Navigation is restricted; navigating to a "
                 "parent of initial path is not allowed."
             )
-
+        folders: list[TypedFileBrowserFileInfo] = []
         files: list[TypedFileBrowserFileInfo] = []
+
         for file in path.iterdir():
```
dmadisetti, line 224: What is the list order of iterdir? Is it automatically alphabetized?

Think there's a weird edgecase where `> self._limit` starts to cause weird things.

e.g.

c
d
--- # limit
a
b

`a` `b` are never seen, despite maybe being first

Maybe this is fine. I think GH prompts that list orderings may be invalid for > 1000 files- but also maybe not a super common occurrence

marimo/_utils/files.py
```diff
@@ -0,0 +1,12 @@
+import re
+from typing import Union
+
+
+def natural_sort(filename: str) -> list[Union[int, str]]:
```
dmadisetti, line 5: nit: natural_order

Not sort

marimo/_utils/files.py
```diff
@@ -0,0 +1,12 @@
...
+    def convert(text: str) -> Union[int, str]:
+        return int(text) if text.isdigit() else text.lower()
+
+    def alphanum_key(key: str) -> list[Union[int, str]]:
```
dmadisetti, line 9: Surprised this works, I didn't know lists were comparable, thought only tuples

dmadisetti (approved):
Test LGTM. Just nits, but it's fine

PR #5358, opened 2025-06-19
feat: Add uv-managed "tree" view for package manager

manzt:
This PR introduces a new tree-based interface for managing dependencies in uv-managed environments, shifting from the traditional pip-style list interface to a more structured approach that aligns with uv's (high-level) dep management philosophy.

The tree view automatically appears when marimo detects a uv-managed environment (either within "sandbox" or within a uv project) and provides a hierarchical visualization of package dependencies. Unlike the existing list view where users directly manipulate the Python environment, the tree view encourages users to manage only their top-level dependencies (specifying their high level requirements) while viewing the complete dependency graph.

The interface distinguishes between two operational modes: "sandbox" mode for script-level dependency management and "project" mode when working within a uv-managed project. Users can "Add", "Remove", and "Upgrade" top-level packages through this interface, which translates to `uv add` and `uv remove` commands rather than direct environment manipulation via `uv pip install/uninstall`.

We hide this view/capabilities in unsupported package managers, but is the default for uv.

### Sandbox mode

https://github.com/user-attachments/assets/d4f00ba6-d635-4825-b20a-4baee7bb4736

### Project mode  

https://github.com/user-attachments/assets/9657e5ba-29e4-4017-955a-10df6f5d81b0

### TODO:

- [x] Add some tests for the parsing of various outputs of `uv tree`. Ideally we could use uv's own structured output in the future (https://github.com/astral-sh/uv/issues/12541, https://github.com/astral-sh/uv/issues/4711)


frontend/src/components/editor/chrome/panels/packages-panel.tsx
```diff
@@ -157,8 +196,63 @@ export const PackagesPanel: React.FC = () => {
...
+              className="items-center border px-2 py-0.5 text-xs transition-colors focus:outline-none focus:ring-2 focus:ring-ring focus:ring-offset-2 text-foreground rounded-sm text-ellipsis block overflow-hidden max-w-fit font-medium"
+              title={treeData?.tree?.name === "<root>" ? "sandbox" : "project"}
+            >
+              {treeData?.tree?.name === "<root>" ? "sandbox" : "project"}
+            </div>
+            {treeData?.tree?.name !== "<root>" && treeData?.tree?.name && (
+              <span className="text-xs text-muted-foreground">
+                {treeData.tree.name}
+                {treeData?.tree?.version && ` v${treeData.tree.version}`}
+              </span>
```
mscolnick, line 225: nit: id maybe pull out some of this into the render portion of this function.

e.g. 
```
const name = treeData?.tree?.name;
const version = treeData?.tree?.version;
```

mscolnick (dismissed):

tests/_utils/uv_tree.py
```diff
@@ -0,0 +1,89 @@
...
+@pytest.mark.skipif(UV_BIN is None, reason="requires uv executable.")
+def test_complex_project_tree(tmp_path: pathlib.Path) -> None:
+    uv(["init", "blah"], cwd=str(tmp_path))
+    project_dir = tmp_path / "blah"
+    uv(["add", "anywidget", "marimo"], cwd=str(project_dir))
+    uv(["add", "--dev", "pytest"], cwd=str(project_dir))
+    uv(["add", "--optional", "bar", "pandas"], cwd=str(project_dir))
+    raw = uv(["tree", "--no-dedupe"], cwd=str(project_dir))
+    tree = parse_uv_tree(raw)
+    snapshot_test("complex_project_tree.json", serialize(tree))
```
manzt: @mscolnick what do you think of this testing? I think this would catch drift and also see if we are resiliant to stuff int he output.

mscolnick (dismissed):

frontend/src/core/wasm/bridge.ts
```diff
@@ -514,6 +514,11 @@ export class PyodideBridge implements RunRequests, EditRequests {
     return response;
   };
 
+  getDependencyTree = async () => {
+    // WASM doesn't support dependency trees yet
+    return { tree: undefined };
```
dmadisetti, line 519: Just wondering what the limitation is. Because micropip vs uv?

manzt: the tree is parsed from the output of `uv tree`, so it is uv-specific.

PR #5359, opened 2025-06-19
improvement: simplify focus and delay on start

mscolnick:
This removes the delay to show cells when in edit mode (only needed for run mode).

### DIFF
--- a/frontend/src/components/editor/renderers/vertical-layout/__tests__/useDelayVisibility.test.ts
+++ b/frontend/src/components/editor/renderers/vertical-layout/__tests__/useDelayVisibility.test.ts
@@ -27,26 +27,44 @@ describe("useDelayVisibility", () => {
 
     expect(result.current.invisible).toBe(true);
 
-    // Advance timers by the calculated delay (5-1)*15 = 60ms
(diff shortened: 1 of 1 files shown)


frontend/src/components/editor/renderers/vertical-layout/useDelayVisibility.ts
```diff
@@ -1,109 +1,27 @@
...
 export function useDelayVisibility(numCells: number, mode: AppMode) {
   // Start the app as invisible and delay proportional to the number of cells,
   // to avoid most of the flickering when the app is loaded (b/c it is
-  // streamed). Delaying also helps prevent cell editors from stealing focus.
+  // streamed).
   const [invisible, setInvisible] = useState(true);
   useOnMount(() => {
+    // Only do this on read mode
+    if (mode !== "read") {
+      return;
```
Light2Dark, line 19: aside from the type error, i think this should `setInvisible(false)` before return.
read mode currently doesn't display anything

mscolnick: yes, thank you

manzt (approved):

PR #5360, opened 2025-06-19
improvement: Ctrl-A in console output selects all the output

mscolnick:
`Ctrl-A` in console output selects all the output. This is nice for copying out out marimo

* also removed eslint for a11y since its covered by biome

### DIFF
--- a/frontend/.eslintrc.cjs
+++ b/frontend/.eslintrc.cjs
@@ -13,8 +13,6 @@ module.exports = {
     "plugin:@typescript-eslint/recommended-type-checked",
     "plugin:@typescript-eslint/stylistic-type-checked",
     "plugin:@typescript-eslint/strict-type-checked",
-    // Accessibility
-    "plugin:jsx-a11y/strict",
     // React
     "plugin:react-hooks/recommended-legacy",
     "plugin:react/recommended",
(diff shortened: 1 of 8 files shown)


frontend/.eslintrc.cjs
```diff
@@ -13,8 +13,6 @@ module.exports = {
     "plugin:@typescript-eslint/recommended-type-checked",
     "plugin:@typescript-eslint/stylistic-type-checked",
     "plugin:@typescript-eslint/strict-type-checked",
-    // Accessibility
-    "plugin:jsx-a11y/strict",
```
mscolnick, line 17: covered by biome

frontend/src/hooks/useSelectAllContent.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Hook that enables Ctrl/Cmd-A to select all text content within a specific element
+ * @param ref - React ref to the element whose content should be selectable
+ * @param enabled - Whether the hook should be active (defaults to true)
+ */
+export function useSelectAllContent<T extends HTMLElement>(
+  ref: RefObject<T | null>,
+  enabled = true,
+) {
+  const handleKeyDown = (event: KeyboardEvent) => {
```
manzt, line 14: Would we want a useCallback for this? Not sure if `enabled` is dynamic.

manzt (approved):
Nice! This is the user behavior i'd expect. Also nice to prefer biome over eslint for some of this stuff.

frontend/src/hooks/useSelectAllContent.ts
mscolnick: we use react-compiler so don't need `useCallback` anymore. also the `useEventHandler` does handle keeping a stable ref

PR #5375, opened 2025-06-20
style: runtime connecting indicator

akshayka:
A more subtle, but still quite visible, connecting alert. The previous connecting alert clipped over the first cell / output, and distracted from looking at the notebook.

(cc @dmadisetti) 

https://github.com/user-attachments/assets/47cf0a36-5637-4cff-958e-04dadfb136ad


manzt (approved):
Nice, I really like this! The use of `size` + `gap` (as a specific tailwind class) feels a bit exposed, but I can't think of a cleaner way around it. That said, the JSDoc comments make the API clear.

PR #5377, opened 2025-06-20
Fix matplotlib plots not rendering in HTML export

### LINKED ISSUE #5370: Export html broken?
### Describe the bug

I've been using `marimo export html file.py -o file.html` for the last couple of days with no issues, generating html files that have the same cell order and embed and render images and matplotlib outputs.

For example see - https://seanmcleod70.github.io/FlightDynamicsCalcs/

Now, I'm assuming after upgrading to marimo 0.14.0, the html output is FUBAR.

The order of the cells in the html file are not in the same order as they appear in the marimo notebook and all the matplotlib outputs are now rendered as base64 strings!

```
data:image/png;base64,iVBORw0KG.....
```

Embedded images still render fine, looks like just the matplotlib outputs aren't rendering.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.0",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 158 Stepping 12, GenuineIntel",
  "Python Version": "3.9.21",
  "Binaries": {
    "Browser": "136.0.7103.114",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.17.1",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.25.0",
    "packaging": "24.2",
    "psutil": "5.9.0",
    "pygments": "2.15.1",
    "pyyaml": "6.0.2",
    "starlette": "0.45.3",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "14.2"
  },
  "Optional Dependencies": {
    "altair": "5.0.1",
    "nbformat": "5.3.0",
    "openai": "1.46.0",
    "pandas": "1.4.2",
    "pyarrow": "19.0.0",
    "pytest": "7.1.1",
    "python-lsp-server": "1.12.0",
    "ruff": "0.9.4"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

manzt:
The issues experienced in #5370 seem to be related to two separate things:

1. Cell outputs appear to be in topological order instead of document order
2. Rich media console outputs aren't rendered in the HTML export

The latter seems to be related to our recent schema/serialization work where matplotlib plots get stuck in console outputs instead of being promoted to main outputs for frontend rendering.

I think I'm close to a fix but I haven't been able to run a local build to verify, so marking this as draft.

Fixes #5370

### DIFF
--- a/marimo/_server/session/serialize.py
+++ b/marimo/_server/session/serialize.py
@@ -201,7 +201,7 @@ def serialize_notebook(
     """Convert a SessionView to a Notebook schema."""
     cells: list[NotebookCell] = []
 
-    for cell_id in view.cell_operations.keys():
+    for cell_id in cell_manager.cell_ids():
         # Get the code from last_executed_code, fallback to empty string
         code = view.last_executed_code.get(cell_id, "")
         cell_data = cell_manager.get_cell_data(cell_id)
--- a/marimo/_server/session/serialize.py
+++ b/marimo/_server/session/serialize.py
@@ -27,11 +27,13 @@ from marimo._schemas.notebook import (
 from marimo._schemas.session import (
     VERSION,
     Cell,
+    ConsoleType,
     DataOutput,
     ErrorOutput,
     NotebookSessionMetadata,
     NotebookSessionV1,
     OutputType,
+    StreamMediaOutput,
     StreamOutput,
 )
 from marimo._server.session.session_view import SessionView
@@ -67,7 +69,7 @@ def serialize_session_view(view: SessionView) -> NotebookSessionV1:
 
     for cell_id, cell_op in view.cell_operations.items():
         outputs: list[OutputType] = []
-        console: list[StreamOutput] = []
+        console: list[ConsoleType] = []
 
         # Convert output
         if cell_op.output:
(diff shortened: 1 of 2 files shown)


marimo/_server/session/serialize.py
```diff
@@ -201,7 +222,9 @@ def serialize_notebook(
     """Convert a SessionView to a Notebook schema."""
     cells: list[NotebookCell] = []
 
-    for cell_id in view.cell_operations.keys():
+    # Use document order from cell_manager instead of execution order from session_view
+    # to ensure cells appear in the correct sequence in HTML export
+    for cell_id in cell_manager.cell_ids():
```
manzt, line 227: Could this cause some reconciliation issue in the front end?

akshayka: This looks correct to me, since the notebook snapshot is only used used in the HTML export.

akshayka:
@manzt I generated the new schemas and API. I *think* this change of yours fixes the issue (`marimo development preview` shows something now, at least. No outputs in the preview but maybe that's a bug in the development command. On main I just see a "Connecting" spinner.).

But the following test is failing:

https://github.com/marimo-team/marimo/blob/75f8d9ec332fb426b41d59050d63b9b30d2bea09/tests/_server/session/test_serialize_session.py#L285-L314

I don't understand the logic in that test — if a cell were in the session view but not in the cell manager, then I don't think it makes sense to serialize it in the notebook (we have no idea where in the notebook it goes). But perhaps I am mistaken.

@dmadisetti can you comment on whether that test actually makes sense?

trevor:
> @trevor I generated the new schemas and API.

Nice effort! Great job so far!

akshayka:
> > @trevor I generated the new schemas and API.
> 
> 
> 
> Nice effort! Great job so far!

Oops, sorry for the wrong tag. Appreciate the humor :)

akshayka (approved):
I think this is correct, or at least "more correct" than what's on main (ie hopefully restoring HTML exports). I'm going to merge so we can see if it fixes the issue (@dmadisetti to review post merge).

PR #5381, opened 2025-06-20
fix: query parameters in ws

### LINKED ISSUE #5376: Marimo 0.14 breaks query params (at least locally)
### Describe the bug

import marimo as mo

query_params = mo.query_params()

query_params

The above 3 lines of code no longer show query params when you set them locally in 0.14 and it used to work in 0.13.15 and all versions before

### Will you submit a PR?

- [ ] Yes

### Environment

0.14

### Code to reproduce

import marimo as mo

query_params = mo.query_params()

query_params

akshayka:
This fixes a regression that caused `mo.query_params()` to stop working. We need to send all query parameters to the backend, not just known ones, as part of the websocket URL.

Tested by playing with the query parameters smoke test (`_marimo/smoke_tests`).

Fixes #5376.

### DIFF
--- a/frontend/src/core/runtime/runtime.ts
+++ b/frontend/src/core/runtime/runtime.ts
@@ -52,14 +52,10 @@ export class RuntimeManager {
       }
     }
 
-    // Move over window level parameters to the WebSocket URL
-    // if they are "known" query params.
-    for (const lookup in KnownQueryParams) {
-      const key = KnownQueryParams[lookup as keyof typeof KnownQueryParams];
-      const value = currentParams.get(key);
-      if (value !== null) {
-        baseUrl.searchParams.set(key, value);
-      }
+    // Move over all window-level parameters to WebSocket URL.
+    // This is needed for user-defined query parameters.
+    for (const [key, value] of currentParams.entries()) {
+      baseUrl.searchParams.set(key, value);
     }
 
     const cleanPath = baseUrl.pathname.replace(/\/$/, "");
--- a/frontend/src/core/runtime/runtime.ts
+++ b/frontend/src/core/runtime/runtime.ts
@@ -38,7 +38,11 @@ export class RuntimeManager {
   /**
    * The base URL of the runtime.
    */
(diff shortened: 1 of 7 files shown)


frontend/src/core/runtime/runtime.ts
```diff
@@ -52,14 +52,10 @@ export class RuntimeManager {
       }
     }
 
-    // Move over window level parameters to the WebSocket URL
```
akshayka, line 55: @dmadisetti Was there a reason for stripping unknown query parameters? This breaks `mo.query_params()`.

Was this code too aggressively refactored, i.e. is there a codepath that relies on query parameter stripping?

frontend/src/core/runtime/runtime.ts
```diff
@@ -52,14 +56,14 @@ export class RuntimeManager {
...
-      const value = currentParams.get(key);
-      if (value !== null) {
-        baseUrl.searchParams.set(key, value);
+    for (const [key, value] of currentParams.entries()) {
+      if (
+        restrictToKnownQueryParams &&
+        !Object.values(KnownQueryParams).includes(key)
+      ) {
+        continue;
       }
```
akshayka, line 65: I included this in case other code paths do need to restrict to known query parameters.

frontend/src/core/runtime/runtime.ts
dmadisetti: A little too conservative - with the refactor query params were not passed down at all. Thanks for the general fix

frontend/src/core/runtime/runtime.ts
dmadisetti: Likely not needed

dmadisetti (approved):
Thanks for the unit test too

frontend/src/core/runtime/runtime.ts
akshayka: Got it, so just don't filter known query params, pass all through?

Edit: nvm, saw your approval.

PR #5390, opened 2025-06-21
fix: wasm column summary charts

Light2Dark:
## 📝 Summary

CleanShot 2025-06-21 at 16 33 19@2x

At some point, our data format for charts turned to arrow. While this worked for urls (@file, http:...), it didn't work for raw base64 strings which happened in wasm mode. This adds some handling for base64 + arrow format.

## 📋 Checklist

## 📜 Reviewers

@akshayka @manzt

### DIFF
--- a/frontend/src/components/data-table/chart-spec-model.tsx
+++ b/frontend/src/components/data-table/chart-spec-model.tsx
@@ -1,15 +1,24 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
 import { mint, orange, slate } from "@radix-ui/colors";
+// @ts-expect-error vega-typings does not include formats
+import { formats } from "vega";
 import type { TopLevelSpec } from "vega-lite";
 import { asRemoteURL } from "@/core/runtime/config";
 import type { TopLevelFacetedUnitSpec } from "@/plugins/impl/data-explorer/queries/types";
+import { arrow } from "@/plugins/impl/vega/formats";
 import { parseCsvData } from "@/plugins/impl/vega/loader";
 import { logNever } from "@/utils/assertNever";
 import type { ColumnHeaderStats, ColumnName, FieldTypes } from "./types";
 
 // We rely on vega's built-in binning to determine bar widths.
 const MAX_BAR_HEIGHT = 20; // px
+
+// Arrow formats have a magic number at the beginning of the file.
+const ARROW_MAGIC_NUMBER = "ARROW1";
+
+// register arrow reader under type 'arrow'
+formats("arrow", arrow);
 
 export class ColumnChartSpecModel<T> {
   private columnStats = new Map<ColumnName, ColumnHeaderStats>();
@@ -54,11 +63,24 @@ export class ColumnChartSpecModel<T> {
           values: parseCsvData(decoded) as T[],
         };
         this.sourceName = "data_0";
+      } else if (this.data.startsWith("data:text/plain;base64,")) {
+        const decoded = atob(this.data.split(",")[1]);
+        if (decoded.startsWith(ARROW_MAGIC_NUMBER)) {
+          const arrowBuffer = Uint8Array.from(decoded, (c) => c.charCodeAt(0));
+          this.dataSpec = {
+            values: arrowBuffer,
+            // @ts-expect-error vega-typings does not include arrow format
+            format: { type: "arrow" },
+          };
+          this.sourceName = "data_0";
+        } else {
+          // Assume it's a CSV string
+          this.parseCsv(decoded);
+          this.sourceName = "data_0";
+        }
       } else {
         // Assume it's a CSV string
-        this.dataSpec = {
-          values: parseCsvData(this.data) as T[],
-        };
+        this.parseCsv(this.data);
         this.sourceName = "data_0";
       }
     } else {
@@ -80,6 +102,12 @@ export class ColumnChartSpecModel<T> {
       stats: this.columnStats.get(column),
       type: this.fieldTypes[column],
       spec: this.opts.includeCharts ? this.getVegaSpec(column) : undefined,
+    };
+  }
+
+  private parseCsv(data: string) {
+    this.dataSpec = {
+      values: parseCsvData(data) as T[],
     };
   }
 
--- a/frontend/src/components/data-table/chart-spec-model.tsx
+++ b/frontend/src/components/data-table/chart-spec-model.tsx
@@ -9,6 +9,11 @@ import type { TopLevelFacetedUnitSpec } from "@/plugins/impl/data-explorer/queri
 import { arrow } from "@/plugins/impl/vega/formats";
 import { parseArrowData, parseCsvData } from "@/plugins/impl/vega/loader";
 import { logNever } from "@/utils/assertNever";
+import {
+  extractBase64FromDataURL,
+  isDataURLString,
+  typedAtob,
+} from "@/utils/json/base64";
 import type { ColumnHeaderStats, ColumnName, FieldTypes } from "./types";
 
 // We rely on vega's built-in binning to determine bar widths.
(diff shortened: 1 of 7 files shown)


frontend/src/components/data-table/chart-spec-model.tsx
```diff
@@ -54,11 +63,23 @@ export class ColumnChartSpecModel<T> {
           values: parseCsvData(decoded) as T[],
         };
         this.sourceName = "data_0";
+      } else if (this.data.startsWith("data:text/plain;base64,")) {
```
manzt: Arrow isn't a text format, so the mimetype here is a little surprising to me. Thinking it's outside the scope of this PR, but it would be good to find where we are generating these data urls on the Python side and register a more descriptive mimetype:

- [`application/vnd.apache.arrow.stream`](https://www.iana.org/assignments/media-types/application/vnd.apache.arrow.stream)
- `application/octet-stream` 

Using the explicit mimetype would also allow us to delineate between CSV vs. Arrow without having to sniff the contents. e.g., eventually something like:


```ts
if (isDataURLString(this.data)) {
  const {mimetype, data} = parseDataURL(this.data);
  if (mimetype === "application/vnd.apache.arrow.stream") {
    const decodedBytes = decodeBase64StringBinary(data);
    invariant(includesArrowMagic(decodedBytes), "Missing ARROW magic.");
    this.dataSpec = {
      values: decodedBytes,
      format: { type: "arrow" },
    };
  } else if (mimetype === "text/csv") {
    const decodedText = decodeBase64StringText(content);
    this.parseCsv(decodedText);
  } else {
    assertNever("Unknown mimetype");
  }
}
```

manzt (changes requested):
Nice! I love that we are switching more to arrow format. I had some smaller nits inline but at a high level I think it would be nice to move some of the data handling into more general utilities in `base64.ts`.

I think Myles added some nice branded types for base64 handling but we're duplicating the parsing logic here. Since we can't delineate between mimetypes right now (both CSV and Arrow come as `text/plain`), we still need to decode and check for the Arrow magic string. We could add/use some existing utilities:

```typescript
// Add to base64.ts:
export function isDataURLString(x: string): x is DataURLString {
  return x.startsWith("data:") && x.includes(";base64,");
}

export function extractBase64FromDataURL(dataUrl: DataURLString): Base64String {
  const [, base64Data] = dataUrl.split(",", 2);
  return base64Data as Base64String;
}

export function byteStringToBinary<T>(bytes: ByteString<T>): Uint8Array {
  return Uint8Array.from(bytes, (c) => c.charCodeAt(0));
}
```

Then use them in the chart model:

```typescript
import {
  isDataURLString,
  extractBase64FromDataURL,
  typedAtob,
  byteStringToBinary
} from "@/utils/json/base64";

if (isDataURLString(this.data)) {
  const base64 = extractBase64FromDataURL(this.data);
  const decoded = typedAtob(base64); // Returns ByteString

  if (decoded.startsWith(ARROW_MAGIC_NUMBER)) {
    this.dataSpec = {
      values: byteStringToBinary(decoded), // Convert to Uint8Array
      format: { type: "arrow" },
    };
  } else {
    // Handle as CSV
    this.parseCsv(decoded);
  }
}
```

Then we probably don't need a `parseArrowData` util.

frontend/src/components/data-table/chart-spec-model.tsx
Light2Dark: The code is from [here](https://github.com/marimo-team/marimo/blob/main/marimo/_plugins/ui/_impl/table.py#L754) but I haven't been able to pinpoint why it's returning base64 in wasm only. (I'm also facing an issue where the requests are being cached imo, and I can't update the backend code...)

Light2Dark:
Thanks for the suggestions! I think it's cleaner now, I maintained the `parseArrowData` function so I could add some tests to verify it works with edge cases (like with csv). But I'm okay to remove it too.

frontend/src/plugins/impl/vega/loader.ts
```diff
@@ -233,6 +240,14 @@ export async function vegaLoadData<T = object>(
   }
 }
 
+export function parseArrowData(data: string): Uint8Array {
+  const decoded = isDataURLString(data)
+    ? typedAtob(extractBase64FromDataURL(data))
+    : (data as ByteString);
+
+  return byteStringToBinary(decoded);
+}
```
manzt: Totally makes sense to keep this function focused on Arrow-specific handling, but I think we might want to tighten the input type to avoid accepting arbitrary strings.

In practice, it looks like we’re already extracting the byte string before calling this, so the conditional/branching may not be necessary. If we don’t expect raw Data URLs here, we could simplify the implementation and lean on TypeScript to enforce the invariant:

```ts
export function parseArrowData(data: ByteString): Uint8Array {
  invariant(data.startsWith("ARROW"), "Invalid Arrow data.");
  return byteStringToBinary(decoded);
}
```

By narrowing the input to a branded `ByteString` (instead of arbitrary string), we push callers toward the correct flow with TypeScript. 

For example, the only “safe” path to a `ByteString` would be something like:

```ts
if (isDataURLString(data)) {
  // data is now DataURLString
  const encoded: Base64String  = extractBase64FromDataURL(data);
  const decoded: ByteString = typedAtob(encoded);
  const arrowBuffer = parseArrowData(decoded);
}
```

Notice how there are no type-assertions (e.g., `as ByteString`). Using our utilities, the only "safe" way to get a `Base64String` type is to have a `DataURLString`, and the only way to get a `ByteString` is to have a `Base64String`.

frontend/src/utils/json/base64.ts
```diff
@@ -37,3 +37,15 @@ export function typedAtob<T>(base64: Base64String<T>): ByteString<T> {
 export function typedBtoa<T>(bytes: ByteString<T>): Base64String<T> {
   return window.btoa(bytes) as Base64String<T>;
 }
+
+export function isDataURLString(str: string): boolean {
+  return str.startsWith("data:");
+}
```
manzt: We can make a custom [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) here:

```suggestion
export function isDataURLString(str: string): str is DataURLString {
  return str.startsWith("data:") && str.includes(";base64,")
}
```

Type guards are a way to “hook” into the type system by telling TypeScript: “if this condition passes, you can safely treat the value as a narrower type." Our `DataURLString` type looks like this:

```ts
export type DataURLString = `data:${string};base64,${Base64String<string>}`;
```

So we want to check both the properties about the arbitrary string. Before, if this just returned `boolean`, we still have to cast or assert manually:

```ts
if (isDataURLString(str)) {
  // str is still of type "string"
  extractBase64FromDataURL(str as DataURLString); // manual cast
}
```

Now, with the `str is DataURLString` return type, TypeScript understands the refinement automatically:

```ts
if (isDataURLString(str)) {
  // str is now of type "DataURLString"
  extractBase64FromDataURL(str); // no need to cast
}
```

frontend/src/utils/json/base64.ts
```diff
@@ -37,3 +37,15 @@ export function typedAtob<T>(base64: Base64String<T>): ByteString<T> {
...
+
+export function extractBase64FromDataURL(str: string): Base64String {
+  return str.split(",")[1] as Base64String;
+}
```
manzt: Since this function assumes a specific structure (`data:*;base64,...`), we can put more requirements on the input type to be a `DataURLString`. That way, we make the invariant explicit and avoid needing an as cast inside the function:

```suggestion
export function extractBase64FromDataURL(str: DataURLString): Base64String {
  return str.split(",")[1] as Base64String;
}
```

This pairs nicely with the `isDataURLString` type guard, since callers can refine from string to `DataURLString`, and this function can then safely extract the payload.

frontend/src/plugins/impl/vega/__tests__/loader.test.ts
```diff
@@ -71,3 +71,75 @@ describe("DATE_MIDDLEWARE", () => {
     expect(data[1].Date).toBe("2024-02-30");
   });
 });
+
+describe("Arrow data", () => {
```
manzt: I might be misunderstanding, but my understanding is that Arrow is a strict binary format—so just prepending "ARROW" to some CSV content isn’t really test/generating Arrow data. The loader tests here seem to be more about going between strings and binary data, rather than testing anything specific to Arrow itself. If that’s the case, we might be able to simplify the tests or clarify the intent a bit.

```typescript
describe("parseArrowData", () => {
  it("should convert ByteString to Uint8Array", () => {
    const byteString = "ARROW1\nsome data" as ByteString;
    const result = parseArrowData(byteString);

    expect(result).toBeInstanceOf(Uint8Array);
    expect(String.fromCharCode(...result)).toBe("ARROW1\nsome data");
  });

  it("should throw if missing ARROW magic number", () => {
    const invalidByteString = "invalid data" as ByteString;
    expect(() => parseArrowData(invalidByteString)).toThrow("Invalid Arrow format");
  });

  it("should handle empty ByteString", () => {
    const emptyByteString = "" as ByteString;
    expect(() => parseArrowData(emptyByteString)).toThrow();
  });
});
```

manzt (commented):
Looking great overall! I left a few inline comments, mostly around using our branded types a bit more effectively. The main idea is to avoid relying on [type assertions](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions), which can act as easy escape hatches and are [generally best avoided](https://trevorma.nz/blog/be-assertive#type-assertions).

Rather than asserting these types directly (which sidesteps type safety), we encourage using small, testable utilities (like type guards or parsing functions) that safely produce or refine these types. By doing this, we let TypeScript do more of the work for us, and reduce the risk of runtime errors caused by incorrect assumptions.

manzt (dismissed):
Looks great!

PR #5392, opened 2025-06-21
feat: add chart builder / chart-to-cell functionality within table

Light2Dark:
## 📝 Summary

https://github.com/user-attachments/assets/c7c35b46-9424-4184-935f-f7ee86ed3761

CleanShot 2025-07-03 at 01 49 56

- Add chart to notebook
- code gen fixes
- rename placeholder variable with `replace_with_df`, I'm not sure which is better `df` or `replace_with_df`, but I did see Vincent get confused with just df. This may help.
- collapsible sidebar
- better tooltips for temporal units

## 📋 Checklist


Light2Dark:
For context @manzt , this feature is under a feat. flag, you're able to build charts right from tables using Vega lite. The vega spec generation happens entirely on the frontend. Code may be hard to parse since it's mainly vega spec generation.

Maybe review on the `readonly_python_code.tsx` would help since that is a shared component.

frontend/src/components/data-table/charts/charts.tsx
```diff
@@ -243,6 +246,12 @@ export const ChartPanel: React.FC<{
 
   const [selectedChartType, setSelectedChartType] =
     useState<ChartType>(chartType);
+  const [formCollapsed, setFormCollapsed] = useState(false);
+
+  const chartContainerRef = useRef<HTMLDivElement>(null);
+  useResizeObserver({
+    ref: chartContainerRef as React.RefObject<Element>,
+  });
```
manzt: Would this work to avoid the type assertion? Looks to be an option in [the docs](https://www.npmjs.com/package/use-resize-observer).

```suggestion
  const { ref: chartContainerRef } = useResizeObserver<HTMLDivElement>();
```

frontend/src/components/editor/code/readonly-python-code.tsx
```diff
@@ -22,12 +23,25 @@ const pythonExtensions = [
...
+/**
+ * A readonly code component that can be used to display code in a readonly state.
+ *
+ * @param props.className - The class name to apply to the component.
+ * @param props.code - The code to display.
+ * @param props.initiallyHideCode - Whether to initially hide the code.
+ * @param props.showHideCode - Whether to show the hide code button.
+ * @param props.insertNewCell - Whether to add a insert new cell button; when clicked will add a new cell next to the current cell or at the end of the file
+ * @param props.language - The language of the code. Default is "python".
+ */
```
manzt, line 35: Nice docs!

manzt (approved):
Looks great!

frontend/src/components/data-table/charts/charts.tsx
Light2Dark: great suggestion, thanks!

frontend/src/components/data-table/charts/chart-spec/spec.ts
```diff
@@ -240,14 +271,19 @@ function getBaseSpec(
   width: number | "container",
   height: number,
   title?: string,
-) {
+): BaseSpec {
   return {
     $schema: "https://vega.github.io/schema/vega-lite/v5.json",
-    background: theme === "dark" ? "dark" : "white",
+    background: theme === "dark" ? "dark" : slate.slate2,
```
manzt: If this is allowed to be a css value, I'm wondering if we could use a css variable (e.g., `"var(--background)"`), or do we want a distinct color?

frontend/src/components/data-table/charts/charts.tsx
```diff
@@ -303,37 +308,44 @@ export const ChartPanel: React.FC<{
     if (typeof specWithoutData !== "string") {
       altairCodeSnippet = generateAltairChartSnippet(
         specWithoutData,
-        "df",
+        "replace_this_df",
```
manzt: I get that `df` might be unclear, but aesthetically (IMO) `replace_this_df` doesn't feel great either. 

Ideally, we'd know the data source rather than asking them to inline their own. In the meantime, maybe we could stick with `df` and add an inline comment to guide them? e.g.,

```python
_chart = (
    alt.Chart(df)  # ← replace `df` with your data source
    .mark_bar()
    .encode(...)
)
_chart
```

frontend/src/components/data-table/charts/forms/common-chart.tsx
```diff
@@ -137,7 +157,10 @@ export const OtherOptions: React.FC = () => {
     <Accordion type="multiple">
       <AccordionFormItem value="facet">
         <AccordionFormTrigger className="pt-0">
-          <Title text="Faceting" />
+          <Title
+            text="Faceting"
+            tooltip="Faceting creates multiple small charts by grouping data on specified fields"
```
manzt: If the text already says "Faceting" I think this could be more concise:

```suggestion
            tooltip="Repeat the chart for each unique field value."
```

manzt (dismissed):
Just some minor nits, thanks for your patience in my review!

akshayka (dismissed):

manzt:
@Light2Dark I updated the name of this PR to try to be more descriptive to end users in release notes. Feel free to update if you think it's not accurate or could be improved!

frontend/src/components/data-table/charts/chart-spec/spec.ts
Light2Dark: unfortunately, it doesn't work. I think it may not accept hsl values.

akshayka: Ah interesting. Why `slate2`? I think `white` looks good and is the default for light?

Light2Dark: Yeah, I think I switched to have some visual separation between the left and right side. But it does look like white is fine. Reverted

akshayka (dismissed):
Feel free to merge when CI passing

akshayka (approved):

PR #5397, opened 2025-06-21
fix: CLI print for non UTF-8 terminals

akshayka:
Replace common unicode characters with ASCII on UnicodeEncodeError encountered while printing to console.

Should fix failing CLI test on CI.

### DIFF
--- a/marimo/_cli/print.py
+++ b/marimo/_cli/print.py
@@ -76,13 +76,32 @@ def muted(text: str) -> str:
     return "\033[37;2m" + text + "\033[0m" if _USE_COLOR else text
 
 
-def echo(*args: Any, **kwargs: Any) -> None:
-    if GLOBAL_SETTINGS.QUIET:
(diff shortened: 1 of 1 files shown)


marimo/_cli/print.py
```diff
@@ -76,13 +76,32 @@ def muted(text: str) -> str:
...
+        # terminals.
+        ascii_args = []
+        for arg in args:
+            if isinstance(arg, str):
+                ascii_arg = arg.replace("→", "->").replace("←", "<-")
+                ascii_arg = ascii_arg.replace("✓", "v").replace("✗", "x")
+                ascii_arg = ascii_arg.replace("•", "*").replace("…", "...")
+                ascii_args.append(ascii_arg)
+            else:
+                ascii_args.append(arg)
```
dmadisetti, line 106: Are those the only possible characters? Seems like there could be others

PR #5398, opened 2025-06-21
fix: Serialization fix

dmadisetti:
## 📝 Summary

Looks like there are 2 issues in #3226 

 - [x] 1. Serialization is not in correct order
 - [] 2. The first cell is dropped.

Unsure why 2 occurs, but might be fixed with 1. Have not tested

Feel free to hack on this in the meantime. Will be back online at ~2

fixes #3226

### DIFF
--- a/marimo/_server/session/serialize.py
+++ b/marimo/_server/session/serialize.py
@@ -67,15 +67,22 @@ def serialize_session_view(view: SessionView) -> NotebookSessionV1:
     """Convert a SessionView to a NotebookSession schema."""
     cells: list[Cell] = []
 
-    # Attempt Top-Bottom sorting
-    cell_ids = view.cell_ids if view.cell_ids else view.cell_operations.keys()
+    if view.cell_ids is not None:
+        cell_ids = view.cell_ids.cell_ids
+    else:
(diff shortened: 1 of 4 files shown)


marimo/_server/session/serialize.py
```diff
@@ -67,7 +70,25 @@ def serialize_session_view(view: SessionView) -> NotebookSessionV1:
     """Convert a SessionView to a NotebookSession schema."""
     cells: list[Cell] = []
 
-    for cell_id, cell_op in view.cell_operations.items():
+    cell_ids: Iterable[CellId_t]
+    if view.cell_ids is not None:
+        cell_ids = view.cell_ids.cell_ids
```
akshayka: @dmadisetti when running `marimo export html`, `view.cell_ids` is `None`, so I think this won't quite solve the problem.

I'm not sure I understand the reconciliation logic. Not sure why the frontend can't match on Cell IDs between the session view and the HTML.

akshayka: Updated `run_app_until_completion` elsewhere to populate `view.cell_ids`. But this does feel brittle to me.

PR #5403, opened 2025-06-21
fix: marimo development preview

akshayka:
This PR fixes `marimo development preview` for previewing static HTML using local assets. Previously the generated HTML didn't include outputs (ie the notebook wasn't run), and also got stuck on attempting to connect to the frontend ("Connecting..." spinner)

* actually generate the notebook session and outputs
* for static notebooks, initialize the connection as open


frontend/src/core/network/connection.ts
```diff
@@ -1,14 +1,15 @@
...
 import { type ConnectionStatus, WebSocketState } from "../websocket/types";
 
 /**
  * Atom for storing the connection status.
- * Initialized to CONNECTING.
+ * Initialized to CONNECTING for normal mode, OPEN for static mode.
  */
 export const connectionAtom = atom<ConnectionStatus>({
-  state: WebSocketState.CONNECTING,
+  state: isStaticNotebook() ? WebSocketState.OPEN : WebSocketState.CONNECTING,
```
akshayka: This feels suspicious because it only appears to be needed when serving HTML with local assets; might not be the right fix.

dmadisetti: Will test locally, but looks fine

manzt (approved):
This looks great! Wondering if instead of checking the type of notebook globally in the connection atom, we can try to "fix" the static case at the websocket implementation level?

```typescript
// In frontend/src/core/websocket/StaticWebsocket.ts
addEventListener(type: string, listener: EventListener): void {
  if (type === "open") {
    this.onopen = listener;
    // Trigger open event immediately since static websockets are always "open"
    queueMicrotask(() => this.onopen?.(new Event("open")));
  }
}
```

This way the `StaticWebsocket` properly simulates a WebSocket that immediately opens, and we avoid leaking static mode awareness into the global connection state.

akshayka:
Yea that's much better. I'll make the change before merging. Thank you!

manzt:
Hmm... just tried my change locally and still see the spinner. I can poke around a bit, but guessing the `queueMicroTask` probably introduces a race condition.

akshayka:
@manzt let me know if you want changes made, or if okay to merge as is

manzt:
Still getting used to procedure with taking over other's PRs... I was able to check this out locally and get the following to work with:

```sh
wget https://gist.githubusercontent.com/manzt/f2a9299d913dc9801c639134db6e06e8/raw/f9ca2d565a3df3597a0eb85fbec230eac71b17f7/mo-3226.py
make fe
uv run --with matplotlib,numpy marimo development preview mo-3226.py
``` 

```patch
diff --git i/frontend/src/core/network/connection.ts w/frontend/src/core/network/connection.ts
index daa8d339..67f58b5b 100644
--- i/frontend/src/core/network/connection.ts
+++ w/frontend/src/core/network/connection.ts
@@ -1,7 +1,6 @@
 import { atom } from "jotai";
 import { waitFor } from "../state/jotai";
-import { isStaticNotebook } from "../static/static-state";
 import { type ConnectionStatus, WebSocketState } from "../websocket/types";
 
 /**
@@ -9,7 +8,7 @@ import { type ConnectionStatus, WebSocketState } from "../websocket/types";
  * Initialized to CONNECTING for normal mode, OPEN for static mode.
  */
 export const connectionAtom = atom<ConnectionStatus>({
-  state: isStaticNotebook() ? WebSocketState.OPEN : WebSocketState.CONNECTING,
+  state: WebSocketState.CONNECTING,
 });
 
 export function waitForConnectionOpen() {
diff --git i/frontend/src/core/websocket/StaticWebsocket.ts w/frontend/src/core/websocket/StaticWebsocket.ts
index d4120366..981a23e2 100644
--- i/frontend/src/core/websocket/StaticWebsocket.ts
+++ w/frontend/src/core/websocket/StaticWebsocket.ts
@@ -17,9 +17,20 @@ export class StaticWebsocket implements IReconnectingWebSocket {
   onmessage = null;
   onopen = null;
 
-  addEventListener(type: unknown, callback: unknown, options?: unknown): void {
-    // Noop
+  addEventListener(
+    type: string,
+    callback: EventListener,
+    _options?: unknown,
+  ): void {
+    // Normally this would be a no-op in a mock, but we simulate a synthetic "open" event
+    // to mimic the WebSocket transitioning from CONNECTING to OPEN state.
+    if (type === "open") {
+      queueMicrotask(() => {
+        callback(new Event("open"));
+      });
+    }
   }
+
   removeEventListener(
     type: unknown,
     callback: unknown,
```

I'm happy to apply the diff directly if that is preferred!

akshayka:
@manzt thanks! just applied it

PR #5413, opened 2025-06-22
Simplify and tighten TypeScript configuration

manzt:
This PR refactors our `tsconfig.json` to use a simplified setup based on Vite's default config (`npx create vite`). Many previously listed options are already covered by `strict` setting, so I've removed the redundant ones for clarity. JavaScript support has also been explicitly disabled to enforce a stricter TypeScript-only codebase.


frontend/src/components/editor/RecoveryButton.tsx
```diff
@@ -1,10 +1,10 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
 import type { Notebook } from "@marimo-team/marimo-api";
-import { Button } from "components/ui/button";
```
manzt, line 4: These kind of imports were technically allowed because `baseUrl` was set to `"src"`, but the new config requires any "root" imports are imported with `@/*` prefix. Better to be consistent.

frontend/src/components/slides/slides-component.tsx
```diff
@@ -14,12 +14,6 @@ import { useEventListener } from "@/hooks/useEventListener";
 import { cn } from "@/utils/cn";
 
 import "./slides.css";
-import "swiper/css";
-import "swiper/css/virtual";
-import "swiper/css/keyboard";
-import "swiper/css/navigation";
-import "swiper/css/pagination";
-import "swiper/css/scrollbar";
```
manzt, line 22: These CSS imports quietly worked before because we had `allowJs: true`, which let TypeScript accept non-TypeScript imports in TS files. 

Moving them to the CSS file makes this behavior explicit and allows us to safely disable `allowJs` across the reset of the code base.

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    /* Interop Constraints */
-    // "isolatedModules": true,                          /* Ensure that each file can be safely transpiled without relying on other imports. */
-    // "allowSyntheticDefaultImports": true,             /* Allow 'import x from y' when a module doesn't have a default export. */
-    "esModuleInterop": true /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables 'allowSyntheticDefaultImports' for type compatibility. */,
-    // "preserveSymlinks": true,                         /* Disable resolving symlinks to their realpath. This correlates to the same flag in node. */
-    "forceConsistentCasingInFileNames": true /* Ensure that casing is correct in imports. */,
-
-    /* Type Checking */
-    "strict": true /* Enable all strict type-checking options. */,
-    "noImplicitAny": true /* Enable error reporting for expressions and declarations with an implied 'any' type. */,
```
manzt, line 86: Default `true` when `strict: true`

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    "strictNullChecks": true /* When type checking, take into account 'null' and 'undefined'. */,
```
manzt, line 87: Default `true` when `strict: true`

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    "strictFunctionTypes": true /* When assigning functions, check to ensure parameters and the return values are subtype-compatible. */,
```
manzt, line 88: Default `true` when `strict: true`

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    "strictBindCallApply": true /* Check that the arguments for 'bind', 'call', and 'apply' methods match the original function. */,
```
manzt, line 89: Default `true` when `strict: true`

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    "strictPropertyInitialization": true /* Check for class properties that are declared but not set in the constructor. */,
```
manzt, line 90: Default `true` when `strict: true`

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    "noImplicitThis": true /* Enable error reporting when 'this' is given the type 'any'. */,
```
manzt, line 91: Default `true` when `strict: true`

frontend/tsconfig.json
```diff
@@ -1,113 +1,40 @@
...
-    "useUnknownInCatchVariables": true /* Default catch clause variables as 'unknown' instead of 'any'. */,
```
manzt, line 92: Default `true` when `strict: true`

akshayka (commented):
Noticed FE tests are failing?

manzt:
> Noticed FE tests are failing?

I believe some of the more strict tsconfig options (e.g., verbatimModuleSyntax, removing `allowJs`) may have caused some issues with disallowing some circular imports. I moved around a couple of imports to fix the error. With a bundle, there is no difference.

With these changes, I'm also able to typecheck the codebase with the new typescript compiler (written in Go), which deprecated some legacy TS options, with only a few issues. Should make the migration easy when we are ready to make it.

frontend/src/core/datasets/engines.ts
```diff
@@ -0,0 +1,10 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+import type { TypedString } from "@/utils/typed";
+
+export type ConnectionName = TypedString<"ConnectionName">;
+
+// DuckDB engine is treated as the default engine
+// As it doesn't require passing an engine variable to the backend
+// Keep this in sync with the backend name
+export const DUCKDB_ENGINE = "__marimo_duckdb" as ConnectionName;
+export const INTERNAL_SQL_ENGINES = new Set([DUCKDB_ENGINE]);
```
manzt, line 10: Separating this module without imports seemed to do the trick. It likely only matters for vitest, where there is no bundling but native esm used.

akshayka (approved):

PR #5414, opened 2025-06-22
chore(frontend): Add `invariant` debugger statement for development

manzt:
I do this often during development and find it really useful for debugging invariant failures. Thought I'd share this either to show other folks or check it into the repo so other developers can use it too.

The conditional `debugger` statement to the `invariant()` utility only runs in development. When an invariant fails, it automatically triggers a breakpoint in the browser before throwing the error. This immediately allows you to step into where some assertion failed.

The nice property is that the entire `if (import.meta.env.DEV)` block gets stripped out in production builds (e.g., `vite build`) since the bundler resolves to `if (false) {}`.


akshayka (approved):
This looks great to me, thanks!

PR #5415, opened 2025-06-22
feat(frontend): Inline cell dependency highlighting

manzt:
One of my favorite details in ObservableHQ notebooks is the subtle visual cue for dependencies between cells (bold/underline).

This PR introduces an experiment to bring a similar experience to marimo. When editing a cell, any references to variables defined in other cells are highlighted as dependencies. These highlights update as the user edits or as the global variable state changes.


manzt:
Just an idea and I know something we haven't discussed. I'd also be happy to put behind some user setting to opt-in...

akshayka (commented):
I can see this information being helpful in a few contexts:
* It tells the user: this cell has the potential to run reactively.
* When writing functions, it tells you if you've closed over anything (which can happen accidentally, and which will prevent making the function a "reusable" function).
* It hints to the user that they can click on the variable to go to the defining cell (but more on this in a bit).

I can also see it being a bit confusing to newcomers without additional context (why is this variable emphasized?).

While a cool feature, I'm not quite convinced that the effort required to land this is worth it right now. Some comments. 

* The implementation would need to handle lexical scoping. See here for an example that the implementation does not yet cover. `amplitude` and `period` should not be emphasized because they shadow a global.

image

* Ideally there would be a tooltip that tells the user that they can jump to the defining cell with Cmd/Ctrl+Click; this would be enough additional context for me, to make it clear to the user why this variable is emphasized ("oh, it's a link to another cell" — they might not think in terms of refs and defs, but "this was defined elsewhere" will make sense). But I don't know how to do that because our tooltip shows the docstring / typing information. Maybe the tooltip just goes on the underline, but that seems kind of clunky (you might accidentally ping/pong across the two different tooltips). Observable shows a tooltip but doesn't show docs.

image

* The double underline when hovering with `Cmd/Ctrl` held is awkward 
image

* Font weight 700 looks too heavy in light mode (somehow it's fine in dark). In light normal / 400 font-weight looks just fine to me.

* Emphasizing modules (e.g., `mo`) reads noisily to me. We have logic that prevents automatic re-running when incrementally adding imports to a cell, so I think it's not really needed. I'd like to filter modules out (this information is available in the frontend). 

* Variables defined in the setup cell (if any) likely shouldn't be emphasized. (I would find this feature useful for telling me at a glance why my function was not pure, but setup-cell references are okay.).

* Should the emphasis only be shown while a cell is focused, instead of always?

manzt:
> - [x] The implementation would need to handle lexical scoping. See here for an example that the implementation does not yet cover. amplitude and period should not be emphasized because they shadow a global.

Good call out, this is the most complicated part of the implementation. I wrote a fairly comprehensive test suite but I bet the algorithm could take another pass. Doing this, I actually found a couple of bugs in our go to definition heuristics and improved the goto definition lookup for "dependency" variables.

> - [x] Ideally there would be a tooltip that tells the user that they can jump to the defining cell with Cmd/Ctrl+Click; this would be enough additional context for me, to make it clear to the user why this variable is emphasized ("oh, it's a link to another cell" — they might not think in terms of refs and defs, but "this was defined elsewhere" will make sense). But I don't know how to do that because our tooltip shows the docstring / typing information. Maybe the tooltip just goes on the underline, but that seems kind of clunky (you might accidentally ping/pong across the two different tooltips). Observable shows a tooltip but doesn't show docs.

Agreed. I have added this to the hover documentation.

image

> - [x] The double underline when hovering with Cmd/Ctrl held is awkward

There is no longer a double underlines and there is a distinct hover state for each.

> - [x] Font weight 700 looks too heavy in light mode (somehow it's fine in dark). In light normal / 400 font-weight looks just fine to me.

Changed to weight 500 (I couldn't distinguish 400)

> - [x] Emphasizing modules (e.g., mo) reads noisily to me. We have logic that prevents automatic re-running when incrementally adding imports to a cell, so I think it's not really needed. I'd like to filter modules out (this information is available in the frontend).

Modules are now excluded from highlighting.

> - [x] Variables defined in the setup cell (if any) likely shouldn't be emphasized. (I would find this feature useful for telling me at a glance why my function was not pure, but setup-cell references are okay.).

Setup cells are now excluded from highlighting.


> - [ ] Should the emphasis only be shown while a cell is focused, instead of always?

Not sure about this one. I like it always, but might be a personal preference.

manzt:
Here is the notebook I have been tinkering with (copied cases from the test suite):

https://gist.github.com/manzt/3ebdc6b102c85859a1bb7d6f282368fa

akshayka (commented):
Nice fixes! Pretty comprehensive.

* Can we do font-weight 400 (ie normal?), at least in light mode? I find 500 (which ends up being the same as 700, I don't think we have the intermediate weights) to be quite distracting.
* Let's make this a user configuration, either `Editor` or `Display`. I can see people being pretty split on this. Maybe we default off, announce, and get feedback from users in Discord; if people really like it we can consider default on. 

After these two things, LGTM.

manzt:
> Can we do font-weight 400 (ie normal?), at least in light mode? I find 500 (which ends up being the same as 700, I don't think we have the intermediate weights) to be quite distracting.

I don't really feel strongly, although my preference is bolder. But something I could change with my own custom CSS :) Switched to 400 for now!

> Let's make this a user configuration, either Editor or Display. I can see people being pretty split on this. Maybe we default off, announce, and get feedback from users in Discord; if people really like it we can consider default on.

I made it a Display configuration, with additional actions to enable / disable. The feature **disabled by default**. I also named the feature "Reference highlighting":

image

There is also a quick action now in the command to enable / disable:

image

manzt:
Hmm seems like I'm running into something similar to #5422. I've been rebasing and re-generating the schema.

akshayka (approved):
Nice work — not an easy change, and it feels nice. Really like the improved go-to-definition (and its discoverability).

The schema test failure seems unrelated so I think this is good to merge.

PR #5423, opened 2025-06-23
logging: downgrade warning to debug for session cache serialization

dmadisetti:
Demote log warning when serializing the session view to a log info.
* The session cache writer does not seem to know the notebook-order of cells. This doesn't seem to be a problem in practice (can run a notebook and hydrate from a cached session view object just fine). But it's odd that not having the notebook order does pose a problem when hydrating an exported HTML object.

@akshayka

### DIFF
--- a/marimo/_server/session/serialize.py
+++ b/marimo/_server/session/serialize.py
@@ -82,7 +82,8 @@ def serialize_session_view(
         if view.cell_ids is not None:
             cell_ids = view.cell_ids.cell_ids
         else:
-            LOGGER.warning(
+            # TODO: Reconcile HTML, and json exports. Something seems a bit off.
+            LOGGER.debug(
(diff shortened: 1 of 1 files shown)


marimo/_server/session/serialize.py
```diff
@@ -82,7 +82,8 @@ def serialize_session_view(
         if view.cell_ids is not None:
             cell_ids = view.cell_ids.cell_ids
         else:
-            LOGGER.warning(
+            # TODO: Reconcile HTML, and json exports. Something seems a bit off.
```
akshayka:
```suggestion
            # TODO: This is a problem for exporting to HTML, but not for caching the session view for replays. Something seems off.
```

akshayka (approved):

PR #5427, opened 2025-06-24
refactor: make websocket-jsonrpc server bridge generic

winstxnhdw:
## 📝 Summary

Required by #5302

This PR introduces a rewrite of the server bridge, which will allow future LSPs (such as `ty`) to spin up a dedicated bridge for their processes.

The current WebSocket to JSON-RPC bridge is tightly coupled to Copilot with a lack of separation of concerns. This means any modification to `lsp.py` will also lead to changes in the bridge server implementation.

2025-06-24_02-07-07-ezgif com-crop

## 🔍 Description of Changes

- Added a `--lsp` flag
- Add bi-directional socket clean-up
- Get rid of every `any`
- Get rid of globals

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

// cc: @nojaf

### DIFF
--- a/lsp/index.ts
+++ b/lsp/index.ts
@@ -92,6 +92,12 @@ function startWebSocketServer(
 
 function main(): void {
   const argv = parseArgs(process.argv.slice(2));
+
+  if (argv.help) {
+    console.log('Usage: node index.cjs --lsp "<command>" [--port <port>]');
+    return;
+  }
+
   const serverPort = Number.parseInt(argv.port) || 3000;
   const languageServerCommand = argv.lsp.split(" ");
 
--- a/lsp/index.ts
+++ b/lsp/index.ts
@@ -1,4 +1,6 @@
+import { appendFile, mkdir, writeFile } from "node:fs/promises";
 import type { IncomingMessage } from "node:http";
+import { dirname } from "node:path";
 import type { IWebSocket } from "@sourcegraph/vscode-ws-jsonrpc";
 import { forward } from "@sourcegraph/vscode-ws-jsonrpc/lib/server/connection.js";
 import {
@@ -9,25 +11,75 @@ import parseArgs from "minimist";
 import type { CloseEvent, Data, ErrorEvent, MessageEvent, WebSocket } from "ws";
 import { WebSocketServer } from "ws";
 
-class WebSocketAdapter implements IWebSocket {
-  private webSocket: WebSocket;
+class Logger {
+  private constructor(private readonly logFilePath: string) {
+    this.logFilePath = logFilePath;
+  }
 
-  constructor(webSocket: WebSocket) {
+  static async create(logFilePath: string): Promise<Logger> {
+    await mkdir(dirname(logFilePath), { recursive: true });
+    await writeFile(logFilePath, "");
+
+    return new Logger(logFilePath);
+  }
+
+  private async appendToLogFile(...args: unknown[]): Promise<void> {
+    const log = args.join(" ");
+
+    try {
+      await appendFile(this.logFilePath, `${log}\n`);
+    } catch (error) {
+      console.error("Failed to write to log file:", error);
+    }
+  }
+
+  debug(...args: Parameters<typeof console.log>): void {
+    console.log(...args);
+    void this.appendToLogFile("[DEBUG]", ...args);
+  }
+
+  log(...args: Parameters<typeof console.log>): void {
+    console.log(...args);
+    void this.appendToLogFile("[INFO]", ...args);
+  }
+
+  error(...args: Parameters<typeof console.error>): void {
+    console.error(...args);
+    void this.appendToLogFile("[ERROR]", ...args);
+  }
+}
+
+class WebSocketAdapter implements IWebSocket {
+  constructor(
+    private readonly webSocket: WebSocket,
+    private readonly logger: Logger,
+  ) {
     this.webSocket = webSocket;
+    this.logger = logger;
   }
 
   send(content: string): void {
-    this.webSocket.send(content);
+    try {
+      this.webSocket.send(content);
+    } catch (error) {
+      this.logger.error("Failed to send message:", error);
+    }
   }
 
   onMessage(callback: (data: Data) => void): void {
     this.webSocket.onmessage = (event: MessageEvent) => {
-      callback(event.data);
+      try {
+        this.logger.debug("Received message:", event.data);
+        callback(event.data);
+      } catch (error) {
+        this.logger.error("Error handling message:", error);
+      }
     };
   }
 
   onError(callback: (message: string) => void): void {
     this.webSocket.onerror = (event: ErrorEvent) => {
+      this.logger.error("WebSocket error:", event);
       if ("message" in event) {
         callback(event.message);
       }
@@ -36,6 +88,10 @@ class WebSocketAdapter implements IWebSocket {
 
   onClose(callback: (code: number, reason: string) => void): void {
     this.webSocket.onclose = (event: CloseEvent) => {
+      this.logger.log(
+        `WebSocket closed with code ${event.code}: ${event.reason}`,
+      );
+
       callback(event.code, event.reason);
     };
   }
@@ -47,6 +103,7 @@ class WebSocketAdapter implements IWebSocket {
 
 function handleWebSocketConnection(
   languageServerCommand: string[],
+  logger: Logger,
   webSocket: WebSocket,
   _: IncomingMessage,
 ): void {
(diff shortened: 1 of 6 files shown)


tests/_server/test_lsp.py
```diff
@@ -123,7 +122,7 @@ def test_pylsp_server():
 def test_copilot_server():
     server = CopilotLspServer(port=8000)
     assert isinstance(server.validate_requirements(), (str, bool))
-    if Path(server._lsp_bin()).exists():
+    if (server._lsp_dir() / "index.cjs").exists():
```
winstxnhdw: It’s probably better that I preserve `_lsp_bin()` since `index.cjs` is an implementation detail.

lsp/index.ts
```diff
@@ -1,188 +1,107 @@
...
-      };
-    },
-    dispose: () => {
-      try {
-        webSocket.close();
-      } catch (error) {
-        Logger.error("Error closing WebSocket:", error);
-      }
-    },
-  };
```
manzt: I don't see anything different from this and the class-based adapter (`WebSocketAdapter`). I'd prefer to keep this to reduce the diff and also preserve logging.

lsp/index.ts
```diff
@@ -1,188 +1,107 @@
...
+  const webSocketServer = new WebSocketServer({
+    port,
+    perMessageDeflate: false,
+  });
 
-    let langServer: string[] | undefined;
+  webSocketServer.on("connection", (webSocket, request) =>
+    handleWebSocketConnection(languageServerCommand, webSocket, request),
+  );
+}
```
manzt: It would be good to preserve the logging from before here and elsewhere.

lsp/index.ts
```diff
@@ -1,188 +1,107 @@
...
+
+  if (argv.help) {
+    console.log('Usage: node index.cjs --lsp "<command>" [--port <port>]');
+    return;
+  }
+
+  const serverPort = Number.parseInt(argv.port) || 3000;
+  const languageServerCommand = argv.lsp.split(" ");
+
+  startWebSocketServer(serverPort, languageServerCommand);
```
manzt: We should consider adding a `--log-file` flag here as well, so the caller can define where logs should be written (e.g., we do this with pylsp).

manzt:
I think rebasing changes on main should hopefully fix some of the schema errors.

winstxnhdw:
Agree with the changes. Will work on this soon.

lsp/index.ts
winstxnhdw: I tried writing something like this and found that the diff didn't get much better because of the added type annotations and how we are passing the logger in the rewrite, so I went and reverted it back to the class-based adapter, just so we could use more TypeScript-like keywords like `private` and `readonly`. I am aware that not many devs might care about such patterns, and if you still prefer the adapter to be written in this way, I am happy to make the change again.

winstxnhdw:
Hey @manzt, sorry for the delay. This PR is now ready for another review.

marimo/_server/lsp.py
```diff
@@ -150,22 +151,37 @@ def validate_requirements(self) -> Union[str, Literal[True]]:
...
+            "--lsp",
+            f"node {copilot_bin} --stdio",
+            "--log-file",
+            str(
+                Path(environ.get("XDG_CACHE_HOME", Path.home()))
+                / ".cache"
+                / "marimo"
+                / "logs"
+                / "github-copilot-lsp.log"
+            ),
```
manzt: I think like in `pylsp` we could probably do:

```python
log_file = _loggers.get_log_directory() / "github-copilot-lsp.log"
```

```suggestion
            str(log_file)
```

lsp/package.json
```diff
@@ -13,6 +13,7 @@
     "@github/copilot-language-server": "^1.322.0"
   },
   "devDependencies": {
+    "@biomejs/biome": "^2.0.5",
```
manzt: Let's keep this PR scoped to the changes in `index.ts`. We could have a follow up to add linting/formatting. Would you mind restoring this file and `pnpm-lock.yaml` back to what is on `main`?

```sh
git co main -- lsp/package.json
git co main -- lsp/pnpm-lock.yaml
```

lsp/index.ts
```diff
@@ -1,188 +1,184 @@
...
   });
-} catch (error) {
-  Logger.error("[FATAL] Failed to start WebSocket server:", error);
-  process.exit(1);
 }
+
+async function main(): Promise<void> {
+  const argv = parseArgs(process.argv);
+
+  if (argv.help) {
```
manzt, line 169: nice!

lsp/index.ts
manzt: I see. Personally, I tend to avoid TypeScript features that aren't [erasable syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#the---erasablesyntaxonly-option), but since we don’t have linting rules around that, it's fine to use your preference.

Originally, I imagined a version where `LOG_FILE: str | undefined` is set globally in main to keep the diff smaller, but I’ve reviewed the changes closely and I’m happy with this approach. I think it’s better overall—just took a little longer to understand the changes on my part :) I might revisit it later, but no blockers from me.

manzt (commented):
No apologies necessary! Thanks for taking the time and for your contributions. Logging is working well for me; this is a great addition. Just a few minor nits, and we should be good to merge!

marimo/_server/lsp.py
winstxnhdw: Yeah, totally agree. No idea how I missed this. Thanks!

winstxnhdw:
Should be good now!

lsp/index.ts
```diff
@@ -1,188 +1,184 @@
...
-    void appendToFileLog("[INFO]", ...args);
-  },
-  error: (...args: any[]) => {
-    console.error(...args);
-    void appendToFileLog("[ERROR]", ...args);
-  },
-};
-
-// Adapted from https://github.com/wylieconlon/jsonrpc-ws-proxy
-const COPILOT_LSP_PATH = path.join(__dirname, "copilot", "language-server.js");
```
dmadisetti, line 47: Note on this, we've had bundling errors before- would appreciate some sort of check

winstxnhdw: How should I check? I can see that the `copilot` directory is indeed copied over into `marimo` correctly.

winstxnhdw: image

I did `hatch build` and unzipped the wheel. You can see it here.

manzt: > Note on this, we've had bundling errors before- would appreciate some sort of check

Would you be able to elaborate re: bundling errors?

My understanding is that copilot (`language-server.js`) isn't a part of the bundle at all, and is just copied into `dist/` along side the bundled `index.ts`:

https://github.com/marimo-team/marimo/blob/cfea07c38c13508f5af2fad420f00e5273408d24/lsp/package.json#L7-L7

Previously, it was important that `./copilot/language-server.js` remained relative to `index.cjs`, since the path was hardcoded. This PR makes `index.cjs` more general, so the path to any LSP (not just Copilot) can now be passed in as a CLI argument from Python: ( [marimo/_server/lsp.py](https://github.com/marimo-team/marimo/pull/5427/files/ebb10f83ef6ace548626bbe66cc6b72af8a704af#diff-924fb831b6a140171580f271323e105d24ebdc812b2d2a43de30975006a954c2)).

manzt: If we really want to assert the file is copied over into `marimo/`. We could lift `_lsp_bin` into a function and add a static property (e.g., `CopilotLspServer.bin`) and assert it exists in some test:

```python
class CopilotLspServer(BaseLspServer):
    
    @staticmethod
    def bin() -> Path:
        return _lsp_bin().parent / "copilot" / "language-server.js"
```

But a plain pytest smoke test might be misleading. It could pass in dev (`uv run pytest`) but fail elsewhere if static files like the `.js` binary are missing. The more reliable approach would be to test from an actual build of marimo, import `CopilotLspServer`, and assert the file exists there.

But that seems a bit ovekill. I don't think we assert other static assets exist? @dmadisetti

dmadisetti: CI build chain and local build chain bifurcated at some point, this was just an easy sanity check that the bundling worked. This should be fine as long as

`node marimo/_lsp/index.cjs --help > /dev/null`

works or `scripts/test-lsp.sh` is updated accordingly. It was just a moment that lead to a bit of confusion and a released broken copilot

winstxnhdw: There's a test case for this and it passed.

dmadisetti: Cool- the test is run on PR. Just weary from the previous time

Issue #5323, opened 2025-06-17
Add the option to disable the "Publish HTML to web" feature


kyrre:
### Description

Provide a way to disable the "Share -> Publish HTML to web feature". 

Context: We are using marimo to handle security incidents and it would be very bad if we were to inadvertently publish an undeltable (?) HTML notebook.

### Suggested solution

Add a field to marimo.toml which disables the functionality and removes it from the UI.

### Will you submit a PR?

- [ ] Yes

### Alternative

On discord it was proposed to add custom CSS for hiding the option.

kyrre:
I will attempt to disable the functionality using custom CSS as suggested!

PR #5432, opened 2025-06-24
feat: configuration to disable sharing actions

kyrre:
## 📝 Summary

This adds an option to disable (hide) all the sharing functionality, e.g., "Share > Publish HTML to web". The issue is tracked in #5323 

image
image

## 🔍 Description of Changes

Introduces a new configuration setting "sharing", which is enabled by default. When it is disabled all the sharing related actions are hidden in the UI.

## 📋 Checklist

## 📜 Reviewers

@mscolnick


kyrre:
Am I supposed to commit the OpenAPI code that is generated? 🤔

kyrre:
recheck

kyrre:

Light2Dark:
Yeap, you can commit it, there is also a pre-commit hook that is failing (& merge conflict), so you may want to fix that.

marimo/_config/config.py
```diff
@@ -381,6 +381,19 @@ class DatasourcesConfig(TypedDict):
...
+class SharingConfig(TypedDict):
+    """Configuration for sharing features.
+
+    **Keys.**
+
+    - `enabled`: if `False`, sharing options will be hidden from the UI
+    """
+
+    enabled: bool
+
```
akshayka: I think this could be more granular:

* html
* wasm

both booleans. wasm doesn't leak (we don't save code).

frontend/src/components/app-config/user-config-form.tsx
```diff
@@ -554,6 +554,41 @@ export const UserConfigForm: React.FC = () => {
...
+                    </FormItem>
+                    <FormDescription>
+                      When unchecked, sharing options (publish HTML, create
+                      WebAssembly link) will be hidden from the notebook actions
+                      menu.
+                    </FormDescription>
+                  </div>
+                )}
+              />
+            </SettingGroup>
```
akshayka: I might remove this. If an org is disabling these feature for their team (via that `pyproject.toml`), likely you don't want to make it easy for developers to turn it back on in the UI.

akshayka (changes requested):
Thanks for the PR! A couple of comments.

frontend/src/components/app-config/user-config-form.tsx
kyrre: > I might remove this. If an org is disabling these feature for their team (via that `pyproject.toml`), likely you don't want to make it easy for developers to turn it back on in the UI.

That's a good point, but would it even be possible for the user to do so? I was just reading the `pyproject.toml` setting in the [Configuration documentation](https://docs.marimo.io/guides/configuration/#overriding-settings-with-pyprojecttoml) and it states that:

> Settings overridden in pyproject.toml or script metadata cannot be changed through the marimo editor's settings menu. Any changes made to overridden settings in the editor will not take effect.

But, anyway, it's not a very useful UI feature (how often would the user actually want to change it?) so I'll just remove it from the PR.

kyrre:
recheck

kyrre:

kyrre:
> Yeap, you can commit it, there is also a pre-commit hook that is failing (& merge conflict), so you may want to fix that.

I think it's only the CLA that is failing now and I can't get it working 🤷🏻

akshayka:
Thanks for the PR! Not sure why the CLA bot isn't happy but that's fine, we have a record in the form of your comment here.

kyrre:
> Premature approval, noticed a test is failing.
> 
> Also, if the sharing config is not required, then it shouldn't have a default value `config.py` / `config-schema.ts` I believe.

Sorry about that, I'm still getting familiar with the marimo codebase 😊

I've removed the default values from the config files. Since that means sharing.* can now be undefined, I added fallback logic in the UI: 

```js
// Fallback: if sharing is undefined, both are enabled by default
const sharingHtmlEnabled = resolvedConfig.sharing?.html ?? true;
const sharingWasmEnabled = resolvedConfig.sharing?.wasm ?? true;
```

Does this work for you, or did you have something else in mind? 🤔

PR #5434, opened 2025-06-24
improvement: install packages directly

Light2Dark:
## 📝 Summary

Just a refactor to install packages with one-click rather than opening the sidebar to install them.

## 📋 Checklist

## 📜 Reviewers

@manzt

### DIFF
--- a/frontend/src/components/data-table/column-explorer-panel/column-explorer.tsx
+++ b/frontend/src/components/data-table/column-explorer-panel/column-explorer.tsx
@@ -78,7 +78,7 @@ export const ColumnExplorerPanel = ({
           className="h-3 w-3 ml-1 mt-0.5"
         />
       </span>
-      <Command className="h-5/6" shouldFilter={false}>
+      <Command className="h-5/6 bg-background" shouldFilter={false}>
         <CommandInput
           placeholder="Search columns..."
           value={searchValue}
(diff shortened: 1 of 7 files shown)


manzt:
Would you mind including a screenshot / video showing how this feature works? I checked out the branch and can't seem to find the "one-click" flow this introduces.

Light2Dark:
https://github.com/user-attachments/assets/b272fdfe-dfb5-4330-b2ae-14455fa80604

Previously it would open up the Packages panel, auto input the packages to install and wait for the user to click 'Add'. I think this smoothens that process. And also makes a reusable hook.

~Next step would be to refresh the chart view upon install. We should do this for the Datasources panel too.~

manzt (approved):
Thanks for the screenshot—this looks great!

I noticed that the new custom hook uses separate `useState` calls for `loading`, `error`, and `success`, but only `loading` is currently being used in the UI.

For now, I'd suggest one of the following approaches:


1.) Remove the unused error and success states to keep the implementation minimal.
2.) Adopt a state machine pattern (e.g., using `useAsyncData` or `AsyncDataResult` or something similar) for better type safety and clearer state transitions.

Using separate `useState` calls can result in inconsistent combinations (e.g., loading and error both being true), and proper handling of success/error states usually involves things like cancellation support. For more context, see: https://github.com/marimo-team/marimo/pull/5272.

My feeling here would be to just go with the former since we aren't using the additional states in the code.

Light2Dark:
removed the states and added the refetch feature.

manzt (approved):
Looks great!

PR #5436, opened 2025-06-24
dev: Use islands as local asset server

dmadisetti:
## 📝 Summary

quarto `0.14.7` is broken.

I'm having trouble confirming this is the issue by replicating it locally, since I can't replicate the breakage. My dev served assets seem to work in general. But `being islands` != `being wasm` as is, and I think some of the logic around sockets gets lost (0.14.6 works, so this may be the offending PR: #5403)

@akshayka

### DIFF
--- a/frontend/vite.config.mts
+++ b/frontend/vite.config.mts
@@ -13,6 +13,7 @@ const TARGET = `http://${HOST}:${SERVER_PORT}`;
 const isDev = process.env.NODE_ENV === "development";
 const isStorybook = process.env.npm_lifecycle_script?.includes("storybook");
 const isPyodide = process.env.PYODIDE === "true";
+const isIslands = process.env.VITE_MARIMO_ISLANDS === "true";
 
 const htmlDevPlugin = (): Plugin => {
   return {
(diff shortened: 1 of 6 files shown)


akshayka (approved):
Lgtm modulo type check / ci

frontend/vite.config.mts
```diff
@@ -252,6 +253,8 @@ export default defineConfig({
       : {},
   },
   define: {
+    // Used when serving local assets in dev mode for islands.
+    "import.meta.env.VITE_MARIMO_ISLANDS": JSON.stringify(isIslands),
```
manzt: nit: Vite automatically exposes `VITE_`-prefixed [env vars](https://vite.dev/guide/env-and-mode.html#env-variables) to `import.meta.env` as strings, so there's no need to manually inject `VITE_MARIMO_ISLANDS` via `define` config. 

Using `define` with `JSON.stringify(bool)` changes this behavior by inlining a boolean intead of a string, which can break type expectations (I believe TypeScript assumes `import.meta.env.VITE_*` values are string). 

We should update `isIslands()` from:

```ts
export function isIslands() {
  return import.meta.env.VITE_MARIMO_ISLANDS === true;
}
```

to:

```ts
export function isIslands() {
  return import.meta.env.VITE_MARIMO_ISLANDS === "true";
}
```

manzt (approved):
Small nit regarding env vars.

frontend/vite.config.mts
dmadisetti: Oh nice. Thank you

PR #5438, opened 2025-06-24
fix: disable controls when not connected

akshayka:
Disable the following controls when the frontend is not connected to a backend:

* shutdown
* user configuration
* notebook actions

Some of the actions in notebook actions may be safe when not connected, but coarsely disabling the entire menu is the simplest solution for now.

### DIFF
--- a/frontend/src/components/editor/inputs/Inputs.styles.ts
+++ b/frontend/src/components/editor/inputs/Inputs.styles.ts
@@ -18,7 +18,7 @@ export const button = cva(
         yellow: "mo-button yellow",
         // for actions that will change state but are not destructive
         "hint-green": "mo-button hint-green",
-        disabled: "mo-button disabled",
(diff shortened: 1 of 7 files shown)


frontend/src/components/editor/inputs/Inputs.styles.ts
```diff
@@ -18,7 +18,7 @@ export const button = cva(
         yellow: "mo-button yellow",
         // for actions that will change state but are not destructive
         "hint-green": "mo-button hint-green",
-        disabled: "mo-button disabled",
+        disabled: "mo-button disabled active:shadow-xsSolid",
```
akshayka, line 21: This is a small detail, but it prevents the user from visually depressing the button when it is disabled.

akshayka:
Marking as a draft, forgot about the use of these controls on the home page.

PR #5440, opened 2025-06-24
chore(frontend): Remove simple cases of `as any`

manzt:
Replaces `as any` with safer alternatives or runtime checks to make
TypeScript happier.


frontend/src/components/data-table/columns.tsx
```diff
@@ -152,8 +152,7 @@ export function generateColumns<T>({
       // may have periods in them ...
       // https://github.com/TanStack/table/issues/1671
       accessorFn: (row) => {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        return (row as any)[key];
+        return row[key as keyof T];
```
manzt, line 155: Not much better, but:

```ts
row[key as keyof T] // is of type T[keyof T]
```

whereas:

```ts
(row as any)[key] // is any
```

Light2Dark (approved):

PR #5444, opened 2025-06-25
Add light/dark/system settings from command 

koaning:
When a user types "dark" they will find the dark mode setting, but they won't when they type "light". Figured we should add that extra keyword to make it easy for the user to search for it.

### DIFF
--- a/frontend/src/components/editor/actions/useConfigActions.tsx
+++ b/frontend/src/components/editor/actions/useConfigActions.tsx
@@ -35,7 +35,7 @@ export function useConfigActions() {
         },
       })),
     {
-      label: "Config > Set dark mode",
+      label: "Config > Set theme: dark",
       handle: () => {
         handleUserConfig({
           ...config,
@@ -47,7 +47,7 @@ export function useConfigActions() {
       },
     },
     {
-      label: "Config > Set light mode",
+      label: "Config > Set theme: light",
       handle: () => {
         handleUserConfig({
           ...config,
@@ -59,7 +59,7 @@ export function useConfigActions() {
       },
     },
     {
-      label: "Config > Set light/dark mode to system setting",
+      label: "Config > Set theme: system",
       handle: () => {
         handleUserConfig({
           ...config,


frontend/src/components/editor/actions/useConfigActions.tsx
```diff
@@ -4,11 +4,9 @@ import { useAppConfig, useResolvedMarimoConfig } from "@/core/config/config";
 import type { AppConfig, UserConfig } from "@/core/config/config-schema";
 import { getAppWidths } from "@/core/config/widths";
 import { saveAppConfig, saveUserConfig } from "@/core/network/requests";
-import { useTheme } from "@/theme/useTheme";
 import type { ActionButton } from "./types";
 
 export function useConfigActions() {
-  const { theme } = useTheme();
```
manzt, line 11: I think it would be nice if we actually used the `theme` below to set the optional `hidden` property on each of the options. That way, we could show only the other options when someone types in "theme". For example:

```diff
    {
      label: "Config > Set theme: light",
++    hidden: theme === "light",
      handle: () => {
        handleUserConfig({
          ...config,
          display: {
            ...config.display,
            theme: "light",
          },
        });
      },
    },
```

manzt (approved):
Looks great! Just an idea to conditionally hide the option for the current theme.

frontend/src/components/editor/actions/useConfigActions.tsx
koaning: I've personally never been a fan of this because it makes me think "oh? I guess that command didn't exist because when I type the keyword it does not appear". I'm open to be convinced otherwise, but want to push back a little bit because I do think there's a consistency argument to be made here, especially because we don't have that *that* many settings.

manzt: I appreciate the push back and don't feel strongly. In the other cases where we enable `hidden` I guess it's to enable/disable a boolean ("enable copilot" / "disable copilot"), but here the commands set a value. Let's just merge as is.

PR #5452, opened 2025-06-26
feat: Build plots and charts from tables, with Python code generation

Light2Dark:
## 📝 Summary

https://github.com/user-attachments/assets/fc57c39c-565f-4433-b586-7052b52941b5

- Release chart builder in tables. 
- Introduce some QOL improvements in charts (better specification defaults, tooltips, collapsible sidebar) and some bug fixes.
- Table panels (bottom left) now conditionally appear depending on the mode marimo is run in.
- In edit mode, all panels are available while in other modes only the row-viewer panel appears.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/charts/charts.tsx
+++ b/frontend/src/components/data-table/charts/charts.tsx
@@ -220,6 +220,8 @@ export const TablePanel: React.FC<TablePanelProps> = ({
   );
 };
 
+const CHART_PLACEHOLDER_CODE = "X and Y columns are not set";
+
 export const ChartPanel: React.FC<{
   chartConfig: ChartSchemaType | null;
   chartType: ChartType;
@@ -302,7 +304,7 @@ export const ChartPanel: React.FC<{
   const developmentMode = import.meta.env.DEV;
 
   const renderChartDisplay = () => {
-    let altairCodeSnippet = "X and Y columns are not set";
+    let altairCodeSnippet = CHART_PLACEHOLDER_CODE;
     if (typeof specWithoutData !== "string") {
       altairCodeSnippet = generateAltairChartSnippet(
         specWithoutData,
@@ -342,7 +344,11 @@ export const ChartPanel: React.FC<{
           {memoizedChart}
         </TabsContent>
         <TabsContent value="code">
-          <CodeSnippet code={altairCodeSnippet} language="python" />
+          <CodeSnippet
+            code={altairCodeSnippet}
+            insertNewCell={altairCodeSnippet !== CHART_PLACEHOLDER_CODE}
+            language="python"
+          />
         </TabsContent>
         {developmentMode && (
           <>
--- a/frontend/src/components/data-table/charts/components/layouts.tsx
+++ b/frontend/src/components/data-table/charts/components/layouts.tsx
@@ -101,8 +101,9 @@ export const AccordionFormContent: React.FC<{
 
 export const CodeSnippet: React.FC<{
   code: string;
+  insertNewCell?: boolean;
   language?: "python" | "sql";
-}> = ({ code, language }) => {
+}> = ({ code, insertNewCell, language }) => {
   return (
     <ReadonlyCode
       minHeight="330px"
@@ -111,7 +112,7 @@ export const CodeSnippet: React.FC<{
       language={language}
       showHideCode={false}
       showCopyCode={false}
-      insertNewCell={true}
+      insertNewCell={insertNewCell}
     />
   );
 };
--- a/frontend/src/components/data-table/charts/forms/common-chart.tsx
+++ b/frontend/src/components/data-table/charts/forms/common-chart.tsx
@@ -44,6 +44,16 @@ export const CommonChartForm: React.FC = () => {
 
   return (
     <>
+      <Tooltip
+        delayDuration={100}
+        content="To persist a chart, add the generated Python code to a new cell."
+      >
+        <div className="flex items-center gap-1">
+          <TriangleAlert className="h-3.5 w-3.5 mb-0.5 text-muted-foreground" />
+          <p className="text-sm text-muted-foreground">Charts are not saved.</p>
+        </div>
+      </Tooltip>
+
       <XAxis />
       <YAxis />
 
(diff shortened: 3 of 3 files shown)


marimo/_plugins/ui/_impl/table.py
```diff
@@ -289,6 +292,13 @@ def style_cell(_rowId, _columnName, value):
             If "stats", only show stats. If "chart", only show charts.
         show_download (bool, optional): Whether to show the download button.
             Defaults to True for dataframes, False otherwise.
+        show_page_size_selector (Union[List[int], Literal[False]], optional): The page sizes to show in the page size selector.
```
akshayka: I think `table` has too many arguments as it is. My preference is to not add these additional arguments, and only add them if we really really need them.

marimo/_plugins/ui/_impl/table.py
```diff
@@ -473,6 +493,25 @@ def __init__(
                 TableManager.DEFAULT_SUMMARY_STATS_ROW_LIMIT
             )
 
+        app_mode = get_mode()
+        if app_mode != "edit":
+            show_column_explorer = False
+            show_chart_builder = False
```
akshayka: Should we hide the row viewer too?

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1182,3 +1225,7 @@ def _validate_column_formatting(
             raise ValueError(
                 f"Column '{next(iter(invalid))}' not found in table."
             )
+
```
akshayka: Delete dead code

marimo/_plugins/ui/_impl/table.py
Light2Dark: Got it, we could also move this to a global config instead if needed in the future.

marimo/_plugins/ui/_impl/table.py
Light2Dark: I think row viewer is quite useful & is not reliant on external deps. So, we could keep it imo.

akshayka (changes requested):
Not necessarily changes to the code — but since this PR GAs charts, we should definitely highlight it in release notes and feature this PR prominently there.

I've added the release highlight tag to the PR.

Can you:
* Update the PR description to describe the entire charts feature in a few paragraphs.
* Add media, such as a video, that demos the feature?

akshayka (commented):
Some comments from playing with charts.

* Can we use `ChartSpline` instead of `ChartBar` for the icon in the table footer?
* Bar chart aggregation dropdown menu. I feel like `Bin` is going to be the most common aggregation. Can we move it to the top?
* Faceting section. Can you add a `?` icon with a tooltip that explains what faceting is?
* Faceting. Does it really make sense to facet by numerical columns? I feel like only categorical columns should support faceting.
* More generally, has much thought gone into eliminating values from various chart fields when they don't make sense? For example, for a line chart, the only y-value that makes sense to me is numerical.
* We should tell the user somehow that charts won't be persisted. It would be a sad first impression if a user carefully crafted lots of charts but didn't save them to their notebook (not knowing they wouldn't be persisted) and lost them all.
* Scatter. Sorting the x-axis by ascending/descending doesn't appear to do anything (my axes are both numerical). Sorting doesn't make much sense to me as an operation for a scatter chart anyway. It also doesn't seem to do anything on Heatmap. Can we just remove the axis sorting?
* Scatter color-by stack doesn't make sense to me. For example here the y-axis is wrong (the max displacement in this plot is ~350, and yet when I stack the values go up to 8000)

Releasing this feature is a big (awesome!) change. We should also have a docs page under > Working with data (either a new page, or a subsection in  "Plotting"). The docs can be in a separate PR.

Light2Dark:
Thanks for testing @akshayka , I've made some changes in https://github.com/marimo-team/marimo/pull/5392 and going to mark this as draft until those are merged.

Generally, we can include those fields (eg. faceting numerical and categorical y-axis), but we need to set better defaults. Numerical faceting should be binned for example.

Additionally, I'd prefer a soft release because I expect a lot of bugs to be ironed out, but adding to the docs and having some media is a good idea.

manzt (approved):
Looks good to me 👍 I was initially going to suggest wrapping these settings in a `React.Context`, but I noticed the rest of the props are passed down in a similar way and would prefer to be consistent.

akshayka:
> Thanks for testing @akshayka , I've made some changes in https://github.com/marimo-team/marimo/pull/5392 and going to mark this as draft until those are merged.

Thanks, taking a look (edit: approved)

akshayka:
@Shahmir one more piece of feedback:

* For scatter plots, `grid` should be `True`.

> Some comments from playing with charts.
> 
> * Can we use `ChartSpline` instead of `ChartBar` for the icon in the table footer?
> * Bar chart aggregation dropdown menu. I feel like `Bin` is going to be the most common aggregation. Can we move it to the top?
> * Faceting section. Can you add a `?` icon with a tooltip that explains what faceting is?
> * Faceting. Does it really make sense to facet by numerical columns? I feel like only categorical columns should support faceting.
> * More generally, has much thought gone into eliminating values from various chart fields when they don't make sense? For example, for a line chart, the only y-value that makes sense to me is numerical.
> * We should tell the user somehow that charts won't be persisted. It would be a sad first impression if a user carefully crafted lots of charts but didn't save them to their notebook (not knowing they wouldn't be persisted) and lost them all.
> * Scatter. Sorting the x-axis by ascending/descending doesn't appear to do anything (my axes are both numerical). Sorting doesn't make much sense to me as an operation for a scatter chart anyway. It also doesn't seem to do anything on Heatmap. Can we just remove the axis sorting?
> * Scatter color-by stack doesn't make sense to me. For example here the y-axis is wrong (the max displacement in this plot is ~350, and yet when I stack the values go up to 8000)
> 
> image
> 
> Releasing this feature is a big (awesome!) change. We should also have a docs page under > Working with data (either a new page, or a subsection in "Plotting"). The docs can be in a separate PR.

Light2Dark:
- changed to chartspline
- bin moved to the top
- added binning to numerical faceting by default
- scatter charts have grid on by default
- removed sorting for scatter, heatmap charts
- stacking only allowed for bar and line charts
- for categorical fields on y-axis, a default aggregation is set to count
- added a tooltip to provide persistence info

CleanShot 2025-07-04 at 00 52 39

So, I don't think anything is missed. I think there could be more polish (eliminating nonsensical fields, edge cases). I will create a general ticket for this

akshayka (approved):
I pushed two small things:

* Updated the tooltip.
* Don't show the add code button when there's no code to add.

I also updated the PR title to help with release notes.

PR #5453, opened 2025-06-26
chore(frontend): Remove "vega" as top-level dependency

manzt:
All runtime uses of "vega" were replaced with "vega-loader" in #5449,
and now the only remaining usage is for types.

This change removes "vega" entirely as a direct dependency and switches
to extracting necessary types from "vega-lite" (which we are using
directly and extensively). It should also simplify future updates, since
we no longer need to keep "vega" and "vega-lite" in sync: all types are
now derived from what "vega-lite" exports. Should also prevent
accidental re-imports from "vega" in the future.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -150,7 +150,6 @@
     "timestring": "^7.0.0",
     "typescript-memoize": "^1.1.1",
     "use-resize-observer": "^9.1.0",
-    "vega": "^5.33.0",
     "vega-lite": "^5.23.0",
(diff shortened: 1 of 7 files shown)


Light2Dark:
nice!

Light2Dark:
lgtm, I spotted 2 issues (unrelated) with altair charts when I tested `marimo/_smoke_tests/altair/vegafusion_nb.py`. Will file a ticket

PR #5474, opened 2025-06-27
register UIElement on change if it is missing from the UIElementRegistry

### LINKED ISSUE #5465: Cached UI elements removed from frontend UIElementRegistry
### Describe the bug

When a notebook automatically reloads from a change in an imported module, UI elements that are returned from a cached function get erased from the `UIElementRegistry`, meaning that changes in the element no longer flow through to other cells.

https://github.com/user-attachments/assets/622411fd-3904-426e-b816-3a315beea9d4

I forgot to record it, but if I had manually run the last cell it still would have shown the old values.

I originally ran into this in the context of module change based reloading, but after some further investigation this is much easier to trigger than I thought. Simply re-running the cell that calls the cached function is enough to trigger the behaviour.

### What I think is going wrong

I suspect this is because the UI elements get removed on cell invalidation, but only get registered on UI element creation. These UI elements aren't being created new, and so aren't being registered anew when the cell reloads. This is only my intuition though, I haven't dug through the implementation in enough detail to know that's what's happening.

I have somewhat validated that assumption by commenting out the entry deletion here and verifying that the above behaviour no longer occurs.

https://github.com/marimo-team/marimo/blob/ae6a394c0335e68d0077a71c95bf8688ed6205e3/frontend/src/core/dom/uiregistry.ts#L121

### Will you submit a PR?

- [x] Yes

### Environment

<details>

```
{
  "marimo": "0.14.0",
  "OS": "Darwin",
  "OS Version": "24.5.0",
  "Processor": "arm",
  "Python Version": "3.11.12",
  "Binaries": {
    "Browser": "137.0.7151.120",
    "Node": "v24.1.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8",
    "narwhals": "1.41.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "starlette": "0.46.2",
    "tomlkit": "0.13.2",
    "typing-extensions": "4.14.0",
    "uvicorn": "0.34.3",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "anywidget": "0.9.18",
    "pandas": "2.2.3",
    "pyarrow": "20.0.0",
    "loro": "1.5.1",
    "pytest": "8.4.0",
    "sqlglot": "26.28.0"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.14.0"

@app.cell
def _():
    import marimo as mo

    import example

    return (
        mo,
        example,
    )


@app.cell
def _(mo):
    @mo.cache
    def get_multiselect(thing1, thing2):
        return mo.ui.array([
            mo.ui.multiselect(options=thing1, value=thing1, label="Thing 1"),
            mo.ui.multiselect(options=thing2, value=thing2, label="Thing 2"),
        ])
    return (get_multiselect,)


@app.cell
def _(get_multiselect):
    filters = get_multiselect(example.thing1, example.thing2)
    filters
    return (filters,)


@app.cell
def _(filters):
    print(filters[0].value)
    return
```

example.py

```python
thing1 = ["a", "b", "c"]
thing2 = ["d", "e", "f"]

def other():
    pass
```

buckley-w-david:
## 📝 Summary

fixes #5465 

When a notebook uses `marimo.cache` for a function that returns UI elements, those UI elements become disconnected from the rest of the notebook after the cell that calls the function re-runs.

I believe this is because UI elements do not end up being disconnected/reconnected in this case, meaning they are never re-added to the UIElementRegistry after they were removed on cell invalidation.

## 🔍 Description of Changes

There may be a more elegant way to solve this, but the best I could come up with was the register the UI element on change if it was missing from the UIElementRegistry.

Before

https://github.com/user-attachments/assets/b0b5c1d2-e999-4ecd-a6bd-049a82914a34

After

https://github.com/user-attachments/assets/a59a6771-a9b7-47ee-8d4e-316feeb209e2

## 📋 Checklist

## 📜 Reviewers

@akshayka OR @mscolnick

### DIFF
--- a/frontend/src/core/dom/ui-element.ts
+++ b/frontend/src/core/dom/ui-element.ts
@@ -100,9 +100,12 @@ export function initializeUIElement() {
         // broadcast? that would still let other elements cancel the event
         // while also reducing the number of event listeners on the document
         if (objectId !== null && e.detail.element === this.firstElementChild) {
-          // A UIElement may be missing from the registery if it was returned from a function that caches return values.
+          // A UIElement may be missing from the registry if it was returned from a function that caches return values.
           if (!UI_ELEMENT_REGISTRY.has(objectId)) {
-            UI_ELEMENT_REGISTRY.registerInstance(objectId, child as HTMLElement);
+            UI_ELEMENT_REGISTRY.registerInstance(
+              objectId,
+              child as HTMLElement,
+            );
           }
 
           UI_ELEMENT_REGISTRY.broadcastValueUpdate(


dmadisetti (approved):
Thanks! LGTM. I might followup with a test just because these things can slip between the cracks on changes.

Will tag you on it

akshayka:
@dmadisetti marked this for bash focus, as this means all elements, even those that are not assigned to globals, will be registered (IIUC). I am not sure what implications this will have, it's a long time since I touched this code (will they still be cleaned up from the FE, for example)

PR #5481, opened 2025-06-27
Fix shift-enter not focusing new cell at end

manzt:
Fixes regression from #5344 where shift-enter on the last cell created a new
cell but didn't focus it. `moveToNextCell` now sets `scrollKey` for new cells
at notebook boundaries, restoring focus behavior without (I think)
reintroducing autoscroll.


akshayka (approved):
Perfect, this won't reintroduce the autoscroll bug. Thanks!

PR #5484, opened 2025-06-28
Fix go-to-definition for function declarations

manzt:
Removed overly aggressive early return introduced in #5415. Adds tests for
variables in different scopes, function declarations.

### DIFF
--- a/frontend/src/core/codemirror/go-to-definition/__tests__/commands.test.ts
+++ b/frontend/src/core/codemirror/go-to-definition/__tests__/commands.test.ts
@@ -6,6 +6,10 @@ import { EditorView } from "@codemirror/view";
 import { afterEach, describe, expect, test } from "vitest";
 import { goToVariableDefinition } from "../commands";
 
+async function tick(): Promise<void> {
+  await new Promise((resolve) => requestAnimationFrame(resolve));
+}
+
 function createEditor(content: string) {
   const state = EditorState.create({
     doc: content,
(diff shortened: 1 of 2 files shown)


frontend/src/core/codemirror/go-to-definition/commands.ts
```diff
@@ -44,14 +44,6 @@ export function goToVariableDefinition(
...
       } // Stop traversal if found
 
-      // Skip function/lambda bodies entirely
-      if (
-        node.name === "LambdaExpression" ||
-        node.name === "FunctionDefinition"
-      ) {
-        return false;
-      }
-
```
manzt, line 54: I added this block in #5415 while trying to fix go-to-definition for scoped variables, but I think the issue was resolved elsewhere. I haven’t been able to reproduce the original problem, so I suggest we revert it.

frontend/src/core/codemirror/go-to-definition/__tests__/commands.test.ts
```diff
@@ -31,30 +35,170 @@ afterEach(() => {
...
+      "
+      def x():
+          ^
+          print("hi")
+
+      def my_func(x):
+          print(x)
+
+      print(x)
+      "
```
manzt, line 127: I switched these tests to use snapshotting because:

- a.) they are so easy to add
- b.) the diffs give better context to understanding failures



image

akshayka (approved):
Nice — great tests.

manzt:
Shoot, sorry missed your review. I just cleaned up some of the tests.

akshayka (approved):

PR #5489, opened 2025-06-29
enhancement: add a signifier for the create cell context menu

akshayka:
Users can right-click the create cell button to get a context menu with Python, SQL, and Markdown cell types, but I would guess most users don't know this. This change adds a signifier on hover letting them know about the context menu.

This change also fixes a usability bug in the create markdown and sql actions by importing marimo as needed when the markdown/sql cells are created.

Screenshot 2025-06-29 at 9 44 38 AM

This is the context menu (which already existed):

image


manzt (approved):
Nice improvement

PR #5491, opened 2025-06-29
Ensure cell action tooltip appears above sidebar panel

### LINKED ISSUE #5490: "New cell above/below" tooltip appears behind side panel
Noticed when reviewing #5489, but confirmed issue was there before. On "medium" and "full" width notebooks, the tooltip for cell actions can appear behind the sidebar panel. It should be above.

Image

I believe the `PanelsWrapper` has `absolute inset-0` which creates a new stacking context, even though the tooltip has `z-50`, it is likely rendered in a different context.

manzt:
Fixes #5490

The create cell button tooltip was appearing behind the sidebar panel because it had `usePortal={false}`, which kept it within the button's stacking context.

Removing this prop allows the tooltip to render in a portal at the document root, ensuring it appears above all panels. I'm not sure of the original reason for disabling the portal, but in my testing, removing seems to resolve the layering issue and doesn't introduce any obvious regressions.

### DIFF
--- a/frontend/src/components/editor/cell/CreateCellButton.tsx
+++ b/frontend/src/components/editor/cell/CreateCellButton.tsx
@@ -45,7 +45,7 @@ export const CreateCellButton = ({
 
   return (
     <CreateCellButtonContextMenu onClick={onClick}>
-      <Tooltip content={finalTooltipContent} usePortal={false}>
+      <Tooltip content={finalTooltipContent}>
         <Button
           onClick={() => onClick?.({ code: "" })}
           className={cn(


manzt:
https://github.com/user-attachments/assets/54afbabd-47c7-4e1d-a784-5d258b56e263

akshayka (commented):
When we used portaling before, the tooltip would sometimes become "stuck" on (IIRC) the top left of the viewport.

If you played with this for a while (try triggering tooltips rapidly, and switching to app view and back?) and didn't see that behavior, then I'm good with this change (though we could label as a focus area for the next bug bash). Otherwise we could raise the z-index, even if it feels hacky?

manzt:
> If you played with this for a while (try triggering tooltips rapidly, and switching to app view and back?) and didn't see that behavior, then I'm good with this change (though we could label as a focus area for the next bug bash).

Thanks for the context, I'll take a closer look. I figured it wasn't there without reason.

> Otherwise we could raise the z-index, even if it feels hacky?

I actually tried something along those lines (though maybe I didn’t go high enough with the z-index), but it didn’t seem to resolve the issue. I’m wondering if it’s because the panel is absolutely positioned, which might be affecting stacking in an unexpected way

manzt:
I played around with this a bit (video attached) and nothing jumped out to me. I'll add the `bash-focus` tag.

https://github.com/user-attachments/assets/1ddff816-b193-4d0d-a90e-bfb7497143ec

PR #5492, opened 2025-06-29
Convert import helpers to use named parameters

manzt:
Used TypeScript's "Convert parameters to destructured object" code action to clarify arguments at call sites. Changes:

```ts
maybeAddMarimoImport(true, createNewCell, lastFocusedCellId);
```

to

```ts
maybeAddMarimoImport({
  autoInstantiate: true,
  createNewCell,
  fromCellId: lastFocusedCellId,
});
```

JS's lack of keyword arguments can make reading call sites challenging. I'd prefer to avoid ambiguous arguments. It probably be best to enforce with a lint rule, though, I suspect it would be noisy across the codebase, so just updated these helpers (seen in #5489 review).

### DIFF
--- a/frontend/src/components/chat/markdown-renderer.tsx
+++ b/frontend/src/components/chat/markdown-renderer.tsx
@@ -95,7 +95,11 @@ const CodeBlock = ({ code, language }: CodeBlockProps) => {
     const result = maybeTransform(language, value);
 
     if (language === "sql") {
-      maybeAddMarimoImport(autoInstantiate, createNewCell, lastFocusedCellId);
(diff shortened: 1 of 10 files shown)


Light2Dark (dismissed):

mscolnick (dismissed):

mscolnick:
@manzt definitely would be down for a lint rule. if it's noisy, we can make it a warning to "stop the bleeding". and as we see it in the codebase, we can do the TS quick-fix to refactor them.

mscolnick (approved):

PR #5500, opened 2025-07-01
Don't highlight keyword argument names as reactive references

manzt:
In `foo(a, b=b)`, the reactive reference analyzer was incorrectly
highlighting both instances of `b`:

```python
foo(a, b=b)
       ^ ^
```

The first `b` is a keyword argument name, not a variable reference.
After this fix, only actual variable references are highlighted:

```python
foo(a, b=b)
         ^
```

This change adds a check to skip keyword argument names by detecting
when a `VariableName` node is immediately followed by an `AssignOp` in
the AST.


akshayka (approved):
Nice catch

manzt:
Sorry, added an additional test with whitespacing.

PR #5507, opened 2025-07-01
fix: include all query params in WebSocket connection

harterrt-shopify:
# Fix: Include all query params in WebSocket connection

## Summary
- Fixed an issue where `mo.query_params()` was only returning the first query parameter
- The frontend now passes all query parameters from the browser URL to the WebSocket connection
- Previously, only special parameters like `session_id` and `file` were being passed through

## What this PR does

When visiting a marimo notebook with multiple query parameters like:
```
http://localhost:2718/?file=test.py&search=test&other=last
```

Previously, `mo.query_params()` would only return:
```python
{"file": "test.py"}
```

Now it correctly returns all parameters:
```python
{
  "file": "test.py",
  "search": "test", 
  "other": "last"
}
```

## How to test

1. Create a simple test notebook:
```python
import marimo as mo

@app.cell
def _(mo):
    params = mo.query_params()
    params
```

2. Run the notebook and open with query parameters:
```
http://localhost:2718/?file=test.py&search=test&other=last
```

3. Verify that all three parameters are displayed

## Design decisions

- Modified both `getWsURL()` and `getWsSyncURL()` for consistency
- Current page query params are merged with base URL params, with base URL params taking precedence
- Added a test case to verify the behavior

### DIFF
--- a/frontend/src/core/runtime/__tests__/createWsUrl.test.ts
+++ b/frontend/src/core/runtime/__tests__/createWsUrl.test.ts
@@ -38,8 +38,24 @@ describe("RuntimeManager.getWsURL", () => {
     const sessionId = "1234" as SessionId;
     const url = runtime.getWsURL(sessionId);
     expect(url.toString()).toBe(
-      "ws://marimo.app/nested/ws?foo=bar&session_id=1234&file=test.py",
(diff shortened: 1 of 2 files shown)


harterrt-shopify:
Not sure why the vercel test is failing, and I don't have access to investigate further. Seems like this is failing for other PRs as well, so I suspect it's upstream of this change.

PR #5509, opened 2025-07-02
fix: Invalid base url in api resolution

dmadisetti:
## 📝 Summary

I traced failing endpoints in WASM to malformed urls (e.g. `marimo.app/eapi`). Quick fix to prevent this, with a sanity check test to prevent regression

### DIFF
--- a/frontend/src/core/network/__tests__/api.test.ts
+++ b/frontend/src/core/network/__tests__/api.test.ts
@@ -1,0 +1,64 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { API } from "../api";
+
+// Mock fetch globally
+const mockFetch = vi.fn();
+global.fetch = mockFetch;
+
+// Default mock for getRuntimeManager
+let baseUrl = "http://localhost:8000";
+vi.mock("@/core/runtime/config", () => ({
+  getRuntimeManager: () => ({
+    get httpURL() { return new URL(baseUrl); },
+    headers: () => ({}),
+  }),
+}));
+
+describe("API", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+    baseUrl = "http://localhost:8000";
+  });
+
+  it("API.post calls fetch with POST and correct URL", async () => {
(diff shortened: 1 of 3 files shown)


frontend/src/core/network/api.ts
```diff
@@ -13,6 +13,10 @@ function getBaseUriWithoutQueryParams(): string {
   return url.toString();
 }
 
+function withTrailingSlash(url: string): string {
+  return url.endsWith("/") ? url : `${url}/`;
+}
```
Light2Dark: do you think this function should live in `frontend/src/utils/strings.ts`? Or `frontend/src/core/runtime/utils.ts`. I see related logic in both codepaths.

frontend/src/core/network/api.ts
```diff
@@ -26,7 +30,10 @@ export const API = {
       baseUrl?: string;
     } = {},
   ): Promise<RESP> {
-    const baseUrl = opts.baseUrl ?? getBaseUriWithoutQueryParams();
+    const baseUrl = withTrailingSlash(
+      opts.baseUrl ?? getBaseUriWithoutQueryParams(),
+    );
+    // Ensure the URL ends with a slash
```
Light2Dark: imo, remove the comment and put it as a docstring in the function.
lgtm otherwise!

Light2Dark (dismissed):

frontend/src/core/network/__tests__/api.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+    await API.get("/bar");
+    expect(mockFetch).toHaveBeenCalledWith(
+      "http://localhost:8000/api/bar",
+      expect.objectContaining({ method: "GET" }),
+    );
+  });
+
+  it("API.post handles base URL with path correctly", async () => {
+    baseUrl = "http://example.com/e";
```
manzt, line 54: Would we also want a sanity check with a trailing slash just to make sure we don't add an extra one?

manzt, line 54: Would we also want a test with a baseUrl that includes a trailing slash?

manzt (commented):
Looks great!

PR #5510, opened 2025-07-02
chore: Prevent noisy health calls in wasm situations

dmadisetti:
## 📝 Summary

WASM continually polls healthy- even though it's essentially a no op. This creates a noisy experience in quarto and some of our logs


manzt:
If this is user facing, maybe tag with `enhancement`? Internal won't show up in our release notes.

dmadisetti:
I still consider it more of a chore. Maybe we need a `fix` flag? Or is that conceding our product is broken :)

Issue #5219, opened 2025-06-06
Clarify and improve sandbox behavior for multiple scripts


manzt:
Follow up to https://github.com/marimo-team/marimo/pull/5216#discussion_r2132841892

### Describe the bug

Right now, `--sandbox` works well when launching a single script:

```sh
uvx marimo edit --sandbox foo.py
```

In that case, we read the PEP 723 metadata from `foo.py` and create a sandboxed environment accordingly.

But when launching with a directory:

```sh
uvx marimo edit --sandbox .
```

marimo allows switching between multiple notebooks in the UI. This is also possible in the `marimo edit --sandbox foo.py` case, picking a another notebook from the file tree. Each of these notebooks may have different (or no) PEP 723 metadata – but the same environment is reused across them, without resetting or isolating per-notebook state.

This leads to subtle and inconsistent behavior. A few issues I’ve run into:

### Case 1: Two notebooks, no PEP 723 metadata

* Open notebook A and import `depa` → marimo writes metadata for `depa`.
* Open notebook B and import `depb` → marimo writes metadata for `depb`.
* Later (in same session), import `depa` in notebook B → it works (since it's already in the environment), but no metadata is written.
* If I exit the session and launch notebook B directly, it fails. `depa` isn't in its metadata.

### Case 2: Overriding existing metadata

* Open notebook A with no dependencies.
* Switch to notebook B, which *does* have PEP 723 metadata for `depa`.
* marimo doesn't parse or install based on the existing metadata. Instead, it waits for a failed import and then overwrites the metadata.
* This can ignore pinned versions or overwrite valid metadata unintentionally.

These are just a couple of examples. I'm sure there are many other invalid states that can arise from reusing the same virtual environment across notebooks.

--- 

We should clarify what sandboxing means when multiple notebooks are involved. My preference would be to:

* Disallow directory targets entirely for `--sandbox` (e.g., raise an error if the target isn’t a script), or
* Disable opening additional notebooks in the frontend when in sandbox mode.

If we *do* want to support multiple notebooks in a single sandbox session, I think the correct behavior would be to reset and recreate the environment from scratch whenever switching to a new notebook so that each notebook is treated independently, as if it were launched with `--sandbox` on its own.

Until then, the current behavior feels inconsistent and hard to reason about.

### Will you submit a PR?

- [ ] Yes

### Environment

```
{
  "marimo": "0.13.15",
  "OS": "Darwin",
  "OS Version": "24.5.0",
[...]

Issue #2598, opened 2024-10-11
make sandbox compatibel with edit server (multiple notebooks)


simonmaris:
### Description

Currently --sandbox works great for single notebooks. 

However, when you start the edit server, two problems occur: 
- we have one isolated environment for all the notebooks
- the script metadata per PEP 723 is ignored on notebook launch, instead required packages are auto-discovered.

It would be lovely to make sandbox compatibel with this mode, since we can then really utilize the power of UV with it's global cache and users can switch projects easily without worrying about dependencies and environments. Really beginner friendly that way and no terminal needed to create / activate manually managed environments. 

### Suggested solution

Create some sort of logic that activates and deactivates a uv isolated environment on notebook launch and shutdown from the edit server.

gabrielgrant:
Oh, interesting... didn't realize this was an issue. So what is actually happening right now when I run `marimo edit --sandbox` is that packages for all notebooks are using the same env?

And it sounds like this would explain why I am being asked to re-install packages when reopening a notebook? 

FWIW, it does seem to be working-ish:

```
$ marimo edit --sandbox
Running in a sandbox: uv run --isolated --no-project --with-requirements /tmp/tmpz2c8w8sx.txt marimo edit


        Create or edit notebooks in your browser 📝

        URL: http://localhost:2718?access_token=lGtxh53-cAaCqe8l1yQzGA

Resolved 5 packages in 322ms
Prepared 3 packages in 106ms
Installed 3 packages in 20ms
 + charset-normalizer==3.4.0
 + requests==2.32.3
 + urllib3==2.2.3

Resolved 6 packages in 441ms
Prepared 6 packages in 2.62s
Installed 6 packages in 405ms
 + numpy==2.1.2
 + pandas==2.2.3
 + python-dateutil==2.9.0.post0
 + pytz==2024.2
 + six==1.16.0
 + tzdata==2024.2
```

mscolnick:
@gabrielgrant - it does re-install the packages, but since they are cached by `uv`, it should be pretty fast

sanzoghenzo:
I'm facing the same issue; while it is not a big problem having to reinstall the dependencies each time (as stated by @mscolnick, uv cache helps a lot here), it does break things when the import module doesn't have the same name as the package.

My user case: I use the `python-docx` package that is imported with `import docx`. The dependency is there

```
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "marimo",
#     "pandas==2.2.3",
[...]

PR #5511, opened 2025-07-02
enhancement: raise a UsageError on marimo edit --sandbox

akshayka:
Today, We (silently) don't support multi-notebook edit servers in sandbox mode. Running

marimo edit --sandbox

yields confusing behavior in which notebooks all share the same environment, even though their inline script metadata is kept separate. This isn't what the users want, and can lead to dependencies missing in notebook metadata.

Instead of failing silently, this change raises a UsageError if the user attempts to run the edit server in sandbox mode without a named notebook argument. It also links to the associated GitHub issue.

Here is the usage error:

```
Error: marimo's package sandbox requires a notebook name:

    * marimo edit --sandbox my_notebook.py

  Multi-notebook sandboxed servers (marimo edit --sandbox) are not supported.
  Follow this issue at: https://github.com/marimo-team/marimo/issues/2598.
```

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -405,13 +405,24 @@ def edit(
         )
         return
 
-    # Set default, if not provided
     if sandbox is None:
+        # When the sandbox flag is omitted we infer whether to
+        # to start in sandbox mode by examining the notebook file and
+        # prompting the user.
         from marimo._cli.sandbox import maybe_prompt_run_in_sandbox
 
         sandbox = maybe_prompt_run_in_sandbox(name)
+    elif sandbox and name is None:
+        raise click.UsageError(
+            """marimo's package sandbox requires a notebook name:
 
-    if sandbox:
+    * marimo edit --sandbox my_notebook.py
+
+  Multi-notebook sandboxed servers (marimo edit --sandbox) are not supported.
+  Follow this issue at: https://github.com/marimo-team/marimo/issues/2598."""
+        )
+
+    elif sandbox:
         from marimo._cli.sandbox import run_in_sandbox
 
         # TODO: consider adding recommended as well
--- a/tests/_cli/test_cli.py
+++ b/tests/_cli/test_cli.py
@@ -582,7 +582,6 @@ def test_cli_sandbox_edit_new_file() -> None:
 @pytest.mark.skipif(not HAS_UV, reason="uv is required for sandbox tests")
 def test_cli_sandbox_edit_not_supported() -> None:
     port = _get_port()
-    d = tempfile.TemporaryDirectory()
     p = subprocess.Popen(
         [
             "marimo",
--- a/tests/_cli/test_cli.py
+++ b/tests/_cli/test_cli.py
@@ -579,6 +579,25 @@ def test_cli_sandbox_edit_new_file() -> None:
     _check_contents(p, b"edit", contents)
 
 
+@pytest.mark.skipif(not HAS_UV, reason="uv is required for sandbox tests")
+def test_cli_sandbox_edit_not_supported() -> None:
+    port = _get_port()
+    d = tempfile.TemporaryDirectory()
+    p = subprocess.Popen(
+        [
+            "marimo",
+            "edit",
+            "-p",
+            str(port),
+            "--headless",
+            "--no-token",
+            "--sandbox",
+        ]
+    )
+    contents = _try_fetch(port)
+    _check_contents(p, b"not supported", contents)
+
+
 @pytest.mark.skipif(is_windows(), reason="Windows will prompt for Docker")
 def test_cli_edit_by_url() -> None:
     port = _get_port()


manzt:
I think related to 

- #5219

akshayka:
> I think related to
> 
> * [Clarify and improve sandbox behavior for multiple scripts #5219](https://github.com/marimo-team/marimo/issues/5219)

Oh yes, definitely. I forgot you raised that issue (thanks!).

gabrielgrant:
IIUC this breaking change would be a (significant, for me) loss of functionality -- this is how I've generally been using marimo since last year. AFAICT although the current behavior isn't ideal (packages need to be reinstalled when switching notebooks), it isn't completely broken. This `edit --sandbox`-with-no-specificed-notebook usage pattern is discussed here, with confirmation from @mscolnick that this isn't a totally crazy thing to do: https://github.com/marimo-team/marimo/issues/2598#issuecomment-2455458558

@akshayka Would you be open to making this a warning, rather than a fatal error? Or, at least having an escape hatch (like a `--force-edit-sandbox` flag) until the fundamental issue is resolved by [spawning each notebook in a separate subprocess](https://github.com/marimo-team/marimo/issues/2598#issuecomment-3024618914)

akshayka:
> IIUC this breaking change would be a (significant, for me) loss of functionality -- this is how I've generally been using marimo since last year. AFAICT although the current behavior isn't ideal (packages need to be reinstalled when switching notebooks), it isn't completely broken. This `edit --sandbox`-with-no-specificed-notebook usage pattern is discussed here, with confirmation from @mscolnick that this isn't a totally crazy thing to do: [#2598 (comment)](https://github.com/marimo-team/marimo/issues/2598#issuecomment-2455458558)
> 
> @akshayka Would you be open to making this a warning, rather than a fatal error? Or, at least having an escape hatch (like a `--force-edit-sandbox` flag) until the fundamental issue is resolved by [spawning each notebook in a separate subprocess](https://github.com/marimo-team/marimo/issues/2598#issuecomment-3024618914)

Hey @gabrielgrant, thanks for chiming in.

Could you instead use `uv` explicitly? I think this would do it (@manzt can let us know if my command isn't quite right):

```bash
uv run --isolated --with marimo marimo edit
```

If all you want is to hit `uv`s cache while using an isolated environment, I believe this does the trick.

`marimo edit --sandbox` actually is broken in a few ways, described here: https://github.com/marimo-team/marimo/issues/5219

gabrielgrant:
Ah, interesting, I had run into the first issue mentioned in #5219 (metadata not being written, which is not ideal, but not the end of the world: when opened in the future, marimo installs the missing package and adds to metadata). But hadn't come across the second issue (valid metadata being overwritten), and see how that one could potentially be more critical

When running `uv run --isolated --with marimo marimo edit` , would that still trigger header-defined deps to be installed? Or missing packages that marimo installs to be written into the header? (ie same behavior as `edit --sandbox` today) Or what is the overall workflow for handling notebook deps you're envisioning?

manzt:
> Ah, interesting, I had run into the first issue mentioned in #5219 (metadata not being written, which is not ideal, but not the end of the world: when opened in the future, marimo installs the missing package and adds to metadata). But hadn't come across the second issue (valid metadata being overwritten), and see how that one could potentially be more critical

I would unfortunately consider this (although usable) a very unfortunate incorrectness that I don't feel great about shipping. I'm in favor of raising a loud error because we haven't defined what multi-file mode is for `--sandbox`, and the fact that switching between files doesn't "sync" but rather just appends to your current environment creates awkward/undefined behavior.

Another thing that comes up is: say the notebook you switch to has a different version of package included in the inline script metadata? marimo will only prompt to install a new package if an import fails, so again, we silently allow the notebook to run with the wrong version. There are several other scenarios I can imagine that fall out from this mutation of the active environment.

I think the only correct way for us to have a multi-file sandbox mode (mentioned in the issue), would be to either entirely reset and "sync" the active environment to the current focused notebook, or support multiple isolated kernels. In marimo, we want our sandbox notebooks to be reproducible, and I would prefer not to have a mode that fundamentally breaks this guarantee.

Regarding the workaround @akshayka suggested:

```bash
uv run --isolated --with marimo marimo edit
```

This won't handle the PEP 723 metadata (the script headers) - it just gives you an isolated environment with marimo. You'd need to manually install dependencies as you go, and marimo won't be able to write metadata back to your notebooks.

However, uv supports supplying [`--with-requirements` from stdin](https://github.com/astral-sh/uv/pull/10447), meaning you could "bootstrap" your notebook session by composing with `uv export`:

```sh
uv export --script start_notebook.py | uv run --with-requirements=- --with=marimo marimo edit 
```

mscolnick:
This might break the marimo vscode extension as well, for some configuration settings.

akshayka:
> This might break the marimo vscode extension as well, for some configuration settings.

Thanks, we can look into this.

@gabrielgrant, does @manzt's elaboration make sense? Does the proposed workaround work for you, or would you still want to use `marimo edit --sandbox`, footguns notwithstanding?

gabrielgrant:
@akshayka : @manzt 's suggested `uv export --script start_notebook.py | uv run --with-requirements=- --with=marimo marimo edit` gets _closer_ to recreating the current useful functionality of `marimo edit --sandbox` (@manzt : thanks for the workaround and detailed explanations!). This seems like a safe option to suggest people use broadly. But, beyond being quite a "mouthful" to type, this still doesn't let me opt in to the occasionally-broken-but-generally-very-useful metadata writing.

Definitely understand the hesitation to continue shipping something that puts people at risk of these foot-guns. And so yes, agree that, given today's behavior, an error makes sense to interrupt people from doing metadata writing without understanding and explicitly opting in to the risks. But I do think it's worth considering that, if removed completely, the functionality would be missed, because in the vast majority of cases `marimo edit --sandbox` mostly works really well today.

Almost every notebook I've seen has a bunch of extremely-common shared deps (NumPy, scipy, Matplotlib, pandas, scikit‑learn, etc) where APIs are pretty stable and specific versions rarely matter. They may also import a few specific libs with more API churn, but those rarely have any impact on other notebooks -- the odds of meaningful dep version conflicts between notebooks being edited within the same session seems very (very) small. And in these happy-path use-cases (well over 90% IME), `marimo edit --sandbox` is great (occasional need to accept re-install of a dep on startup that didn't get written out in a previous session notwithstanding).

I totally, _totally_ get the desire to not have confusing, undefined edge-case behavior, especially with dependency management & reproducibility. As an engineer (esp one who has dedicated a _lot_ of time to tooling in these exact areas as an early team member at [Docker](https://www.docker.com/) and [Pachyderm](https://docs.pachyderm.com/products/mldm/latest/learn/key-features/)), it makes my skin crawl. But, putting on my pragmatic product person hat, I'd say that, while marimo can be used in lots of contexts where absolute correctness is critical, its _core_ use-case (AFAIU) is still as a Jupyter alternative for exploratory analysis. And a set of features that make life significantly easier in 90%+ of these core use-cases is a huge win.

Maybe there's a good alternative to dep management in multi-notebook edit mode that I'm unaware of, but it seems the main one is just running it all in one venv, which has basically the same cross-contamination downsides, but is just more annoying. To keep things separate, expecting users to manually manage the metadata header (which afaict would be required for @manzt 's workaround to continue working with new/edited notebooks?) seems like a non-starter.

So wondering if there's a happy middle path? Ideally there are some easy-ish ways to mitigate the footgun[1] (perhaps even enough to make it a safe default). But even without mitigation, and knowing the risks, I'd personally still choose to continue using `marimo edit --sandbox` with today's behavior. Would you be open to allowing this potentially-risky (but generally very useful) behavior to be explicitly opted in to (eg `marimo edit --dangerously-broken-sandbox`)?

[1]: Specific footgun mitigation ideas, in rough increasing order of implementation complexity & user-experience benefit:
-  loud & frequent reminders that if you're coming across an occasional head-scratcher, broken sandbox behavior may be the cause
- checking requested versions against what's in the kernel on notebook load, and throwing an in-notebook warning message if incompatibilities are found
- unloading notebook-specific deps when switching from one sandboxed notebook to another (eg with `sys.modules.pop`, tho that still isn't perfect: https://stackoverflow.com/a/57891909 )

Apologies for the essay, thanks for reading (if you got through it) and will certainly respect whatever decision you make -- you folks are doing amazing work with marimo, that I hope/believe is the future how most Python code gets written ❤️

akshayka:
@gabrielgrant, thanks so much for the very thoughtful feedback. I truly appreciate it!

Taking your feedback into consideration, we won't remove `--sandbox` outright for multi-notebook edit servers, or at least we won't remove its functionality. We may instead adopt your suggestion of  adding a "dangerous" flag, such as

```
marimo edit --dangerous-sandbox
```

and in any case we'll make sure to ping you before merging (it'll be in this PR).

Thanks again for chiming in!

marimo/_cli/cli.py
```diff
@@ -159,8 +159,10 @@ def _resolve_token(
 )
 
 sandbox_message = (
-    "Run the command in an isolated virtual environment using "
-    "`uv run --isolated`. Requires `uv`."
+    "Run the notebook in an isolated virtual environment, tracking "
+    "installed dependencies in the notebook file as inline script metadata. If "
+    "the notebook file already has tracked dependencies, automatically install "
+    "them into the environment. Requires uv."
```
manzt: Nit, maybe mention PEP 723 specifically (so one could easily lookup "inline metadata")? E.g., 

"Run the notebook in an isolated environment, with dependencies tracked via PEP 723 inline metadata. If already declared, dependencies will install automatically. Requires uv."

manzt (dismissed):
Looks great. ~I recommend a follow up PR for `--dangerous-sandbox` before the next release.~

Whoops, missed that we want to add to this PR :) just ping me when ready.

marimo/_cli/cli.py
```diff
@@ -377,6 +389,7 @@ def edit(
     allow_origins: Optional[tuple[str, ...]],
     skip_update_check: bool,
     sandbox: Optional[bool],
```
mscolnick, line 393: could make `sandbox` hidden

marimo/_cli/cli.py
```diff
@@ -336,6 +337,17 @@ def _get_stdin_contents() -> str | None:
...
 )
+@click.option(
+    "--dangerous-sandbox/--no-dangerous-sandbox",
+    is_flag=True,
+    default=None,
+    show_default=False,
+    type=bool,
+    hidden=True,
+    help="""Enables the usage of package sandboxing when running a multi-edit
+notebook server; this behavior is dangerous, for reasons described in
```
mscolnick: "this behavior is dangerous". maybe `this behavior is misleading`. not sure if its actually "dangerous"

marimo/_cli/cli.py
akshayka: Done!

marimo/_cli/cli.py
akshayka: I replaced dangerous with a more specific statement, thanks

marimo/_cli/cli.py
akshayka: The `edit` command is shared for multi-notebook servers and for named edit sessions (`marimo edit foo.py`), so we can't hide `sandbox`.

manzt (approved):
lgtm

PR #5515, opened 2025-07-02
fix: reuse ports that are available but recently closed

akshayka:
This fixes find free port to actually use addr/ports that are available but whose sockets were recently closed. The fix is based on logic from Tornado and Jupyter.

Previously,

* `marimo new` -> port 2718
* shut down marimo
* `marimo new` -> port 2719
* shut down marimo
* `marimo new` -> port 2720
* ...

Now, all instances of `marimo new` use 2718.

This fix also parametrizes find_free_port with an address. Users can still search on any interface if they start marimo with --host 0.0.0.0, otherwise our CLI will default to localhost (as it used to do until recently).

**Tested**: In manual testing, this appears to do the right thing. Unfortunately a robust CLI test will likely take some time.

### DIFF
--- a/marimo/_server/utils.py
+++ b/marimo/_server/utils.py
@@ -2,7 +2,9 @@
 from __future__ import annotations
 
 import asyncio
+import errno
 import os
+import socket
 import sys
 from typing import TYPE_CHECKING, Any, TypeVar
 
@@ -11,6 +13,7 @@ from marimo._utils.marimo_path import MarimoPath
 
 if TYPE_CHECKING:
     from collections.abc import Coroutine
+    from typing import Optional
 
 # use spaces instead of a tab to play well with carriage returns;
 # \r\t doesn't appear to overwrite characters at the start of a line,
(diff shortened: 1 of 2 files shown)


marimo/_server/utils.py
```diff
@@ -31,32 +34,199 @@ def canonicalize_filename(filename: str) -> str:
...
 
 
-def find_free_port(port: int, attempts: int = 100) -> int:
-    """Find a free port or move to the next one recursively"""
+_DEFAULT_BACKLOG = 128
 
-    import socket
+
+# From Tornado (Apache 2.0)
+# https://github.com/tornadoweb/tornado/blob/8a953697888d463f48090d500268892a7384e6b1/tornado/netutil.py#L56
```
dmadisetti, line 41: Worth refactoring into a util file? Now this is over half the file

dmadisetti (commented):
Probably worth splitting up, but LGTM

marimo/_server/utils.py
akshayka: I'm generally not a fan of creating tons of files. This file is still only 290 lines long.

PR #5518, opened 2025-07-02
feat: statically served page as landing for callback

dmadisetti:
## 📝 Summary

Provides a generally vanilla landing page to set localhost values on general callbacks. `/static/callback.html?__values=stored&__if=prefixed&not=otherwise`

May be subject to change, do not explicitly rely on this feature outside of `moutils` unless you have a working relationship with the marimo-team @Talador12

### DIFF
--- a/frontend/public/files/callback.html
+++ b/frontend/public/files/callback.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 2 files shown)


dmadisetti:
Or we can just support this on `marimo.app` and dashboard. Which might be better, because we an do styling etc. etc.

dmadisetti:
Yeah, I think it's best not to touch FOSS. Will expund in other issue

PR #5477, opened 2025-06-27
Automate release notes generation using Git and GitHub metadata

manzt:
A Python script automates the creation of structured release notes by:

- Fetching all commits to `main` since the last tag
- Extracting PR numbers from (squashed PR) commit messages
- Batch-fetching recent PR details (title, labels, body) from GitHub
- Categorizing PRs using labels: `enhancement`, `preview`, `bug`, and other (e.g., `internal`)
- Identifying top-level highlights using:
  - `release-highlight` (important user-facing features)
  - `includes-media` (PRs with visual content)
- Formatting the categories and sections into markdown

## Usage

```bash
uv run scripts/generate_release_notes.py <since-tag>
# Example: python generate_release_notes.py 0.14.7
```

PR #5534, opened 2025-07-04
chore: add pr title to infer which category the PR falls in

Light2Dark:
## 📝 Summary

Add PR title in the inference of the category in release notes, so imo, making the process more automatic. 

Comparison from running `uv run scripts/generate_release_notes.py 0.14.7`

CleanShot 2025-07-05 at 02 06 37@2x

## 📋 Checklist

## 📜 Reviewers

@manzt

### DIFF
--- a/scripts/generate_release_notes.py
+++ b/scripts/generate_release_notes.py
@@ -173,8 +173,8 @@ def categorize_entries(
         pr: PullRequest = entry.pr
         label_names = {label.name for label in pr.labels}
 
-        # Skip entries labeled as "internal"
-        if "internal" in label_names or pr.title.startswith("chore"):
+        # Internal PRs can be skipped to focus on public-facing changes
+        if "internal" in label_names or pr.title.startswith(("chore")):
             continue
 
         if "release-highlight" in label_names or pr.title.startswith("feat"):
--- a/scripts/generate_release_notes.py
+++ b/scripts/generate_release_notes.py
@@ -170,22 +170,23 @@ def categorize_entries(
             categories["other"].append(entry)
             continue
 
-        label_names = {label.name for label in entry.pr.labels}
+        pr: PullRequest = entry.pr
+        label_names = {label.name for label in pr.labels}
 
         # Skip entries labeled as "internal"
-        if "internal" in label_names:
+        if "internal" in label_names or pr.title.startswith("chore"):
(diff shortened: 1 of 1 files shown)


Light2Dark:
Or maybe we could force PRs to have labels: https://github.com/marketplace/actions/enforce-pr-labels

manzt (commented):
Thanks for flagging my review! I'd defer to @akshayka / @mscolnick for a final call, but my preference is that we stick with either labels _or_ conventional commits — not both. Whichever direction we choose, it should be [clearly documented](https://github.com/marimo-team/marimo/pull/5477#issuecomment-3013871610) how preparing a PR affects releases.

For context: I intentionally made this script to rely only on labels so we could decouple the release semantics (e.g., bug fix, feature, preview) from the PR title. I know adding a label is one extra step (a click in the GitHub UI or at the end of `gh pr create`) but I hope it's minimal.

On a personal note, I’ve used conventional commits before, but I’ve come to prefer GitHub labels after "discovering" it as an alternative. They give us more flexibility in when we define the semantics of a change — whether at PR creation, during iteration on the PR, or, and most importantly, even after merging. Ideally, we’d know the right categorization upfront, but in practice, I've found it so helpful to have that flexibility. It also gives release managers the ability to review and reclassify changes based on how they should be presented to users. With conventional commits, those semantics are baked into git history, making them harder to adjust later.

My main concern with using both labels and conventional commits is the added complexity from their interaction. Which one takes priority? For example, with the proposed changes in this PR, any future PR prefixed with "chore:" will be omitted from the release notes entirely, even if we later decide it should be labeled as a bug fix or enhancement.

akshayka:
Thanks @light2dark for the PR and @manzt for the thoughtful writeup.

Let's continue to use labels exclusively. I have found the process to work well.

manzt:
> Or maybe we could force PRs to have labels: https://github.com/marketplace/actions/enforce-pr-labels

We didn't require labels previously, so I’m not too concerned if one gets missed (we can always add after a PR has been merged and before releasing). Unlabeled PRs still show up in the “Other” section of the release notes, so nothing gets silently dropped. That section acts as a fallback, similar to how things worked before we started categorizing PRs. Only PRs explicitly labeled as “internal” are excluded entirely.

PR #5544, opened 2025-07-05
enhancement: don't auto scroll when sending cell to top/bottom from menu

akshayka:
The user is likely just trying to get the cell out of the way, especially when moving it to the bottom. So don't scroll to the bottom.

Scrolling still might still sense if using a hotkey since the user may type it accidentally (and then wonder where the cell went); or, when using a hoteky, the user is more likely to be actively editing the cell.

This is just a small usability change, but something I noticed while working on slides.


manzt (approved):

manzt:
I'm not entirely sure I grasp the exact context, but the changes seem minimal and straightforward. I trust your judgment on the usability aspect.

PR #5560, opened 2025-07-07
chore: add js docs to useAsyncData

mscolnick:
Add some documentation to `useAsyncData`. I created a base interface so the docs get carried over to the specific result case

### DIFF
--- a/frontend/src/hooks/useAsyncData.ts
+++ b/frontend/src/hooks/useAsyncData.ts
@@ -14,7 +14,7 @@ import { invariant } from "@/utils/invariant";
  *
  * @template T - The type of data being fetched
  */
-export interface AsyncBaseResult<T> {
+interface AsyncBaseResult<T> {
   /**
    * The current status of the async operation.
    * - `pending`: Initial state, no data has been fetched yet
--- a/frontend/src/hooks/useAsyncData.ts
+++ b/frontend/src/hooks/useAsyncData.ts
@@ -9,7 +9,54 @@ import {
 import useEvent from "react-use-event-hook";
 import { invariant } from "@/utils/invariant";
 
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/useAsyncData.ts
```diff
@@ -9,31 +9,87 @@ import {
...
+   * - `true` when actively fetching data (pending or loading states)
+   * - `false` when not fetching (success or error states)
+   */
+  isFetching: boolean;
+}
+
+/**
+ * Represents the initial state when no data has been fetched yet.
+ */
+interface PendingResult<T> extends AsyncBaseResult<T> {
```
manzt, line 59: Either is fine, but since the generic param isn't used we could just use `never` I think:

```suggestion
interface PendingResult extends AsyncBaseResult<never> {
```

frontend/src/hooks/useAsyncData.ts
```diff
@@ -9,31 +9,87 @@ import {
 import useEvent from "react-use-event-hook";
 import { invariant } from "@/utils/invariant";
 
-interface PendingResult {
+/**
+ * Base result interface containing common properties for all async data states.
+ *
+ * @template T - The type of data being fetched
+ */
+export interface AsyncBaseResult<T> {
```
manzt: This is ideal for documentation, but I think we should avoid exporting and encourage the use of only `AsyncDataResult<T>` outside of this module.

```suggestion
interface AsyncBaseResult<T> {
```

manzt (approved):
Sweet, awesome to have this so nicely documented. I thought that inheriting from `AsyncBaseResult` would break the ability to discriminate between the different states but I checked it out and it still works.

PR #5565, opened 2025-07-07
fix: allow select on table cells

Light2Dark:
## 📝 Summary

https://github.com/user-attachments/assets/92337c8d-1250-4bc1-92b4-e3291f3fae5b

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/renderers.tsx
+++ b/frontend/src/components/data-table/renderers.tsx
@@ -11,6 +11,7 @@ import {
   type Table,
   type Table as TanStackTable,
 } from "@tanstack/react-table";
+import { useAtomValue } from "jotai";
 import type { JSX } from "react";
 import {
   TableBody,
@@ -20,6 +21,7 @@ import {
   TableRow,
 } from "@/components/ui/table";
 import { cn } from "@/utils/cn";
+import { isSelectingMultipleCellsAtom } from "./range-focus/atoms";
 import { CellRangeSelectionIndicator } from "./range-focus/cell-selection-indicator";
 import { useCellRangeSelection } from "./range-focus/use-cell-range-selection";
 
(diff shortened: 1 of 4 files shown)


mscolnick:
i think this is a lesser of two-evils option for now. 

something we can consider is if the current drag selects more than 1 cells (i.e. once its more than 1), clear the selection:

```ts
// from https://stackoverflow.com/questions/3169786/clear-text-selection-with-javascript
if (window.getSelection) {
  if (window.getSelection().empty) {  // Chrome
    window.getSelection().empty();
  } else if (window.getSelection().removeAllRanges) {  // Firefox
    window.getSelection().removeAllRanges();
  }
}
```

Light2Dark:
Thanks, I added the atom to check the state and func to clear the selection. I think its a good middle ground (perf and UX)

frontend/src/components/data-table/renderers.tsx
```diff
@@ -88,6 +90,12 @@ export const DataTableBody = <TData,>({
     handleCellsKeyDown,
   } = useCellRangeSelection({ table });
 
+  const isSelectingMultipleCells = useAtomValue(isSelectingMultipleCellsAtom);
+
+  if (isSelectingMultipleCells && window.getSelection) {
+    window.getSelection()?.empty();
```
mscolnick: can we put this in the reducer? this may mean if you select somewhere else in the page and this re-renders for whatever reason, then the selection is cleared.

it would be better to do this in the reducer once we select more.

this could be in each action, or you can potentially use the reducer middleware

mscolnick (approved):

frontend/src/components/data-table/renderers.tsx
Light2Dark: that's awesome, didn't think it would be possible, but since the window is global, it makes sense.

PR #5567, opened 2025-07-07
tests: get playwright running again

mscolnick:
Refactors and configuration improvements to playright to get it passing again.

* Also skips the flakey e2e test, but adds a new one that tests the same functionality

### DIFF
--- a/frontend/.eslintrc.cjs
+++ b/frontend/.eslintrc.cjs
@@ -168,6 +168,7 @@ module.exports = {
         "@typescript-eslint/require-await": "off",
         "@typescript-eslint/no-unsafe-member-access": "off",
         "@typescript-eslint/no-unsafe-return": "off",
+        "no-console": "off",
       },
     },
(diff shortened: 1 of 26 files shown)


frontend/.eslintrc.cjs
```diff
@@ -168,6 +168,7 @@ module.exports = {
         "@typescript-eslint/require-await": "off",
         "@typescript-eslint/no-unsafe-member-access": "off",
         "@typescript-eslint/no-unsafe-return": "off",
+        "no-console": "off",
```
manzt, line 171: I notice we have disabled in the individual tests. Maybe can remove here?

```suggestion
```

manzt (approved):

frontend/.eslintrc.cjs
mscolnick: console should be fine / helpful in e2e tests anyways

PR #5571, opened 2025-07-08
internal: request utils, make remote file send marimo user-agent

mscolnick:
Add a few request utils for a better API for fetching (similar to the `requests` lib).
Pass `marimo/__version__` as the user agent when fetching remote notebooks

### DIFF
--- a/marimo/_utils/requests.py
+++ b/marimo/_utils/requests.py
@@ -51,14 +51,19 @@ class Response:
         """
         return self.content.decode("utf-8")
 
-    def raise_for_status(self) -> None:
-        """Raise an exception for non-2xx status codes."""
+    def raise_for_status(self) -> "Response":
+        """Raise an exception for non-2xx status codes.
+
+        Returns:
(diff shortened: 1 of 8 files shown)


marimo/_utils/requests.py
```diff
@@ -0,0 +1,185 @@
...
+        self.content = content
+        self.headers = headers
+
+    def json(self) -> Any:
+        """Parse response content as JSON."""
+        return json.loads(self.text())
+
+    def text(self) -> str:
+        """Get response content as text."""
+        return self.content.decode("utf-8")
```
akshayka: Requests has logic to infer the encoding. Users can also explicitly set the encoding.

I guess we don't need that today. But I would add in a docstring to `text` and `json` that `Response` assumes `utf-8` encoded bytes.

mscolnick: thanks added

akshayka (approved):
LGTM if tests are green

PR #5580, opened 2025-07-08
fix: show additional loading indicator after 5 seconds

mscolnick:


dmadisetti:
A little unrelated- but I can follow up. We need an isHealthy atom, to show the runtime is broken and replace  if we don't connect

We currently only wait a minimum of 6s on the healthy side:
https://github.com/marimo-team/marimo/blob/1daf0b70b3282c48e18136ca3e023f77e1d644bc/frontend/src/core/runtime/runtime.ts#L180-L181

So 5s delay might be a bit long

mscolnick:
@dmadisetti 

> We currently only wait a minimum of 6s on the healthy side:

That is incorrect, it does exponential backoff. We may want a different backoff strategy because we could be waiting when it is actually healthy

akshayka:
Oh I saw that the alert only comes after 5s, that's fine

PR #5586, opened 2025-07-08
fix: patch matplotlib interactive to not steal focus on mount

mscolnick:


marimo/_plugins/stateless/mpl/_mpl.py
```diff
@@ -509,3 +509,17 @@ def dispose(self, context: RuntimeContext, deletion: bool) -> bool:
     filter: invert(0.3);
 }
 """.strip()
+
+
+def patch_javascript(javascript: str) -> str:
```
manzt, line 514: I'm guessing we are doing this because we don't want to vendor this file into marimo?

tests/_plugins/stateless/test_mpl.py
```diff
@@ -64,3 +64,19 @@ async def test_mpl_show(k: Kernel, exec_req: ExecReqProvider) -> None:
...
+def test_patch_javascript() -> None:
+    from matplotlib.backends.backend_webagg_core import FigureManagerWebAgg
+
+    from marimo._plugins.stateless.mpl._mpl import patch_javascript
+
+    javascript: str = str(FigureManagerWebAgg.get_javascript())  # type: ignore[no-untyped-call]
+    assert javascript is not None
+    javascript = patch_javascript(javascript)
+    assert javascript.count("// canvas.focus();") == 1
+    assert javascript.count("// canvas_div.focus();") == 1
```
manzt, line 82: I _think_ this is farily robust to ensure the script doesn't change out from underneath us. Would you want a find+replace with more context to ensure we make the change in one spot?

mscolnick: i think this is enough. not sure if adding more context makes it more fragile or less fragile. this just removes focus altogether

marimo/_plugins/stateless/mpl/_mpl.py
mscolnick: id rather not vendor it since it may forgo other important updates made to the upstream lib

tests/_plugins/stateless/test_mpl.py
manzt: Sure, I also don't know if more or less fragile so sticking with it makes sense.

PR #5592, opened 2025-07-09
fix: autofocus generate with AI

mscolnick:


Light2Dark (dismissed):
made code improvement in #5597, but otherwise this looks good

PR #5563, opened 2025-07-07
Make reactivity options in footer responsive for small screen widths

manzt:
Fixes #5495

Fixes bottom toolbar overflow issue on small displays by implementing
responsive design that groups runtime settings into a settings dropdown.
When screen width is below `md` (768px), the footer shows a settings icon
instead of individual runtime controls, preventing text overlap.

https://github.com/user-attachments/assets/805819de-a8b8-4d9e-8ef0-b6c26a72961b

PR #5593, opened 2025-07-09
style: fix overflow on footer styles

mscolnick:


manzt (approved):
I'm _thinking_ this avoids the awkward horizontal scrollbar when resizing the window (seen briefly in video in #5563)?

mscolnick:
@manzt  its actually the vertical scroll bar which is also in the video

PR #5595, opened 2025-07-09
improvement: scroll table with keyboard navigation

mscolnick:


mscolnick:
@Light2Dark i did try smooth. didn't like it as much, but i didn't hate it either. ill defer to you.

PR #5607, opened 2025-07-10
docs: Update cell docstring to better express current abillities

### LINKED ISSUE #5336: Running named cells that use @persistent_cache or query_params() outside notebooks
### Documentation is

- [x] Missing
- [ ] Outdated
- [x] Confusing
- [ ] Not sure?

### Explain in Detail

If I have an app and a named cell within it, the [current `Cell` doc](https://docs.marimo.io/api/cell/#marimo.Cell) says:

> Cells can be named via the marimo editor in the browser, or by changing the cell's function name in the notebook file. Named cells can then be executed for use in other notebooks, or to test in unit tests.
> 
> For example:
>
> ```py
> from my_notebook import my_cell
> 
> output, definitions = my_cell.run()
> ```

The phrase "or to test in unit tests" doesn't mention that those unit tests must apparently also be [marimo-embedded](https://docs.marimo.io/guides/testing/).

I figured out the way to truly run a Marimo cell outside of any notebook, which was:
```py
from my_notebook import app, my_cell
from marimo._ast.app import InternalApp
from marimo._runtime.context.script_context import initialize_script_context
from marimo._messaging.types import NoopStream
initialize_script_context(app=InternalApp(app), stream=NoopStream(), filename=__file__)
my_cell.run(...)
```

This should either be documented, or better wrapped into a simpler function or construct, or cells should figure out themselves whether they are in a context of a notebook and be able to run outside of notebooks.

### Will you submit a PR?

- [ ] Yes

dmadisetti:
## 📝 Summary

fixes #5336 mostly by documentation

We do not mention that calling `my_cell()` directly is possible. Also the language around testing was a little confusing

### DIFF
--- a/marimo/_ast/cell.py
+++ b/marimo/_ast/cell.py
@@ -393,14 +393,31 @@ class CellImpl:
 class Cell:
     """An executable notebook cell
 
-    A `Cell` object can be executed as a function via its `run()` method, which
-    returns the cell's last expression (output) and a mapping from its defined
-    names to its values.
+    Cells are the fundamental unit of execution in marimo. They represent
+    a single unit of execution, which can be run independently and reused
+    across notebooks.
 
-    Cells can be named via the marimo editor in the browser, or by
-    changing the cell's function name in the notebook file. Named
-    cells can then be executed for use in other notebooks, or to test
-    in unit tests.
+    Cells are defined using the `@app.cell` decorator, which registers the
+    function as a cell in marimo.
+
+    For example:
+
+    ```python
+    @app.cell
+    def my_cell(mo, x, y):
+        z = x + y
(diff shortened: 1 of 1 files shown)


akshayka (dismissed):
Looks great, thank you!

marimo/_ast/cell.py
```diff
@@ -393,14 +393,31 @@ def output(self) -> Any:
...
+
+    ```python
+    @app.cell
+    def my_cell(mo, x, y):
+        z = x + y
+        mo.md(f"The value of z is {z}")  # This will output markdown
+        return (z,)
+    ```
+
+    Cells can invoked as functions, and picked up by test frameworks like
```
dmadisetti:
```suggestion
    Cells can be invoked as functions, and picked up by test frameworks like
```

PR #5613, opened 2025-07-11
fix: variable kwargs did not invalidate cache

### LINKED ISSUE #5310: persistent_cache() bug: doesn't capture **kwargs into the cache key properly
### Describe the bug

I have a persistently-cached function with kwargs, the parameters are not put into the cache key properly:

```py
@app.cell
def _(mo):
    @mo.persistent_cache
    def my_cached_func(*args, **kwargs):
        # do something
    
    kw = {"some_kw_arg": 1}
    # Expect (persistent) caching correctly capture changes in kw, but it doesn't.
    res = my_cached_func(some_arg1, **kw)
```

The bug manifest when my_cached_func's args are completely expanded, so this bug should definitely relate to kwargs handling.

The bug is most likely in these lines:

https://github.com/marimo-team/marimo/blob/dae55fd52056d271738cdaef4a0fe5d9eebab811/marimo/_save/hash.py#L276-L282

Or, in some closely preceding code on the `content_cache_attempt_from_base() -> get_and_update_context_from_scope()` call path, or in the subsequent call to `hasher.extract_ref_state_and_normalize_scope()`.

FYI @dmadisetti 

### Will you submit a PR?

- [ ] Yes

### Environment

{
  "marimo": "0.13.15",
  "OS": "Darwin",
  "OS Version": "21.6.0",
  "Processor": "i386",
  "Python Version": "3.12.7",
  "Binaries": {
    "Browser": "137.0.7151.104",
    "Node": "v20.18.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8",
    "narwhals": "1.41.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "starlette": "0.46.2",
    "tomlkit": "0.13.3",
    "uvicorn": "0.34.3",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "openai": "1.76.2",
    "pandas": "2.2.3",
    "pyarrow": "18.1.0",
    "loro": "1.5.1",
    "ruff": "0.11.7"
  },
  "Experimental Flags": {}
}

### Code to reproduce

_No response_

dmadisetti:
## 📝 Summary

fixes #5310 

variable `**kwargs` are captured as a positional argument by inspect, leading to incorrect scope creation for hashing- leading to false postive cache hits.

This PR explicitly sets the `**kwarg` positional variable, and adds further tests to enumerate other potential function signatures.

### DIFF
--- a/marimo/_save/save.py
+++ b/marimo/_save/save.py
@@ -128,8 +128,6 @@ class _cache_call:
                 inspect.Parameter.POSITIONAL_ONLY,
             )
         ]
-        # Both varargs and kwargs are are lumped under co_flags;
-        # So extract them if they exist, and handle them explicitly.
         self._var_arg = next(
             (
                 param.name


marimo/_save/save.py
```diff
@@ -116,7 +118,15 @@ def _set_context(self, fn: Callable[..., Any]) -> None:
...
         self.fn = fn
-        self._args = list(self.fn.__code__.co_varnames)
+        code = fn.__code__
+        has_var_args = bool(code.co_flags & inspect.CO_VARARGS)
+        has_var_kwargs = bool(code.co_flags & inspect.CO_VARKEYWORDS)
+        self._args = list(code.co_varnames)
+        if has_var_kwargs:
+            self._var_kwarg = self._args.pop()
+        if has_var_args:
+            self._var_arg = self._args.pop()
```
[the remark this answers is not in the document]
dmadisetti: Not really more clear imo

marimo/_save/save.py
```diff
@@ -116,7 +118,19 @@ def _set_context(self, fn: Callable[..., Any]) -> None:
         )
 
         self.fn = fn
-        self._args = list(self.fn.__code__.co_varnames)
+        code = fn.__code__
```
akshayka: This might fail in a couple ways:

* Methods don't have a `__code__` object. But I believe you can retrieve the code from `method.__func__.__code__`.
* It might make `@mo.cache` not composable, which is okay

Does caching support methods, by the way? It fails with a non-graceful error right now. If method support is not planned we can (in a future PR) raise a better error.

image

If we decide not to support methods, and to make mo.cache not composable, let's document that somewhere.

tests/_save/test_cache.py
```diff
@@ -859,6 +859,394 @@ def fib(n):
         assert k.globals["a"] == 5
         assert k.globals["b"] == 55
 
+    async def test_cache_decorator_with_kwargs(
```
akshayka, line 862: Should we add a test with `persistent_cache` too, since that's what the original issue was about?

marimo/_save/save.py
dmadisetti: Methods are not currently handled.

I think we ideally want drop in replacement for functools, so good comments.

I can tackle these in an additional PR or tack them on

tests/_save/test_cache.py
dmadisetti: Observed in `persistent_cache`, but the code path is shared and reproducible in `cache`. Checking persistent cache explicitly just adds more boilerplate for file cleanup and/or injecting the memory loader in, but I'll add one for a sanity check

marimo/_save/save.py
akshayka: Another PR sounds good, thanks!

marimo/_save/save.py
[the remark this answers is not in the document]
akshayka: @dmadisetti inspect.signature would I think work for both methods and callable objects, whereas using `__code__` doesn't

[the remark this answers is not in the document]
dmadisetti: Good call. Replaced

marimo/_save/save.py
```diff
@@ -116,7 +118,35 @@ def _set_context(self, fn: Callable[..., Any]) -> None:
...
+            param.name
+            for param in sig.parameters.values()
+            if param.kind
+            in (
+                inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                inspect.Parameter.POSITIONAL_ONLY,
+            )
+        ]
+        # Both varargs and kwargs are are lumped under co_flags;
+        # So extract them if they exist, and handle them explicitly.
```
akshayka: Remove stale comment
```suggestion
```

akshayka (dismissed):
LGTM with one small comment

PR #5614, opened 2025-07-12
improvement: better focus navigation between cell and cell-editor

mscolnick:
This refactors cell navigation with hotkeys. Now we have 2 levels of focus: cell and cell-editor. 

- `Escape` to move from cell-editor to cell
- `Enter` to move from cell to cell-editor
- up/down arrows work th esame
- refactor of hidden cell code (and some bug fixes), so feels more natural entering and leaving editors that are temporarily hidden


akshayka:
@manzt, when you take a look at this, here are some things I suggest you consider and would appreciate your opinions on:

* The primary question we should have a discussion on and informed opinion on: How does modal editing feel to the non-expert user, and in particular should it be the default or not? As far as I understand [modern HCI has made a push for modeless editing](https://worrydream.com/refs/Tesler_2012_-_A_Personal_History_of_Modeless_Text_Editing_and_Cut-Copy-Paste.pdf). I like modal editing (vim) but I consider that to be an "advanced" kind of editing and understand that most people don't like vim keybindings. Are there compelling reasons that the default experience for a notebook should be different, or could these same things be achieved by simpler keyboard shortcuts?
* When clicking around the page and cells, is it intuitive what mode the user will enter into?
* Secondary (if we think modal editing should be a default), is the visual indicator for what mode you are in clear, and also importantly does it fit our product's design language?

@manzt after you do a review pass and share your thoughts, I will also take a look. Thanks! (Edit: Couldn't help but try it, and it feels nice and intuitive to me personally. I think the styling needs updating but the experience felt mostly intuitive to me. Still curious for your opinion.)

manzt:
Looking forward to reviewing and giving this a try.

mscolnick:
Fwiw, we've mostly had these two levels of navigation, there were some bugs and uncompleted implementation that made it not as apparent. Its still possible to click around without even knowing you can navigate at the cell level 

As far as command mode hotkeys (not implemented yet), i was thinking of making that opt in via config

akshayka (dismissed):
This feels really great! Navigation feels much better, and opening hidden cells is also much nicer :) I just slightly updated the focus style and I'm happy with that too now.

I didn't review the code (I'll let others do that), but from UI/UX perspective LGTM!

manzt:
sorry wasn't quick enough to review. looking forward to test driving :)

freethebee:
Is there an option to switch this behaviour off, I searched around, but didn't find one? I really enjoy Marimo and always use the vim bindings for editing, but this extra layer of normal mode tends to get in the away a bit. As a vim user my instinct is to always end in normal mode, but now that there are two, this leads to unintended actions, for example, I hit gg to go to the top of a cell and find myself at the beginning of the document instead.
I understand for many it is a very nice enhancement, but a toggle would be great.

mscolnick:
@freethebee, currently it is not possible to turn this off. if you would like to see this added though, please file a feature request so we can prioritize or discuss workarounds. 

just to be clear (and feel free to answer in the issue), you don't want any keyboard navigation for moving between the cells? do you use your mouse exclusively for this?

PR #5616, opened 2025-07-12
Fix handling of class attribute references for reactive variables

manzt:
Improves how class-scoped assignments are resolved when analyzing
reactive references. Previously, class attributes like `bar = 10` could
be incorrectly highlighted as reactive if a global `bar` existed, due to
improper scoping resolution. The fix ensures that class-level variables
are interpreted in declaration order, mirroring Python's evaluation
semantics. This avoids false positives in cases involving
self-references or forward references within the class body. Additional
tests cover sequential assignment, references to undefined variables,
and repeated assignments. If anyone has more pathological examples, I'm
happy to add them to the test suite.


mscolnick (approved):
Looks great

Issue #4539, opened 2025-04-15
Autocomplete/auto suggest on subscriptable objects


astrowonk:
### Description

Jupyter and ipython support autocompletion of subscriptable objects via defining a `_ipython_key_completions_` method. [This](https://ipython.readthedocs.io/en/stable/config/integrating.html) allows any class that creates a subscriptable instance to define what can be autocompleted. 

I use this extensively to make custom objects that are subscriptable and that return data structures I access regularly but can't remember all the names. i.e.  `DataGrabber['table_name']` or what have you.

Marimo's autocomplete works perfectly on dictionary, but fails on custom subscriptable objects.

I also notice that there is no autocomplete or auto-suggest for pandas dataframes, failing to autocomplete or suggest the column names.

I tried this with and without LSP support installed and turned on.

![Image](https://github.com/user-attachments/assets/fdc7cf90-ceb4-4466-8942-ee5298bda064)

### Suggested solution

Custom objects and pandas dataframes should show autocomplete suggestions for items within just like a dictionary.

mscolnick:
@astrowonk - is this a more appropriate feature-request for [jedi](https://github.com/davidhalter/jedi) or [python-lsp-server](https://github.com/python-lsp/python-lsp-server), if these libraries can support it, we can support it.

mscolnick:
We use an LSP now (opt-in) for our completions. We use python-lsp-server.

When it is not installed, we fallback to `jedi` with some of our own logic (heavy lifting done by jedi). You can find that in `marimo/_runtime/complete.py`

gundersona:
Lack of autocomplete suggestions for pandas dataframes is big pain-point for me. Looking forward to this enhancement!

zilto:
Bump. It's a valuable feature to implement. Many libraries implement useful autocomplete

> is this a more appropriate feature-request for [jedi](https://github.com/davidhalter/jedi) or [python-lsp-server](https://github.com/python-lsp/python-lsp-server), if these libraries can support it, we can support it.

jedi and LSP are intended for static analysis (e.g., format the code, infer from hints). Here, autocompletion requires runtime inspection (e.g., you need the `pd.DataFrame` in-memory to know the potential columns autocomplete).

These features are built on top of IPython, more specifically the [core.completer](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.completer.html#) module.

### Why it's cool

[...]

PR #4683, opened 2025-04-26
feat: Ibis SQL engine

zilto:
## 📝 Summary

Implements #4546 

A `IbisEngine` class was added, mimicking the `SQLAlchemyEngine`. It automatically pickups Ibis `Backend` objects and displays them in the database panel.

## 🔍 Description of Changes

The code reimplements the logic of the `SQLAlchemyEngine`. It was done by implementing the `IbisEngine` and adding code wherever `SQLAlchemyEngine` was referenced.

 ### Notes
 - Ibis doesn't use the hierarchy `table <- schema <- database`. Instead, it uses `table <- database <- catalog`
 - Could add a `nested` / `complex` / `object` type in the UI. Right now, everything defaults to `string`
 - Could add a `Add table to notebook` in Python instead of SQL. It would insert the statement
   ```python
   table = con.table("table_name")
   ```
 - More thorough testing could be done to verify individual Ibis backends.
 - Could add a frontend component with a `Ibis` button that adds code to the cell (similar to SQLAlchemy) 

image

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

## 📜 Reviewers

I chatted with @mscolnick and see that @Light2Dark assigned the issue two days ago

PR #5618, opened 2025-07-13
feat: interactive key completion

zilto:
## 📝 Summary

Implements key autocompletion (issue #4539). Now, `dataframe["` will suggest column names, dictionaries will suggest keys, etc.

## 🔍 Description of Changes

Existing autocompletion features rely on `jedi`, which is a static analysis tool. It works using the source code / AST or by loading and inspecting an isolated copy of the module. It never accesses the user's namespaces (i.e., where the variable from the notebook live). Key completion requires accessing the object instance in the user namespace. 

`_maybe_get_key_options` was added and is called from the existing `complete()` function (main entrypoint). This identifies the name of the variable being completed, retrieves it from the already available `glbls` namespace and retrieves keys from the instance

### Limitations

- non-`str` dictionary keys aren't show as suggestions. When those are involved (`int`, `float`,  `tuple`), debug mode shows serialization warnings. Even if those values are converted to strings explicitly, something is failing
- the changes explicitly 3 cases to retrieve keys (has `_ipython_key_completions_()`, is `Mapping`, has `.keys()`). However, there's no standard about "how to retrieve valid values to pass to `.__getitem__()`

## 📋 Checklist

## Screenshots

image

image

image

### DIFF
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -9,6 +9,7 @@ from functools import lru_cache
 from typing import TYPE_CHECKING, Any, cast
 
 import jedi  # type: ignore # noqa: F401
+import jedi.api  # type: ignore # noqa: F401
 
 from marimo import _loggers as loggers
 from marimo._dependencies.dependencies import DependencyManager
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -2,10 +2,12 @@
 from __future__ import annotations
 
 import html
+import sys
 import threading
 import time
 from collections.abc import Mapping
 from functools import lru_cache
+from types import ModuleType
 from typing import TYPE_CHECKING, Any, cast
 
 import jedi  # type: ignore # noqa: F401
@@ -344,6 +346,47 @@ def _get_completions_with_interpreter(
         completions = script.complete()
     return script, completions
 
+# TODO move this to a global utility module
+def _isinstance_external(obj: Any, *, class_ref: str) -> bool:
+    import_parts = class_ref.split(".")
+    module_name = import_parts[0]
+
+    if module_name not in sys.modules:
+        return False
+    
+    module: ModuleType = sys.modules[module_name]
+    target_class: Any = module
+    for part in import_parts:
+        target_class = getattr(target_class, part)
+
+    return isinstance(obj, target_class)
+
+
+def _key_options_from_ipython_method(obj: Any) -> list[str]:
+    return obj._ipython_key_completions_()
+
+
+def _key_options_from_mapping(obj: Mapping) -> list[str]:
+    return [str(key) for key in obj.keys()]
+
+
+def _key_options_from_pandas(obj: Any) -> list[str]:
+    return _key_options_from_mapping(obj)
+
+
+def _key_options_dispatcher(obj: Any) -> list[str]:
+    if getattr(obj, "_ipython_key_completions_", False):
+        return _key_options_from_ipython_method(obj)
+    elif isinstance(obj, Mapping):
+        return _key_options_from_mapping(obj)
+    elif _isinstance_external(obj, class_ref="pandas.DataFrame"):
+        return _key_options_from_pandas(obj)
+        
+    LOGGER.debug(
+        f"No matching handlers found to retrieve keys from type `{type(obj)}`"
+    )
+    return []
+
 
 def _get_key_options(
     script: jedi.Script, glbls: dict[str, Any]
@@ -369,20 +412,7 @@ def _get_key_options(
         LOGGER.debug(f"Failed to find `{obj_name=:}` in `glbls`")
         return []
 
-    # TODO refactor handler to separate metho as complexicty grows
-    if getattr(obj, "_ipython_key_completions_", False):
-        keys = obj._ipython_key_completions_()
-    elif isinstance(obj, Mapping):
-        keys = obj.keys()
-    # for instance, pandas.DataFrame doesn't inherit from Mapping
-    elif getattr(obj, "keys", False):
-        keys = obj.keys()
-    else:
-        LOGGER.debug(
-            f"No matching handlers found to retrieve keys from type `{type(obj)}`"
-        )
-        keys = []
-
+    keys = _key_options_dispatcher(obj)
     # TODO currently unreliable for non-string keys, even if stringified
     # seems to be related to serialization issue if include `"(True, False)` (no closing quote)
     return [
@@ -394,12 +424,15 @@ def _get_key_options(
 
 
 def _maybe_get_key_options(
-    request: CodeCompletionRequest, script: jedi.Script, glbls: dict[str, Any]
+    request: CodeCompletionRequest, script: jedi.Script, glbls: dict[str, Any], glbls_lock
 ) -> list[CompletionOption]:
     """Call key completions methods if the request contains the trigger"""
     triggers = ('["', "['")  # explicitly handle both quotation characters
     if request.document[-2:] in triggers:
-        return _get_key_options(script, glbls)
+        locked = False
+        completions: list[jedi.api.classes.Completion] = []
+        locked = glbls_lock.acquire(blocking=False)
+        return _get_key_options(script, glbls, locked)
 
     return []
 
@@ -475,7 +508,12 @@ def complete(
 
         # TODO we can probably return early if we have key completion;
         # it has different triggers and retrieval should never overlap with attribute completion
-        key_options = _maybe_get_key_options(request, script, glbls)
+        key_options = _maybe_get_key_options(
+            request,
+            script,
+            glbls=glbls,
+            glbls_lock=glbls_lock,
+        )
 
         if (
             prefix_length == 0
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -4,11 +4,11 @@ from __future__ import annotations
 import html
 import threading
 import time
+from collections.abc import Mapping
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any, cast
 
 import jedi  # type: ignore # noqa: F401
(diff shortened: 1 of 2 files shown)


zilto:
(CLA assistant is failing, but I previously agreed [here](https://github.com/marimo-team/marimo/pull/4683) where I was facing issues)

marimo/_runtime/complete.py
```diff
@@ -344,6 +345,65 @@ def _get_completions_with_interpreter(
...
+        LOGGER.debug(f"Failed to find `{obj_name=:}` in `glbls`")
+        return []
+
+    # TODO refactor handler to separate metho as complexicty grows
+    if getattr(obj, "_ipython_key_completions_", False):
+        keys = obj._ipython_key_completions_()
+    elif isinstance(obj, Mapping):
+        keys = obj.keys()
+    # for instance, pandas.DataFrame doesn't inherit from Mapping
+    elif getattr(obj, "keys", False):
```
akshayka: Instead of duck typing here, we should enumerate specific types where we know it is safe (inexpensive, not side-effectful) to call `keys`).

We can explicitly check if the object is a Pandas dataframe (after first checking that Pandas has been imported ie is in sys modules).

marimo/_runtime/complete.py
```diff
@@ -344,6 +345,65 @@ def _get_completions_with_interpreter(
...
+
+    obj_name = names[-1].name
+    obj = glbls.get(obj_name)
+    if obj is None:
+        LOGGER.debug(f"Failed to find `{obj_name=:}` in `glbls`")
+        return []
+
+    # TODO refactor handler to separate metho as complexicty grows
+    if getattr(obj, "_ipython_key_completions_", False):
+        keys = obj._ipython_key_completions_()
```
akshayka: Nice!

marimo/_runtime/complete.py
```diff
@@ -344,6 +345,65 @@ def _get_completions_with_interpreter(
...
 
+def _get_key_options(
+    script: jedi.Script, glbls: dict[str, Any]
+) -> list[CompletionOption]:
+    """Get completion values for trigger `["` or `['`. Values are meant to be
+    passed to `.__getitem__()`
+
+    This implementation assumes that marimo's `CompletionRequest.document`
+    only includes current line, up to the completion trigger.
+    """
```
akshayka, line 427: This method should only be called when `globals_lock` is held. Unlike `ipykernel`, in marimo the completer runs in a separate thread, so that we can fetch completions even while the kernel is running user code. But this means that the globals dictionary should only be accessed while the lock is held, otherwise we risk race conditions.

See the docstring of `complete` for information the lock.

The completer should also never wait for the runtime to finish, meaning if the lock is already held, then we should instead just bail on getting dynamic completions. For an example, see how the lock is used here: https://github.com/marimo-team/marimo/blob/c70b0ef831ab8e88efe7b3c3cf12e64280a8a83b/marimo/_runtime/complete.py#L327-L344

marimo/_runtime/complete.py
```diff
@@ -344,6 +345,65 @@ def _get_completions_with_interpreter(
...
+    obj = glbls.get(obj_name)
+    if obj is None:
+        LOGGER.debug(f"Failed to find `{obj_name=:}` in `glbls`")
+        return []
+
+    # TODO refactor handler to separate metho as complexicty grows
+    if getattr(obj, "_ipython_key_completions_", False):
+        keys = obj._ipython_key_completions_()
+    elif isinstance(obj, Mapping):
+        keys = obj.keys()
```
akshayka: Completion for dictionaries appears to already be implemented in 0.14.10, thanks to `_get_completions_with_interpreter` (https://github.com/marimo-team/marimo/blob/c70b0ef831ab8e88efe7b3c3cf12e64280a8a83b/marimo/_runtime/complete.py#L327).

For example:

image

akshayka (changes requested):
This change introduces a race condition since marimo's completer runs in a separate thread to provide completions even when the kernel is running; more information in comments.

marimo/_runtime/complete.py
zilto: This is slightly different. Jedi only does static analysis, even with the `Interpreter`. In this screenshot, it infers keys based on values harded-code in the cell. 

In the initial post, I show autocompletion for a dictionary with `random` keys defined at runtime. Similarly, dataframe columns are typically not defined in the sourec

Here's a short clip where I toggle features from this PR. (I should be a on a fork of `0.14.10`)

https://github.com/user-attachments/assets/9efb80de-08b7-402b-9f64-49670dd9d61e

zilto:
The test that failed in CI `tests/_server/api/endpoints/test_files.py::test_save_app_config` does succeed locally

marimo/_runtime/complete.py
```diff
@@ -344,6 +347,118 @@ def _get_completions_with_interpreter(
...
+    """Get completion values for trigger `["` or `['`. Values are meant to be
+    passed to `.__getitem__()`
+
+    This implementation assumes that marimo's `CompletionRequest.document`
+    only includes current line, up to the completion trigger.
+    """
+    names = script.get_names(
+        references=True, definitions=False, all_scopes=False
+    )
+    if len(names) == 0:
```
akshayka: Nit:

```suggestion
    if not names:
```

marimo/_runtime/complete.py
```diff
@@ -344,6 +347,118 @@ def _get_completions_with_interpreter(
...
+    obj = glbls.get(obj_name)
+    if obj is None:
+        LOGGER.debug(f"Failed to find `{obj_name=:}` in `glbls`")
+        return []
+
+    keys = _key_options_dispatcher(obj)
+    # TODO currently unreliable for non-string keys, even if stringified
+    # seems to be related to serialization issue if include `"(True, False)` (no closing quote)
+    return [
+        CompletionOption(name=key, type="property", completion_info="property")
```
akshayka:
```suggestion
        CompletionOption(name=key, type="property", completion_info="key")
```

marimo/_runtime/complete.py
```diff
@@ -344,6 +347,118 @@ def _get_completions_with_interpreter(
...
+    based on its methods and type.
+
+    Returns a list of strings to later create marimo `CompletionOption`
+    """
+    if getattr(obj, "_ipython_key_completions_", False):
+        return _key_options_from_ipython_method(obj)
+    elif isinstance(obj, Mapping):
+        return _key_options_from_mapping(obj)
+    elif _isinstance_external(obj, class_ref="pandas.DataFrame"):
+        return _key_options_from_pandas(obj)
```
akshayka: It would be great to support Polars too, but we can do that in a follow up

marimo/_runtime/complete.py
```diff
@@ -344,6 +347,118 @@ def _get_completions_with_interpreter(
...
+    if request.document[-2:] not in triggers:
+        return []
+
+    # TODO ideally, we don't acquire the lock twice with `_get_completions_with_interpreter()` and `_maybe_get_key_options`
+    # however, the two functions are applied on different triggers
+    # also, this other function returns Jedi `Completion` while this returns native `CompletionOption`
+    locked = False
+    locked = glbls_lock.acquire(blocking=False)
+    if locked:
+        return _get_key_options(script, glbls)
```
akshayka: Since we might acquiring the lock in two places now, we now need to make sure that the lock gets released the same number of times that it is acquired. Otherwise other threads may deadlock (https://docs.python.org/3/library/threading.html#rlock-objects).

```suggestion
    locked = False
    locked = glbls_lock.acquire(blocking=False)
    if locked:
        try:
            return _get_key_options(script, glbls)
        finally:
            glbls_lock.release()
```

zilto:
> Additionally, I noticed keys completion doesn't appear to work on mappings when have at least one string literal key.

Ok, so the "bug" is due to different options requiring different `prefix_length`. Jedi has one value for its completions while this new built-in mechanism always has `prefix_length=0`. Results are sent to the frontend, but it probably breaks some display assumptions (e.g., `CompletionOption` message only accepts a single `prefix_length`).

This was fixed by separating the return values for key completion (new feature) and attribute completion (existing feature).

I'm very happy with the test coverage. Although a bit messy, `test_complete_main_entrypoint()` could be expanded to support other Jedi results

zilto:
The failing test `tests/_server/api/endpoints/test_files.py::test_read_code` seems unrelated to the changes and passes locally. How should I proceed?

marimo/_runtime/complete.py
```diff
@@ -443,12 +594,11 @@ def complete(
                 )
                 return
 
-        if not completions:
+        if not completions and not key_options:
```
akshayka: nit: I believe `key_options` is guaranteed to be empty here given the early return above

```suggestion
        if not completions:
```

akshayka (dismissed):
One small nit. Otherwise LGTM. Thank you for this PR!

marimo/_runtime/complete.py
zilto: Good catch! Forgot to remove this check

mscolnick:
@zilto can you try appeasing the CLA bot again, just so it doesn't keep propping up

zilto:

edit: It seemed to have worked now that I cleaned up my local git credentials!

liquidcarbon:
should the options here be "x", "y", "z"?

image

zilto:
@liquidcarbon Interesting case. It seems to apply to `dict` and all other supported types. 

The completion trigger needs to be modified. Will work on a fix.

liquidcarbon:
The use case for this might be:
1) quick lookup of possible values within a column
2) autocomplete for creating filters - as in `df["a"] == <choices>` - could this work?
One might have to consider limiting the choices on high-cardinality fields.

The other day found myself wishing for `df["a"]["x"]` as a substitute for "give me row(s) where `df.a == "x"` but it's a different story

PR #5620, opened 2025-07-13
improvement: navigation hotekys when on a cell

mscolnick:


frontend/src/components/editor/focus/focus.ts
```diff
@@ -37,35 +46,97 @@ export function useCellNavigationProps(cellId: CellId) {
...
+      if (evt.key === "b" && !Events.hasModifier(evt)) {
+        actions.createNewCell({
+          cellId,
+          before: false,
+          autoFocus: true,
+        });
       }
 
       // j/k movement in vim mode.
       if (keymapPreset === "vim") {
```
mscolnick, line 146: should we enable `j/k` outside vim mode? it might be nice to have

Light2Dark:
lgtm, suggestion once you merge the other PR into here. 
- save notebook to be the same as Ctrl-S (?), we don't need `s` to save, I think.
- we could add a delete cell (doesn't have to be now)

mscolnick:
> save notebook to be the same as Ctrl-S (?), we don't need s to save, I think.

I added `s` for jupyter parity

> we could add a delete cell (doesn't have to be now)

yea we could next. i didn't want to yet since it can be more destructive, might need more work than just adding the shortcut

Light2Dark (approved):

PR #5630, opened 2025-07-14
fix: Enable cache decoration on class methods

dmadisetti:
## 📝 Summary

Follow up from #5613 

This enables `@mo.cache` and `@mo.persistent_cache` to be used on class objects.

```python
class MyClass:
    def __init__(self, x=0):
        self.x = x

    @mo.cache
    def expensive_function(self, args):
        """NB. Pickle of self is invalidating"""
```

Notably, the pickle of the object instance is invalidating. That is:

```python
MyClass(x=0).expensive_function(1) # Cache miss
MyClass(x=1).expensive_function(1) # Cache miss
MyClass(x=0).expensive_function(1) # Cache hit
```

Non-pickable object instances may lead to undefined behavior.

### DIFF
--- a/marimo/_save/save.py
+++ b/marimo/_save/save.py
@@ -69,19 +69,36 @@ if TYPE_CHECKING:
 class _cache_call:
     """Like functools.cache but notebook-aware. See `cache` docstring"""
 
+    __slots__ = (
+        "base_block",
+        "scope",
+        "scoped_refs",
+        "pin_modules",
+        "hash_type",
+        "_args",
+        "_var_arg",
+        "_var_kwarg",
+        "_loader",
+        "_loader_partial",
+        "_bound",
+        "_last_hash",
+        "_frame_offset",
+        "__wrapped__",
+    )
+
     base_block: BlockHasher
     scope: dict[str, Any]
     scoped_refs: set[str]
     pin_modules: bool
     hash_type: str
     _args: list[str]
-    _var_arg: Optional[str] = None
-    _var_kwarg: Optional[str] = None
-    _loader: Optional[State[Loader]] = None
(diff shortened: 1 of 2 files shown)


marimo/_save/save.py
```diff
@@ -63,23 +63,27 @@
...
     _var_kwarg: Optional[str] = None
     _loader: Optional[State[Loader]] = None
     _loader_partial: LoaderPartial
-    name: str
-    fn: Optional[Callable[..., Any]]
+    _bound: Optional[dict[str, Any]]
+    _last_hash: Optional[str] = None
+    _frame_offset: int = 0
+    # Consistent with functools.cache
+    __wrapped__: Optional[Callable[..., Any]]
```
akshayka: Is there a reason these are class variables instead of instance variables?

marimo/_save/save.py
```diff
@@ -206,9 +211,41 @@ def loader(self) -> Loader:
         assert self._loader is not None, UNEXPECTED_FAILURE_BOILERPLATE
         return self._loader()
 
+    def __get__(
+        self, instance: Any, _owner: Optional[type] = None
+    ) -> _cache_call:
```
akshayka, line 236: Since this is somewhat complex logic, can you leave a 1 or 2 sentence docstring for posterity? :)

marimo/_save/save.py
dmadisetti: Woops. Meant to add `__slots__`

marimo/_save/save.py
dmadisetti: Done! Made me explore the behavior a bit more too. Not the most intuitive path, but I think it is fine

PR #5631, opened 2025-07-14
consolidate / cleanup cell keyboard props

mscolnick:
Some refactors to move all the cell shortcuts into the same files, and renaming some files


mscolnick:
merging since i am build off this, can do a post-review

PR #5636, opened 2025-07-14
Add vim bindings for top/bottom cell navigation: `gg` and `G`

manzt:
Adds `gg` to jump to first cell and `G` to jump to last cell in command mode,
matching standard vim behavior. Also refactors vim keybindings into a dedicated
handler with declarative binding definitions, making it easy to add more vim
commands in the future.

### DIFF
--- a/frontend/src/components/editor/navigation/navigation.ts
+++ b/frontend/src/components/editor/navigation/navigation.ts
@@ -23,6 +23,7 @@ import {
   useIsCellSelected,
 } from "./selection";
 import { temporarilyShownCodeAtom } from "./state";
+import { handleVimKeybinding } from "./vim-bindings";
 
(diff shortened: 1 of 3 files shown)


manzt:
@mscolnick sounds good, the changes here are pretty simple so I can roll them back and rebase the same idea on whatever we get merged. Most of the work was trying to understand where / how to define these things in the future.

mscolnick (approved):
this is great, nice implementation. maybe a few tests for the buffer part?

mscolnick:
this does lead to duplication logic for the shortkeys and additional vim ones. in my branch, I have a way to dedupe the logic, so that may change this implementation

manzt:
that's fine. i'll just wait until we get whatever you have merged and figure it out!

PR #5647, opened 2025-07-15
feat: allow auto download to be configured globally

dmadisetti:
## 📝 Summary

Setting `default_auto_download` in your `marimo.toml` will automatically set exports for all your notebooks.

For instance,

```toml
[runtime]
...
default_auto_download=["html"]
...
```

will download save html artifacts in `__marimo__/`. Read more about exports here: https://docs.marimo.io/guides/expensive_notebooks/#disable-individual-cells

an aside, this can also be set with `_MARIMO_APP_OVERLOAD_AUTO_DOWNLOAD="[html]"`

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -95,6 +95,14 @@ OnCellChangeType = Literal["lazy", "autorun"]
 ExecutionType = Literal["relaxed", "strict"]
 
 
+# TODO(akshayka): remove normal, migrate to compact
+# normal == compact
+WidthType = Literal["normal", "compact", "medium", "full", "columns"]
+Theme = Literal["light", "dark", "system"]
+ExportType = Literal["html", "markdown", "ipynb"]
+SqlOutputType = Literal["polars", "lazy-polars", "pandas", "native", "auto"]
+
+
 @mddoc
 @dataclass
 class RuntimeConfig(TypedDict):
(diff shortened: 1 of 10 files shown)


dmadisetti:
One caveat is that this doesn't backfill. Only for new notebooks. The `env` is a blanket override though

marimo/_config/config.py
```diff
@@ -133,6 +141,9 @@ class RuntimeConfig(TypedDict):
     - `default_sql_output`: the default output format for SQL queries. Can be one of:
         `"auto"`, `"native"`, `"polars"`, `"lazy-polars"`, or `"pandas"`.
         The default is `"auto"`.
+    - `default_auto_download`: an Optional list of export types to automatically:
```
akshayka:
```suggestion
    - `default_auto_download`: an Optional list of export types to automatically snapshot your notebook as:
```

marimo/_config/config.py
```diff
@@ -133,6 +141,9 @@ class RuntimeConfig(TypedDict):
...
+       `html`, `markdown`, `ipynb`.
```
akshayka, line 145:
```suggestion
       `html`, `ipynb`.
```

I suggest we remove markdown because markdown doesn't have outputs (it's also not an option in the frontend UI). It's cheap to export as markdown from the CLI (but not cheap to go to HTML or IPYNB since that runs the notebook)

akshayka (approved):
Awesome! Two requests but otherwise looks good to me. In another PR we could follow-up with a frontend change for the user config menu?

marimo/_config/config.py
mscolnick: we already support this in `auto_download`, so i think it makes sense for parity (we don't need to expose it in the UI though)

dmadisetti: I'm going to leave it in the code logic- but fine removing it from the docs

PR #5662, opened 2025-07-16
fix: mpl.interactive with remote backends

mscolnick:
This fixes `mo.mpl.interactive()` to work properly with remote backends. Before `mpl.interactive` was using a relative path `/` which is just where the frontend was served. This instead adds a header param for the remote backend URL (from the user's perspective) that the mpl widget can pick up to serve the interactive websocket at with an absolute URL instead of relative.

### DIFF
--- a/frontend/src/core/runtime/__tests__/runtime.test.ts
+++ b/frontend/src/core/runtime/__tests__/runtime.test.ts
@@ -244,6 +244,7 @@ describe("RuntimeManager", () => {
           "Authorization": "Bearer test-token",
           "Marimo-Server-Token": "",
           "Marimo-Session-Id": "test-session-id",
+          "x-runtime-url": "https://example.com/",
         }
       `);
     });
@@ -255,6 +256,7 @@ describe("RuntimeManager", () => {
       expect(headers).toEqual({
         "Marimo-Session-Id": "test-session-id",
         "Marimo-Server-Token": "",
+        "x-runtime-url": "http://localhost:8080/",
       });
     });
   });


marimo/_plugins/stateless/mpl/_mpl.py
```diff
@@ -101,11 +102,34 @@ def _get_secure() -> bool:
...
+    if not request:
+        return ""
+
+    base_url = request.headers.get("x-runtime-url")
+    if not base_url:
+        return ""
+    return base_url.rstrip("/")
+
+
+def _convert_scheme_to_ws(url: str) -> str:
```
dmadisetti, line 116: Do we have a util for this?

marimo/_server/main.py
```diff
@@ -118,21 +118,54 @@ def create_starlette_app(
...
+        if not remaining.startswith("/"):
+            raise ValueError(f"Invalid MPL path format: {path}")
+
+        # Parse /<port>/rest/of/path to get the port
+        parts = remaining.split("/")
+        if len(parts) < 2 or not parts[1]:
+            raise ValueError(f"Invalid MPL path format: {path}")
+
+        port = parts[1]
         return f"http://localhost:{port}"
```
dmadisetti, line 145: maybe needs `--proxy`
Worth leaving a note even if we don't change it rn?

marimo/_plugins/stateless/mpl/_mpl.py
mscolnick: i dont see one

marimo/_server/main.py
mscolnick: i dont think this needs `--proxy`. this is already coming from what the user sees (and BE they are connected to)

dmadisetti: The server could be started with something like `marimo edit --host=198.x.x.x --proxy=entrydomain.ex`

Sure an edgecase, but blanket `localhost` is not correct. Proxy should represent the route the user expects to be followed

mscolnick: yes, localhost may not be correct, but in this case i think `198.x.x.x` would be correct because this is internal. going to `entrydomain.ex` does not seem right

mscolnick: anyways, we can make this change separately

dmadisetti: Cool, just wanted to make a note. With something like nginx on top, the example is fine

dmadisetti (approved):
Looks good enough to unblock. We'll circle back for the relative path

PR #5664, opened 2025-07-16
Add destructive cell deletion confirmation

### LINKED ISSUE #5658: Jupyter like key bindings for deleting cells
### Description

Recently in marimo 0.14.11 the jupyter key bindings like a, b, shift + enter was added, thank you!

i believe dd was missed for deleting cells?

can you add this in please?

### Suggested solution

can you add the same as found in jupyter please?

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

manzt:
Fixes #5658

This change adds a keymap configuration (`destructive_delete`, default **True**) to enable controlled, destructive deletion of cells via keyboard shortcut. Video:

- First deletion is "cheap" (`d = 10`), and automatically deleted
- Section deletion is "expensive", and adds confirmation step

The idea is to allow deletion of cells, but to have a confirmation step (i.e., "are you sure you want to do that?") when we consider a deletion _could_ be "expensive". The definition of expensive is up for question but right now we can look at:

- Whether a cells execution time is above some threshold (e.g., 2s)
- Whether a cell declares variables used by other cells

If neither of those conditions are met, we automatically delete the cell.

What I hope this unlocks is the ability to readily delete cells for the common editing use case, and give pause for the moments where a deletion might be costly or unintentional. My preference is to have opt-out unless the user explicitly prefers stricter safeguards.

### DIFF
--- a/frontend/src/core/cells/__tests__/pending-delete-service.test.tsx
+++ b/frontend/src/core/cells/__tests__/pending-delete-service.test.tsx
@@ -1,0 +1,202 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+
+import { act, renderHook, waitFor } from "@testing-library/react";
+import { createStore, Provider } from "jotai";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { MockNotebook } from "@/__mocks__/notebook";
+import { notebookAtom } from "@/core/cells/cells";
+import { CellId } from "@/core/cells/ids";
+import { createCellRuntimeState } from "@/core/cells/types";
+import { variablesAtom } from "@/core/variables/state";
+import type { VariableName } from "@/core/variables/types";
+import type { Milliseconds } from "@/utils/time";
+import {
+  usePendingDelete,
+  usePendingDeleteService,
+} from "../pending-delete-service";
+
+const mockDeleteCell = vi.fn();
+const mockDeleteManyCells = vi.fn();
+vi.mock("@/components/editor/cell/useDeleteCell", () => ({
+  useDeleteCellCallback: () => mockDeleteCell,
+  useDeleteManyCellsCallback: () => mockDeleteManyCells,
+}));
+
+function createTestWrapper() {
+  const store = createStore();
+  const wrapper = ({ children }: { children: React.ReactNode }) => (
+    <Provider store={store}>{children}</Provider>
+  );
+  return { wrapper, store };
+}
+
+// Don't clear all mocks - just reset the specific one we care about
+beforeEach(() => {
+  vi.resetAllMocks();
+});
+
+describe("pending-delete-service", () => {
+  it("auto-deletes simple cells (no dependencies, short execution time)", async () => {
+    const { wrapper, store } = createTestWrapper();
+
+    const cellId = CellId.create();
+    const notebook = MockNotebook.notebookState({
+      cellData: {
+        [cellId]: { code: "print('hello')" },
+      },
+    });
+
+    notebook.cellRuntime[cellId] = createCellRuntimeState({
+      runElapsedTimeMs: 1 as Milliseconds,
+    });
+
+    store.set(notebookAtom, notebook);
+    store.set(variablesAtom, {});
+
+    const { result } = renderHook(
+      () => {
+        const service = usePendingDeleteService();
+        const deleteState = usePendingDelete(cellId);
+        return { service, deleteState };
+      },
+      { wrapper },
+    );
+
+    act(() => {
+      result.current.service.submit([cellId]);
+    });
+
+    // should flush single cell
+    await waitFor(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/editor/cell/PendingDeleteConfirmation.tsx
```diff
@@ -0,0 +1,193 @@
...
+  }));
+
+  const hasExpensiveExecution = executionTimeMs > 2000;
+  const hasDependencies = downstreamCells.length > 0;
+  const isExpensiveOrHasDeps = hasExpensiveExecution || hasDependencies;
+  const isMultiPending = pendingCells.size > 1;
+
+  // Auto-delete if not expensive and not multi-pending
+  useEffect(() => {
+    if (!isMultiPending && !isExpensiveOrHasDeps) {
```
mscolnick: maybe pull this out `const autoDelete = !isMultiPending && !isExpensiveOrHasDeps` and reuse below

frontend/src/components/editor/cell/PendingDeleteConfirmation.tsx
```diff
@@ -0,0 +1,193 @@
...
+  if (hasExpensiveExecution && hasDependencies) {
+    warningMessage = `This cell took ${formattedTime} to run and contains variables referenced by other cells.`;
+  } else if (hasExpensiveExecution) {
+    warningMessage = `This cell took ${formattedTime} to run.`;
+  } else if (hasDependencies) {
+    warningMessage = "This cell contains variables referenced by other cells.";
+  }
+
+  // Variables declared by this cell
+  const defs = Object.values(variables)
```
mscolnick: nit: this could be combined in the same loop as above where we get the `downstream` cell ids

manzt: was able to clean up the logic quite a bit

manzt:
@mscolnick This currently is a little awkward with multi-select in `navigation.ts` (dd or Shift +delete). I for now, we could just do a no-op for `shortcuts["cell.delete"]` if there are multiple cells?

mscolnick:
@manzt find to not support multi-delete for now (it is still supported via click). i think we can pull out the callback into 

```diff
+ const handleDelete = (cellId) => {
+   if (selection.size > 1) {
+.     return false;
+.  }

    // Only handle if destructive_delete is enabled
    if (!userConfig.keymap.destructive_delete) {
      return false;
    }
    // Cannot delete running cells
    const notebook = store.get(notebookAtom);
    const cellData = notebook.cellRuntime[cellId];
    const hasRunningCell =
      cellData.status === "running" || cellData.status === "queued";
    if (hasRunningCell) {
      return false;
    }
    setPendingCells(new Set([cellId]));
    return true;
  }
```

in a followup we could the instead change the ui of `multi-cell-action-toolbar.tsx` to render a different body

frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -429,10 +422,56 @@ export function useCellNavigationProps(
...
+
+          // Only handle if destructive_delete is enabled
+          if (!userConfig.keymap.destructive_delete) {
+            return false;
+          }
+          // Cannot delete running cells
+          const notebook = store.get(notebookAtom);
+          const cellData = notebook.cellRuntime[cellId];
+          const hasRunningCell =
+            cellData.status === "running" || cellData.status === "queued";
```
Light2Dark: Non-blocker:

Would it be safe to delete running cells, sometimes `Stop` doesn't work and force deleting could help. 

I also notice while I can't use the shortcut to delete and the button is disabled, I can expand the cell actions and click delete. Is this intentional (more clicks to force delete)?

Light2Dark:
Since this is enabled by default, could the keyboard shortcut name be changed to `delete cell`
CleanShot 2025-07-17 at 15 50 02@2x

mscolnick (dismissed):

akshayka:
UI/UX LGTM — very nice!!

frontend/src/components/editor/navigation/navigation.ts
manzt: > Would it be safe to delete running cells, sometimes Stop doesn't work and force deleting could help.

I'm not totally sure about this. I think stoping the kernel is a different action and I would worry that also deleting could lead to some race condition (/ not actually stop the kernel).

> I also notice while I can't use the shortcut to delete and the button is disabled, I can expand the cell actions and click delete. Is this intentional (more clicks to force delete)?

I think this should be changed now. Originally yes, just required more clicks.

akshayka: Deleting running cells would hide the fact that the cell is still running (the delete will only be processed in the backend after the cell finishes running), which could cause strange "action-at-a-distance" behavior.

PR #5666, opened 2025-07-16
'marimo env' shows if it is an editable install

mscolnick:


dmadisetti (dismissed):

dmadisetti:
test looks unrelated

dmadisetti:
Actually wrong json format

image

mscolnick:
@dmadisetti the keys are all strings. doesn't need to be a boolean. its still correct json

dmadisetti:
The other keys are Boolean, I'm not sure why you would mix

PR #5668, opened 2025-07-17
Migrate to pnpm workspace layout

manzt:
This PR lifts shared packages into a proper monorepo structure using pnpm workspaces.

- `lsp/` → `packages/lsp`
- `openapi/` → `packages/openapi`
- Centralized root `package.json`, `tsconfig.json`, and `pnpm-lock.yaml`
- Updated `Makefile` to use `pnpm --filter` instead of manual `cd`
- Added shared `.gitignore`
- Removed vestigial lockfiles


package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "scripts": {
+    "typecheck": "pnpm run -r typecheck",
+    "format": "pnpm biome check --write",
```
manzt: Could also call `fix`

packages/lsp/index.test.ts
```diff
@@ -1,13 +1,17 @@
...
   it("should start and be killable", async () => {
-    const process = spawn("node", [
+    const process = childProcess.spawn("node", [
       "./dist/index.cjs",
       "--lsp",
-      "echo",
-      "hello",
+      "echo hello",
+      "--log-file",
+      path.join(os.tmpdir(), "lsp-server-test.log"),
```
manzt, line 14: This test was broken but not run anywhere...

PR #5676, opened 2025-07-17
Change "cell.delete" hotkey description from "Delete empty cell" to "Delete cell"

manzt:
Follow up to #5664. We now support deleting cells destructively using
hotkeys.


manzt:
cc: @Light2Dark , thanks for the attention to detail!

mscolnick (approved):

Light2Dark (approved):

PR #5677, opened 2025-07-17
fix: prompt user to open invalid files

dmadisetti:
## 📝 Summary

Previously, we would just overwrite files if they were not recognized as a marimo file.
Now we ask for confirmation if it is not a valid file

image

Otherwise we show a warning

image

Otherwise continue as normal. Also fixed a comparison bug that meant we never detected the run guard properly.

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -14,7 +14,7 @@ import click
 import marimo._cli.cli_validators as validators
 from marimo import __version__, _loggers
 from marimo._ast import codegen
-from marimo._ast.load import notebook_is_openable
+from marimo._ast.load import get_notebook_status
 from marimo._cli.config.commands import config
 from marimo._cli.convert.commands import convert
 from marimo._cli.development.commands import development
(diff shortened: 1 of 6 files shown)


marimo/_cli/cli.py
```diff
@@ -70,6 +70,18 @@ def check_app_correctness(filename: str) -> None:
         click.echo(f"Failed to parse notebook: {filename}\n", err=True)
         raise click.ClickException(traceback.format_exc(limit=0)) from None
 
+    if status == "invalid":
+        click.confirm(
+            "The notebook is invalid. Do you want to open it anyway?",
```
mscolnick: could it be something like `"The notebook is invalid or not a marimo notebook. Do you want to open it anyway?"`

either we should say something like `This may overwrite its contents.` or We should put the contents in the first cell

dmadisetti: Suggested the user use `marimo convert` over putting it in the first cell.

image

Definitely need to be more explicit about potential overwriting.

mscolnick (dismissed):
maybe a few more tests for testing the result type of the parse (not sure all possible types are covered)

PR #5684, opened 2025-07-18
fix: don't write snapshots on empty session views

mscolnick:
If session views are empty, just skip writing it. 

Refactor to clean up the args to just pass down the `App` instead of `AppFileManager`

### DIFF
--- a/marimo/_server/session/session_view.py
+++ b/marimo/_server/session/session_view.py
@@ -323,6 +323,9 @@ class SessionView:
                 all_ops.extend(messages)
         return all_ops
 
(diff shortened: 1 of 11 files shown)


marimo/_server/session/session_view.py
```diff
@@ -323,6 +323,9 @@ def operations(self) -> list[MessageOperation]:
                 all_ops.extend(messages)
         return all_ops
 
+    def is_empty(self) -> bool:
+        return len(self.cell_operations) == 0
```
dmadisetti: I think outputs might be better like you said

PR #5686, opened 2025-07-18
Prevent accidental overwrites and converter for non-marimo Python scripts

manzt:
Previously, `marimo edit script.py` would silently overwrite any Python file, potentially destroying user code ([link](https://discord.com/channels/1059888774789730424/1059891311190229082/1395424062339022941)).

#5677 introduced a prompt asking users to confirm before overwriting invalid files. This PR takes a stronger stance by making `marimo edit` **error for non-marimo Python files** and providing a conversion path.

It introduces two key changes:

1. **Protection against accidental overwrites**: `marimo edit` now errors when attempting to open a non-marimo Python file, rather than prompting for confirmation. Users are directed to use `marimo convert` instead.

2. **Python script conversion support**: `marimo convert script.py` now supports converting regular Python scripts to marimo notebooks. The converter:
   - Detects and converts py:percent format if used (requires jupytext), otherwise uses a single marimo cell
   - Transforms `if __name__ == "__main__":` blocks into `_main_()` functions to maintain marimo's variable scoping
   - Preserves file headers (docstrings/comments)

Importantly the notebook pipeline reuses the ipynb converter transformations for multiple cells.

Running `marimo edit regular_script.py` now shows:

```sh
Error: Python script not recognized as a marimo notebook.

  Tip: Try converting with

    marimo convert regular_script.py -o regular_script_nb.py

  then open with marimo edit regular_script_nb.py
```


dmadisetti:
Oh no, potentially closed by https://github.com/marimo-team/marimo/pull/5677

dmadisetti:
There's actually a fair bit of other logic in here. Will do a review on merge res if you want to keep it open

manzt:
if it makes the review easier, i'm happy to break into two separate PRs:

- one that checks and errors on `marimo edit unknown_py_script.py`
- one that add `marimo convert` and updates the error message from above with a Tip

manzt:
Most of the logic is in the second commit for adding a new converter. The error is just adding another violation that we error on inside marimo edit.

dmadisetti:
I think splitting these out make sense, will make the merge res easier

willingc:
Thanks @manzt @dmadisetti. Simple is fine by me. It was a bit of a surprise when it happened. As an FYI, I could see folks wanting to be able to bulk convert a directory of `.py` files to marimo notebooks, such as when creating examples for docs, which is how I discovered the behavior.

dmadisetti:
Thanks @willingc, hope you didn't lose any important files 😬 
The bulk util does make sense, a general pattern matching, or repeating arg might be reasonable

mscolnick:
calling it `unknown python` script feels a bit weird to me. especially if we introduce `marimo convert pypercent`, this will conflict with this code path. and in future, if/when we do add `marimo convert pypercent`, we can treat these "unknown scripts" as just a pypercent with 1 cell.

willingc:
@dmadisetti I didn't lose anything since everything is under version control.


I agree with @mscolnick "unknown python script" is odd phrasing to a user. If everything is a `.py` file, it should behave like a `.py` file whether or not it is a marimo notebook.

Ultimately, I would love to see (some of which you already have):
- A regular `.py` file (code or script) opened in a marimo notebook (going to call it `.mpy`) without data loss
- A regular `.py` file (code or script) converted to a marimo notebook (`.mpy`)
- A marimo notebook (`.mpy`) stripped and converted to a regular `.py` file without marimo code included (IMHO, cleaner for using in a pipeline)
- A Jupyter notebook (`.ipynb`) converted to a marimo notebook (`.mpy`) with cells similar to the initial Jupyter notebook
- A marimo notebook (`.mpy`) converted to a Jupyter notebook (`.ipynb`)
- A marimo notebook exportable as a PDF
- Bonus points: being able to do that using pandoc :wink:

marimo/_ast/load.py
```diff
@@ -8,7 +8,11 @@
 
 from marimo import _loggers
 from marimo._ast.app import App, InternalApp
-from marimo._ast.parse import MarimoFileError, parse_notebook
+from marimo._ast.parse import (
+    MarimoFileError,
+    is_unknown_python_script,
```
willingc: Perhaps `is_python_file` as it doesn't need to be a script. Ideally it is later checked to make sure that it contains no marimo code.

willingc: Longer but better `is_python_file_without_marimo`

marimo/_ast/load.py
```diff
@@ -85,6 +94,7 @@ def notebook_is_openable(filename: str) -> Literal[True]:
 
     Raises:
         SyntaxError: If the file contains a syntax error
+        UnknownPythonScriptError: If the file is an unrecognized Python script format
```
willingc: Perhaps `PythonFileWithoutMarimoError`

marimo/_ast/load.py
willingc: My 2 cents but it feels a bit diminishing to label a standard python file as unknown. Python files are a superset of marimo files which are Python files that contain marimo code.

willingc (commented):
I love the effort here. I would be fine with a simple warning that the `.py` file does not include marimo notebook code. See this link on creating or converting... would be fine short term.

tests/_ast/codegen_data/test_get_codes_non_marimo_python_script.py
```diff
@@ -0,0 +1,11 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "altair==5.4.1",
+#     "duckdb==1.1.3",
+#     "marimo",
+# ]
+# ///
```
mscolnick, line 8: i am curious if we could write a test. is this valid? should it be? ive started notebook like this before (nothing but inline-metadata) and it got overwritten

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "marimo",
# ]
# ///
```

marimo/_convert/non_marimo_python_script.py
```diff
@@ -0,0 +1,56 @@
...
+            raise ImportError(
+                "Converting py:percent format requires jupytext"
+            ) from e
+        notebook = jupytext.reads(source, fmt="py:percent")
+    else:
+        notebook = {"cells": [{"source": source, "cell_type": "code"}]}
+
+    notebook_str = json.dumps(notebook)
+
+    ir = convert_from_ipynb_to_notebook_ir(notebook_str)
```
mscolnick, line 36: this would require `nbformat` is cases where you probably don't really need it.

it might be easier to just create the `ir` in the `else` block

tests/_ast/test_load.py
```diff
@@ -216,6 +217,13 @@ def test_get_codes_app_with_only_comments(load_app) -> None:
         app = load_app(get_filepath("test_app_with_only_comments"))
         assert app is None
 
+    @staticmethod
+    def test_get_codes_non_marimo_python_script(static_load) -> None:
+        with pytest.raises(MarimoFileError, match="is not a marimo notebook."):
+            static_load(
+                get_filepath("test_get_codes_non_marimo_python_script")
+            )
```
manzt, line 225: @mscolnick, this is the test... it's expected to raise an exception (and uv run marimo edit) will print out the conversion message.

marimo/_convert/non_marimo_python_script.py
manzt: > this would require nbformat is cases where you probably don't really need it.

I don't think it does? the file appears to be just using `json.loads` I think we only need `nbformat` for export.

manzt: https://github.com/marimo-team/marimo/blob/bd8d1315c3e00c89cbdb974a14a3a369c49e4c84/marimo/_convert/ipynb.py#L748-L790

mscolnick: ohh nice, good call

manzt:
Thanks for the feedback and reviews everyone. I've updated the PR description with the latest changes, and avoided the use of "unknown" python script.

marimo/_cli/convert/commands.py
```diff
@@ -41,27 +51,63 @@ def convert(
...
+    After conversion:
 
         marimo edit your_nb.py
 
-    Since marimo is different from traditional notebooks, once in the editor,
-    you may need to fix errors like multiple definition errors or cycle
-    errors.
+    Note:
+    Since marimo differs from traditional notebooks, expect to fix issues like
+    multiple definitions or cyclic dependencies after conversion.
```
akshayka: We may not need this anymore, we fix multiple definitions for users, and cyclic dependencies are rare.

The thing to "fix" or eliminate is mutation (mutating a dataframe across multiple cells, for example, which is common in Jupyter but will lead to strange behavior in marimo). Unsure whether to document here but just calling it out.

marimo/_ast/parse.py
```diff
@@ -838,6 +838,17 @@ def parse_notebook(contents: str) -> Optional[NotebookSerialization]:
                 lineno=1,
             )
         )
+
+        remaining = parser.extractor.contents[len(header.value) :]
+        if len(remaining.strip()) != 0:
```
akshayka: Nit: idiomatic Python uses the fact that empty sequences are false. This comes from PEP-8.

image

```suggestion
        if remaining.strip():
```

Your version is not "wrong" but I think Python developers just find it easier to read idiomatic Python.

marimo/_cli/convert/commands.py
```diff
@@ -28,10 +28,20 @@ def convert(
...
     output: Optional[Path],
 ) -> None:
-    r"""Convert a Jupyter notebook or Markdown file to a marimo notebook.
+    r"""Convert a Jupyter notebook, Markdown file, or Python script to a marimo notebook.
 
-    The argument may be either a path to a local .ipynb/.md file,
-    or an .ipynb/.md file hosted on GitHub.
+    Supported input formats:
+    - `.ipynb` (local or GitHub-hosted)
+    - `.md` files with `{python}` code fences
```
akshayka, line 35: Maybe unrelated. So feel free to ignore. But should we give an example of what a `{python}` fenced code block looks like?

marimo/_cli/convert/commands.py
```diff
@@ -41,27 +51,63 @@ def convert(
...
+                and getattr(e.__cause__, "name", None) == "jupytext"
+            ):
+                from marimo._cli.print import green
+
+                raise click.ClickException(
+                    f"{e}\n\n"
+                    f"  {green('Tip:')} If you're using uv, run:\n\n"
+                    f"    uvx --with=jupytext marimo convert {filename}"
+                ) from e
+            raise
```
dmadisetti, line 119: Checking the name just feels a little brittle, maybe better to check explictly

```suggestion
        try:
            # Or use dependency manager
            import jupytext
        except ImportError as e:
            from marimo._cli.print import green

            raise click.ClickException(
                    f"{e}\n\n"
                    f"  {green('Tip:')} If you're using uv, run:\n\n"
                    f"    uvx --with=jupytext marimo convert {filename}"
            ) from e

        notebook = MarimoConvert.from_non_marimo_python_script(
                text
            ).to_py()
```

dmadisetti: Oh, I see the import is conditional. Eh, this is fine then

tests/_convert/test_convert_non_marimo_python_script.py
```diff
@@ -0,0 +1,100 @@
...
+        snapshot_test("minimal_script.py.txt", converted)
+
+    @pytest.mark.skipif(not HAS_JUPYTEXT, reason="jupytext not installed")
+    def test_pypercent_format(self) -> None:
+        """Test conversion of pypercent format file."""
+        source = '''"""Pypercent format notebook."""
+
+import numpy as np
+
+# %% [markdown]
```
dmadisetti, line 50: pypercent also supports

```python
# %% [markdown]
"""This is a doc string, but also markdown"""
```

Be nice to confirm that works as well

marimo/_ast/parse.py
manzt: TIL, thanks for teaching me!

tests/_convert/test_convert_non_marimo_python_script.py
manzt: all py:percent support comes from whatever is supported by jupytext. I'll add a test, but if it doesn't work I think it's out of scope for the PR to make it work.

marimo/_cli/convert/commands.py
manzt: yeah, we purposely throw a nested import error. I could throw something more specific.

manzt:
> I might followup to make confirm non-blocking in non-tty contexts.

I don't think these changes introduce any new interactive code paths, unless I'm misunderstanding. Are you referring to improving the existing overwrite confirmation to be non-blocking in non-TTY contexts?

tests/_convert/test_convert_non_marimo_python_script.py
manzt: Nice! looks like jupytext supports it.

dmadisetti:
> I don't think these changes introduce any new interactive code paths, unless I'm misunderstanding. Are you referring to improving the existing overwrite confirmation to be non-blocking in non-TTY contexts?

Yes, related to this PR, but yes nothing this code explicitly touches

mscolnick (dismissed):

marimo/_cli/cli.py
```diff
@@ -71,6 +71,19 @@ def check_app_correctness(filename: str) -> None:
...
+        raise click.ClickException(
+            f"Python script not recognized as a marimo notebook.\n\n"
+            f"  {green('Tip:')} Try converting with"
+            "\n\n"
+            f"    marimo convert {filename} -o {stem}_nb.py\n\n"
+            f"  then open with marimo edit {stem}_nb.py"
+        ) from None
+
     # Only show the tip if we're in an interactive terminal
     if status == "invalid" and sys.stdin.isatty():
```
dmadisetti, line 88: Thanks for the rebase.
This might be dead code now? Either way the ideal behavior for molab is:

`if not tty and invalid: continue # as normal (regardless of .py ending)`

If this is too dangerous, we can hide behind a another flag. An auto convert flag could be reasonable

`if invalid and autoconvert: convert and continue`

manzt: The above condition is:

```py
if status == "invalid" and filename.endswith(".py"):
```

I assume non-Python files could also result in an `invalid` notebook status, making this code not dead in that context? I don't know the molab context well enough and trust whatever you think.

PR #5688, opened 2025-07-18
Auto sync deps before running pnpm run/exec

manzt:
TIL: https://pnpm.io/next/settings#verifydepsbeforerun

This setting allows the checking of the state of dependencies before
running scripts. The check runs on `pnpm run` and `pnpm exec` commands,
automatically running install if `node_modules` is not up to date.

Means you can also just clone marimo and run `pnpm build` without install step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,6 @@ help:
 # 🚀 First-time setup: Install all dependencies (frontend & Python)
 install-all: fe py
 
-.PHONY: pnpm-install
-pnpm-install:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -20,11 +20,6 @@ help:
 # 🚀 First-time setup: Install all dependencies (frontend & Python)
 install-all: fe py
 
-.PHONY: pnpm-install
-pnpm-install:
-	@command -v pnpm >/dev/null 2>&1 || { echo "pnpm is required. See https://pnpm.io/installation"; exit 1; }
```
manzt, line 25: We already have this bail/check in install-all

PR #5690, opened 2025-07-18
Properly focus (and unfocus) multi-cell delete confirmation

manzt:
We didn't support focusing the multi-cell deletion confirmation before
with `FocusScope` because `autoFocus` triggered the cells' blur
handlers, which in turn cleared the pending delete state. When
attempting to focus the confirmation UI, blur events from the selected
cells would clear the pending delete service before the dialog could
render.

The fix removes `pendingDeleteService.clear()` from the cell's
`onBlurWithin` handler to prevent premature clearing and wraps the
confirmation buttons in `FocusScope` with `restoreFocus` to
automatically return focus to the previously focused element when the
dialog is dismissed.

([react-aria](https://react-spectrum.adobe.com/react-aria/index.html) / `FocusScope` is really nice quality of life improvement)

### DIFF
--- a/frontend/src/components/editor/navigation/multi-cell-action-toolbar.tsx
+++ b/frontend/src/components/editor/navigation/multi-cell-action-toolbar.tsx
@@ -19,6 +19,7 @@ import {
   ZapOffIcon,
 } from "lucide-react";
 import React from "react";
+import { FocusScope } from "react-aria";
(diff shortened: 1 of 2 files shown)


Light2Dark (approved):
nice!

manzt:
rebased, should be good to go

PR #5692, opened 2025-07-19
fix: fetch field types for lazy tables

Light2Dark:
## 📝 Summary

Previously, we did not fetch field_types for lazy tables, which prevented dtypes from being visible. Fetching dtypes shouldn't be expensive, we use `collect_schema` to fetch for lazy df's.

## 📋 Checklist

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -41,7 +41,6 @@ from marimo._plugins.ui._impl.tables.selection import (
 )
 from marimo._plugins.ui._impl.tables.table_manager import (
     ColumnName,
-    FieldTypes,
     RowId,
     TableCell,
     TableCoordinate,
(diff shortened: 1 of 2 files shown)


mscolnick:
just a heads up, i don't think `collect_schema` is less expensive than `.schema`. it is the same, but they just wanted to make it clear that you were doing something that was not "free" or pre-calculated

Light2Dark:
Digging into this a bit, I think we should not add this feature now.
`collect_schema` / `schema` are potentially expensive operations for lazyframes: https://github.com/pola-rs/polars/issues/16328#issuecomment-2119967467

> This performance regression has significantly impacted my use case. I’ve observed the regression between versions 0.20.21 and 0.20.22 and through to the latest 1.8.2 with the .collect_schema().names() syntax.

> LazyFrame.columns performs much slower from version 0.20.23. In my simple example above, 0.20.23 is nearly 10x slower than 0.20.21.

PR #5705, opened 2025-07-21
More command keywords

koaning:
It felt like these two were the main culprits, at least in my workflow.

### DIFF
--- a/frontend/src/components/editor/actions/useNotebookActions.tsx
+++ b/frontend/src/components/editor/actions/useNotebookActions.tsx
@@ -289,7 +289,7 @@ export function useNotebookActions() {
     },
     {
       icon: <TextSearchIcon size={14} strokeWidth={1.5} />,
-      label: "View live docs",
+      label: "Helper panel > View live docs",
       description: "Show the documentation helper panel",
       handle: () => toggleApplication("documentation"),
     },
--- a/frontend/src/components/editor/actions/useNotebookActions.tsx
+++ b/frontend/src/components/editor/actions/useNotebookActions.tsx
@@ -35,6 +35,7 @@ import {
   PresentationIcon,
   SettingsIcon,
   Share2Icon,
+  TextSearchIcon,
   Undo2Icon,
   XCircleIcon,
   YoutubeIcon,
@@ -285,6 +286,12 @@ export function useNotebookActions() {
           handle: () => toggleApplication(type),
         };
       }),
+    },
+    {
+      icon: <TextSearchIcon size={14} strokeWidth={1.5} />,
+      label: "View live docs",
+      description: "Show the documentation helper panel",
+      handle: () => toggleApplication("documentation"),
     },
 
     {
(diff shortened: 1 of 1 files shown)


Light2Dark:
Using the tooltip looks okay to me imo and is more descriptive in the command pallete. We shouldn't add another command when it already exists.

CleanShot 2025-07-21 at 22 16 17@2x

Another option is to add an optional label param and add it for just these 2 panels.

manzt:
I agree with @Light2Dark, think the tooltip would be fine to use as a label. I think we have more freedom to change these kind of "on-demand" behaviors in the UI.

May I suggest we use a more descriptive PR title like "Add additional command palette actions for documentation and AI chat panels"? The PR titles are copied into release notes so it would be useful to help discoverability by end users of these new additions.

Light2Dark (approved):

mscolnick (changes requested):
i think we can add something call `extraDescription` that is just used for searching. using `tooltip` as the label will may lead to some other side effects

koaning:
I added a key `extraDescription` per comment from @mscolnick that also lets us add extra wording. In the case of the documentation panel, just to name one example, we're also able to refer to terms that people may remember from Jupyter like "contextual helper". It's the same feature, but in Juypter it has another name.

frontend/src/components/editor/actions/useNotebookActions.tsx
```diff
@@ -274,16 +274,19 @@ export function useNotebookActions() {
       icon: <PanelLeftIcon size={14} strokeWidth={1.5} />,
       label: "Helper panel",
       handle: NOOP_HANDLER,
-      dropdown: PANELS.flatMap(({ type, Icon, hidden }) => {
+      dropdown: PANELS.flatMap(({ type, Icon, hidden, extraDescription }) => {
```
mscolnick, line 277: this is going to create a lot of dropdowns (which we don't want). we should instead fire through a custom `filter` prop to in `frontend/src/components/editor/controls/command-palette.tsx` to `<CommandDialog` that uses these extra descriptions

koaning: I feel inclined to push back a little bit, but this might just be a lack of typescript/react knowledge on my part, so feel free to push back too. 

My concern is that if a user types "contextual helper" and they see "documentation" that they may experience that as a "oh, i guess this feature is not here". That's the thing I really want to prevent. I think the only way to get there is to have extra dropdown items because if we just show the "documentation" element without any matching terms the user may still get confused.

mscolnick: the problem is, this is what your change is doing

image

koaning: Ah! Now I understand!

mscolnick:
Closed by https://github.com/marimo-team/marimo/pull/8600

PR #5706, opened 2025-07-21
An upgrade for VIM users

koaning:
I am learning vim-motions this week and I really wanted to do that in marimo. f

Figured I might make it easier for myself... 

CleanShot 2025-07-21 at 14 33 50


koaning:
We need to read from `vimrc` if you want to do this properly and that makes this a bigger PR. It also means that it's probably wiser to wait and see if the community demands this because there are plenty of other vim motions that you could use that don't require the line number diff. 

Am closing this for now.

PR #5707, opened 2025-07-21
Add reactive notebook minimap for navigating and visualizing dataflow

manzt:
This PR introduces a reactive notebook minimap (similar in spirit to [Observable](https://observablehq.com/documentation/debugging/minimap)), but adapted marimo where multiple variables can be declared across cells. Selecting a cell highlights its upstream and downstream connections, visualizing data flow through the notebook.

For now, it's rendered as a floating window for testing. The current goal is to refine the UX and determine its final placement in the UI. I'd like to avoid burying it in the left panel and instead possibly have a floating window triggered either by button or keyboard shortcut.

Some nuance with Observable's minimap: Since our cells can declare multiple variables, our dataflow tracing is asymmetric. For upstream variabes, we can trace granularly which variables used by the selected cell and find where they were declared. For downstream tracing, we coarsely mark all variables that are in a dependent cell as connected (since they are all dependents).

TODO:
- [x] ~Handle unselected state. Right now, the minimap always shows a focused cell. We should have a state for when nothing is focused.~ This state is handled, but never shows up because `lastFocusedCell` is always defined. Would be nice to have some way to select/deselect items in the minimap. Wondering how it should play with command/edit modes.
- [x] ~Fix click selection behavior. We're currently using `actions.focusCell({ cellId, before: false })`, which causes the next/previous cell to be focused depending on cell state. Instead, we want to focus _exactly_ the clicked cell regardless of position.~ Updated `actions.focusCell({ cellId, mode: "before" | "after" | "exact" })`.
- [ ] Documentation. We can lean on aligning with observable, but this feature should be well documented with _how_ to actually read the minimap. It takes a little adjustment, but once you grok it it's a very efficient encoding of the information.

https://github.com/user-attachments/assets/214b3d9e-daa4-4e2e-9970-f3517cb9b455

### DIFF
--- a/frontend/src/components/editor/chrome/wrapper/minimap.tsx
+++ b/frontend/src/components/editor/chrome/wrapper/minimap.tsx
@@ -1,13 +1,16 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
 import { useAtomValue } from "jotai";
-import React from "react";
+import { XIcon } from "lucide-react";
+import * as React from "react";
+import { Button } from "@/components/ui/button";
 import { useCellActions, useNotebook } from "@/core/cells/cells";
 import { cellFocusAtom, useCellFocusActions } from "@/core/cells/focus";
 import type { CellId } from "@/core/cells/ids";
 import { useVariables } from "@/core/variables/state";
 import type { VariableName } from "@/core/variables/types";
 import { cn } from "@/utils/cn";
+import { useChromeActions, useChromeState } from "../state";
 import {
   type CellGraph,
   cellGraphsAtom,
@@ -122,13 +125,14 @@ const MinimapCell: React.FC<MinimapCellProps> = (props) => {
   );
 };
 
-export const Minimap: React.FC<{ className?: string }> = ({ className }) => {
+const MinimapInternal: React.FC<{
+  open: boolean;
+  setOpen: (update: boolean) => void;
+}> = ({ open, setOpen }) => {
   const notebook = useNotebook();
-
   const cellPositions: Record<CellId, number> = Object.fromEntries(
     notebook.cellIds.inOrderIds.map((id, idx) => [id, idx]),
   );
-
   const columnBoundaries: number[] = [];
   let cellCount = 0;
   for (const [idx, column] of notebook.cellIds.getColumns().entries()) {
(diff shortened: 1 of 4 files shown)


frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
```diff
@@ -244,6 +245,7 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
         </Panel>
         <ContextAwarePanel />
       </PanelGroup>
+      <FloatingMinimap />
```
manzt: Need to move/change before merging

dmadisetti:
What do the dot sizes mean? (Some are big, some are small)

frontend/src/components/editor/chrome/wrapper/floating-minimap.tsx
```diff
@@ -0,0 +1,285 @@
...
+  );
+};
+
+export const FloatingMinimap: React.FC<{ className?: string }> = ({
+  className,
+}) => {
+  const notebook = useNotebook();
+  const actions = useCellActions();
+  const handleCellClick = (cellId: CellId) => {
+    actions.focusCell({ cellId, before: false });
```
manzt: need to have as autofocus for the exact cell not before or after

manzt:
>  What do the dot sizes mean? (Some are big, some are small)

I followed this description of the feature from the Observable docs as closely as I could: https://observablehq.com/documentation/debugging/minimap#connection-types

image

small dot == independent, not in dependency tree. large dot == in dependency tree

frontend/src/components/editor/chrome/wrapper/minimap.tsx
```diff
@@ -0,0 +1,338 @@
...
+      data-node-id={cellId}
+      className={cn(
+        "group bg-transparent text-left w-full flex relative justify-between items-center",
+        "bg-white border-none rounded cursor-pointer",
+        "h-[21px] pl-[51px] font-inherit",
+        isSelected
+          ? "text-white"
+          : runtime.errored
+            ? "text-red-400"
+            : "text-gray-300",
```
mscolnick: How does this look in dark mode? If you use radix colors, dark mode is a bit more free

frontend/src/components/editor/chrome/wrapper/minimap.tsx
```diff
@@ -0,0 +1,338 @@
...
+            />
+          );
+        })}
+      </div>
+    </div>
+  );
+};
+
+function codePreview(code: string): string | undefined {
+  return code.split("\n")[0].trim() || undefined;
```
mscolnick, line 195: Would be nice to trim sql and markdown. Or even color it differently

frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -186,13 +186,13 @@ export function useCellNavigationProps(
         },
         // Move up
         ArrowUp: () => {
-          actions.focusCell({ cellId, before: true });
+          actions.focusCell({ cellId, mode: "before" });
```
mscolnick: nit: maybe "where" is better than "mode".

akshayka:
This looks amazing!

manzt:
I added a very subtle indicator for multicolumn:

image

frontend/src/components/editor/chrome/wrapper/minimap.tsx
manzt: I like the idea of trimming down markdown/sql, but maybe something we could explore outside this PR? I tried a few different designs and didn't really come up with something I like. I worry color could get a little noisy.

manzt:
I'm happy where this is at. I added a launcher to the bottom. I'm wondering if a hot key for toggling (ctrl + shift + m) could be preferable to the command pallete. I was experimenting with it locally but we could add in a second PR.

mscolnick (approved):

PR #5719, opened 2025-07-22
Fix stale variable dependencies in frontend

manzt:
Found while implementing #5707 

The frontend wasn't updating when a cell's variable dependencies changed
without affecting the dataflow graph structure. For example, changing and executing a cell from:

```py
print(a)
print(b)
```

to

```py
print(a)
```

the frontend would still show the old dependency information.

The issue occurred because the backend only sent variable updates when
cell-to-cell edges changed or when variables were created/deleted. But a cell's
`used_by` information _can_ change even when the graph edge remains intact. The
cell still imports from the same parent, just using different variables.

This change ensures the frontend always receives the latest variable dependency
information by broadcasting `Variables` updates after any graph mutation. An
alternative would be to track more granular changes to variable usage (like
before), but I think that kind of optimization is why this bug popped up in the
first place  it seems safest to always emit these updates and special case if
it's a performance concern.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -13,7 +13,7 @@ import sys
 import threading
 import time
 import traceback
-from copy import copy, deepcopy
+from copy import copy
 from functools import cached_property
 from multiprocessing import connection
 from pathlib import Path
@@ -1113,8 +1113,6 @@ class Kernel:
         LOGGER.debug("Current set of errors: %s", self.errors)
         cells_before_mutation = set(self.graph.cells.keys())
         cells_with_errors_before_mutation = set(self.errors.keys())
-        edges_before = deepcopy(self.graph.children)
-        definitions_before = set(self.graph.definitions.keys())
         cells_starting_stale = (
             set() if cells_starting_stale is None else cells_starting_stale
         )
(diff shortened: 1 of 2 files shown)


manzt:
MRE: Open the side panel dependency viewer:

```py
a = 1
b = 2
```

```py
print(a, b)
```

then changing cell 2 to:

```py
print(a)
```

and after running, it will still say cell-2 depends on a AND b.

mscolnick (commented):
could we capture this in a test?

manzt:
> could we capture this in a test?

done, added (and updated previous)

marimo/_runtime/runtime.py
```diff
@@ -1271,27 +1269,22 @@ def mutate_graph(
...
-                        used_by=list(
-                            self.graph.get_referring_cells(
-                                variable, language="python"
-                            )
-                        ),
-                    )
-                    for variable, declared_by in self.graph.definitions.items()
-                ]
-            ).broadcast()
+        # Always broadcast Variables message after graph mutation to ensure
```
dmadisetti, line 1272: This feels expensive, although I'm certain it probably isn't. We do determine references on a per variable level- so in theory we could capture the more granular cases

dmadisetti (approved):
LGTM. Maybe room for improvement, but I would bet that it's marginal (and definitely more bug prone)

PR #5734, opened 2025-07-23
Remove global `hatch` CLI as required development dependency

manzt:
Simplifies the development toolchain by removing the requirement to install `hatch` separately. Devs are still free to install hatch CLI directly (`uv tool install hatch`), but not required.

All hatch commands can now run through `uvx hatch`, and building uses `uv build` directly. This reduces the number of tools required for developers install.

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -44,8 +44,8 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-      - name: 🥚 Install Hatch
-        uses: pypa/hatch@install
+      - name: 🐍 Setup uv
+        uses: astral-sh/setup-uv@v6
 
       # patch __init__.py version to be of the form
       # X.Y.<Z+1>-dev{n-commits-since-last-tag}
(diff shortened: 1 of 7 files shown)


mscolnick (approved):
this if fine for ci, but i find it more inconvenient to type 2 things (uv and hatch) instead of one when i need to run tests.

also if we need to pin a version of hatch, this could make it slightly more difficult

mscolnick:
> This reduces the number of tools developers need to install

This does not need to be the goal. You type commands more than you install

manzt:
To be clear, you can still `uv tool install hatch`, to have a global CLI, it's just for CI and the Makefile that we use `uvx hatch` instead.

manzt:
There is no version of hatch currently pinned. If we wanted to pin a version of hatch, we could add to the project dependency group:

```sh
uv add --dev hatch
```

and then `uv run hatch` would use whatever is in the `uv.lock` file.

or update the `uvx` invocations to use a specific version with something like [ruplacer](https://github.com/your-tools/ruplacer):

```sh
ruplacer "uvx hatch" "uvx hatch@version"
```

manzt:
I updated the description & title to make this clear.

PR #5739, opened 2025-07-24
Add text wrapping toggle for console output

### LINKED ISSUE #4968: Option to wrap text in the output
### Description

I am working as a data journalist, I just discovered marimo. Me any my colleagues love it and are considering switching to it entirely. However, one issue:

We do a lot of automated/machine assisted text analysis. For that it is nessecary to frequently have a look at the text you're working with. Since marimo does not wrap the output this is highly impractical:

Printing a single doc in marimo

### Suggested solution

Coming from jupyter (which has many other disadvantages for our use cases which is why we are considering marimo) I expect something like this:

Printing in Jupyter

The output being wrapped to the viewwidth.

Since this is also not an ideal behavior all the time: **an option to toggle an output wrap on or off**

### Will you submit a PR?

- [ ] Yes

### Alternative

I thought about using marimo's markdown abilities to solve this:

Using mo.md to display a single doc

However, this has the disadvantage that it will then not be displayed when running the .py file just via the terminal. Additionally, it is not appllicable for code like this as far as I see it:

Printing multiple docs in a loop in marimo

### Additional context

_No response_

manzt:
Fixes #4968

Adds a button to toggle text wrapping in console outputs, making it easier to read and analyze text without horizontal scrolling.

The toggle appears on hover in the top-right corner of console outputs. Its state is currently persisted in localStorage, making it consistent across cells (and sessions). Alternatively, we could:

- Add a marimo-wide configuration option, or
- Not persist anything and have each output control it's wrapping state

https://github.com/user-attachments/assets/96aa6ccd-96ed-4ece-8d4d-2d9f570946ea

### DIFF
--- a/frontend/src/components/editor/Output.tsx
+++ b/frontend/src/components/editor/Output.tsx
@@ -47,8 +47,9 @@ export const OutputRenderer: React.FC<{
   message: Pick<OutputMessage, "channel" | "data" | "mimetype">;
   cellId?: CellId;
   onRefactorWithAI?: (opts: { prompt: string }) => void;
+  wrapText?: boolean;
 }> = memo((props) => {
-  const { message, onRefactorWithAI, cellId } = props;
+  const { message, onRefactorWithAI, cellId, wrapText } = props;
   const { theme } = useTheme();
 
   // Memoize parsing the json data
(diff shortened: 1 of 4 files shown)


mscolnick (commented):
this is exactly how I was thinking to store the state: global and in local storage. I think that is a good default

Light2Dark (approved):

mscolnick (approved):

PR #5746, opened 2025-07-25
tests: fix broken tests on main

mscolnick:


manzt (dismissed):
thank you

tests/_cli/test_cli_export.py
```diff
@@ -922,8 +923,8 @@ def test_export_ipynb_with_multiple_definitions(
         assert p.stdout.decode() == ""
 
     @pytest.mark.skipif(
-        not DependencyManager.nbformat.has(),
-        reason="This test requires nbformat.",
+        not DependencyManager.nbformat.has() or is_windows(),
```
dmadisetti, line 926: I think this is an actual error, but just cosmetic so not worried

dmadisetti (approved):

PR #5758, opened 2025-07-25
tests: cache invaliadation for external decorators

dmadisetti:
## 📝 Summary

External functions are assumed to be pure. If an external function wraps a notebook function, then our test was unable to pick up the notebook change. This PR adds test to catch this behavior, and adds an appropriate ix (explicitly checking to see if the external function is used as a wrapper)


mscolnick (approved):

dmadisetti:
@akshayka

akshayka (approved):

PR #5759, opened 2025-07-26
Fix cell movement shortcuts on non-Mac keyboards

manzt:
Cell movement shortcuts (`Ctrl+Shift+7/8/9/0`) weren't working on Windows and Linux because these keyboards produce shifted characters like "&" instead of "7" when Shift is held. The shortcut parser was comparing against the literal digit, causing matches to fail.

This fix uses the physical key code for shifted digit combinations, making the shortcuts work consistently across keyboard layouts. Note: `Ctrl+Shift+0` may still be intercepted at the system-level Windows.

### DIFF
--- a/frontend/src/components/editor/navigation/__tests__/navigation.test.ts
+++ b/frontend/src/components/editor/navigation/__tests__/navigation.test.ts
@@ -645,6 +645,7 @@ describe("useCellNavigationProps", () => {
 
       const mockEvent = Mocks.keyboardEvent({
         key: "9",
+        code: "Digit9",
         ctrlKey: true,
         shiftKey: true,
       });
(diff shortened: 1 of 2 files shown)


manzt:
@akshayka would you mind trying out on linux?

akshayka:
Somehow moving `Ctrl+Shift+7/8` are still not working for me on Linux. I'm not sure why

mscolnick (approved):
nice fix!

mscolnick (changes requested):
whoops seeing the comment it still doesn't work

akshayka:
I noticed this now works in command mode on my linux machine. Just not "edit" mode. Remapping say move left to other key combinations (such as ctrl + shift + X) still doesn't fix it

Regardless since it's working in command mode at least, this is an improvement that I'm fine with merging.

manzt:
Edit mode supported in:

- #5808

Thanks, @mscolnick !

mscolnick (approved):

PR #5760, opened 2025-07-26
Support keyboard navigation across columns in command mode

manzt:
Adds support for navigating between columns using left/right arrow keys in command mode. For vim users, `h` and `l` also work to move left and right between columns.

### DIFF
--- a/frontend/src/components/editor/navigation/navigation.ts
+++ b/frontend/src/components/editor/navigation/navigation.ts
@@ -195,6 +195,43 @@ export function useCellNavigationProps(
           selectionActions.clear();
           return true;
         },
+        // Move left across columns
+        ArrowLeft: () => {
+          if (canMoveX) {
+            const notebook = store.get(notebookAtom);
+            const column = notebook.cellIds.findWithId(cellId);
+            const columnIndex = notebook.cellIds.indexOf(column);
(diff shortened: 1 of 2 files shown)


Light2Dark (approved):

manzt:
@Light2Dark would you mind trying out the latest changes? I took a stab a making the navigation size aware.

https://github.com/user-attachments/assets/744e24c2-a4c7-4d07-b9ed-8f29103a1237

frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -576,3 +616,33 @@ export function useCellEditorNavigationProps(cellId: CellId) {
...
+  const current = document.getElementById(HTMLCellId.create(currentCellId));
+
+  if (!current) {
+    // fallback to first
+    return adjacentColumn.first();
+  }
+
+  const currentRect = current.getBoundingClientRect();
+  // first to to either overlap or contain the current cell's top edge
+  for (const candidateId of adjacentColumn.inOrderIds) {
```
mscolnick: likely should be `adjacentColumn.topLevelIds` which are just the visible ones

manzt: Good call

Light2Dark:
feels good!

PR #5761, opened 2025-07-26
Support keyboard navigation across columns in command mode

Light2Dark:
Instead of going to first, it finds the closest adjacent cell by index. It's not perfect but imo slightly better than going to first, especially for long column notebooks.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/editor/navigation/__tests__/navigation.test.ts
+++ b/frontend/src/components/editor/navigation/__tests__/navigation.test.ts
@@ -121,7 +121,7 @@ const setupSelection = () => {
   return selectionTesting.createActions(dispatch);
 };
 
-const [cellId1, cellId2, cellId3] = MockNotebook.cellIds();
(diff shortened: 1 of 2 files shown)


mscolnick (approved):

manzt:
Sweet, thank you! I wonder if this could potentially be a bit more disorienting when cells become differently sized, so switching from cell 3 in column 1 -> cell 3 in column 2 would just jump to a different spot. I might play around with the nearest function to see if we can try to find what cell is adjacent.

Light2Dark:
> Sweet, thank you! I wonder if this could potentially be a bit more disorienting when cells become differently sized, so switching from cell 3 in column 1 -> cell 3 in column 2 would just jump to a different spot. I might play around with the nearest function to see if we can try to find what cell is adjacent.

Yeah, happy to revert if not. I wasn't sure if there was an easy way to find the height (maybe length of code, outputs). Jump to first is fine too imo. Better than nothing

PR #5765, opened 2025-07-26
improvement: add copy button in console, and fix ctrl-a to copy all

mscolnick:


frontend/src/components/editor/output/ConsoleOutput.tsx
```diff
@@ -110,23 +113,43 @@ const ConsoleOutputInternal = (props: Props): React.ReactNode => {
...
-          {wrapText ? "Disable wrap text" : "Wrap text"}
+      <div className="absolute top-1 right-5 z-10 opacity-0 group-hover:opacity-100 flex gap-1">
+        <Tooltip content="Copy all">
+          <span>
+            <button
+              aria-label="Copy all console output"
+              className="p-1 rounded bg-transparent text-muted-foreground hover:text-foreground"
+              type="button"
+              onClick={() => {
+                const text = reversedOutputs
```
Light2Dark, line 124: I think using `consoleOutputs` rather than reversed feels more natural for the user 🤔

mscolnick: most recent log is last, so reversed it better:

print("1")
print("2")
print("3")
print("4")

will copy: `1, 2, 3, 4`

Light2Dark: oops, i saw it was copying the debugger outputs first, but without it looks fine.
```python
[34m[D 250726 23:27:31 cell_runner:602](B[m Running post_execution hooks in context
[34m[D 250726 23:27:31 hooks_post_execution:64](B[m Acquiring graph lock to update cell import workspace
[34m[D 250726 23:27:31 hooks_post_execution:66](B[m Acquired graph lock to update import workspace.

1
2
3
4
```

frontend/src/components/editor/output/ConsoleOutput.tsx
```diff
@@ -110,23 +113,43 @@ const ConsoleOutputInternal = (props: Props): React.ReactNode => {
...
-          onChange={setWrapText}
-        >
-          <WrapTextIcon className="h-4 w-4" />
-        </ToggleButton>
-        <Tooltip className="z-50 overflow-hidden rounded-md border bg-popover px-3 py-1.5 text-sm text-popover-foreground shadow-xs data-[side=bottom]:slide-in-from-top-1 data-[side=left]:slide-in-from-right-1 data-[side=right]:slide-in-from-left-1 data-[side=top]:slide-in-from-bottom-1">
-          {wrapText ? "Disable wrap text" : "Wrap text"}
+      <div className="absolute top-1 right-5 z-10 opacity-0 group-hover:opacity-100 flex gap-1">
+        <Tooltip content="Copy all">
+          <span>
+            <button
```
manzt, line 119: Would we want to use `Button` from react aria here?

manzt (approved):

frontend/src/components/editor/output/ConsoleOutput.tsx
mscolnick: not at the moment, my thought is to not directly use `react-aria-components` in app code, and should create wrappers in the `/ui` folder liek we do for radix (and we should move off radix), and consolidate all our buttons and variations

PR #5766, opened 2025-07-26
Restrict `Shift+Delete` in edit mode to empty cells only

manzt:
I optimistically enabled the "destructive delete" keybindings for non-empty cells outside of command mode (i.e., editing), but I've personally found `Shift+Backspace` is too easy to trigger accidentally when typing (I believe @mscolnick has ran into this too). These changes require entering command mode to use the keybinding to delete non-empty cells, matching vim's `dd` behavior in editing mode, and behavior before #5664.


akshayka (approved):
Nice catch, I ran into this too.

PR #5772, opened 2025-07-27
Revamp column UI with kanban-inspired design

manzt:
/sunday-fun

The existing column UI (subjectively) felt a little inconsistent with marimo's design language. I thought I'd give a go... the changes introduces an attempt at a Linear/GitHub-inspired kanban style.


manzt:
One design consideration is that this adds a consistent, very subtle gray background to the columns. One option is to only show that container on hover... I personally like having it always on/consistent since there is a lot that happens already visually on hover, but here's how it looks only on hover:

https://github.com/user-attachments/assets/c7374102-1732-43e4-bd21-526f91e90a28

mscolnick (approved):
i like this change / design improvement. i do prefer the background always being there, instead of on hover.

PR #5775, opened 2025-07-28
fix: dynamic SQL keyword completion based on dialect

mscolnick:


frontend/src/core/codemirror/language/languages/sql.ts
```diff
@@ -221,135 +207,185 @@ export class SQLLanguageAdapter
...
+              // Otherwise nest under database name
+              dbToTablesMap[db.name] = dbToTablesMap[db.name] || {};
+              dbToTablesMap[db.name][table.name] = columns;
             }
           }
         }
+      }
+
+      return {
+        shouldAddLocalTables: false,
```
mscolnick, line 270: @Light2Dark is this the right thing? 

i noticed previously, this early return only included `schema: dbToTablesMap,` while the later return includes `schema: { ...databaseMap, ...schemaMap, ...tablesMap },`

so i am not sure if this was an existing bug or this is what i should correctly return.

the tests currently pass either way if i make `shouldAddLocalTables` true or false

Light2Dark: I think it looks correct to me, since it's a schemaless db, we don't need to pass schemaMap separately. And since we specified defaultSchema, tables can be completed at the top level.

I will test with some db's

Light2Dark (approved):
thanks!

PR #5778, opened 2025-07-28
Switch to uv_build as PEP 517 build-system

manzt:
Don't know if this is actually what we want to do, but was telling @mscolnick the other day and thought I'd share a working branch.

This PR changes our PEP 517 `build-system` from `hatchling` to [`uv_build`](https://docs.astral.sh/uv/concepts/build-backend/#modules). This substantially improves startup time for local dev (i.e., `uv run marimo` and `hatch run test`) and actually building marimo for pypi (i.e., `uv build`). 

```sh
hyperfine --warmup 1 --min-runs 3 \
    --prepare 'jj edit -r y' \
    --command-name "hatchling" \
      'rm -rf dist && uv build' \
    --prepare 'jj edit -r u' \
    --command-name "uv_build" \
      'rm -rf dist && uv build'
Benchmark 1: hatchling
  Time (mean ± σ):     17.452 s ±  0.090 s    [User: 16.436 s, System: 0.989 s]
  Range (min … max):   17.376 s … 17.552 s    3 runs

Benchmark 2: uv_build
  Time (mean ± σ):      3.439 s ±  0.018 s    [User: 2.909 s, System: 0.247 s]
  Range (min … max):    3.418 s …  3.451 s    3 runs

Summary
  uv_build ran
    5.08 ± 0.04 times faster than hatchling
```

The main tradeoff with these changes is that this drops support for `build_hatch.py` hook which integrates building our front end as a part of the build process. Astral team says [plugins and dynamic metadata are not in scope](https://github.com/astral-sh/uv/issues/14561) for their build backend, so I don't think it's a feature that will come soon. This limitations means that we no longer have the ability to send a github URL to build a branch (see section removed from CONTRIBUTING.md).

That said, with my tinkering I've found this to be a much nicer improvement (I rarely way on `uv run marimo`) when testing local changes. (Normally i'd switch tabs or do something else, so it subjectively feels liek a productivity boost).


pyproject.toml
```diff
@@ -305,6 +287,12 @@ dependencies = [
 [tool.uv.sources]
 marimo_docs = { path = "./docs", editable = true }
 
+[tool.uv.build-backend]
+module-name = "marimo"
+module-root = ""
+source-include = ["third_party.txt", "third_party_licenses.txt"]
+wheel-exclude = ["marimo/_smoke_tests/**"]
```
mscolnick, line 294: 1. do we need to includ `"marimo/_static/", "marimo/_lsp/"`?
2. we will need to update `scripts/modify_pyproject_for_marimo_base.sh` to work with this new format. this script strips the assets (except index.html) to reduce the size of the build

manzt: > do we need to includ "marimo/_static/", "marimo/_lsp/"?

these are automatically included

```sh
unzip -l dist/marimo-0.14.13-py3-none-any.whl | grep static | wc -l
     294
 unzip -l dist/marimo-0.14.13-py3-none-any.whl | grep _lsp | wc -l
      59
```

manzt: > we will need to update scripts/modify_pyproject_for_marimo_base.sh to work with this new format. this script strips the assets (except index.html) to reduce the size of the build

I just pushed a new script for this and updated the associated workflows

marimo/__init__.py
```diff
@@ -84,9 +84,8 @@
     "video",
     "vstack",
     "watch",
+    "__version__",
 ]
-__version__ = "0.14.13"
```
dmadisetti, line 88: I suppose dynamic versioning cleans things up, but wouldn't mind some marker in source for reference

manzt: `pyproject.toml` is the canonical, static source of truth and required in both `sdist` and `wheel`. uv build-system doesn't allow for 'dynamic' metadata (like deriving a version from python source) because they think it's an anti-pattern to have dynamic package metadata (see removal of `dynamic` fields). Importing from `importlib.metadata` is reading static metadata from a TOML data file.

mscolnick (approved):

PR #5783, opened 2025-07-28
fix: set base-url if supplied

mscolnick:
If we have a base-url (subpath or remote), we should include it the html as `<base href>`


dmadisetti:
Seems fine, what's the blocker?

mscolnick:
this is a better approach https://github.com/marimo-team/marimo/pull/5787, since this one does not work with redirects

PR #5787, opened 2025-07-28
fix: update base-url on remote healthy

mscolnick:


frontend/src/core/runtime/runtime.ts
```diff
@@ -167,14 +167,49 @@ export class RuntimeManager {
...
       if (response.redirected) {
         // strip /health from the URL
-        this.config.url = response.url.replace(/\/health$/, "");
+        const baseUrl = response.url.replace(/\/health$/, "");
+        this.config.url = baseUrl;
       }
-      return response.ok;
+
+      const success = response.ok;
+      if (success) {
```
dmadisetti, line 175: Is it possible for `response.redirected == False && response.ok == True`? Because this will then set the DOMBaseUri to be a noop? Maybe this is fine

frontend/src/core/runtime/runtime.ts
```diff
@@ -167,14 +167,49 @@ export class RuntimeManager {
...
+   * Side effects:
+   * - Changes how relative URLs are resolved in the document.
+   * - May affect the behavior of scripts, styles, and other resources that use relative URLs.
+   */
+  private setDOMBaseUri(uri: string) {
+    // Make sure there is a trailing slash
+    if (!uri.endsWith("/")) {
+      uri += "/";
+    }
+    const baseURI = document.querySelector("base");
```
dmadisetti:
```suggestion
    const base = document.querySelector("base");
```

since here; nit, can set `base.setAttribute("href", uri)` in one spot no branch

frontend/src/core/runtime/runtime.ts
```diff
@@ -167,14 +167,49 @@ export class RuntimeManager {
...
+   * If a `<base>` element already exists, its `href` attribute is updated.
+   * Otherwise, a new `<base>` element is created and appended to the `<head>`.
+   *
+   * Side effects:
+   * - Changes how relative URLs are resolved in the document.
+   * - May affect the behavior of scripts, styles, and other resources that use relative URLs.
+   */
+  private setDOMBaseUri(uri: string) {
+    // Make sure there is a trailing slash
+    if (!uri.endsWith("/")) {
```
dmadisetti, line 203: What about query params?

frontend/src/core/runtime/runtime.ts
mscolnick: it may not be a noop. the HTML can be served from a CDN and the remote URL not already set as the baseURL (and the remote url was not a redirect)

frontend/src/core/runtime/runtime.ts
mscolnick: oh good call, we can remove the query params

dmadisetti (approved):

PR #5794, opened 2025-07-28
Use explicit `uv run` for dev build script in CI

manzt:
The shebang in the `scripts/modify_pyproject_for_marimo_base.py` from #5778 is failing in CI. We should just be able to call `uv run`.


manzt:
think GH actions are just borked right now

PR #5802, opened 2025-07-29
Adding snippet for embedding visualisation

koaning:
The snippet in this PR contains a minimum example of a bulk labelling workflow, which doubles as a nice example of a embedding visualization technique.

### DIFF
--- a/marimo/_snippets/data/embeddings-0.py
+++ b/marimo/_snippets/data/embeddings-0.py
@@ -22,7 +22,7 @@ def _(mo):
         r"""
         # Visualization: Embedding Summary and Bulk Selection 
 
-        Create interactive dashboards using `mo.altair_chart` and `UMAP`.
+        Create interactive scatterplots using `mo.altair_chart` and `UMAP`.
         This technique is generally useful for any kind of embedding, but we're demonstrating it with text embeddings below.
         """
     )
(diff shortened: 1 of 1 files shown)


marimo/_snippets/data/altair-15.py
```diff
@@ -0,0 +1,74 @@
+# /// script
```
mscolnick: thoughts on renaming these series to `embedding-1.py`? maybe for more discoverability (also not teaching you altair, but rather using some tools for embedding)

koaning: Sure. In my mind this is the main visualisation technique for embeddings, but I agree it's less of an altair thing at this point

marimo/_snippets/data/embeddings-0.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@app.cell
+def _(StaticModel, UMAP, pl):
+    DATASET = "https://calmcode.io/static/data/clinc.csv"
+    TEXT_COL = "text"
+
+    df = pl.read_csv(DATASET).sample(10_000)
+    
+    # We're using Model2Vec because it so lightweight, sentence-transformers will also work!
```
akshayka:
```suggestion
    # We're using Model2Vec because it's lightweight; sentence-transformers will also work
```

marimo/_snippets/data/embeddings-0.py
```diff
@@ -0,0 +1,74 @@
...
+app = marimo.App(width="columns")
+
+
+@app.cell
+def _(mo):
+    mo.md(
+        r"""
+        # Visualization: Embedding Summary and Bulk Selection 
+
+        Create interactive dashboards using `mo.altair_chart` and `UMAP`.
```
akshayka:
```suggestion
        Create interactive scatterplots using `mo.altair_chart` and `UMAP`.
```

?

PR #5822, opened 2025-07-30
ai: improve prompt to not comment on the code

mscolnick:
No evals, just vibes. This does better than previously (tested on gpt-4o).

Somewhere claude recommended for important rules to include them in the bottom as well.


Light2Dark (approved):
maybe we should start being aggresive 😆
https://simonwillison.net/2025/Feb/25/leaked-windsurf-prompt/

PR #5824, opened 2025-07-30
Identify ddl, dmls and run clickhouse.command instead of query

Light2Dark:
## 📝 Summary

Allows running DDLs and DMLs on clickhouse without throwing errors.

This does change clickhouse behaviour, it requires `sqlglot`. I think it's okay and should be encouraged to install this to enable smooth execution.

Alternatively, we can just check if the lib exists and execute normally otherwise.

## 📋 Checklist

### DIFF
--- a/marimo/_ast/sql_utils.py
+++ b/marimo/_ast/sql_utils.py
@@ -1,6 +1,6 @@
 # Copyright 2025 Marimo. All rights reserved.
 
-from typing import Literal, Optional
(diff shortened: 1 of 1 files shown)


Light2Dark:
Ah yes, I think it might be good to bring in some of those sqlglot parsing benefits to other engines too (instead of just duckdb atm).

So we get the reactivity and can auto update the datasource panel. A future improvement.

PR #5837, opened 2025-07-30
Change vim mode command entry from `Shift+Escape` to `Ctrl/Cmd+Escape`

manzt:
On Linux systems, `Shift+Escape` triggers unintended behavior that interferes with marimo's command mode transition. This makes the current keybinding unusable for Linux users trying to exit vim's insert mode to marimo's command mode.

Following the changes from #5815 which introduced `Shift+Escape` to prevent accidental exits from vim mode, we're now switching to `Ctrl+Escape` (or `Cmd+Escape` on macOS) as a more reliable cross-platform solution.

### DIFF
--- a/frontend/src/components/editor/navigation/navigation.ts
+++ b/frontend/src/components/editor/navigation/navigation.ts
@@ -606,9 +606,9 @@ export function useCellEditorNavigationProps(cellId: CellId) {
 
   const { keyboardProps } = useKeyboard({
     onKeyDown: (evt) => {
-      // For vim mode, require Ctrl+Escape to exit to command mode
+      // For vim mode, require Ctrl+Escape (or Cmd+Escape on Mac) to exit to command mode
       if (keymapPreset === "vim") {
-        if (evt.key === "Escape" && evt.ctrlKey) {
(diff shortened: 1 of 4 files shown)


frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -606,9 +606,9 @@ export function useCellEditorNavigationProps(cellId: CellId) {
 
   const { keyboardProps } = useKeyboard({
     onKeyDown: (evt) => {
-      // For vim mode, require Shift+Escape to exit to command mode
+      // For vim mode, require Ctrl+Escape to exit to command mode
       if (keymapPreset === "vim") {
-        if (evt.key === "Escape" && evt.shiftKey) {
+        if (evt.key === "Escape" && evt.ctrlKey) {
```
akshayka: Should this be `evt.ctrlKey || evt.metaKey`? For consistency with our other keymaps, so that on macOS it will be `cmd+Esc`. Just a thought, up to you

manzt: I like that

manzt: Updated PR body/title

mscolnick (approved):

PR #5839, opened 2025-07-30
Document working with marimo notebooks in existing projects

### LINKED ISSUE #3008: Working with external surrounding project deps (with sandbox?)
### Documentation is

- [x] Missing
- [ ] Outdated
- [ ] Confusing
- [ ] Not sure?

### Explain in Detail

We work with notebooks within a larger project. Marimo is ideal for this, since it doesn't cause the horrendous diffs we get from Jupyter. 

The outer project is managed with poetry. We don't want marimo in the general project deps, and sometimes want additional tools in a notebook for testing/analysis that aren't in project deps. Not finding docs on a good workflow for this

So far been trying out the VSCode extension and pointing it at a separate python venv with marimo installed, then manually adding the project dir to the path within the notebook. This kinda works for simple things, but is fragile and doesn't have project-specific deps. Would be ideal if a workflow/solution would work with the VSCode extension, but not a requirement

It would be great to be able to have the project deps installed and then any notebook-specific deps layered on top all in a single sandbox. Not sure if that's achievable today? Maybe with uv directly? However it's best done, this general pattern of having some notebooks for doing exploratory work within a larger project repo seems like a common patter that it'd be great to see addressed in the docs

manzt:
Fixes #3008

Adds a new guide explaining how to use marimo notebooks within existing Python projects. The guide clarifies two approaches: sandbox notebooks for self-contained examples with isolated dependencies, and project notebooks that share the project's environment.

Think there is probably some overlap with the [uv's integration guide](https://docs.astral.sh/uv/guides/integration/marimo/) and our own [using uv](https://docs.marimo.io/guides/package_management/using_uv/) section. Think this could use a pass to make more agnostic of package manager.

### DIFF
--- a/docs/guides/configuration/runtime_configuration.md
+++ b/docs/guides/configuration/runtime_configuration.md
@@ -86,17 +86,37 @@ Autoreloading comes in two types:
 
 ## Python path
 
-By default, marimo will not add any additional directories to the Python path. This keeps the behavior of `marimo edit nb.py` and `python nb.py` consistent.
+By default, marimo will not add any additional directories to the Python path.
+This keeps the behavior of `marimo edit nb.py` and `python nb.py` consistent.
 
-You can add directories to the Python path by setting the `pythonpath` key in the
-runtime configuration. These directories will be added to the head of `sys.path`,
-similar to how the `PYTHONPATH` environment variable works. This allows Python to
-find and import modules from these directories.
+You can add directories to the Python path by setting the `pythonpath` key in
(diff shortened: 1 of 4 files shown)


manzt:
Thanks! I've added a section mentioning the pythonpath configuration, but kept it brief and emphasized leaning on package management when possible instead of path manipulation. I think it's better to guide users toward evolving Python packaging practices rather than encouraging runtime path changes that we are responsible for.

akshayka (approved):

PR #5851, opened 2025-07-31
Make reusable cell link only clickable on text, not entire block

manzt:
Now only the "reusable" text and the icons are clickable links. Previously, the entire output area block was clickable as a hyperlink when a cell contained a reusable definition. This made it easy to accidentally navigate away when clicking anywhere in that area, even though most of the block is not visually indicated as a link.

### DIFF
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -726,16 +726,16 @@ const EditableCellComponent = ({
             </div>
             {cellOutput === "below" && outputArea}
             {serialization && (
-              <a
-                href="https://links.marimo.app/reusable-definitions"
-                target="_blank"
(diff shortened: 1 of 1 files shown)


manzt:
https://github.com/user-attachments/assets/a438a615-e8a1-4c5b-83e2-61e069c7283a

Light2Dark:
looks great :>, what's with the branch name btw?

Light2Dark (approved):

manzt:
> looks great :>, what's with the branch name btw?

I've been trying our [jj](https://jj-vcs.github.io/jj/latest/), which has [anonymous branches](https://steveklabnik.com/writing/against-names/). Hard to describe, but you don't really think in terms of branches unless you're pushing to GitHub. You can name them... but you can also just `jj git push -c <id>` and it will make a branch for you.

PR #5852, opened 2025-07-31
Support hiding code in setup cells

### LINKED ISSUE #5727: Hide setup cell
### Description

The setup cell is often pretty long (e.g. a lot of module imports). It would be really useful to be able to hide it, just like how other cells can be hidden.

### Suggested solution

Support something like:

```python
with app.setup(hide_code=True):
    ...
```

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

manzt:
Fixes #5727

The setup cell is often used for lengthy imports and configuration that users don't need to see when viewing a notebook. This change allows users to hide the setup cell's code while keeping it functional, just like regular cells.

The implementation maintains backward compatibility by supporting both property access (`with app.setup:`) and method calls (`with app.setup(hide_code=True):`).

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -109,12 +109,13 @@ class _SetupContext:
     See design discussion in MEP-0008 (github:marimo-team/meps/pull/8).
     """
 
-    def __init__(self, cell: Cell):
+    def __init__(self, cell: Cell, hide_code: bool = False):
         super().__init__()
         self._cell = cell
         self._glbls: dict[str, Any] = {}
         self._frame: Optional[FrameType] = None
         self._previous: dict[str, Any] = {}
+        self._hide_code = hide_code
 
     def __enter__(self) -> None:
         if maybe_frame := inspect.currentframe():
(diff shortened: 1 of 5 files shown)


marimo/_ast/app.py
```diff
@@ -452,28 +495,30 @@ class MyStruct: ...
...
-        ```
-        """
-        # Get the calling context to extract the location of the cell
-        frame = inspect.stack()[1].frame
-        cell = self._cell_manager.cell_context(
-            app=InternalApp(self), frame=frame
-        )
-        self._setup = _SetupContext(cell)
-        return self._setup
+    setup: _SetupDescriptor = _SetupDescriptor()
```
dmadisetti: This code path seems a bit more convoluted. Not against it, just not sure I see the added benefit to the extra abstraction (over `property`) but lmk

marimo/_ast/app.py
```diff
@@ -159,6 +159,49 @@ def __exit__(
         return False
 
 
+class _SetupCallable(_SetupContext):
```
dmadisetti: I think you could stuff this in _SetupContext directly with call returning self

marimo/_ast/app.py
manzt: You're totally right. I was overthinking with descriptors.

mscolnick (approved):

PR #5855, opened 2025-07-31
Formatters for Ibis objects

### LINKED ISSUE #5828: Create a formatter for Ibis tables and values
### Description

For the PySpark dataframe type there is a marimo [formatter](https://github.com/marimo-team/marimo/blob/f21930a17d4abf4562d62918d7a7c4610252cf2b/marimo/_output/formatters/df_formatters.py#L159), which will execute the query (with a limit clause) and use `mo.table` to display the results.

We can use the same strategy for Ibis for the `ir.Table` and `ir.Column` types:

#### Eager

If eager execution is enabled (`ibis.options.interactive = True`) then we can just pass it to table (or call `to_pyarrow`) and assume that this is OK for the user, or we can use the limits that Ibis' use for their own [rich Table](https://github.com/ibis-project/ibis/blob/2fc33034412cb6d65a09303fadcee8971f27f61a/ibis/expr/types/_rich.py#L286)

#### Lazy 

If it's a lazy dataframe then we can either:

1. Do what the PySpark formatter does.
2. Let Ibis use it's default renderer:
  
  Image
3. Display something more "creative", for example, we could use `vstack` or `tabs` to render both the preview table and the output from `ibis.to_sql` (is this possible?)

#### Other types

While `ir.Table` and `ir.Column` are the most important objects, there are also `Value` types which interesting. For example, maybe we could use the [json layout]( https://docs.marimo.io/api/layouts/json/) for `MapValues` 😌 

### Suggested solution

Implement a another formatter in `df_formatters.py`. Just start with `ir.Table` and `ir.Column`.

- [x] Yes

### Alternatives

Try to get Ibis to do this on their end.

### Additional context

_No response_

kyrre:
## 📝 Summary

This change adds custom formatters for Ibis tables, columns and scalars. It closes #5828.

## 🔍 Description of Changes

Ibis uses the Rich library to display tables and scalars. However, doesn't do look great in marimo. It would be better to use the marimo's own table and output formatters.

  What this PR adds:

  - Table formatters: Table, Join, GroupedTable, CachedTable - displays as interactive marimo tables when possible
  - Column formatters: All 24+ column types (numeric, text, temporal, geospatial, etc.) - converts to single-column tables

  - Scalar formatters: Simple scalars (text display) and complex scalars like arrays/maps (JSON widgets)

  Eager execution and unbound expressions:
  - Interactive + bound expressions: Shows data as interactive table widgets or executed values
  - Lazy/unbound expressions: Shows expression representation + generated SQL (when backend supports SQL)
   - Error handling: Graceful fallbacks to increasingly coarse string representations

## 📋 Checklist

### DIFF
--- a/marimo/_output/formatters/df_formatters.py
+++ b/marimo/_output/formatters/df_formatters.py
@@ -13,9 +13,12 @@ from marimo._output.formatters.formatter_factory import (
     Unregister,
 )
 from marimo._plugins.stateless.mermaid import mermaid
+from marimo._plugins.stateless.plain_text import plain_text
+
 from marimo._plugins.ui._impl import tabs
 from marimo._plugins.ui._impl.table import get_default_table_page_size, table
 from marimo._runtime.patches import patch_polars_write_json
+
 
 LOGGER = _loggers.marimo_logger()
 
@@ -206,3 +209,56 @@ class PySparkFormatter(FormatterFactory):
             @formatting.opinionated_formatter(pyspark_DataFrame)
             def _show_df(df: pyspark_DataFrame) -> tuple[KnownMimeType, str]:
                 return table.lazy(df)._mime_()
+
+
+class IbisFormatter(FormatterFactory):
+    @staticmethod
+    def package_name() -> str:
+        return "ibis"
+
+    def register(self) -> None:
+
+        import ibis  # type: ignore[import-not-found]
+        import ibis.expr.types as ir  # type: ignore[import-not-found]
+        from ibis.backends.sql import SQLBackend  # type: ignore[import-not-found]
+
+        from marimo._output import formatting
+
+        if not include_opinionated():
+            return
+
+        def _format_ibis_expression(expr) -> tuple[KnownMimeType, str]:
+            """Format Ibis expressions (tables or columns converted to tables).
+            
+            Handles both interactive and non-interactive modes:
+            - Interactive: Shows as interactive table
+            - Non-interactive: Shows expression + SQL for SQL backends
+            """
+            try:
+                # Check if ibis is in interactive mode (eager execution)
+                if ibis.options.interactive:
+                    return table(expr, selection=None, pagination=True)._mime_()
+                else:
+                    backend = expr.get_backend()
+                    supports_sql = isinstance(backend, SQLBackend)
+                    
+                    if supports_sql:
+                        return tabs.tabs(
+                            {
+                                "Expression": plain_text(str(expr)),
+                                "SQL": f"```sql\n{ibis.to_sql(expr)!s}\n```"
+                            }
+                        )._mime_()
+                    else:
+                        return plain_text(str(expr))._mime_()
+
+            except BaseException as e:
+                LOGGER.warning("Failed to format Ibis expression: %s", e)
+                # Fallback to string representation
+                return ("text/plain", str(expr))
+
+        @formatting.opinionated_formatter(ir.Table)
+        def _show_marimo_ibis_table(
+            table_expr: ir.Table,
+        ) -> tuple[KnownMimeType, str]:
+            return _format_ibis_expression(table_expr)
--- a/marimo/_output/formatters/df_formatters.py
+++ b/marimo/_output/formatters/df_formatters.py
@@ -262,3 +262,26 @@ class IbisFormatter(FormatterFactory):
             table_expr: ir.Table,
         ) -> tuple[KnownMimeType, str]:
             return _format_ibis_expression(table_expr)
+
+        def register_column_formatters():
+            """Register formatters for all Ibis column types.
+            
+            Converts columns to tables using .as_table() then formats them.
+            """
+            column_types = [
+                ir.Column, ir.UnknownColumn, ir.NullColumn,
+                ir.NumericColumn, ir.IntegerColumn, ir.FloatingColumn, 
+                ir.DecimalColumn, ir.BooleanColumn, ir.StringColumn,
+                ir.TimeColumn, ir.DateColumn, ir.TimestampColumn, ir.IntervalColumn,
+                ir.GeoSpatialColumn, ir.PointColumn, ir.LineStringColumn, ir.PolygonColumn,
+                ir.MultiLineStringColumn, ir.MultiPointColumn, ir.MultiPolygonColumn,
+                ir.ArrayColumn, ir.SetColumn, ir.MapColumn, ir.StructColumn, ir.JSONColumn,
+                ir.BinaryColumn, ir.MACADDRColumn, ir.INETColumn, ir.UUIDColumn,
+            ]
+
+            for column_type in column_types:
+                @formatting.opinionated_formatter(column_type)
+                def _show_marimo_ibis_column(column_expr):
+                    return _format_ibis_expression(column_expr.as_table())
+
+        register_column_formatters()
--- a/marimo/_output/formatters/df_formatters.py
+++ b/marimo/_output/formatters/df_formatters.py
@@ -12,14 +12,12 @@ from marimo._output.formatters.formatter_factory import (
     FormatterFactory,
     Unregister,
 )
+from marimo._plugins.stateless.json_output import json_output
 from marimo._plugins.stateless.mermaid import mermaid
 from marimo._plugins.stateless.plain_text import plain_text
-from marimo._plugins.stateless.json_output import json_output
-
 from marimo._plugins.ui._impl import tabs
 from marimo._plugins.ui._impl.table import get_default_table_page_size, table
 from marimo._runtime.patches import patch_polars_write_json
-
 
 LOGGER = _loggers.marimo_logger()
 
@@ -213,6 +211,15 @@ class PySparkFormatter(FormatterFactory):
 
 
 class IbisFormatter(FormatterFactory):
+    """Custom formatting for Ibis expressions.
+
+    Provides custom formatters for tables, columns, and scalar values.
+
+    When eager execution is enabled, tables and values are executed/generated.
+    For unbound expressions, shows textual representation and generated SQL if backend supports SQL.
+    For lazy expressions, shows expression representation and SQL instead of executing.
+    """
+
     @staticmethod
     def package_name() -> str:
         return "ibis"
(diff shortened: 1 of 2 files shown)


marimo/_output/formatters/df_formatters.py
```diff
@@ -206,3 +208,368 @@ def _show_connect_df(
...
+                else:
+                    return table(
+                        expr, selection=None, pagination=True
+                    )._mime_()
+
+            except BaseException as e:
+                LOGGER.error("Failed to format Ibis expression: %s", e)
+                # Simple fallback - just show the expression as text
+                try:
+                    return ("text/plain", str(expr))
```
mscolnick: `repr()` might be better than `str()` here

marimo/_output/formatters/df_formatters.py
```diff
@@ -206,3 +208,368 @@ def _show_connect_df(
...
+            ) -> tuple[KnownMimeType, str]:
+                """Format GroupedTable expressions as plain text.
+
+                GroupedTable can be safely converted to a string.
+
+                Examples:
+                --------
+                >>> t.group_by("species")
+                """
+                return ("text/plain", str(grouped_expr))
```
mscolnick: `repr()` might be better than `str()` here

marimo/_output/formatters/df_formatters.py
```diff
@@ -206,3 +208,368 @@ def _show_connect_df(
...
+            pass
+
+        def register_column_formatters() -> None:
+            """Register formatters for Column types.
+
+            Handles numeric, text, temporal, geospatial, and complex column types.
+            Converts columns to tables using .as_table().
+            """
+            column_types = [
+                ir.Column,
```
mscolnick: is there a base type these all extend?

marimo/_output/formatters/df_formatters.py
```diff
@@ -206,3 +208,368 @@ def _show_connect_df(
...
+
+        def register_scalar_formatters() -> None:
+            """Register formatters for Scalar types.
+
+            Simple scalars render as text.
+            Complex scalars (arrays, maps, structs) render using marimo json_output.
+            """
+            # Simple scalars - render as text
+            simple_scalar_types = [
+                ir.Scalar,
```
mscolnick: is there a base type these all extend?

marimo/_output/formatters/df_formatters.py
```diff
@@ -206,3 +208,368 @@ def _show_connect_df(
...
+                            # Simple fallback - just show the scalar as text
+                            try:
+                                return ("text/plain", str(scalar_expr))
+                            except Exception:
+                                return (
+                                    "text/plain",
+                                    f"<{type(scalar_expr).__name__} object - could not display>",
+                                )
+
+        register_scalar_formatters()
```
mscolnick: would it be possible to write some tests for these formatters? you can pattern match off of `tests/_output/formatters/test_pandas.py`

marimo/_output/formatters/df_formatters.py
```diff
@@ -206,3 +208,368 @@ def _show_connect_df(
...
+            except BaseException as e:
+                LOGGER.error("Failed to format Ibis expression: %s", e)
+                # Simple fallback - just show the expression as text
+                try:
+                    return ("text/plain", str(expr))
+                except Exception:
+                    return (
+                        "text/plain",
+                        f"<{type(expr).__name__} object - could not display>",
+                    )
```
mscolnick: i think this can be DRYed up:

```
def render_plain_text():
      try:
          return ("text/plain", str(expr)) 1 1Code has comments. Press enter to view.
      except Exception:
          return (
              "text/plain",
              f"<{type(expr).__name__} object - could not display>",
          )
```

mscolnick (commented):
this looks great! thank you so much for the PR.

also if its helpful, you can include a smoke test in `marimo/_smoke_tests` for us to try out

kyrre:
This turned out to be more complicated than anticipated 😵  

_The commits are in the PR are meant to represent different versions/approaches_. Lets discuss tests if you want to proceed with any of these!

There were a few things that complicated matters:

1. For one, we had to register formatters for a lot of variations of the "same" type, e.g., Table, CachedTable, Join. Not huge issue, but maybe we can work around this.

2. Eager execution, unbound expressions and exceptions

The main issue is that Ibis expressions can be in different states, which requires different handling to avoid exceptions that weren't being display properly by marimo.

For example, if `ibis.options.interactive` is enabled AND the expression is Unbound then you have to make sure Expr.__repr__ does not call `to_pyarrow`, which throws an exception that is obfuscated by marimo, i.e., I couldn't make [this](https://github.com/ibis-project/ibis/blob/6b9f80c78e70c9013633ef89e5073a2081c49990/ibis/expr/types/core.py#L330) bubble back to marimo

It's not super complicated, but it forced me to use some private methods.

kyrre:
> this looks great! thank you so much for the PR.
> 
> also if its helpful, you can include a smoke test in `marimo/_smoke_tests` for us to try out

hey, sure, i was working on this comment when you starting replying (https://github.com/marimo-team/marimo/pull/5855#issuecomment-3140815870)

let me address your comments here:

if we can get away with just registering a formatter for each of the base types:

- Table
- Column
- Scalar 

then we can make it _really_ small. i did try that approach first, but wouldn't you have to change the whole OPINIONATED_FORMATTERS logic to make it work? 

the other issue that complicated things **a lot** is this the way exceptions are handled for formatters.

```
ibis.options.interactive = True 

t1 = ibis.table(dict(value1="float", key1="string", key2="string"), name="table1")
t2 = ibis.table(dict(value2="float", key3="string", key4="string"), name="table2")

joined = t1.left_join(t2, t1.key1 == t2.key3)

joined
```

this throws an IbisError which we want to display:

```
IbisError
Expression contains unbound tables and therefore cannot be executed. Use `<backend>.execute(expr)` to execute against an explicit backend, or rebuild the expression using bound tables instead.

See the console area for a traceback.
```

but something goes wrong (?) in `try_format` and it basically swallows the underlying error:

```
  File "/Users/kyrre.wahl.kongsgaard/projects/deathcon-workshop/.venv/lib/python3.13/site-packages/marimo/_output/formatting.py", line 201, in try_format
    mimetype, data = formatter(obj)
                     ~~~~~~~~~^^^^^
  File "/Users/kyrre.wahl.kongsgaard/projects/deathcon-workshop/.venv/lib/python3.13/site-packages/marimo/_output/formatters/repr_formatters.py", line 69, in f_repr
    "text/html" not in contents
TypeError: argument of type 'NoneType' is not iterable
```

if we can propagate the right error message back, then we don't need all that extra logic for handling interactive+unbound. we can just crash instead, which is fine.

marimo/_output/formatters/df_formatters.py
kyrre: yes, if we could just register [Scalar](https://github.com/ibis-project/ibis/blob/6b9f80c78e70c9013633ef89e5073a2081c49990/ibis/expr/types/generic.py#L1414) that would work. it would be a bonus if we could also register the ones that can be rendered as json

marimo/_output/formatters/df_formatters.py
kyrre: ir.Column 😁

mscolnick: i think that is good enough. it should operate on the base class as well iirc

marimo/_output/formatters/df_formatters.py
mscolnick: you can do a more fine-grained check inside.

```python
@formatting.opinionated_formatter(ir.Scalar)
def _show_scalar(scalar: ir.Scalar):
	if should_show_as_json(scalar):
		return ...
    return 
```

mscolnick:
@kyrre, thanks for picking this up and working through the complexity. 

the approach looks good so far. 

i think if `opinionated_formatter` is causing issues for the scalars or non-tables, you could just do `.formatter` instead. `opinionated_formatter` is just powered by a user-setting (because some people like their old tables). but for scalars, we could force a particular formatter.

> but wouldn't you have to change the whole OPINIONATED_FORMATTERS logic to make it work?

if this does fail for Tables and others, i can fix this. do you want to write a test to verify? if it fails, I can fix the implementation to get the test to pass.

kyrre:
> @kyrre, thanks for picking this up and working through the complexity.
> 
> the approach looks good so far.
> 
> i think if `opinionated_formatter` is causing issues for the scalars or non-tables, you could just do `.formatter` instead. `opinionated_formatter` is just powered by a user-setting (because some people like their old tables). but for scalars, we could force a particular formatter.
> 
> > but wouldn't you have to change the whole OPINIONATED_FORMATTERS logic to make it work?
> 
> if this does fail for Tables and others, i can fix this. do you want to write a test to verify? if it fails, I can fix the implementation to get the test to pass.

Registering just the base types worked when using the non-opinionated formatting decorator. This simplified things quite a bit!

Let me have a look at writing these test suites!

kyrre:

kyrre:
@mscolnick 

I have added some test suites and refactored the code, and think this version is ready for another review.

mscolnick:
@kyrre , couple asks:

1. can you rename `test_ibis` to `test_ibis_formatters.py`, other it conflicts with another file and pytest does not like that.
2. can you try signing the CLA again (by commeting with only "I have read the CLA Document and I hereby sign the CLA")

kyrre:

mscolnick:
there are a couple typechecking errors:
```
marimo/_output/formatters/df_formatters.py:232: error: Cannot find implementation or library stub for module named "ibis.backends.sql"  [import-not-found]
marimo/_output/formatters/df_formatters.py:232: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
marimo/_output/formatters/df_formatters.py:274: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
```

the easiest is to just `type: ignore` the first one

kyrre:
> there are a couple typechecking errors:
> 
> ```
> marimo/_output/formatters/df_formatters.py:232: error: Cannot find implementation or library stub for module named "ibis.backends.sql"  [import-not-found]
> marimo/_output/formatters/df_formatters.py:232: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> marimo/_output/formatters/df_formatters.py:274: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
> ```
> 
> the easiest is to just `type: ignore` the first one

sorry, i had forgotten to run all the checks locally.

for the second issue i was forced to use `Any` in the signature because it's is called with this value `scalar.to_pyarrow().as_py()`

kyrre:
recheck

kyrre:
> @kyrre , couple asks:
> 
> 1. can you rename `test_ibis` to `test_ibis_formatters.py`, other it conflicts with another file and pytest does not like that.
> 2. can you try signing the CLA again (by commeting with only "I have read the CLA Document and I hereby sign the CLA")


@mscolnick The CLA signing process seems to be broken. The same thing happened last time as well 🤷🏻

mscolnick:
@kyrre i just merged. it is because there is another git username that was not associated with your account. just a quirk of the CLA thing

PR #5873, opened 2025-08-01
fix: improved robustness of cache wrapper in module contexts

dmadisetti:
## 📝 Summary

Follow up to #5758

Solves reference dependency issues in an external lib, by creating a partial graph of toplevel functions on invocation. Previously, the "graph" of an external function was allowed to be blank. This is incorrect, since both setup cell values, and other top level definitions can be exposed to the cached function.

For example in tests/_save/external_decorators/transitive_wrappers_1.py

```python
with app.setup:
    ...
    impure_state = ...

@app.function
def my_impure_decorator(func):
    """An impure decorator that depends on impure_state"""

    @functools.wraps(func)
    def wrapper(*args: Any, **kwargs: Any):
        # Decorator depends on impure_state
        wrapper._value = ... # dependent on impure state ...
        return func(*args, **kwargs)

    return wrapper

...

@app.function
@my_impure_decorator
def pure_function():
    # This function itself is pure (no external dependencies)
    return 42

...
```

on `from external_decorators.transitive_wrappers_1 import pure_function` a partial graph is constructed to invalidate pure_function based on its relationship `pure_function` -> `impure_function` -> `impure_state`. Whereas before, we just ignored these relationships because `pure_function` was determined to be an external function and thus 'pure'.


marimo/_save/toplevel.py
```diff
@@ -0,0 +1,48 @@
...
+    TOPLEVEL_CELL_PREFIX,
+)
+from marimo._ast.toplevel import TopLevelExtraction
+from marimo._runtime.dataflow import DirectedGraph
+from marimo._types.ids import CellId_t
+
+LOGGER = _loggers.marimo_logger()
+
+
+def graph_from_app(fn: Callable[..., Any], app: App) -> DirectedGraph:
```
dmadisetti: I wouldn't mind a bit of a sanity check here.

This is the setup:
 - we are loading a marimo notebook as a module
 - a toplevel cache decorated function has dependencies on other top level dependencies.

This function is called on first invocation, where we do a best effort to create the graph thus far with top level definitions.

Notably, function definitions can be out of order (solved by waiting for invocation), but also invoked even before the app is done being constructed (via decorators). We grab `app` from the scope of the function declaration, and construct our own version of the graph with only top level deps.

marimo/_runtime/primitives.py
```diff
@@ -208,8 +208,18 @@ def is_pure_function(
     # We assume all external module function references to be pure. Cache can
     # still be be invalidated by pin_modules attribute. Note this also captures
     # cases like functors from an external module.
-    # TODO: Investigate embedded notebook values.
-    if getattr(value, "__module__", None) != "__main__":
+    module = getattr(value, "__module__", None)
+    is_lib = module != "__main__"
+
+    # If the function was not imported, and has the same name as the ref, then
```
dmadisetti: Another potentially brittle piece of logic. I think this works, but chaing to be explicit.

dmadisetti:
Woops. Saw the fastforward requested everyone. Sorry about that. rebasing

mscolnick (approved):

PR #5875, opened 2025-08-01
feat: --auto-convert to 'aggressively' open unknown file types

dmadisetti:
## 📝 Summary

Attempts to open a file as pypercent, python script, or just test block in that order is `--auto-convert` is passed in

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -56,7 +56,7 @@ def helpful_usage_error(self: Any, file: Any = None) -> None:
         click.echo(self.ctx.get_help(), file=file, color=color)
 
 
-def check_app_correctness(filename: str) -> None:
+def check_app_correctness(filename: str, noninteractive: bool = True) -> None:
     try:
         status = get_notebook_status(filename)
     except SyntaxError:
@@ -85,7 +85,8 @@ def check_app_correctness(filename: str) -> None:
         ) from None
 
     # Only show the tip if we're in an interactive terminal
-    if status == "invalid" and sys.stdin.isatty():
+    interactive = sys.stdin.isatty() and not noninteractive
+    if status == "invalid" and interactive:
         click.echo(
             green("tip")
             + ": Use `"
@@ -107,6 +108,29 @@ def check_app_correctness(filename: str) -> None:
         _loggers.marimo_logger().warning(
             "This notebook has errors, saving may lose data. Continuing anyway."
         )
+
+
+def check_app_correctness_or_convert(filename: str) -> None:
+    from marimo._convert.converters import MarimoConvert
+
+    code = ""
+    try:
+        return check_app_correctness(filename, noninteractive=True)
+    except SyntaxError:
+        # Create a single cell with the broken code
+        with open(filename, encoding="utf-8") as f:
+            code = MarimoConvert.from_text(source=f.read()).to_py()
+        with open(filename, "w", encoding="utf-8") as f:
+            f.write(code)
+    except click.ClickException:
+        # A click exception is raised if a python script could not be converted
+        with open(filename, encoding="utf-8") as f:
+            code = MarimoConvert.from_non_marimo_python_script(
+                source=f.read(), aggressive=True
+            ).to_py()
+
+    with open(filename, "w", encoding="utf-8") as f:
+        f.write(code)
 
 
 click.exceptions.UsageError.show = helpful_usage_error  # type: ignore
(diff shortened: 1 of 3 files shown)


dmadisetti:
Or maybe not internal. It's a hidden flag, but maybe someone will like to use it

marimo/_cli/cli.py
```diff
@@ -398,6 +422,14 @@ def _get_stdin_contents() -> str | None:
     hidden=True,
     help="Remote URL for runtime configuration.",
 )
+@click.option(
+    "--auto-convert",
+    is_flag=True,
+    default=False,
+    type=bool,
+    hidden=True,
+    help="Automatically convert the file to a marimo notebook.",
```
akshayka: This should be better documented

akshayka (commented):
I would just call this `--convert` instead of `--auto-convert`. The `auto` is redundant. Let's keep this hidden for now.

marimo/_cli/cli.py
dmadisetti: A bit more explicit in the message

marimo/_cli/cli.py
```diff
@@ -109,6 +110,29 @@ def check_app_correctness(filename: str) -> None:
...
 
+def check_app_correctness_or_convert(filename: str) -> None:
+    from marimo._convert.converters import MarimoConvert
+
+    code = ""
+    try:
+        return check_app_correctness(filename, noninteractive=True)
+    except SyntaxError:
+        # Create a single cell with the broken code
+        with open(filename, encoding="utf-8") as f:
```
mscolnick: using pathlib will make this code a bit cleaner in my opinion

marimo/_cli/cli.py
```diff
@@ -109,6 +110,29 @@ def check_app_correctness(filename: str) -> None:
...
+            code = MarimoConvert.from_text(source=f.read()).to_py()
+        with open(filename, "w", encoding="utf-8") as f:
+            f.write(code)
+    except click.ClickException:
+        # A click exception is raised if a python script could not be converted
+        with open(filename, encoding="utf-8") as f:
+            code = MarimoConvert.from_non_marimo_python_script(
```
mscolnick: should `from_non_marimo_python_script` and `from_text` be inverted?

marimo/_convert/converters.py
```diff
@@ -60,13 +61,39 @@ def from_non_marimo_python_script(
...
+        if aggressive:
+            notebook_ir = convert_non_marimo_script_to_notebook_ir(source)
+        else:
+            notebook_ir = convert_non_marimo_python_script_to_notebook_ir(
+                source
+            )
+        return MarimoConvert.from_ir(notebook_ir)
+
+    @staticmethod
+    def from_text(
```
mscolnick: `from_text` might be a confusing name

marimo/_convert/non_marimo_python_script.py
```diff
@@ -20,22 +79,31 @@ def convert_non_marimo_python_script_to_notebook_ir(
...
+        return convert_non_marimo_python_script_to_notebook_ir(source)
+    except ImportError:
+        pass
+    try:
+        import ast
+
+        ast.parse(source)  # Validate if it's valid Python code
+        return convert_python_block_to_notebook_ir(source)
+    except SyntaxError:
+        return convert_script_block_to_notebook_ir(source)
```
mscolnick, line 106: instead of all these try/except blocks and check, could we instead create a common convert:

```
class NotebookConverter:
     def try_convert(): IR | None

converters = [OneCoverter, TwoCoverter]
for converter in converts:
   if result := convert.try_convert(source)
           return result
```

also could use some tests for these new functions

marimo/_cli/cli.py
dmadisetti: Syntax errors must be opened as text, while a click error just means it's not a marimo notebook. I reordered the errors to highlight severity though

marimo/_convert/converters.py
dmadisetti: Not a fan either, but not really a fan of the alternatives
- `from_block`
- `from_blob`
- `from_string`

LMK if you have an idea

marimo/_convert/non_marimo_python_script.py
dmadisetti: Added the tests. Agreed that pattern is cleaner, but think it adds a lot more boiler plate for this one off code path. We can revisit if conversion become even hairier

marimo/_convert/converters.py
mscolnick: `from_string` or `from_plain_text`. the doc `Convert text from a script into` is confusing since it may not be a script

mscolnick (approved):

PR #5879, opened 2025-08-01
Improve cache for anywidget:

davidangarita1:
## 📝 Summary

The improved cache works for all types of widgets and avoids a race condition.

## 🔍 Description of Changes
* Created `class WeakCache` to handle caching of widget --> anywidget
* Allows caching of all widgets
* avoids try/except
* avoids further errors present in LOGGING during except
* squashes small/unlikely race condition

ping: @ayjayt

## 📋 Checklist

~- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).~

### DIFF
--- a/marimo/_plugins/ui/_impl/from_anywidget.py
+++ b/marimo/_plugins/ui/_impl/from_anywidget.py
@@ -23,6 +23,7 @@ if TYPE_CHECKING:
         AnyWidget,
     )
 
+
 LOGGER = _loggers.marimo_logger()
 
 
@@ -38,6 +39,9 @@ class WeakCache:
 
     def get(self, k: AnyWidget) -> UIElement[Any, Any] | None:
         return self._data.get(id(k))
+
+    def __len__(self) -> int:
+        return len(self._data)
 
     def _cleanup(self, oid: int) -> None:
         self._data.pop(oid, None)
--- a/marimo/_plugins/ui/_impl/from_anywidget.py
+++ b/marimo/_plugins/ui/_impl/from_anywidget.py
@@ -5,7 +5,7 @@ import hashlib
 import weakref
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any, Generic, Optional, TypeVar
 
 import marimo._output.data.data as mo_data
 from marimo import _loggers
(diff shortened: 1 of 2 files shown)


mscolnick:
@davidangarita1 - thanks for the contribution! do you mind signing the CLA by responding with only "I have read the CLA Document and I hereby sign the CLA"

davidangarita1:

ayjayt:
Hello Marimo team!

We (some of us) met at PyCon :-) I collaborate with Plotly on some of their tooling and I'm the lead author of GeoPozo.

This PR is a collaboration between David (on my team) and I.

So, and particularly for @manzt:

1. This improvement does what it says and allows weak caching of non-hasheable classes, and eliminates a corner-case race condition, and eliminates errors in logging.
2. But neither this improved version nor what we are replacing accomplish their *overall goal*: entries to the cache expire automatically.
 
I believe there is a memory leak in ipywidget:

In our testing, we originally used a very light `anywidget` wrapper as the object of the `WeakCache`, but the `weakref` is *never* destroyed even if we destroy all our own references to the object. Ostensibly something else is storing a never-expiring reference?

The `WeakCache` does work if we eliminate `anywidget` as a base class. I looked at `anywidget` and I didn't see any obvious memory leaks so maybe it's the ipywidget which it's based on?

So, while I recommend accepting this pull request (our code eliminates errors, but neither our code nor the original code works as it is intended):

@davidangarita1, happy birthday, and also could you please prepare **another** pull request using the original `TestWidget(_anywidget.AnyWidget)` testing which will intentionally fail in order to track the internal memory leak?

manzt:
Thanks for the PR and detailed write-up.

> The WeakCache does work if we eliminate anywidget as a base class. I looked at anywidget and I didn't see any obvious memory leaks so maybe it's the ipywidget which it's based on?

Yes, I'm aware of this issue with `ipywidgets`, and believe the suspect the cause of memory leaks is here:

https://github.com/jupyter-widgets/ipywidgets/blob/57ff79dce80b38b6666f66d1c7436153fe7c85b5/python/ipywidgets/ipywidgets/widgets/widget.py#L44-L46

I originally extended from `ipywidgets` for compatibility with the broader widget ecosystem, but it’s a heavy dependency with several problems, including this. In anywidget, we’ve built an experimental [descriptor](https://anywidget.dev/en/experimental/#mimebundle-descriptor) that uses weakrefs to manage comms lifetimes. This could form the basis for a new `AnyWidget` base class.

Long-term, the goal is to drop `ipywidgets` entirely and subclass directly from `traitlets.HasTraits`. The main blocker has been potential edge cases where users rely on parts of the large `ipywidgets` API surface area. But at this point, many widgets have already migrated, and cases like this are motivating to push forward. My thinking is that 99% of widgets would still "just work" with the current APIs, and the latter we could encourage to import from if the migration was painful:

```py
import anywidget.ipywidgets as anywidget

anywidget.AnyWidget # ipywidgets-based
```

And require `ipywidgets` as a dep.

Note, using our descriptor with proper weakrefs has tripped some folks up that have tried to call `display()`: 

- https://github.com/manzt/anywidget/issues/762

ayjayt:
Thanks for the context. I may take a look at it (time depending).

That said, this change is more or less good, solves some other problems, doesn't solve that problem.

Thanks!

manzt:
Thanks for the contribution!

PR #5854, opened 2025-07-31
Fix `__file__` in embedded notebooks

manzt:
Fixes #5850

When embedding notebooks with `app.embed()`, the `__file__` special variable was returning `"<unknown>"` instead of the actual file path. This broke relative path operations in embedded notebooks.

The issue was that `AppKernelRunner` hardcoded `filename = "<unknown>"` when creating the execution context for embedded apps. Now it uses the filename from the app being embedded.

PR #5880, opened 2025-08-01
Forward config parameter from plotly `fig.show()`

### LINKED ISSUE #5338: Plotly I cannot set the configuration ins fig.show(config=config)
### Describe the bug

When I want to apply some configuration in plotly through fig.show(config=config) assuming my figure is named 'fig' and my configuration string is set in 'config', I get strange results and the changes are not applied. The plot also appears strangely at the bottom of the cell (against my settings in marimo). 

In the sample code, I just adjust the 'toImage' property to save as svg.

If I don't want to make changes I can call 'fig' (instead of fig.show()) and the plot appears at the top where is should be. 

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

```python
import plotly.graph_objs as go
import plotly as pll

# Sample data for the demo plot
x_values = [1, 2, 3, 4, 5]
y_values = [10, 15, 13, 17, 21]

# Creating the plotly figure
fig = go.Figure(data=go.Scatter(x=x_values, y=y_values, mode='lines+markers', name='Demo Line Plot'))

# Adding title and labels
fig.update_layout(title='Demo Plot with Plotly', 
                  xaxis_title='X Axis', 
                  yaxis_title='Y Axis',
                  template='plotly_dark')
config = {'toImageButtonOptions': {'format': 'svg', 'filename': 'demo_plot'}}

fig.show(config=config)
```

manzt:
Fixes #5338

The `config` parameter is now forwarded to the frontend in `fig.show()`. Previously, `fig.show(config=config)` ignored the config. This removed the ability to customize plotly button options, image export formats, and other configuration settings.

### DIFF
--- a/marimo/_output/formatters/plotly_formatters.py
+++ b/marimo/_output/formatters/plotly_formatters.py
@@ -31,6 +31,7 @@ class PlotlyFormatter(FormatterFactory):
             @formatting.formatter(plotly.graph_objects.FigureWidget)
             def _show_plotly_figure(
                 fig: plotly.graph_objects.Figure,
+                config: dict[str, Any] | None = None,
             ) -> tuple[KnownMimeType, str]:
                 dragmode = getattr(fig.layout, "dragmode", None)
                 if dragmode is None:
@@ -38,7 +39,7 @@ class PlotlyFormatter(FormatterFactory):
                     fig.update_layout(dragmode="zoom")
                 json_str: str = pio.to_json(fig)
                 plugin = PlotlyFormatter.render_plotly_dict(
-                    json.loads(json_str)
+                    json.loads(json_str), config=config
                 )
                 return ("text/html", plugin.text)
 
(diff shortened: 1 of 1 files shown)


mscolnick (dismissed):

manzt:
@mscolnick we are getting type error because the formatting decorator doesn't allow for args/kwargs. I see two options: 

- 1.) Update `Formatter` to be a protocol:


```py
T = TypeVar("T", contravariant=True)
P = ParamSpec("P")


class Formatter(Protocol[T, P]):
    def __call__(
        self, obj: T, *args: P.args, **kwds: P.kwargs
    ) -> tuple[KnownMimeType, str]: ...
```


or 

- 2.) Just type ignore with a comment explaining why it's ok


I'm a little conflicted because I think the former is more correct but would require shuffling around some code and no other formatters so far require args/kwargs...

manzt:
actually, i'm dumb... there is a much more simple and correct work around

mscolnick (dismissed):

mscolnick (approved):

manzt:
tests were failing from #5854, rebased and should be good to go

PR #5891, opened 2025-08-03
Reorganize runtime settings in user configuration menu

akshayka:
Reorganize settings runtime settings for clarity.

Move and rename "auto output format" from runtime to autosave; even though this is a runtime setting in the backend, from a user perspective it makes more sense in autosave settings. Rename to "Save cell outputs as".


Light2Dark (approved):

akshayka:
@Light2Dark is this test failure related?

Light2Dark:
it's not, might be some mocked state not being cleaned up correctly in the tests.

PR #5897, opened 2025-08-04
Refactor `CellArray`/`Cell` to use atom-based state management

manzt:
Follow up to #5894. Cells now access their state through atoms (via specialized `useCellData` / `useCellRuntime` / `useCellHandle` hooks) instead of props, eliminating the large props interface. This achieves better render isolation – cells only re-render when their own state changes, and the containing `CellArray` no longer has a dependency on the notebook atom.

It's worth noting this refactor makes the `Cell` slightly harder to test in isolation, since you now need to mock a `NotebookState` instead of just passing down props. However, think the rendering benefits are justified and the dataflow is slightly more clear now that we aren't prop-drilling/spreading.

| Before                                                                          | After                                                                           |
| ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
|  |  |

### DIFF
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -543,7 +543,7 @@ const EditableCellComponent = ({
     return undefined;
   };
 
-  const isToplevel = !!(cellRuntime.serialization?.toLowerCase() === "valid");
+  const isToplevel = cellRuntime.serialization?.toLowerCase() === "valid";
 
   return (
     <TooltipProvider>
--- a/frontend/src/components/editor/Cell.tsx
+++ b/frontend/src/components/editor/Cell.tsx
@@ -33,6 +33,7 @@ import type { LanguageAdapterType } from "@/core/codemirror/language/types";
 import { canCollapseOutline } from "@/core/dom/outline";
 import { isErrorMime } from "@/core/mime";
 import type { AppMode } from "@/core/mode";
+import { connectionAtom } from "@/core/network/connection";
 import { sendStdin } from "@/core/network/requests";
 import type { CellConfig, RuntimeState } from "@/core/network/types";
 import { useResizeObserver } from "@/hooks/useResizeObserver";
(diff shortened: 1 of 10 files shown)


frontend/src/components/editor/Cell.tsx
```diff
@@ -598,34 +534,24 @@ const EditableCellComponent = ({
...
+    if (cellRuntime.status === "disabled-transitively") {
       return "This cell has a disabled ancestor";
     }
     return undefined;
   };
 
-  const isToplevel = !!(
-    serialization && serialization.toLowerCase() === "valid"
-  );
+  const isToplevel = !!(cellRuntime.serialization?.toLowerCase() === "valid");
```
Light2Dark: nit:   `const isToplevel = cellRuntime.serialization?.toLowerCase() === "valid";`

frontend/src/core/cells/cells.ts
```diff
@@ -1608,6 +1615,16 @@ export const columnIdsAtom = atom((get) =>
   get(notebookAtom).cellIds.getColumnIds(),
 );
 
+export const cellDataAtom = atomFamily((cellId: CellId) =>
```
Light2Dark: Should we remove values from the atomFamily on cell delete? https://jotai.org/docs/utilities/family#caveat-memory-leaks

Although we don't have infinite params, might be good to cleanup?
```ts
  cellDataAtom.remove(cellId);
  cellRuntimeAtom.remove(cellId);
  cellHandleAtom.remove(cellId);
```

Light2Dark (commented):
this is awesome, I really like the refactor!

frontend/src/core/cells/cells.ts
manzt: I was just thinking this! I think we can add to delete helpers.

frontend/src/components/editor/chrome/wrapper/minimap.tsx
```diff
@@ -32,8 +36,8 @@ const MinimapCell: React.FC<MinimapCellProps> = (props) => {
   const cell = {
     id: props.cellId,
     graph: graphs[props.cellId],
-    code: notebook.cellData[props.cellId].code,
-    hasError: notebook.cellRuntime[props.cellId].errored,
+    code: useAtomValue(cellDataAtom(props.cellId)).code,
```
mscolnick: could probably pull this out: `useAtomValue(cellDataAtom(props.cellId))`, so don't need two subscriptions

manzt: they are different atoms, but i realized from this comment the wrapping in useAtomValue makes it a little less readable. I exported specialized `useCellData` / `useCellRuntime` / `useCellHandle` hooks instead of exporting the atoms.

mscolnick: oh yea, misread that

mscolnick (approved):

PR #5899, opened 2025-08-04
fix: opinionated formatters work on subclasses. improve searching perf for formatters

mscolnick:


marimo/_output/formatting.py
```diff
@@ -42,11 +42,40 @@
...
+        self.formatters[t] = f
+
+    def get_formatter(self, obj: Any) -> Optional[Formatter[Any]]:
+        top_level_type = type(obj)
+        # Top-level formatters
+        if top_level_type in self.formatters:
+            return self.formatters[top_level_type]
+
+        # Search for formatters in the object's type hierarchy
+        for t in top_level_type.mro():
```
dmadisetti, line 66: I think maybe a guard against formatting `object` which is always the last type in mro? Or maybe that's a feature, not a bug?

dmadisetti (commented):
LGTM, one security thought though

marimo/_output/formatting.py
mscolnick: i don't think special casing is necessary. `object` if object is never in the formatters, and it is it, it probably is there for a reason

dmadisetti:
`tests/_output/formatters/test_ai_formatters.py` failing may be relevant. Rerunning tests just in case

PR #5900, opened 2025-08-04
Refactor pending delete service with reducer pattern

manzt:
Replaces React state primitives (`useState`, `useEffect`) with a reducer-based approach, eliminating race conditions from effect-based state synchronization. Auto-deletion of "simple" cells now happens synchronously within the reducer action.

We could move the computed properties into the reducer state, but for now I just wraped the existing API to preserve the computed properties (`idle`, `shouldConfirmDelete`).

### DIFF
--- a/frontend/src/core/cells/pending-delete-service.ts
+++ b/frontend/src/core/cells/pending-delete-service.ts
@@ -1,15 +1,17 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
-import { atom, useAtom, useStore } from "jotai";
-import { useCallback, useEffect } from "react";
+import { useAtomValue, useStore } from "jotai";
+import { useMemo } from "react";
 import {
   useDeleteCellCallback,
   useDeleteManyCellsCallback,
 } from "@/components/editor/cell/useDeleteCell";
 import { notebookAtom } from "@/core/cells/cells";
 import type { CellId } from "@/core/cells/ids";
+import type { JotaiStore } from "@/core/state/jotai";
 import { variablesAtom } from "@/core/variables/state";
 import type { VariableName } from "@/core/variables/types";
+import { createReducerAndAtoms } from "@/utils/createReducer";
 
 type PendingDeleteEntry =
   | {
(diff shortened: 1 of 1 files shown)


frontend/src/core/cells/pending-delete-service.ts
```diff
@@ -23,16 +25,30 @@ type PendingDeleteEntry =
...
+const { valueAtom: pendingDeleteStateAtom, useActions } = createReducerAndAtoms(
+  initialState,
+  {
+    submit: (
+      state,
+      action: {
+        cellIds: CellId[];
+        deleteCell: (args: { cellId: CellId }) => void;
+        deleteManyCells: (args: { cellIds: CellId[] }) => void;
+        store: JotaiStore;
```
manzt, line 45: Not sure if there is a better way to do this dependency injection. I just wrapped the hooks below and this isn't exported.

Light2Dark: we can import store `import { store } from "@/core/state/jotai";` if you want reduce passing that in.

frontend/src/core/cells/pending-delete-service.ts
```diff
@@ -81,53 +97,68 @@ export function usePendingDeleteService() {
...
 
-  return {
-    idle: state.size === 0,
-    shouldConfirmDelete: state.size > 1,
-    submit,
-    clear,
-  };
+export function usePendingDeleteService() {
+  const store = useStore();
+  const { submit, clear } = useActions();
```
Light2Dark, line 129: small thought if you want to rename this for clarity
```ts
const { submit: submitDelete, clear } = useActions(); // or something else
```

manzt: nice suggestion!

frontend/src/core/cells/pending-delete-service.ts
manzt: I think since we are injecting the other actions from another hook I don't mind passing in `store` this way. One benefit of `useStore` is that it lets use easier setup new atoms/store in tests since `useStore` will grab whatever is provided by a jotai Provider if there is one. It will fallback to global store (i believe) if none is provided).


For example, we have this in some tests:

```ts
function createTestWrapper() {
  const store = createStore();
  const wrapper = ({ children }: { children: React.ReactNode }) => (
    <Provider store={store}>{children}</Provider>
  );
  return { wrapper, store };
}
```

```ts
const { wrapper, store } = createTestWrapper();
store.set(notebookAtom, mockNotebook);
const result = await renderHook(<Component />, { wrapper });
```

Light2Dark (approved):

frontend/src/core/cells/pending-delete-service.ts
Light2Dark: yeah, the other option is to mock global store, but this is more explicit.

PR #5903, opened 2025-08-04
Fix Windows multiprocessing test hanging after pytest path fix

manzt:
Changes in #5854 fixed FileStore issues by providing a dummy path when
pytest.exe appears in `__file__`, but this broke multiprocessing tests
because the dummy file didn't exist for child processes to import.

This fix creates an actual dummy Python file that both satisfies
FileStore's need for a valid path and allows multiprocessing to
successfully spawn processes on Windows.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -152,14 +152,30 @@ class MockedKernel:
         self.stdin = MockStdin(self.stream)
         self._main = sys.modules["__main__"]
         # On Windows, sys.modules["__main__"].__file__ might be something like
-        # "pytest.exe\__main__.py" which causes issues. Use a dummy file instead.
+        # "pytest.exe\__main__.py" which causes issues. Use the current test file instead.
         test_file = None
         if (
             sys.platform == "win32"
             and hasattr(sys.modules["__main__"], "__file__")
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -152,14 +152,30 @@ def __post_init__(self) -> None:
         self.stdin = MockStdin(self.stream)
         self._main = sys.modules["__main__"]
         # On Windows, sys.modules["__main__"].__file__ might be something like
-        # "pytest.exe\__main__.py" which causes issues. Use a dummy file instead.
```
manzt, line 155: This is an issue in our `notebook_dir()` handler, which assumes that `Path(__file__).parent` is a directory... but in these tests `Path(__file__).parent` is `pytest.exe`, which is file. We could special case this there, and get rid of the conftest stuff, but i'm really not sure what the ideal approach is. Seems weird to make `pytest.exe/__main__.py` the `__file__` from pytest.

dmadisetti: `pytest.exe/__main__.py` is a result of our patching just in tests right?

Because otherwise it seems like we should be doing:

```python
while not notebook_dir.is_directory:
    notebook_dir = notebook_dir.parent
```

Still very weird `pytest.exe\__main__.py` pops up

PR #5906, opened 2025-08-04
Find first directory parent for `__file__` in `mo.notebook_dir`

manzt:
An alternative to our test patches in #5903 and #5854 is to just find
the first parent that is a directory for the edge case where a parent
path is actually a file (like `pytest.exe\__main__.py`).


manzt:
@dmadisetti following your suggestion as an alternative

dmadisetti (approved):
Oh, so much nicer. Thank you, I wasn't sure if it was just test dependent

PR #5917, opened 2025-08-05
docs: Expound upon mo.cache expectations and limitations

dmadisetti:
## 📝 Summary

This PR documents how marimo's caching utilities are keyed, describes limitations, and outlines best practices.

Our decorators are similar to but not a drop-in replacement for functools.cache, with some differences in how the cache key is constructed to support a larger universe of functions, and to minimize hidden state (such as closed-over variables, which our cache keys on but which functools does not).

When used as a context manager, cache restoration is deterministic and consistent when the code path relies on marimo's APIs and avoids external hidden state (like file system changes, network calls, or non-deterministic operations). In this case, the cache key is computed from function arguments, closed-over variables, and source code, ensuring reliable cache hits when these factors remain unchanged.

### DIFF
--- a/docs/api/caching.md
+++ b/docs/api/caching.md
@@ -254,10 +254,7 @@ def prompt_llm(query, **kwargs):
 
 ### Close over unhashable or un-pickleable arguments
 
-`marimo`'s caching mechanism is accounts for every named variable in your notebook.
-However, function arguments, are not notebook variables- and as such, they must be pickleable for caching to work.
-As a result, there may be some code patterns that will not worth with marimo's caching mechanism.
-For example:
+The [cache key](#cache-key) is constructed in part by hashing or pickle-ing function arguments. When you call a cached function with arguments that cannot be processed in this way, an exception will be raised. To parametrize cached functions with unhashable or un-pickleable arguments, syntactically close over them instead.
 
 **You can't do this:**
 ```python


docs/api/caching.md
```diff
@@ -15,7 +15,8 @@ The resulting cache is similar to `functools.cache`, but with the benefit that
 closed-over values) and isn't invalidated when the cell defining the decorated
 function is simply re-run (because it keys on notebook code). This means that
 like marimo notebooks, [`mo.cache`][marimo.cache] has no hidden state
-associated with the cached function, which makes you more productive while developing iteratively.
+associated with the cached function, which makes you more productive while
+developing iteratively. See [caching expectations](#expectations) for more details.
```
dmadisetti:
```suggestion
developing iteratively. See [caching expectations](#caching-expectations) for more details.
```

docs/api/caching.md
```diff
@@ -35,3 +36,252 @@ memory, letting you pick up where you left off.
...
+
+The performance and reliability of your notebook may depend on the caching functionality—as such, it is important to understand the expectations of marimo's caching utilities.
+
+`mo.cache` context blocks are designed to provide deterministic cache restoration when using marimo's APIs for state management.
+
+- On cache hit, the cached values are restored and the code block is not executed, meaning no side effects like printing, file I/O, or network requests will occur.
+- Rerunning a cell will not remove your cache, enabling cache hits on subsequent runs.
+- Dynamic changes to the notebook will appropriately invalidate the cache, if either:
+    1. the code in the defined block changes,
+    2. the variables with basic data types used by your code block change,
```
mscolnick: i think it would be good to provide examples for 2 and 3.

e.g. `x = 3`
and 
e.g. `pd.read_csv(https://some-external-url)

docs/api/caching.md
```diff
@@ -35,3 +36,252 @@ memory, letting you pick up where you left off.
...
+    3. or the relevant code that defines complex datatype variables changes.
+- External modules are assumed to not mutate variables in a way that would affect the cache.
+    - By setting `pinned_version` to `True`, you can ensure that the cache is invalidated when module versions change.
+    - This limitation does not apply if the external module is a marimo notebook.
+- Return values must be serializable.
```
mscolnick: these bullets feel like a mix of contract vs expectations. maybe this last could be re-worded or split it up

docs/api/caching.md
```diff
@@ -35,3 +36,252 @@ memory, letting you pick up where you left off.
...
+
+`mo.cache` decorators are meant to be drop-in replacements for `functools.cache`, adapted to a dynamic notebook environment. In addition to the expectations of context blocks, the additional limitations apply:
+
+- The wrapped function should be valid in all cases where `functools.cache` applies.
```
mscolnick: is this a limitation?

docs/api/caching.md
```diff
@@ -35,3 +36,252 @@ memory, letting you pick up where you left off.
...
+    and changes. If you mutate variables outside of marimo's APIs, the cache may
+    not be invalidated correctly, leading to stale data.
+
+## Comparison with functools.cache
+
+| Feature | marimo context | marimo decorator | functools.cache |
+|---------|----------------|------------------|-----------------|
+| **Dynamic Redefinition** | ✅ Handles cell re-execution | ✅ Handles function redefinition | ❌ Clears on redefinition |
+| **Global Scope Changes** | ✅ Tracks global dependencies | ✅ Tracks closure variables | ❌ Misses scope changes |
+| **Execution Context** | ✅ Detects external code changes | ✅ Detects external code Changes | ❌ No context awareness |
```
mscolnick: Changes -> changes

docs/api/caching.md
```diff
@@ -35,3 +36,252 @@ memory, letting you pick up where you left off.
...
+
+**Do this instead:**
+```python
+# Cell 1
+my_database_engine = ...
+
+# Cell 2
+@mo.cache
+def query_database(query):
+    return my_database_engine.execute(query)
```
mscolnick: could we add an example of whether or not to do:

```
@mo.cache
def query_database(query, engine):
    return engine.execute(query)

File: docs/api/caching.md [hunk 7]
```diff
@@ -35,3 +36,252 @@ memory, letting you pick up where you left off.
...
+
+```python
+# Cell 1
+from my_lib import my_decorator
+
+@mo.cache
+@my_decorator
+def expensive_function():
+    # ... some computation
+    return "result1"
```
akshayka: Is it safe to do this instead?

```python
@my_decorator
@mo.cache
def expensive_function():
    # ... some computation
    return "result1"
```

dmadisetti: Yes, but the behavior is different and there's no caveat. Worth highlighting?

akshayka: > Yes, but the behavior is different and there's no caveat. Worth highlighting?

Sorry, what do you mean that there's no caveat, and what is the behavior?

If

```python
@my_decorator
@mo.cache
def expensive_function():
    # ... some computation
    return "result1"
```

does what the user wants, then I would say it's worth emphasizing that in a "don't do this", "do this instead" tip.

docs/api/caching.md
```diff
@@ -1,37 +1,346 @@
...
+
+```python
+# Cell 2
+@mo.cache
+def prompt_llm(query, **kwargs):
+    message = {"role": "user", "content": query}
+    return llm_client.chat.completions.create(messages=[message], **kwargs)
+```
+
+### Leverage marimo's closure mechanism for unhashable arguments
```
akshayka:
```suggestion
### Close over unhashable or un-pickleable arguments
```

docs/api/caching.md
```diff
@@ -1,37 +1,346 @@
...
+
+`marimo`'s caching mechanism is accounts for every named variable in your notebook.
+However, function arguments, are not notebook variables- and as such, they must be pickleable for caching to work.
+As a result, there may be some code patterns that will not worth with marimo's caching mechanism.
+For example:
```
akshayka:
```suggestion
The [cache key](#cache-key) is constructed in part by hashing or pickle-ing function arguments. When you call a cached function with arguments that cannot be processed in this way, an exception will be raised. To parametrize cached functions with unhashable or un-pickleable arguments, syntactically close over them instead.
```

docs/api/caching.md
```diff
@@ -1,37 +1,346 @@
...
+    return engine.execute(query)
+```
+
+```python
+# This won't work because my_database_engine is not hashable
+query_database("SELECT * FROM my_table", my_database_engine)
+```
+
+Instead, by referencing the notebook variable directly, you can leverage
+marimo's closure mechanism to cache the function.
```
akshayka:
```suggestion
Instead, you can close over `my_database_engine`:
```

docs/api/caching.md
dmadisetti: Rewrite removes this block

docs/api/caching.md
dmadisetti: Added an LLM client example so that we have more than db

docs/api/caching.md
```diff
@@ -1,37 +1,346 @@
...
+    the Fibonacci sequence is a classic example of using `functools.cache`
+    effectively. On a basic macbook in pure python, `fib(35)` takes 1 second to
+    compute; with `mo.cache` it takes 0.000229 seconds; with `functools.cache`, it
+    takes 0.000025 seconds (x9 faster!!). Although relatively small, the additional overhead of
+    `mo.cache` (and more so `mo.persistent_cache`) is larger than
+    `functools.cache`. If your function takes more than a few milliseconds to
+    compute, the difference is negligible.
+
+
+## Tips for reliable caching
```
akshayka:
```suggestion
## Tips
```

docs/api/caching.md
akshayka: The reason for this suggestion: it is not clear how to leverage marimo's closure mechanism (what is marimo's closure mechanism? a user would ask), but "close over ... arguments" is immediately understandable to anyone who is familiar with Python.

docs/api/caching.md
dmadisetti: In coming steps I have an idea for how to make this more explicit, even in the example you provided

```python
data = pd.read_csv("https://some-external-url/")
```

Should not rely on code because pandas is "hashable". But we'll add this in once we provide an API to check

docs/api/caching.md
akshayka: The original text was confusing. What does it mean to "account for" named variables? What are "notebook variables"? The suggestion makes the language specific, direct, and actionable.

akshayka:
@dmadisetti Here is a suggested PR description:

This PR documents how marimo's caching utilities are keyed, describes limitations, and outlines best practices.

marimo provides two flavors of caching: in memory or to disk. The primary way of interfacing with these caches is through one of two decorators, though we also document a couple of other utilities that marimo provides, such as a context manager.

Our decorators are similar to but not a drop-in replacement for `functools.cache`, with some differences in how the cache key is constructed to support a larger universe of functions, and to minimize hidden state (such as closed-over variables, which our cache keys on but which functools does not).

docs/api/caching.md
dmadisetti: I think weighing in might be more confusing. Top of mind example:

```python
def decorator(fn):
    @functools.wrap(fn) # Optional in this case but still good practice
    def integrate(ts, ic):
        t0 = ts[0]
        response = [ic]
        for t1 in ts[1:]:
            dt = t1 - t0
            response.append(dt * fn((t0 + t1)/2, ...)) # Caching in standard trapeziod is not useful, and actually adds slow down
            # But using backward euler or something
            # _with_ the memoized function, then it is useful
            # dt * (fn(t1) + fn(t0))/2 ... or something more sophisticated
            t0 = t1
        return response
    return integrate

@decorator
@mo.cache # Not useful in naive example, but could be
def drag(...):
    ...
```

The behavior is different, depending on the ordering but it might be intentional. Does the user want to store the end result? or are they using caching on the function level for some sort of speedup?

Cache behaves as expected, there's nothing special to marimo here, and confusion here would stem from the user. Cache may not behave as expected in the decorator `cache(decorator(fn))`, so that's worth calling out

edit: woops, should be passing in response values, not `t` (but the point isn't the example)

dmadisetti:
> marimo provides two flavors of caching: in memory or to disk. 

I don't think this is a great distinction, but fine for now. We have in memory, disk, to s3, to redis + potentially more

The distinction in my mind is function level caching vs context block (work around for cell level) caching. The context block is a little awkward for now- but I think there's room for it to become a first class citizen and live on the cell level

akshayka:
I actually disagree. From the user's perspective, the main thing to know is: is data stored ephemerally or persistently? It is a very important distinction

PR #5925, opened 2025-08-06
improvement: show box drawings in tables better

Light2Dark:
## 📝 Summary

CleanShot 2025-08-06 at 14 02 28

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/columns.tsx
+++ b/frontend/src/components/data-table/columns.tsx
@@ -450,6 +450,19 @@ export function renderCellValue<TData, TValue>(
 
     const parts = parseContent(stringValue);
     const hasMarkup = parts.some((part) => part.type !== "text");
+
+    // Check if the string contains box drawing characters
+    const hasBoxDrawing = /[─│┌┐└┘├┤┬┴]/.test(stringValue);
+
+    // For box drawing characters, always show inline with proper formatting
+    if (hasBoxDrawing) {
+      return (
+        <div onClick={selectCell} className={cellStyles}>
+          <pre className="text-xs max-h-96 overflow-y-auto">{stringValue}</pre>
+        </div>
+      );
+    }
+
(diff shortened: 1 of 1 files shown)


marimo/_plugins/ui/_impl/tables/table_manager.py
```diff
@@ -140,10 +155,18 @@ def get_field_type(
...
-            for column_name in self.get_column_names()
-        ]
+        # Box drawings are only supported for small tables
+        check_box_drawing = self.get_num_columns() <= 3
+
+        field_types = []
+        for column_name in self.get_column_names():
+            field_type, external_type = self.get_field_type(column_name)
+            if check_box_drawing and has_box_drawing_characters(
+                self.get_sample_values(column_name)
```
mscolnick: it feels weird to have to get the samples in `get_field_types` (and maybe has a hidden performance hit) 

this could be fine, but curious if there are other approaches you though of?

Light2Dark: it's really specific to duckdb that we could just get the first value and check.

I tried moving this to the frontend in `renderCellValue` but this check happens on each cell render so not ideal imo.

Ideally i want to move this logic to `sql.py`, we can check if it's an explain query easily.
And we can format it to html probably. 

I can dive deeper into the ideal solution

mscolnick: yea i think we can maybe see if the statment is an explain with sqlglot

Light2Dark: explain statements just start with explain, doesn't need sqlglot.

there's likely a clever way to format the value to html regardless of the table manager.

mscolnick: you could wrap it it in `mo.plain` if it starts with `EXPLAIN`, would would do the default rendering of the table (instead of a fancy table)

Issue #5930, opened 2025-08-06
Cached UI elements not re-registered with UIElementRegistry


buckley-w-david:
### Describe the bug

When a notebook is using `mo.cache` to build UI elements, if the caching function returns an old UI element (one that was previously displayed but since replaced), it is not re-registered with the `UIElementRegistry`, and thus becomes disconnected from the rest of the notebook

https://github.com/user-attachments/assets/520921d2-7cef-478f-8c74-818b44dcf5a1

I had another (much larger) notebook not included in the below minimal reproduction where I collected the following information:

1. I added the follow code block to one of my cells to inspect the current state of the `UIElementRegistry`

```
from marimo._runtime.context import get_context
print(get_context().ui_element_registry._objects)
```

2. I manipulated the notebook such that a dropdown would be generated at the start (call it dropdown 'A'), then be replaced with a new one because the arguments to the caching function that returns it have changed (call it dropdown 'B'), and then changed the parameters back to generate dropdown 'A'

This is a snippet of the output at each of those 3 steps

```
# A
{ ... 'lEQa-0': <weakref at 0x147f48c70; to 'dropdown' at 0x147f1c910> ... }

# B
{ ... 'lEQa-0': <weakref at 0x15ef2d670; to 'dropdown' at 0x15ea8ef90> ... }

# Should be A, but is B
{ ... 'lEQa-0': <weakref at 0x15ef2d670; to 'dropdown' at 0x15ea8ef90> ... }
```

You can see at the last step where we should return to dropdown 'A', our UI element is unchanged from 'B'. The frontend updated correctly, but when it makes requests to update the state of the UI element on change, it ends up updating a dropdown that is no longer being actively used in the notebook, meaning anything depending on the dropdown does not see the new state.

### Will you submit a PR?

- [x] Yes

### Environment

<details>

This is from my local repo that is up to date with `main`

```
{
  "marimo": "0.14.16",
  "editable": true,
  "OS": "Darwin",
  "OS Version": "24.5.0",
  "Processor": "arm",
  "Python Version": "3.10.4",
  "Binaries": {
    "Browser": "138.0.7204.184",
    "Node": "v24.3.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "1.44.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
[...]

PR #5933, opened 2025-08-06
stub and restore cache meta return value

buckley-w-david:
## 📝 Summary

closes https://github.com/marimo-team/marimo/issues/5930

Process `self.meta["return"]` in the same way that `contextual_defs` are handled in `cache.py` so that when caching and restoring UI elements they are re-registered with the `UIElementRegistry`

## 🔍 Description of Changes

https://github.com/marimo-team/marimo/issues/5930 describes a bug from `UIElement`s not being re-registered when they are returned from a caching function.

It seemed like handling `UIElement`s (and some other classes) already handled doing this correctly when it came to "contextual_defs", so I figured that using that same process for the `return` value of the cached function should work as well. It was either that or explicitly re-register `UIElement`s on cache hit, which seemed like a worse solution.

I tried adding some tests for this but got kinda stuck on how to assert that the old and new ui elements were "the same" since the stub loading process technically creates a new instance. If we need tests for this, I would appreciate some advice.

## 📋 Checklist

### DIFF
--- a/marimo/_save/cache.py
+++ b/marimo/_save/cache.py
@@ -104,17 +104,10 @@ class Cache:
         """Restores values from cache, into scope."""
         for var, lookup in self.contextual_defs():
             value = self.defs.get(var, None)
-            # If it's a module we must replace with a stub.
-            if isinstance(value, ModuleStub):
-                scope[lookup] = value.load()
-            elif isinstance(value, FunctionStub):
-                scope[lookup] = value.load(scope)
-            elif isinstance(value, UIElementStub):
-                # UIElementStub is a placeholder for UIElement, which cannot be
-                # restored directly.
-                scope[lookup] = value.load()
-            else:
-                scope[lookup] = self.defs[var]
+            scope[lookup] = self._restore_from_stub_if_needed(value, scope)
+
+        if "return" in self.meta:
+            self.meta["return"] = self._restore_from_stub_if_needed(self.meta["return"], scope)
 
         defs = {**globals(), **scope}
         for ref in self.stateful_refs:
@@ -135,6 +128,24 @@ class Cache:
                     "Unexpected stateful reference type "
                     f"({type(ref)}:{ref})."
                 )
+    
+    def _restore_from_stub_if_needed(self, value: Any, scope: dict[str, Any]) -> Any:
+        """Restore objects from stubs if needed, recursively handling collections."""
+        if isinstance(value, ModuleStub):
+            return value.load()
+        elif isinstance(value, FunctionStub):
+            return value.load(scope)
+        elif isinstance(value, UIElementStub):
+            return value.load()
+        elif isinstance(value, (list, tuple)):
+            # Recursively restore collections
+            restored = [self._restore_from_stub_if_needed(item, scope) for item in value]
+            return type(value)(restored)  # Preserve original type (list/tuple)
+        elif isinstance(value, dict):
+            # Recursively restore dictionary values
+            return {k: self._restore_from_stub_if_needed(v, scope) for k, v in value.items()}
+        else:
+            return value
 
     def update(
         self,
@@ -179,15 +190,30 @@ class Cache:
                     f"({type(value)}:{ref})."
                 )
 
+        # Convert objects to stubs in both defs and meta
         for key, value in self.defs.items():
-            # If it's a module we must replace with a stub.
-            if inspect.ismodule(value):
-                self.defs[key] = ModuleStub(value)
-            elif inspect.isfunction(value):
-                self.defs[key] = FunctionStub(value)
-            elif isinstance(value, UIElement):
-                # UIElement cannot be restored directly, so we store a stub.
-                self.defs[key] = UIElementStub(value)
+            self.defs[key] = self._convert_to_stub_if_needed(value)
+            
+        if "return" in self.meta:
+            self.meta["return"] = self._convert_to_stub_if_needed(self.meta["return"])
+    
+    def _convert_to_stub_if_needed(self, value: Any) -> Any:
+        """Convert objects to stubs if needed, recursively handling collections."""
+        if inspect.ismodule(value):
+            return ModuleStub(value)
+        elif inspect.isfunction(value):
+            return FunctionStub(value)
+        elif isinstance(value, UIElement):
+            return UIElementStub(value)
+        elif isinstance(value, (list, tuple)):
+            # Recursively convert collections
+            converted = [self._convert_to_stub_if_needed(item) for item in value]
+            return type(value)(converted)  # Preserve original type (list/tuple)
+        elif isinstance(value, dict):
+            # Recursively convert dictionary values
+            return {k: self._convert_to_stub_if_needed(v) for k, v in value.items()}
+        else:
+            return value
 
     def contextual_defs(self) -> dict[tuple[Name, Name], Any]:
         """Uses context to resolve private variable names."""


marimo/_save/cache.py
```diff
@@ -179,15 +200,38 @@ def update(
...
-            if inspect.ismodule(value):
-                self.defs[key] = ModuleStub(value)
-            elif inspect.isfunction(value):
-                self.defs[key] = FunctionStub(value)
-            elif isinstance(value, UIElement):
-                # UIElement cannot be restored directly, so we store a stub.
-                self.defs[key] = UIElementStub(value)
+            self.defs[key] = self._convert_to_stub_if_needed(value)
+
+        if "return" in self.meta:
```
dmadisetti: I think this is fair to run on all meta keys. See https://github.com/marimo-team/marimo/pull/5935 where I was tinkering with the API

Sorry I didn't get this in at the time, I hadn't settled on the best way to expose "stubs" Let me know if you have feedback!

buckley-w-david:
> Awesome! Thank you. Do you mind adding a test for the added recursive stuff too?
> 
> Does it survive:
> 
> ```python
> a = []
> a.append(a)
> ```
> 
> ?
> 
> Thanks!

Great callout, it probably doesn't survive self-references like that

tests/_save/test_cache.py
```diff
@@ -9,12 +9,131 @@
...
+        cache = Cache(defs={}, hash="123", cache_type="Pure", stateful_refs=set(), hit=True, meta={})
+        scope = {}
+        ret = []
+        ret.append(ret)
+        cache.update(scope, {"return": ret})
+
+        stored = cache.meta["return"]
+        assert isinstance(stored, list)
+        assert len(stored) == 1
+        assert stored[0] is stored  # Self-reference maintained
```
buckley-w-david, line 54: One part of this I'm worried about @dmadisetti is that with these changes collections are no longer returning the literal same objects on subsequent calls, they're returning ones that are equal (by reconstructing the same thing every time during this stub processing)

So `old_result == new_result` should be true (assuming the objects support sane equality checks), but `old_result is new_result` will not

Not sure if you consider that a "breaking" change or not

buckley-w-david:
@dmadisetti actually I think this is causing a regression in something else I rely on (UI element state not being reset through a cell reload if the data they depend on haven't changed), I might put this into draft and come back at it from another angle

buckley-w-david:
@dmadisetti what are your thoughts on this approach?

https://github.com/marimo-team/marimo/compare/main...buckley-w-david:marimo:david/reregister-on-cache-hit

It's what I was going to do initially (and what I referenced in the PR description as something that seems like a worse implementation), but it doesn't break the use case I'm relying on 😅 

To give expanded context on that, we have some notebooks that dynamically build ui elements based on data from a csv. That CSV is being watched by `mo.watch.file` and is expected to change while you're viewing the notebook. That change reloads much of the notebook.

We are relying on the fact that by wrapping the creation of those ui elements in a function that uses `mo.cache`, they aren't recreated on cell reload, and so any state in them (the users current selection) is not reset as long as the parts that determine those ui elements have not also changed (which is pretty rare)

That is what is breaking with this current implementation (because it's no longer true that the ui elements aren't being recreated), and why I'm back to thinking about this other implementation

marimo/_save/cache.py
```diff
@@ -132,30 +131,45 @@ def restore(self, scope: dict[str, Any]) -> None:
...
-            ]
-            return type(value)(restored)  # Preserve original type (list/tuple)
+            )
         elif isinstance(value, dict):
-            # Recursively restore dictionary values
-            return {
-                k: self._restore_from_stub_if_needed(v, scope)
-                for k, v in value.items()
-            }
+            result = {}
```
dmadisetti, line 173:
```python
result = {restore(v, scope, memo) for v in vars}
memo[obj_id] = id(value)
value.clear()
value.update(result)
```

marimo/_save/cache.py
```diff
@@ -132,30 +131,45 @@ def restore(self, scope: dict[str, Any]) -> None:
...
+            result = value.load(scope)
         elif isinstance(value, UIElementStub):
-            return value.load()
-        elif isinstance(value, (list, tuple)):
-            # Recursively restore collections
-            restored = [
-                self._restore_from_stub_if_needed(item, scope)
+            result = value.load()
+        elif isinstance(value, list):
+            result = []
```
dmadisetti, line 159: see dict case below

```python
result = [
                self._restore_from_stub_if_needed(item, scope, memo)
                for item in value
            ]
value.clear()
value.extend(result)
```

tests/_save/test_cache.py
dmadisetti: You can use the same objects, which I think is preferable (let comments on restore)

marimo/_save/cache.py
```diff
@@ -201,37 +215,53 @@ def update(
...
+            memo = {}
+        
+        # Check for cycles
+        obj_id = id(value)
+        if obj_id in memo:
+            return memo[obj_id]
+        
         if inspect.ismodule(value):
-            return ModuleStub(value)
+            result = ModuleStub(value)
```
dmadisetti, line 249: You only need to memo list / dicts so I think returning is fine and edge case handling in those cases.

reconstructing the module stub is super cheap too, so don't think there's much perf benefit either

dmadisetti:
> @dmadisetti what are your thoughts on this approach?

These approaches aren't mutually exclusive though right? I think properly "registering" under load is ideal though: https://github.com/buckley-w-david/marimo/blob/dc4ef91f7e3e1ede304bcc957546c4379bf38509/marimo/_save/cache.py#L80

dmadisetti:
@buckley-w-david 

I'm going on vacation next week and would like to get this in. If you have a test case for the behavior you'd like to see (it can fail, or just be a sample script, I just want something to work against)- I grab this and get it over the finish line.

Thanks again!

buckley-w-david:
> @buckley-w-david
> 
> I'm going on vacation next week and would like to get this in. If you have a test case for the behavior you'd like to see (it can fail, or just be a sample script, I just want something to work against)- I grab this and get it over the finish line.
> 
> Thanks again!

Hey, sorry have been focused elsewhere today, I'll try to put together an example of what I'm talking about above tonight. Probably will be a small notebook + set of instructions to reproduce what I'm talking about rather than an actual test case

dmadisetti:
Totally fine! Appreciate it, just don't want to break your code (:

buckley-w-david:
Alright here is a demonstration notebook @dmadisetti

https://gist.github.com/buckley-w-david/b9b38c841f979b2162c44a2d5d177339

If you select an option from the dropdown, it will show you some data at the bottom of the notebook. Different data depending on which dropdown item you select. The dropdown items are dynamically built from the data in a file.

If you click on the "Change the file" button, the file will be updated, cells will reload (because we're watching the file with `mo.watch.file`), but _the dropdown selection is persistent_. While the cells reload the selection in the dropdown is not lost since the specific part of the data it was using to build those options was not part of what was changed, and so the `make_ui` function returns the same thing

This specific demonstration is somewhat forced, in my real life usage of this the file changes come from an outside process not in the notebook.

---

All that being said, I just tested the above notebook in this branch it is unexpectedly working :sweat_smile: I did a less thorough check yesterday that I thought was telling me the above behaviour was broken, but maybe I was just checking it badly

I guess false alarm then?

dmadisetti:
🔥  I'll pull this and double check. Persistent cache in particular with new sessions can cause issues.

But thanks for the repro!

dmadisetti:
Closing put for the updated branch. Thanks again!

PR #5948, opened 2025-08-06
fix: formatter for 'type()' and make ibis formatters opinionated

mscolnick:


dmadisetti (approved):
What's the difference between opinionated and not?

PR #5966, opened 2025-08-07
Allow deleting variables defined by other cells

akshayka:
This change makes it possible to delete variables that were defined in another cell.

The motivation for this change is to allow advanced users to reclaim
memory. Especially when dealing with tensors or large datasets, either
reusing a variable name or explicitly deleting it with `del` are
traditionally used to reclaim memory. Reusing variables is not allowed
in marimo, and before this change `del` was not either, since it wasn't
clear how to order cells.

This change allows the `del` operator by introducing the following rule:

  If a cell deletes a variable, it becomes a child of all other cells
  that reference that variable.

For example:

**cell u**
```
x
```

**cell v**
```
del x
```

In this case, **v** becomes a child of **u**.

This has the (desired) consequence that two cells cannot delete the same
variable, as that forms a cycle error.

This change does introduce some path dependency in how cells are
executed. In the above example, if cell u were manually re-run after
cell v ran, the user would get a NameError. We mitigate this by
communicating in the error message that the variable may have been
deleted (if at least one cell deletes the variable).

While the path dependency is unfortunate, using `del` is an advanced pattern that one
only uses when they're in pretty deep (and trying to avoid restarting
their kernel), so trading off execution reproducibility for
unblocking more resource intensive workflows is in my mind worth it.
Without this change, users would need to instead coalesce all
memory-intensive work into a single function (making incremental
computation difficult).

### DIFF
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -202,6 +202,18 @@ class DirectedGraph:
                     if path:
                         self.cycles.add(tuple([(other_id, cell_id)] + path))
                     self.children[other_id].add(cell_id)
+
+            for name in cell.deleted_refs:
+                referring_cells = self.get_referring_cells(
+                    name, language="python"
+                ) - set((cell_id,))
+                for other_id in referring_cells:
+                    parents.add(other_id)
+                    path = self.get_path(cell_id, other_id)
+                    if path:
+                        self.cycles.add(tuple([(other_id, cell_id)] + path))
+                    self.children[other_id].add(cell_id)
+
         LOGGER.debug("Registered cell %s and released graph lock", cell_id)
         if self.is_any_ancestor_stale(cell_id):
             self.set_stale(set([cell_id]))
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -179,6 +179,8 @@ class DirectedGraph:
                     self.parents[child].add(cell_id)
 
             for name in cell.refs:
+                # First, for each referenced variable, we add cells that define
+                # that variable as parents
                 other_ids_defining_name: set[CellId_t] = (
                     self.definitions[name]
                     if name in self.definitions
@@ -202,6 +204,23 @@ class DirectedGraph:
                     if path:
                         self.cycles.add(tuple([(other_id, cell_id)] + path))
                     self.children[other_id].add(cell_id)
+
+                # Next, any cell that deletes this referenced variable is made
+                # a child of this cell. If two cells delete the same variable,
+                # they form a cycle.
+                other_ids_deleting_name: set[CellId_t] = set(
+                    cid
+                    for cid in self.get_referring_cells(
+                        name, language="python"
+                    )
+                    if name in self.cells[cid].deleted_refs
+                ) - set((cell_id,))
+                for v in other_ids_deleting_name:
+                    path = self.get_path(v, cell_id)
+                    if path:
+                        self.cycles.add(tuple([(cell_id, v)] + path))
+                    self.parents[v].add(cell_id)
+                children.update(other_ids_deleting_name)
 
             for name in cell.deleted_refs:
                 referring_cells = self.get_referring_cells(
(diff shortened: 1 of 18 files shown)


marimo/_runtime/runner/cell_runner.py
```diff
@@ -354,9 +354,19 @@ def _run_result_from_exception(
...
+
+                deleted_message = (
+                    " Another cell may have deleted it with the del operator."
+                    if possibly_deleted
+                    else ""
+                )
                 output = MarimoExceptionRaisedError(
-                    f"marimo came across the undefined variable `{ref}` "
-                    "during runtime. Definition expected in cell : ",
+                    f"Name `{ref}` is not defined.{deleted_message} "
```
dmadisetti, line 368: I thought we talked about rerunning cells to get the value again? I think this is OK, but just wondering the scope / behavior of del

marimo/_runtime/dataflow.py
```diff
@@ -202,6 +204,35 @@ def register_cell(self, cell_id: CellId_t, cell: CellImpl) -> None:
...
+                children.update(other_ids_deleting_name)
+
+            for name in cell.deleted_refs:
+                referring_cells = self.get_referring_cells(
+                    name, language="python"
+                ) - set((cell_id,))
+                for other_id in referring_cells:
+                    parents.add(other_id)
+                    path = self.get_path(cell_id, other_id)
+                    if path:
```
dmadisetti, line 259: Can you leave a comment that del refs should always be terminal nodes? Obvious in the context of this PR, but maybe still worth noting

tests/_runtime/test_dataflow.py
```diff
@@ -974,6 +974,33 @@ def test_cycles() -> None:
...
+    graph.register_cell("1", parse_cell("x"))
+    graph.register_cell("2", parse_cell("x = 1"))
+    assert graph.parents["0"] == set(["1", "2"])
+    assert graph.children["0"] == set()
+
+    assert graph.parents["1"] == set(["2"])
+    assert graph.children["1"] == set(["0"])
+
+    assert graph.parents["2"] == set([])
+    assert graph.children["2"] == set(["0", "1"])
```
dmadisetti, line 1012: Would also like to see:

```python
@app.cell
def A():
    x = 1

@app.cell
def B():
    del x
    y = 1

@app.cell
def C():
    z = x + y
```

This creates the cycle error right? Does the error message make sense?

dmadisetti: Think it would also be useful to test the behavior of a dynamic kernel


```python
@app.cell
def A():
    x = 1

@app.cell
def C():
    del x
```

then

```python
@app.cell
def B():
    y = x + 1
```

dmadisetti (commented):
I was a bit more skeptical on first suggestion, reading through it- this seems less aggressive than what I initially understood. I'll pull it down and play with it a bit, but think it would be useful to see more of the behavior codified in tests.

I think this is still potentially dangerous with run all, but can be manually leveraged with caching to the effect I was thinking of. Still think it's a "power user" feature, but also warrants some documentation

marimo/_runtime/runner/cell_runner.py
akshayka: I decided against it. The error message points to the defining cell, so the user can go and run manually.

If we want to improve the developer experience, we could have an inline "warning" similar to destructive delete in the future, in which the user has the option of either running the defining cell or running this particular cell. But that felt out of scope, since it requires substantial changes.

tests/_runtime/test_dataflow.py
akshayka: 100%, adding a test for both. Thanks for the callout.

The error message does make sense (but it's terse). (Caught a bug in the error message with this test, thank you.) I think it's fine for now

image

akshayka:
Feeling pretty good about this change now: it breaks down a wall that advanced users would hit, without compromising the developer experience for novice and intermediate users.

dmadisetti (approved):
I feel good about this too. The NameError provides context in terms of foot gun. Docs look good

Just a note that the runner defined in dataflow (and used by `my_cell.run`) won't be influenced by this- but I think that's fine (and even ideal). Though maybe we should revisit that code all together since I think it may be an alpha remnant

PR #5971, opened 2025-08-08
Refactor frontend request layer to use dependency injection

manzt:
Previously, the request layer used static module-level resolution that eagerly determined the runtime (WASM, Islands, Network) at import time. This required bundling all network implementations even when only one was needed.

Now the request layer uses dependency injection via a Jotai atom. Each entrypoint provides its implementation:

```ts
// In mount.tsx (main app)
store.set(requestClientAtom, resolveRequestClient()); // wasm or not

// In islands/main.ts
store.set(requestClientAtom, IslandsPyodideBridge.INSTANCE);

// Future: VS Code extension
store.set(requestClientAtom, createVSCodeRequests());
```

This allows adding new networking layers without importing others, simplified mocking, and runtime implementation swapping.


frontend/src/mount.tsx
```diff
@@ -266,6 +268,9 @@ function initStore(options: unknown) {
     );
   }
 
+  // Configure networking layer
+  store.set(requestClientAtom, resolveRequestClient());
+
```
manzt, line 273: For marimo "app"

frontend/src/core/islands/main.ts
```diff
@@ -46,6 +46,9 @@ import { dismissIslandsLoadingToast, toastIslandsLoading } from "./toast";
  * Initialize the Marimo app.
  */
 export async function initialize() {
+  // Setup networking
+  store.set(requestClientAtom, IslandsPyodideBridge.INSTANCE);
+
```
manzt, line 51: for islands

PR #5979, opened 2025-08-10
Migrate to Tailwind CSS v4

manzt:
Runs the auto-migrate utility and manually moves some configuration from `tailwind.config.cjs` into `global.css`. Some more of the `config.extend` properties can be moved into globals under the `@theme` section, however I'd like to piece that out incrementally.

The main benefit of these changes is that marimo frontend requires less coordination of build tooling to build for separate entry points. For example for VS Code entrypoint, I can just use the `@tailwindcss/vite` plugin (without PostCSS).

```ts
// vite.config.js
import { defineConfig } from "vite";
import tailwindcss from "@tailwindcss/vite";

export default defineConfig({
  build: {
    lib: {
      entry: { renderer: "./src/renderer.tsx" },
      formats: ["es"],
    },
  },
  plugins: [
    tailwindcss(),
  ],
  resolve: {
    tsconfigPaths: true,
  },
  experimental: {
    enableNativePlugin: true,
  },
});
```

### DIFF
--- a/frontend/src/components/editor/controls/Controls.tsx
+++ b/frontend/src/components/editor/controls/Controls.tsx
@@ -219,7 +219,7 @@ const StopControlButton = ({
 };
 
 const topRightControls =
-  "absolute top-3 right-5 m-0 flex items-center space-x-3 min-h-[28px] no-print pointer-events-auto z-30 print:hidden";
+  "absolute top-3 right-5 m-0 flex items-center gap-2 min-h-[28px] no-print pointer-events-auto z-30 print:hidden";
 
 const bottomRightControls =
   "absolute bottom-5 right-5 flex flex-col gap-2 items-center no-print pointer-events-auto z-30 print:hidden";


frontend/tailwind.config.cjs
```diff
@@ -16,31 +16,6 @@ module.exports = {
...
-          "2px 2px 0px 0px var(--base-shadow-darker), 0px 0px 2px 0px hsl(0deg 0% 50% / 20%)",
-        mdSolid:
-          "4px 4px 0px 0px var(--base-shadow-darker), 0 0px 2px 0px hsl(0deg 0% 60% / 50%)",
-        lgSolid:
-          "5px 6px 0px 0px var(--base-shadow-darker), 0 0px 4px 0px hsl(0deg 0% 75% / 50%)",
-        xlSolid:
-          "7px 8px 0px 0px var(--base-shadow-darker), 0 0px 4px 0px hsl(0deg 0% 85% / 50%)",
-        "2xlSolid":
-          "10px 12px 0px 0px var(--base-shadow-darker), 0 0px 8px 0px hsl(0deg 0% 90% / 50%)",
-      },
```
manzt, line 43: note how these just become variables in the `@theme` section with tailwind v4.

manzt:
ugh there are some conflicting biome and stylelint formatting that i need to sort out..

manzt:
TS errors are unrelated (#5980)

frontend/src/css/app/Cell.css
```diff
@@ -234,22 +243,18 @@
...
-  border-color: transparent;
-  box-shadow: none;
-
-  /* Apply the original styles */
-  &:hover {
-    border: 1px solid var(--gray-4);
-    @apply shadow-smSolid shadow-shade;
-  }
-}
-
```
Light2Dark, line 247: this was added so we didn't miss out on changing the styles across both places.

frontend/src/components/editor/actions/name-cell-input.tsx
```diff
@@ -39,9 +39,9 @@ export const NameCellInput: React.FC<Props> = ({
       return;
     }
 
-    input.addEventListener("blur", onBlur);
+    input.addEventListener("blur-sm", onBlur);
```
Light2Dark: is this correct?

Light2Dark (approved):
nicee! i think spotted small things

frontend/src/components/editor/actions/name-cell-input.tsx
manzt: It's not, good catch. The tailwind convert script did `blur` -> `blur-sm` and `outline` -> `outline-solid` in a few spots that was incorrect. This one got through my pass.

frontend/src/css/app/Cell.css
manzt: I had to remove the `.borderless-common` class because Tailwind v4 doesn't seem to support defining and applying custom utility classes like this inline (causes compilation failure). 

In this case, it seemed like the most simple approach was to just duplicate the code. However, to avoid code duplication between `.marimo-cell.borderless` and `.dark .marimo-cell.borderless` we could use  a more CSS-native solution like the [`:is()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) selector instead relying on tailwind utilities to define shared behavior. E.g.,

```css
:is(.marimo-cell, .dark .marimo-cell).borderless {
  border-color: transparent;
  box-shadow: none;
  
  &:hover {
    border: 1px solid var(--gray-4);
    @apply shadow-sm-solid shadow-shade;
  }
}
```

or with nesting:

```css
.marimo-cell {
  /* ... other styles ... */

  .dark & {
    @apply border-border;
  }

  :is(&, .dark &).borderless {
    border-color: transparent;
    box-shadow: none;

    &:hover {
      border: 1px solid var(--gray-4);
      @apply shadow-sm-solid shadow-shade;
    }
  }
}
 ```

Light2Dark: alright, i can help to do that in subsequent PRs, thanks!

mscolnick (approved):
looks good to me - better to get this out sooner to avoid merge conflicts

PR #5980, opened 2025-08-10
Remove handling for deleted `delete-nonlocal` error type

manzt:
The backend changes in #5966 removed the `delete-nonlocal` error variant from the API schema, but the frontend still had code attempting to handle this error type. This caused TypeScript errors because the type narrowing created a `never` type when filtering for a non-existent variant, making the error object properties inaccessible.

### DIFF
--- a/.github/workflows/test_fe.yaml
+++ b/.github/workflows/test_fe.yaml
@@ -23,8 +23,7 @@ jobs:
           filters: |
             frontend:
               - 'frontend/**'
-              - 'lsp/**'
-              - 'openapi/**'
+              - 'packages/**'
 
   test_frontend:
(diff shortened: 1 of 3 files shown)


akshayka (dismissed):
Thanks for catching this!

manzt:
It was missed by CI because my changes in #5668 failed to update the filters for the new openapi/lsp package location under `packages/**` directory. I added a fix to this PR so we continue to run our frontend CI any time the API types change.

akshayka (approved):

manzt:
Agh, completely missed #5977. Add you as a co-author @Light2Dark, sorry about that.

PR #5985, opened 2025-08-10
Keep box-shadow CSS definitions on single lines

manzt:
Biome formatter was breaking multi-value box-shadow declarations across multiple lines, which caused Tailwind v4 ring utilities and shadow variants to fail. Updated biome-ignore comments to clarify that these definitions must stay on single lines to work with Tailwind variants, not due to tool conflicts.


manzt:
I noticed this because the cell selection rings weren't showing up on `main`.

mscolnick (approved):

PR #5992, opened 2025-08-11
fix: get fully qualified name in sql refs

### LINKED ISSUE #5945: Spurious references to variables from SQL cells can create cycles
Image

Reproduction:

```python
import marimo

__generated_with = "unknown"


@app.cell
def _(equipment, mo, schema, sites):
    df = mo.sql(
        f"""
        select s.site_name, e.equipment_name from schema.sites as s join schema.equipment as e using (site_id)
        """
    )
    return (df,)


@app.cell
def _(df):
    df
    sites = ...
    return (sites,)


if __name__ == "__main__":
    app.run()
```

### Discussed in https://github.com/marimo-team/marimo/discussions/5944

<sup>Originally posted by **waterworthd-cim** August  6, 2025</sup>
I'm evaluating marimo as an alternative to streamlit / python shiny for sharing analysis etc. One thing that puzzled me is I created a cell containing a sql statement, for example

```
select s.site_name, e.equipment_name from schema.sites as s join schema.equipment as e using (site_id)
```

This generated a reactive function that accepted the expected "engine" argument, but also marimo appears to have parsed my sql statement and generated arguments to the generated function, one for each table and schema i.e. ["sites", "equipment", "schema" ...] etc. Is this expected?

This became problematic because I tend not to follow the convention of adding a type suffix to all my variables (i.e. "sites_df" etc). So when in a later cell I tried to create a list of sites for an `mo.ui.dropdown` I did something like

```
sites = df["site_name"].unique().sort().to_list()
sites_dropdown = mo.ui.dropdown(options=sites)
```

This complained that there was a cycle - there was no way of telling why until I opened the .py file and inspected it - I then saw that sites was both an argument to the sql cell (generated based on analysing the sql statement?) and a variable I declared.

The fact that marimo generates variables then hides them from the user - but then complains when the user creates a variable of the same name - is very confusing. Is this expected or is it a bug. I'm using postgres sql with sqlalchemy</div>

Light2Dark:
## 📝 Summary

**Note for bash focus: There might be a perf hit with this PR because we iterate through all definitions.**

- Get SQLRefs
- Fuzzy find other cells that define the same name for SQL cells
- Fuzzy find referring cells

Fixes #5945 . We now pull the qualified names from sql cells instead of catalog, schema and tables separately.
Also remove duckdb.get_table_names as it's redundant.

## 📋 Checklist


marimo/_ast/sql_visitor.py
```diff
@@ -311,17 +311,40 @@ def find_sql_defs(sql_statement: str) -> SQLDefs:
...
+        name = ""
+        if self.catalog is not None:
+            name += self.catalog + "."
+        if self.schema is not None:
+            name += self.schema + "."
+        if self.table is not None:
+            name += self.table
+        return name
+
+    def only_table(self) -> bool:
```
mscolnick: `is_only_table()` or `is_bare_table()`?

marimo/_ast/sql_visitor.py
```diff
@@ -332,44 +355,48 @@ def find_sql_refs(
...
+            table=table_name, schema=schema_name, catalog=catalog_name
+        )
 
     try:
         expression_list = parse(sql_statement, dialect="duckdb")
     except ParseError as e:
         LOGGER.error(f"Unable to parse SQL. Error: {e}")
         return []
 
+    refs: list[SQLRef] = []
```
mscolnick: if you make this a `set`, it should remove duplicates by default

marimo/_runtime/dataflow.py
```diff
@@ -96,11 +96,13 @@ def get_referring_cells(
...
-                for cid, cell in self.cells.items()
-                if name in cell.refs and cell.language == "sql"
-            }
+            cells = set()
+            for cid, cell in self.cells.items():
+                if cell.language == "sql":
+                    for ref in cell.refs:
+                        if name == ref or ("." in ref and name in ref):
+                            cells.add(cid)
+            return cells
```
mscolnick, line 118: btw, the current bugfix referenced in the PR description is in the `else` statement im pretty sure; the failing code is `python`

tests/_ast/test_sql_visitor.py
```diff
@@ -758,20 +815,28 @@ def test_dml_with_subquery() -> None:
...
+
+    def test_read_file_and_urls(self) -> None:
+        sql = "SELECT * FROM 'file.csv'"
+        assert find_sql_refs(sql) == [SQLRef(table="file.csv")]
+
+        sql = "SELECT * FROM 'https://example.com/file.csv'"
         assert find_sql_refs(sql) == [
-            "pdb",
-            "database",
+            SQLRef(table="https://example.com/file.csv")
```
mscolnick: maybe an existing bug, but should these not be picked up?

tests/_runtime/test_dataflow.py
```diff
@@ -703,6 +742,433 @@ def test_attached_db(self):
         # cell
         assert graph.get_referring_cells("my_db", language="sql") == set(["1"])
 
+    def test_get_referring_cells_sql_fuzzy_matching(self) -> None:
```
mscolnick: could we write some tests (even if xfail) that capture some DB hierarchy:

it might be also easier to write them out as a dataclasses

```
class TestCase:
   cells: list[str]
   expected_defs: list[list[Variable]]
   expected_refs: list[list[Variable]]
   edges: list[str]
```

e.g.
1. bare table reference depends on table creation (with schema)
2. bare table reference depends on table creation (with catalog/schema)
3. bare table reference depends on table creation (bare table)
3. table (with schema) reference depends on bare table creation 
4. table (with catalog/schema) reference depends bare  on table creation
5. table (with schema) reference depends on table creation (with another schema)

then we can do some for python as well

tests/_ast/test_sql_visitor.py
Light2Dark: I added some validation logic to filter these out. But turns out, thesd are valid table names lol. As long as they are quoted. CREATE TABLE 'https://...'

marimo/_ast/codegen.py
```diff
@@ -210,7 +213,16 @@ def to_functiondef(
     # already in globals)
     if allowed_refs is None:
         allowed_refs = BUILTINS
-    refs = tuple(ref for ref in sorted(cell.refs) if ref not in allowed_refs)
+
+    # Filter out refs that are not valid Python identifiers
```
Light2Dark: added this to fix an issue where function signatures would include the qualified sql ref

eg: `def _(mo, schema1.table_name)`

dmadisetti: Can you leave this as a comment as well?

marimo/_ast/sql_visitor.py
```diff
@@ -311,17 +311,49 @@ def find_sql_defs(sql_statement: str) -> SQLDefs:
...
+    # Tables are synonymous with views,
+    # since we can't know the difference in queries
+    table: str
+    schema: Optional[str] = None
+    catalog: Optional[str] = None
+
+    def convert_to_name(self) -> str:
+        """Convert a SQLRef to a name to be used as a reference in the visitor"""
+        name = ""
+        if self.catalog is not None:
```
dmadisetti, line 336: Maybe a join? because as is you can have `catalog.schema.`

/ remove the check table is None

Light2Dark: it may look something like this, I think it's messier & doesn't cover all edge cases.

also table shouldn't be None for sqlrefs (or query like statements), hence we log a warning and return None.

```python
        name = ".".join(
            [self.catalog, self.schema, self.table]
            if self.catalog is not None
            else [self.schema, self.table]
            if self.schema is not None
            else [self.table]
        )
```

dmadisetti:
```python
representation = []
if self.catalog is not None:
    representation.append(self.catalog)
if self.schema is not None:
    representation.append(self.schema)
# Table cannot be None since typed str
representation.append(self.table)
return ".".join(representation)
```

Light2Dark: ah thanks! that makes sense

marimo/_ast/sql_visitor.py
```diff
@@ -332,45 +363,54 @@ def find_sql_refs(
...
+
+        if table_name is None:
+            LOGGER.warning("Table name cannot be found in the SQL statement")
+            return None
+
+        # Check if the table name looks like a URL or has a file extension.
+        # These are often not actual table references, so we skip them.
+        # Note that they can be valid table refs, but for now we skip them.
+        if table_name.startswith(
+            ("http://", "https://", "ftp://")
```
mscolnick: `hf://`, `s3://`, `gcs://` are some other common ones

marimo/_ast/sql_visitor.py
```diff
@@ -332,45 +363,54 @@ def find_sql_refs(
...
+        ) or table_name.endswith(
+            (".csv", ".parquet", ".json", ".txt", ".db", ".tsv", ".xlsx")
```
mscolnick: i would maybe pull out both of these lists

marimo/_ast/sql_visitor.py
dmadisetti: `"://" in table_name`

marimo/_ast/sql_visitor.py
dmadisetti: Still a useful list, but "`.`" isn't a valid table character by itself so you might be able to just check for the char

marimo/_ast/visitor.py
```diff
@@ -54,6 +54,15 @@ class AnnotationData:
     refs: set[Name] = field(default_factory=set)
 
 
+@dataclass
+class RefsData:
```
dmadisetti: Wouldn't it be better to expose SQLRef directly? Seems a littl verbose if we are only exposing sql data. Ok with keeping this in, but wondering if there's another RefsData type that would be wrapped

marimo/_runtime/dataflow.py
```diff
@@ -96,17 +97,108 @@ def get_referring_cells(
...
+
+                for ref in cell.refs:
+                    # Direct reference match
+                    if ref == name:
+                        cells.add(cid)
+                        break
+
+                    ref_data = cell.refs_data[ref]
+
+                    # Hierarchical reference match
```
dmadisetti, line 113: `ref_data.matches(ref) -> bool` ? Would be nice to isolate for tests if we need

marimo/_ast/sql_visitor.py
Light2Dark: it can be, i'll update the comment but table_names can have weird chars.
`://` is valid though, very uncommon in table name

dmadisetti (commented):
LGTM

Would like to see a name shadowing case in tests, because not certain what the behavior will exactly be:

`my_table = mo.sql(f'CREATE TABLE schema_one.my_table (name STRING)')"`

`schema_one = mo.sql(f'CREATE TABLE schema_one.my_table (name STRING)')"`

marimo/_ast/visitor.py
Light2Dark: there is `RefData` which is similar but has some more stuff.

Light2Dark:
I think a more robust solution involves having SQLDefs be fully qualified names. or a list of SQLDef. Edge cases will fail with this solution.

marimo/_ast/visitor.py
```diff
@@ -591,7 +620,7 @@ def visit_Call(self, node: ast.Call) -> ast.Call:
             first_arg = node.args[0]
             sql: Optional[str] = None
             if isinstance(first_arg, ast.Constant):
-                sql = first_arg.s
+                sql = first_arg.value
```
dmadisetti, line 613: Nice! Thanks. the only instance I found

marimo/_ast/cell.py
```diff
@@ -156,6 +156,8 @@ class CellImpl:
     mod: ast.Module
     defs: set[Name]
     refs: set[Name]
+    # metadata about refs, currently only tracks SQL refs
+    refs_data: dict[Name, SQLRef]
```
dmadisetti: Name update?

marimo/_runtime/dataflow.py
```diff
@@ -96,17 +97,87 @@ def get_referring_cells(
...
+                    # Direct reference match
+                    if ref == name:
+                        cells.add(cid)
+                        break
+
+                    sql_ref = cell.refs_data.get(ref)
+
+                    # Hierarchical reference match
+                    if sql_ref and name in ref and "." in ref:
+                        if SQLRef.matches_hierarchical_ref(ref, sql_ref):
```
dmadisetti: Oh woops. Not this, but `sql_ref.matches_hierarchical_ref(name)`?
My bad, I think something was lost in communication.

```python
if sql_ref and sql_ref.matches_hierarchical_ref(name):
    cells.add(cid)
    break
```

Also `name in ref` potentially has the issue of `name=name; ref=my_base_name.not_name` so the comparison should be more granular anyway

marimo/_runtime/dataflow.py
```diff
@@ -196,11 +267,27 @@ def register_cell(self, cell_id: CellId_t, cell: CellImpl) -> None:
                     if name in self.definitions
                     else set()
                 ) - set((cell_id,))
+
+                # Handle SQL matching for hierarchical references
+                name_map: dict[Name, Name] = {}
+                sql_ref = cell.refs_data.get(name)
+                if (
```
dmadisetti:
```python
other_ids_defining_name, variable_name = sql_ref.extract_hierarachial_matches(tables=self.table_defs, views=self.views_defs, catalog=self.catalog_defs) # Directly testable

...

@property
def variable_kinds(self) -> dict[Name, str]: # Name, kind
    ...

@property
def table_defs(self) -> set[Name]:
    return {name for name, kind in self.variable_kinds.items() if kind == "table"}

...
```

Like I said, just a bit of tidy- but I don't want this to be a major blocker. PR over all looks solid, I just think we can move some of the sql specific stuff directly to sqlref, and avoid the name splitting since sqlref contains that information.

marimo/_runtime/dataflow.py
Light2Dark: oops, yes, it should be like that. thanks!

marimo/_runtime/dataflow.py
Light2Dark: I might hold off on optimizing this function. I think it will change in the future and so it may look different. I also found it used a lot of parent state hence why it wasn't easy to move over.

dmadisetti:
I'm good with this, there's a refactor to `find_sql_defs` such that the qualified names are preserved for proper redefinition detection / shadow handling, but I think @Light2Dark did a good job documenting the test cases with tests known to fail.

mscolnick (approved):
nice work!

PR #5996, opened 2025-08-11
fix: stub and restore cache meta values 

dmadisetti:
replaces #5933

---

## 📝 Summary

closes https://github.com/marimo-team/marimo/issues/5930

Process `self.meta["return"]` in the same way that `contextual_defs` are handled in `cache.py` so that when caching and restoring UI elements they are re-registered with the `UIElementRegistry`

## 🔍 Description of Changes

https://github.com/marimo-team/marimo/issues/5930 describes a bug from `UIElement`s not being re-registered when they are returned from a caching function.

It seemed like handling `UIElement`s (and some other classes) already handled doing this correctly when it came to "contextual_defs" by using a pattern of storing and loading "stub" classes, so I figured that using that same process for the `return` value of the cached function should work as well. It was either that or explicitly re-register `UIElement`s on cache hit, which seemed like a worse solution.

I tried adding some tests for this but got kinda stuck on how to assert that the old and new ui elements were "the same" since the stub loading process technically creates a new instance. If we need tests for this, I would appreciate some advice.

## 📋 Checklist

### DIFF
--- a/marimo/_save/cache.py
+++ b/marimo/_save/cache.py
@@ -161,10 +161,15 @@ class Cache:
                 for item in value
             )
         elif isinstance(value, set):
+            # Sets cannot be recursive (require hashable items), but keep the
+            # reference.
             result = set(
                 self._restore_from_stub_if_needed(item, scope, memo)
                 for item in value
             )
+            value.clear()
+            value.update(result)
+            result = value
         elif isinstance(value, list):
             memo[obj_id] = value
             result = [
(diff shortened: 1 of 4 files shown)


marimo/_save/cache.py
```diff
@@ -136,6 +134,61 @@ def restore(self, scope: dict[str, Any]) -> None:
...
+            )
+        elif isinstance(value, list):
+            memo[obj_id] = value
+            result = [
+                self._restore_from_stub_if_needed(item, scope, memo)
+                for item in value
+            ]
+            # Keep the original list reference
+            value.clear()
+            value.extend(result)
```
mscolnick, line 181: why are you keeping the original list, but not doing this for tuple or sets?

tests/_save/test_cache.py
```diff
@@ -10,11 +10,222 @@
...
+
+    @staticmethod
+    def test_cache_scope_recursive() -> None:
+        _list = []
+        _list.append(_list)
+        d = {}
+        d["self"] = d
+        scope = {
+            "_list": _list,
+            "_dict": d,
```
mscolnick, line 65: maybe worth having recursive set and tuple as well

marimo/_save/cache.py
dmadisetti: sets and tuples are non-hashable, but require hashable inputs- so you can't have the same recursive pattern. I did add the tuple/ set specific cases in `test_cache_iterable`. I did add a comment though

tests/_save/test_cache.py
dmadisetti: Added a test to ensure references are preserved, see other comment regarding recursive refs

dmadisetti:
Thanks @buckley-w-david

PR #6000, opened 2025-08-11
fix: pandas convert non-string columns to string columns

### LINKED ISSUE #5981: No displayed data for a non-empty Pandas data set
### Describe the bug

I tried to use a Marimo notebook for the most recent Bamboo Weekly. I found that while I could load the Excel file into Pandas, I didn't see any of the data when running `df.head()`. I finally used `to_json` to see that there was indeed data -- and I kept getting error output on my terminal from Marimo, ending with " does not fit any type of the union". (Oodles of text before then!)

[marimo-bw-130.zip](https://github.com/user-attachments/files/21706268/marimo-bw-130.zip)
[bw-130-payems-2.xlsx](https://github.com/user-attachments/files/21706256/bw-130-payems-2.xlsx)

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.16",
  "editable": false,
  "OS": "Darwin",
  "OS Version": "24.5.0",
  "Processor": "i386",
  "Python Version": "3.13.5",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.5.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.0.1",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "anywidget": "0.9.18",
    "nbformat": "5.10.4",
    "pandas": "2.3.1",
    "polars": "1.32.2",
    "pyarrow": "21.0.0",
    "loro": "1.5.3",
    "pytest": "8.4.1"
  },
  "Experimental Flags": {
    "rtc_v2": true
  }
}

```

</details>

### Code to reproduce

_No response_

mscolnick:
Fixes #5981 

Pandas can have non-string index column values such as `int` and `pd.Timestamp`. This breaks a lot of downstream assumptions around string column names. 

There isn't much (or any benefit) in keeping them as `int` or `Timestamp` when displaying, so instead we convert them to string.

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/table_manager.py
+++ b/marimo/_plugins/ui/_impl/tables/table_manager.py
@@ -121,9 +121,6 @@ class TableManager(abc.ABC, Generic[T]):
     def select_columns(self, columns: list[str]) -> TableManager[Any]:
         pass
 
-    def clamp_columns(self, num_columns: int) -> TableManager[Any]:
-        return self.select_columns(self.get_column_names()[:num_columns])
-
     @abc.abstractmethod
     def select_cells(self, cells: list[TableCoordinate]) -> list[TableCell]:
         pass


marimo/_plugins/ui/_impl/tables/table_manager.py
```diff
@@ -121,6 +121,9 @@ def select_rows(self, indices: list[int]) -> TableManager[Any]:
     def select_columns(self, columns: list[str]) -> TableManager[Any]:
         pass
 
+    def clamp_columns(self, num_columns: int) -> TableManager[Any]:
```
Light2Dark: this isn't used

PR #6003, opened 2025-08-11
improvement: use fully qualified model ids

mscolnick:


frontend/src/utils/ai/__tests__/ids.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      });
+
+      it("should parse bedrock qualified id", () => {
+        const modelId = AiModelId.parse("bedrock/titan-text");
+        expect(modelId.providerId).toBe("bedrock");
+        expect(modelId.shortModelId).toBe("titan-text");
+      });
+
+      it("should handle multiple slashes by taking first two parts", () => {
+        const modelId = AiModelId.parse("openai/gpt-4/turbo");
```
Light2Dark, line 70: should this logic be the same as backend. provider = openai, model = gpt-4/turbo

mscolnick: oh you are right, this is a bug

marimo/_server/ai/ids.py
```diff
@@ -0,0 +1,62 @@
+from dataclasses import dataclass
+from typing import NewType
+
+from marimo import _loggers
+
+LOGGER = _loggers.marimo_logger()
+
+AiProviderId = NewType("AiProviderId", str)
+QualifiedModelId = NewType("QualifiedModelId", str)
+ShortModelId = NewType("ShortModelId", str)
```
Light2Dark, line 10: any reason to have this strong typing?

mscolnick: like enumerate all the IDs?

Light2Dark: ah, i just felt it was overengineered to write this: 
```python
AiModelId(provider=AiProviderId(provider), model=ShortModelId(short_id))
```
and we created tests to check if it's a string, so feels over eng'd.

i imagine this would be the same: `AiModelId(provider=provider, model=short_id)`

mscolnick: it can be useful to pass to downstream consumers to make sure that the `ShortModelId` is passed and not the fully qualified one

mscolnick: i can follow-up to either make this useful or remove

PR #6005, opened 2025-08-11
@marimo-team/llm-info package

mscolnick:
This PR adds a new package `@marimo-team/llm-info`. This is mainly to keep things organized and make it easier for contributions. This is a private package and not shared to npm (although could do that if people wanted).

This does not add all the models or providers we use, but mainly a placeholder to start to add more and let others jump in.

This PR also adds:
1. `@marimo-team/tsconfig` for shared config files
2. Adds icons in the settings page as an example:

Screenshot 2025-08-11 at 2 40 27 PM


packages/llm-info/data/models.yml
```diff
@@ -0,0 +1,41 @@
+- name: Claude 3.5 Haiku
```
mscolnick, line 5: non exhaustive (but not used yet)

packages/llm-info/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@marimo-team/llm-info",
```
mscolnick, line 2: open to the name `@marimo-team/ai-info` as well

mscolnick:
@manzt @Light2Dark, i need to build off of this in another branch so going to merge. happy to followup with comments

PR #5938, opened 2025-08-06
preview: Improved SQL completions, linting, and hover

Light2Dark:
## 📝 Summary

Adds the integration under a feature flag.

https://github.com/user-attachments/assets/2dad658f-cb58-4fcb-adcc-f4122db3c945

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [ ] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

PR #6033, opened 2025-08-13
chore: move to rolldown

mscolnick:
Migrate `vite` → `rolldown-vite`

This does not take full advantage of rolldown as we still need to use babel for react-compiler. But other things like resolving and alias can at least be offloaded to a faster and more memory-efficient tool.

### DIFF
--- a/frontend/islands/validate.sh
+++ b/frontend/islands/validate.sh
@@ -9,10 +9,10 @@
 OUT_DIR=$(pwd)/dist
 echo "validating $OUT_DIR"
 
-echo "[validate: no process.env variables in any of the js files]"
-grep -R "process.env" $(pwd)/dist
+echo "[validate: no process.env. variables in any of the js files]"
+grep -R "process.env." $(pwd)/dist
 if [ $? -eq 0 ]; then
-  echo "process.env variables found in js files"
+  echo "process.env. variables found in js files"
   exit 1
 fi
 


frontend/islands/vite.config.mts
```diff
@@ -3,10 +3,8 @@
 import fs from "node:fs";
 import path from "node:path";
 import react from "@vitejs/plugin-react";
-import { defineConfig, type Plugin } from "vite";
-import topLevelAwait from "vite-plugin-top-level-await";
+import { defineConfig, type Plugin } from "rolldown-vite";
```
manzt: Since rolldown-vite will eventually be deprecated in an upcoming release of `vite`, you can do:

```json
  "dependencies": {
    "vite": "npm:rolldown-vite@latest",
  }
```

In the `package.json`. That will lock the specific version of `rolldown-vite` (in our pnpm-lock.yml), but keep `vite` as an alias. That way we can keep imports to:

```ts
import { defineConfig, type Plugin } from "vite";
```

and also run `pnpm vite` to run the rolldown-vite cli.

frontend/islands/vite.config.mts
```diff
@@ -31,11 +29,13 @@ const ReactCompilerConfig = {
 // https://vitejs.dev/config/
 export default defineConfig({
   resolve: {
+    alias: {
+      "@": path.resolve(__dirname, "../src"),
+    },
```
manzt: `rolldown-vite` natively supports resolution using the `paths` from tsconfig:

  ```ts
resolve: {
    tsconfigPaths: true,
  },
  experimental: {
    enableNativePlugin: true,
  },
```

This requires the `enableNativePlugin=true` currently.

File: frontend/vite.config.mts [hunk 3]
```diff
@@ -192,6 +191,9 @@ const ReactCompilerConfig = {
 export default defineConfig({
   // This allows for a dynamic <base> tag in index.html
   base: "./",
+  // experimental: {
+  //   enableNativePlugin: true,
+  // },
```
manzt: Oh I guess you maybe tried it here. Did this not work? I was using the native resolution in marimo-lsp.

package.json
```diff
@@ -19,10 +19,20 @@
...
+  "packageManager": "pnpm@10.12.4",
   "pnpm": {
     "patchedDependencies": {
       "react-plotly.js": "patches/react-plotly.js.patch"
+    },
+    "overrides": {
+      "react": "^19.1.0",
+      "react-dom": "^19.1.0",
+      "vite": "npm:rolldown-vite@latest"
+    },
```
manzt, line 31: Ah, I see the overrides here. I think this will also work, would just be nice to have the imports in the vite configs still from `vite`.

manzt: I think with `pnpm` you only need `overrides` to override another package's import for `vite`. Since we just use it at the top level, it can go under `dependencies` /  `devDependencies` directly.

frontend/vite.config.mts
mscolnick: `enableNativePlugin: "v1"`  is the default, which is currently the same as `enableNativePlugin: true,`, but i am adding `enableNativePlugin: true,` back

package.json
mscolnick: yep, moved to that

mscolnick:
The main bundle builds fine, but does not run due to a loro/wasm error:

```
loro_wasm_bg-BN7J7_cH.js:2 panicked at crates/loro-internal/src/configure.rs:121:36:
called `Result::unwrap()` on an `Err` value: Error { internal_code: 2147483655, description: "Web Crypto API is unavailable" 
```

Can investigate later, but going to deprioritize this. This was to get https://github.com/marimo-team/marimo/pull/5938 building, but maybe there is a shorter path.

Issue #4572, opened 2025-04-17
Support for XDG Base Directory specification


AliSPEDA:
### Description

#2367 made configuration support XDG_CONFIG_HOME, which is great! 
On that PR there was a bit of talk of broader support for XDG Base Dirs, but I can't tell what the wish/plans are for this.

Currently I can see some state information (state.toml) in ~/.marimo, and this directory is probably also used for other files? (I recently deleted the directory in a clean up and have only used marimo a couple of times since)

### Suggested solution

I would love if marimo respected XDG environment variables if set and otherwise fallback on ~/.marimo (or anything else is preferred). But will there be support for such a change? 
Would love to take a shot at a PR, if maintainers would like/accept such a change.

### Additional context

https://wiki.archlinux.org/title/XDG_Base_Directory

akshayka:
Thanks for opening this issue. We're open to it, but what specifically did you have in mind? Moving ~/.marimo/ to XDG_CACHE_HOME, or something else?

a-alak:
(Switching to my private GitHub account, posted from work account by mistake 😅)

Yes, or whatever other XDG directory makes more sense.
Currently I only have state.toml in there, with data that looks more cache like than state like.

But also if marimo writes any other file in ~/.marimo I think it should go to the right XDG directory (CACHE, STATE, DATA, CONFIG) if the XDG environment variables is set.

twio142:
I'd cast a vote for XDG_STATE_HOME

PR #6048, opened 2025-08-15
Using XDG_STATE_HOME environment variable to support the Linux XDG Base Directory

TDYQ-Liu:
## 📝 Summary

Closes https://github.com/marimo-team/marimo/issues/4572

The default root dir is ".marimo" and cannot be modified, it will autocreate ".marimo" folder at the user's home in Linux platform.

Add the environment variable XDG_STATE_HOME to support the Linux XDG Base Directory.

## 🔍 Description of Changes

1. Get XDG_STATE_HOME environment variable as ROOT_DIR value
2. If it is not set by the user, using the default XDG_STATE_HOME value “~/.config/state”.
3. If the user's system is Windows or others, using default value ".marimo".
4. I don't know if there is a standard for Dir in Windows or other systems, so I use the default value and make it easy for other contributors to modify it.

## 📋 Checklist

### DIFF
--- a/marimo/_utils/config/config.py
+++ b/marimo/_utils/config/config.py
@@ -3,13 +3,20 @@ from __future__ import annotations
 
 import os
 from dataclasses import asdict
+from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Optional, TypeVar
 
 from marimo._utils.parse_dataclass import parse_raw
 from marimo._utils.toml import is_toml_error, read_toml
 
-ROOT_DIR = os.environ.get("MARIMO_ROOT_DIR", ".marimo")
+if os.name == "posix":
+    # for Linux/macOS/Unix
+    _state_home = os.environ.get("XDG_STATE_HOME", Path.home() / ".local/state")
+    ROOT_DIR = Path(_state_home) / "marimo"
+elif os.name == "nt":
+    # for Windows
+    ROOT_DIR = Path.home() / ".marimo"
 
 T = TypeVar("T")
 
--- a/marimo/_utils/config/config.py
+++ b/marimo/_utils/config/config.py
@@ -9,10 +9,7 @@ from typing import Any, Optional, TypeVar
 from marimo._utils.parse_dataclass import parse_raw
 from marimo._utils.toml import is_toml_error, read_toml
 
-if (_root_dir := os.environ.get("MARIMO_ROOT_DIR")) is None:
-    ROOT_DIR = ".marimo"
-else:
-    ROOT_DIR = _root_dir
+ROOT_DIR = os.environ.get("MARIMO_ROOT_DIR", ".marimo")
 
 T = TypeVar("T")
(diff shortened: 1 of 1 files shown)


marimo/_utils/config/config.py
```diff
@@ -9,7 +9,10 @@
 from marimo._utils.parse_dataclass import parse_raw
 from marimo._utils.toml import is_toml_error, read_toml
 
-ROOT_DIR = ".marimo"
+if (_root_dir := os.environ.get("MARIMO_ROOT_DIR")) is None:
+    ROOT_DIR = ".marimo"
+else:
+    ROOT_DIR = _root_dir
```
mscolnick: 1. what is the use case for this?
2. does `ROOT_DIR = os.environ.get("MARIMO_ROOT_DIR", ".marimo")` work instead?

TDYQ-Liu: You’re right it works and is simpler. I’ve updated the commit accordingly, thanks!

mscolnick: what is the use case for this?

TDYQ-Liu: > The default root dir is ".marimo" and cannot be modified, it will autocreate ".marimo" folder at the user's home in Linux platform.

I don't want it in my home. I hope it can be changed to another location in order to apply XDG Base Dir.

mscolnick: Oh I didn't realize it went to your home dir. it should go to your xdg dir? Maybe that is not set?

TDYQ-Liu: I found the ".marimo” folder contains some files like recently opened files or states file. I tried to change it to the default xdg config dir `~/.config/marimo/`, then run marimo, it could work normally.

mscolnick: Yeah, we want to move those there, but haven't gotten around to it. 

would you be open to making that change, instead? something like:

```python
if (_root_dir := os.environ.get("XDG_CONFIG_HOME")) is None:
     ROOT_DIR =  Path(_root_dir) / "marimo"
else:
    ...old path
```

mscolnick: actually here is the original info. https://github.com/marimo-team/marimo/issues/4572

`XDG_STATE_HOME` seems like correct variable

TDYQ-Liu: Got it. I’ve updated the commit accordingly.

mscolnick:
@TDYQ-Liu looks great! do you mind just replying with only "I have read the CLA Document and I hereby sign the CLA".

also, could you update the title and description to match the new changes

TDYQ-Liu:

mscolnick:
@TDYQ-Liu, apologies for the followups. it looks like a test is failing:

```
FAILED tests/_cli/test_upgrade.py::test_version_comparison_edge_cases - AssertionError: Expected 'echo' to have been called.
```

PR #6075, opened 2025-08-19
fix: clean up finder/module logic for recursion

### LINKED ISSUE #6072: RecursionError after too many refresh (deployed using FastApi)
### Describe the bug

Hi team

so we are using Marmio server deployed following FastAPI guide: https://docs.marimo.io/guides/deploying/programmatically/#fastapi-example

We found that starting (or simply refreshing) a notebook too many times causing **RecursionError** like below, when a new module is imported.

```
    from . import types
  File \"<frozen importlib._bootstrap>\", line 1360, in _find_and_load
  File \"<frozen importlib._bootstrap>\", line 1322, in _find_and_load_unlocked
  File \"<frozen importlib._bootstrap>\", line 1262, in _find_spec
  File \"/opt/app-root/.pixi/envs/production/lib/python3.12/site-packages/marimo/_output/formatters/formatters.py\", line 156, in find_spec
    spec = original_find_spec(fullname, path, target)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File \"/opt/app-root/.pixi/envs/production/lib/python3.12/site-packages/marimo/_output/formatters/formatters.py\", line 156, in find_spec
    spec = original_find_spec(fullname, path, target)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File \"/opt/app-root/.pixi/envs/production/lib/python3.12/site-packages/marimo/_output/formatters/formatters.py\", line 156, in find_spec
    spec = original_find_spec(fullname, path, target)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 288 more times]
RecursionError: maximum recursion depth exceeded", 
```

so we traced it down to [formatters.py](https://github.com/marimo-team/marimo/blob/0.14.17/marimo/_output/formatters/formatters.py#L190), as `find_spec` is wrapped each time this function is called.

Eventually, it will be recursively wrapped more than python's recursion limit. When this happens, importing a new module causes `find_spec` to be called, and each of the wrapped layer is now being called, thus breaching the limit.

We also noticed every time we start/refresh a notebook, another instance of `MicropipFinder` is appended to `sys.meta_path`

```
# printing sys.meta_path
[
<_distutils_hack.DistutilsMetaFinder object at 0x7f54f2f72d20>
<class '_frozen_importlib.BuiltinImporter'>
<class '_frozen_importlib.FrozenImporter'>
<class '_frozen_importlib_external.PathFinder'>
<six._SixMetaPathImporter object at 0x7f54f1f08c80>
<importlib_metadata.MetadataPathFinder object at 0x7f54f0cf9fa0>
<__main__._MicropipFinder object at 0x7f54e7f56450>
<__main__._MicropipFinder object at 0x7f54e468ed20>
<__main__._MicropipFinder object at 0x7f54ec197e90>
<__main__._MicropipFinder object at 0x7f54e4291e80>
<__main__._MicropipFinder object at 0x7f54e412b470>
<__main__._MicropipFinder object at 0x7f54e7f0e0c0>
]
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.17",
  "editable": false,
  "OS": "Linux",
  "OS Version": "4.18.0-553.64.1.el8_10.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.13.5",
  "Binaries": {
    "Browser": "138.0.7204.183",
    "Node": "v20.11.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.1.2",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.5.4",
    "ruff": "0.12.9"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

dmadisetti:
## 📝 Summary

Fixes #6072

Register formatter is called multiple times when running as an app. The expectation of `register_formatters` and module patching is that it they are only called once. Added some recursion guards and testing to better handle multiple calls

### DIFF
--- a/marimo/_runtime/functions.py
+++ b/marimo/_runtime/functions.py
@@ -40,19 +40,13 @@ class Function(Generic[S, T]):
         arg_cls: type[S],
         function: Callable[[S], T],
     ) -> None:
-        from marimo._runtime.context import (
-            ContextNotInitializedError,
-            get_context,
-        )
+        from marimo._runtime.context import safe_get_context
 
         self.name = name
         self.arg_cls = arg_cls
         self.function = function
 
-        try:
-            ctx = get_context()
-        except ContextNotInitializedError:
-            ctx = None
+        ctx = safe_get_context()
 
         if ctx is not None and ctx.execution_context is not None:
             self.cell_id = ctx.execution_context.cell_id
--- a/tests/_plugins/stateless/test_routes.py
+++ b/tests/_plugins/stateless/test_routes.py
@@ -1,8 +1,6 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
-import gc
-
 from marimo._runtime.context import get_context
 from marimo._runtime.context.types import RuntimeContext
 from marimo._runtime.runtime import Kernel
@@ -28,7 +26,6 @@ async def test_routes_lazy(k: Kernel, exec_req: ExecReqProvider) -> None:
         ]
     )
 
-    gc.collect()
     context: RuntimeContext = get_context()
     # 4 functions, 1 for each route
     assert len(context.function_registry.namespaces.values()) == 4


marimo/_runtime/patches.py
```diff
@@ -143,7 +143,13 @@ def set_index_urls(urls):
     # append the finder to the end of meta_path, in case the user
     # already has a package called micropip
     exec(
-        "import sys; sys.meta_path.append(_MicropipFinder()); del sys",
+        textwrap.dedent(r"""
+import sys
+# If micropip is already in sys.meta_path, we don't need to add it
+# again.
+if not sys.meta_path or sys.meta_path[-1].__class__.__name__ != '_MicropipFinder':
```
mscolnick: can we skip all this logic above as well? could we just add an annotaiton `@once`

mscolnick:
unrelated, but do we need `patch_webbrowser` outside wasm?

dmadisetti:
re webbrowser: It's a fallback for no webbrowser at all too (e.g. in docker etc)

marimo/_runtime/patches.py
dmadisetti: Yep, good call. Changed to jump out early- but since explicitly use this function in testing, not opting for `@once`

marimo/_output/formatters/formatters.py
```diff
@@ -84,6 +84,72 @@
...
+        third_party_factories = THIRD_PARTY_FACTORIES
+    # Note: "Vendored" dependencies may not have a find_spec method.
+    # E.g. `six` bundled with a project.
+    original_find_spec = getattr(finder, "find_spec", None)
+    if original_find_spec is None:
+        return
+
+    # We include `original_find_spec` as a kwarg to force it to be bound
+    # to the new `find_spec` method; this is needed because closures are
+    # late-binding and we're in a for loop ...
```
akshayka: This comment is now stale? No longer in a for loop

marimo/_output/formatters/formatters.py
```diff
@@ -84,6 +84,72 @@
...
+        spec = original_find_spec(fullname, path, target)
+        if spec is None:
+            return spec
+
+        if spec.loader is not None and fullname in third_party_factories:
+            # We're now in the process of importing a module with
+            # an associated formatter factory. We'll hook into its
+            # loader to register the formatters.
+            original_exec_module = spec.loader.exec_module
+            factory = THIRD_PARTY_FACTORIES[fullname]
```
akshayka: Should this be `third_party_factories`?

marimo/_output/formatters/formatters.py
```diff
@@ -84,6 +84,72 @@
...
+
+        return spec
+
+    # Recursive wrap can lead to a stack overflow in long running apps.
+    # `find_spec` is dynamic, so just compare the __module__s
+    module_name = getattr(finder.find_spec, "__module__", None)
+    if hasattr(finder.find_spec, "__func__"):
+        module_name = finder.find_spec.__module__
+
+    if module_name != find_spec.__module__:
```
akshayka, line 160: Suggested comment to see if my understanding is correct

```suggestion
    # Only patch the finder if the module it was defined in is
    # different from the current module (`find_spec.__module__`)
    # i.e., only patch the finder if it isn't already patched
    if module_name != find_spec.__module__:
```

marimo/_output/formatters/formatters.py
dmadisetti: +1 The work around isn't even needed anymore

marimo/_output/formatters/formatters.py
dmadisetti: third_party_factories is a subset of THIRD_PARTY_FACTORIES, but yes

marimo/_output/formatters/formatters.py
```diff
@@ -84,6 +84,85 @@
...
+
+        return spec
+
+    # Recursive wrap can lead to a stack overflow in long running apps.
+    # `find_spec` is dynamic, so just compare the __module__s
+    module_name = getattr(finder.find_spec, "__module__", None)
+    if hasattr(finder.find_spec, "__func__"):
+        module_name = finder.find_spec.__module__
+
+    # Only patch the finder if the module it was defined in a
```
akshayka:
```suggestion
    # Only patch the finder if the module it was defined in is
```

akshayka (dismissed):
LGTM if tests are passing

dmadisetti:
Can't repro the failure locally. Doesn't look related, but is consistent. Will dig into it a bit more, and follow up with a flakey PR if that seems to be the case

akshayka:
> Can't repro the failure locally. Doesn't look related, but is consistent. Will dig into it a bit more, and follow up with a flakey PR if that seems to be the case

It's an odd failure. But if it's consistent and not reflected in other PRs, then worth digging into instead of marking as flakey?

PR #6079, opened 2025-08-19
fix: Don't do health check in static notebooks

dmadisetti:
## 📝 Summary

Health check in static notebooks is blocking and invalid

### DIFF
--- a/frontend/src/core/runtime/config.ts
+++ b/frontend/src/core/runtime/config.ts
@@ -1,9 +1,9 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 import { atom, useAtomValue } from "jotai";
+import { isStaticNotebook } from "@/core/static/static-state";
 import { store } from "../state/jotai";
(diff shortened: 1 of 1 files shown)


frontend/src/core/runtime/config.ts
```diff
@@ -18,7 +19,10 @@ export const DEFAULT_RUNTIME_CONFIG: RuntimeConfig = {
 export const runtimeConfigAtom = atom<RuntimeConfig>(DEFAULT_RUNTIME_CONFIG);
 const runtimeManagerAtom = atom<RuntimeManager>((get) => {
   const config = get(runtimeConfigAtom);
-  return new RuntimeManager(config);
+  // "lazy" means that the runtime manager will attempt to connect to a
```
akshayka: will "not"?

```suggestion
  // "lazy" means that the runtime manager will not attempt to connect to a
```

akshayka (approved):

mscolnick (approved):

PR #6092, opened 2025-08-20
Correct command mode handling for vim vs non-vim modes

manzt:
In vim mode we incorrectly fell back to plain `Escape`, ignoring the configurable command shortcut. Now Escape is only used in non-vim modes, while vim uses its dedicated shortcut.


manzt:
@akshayka I hard coded the linux/mac/windows overrides. If this still doesn't work, i'll make a separate PR changing how we determine platform.

akshayka:
@manzt thanks, still not working unfortunately

manzt:
yeah i'm thinking it's because platform overrides everything (and linux is always windows). i'll make the "right" fix in a separate pr

akshayka (approved):

PR #6096, opened 2025-08-21
improvement: more lenient AI config, and more descriptive error

mscolnick:
More lenient AI config for things that don't require configuration, like `ollama`


akshayka (approved):
If it's easy, it'd be nice if ` Go to Settings > AI to configure.` was a link button that opened the user settings. If it's not easy, then never mind

PR #6099, opened 2025-08-21
Properly detect Linux platform for hotkeys

manzt:
Linux was incorrectly treated as Windows for keyboard shortcuts. Now Linux, Windows, and Mac are properly distinguished, allowing platform-specific keybindings to work correctly.

Should be merged before #6092 and that PR rebased on these changes.

### DIFF
--- a/frontend/src/core/hotkeys/hotkeys.ts
+++ b/frontend/src/core/hotkeys/hotkeys.ts
@@ -1,10 +1,10 @@
 /* Copyright 2024 Marimo. All rights reserved. */
-import { isPlatformMac } from "@/core/hotkeys/shortcuts";
+import { isPlatformMac, isPlatformWindows } from "@/core/hotkeys/shortcuts";
 import { Objects } from "@/utils/objects";
 
 export const NOT_SET: unique symbol = Symbol("NOT_SET");
 
-interface Hotkey {
+export interface Hotkey {
   name: string;
   /**
    * Grouping for the command palette and keyboard shortcuts page.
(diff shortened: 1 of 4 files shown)


frontend/src/core/hotkeys/hotkeys.ts
```diff
@@ -432,21 +433,17 @@ export interface IHotkeyProvider {
...
-  static create(isMac?: boolean): HotkeyProvider {
-    return new HotkeyProvider(DEFAULT_HOT_KEY, isMac);
+  static create(platform?: Platform): HotkeyProvider {
+    return new HotkeyProvider(DEFAULT_HOT_KEY, platform);
   }
 
   constructor(
     private hotkeys: Record<HotkeyAction, Hotkey>,
-    isMac?: boolean,
+    private platform: Platform = resolvePlatform(),
```
Light2Dark: a smal thought is that we can mark/comment this parameter for testing use only. Since we don't & shouldn't specify it elsewhere I think

but it's not a big deal

Light2Dark (approved):

frontend/src/core/hotkeys/hotkeys.ts
manzt: Done, thanks!

frontend/src/core/hotkeys/hotkeys.ts
```diff
@@ -431,22 +432,33 @@ export interface IHotkeyProvider {
...
+
 export class HotkeyProvider implements IHotkeyProvider {
-  private mod: string;
+  private mod: ModKey;
   private platform: Platform;
 
-  static create(isMac?: boolean): HotkeyProvider {
-    return new HotkeyProvider(DEFAULT_HOT_KEY, isMac);
+  /**
+   * @param platform - See {@link HotkeyProviderOptions.platform}.
```
Light2Dark, line 450: wow :o, cool!

PR #6101, opened 2025-08-21
cleanup: xdg config commons and migrate utils to use Path

mscolnick:
Follow-up and consolidation after #6048

### DIFF
--- a/marimo/_utils/xdg.py
+++ b/marimo/_utils/xdg.py
@@ -1,0 +1,93 @@
+# Copyright 2024 Marimo. All rights reserved.
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+
+def home_path() -> Path:
+    """Get home directory or temp directory if home directory is not available.
+
+    Returns:
+        Path: The home directory.
+    """
+    try:
+        return Path.home().resolve()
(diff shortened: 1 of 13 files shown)


marimo/_utils/xdg.py
```diff
@@ -0,0 +1,93 @@
...
+    """Get home directory or temp directory if home directory is not available.
+
+    Returns:
+        Path: The home directory.
+    """
+    try:
+        return Path.home().resolve()
+    except RuntimeError:
+        # Can't get home directory, so use temp directory
+        return Path("/tmp")
```
mscolnick, line 18: this is the only logic change (at least intentionally). instead of returning `None` in some code paths and bubbling up, we instead return a `tmp` dir

dmadisetti: Seems reasonable to me. `/tmp` may be RO in extreme situations, but then they should just set the env variables.

Is there a better place for this function to live though?

dmadisetti (dismissed):
Nice refactor

marimo/_utils/xdg.py
mscolnick: i think this is a good place for this function

PR #6110, opened 2025-08-21
fix: tools being empty

mscolnick:


Light2Dark:
I tested with deepseek 3.1 and it works

akshayka (approved):

PR #6115, opened 2025-08-21
fix: Catch script edgecase for convert

dmadisetti:
## 📝 Summary

Notebooks that imported marimo but were not apps failed. Added catch, and appropriate tests


dmadisetti:
Unsure why we run cli-test on the pypi version? cli tests pass locally

dmadisetti:
Merging. Wheel pulled from testing, but can revert if persistently breaks CI

PR #6121, opened 2025-08-22
Minor fix to docs: show bullet list. 

koaning:
The bullet list does not show unless there is a newline. 

CleanShot 2025-08-22 at 10 33 56


akshayka (approved):
Thanks!

PR #6130, opened 2025-08-23
improvement: check node version

mscolnick:
This checks the node version >=20, which needed for GH Copilot.

### DIFF
--- a/marimo/_server/lsp.py
+++ b/marimo/_server/lsp.py
@@ -168,13 +168,13 @@ class CopilotLspServer(BaseLspServer):
                     )
             else:
                 # Fail open: If the node version check fails we don't want to not start the server.
-                # If it fails again, the user will be alerted again..
+                # If it fails again, the user will be alerted again.
                 LOGGER.info(
                     "Failed to get Node.js version, stderr: %s", result.stderr
                 )
         except Exception as e:
             # Fail open: If the node version check fails we don't want to not start the server.
-            # If it fails again, the user will be alerted again..
+            # If it fails again, the user will be alerted again.
             LOGGER.info("Failed to check Node.js version: %s", e)
 
         return True


marimo/_server/lsp.py
```diff
@@ -142,11 +142,40 @@ class CopilotLspServer(BaseLspServer):
...
+                        "Please upgrade at https://nodejs.org/."
+                    )
+            else:
+                # Fail open, if the node command failed it will alert on start.
+                LOGGER.info(
+                    "Failed to get Node.js version, stderr: %s", result.stderr
+                )
+        except Exception as e:
+            # Fail open, if the node command failed it will alert on start.
+            LOGGER.info("Failed to check Node.js version: %s", e)
```
dmadisetti, line 178: Shouldn't we return here too?

dmadisetti, line 178:
```suggestion
            return f"Failed to check Node.js version: {e}"
```

marimo/_server/lsp.py
```diff
@@ -142,11 +142,40 @@ class CopilotLspServer(BaseLspServer):
...
+                if major_version < 20:
+                    return (
+                        f"Node.js version {version_str} is too old. "
+                        "GitHub Copilot requires Node.js version 20 or higher. "
+                        "Please upgrade at https://nodejs.org/."
+                    )
+            else:
+                # Fail open, if the node command failed it will alert on start.
+                LOGGER.info(
+                    "Failed to get Node.js version, stderr: %s", result.stderr
```
dmadisetti, line 173:
```suggestion
                return (
                    f"Failed to get Node.js version, stderr: {result.stderr}"
```

marimo/_server/lsp.py
mscolnick: i want to fail open. just because we can't get the version. i don't want to not start the node server.

marimo/_server/lsp.py
```diff
@@ -142,11 +142,40 @@ class CopilotLspServer(BaseLspServer):
...
+                # Parse major version
+                major_version = int(version_str.split(".")[0])
+                if major_version < 20:
+                    return (
+                        f"Node.js version {version_str} is too old. "
+                        "GitHub Copilot requires Node.js version 20 or higher. "
+                        "Please upgrade at https://nodejs.org/."
+                    )
+            else:
+                # Fail open, if the node command failed it will alert on start.
```
dmadisetti:
```suggestion
                LOGGER.info(
                    "Failed to get Node.js version, stderr: %s", result.stderr
                )
        except Exception as e:
            # Log, but allow falling through such that an error will display on LSP start.
            LOGGER.info("Failed to check Node.js version: %s", e)
```

marimo/_server/lsp.py
dmadisetti: Got you, didn't really get the comment

PR #6131, opened 2025-08-23
fix: GitHub copilot resiliency with LSP proxy retry

mscolnick:
I can reliably reproduce an error when the server starts with copilot off and then turn it on in the settings. this fixes that code path.

### DIFF
--- a/marimo/_server/api/middleware.py
+++ b/marimo/_server/api/middleware.py
@@ -425,39 +425,41 @@ class ProxyMiddleware:
 
             # Try to connect to the upstream WebSocket with retries
             max_retries = 3
-            retry_delay = 0.5  # seconds
             exponential_backoff = 1.5
 
-            ws_client: Optional[ClientConnection] = None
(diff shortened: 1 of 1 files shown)


marimo/_server/api/middleware.py
```diff
@@ -422,7 +422,46 @@ async def _proxy_websocket(
...
+                        LOGGER.error(
+                            f"Failed to connect to {ws_url} after {max_retries} attempts. Final error: {e}"
+                        )
+                        # Close the client WebSocket with a meaningful error
+                        if (
+                            websocket.client_state
+                            != WebSocketState.DISCONNECTED
+                        ):
+                            await websocket.close(
+                                code=1011,
```
dmadisetti: What's this code?

dmadisetti:
Minus type check

marimo/_server/api/middleware.py
mscolnick: unexpected error. moving to a constant

PR #6136, opened 2025-08-23
fix: use 'uv tree' to list packages if available

### LINKED ISSUE #6122: LSP warnings in settings but (maybe) running anyway
### Describe the bug

In settings I am seeing 

Image

I have run `uv add "marimo[lsp]"`

My project.toml file:

```
[project]
name = "fund3"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "marimo[lsp]>=0.14.17",
    "numpy>=2.3.2",
    "pandas>=2.3.1",
]

# Language server configuration
[tool.marimo.language_servers.pylsp]
enabled = true               # Enable/disable the Python language server
enable_mypy = true           # Type checking with mypy (enabled by default, if installed)
enable_ruff = true           # Linting with ruff (enabled by default, if installed)
enable_flake8 = false        # Linting with flake8
enable_pydocstyle = false    # Check docstring style
enable_pylint = false        # Linting with pylint
enable_pyflakes = false      # Syntax checking with pyflakes

# Diagnostics configuration
[tool.marimo.diagnostics]
enabled = true               # Show diagnostics in the editor
```

`uv add python-lsp-server` doesn't seem to fix it.

FWIW I am getting lint errors in the editor, so maybe it is working.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

_No response_

mscolnick:
Fixes #6122

I _think_ this is correct. `list_packages()` just called `uv pip list` will grab another environment where instead we want to grab all the packages from `pyproject.toml` (using `uv tree`)

### DIFF
--- a/marimo/_runtime/packages/pypi_package_manager.py
+++ b/marimo/_runtime/packages/pypi_package_manager.py
@@ -382,10 +382,21 @@ class UvPackageManager(PypiPackageManager):
         tree = self.dependency_tree()
         if tree is not None:
             LOGGER.info("Listing packages with 'uv tree'")
-            return [
-                PackageDescription(name=pkg.name, version=pkg.version or "")
-                for pkg in tree.dependencies
-            ]
+            seen: set[str] = set()
+            packages: list[PackageDescription] = []
+            stack = list(tree.dependencies)
+            while stack:
+                pkg = stack.pop()
+                if pkg.name not in seen:
+                    packages.append(
+                        PackageDescription(
+                            name=pkg.name, version=pkg.version or ""
+                        )
+                    )
+                    seen.add(pkg.name)
+                    # Add dependencies to stack for recursion
+                    stack.extend(pkg.dependencies)
+            return sorted(packages, key=lambda pkg: pkg.name)
 
         LOGGER.info("Listing packages with 'uv pip list'")
         cmd = [self._uv_bin, "pip", "list", "--format=json", "-p", PY_EXE]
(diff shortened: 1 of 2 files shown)


marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -378,6 +378,26 @@ async def uninstall(self, package: str) -> bool:
         )
 
     def list_packages(self) -> list[PackageDescription]:
+        # First try with `uv tree`
+        tree = self.dependency_tree()
+        if tree is not None:
+            LOGGER.info("Listing packages with 'uv tree'")
+            seen: set[str] = set()
```
dmadisetti, line 385:
```python
seen: dict[Name, set[Version]] = {}
...

if pkg.name in seen:
    seen[pkg.name].add({pkg.version})
else:
    seen[pkg.name] = set({pkg.version})
...

packages = [PackageDescription(name, ",".join(sorted(versions))) for pkg, versions in seen.items()]
```

Are there many dupes? We likely want to report them all rights?

mscolnick: there are dups, but that is expected since we flatten the tree. they will all have the same version though.

manzt: There can only be one version of a Python dependency when uv solves an environment, it will fail otherwise. Thus, a dependency can be found in multiple parts of the tree, but will have the same version, so this code seems correct to me.

manzt (approved):
Looks good to me!

PR #6140, opened 2025-08-23
improvement: add tab to configure "Displayed models" in the AI configuration

mscolnick:
- Add tab to configure "Displayed models" in the AI configuration
- Debounce config to avoid DOS the backend on rapid config changes
- Edge case where displayed models could filter out all models. In this case, we revert to showing all models.

image


Light2Dark:
CleanShot 2025-08-25 at 13 19 15

use react aria button to improve accessibility (will throw warning if not, seems the slot=chevron is unique to aria button).
*i can change the component name to AriaButton

PR #6142, opened 2025-08-24
perf: start LSP and MCP in the background, avoid repetitive io reads

mscolnick:


marimo/_server/lsp.py
```diff
@@ -64,7 +66,7 @@ def start(self) -> Optional[Alert]:
                 return None
 
             LOGGER.debug("... running command: %s", cmd)
-            self.process = subprocess.Popen(
+            self.process = subprocess.Popen(  # noqa: ASYNC220
```
dmadisetti, line 69: Think it's worth still using asyncio here

https://docs.astral.sh/ruff/rules/create-subprocess-in-async-function/

marimo/_utils/timer.py
```diff
@@ -0,0 +1,22 @@
+import functools
+import time
+from typing import Any, Callable
+
+
+def timer(func: Callable[..., Any]) -> Callable[..., Any]:
```
dmadisetti, line 6: dead code? looks like this might have snuck in from some other benchmarking stuff

mscolnick: this is a debug util, but yes it is not used. i want to keep it around for others to use

marimo/_server/lsp.py
mscolnick: i tried, but `create_subprocess_shell` is not supported in windows, for the version range of python we were using

PR #6153, opened 2025-08-25
improvement: handle errors in stream responses

mscolnick:
This handles where the stream response is successful, but the first stream chunk is not.

### DIFF
--- a/marimo/_ai/_convert.py
+++ b/marimo/_ai/_convert.py
@@ -432,7 +432,8 @@ def convert_to_ai_sdk_messages(
 ) -> str:
     """
     Format text events for the AI SDK stream protocol.
-    See: https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol
+    This follows the data-stream v1 protocol
+    See: https://v4.ai-sdk.dev/docs/ai-sdk-ui/stream-protocol
     """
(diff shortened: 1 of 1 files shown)


marimo/_ai/_convert.py
```diff
@@ -441,6 +442,7 @@ def convert_to_ai_sdk_messages(
     TOOL_CALL_PREFIX = "9:"
     TOOL_RESULT_PREFIX = "a:"
     FINISH_REASON_PREFIX = "d:"
+    ERROR_PREFIX = "3:"
```
dmadisetti, line 446: Can you update the comment to reflect v4: https://v4.ai-sdk.dev/docs/ai-sdk-ui/stream-protocol

newest is now v5, which we might consider bumping to: https://v5.ai-sdk.dev/docs/ai-sdk-ui/stream-protocol

dmadisetti: others we might consider handling:

source `h`, filepart `k`, datapart `2`, annotation `8`, step `f`

mscolnick: its actually v1. and these are all valid, but unhandled currently

dmadisetti (approved):

PR #6158, opened 2025-08-26
Add "Notebook Errors" as a context in any AI input

mscolnick:
This adds a new context provider for Notebook Errors. It currently grabs all errors as a single entry. In the future we could group errors by type or single errors, but i don't think we need to explode out the errors into more choices.

Screenshot 2025-08-25 at 11 04 09 PM

Screenshot 2025-08-25 at 11 03 55 PM

### DIFF
--- a/frontend/src/components/editor/ai/__tests__/completion-utils.test.ts
+++ b/frontend/src/components/editor/ai/__tests__/completion-utils.test.ts
@@ -52,41 +52,36 @@ describe("getAICompletionBody", () => {
     const input = "Use @data://dataset1 and @data://dataset2 for analysis";
     const result = getAICompletionBody({ input });
 
-    expect(result).toEqual({
-      includeOtherCode: "// Some other code",
-      context: {
-        schema: [
-          {
-            name: "dataset1",
-            columns: [
-              { name: "col1", type: "number" },
-              { name: "col2", type: "string" },
-            ],
-          },
-          {
-            name: "dataset2",
-            columns: [
-              { name: "col3", type: "boolean" },
-              { name: "col4", type: "date" },
-            ],
-          },
-        ],
-        variables: [],
-      },
-    });
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "context": {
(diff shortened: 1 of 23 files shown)


frontend/src/core/ai/context/__tests__/utils.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      type: "data",
+      data: {
+        name: "dataset<>&\"'",
+        description: "Contains special chars",
+      },
+    };
+
+    const result = contextToXml(context);
+    expect(result).toBe(
+      '<data name="dataset&lt;&gt;&amp;&quot;&#39;" description="Contains special chars"></data>',
```
Light2Dark: cmiiw, but this is going to be passed as strings to the backend. Might seem gibberish for the AI on these data structures. 

We could just use raw values even if invalid xml imo / parse it correctly on the backend. Or use another format?

frontend/src/components/editor/ai/__tests__/completion-utils.test.ts
```diff
@@ -52,41 +52,40 @@ describe("getAICompletionBody", () => {
...
-      context: {
-        schema: [],
-        variables: [],
-      },
-    });
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "context": {
+          "plainText": "",
+          "schema": [],
```
Light2Dark, line 83: cool that we are moving the formatting to the frontend, does this mean we plan to no longer use schema & variables.

frontend/src/core/ai/context/__tests__/utils.test.ts
mscolnick: yea you are probably right

frontend/src/components/editor/ai/__tests__/completion-utils.test.ts
mscolnick: yea, we are not, but i dont want to remove that completely just yet in this PR

PR #6165, opened 2025-08-26
fix: _repr_mimebundle_ for img and tuple outputs

### LINKED ISSUE #6162: marimo does not exhaustively handle the output of obj._repr_mimebundle_
I'm now certain this is a bug.

> While reviewing the plotnine [PR](https://github.com/has2k1/plotnine/pull/969), I could not get it to work for marimo (v0.15). I don't know if there is something wrong with my installation!
> 
> Here is a reproducible example.
> 
> ```python
> from io import BytesIO
> from PIL import Image, ImageDraw
> 
> class BlueOnGray:
>     def __init__(self, text, retina=False):
>         self.text = text
>         self.retina = retina
> 
>     def _repr_mimebundle_(self, include=None, exclude=None):
>         w, h = 200, 80
>         f = 2 if self.retina else 1
>         img = Image.new("RGB", (w, h), color="lightgray")
>         metadata = {"image/png": {"width": w // f, "height": h // f}}
>         draw = ImageDraw.Draw(img)
>         draw.text((10, 30), self.text, fill="blue")
>         buffer = BytesIO()
>         img.save(buffer, format="PNG")
>         return {
>             "image/png": buffer.getvalue(),
>             "text/html": f"<span style='color: blue; background-color: gray'>{self.text}</span>",
>             #"text/plain": self.text
>         }, metadata
> 
> BlueOnGray("Blue text on gray background")
> ```
> Two things for marimo and `_repr_mimebundle_`.
> 
> 1. Cannot deal with images (tested for `png` and `jpeg`).
> 2. Fails for all mimetypes if the return value is a tuple that includes the metadata.
> 
> On the otherhand, using a `_mime_` method and returning `("image/png", buffer.getvalue())` works! 

 _Originally posted by @has2k1 in [#5953](https://github.com/marimo-team/marimo/issues/5953#issuecomment-3215991143)_

mscolnick:
Fixes #6162

## Summary

- Fix `_repr_mimebundle_` handling when it returns tuple format
- Add support for Jupyter spec-compliant `(data, metadata)` tuple returns
- Convert binary image data to data URLs for web display
- Add comprehensive tests for both dict and tuple return formats

## Background

Issue #6162 reported that marimo fails to handle `_repr_mimebundle_` output when libraries (like plotnine) return the Jupyter-compliant tuple format `(data_dict, metadata_dict)` instead of just the data dict. Additionally, binary image data wasn't being properly converted for web display.

## Changes

**Core Fix (repr_formatters.py:55-69):**
- Added handling for tuple return format from `_repr_mimebundle_`
- Extract data portion when method returns `(data, metadata)` tuple
- Convert binary image data to data URLs using existing `io_to_data_url` utility
- Maintain backward compatibility with existing dict-only returns

### DIFF
--- a/docs/guides/integrating_with_marimo/displaying_objects.md
+++ b/docs/guides/integrating_with_marimo/displaying_objects.md
@@ -76,6 +76,8 @@ We support the following methods:
 - `_repr_latex_`
 - `_repr_text_`
 
+**Note:** marimo currently does not handle any optional metadata returned by `_repr_mimebundle_`.
+
 ## Option 3: Implement a `_mime_` method
 
 When displaying an object, marimo's media viewer checks for the presence of a
--- a/marimo/_output/formatters/repr_formatters.py
+++ b/marimo/_output/formatters/repr_formatters.py
@@ -1,12 +1,19 @@
 # Copyright 2024 Marimo. All rights reserved.
 from __future__ import annotations
 
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, cast
 
 from marimo._messaging.mimetypes import KnownMimeType
 from marimo._output.formatters.iframe import maybe_wrap_in_iframe
 from marimo._plugins.core.media import io_to_data_url
 from marimo._utils.methods import is_callable_method
+
+MEDIA_MIME_PREFIXES = (
+    "image/",
+    "audio/",
+    "video/",
+    "application/pdf",
+)
 
 
 def maybe_get_repr_formatter(
@@ -44,6 +51,7 @@ def maybe_get_repr_formatter(
                     continue
 
                 method = getattr(obj, attr)
+                contents: Any
                 # Try to call _repr_mimebundle_ with include/exclude parameters
                 if attr == "_repr_mimebundle_":
                     try:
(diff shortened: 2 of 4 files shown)


docs/guides/integrating_with_marimo/displaying_objects.md
```diff
@@ -76,6 +76,8 @@ We support the following methods:
 - `_repr_latex_`
 - `_repr_text_`
 
+**Note:** marimo currently does not handle any optional metadata returned by `_repr_mimebundle_`.
```
dmadisetti, line 79: What does this mean?

```suggestion
**Note:** marimo currently does not handle optional metadata (such as widget dimension) returned by `_repr_mimebundle_`.
```

dmadisetti: Or maybe give some context to the user. Is there a common case where this happens? (e.g. plotnine metadata sizing)

marimo/_output/formatters/repr_formatters.py
```diff
@@ -44,13 +51,35 @@ def f_repr(obj: Any) -> tuple[KnownMimeType, str]:
                     continue
 
                 method = getattr(obj, attr)
+                contents: Any
```
dmadisetti, line 54: Maybe some stronger typing here?

dmadisetti: Non-blocking but a nice to have

PR #6168, opened 2025-08-27
fix: Enable qualified sql references to shadow regular variables

dmadisetti:
## 📝 Summary

Followup from #5992- currently SQL references and python references share the same "definition" namespace space, which can lead to unexpected variable cycles for the user. This propagates the qualified_name from visitor to graph construction, such that we can distinguish between python and sql refs and prevent false positive cycles


tests/_runtime/test_dataflow_cases.py
```diff
@@ -306,9 +306,6 @@ def __post_init__(self) -> None:
         },
         # TODO: What should the defs be?
```
mscolnick: can we remove this TODO?

mscolnick: there are a few other comments either from @Light2Dark or me `# expected_refs`. do you mind cleaning those up and verifying they look correct to you

mscolnick (dismissed):
great stuff! this is awesome

Light2Dark:
thanks Dylan! I like the approach of having SQLRef in find_sql_defs. The code change isn't too much this way.

Issue #6179, opened 2025-08-27
Websocket noisyness (deployed using fastapi)


oscarrobertson:
### Describe the bug

Hi team

We are using Marimo server deployed following FastAPI guide: https://docs.marimo.io/guides/deploying/programmatically/#fastapi-example
I can also reproduce this issue locally using `marimo run`

I've discovered that _something_ is accumulating over time when I refresh certain notebooks but it's not clear what exactly that is! The end result is that the same notebook when the deployment is in a good state (ie. I just bounced it) is snappy and responsive and the stream of websocket messages looks sensible. When it's in a bad state particular cells with certain behaviour spam 100s of error messages over websockets before eventually succeeding, but all these extra websocket messages slow down the notebook a lot. Example image of the errors below, they seem to be ast parsing failures of some kind. Bigger notebooks end up getting slowed down more so in my minimal repros the time delay isn't as noticeable. 

I've also noticed that each refresh of the notebook adds one more error websocket message for each of the offending cells.

I think the cells that this ends up happening to are those that pop up the warning that they are non reusable that links to here: https://docs.marimo.io/guides/reusing_functions/ - but in my reading this should only be relevant if we're trying to import these things into other files, not importing into other cells in the same notebook.

I have a very minimal repro notebook below, on a fresh deployment this works absolutely fine. If I keep refreshing this notebook I start to get the behaviour described. I cannot reproduce this running in a local marimo edit

<img width="1861" height="467" alt="Image" src="https://github.com/user-attachments/assets/dd5eac21-a122-424b-b52f-0a354b3d40d6" />

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.15",
  "editable": false,
  "OS": "Linux",
  "OS Version": "4.18.0-553.64.1.el8_10.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.12.11",
  "Binaries": {
    "Browser": "138.0.7204.183",
    "Node": "v18.18.2"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.1.2",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
[...]

PR #6180, opened 2025-08-27
fix: turn off toplevel function post hook in run mode

dmadisetti:
## 📝 Summary

Mitigates #6179 by only running top level function analysis in edit mode. A bit worrying they also noticed a performance issue, but this is a reasonable stopgap


dmadisetti:
We don't have any explicit tests for `launch_kernel`, code path is invoked through `session.start_kernel` and the kernel object is not directly accessible to check

We can add tests, but I think it would be a bigger effort than this quick change

PR #6184, opened 2025-08-28
fix: fixes for functions and anywidget virtual file when auto-instantiate=off

mscolnick:


dmadisetti:
Rebuilt and may still be seeing the error. Will dig in a bit deeper

dmadisetti:
Got you. So subsequent runs work, which was the blocker. The error is still rendered on first pass which might be a bit misleading?

image


I think a "Output area requires initial run" error should be possible if it's a first pass render

mscolnick:
Yep exactly

Issue #6163, opened 2025-08-26
Timeout after inactivity


tgbrooks:
### Description

I would like to be able to configure marimo to close automatically after a certain amount of time without any activity. This is useful since I make a job on our cluster to run marimo and want the job to auto exit when I'm no longer using it and forget to manually shut off marimo.

### Suggested solution

For example, to timeout after 60 minutes:

`marimo edit --timeout 60 example.py`

### Are you willing to submit a PR?

- [ ] Yes

mscolnick:
hey @tgbrooks - this is a great suggestion. would you be open to making this contribution? not sure we have the bandwidth currently to take this on.

tgbrooks:
Yes, I can work on it.

mscolnick:
Closed by https://github.com/marimo-team/marimo/pull/6196 (thanks @tgbrooks!)

PR #6196, opened 2025-08-29
Implement timeout option

tgbrooks:
## 📝 Summary

Add `--timeout` CLI flag which will cause the server to shutdown after the specified duration if no connections are made or active. See #6163

## 🔍 Description of Changes

Added a Starlete Middleware, `TimeoutMiddleware`, which increments a timer on the app state to record when the most recent activity was. Every 10 seconds (currently not configurable), it compares to the current time and shuts down the server if necessary.

## 📋 Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable). #6163

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -476,9 +476,9 @@ def edit(
     skew_protection: bool,
     remote_url: Optional[str],
     convert: bool,
+    timeout: Optional[float],
     name: Optional[str],
     args: tuple[str, ...],
-    timeout: Optional[float],
 ) -> None:
     # We support unix-style piping, e.g. cat notebook.py | marimo edit
     if name is None and (stdin_contents := _get_stdin_contents()) is not None:
@@ -722,8 +722,8 @@ def new(
     base_url: str,
     sandbox: Optional[bool],
     skew_protection: bool,
-    prompt: Optional[str],
     timeout: Optional[float],
+    prompt: Optional[str],
 ) -> None:
     if sandbox:
         from marimo._cli.sandbox import run_in_sandbox


tgbrooks:

tgbrooks:
This was my first contribution to this project and my first time working with Starlette. I have a couple points that a reviewer should keep in mind:

1. Is a Starlette middleware the appropriate way to monitor activity and is there a situation where a user could be active without any activity that the middleware detects? I assume activity through websockets is not tracked, but I don't know if there is a situation where users only have activity over websockets.
2. The timeout monitor checks every 10 seconds. Currently this value isn't configurable. Should it be configurable and if so, where are values like this configured? And should the default be changed?

marimo/_server/main.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2024 Marimo. All rights reserved.
+# Middleware(TimeoutMiddleware),
```
dmadisetti: Wayward comment?

marimo/_server/api/middleware.py
```diff
@@ -533,3 +536,53 @@ async def upstream_to_client() -> None:
...
                 await websocket.close(code=WebSocketCodes.UNEXPECTED_ERROR)
             raise
+
+
+class TimeoutMiddleware(BaseHTTPMiddleware):
+    def __init__(
+        self,
+        app: ASGIApp,
+        dispatch: DispatchFunction | None = None,
+        app_state: Any = None,
```
dmadisetti: Can we more strongly type this?

marimo/_server/api/middleware.py
```diff
@@ -533,3 +536,53 @@ async def upstream_to_client() -> None:
...
+
+        request.app.state.timeout_tracker = time.time()
+
+        LOGGER.info(f"Connection detected {request.app.state.timeout_tracker}")
+
+        return await self.app(scope, receive, send)
+
+    async def monitor(self) -> None:
+        while True:
+            await asyncio.sleep(10)
```
dmadisetti: You can sleep based on the timeout itself. e.g

If the timeout is 300s, and the user is already inactive 200 seconds, then try waiting 101 seconds

marimo/_server/api/middleware.py
```diff
@@ -533,3 +536,53 @@ async def upstream_to_client() -> None:
...
+            LOGGER.info("Checking time!")
```
dmadisetti: Log message should be a bit more meaningful and maybe debug

marimo/_server/print.py
```diff
@@ -155,3 +155,7 @@ def print_experimental_features(config: MarimoConfig) -> None:
     print_tabbed(
         f"{_utf8('🧪')} {green('Experimental features (use with caution)')}: {', '.join(keys)}"
     )
+
+
+def print_timeout() -> None:
```
dmadisetti: Hmm, maybe not needed as its own function?

tests/_cli/test_cli.py
```diff
@@ -1215,3 +1215,22 @@ def test_cli_edit_with_convert(
...
+            "--headless",
+            "--timeout",
+            "0.01",  # very short timeout
+        ],
+        stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        text=True,
+    )
+
+    stdout, _ = p.communicate(timeout=60)
```
dmadisetti: I think you can lower this to 5

marimo/_server/main.py
```diff
@@ -116,6 +119,13 @@ def create_starlette_app(
             ModuleNotFoundError: handle_error,
         },
     )
+    if timeout is not None:
```
dmadisetti, line 121: `final_middlewares.append` ? Opposed to post app creation

dmadisetti (changes requested):
Thanks for the PR!

Middleware is the correct mechanism to shutdown the whole system. I left a few highlevel comments

marimo/_server/main.py
tgbrooks: I couldn't do that since I needed to store a reference to `app.state`, which doesn't exist yet at the point that `final_middlewares` is used. I couldn't find a way to extract the app's state otherwise (the `app` parameter to `__init__` is actually another middleware and there doesn't seem to be a way to chain it through to the actual app). Maybe there's an alternative I'm not thinking of

marimo/_cli/cli.py
```diff
@@ -451,6 +451,14 @@ def _get_stdin_contents() -> str | None:
...
 )
 @click.argument("args", nargs=-1, type=click.UNPROCESSED)
+@click.option(
+    "--timeout",
+    required=False,
+    default=None,
+    show_default=False,
+    type=float,
+    help="Enable a global timeout to shut down the server after specified number of minutes of no connection",
+)
```
mscolnick, line 455: Im not sure if its required, but can you move this before the positional `args` and `name`

PR #6202, opened 2025-08-30
If the gpu stats command fails, don't run it again on every usage check

loginoff:
## 📝 Summary

On a machine where the `nvidia-smi` command exists, but it's execution fails for whatever reason (it cannot communicate with the driver, for example), marimo's console logs are spammed with
`[W 250830 16:25:58 health:265] Failed to extract GPU stats: Command '['nvidia-smi', '--query-gpu=index,name,memory.total,memory.used,memory.free', '--format=csv,noheader,nounits']' returned non-zero exit status 9`
This commit add a check further up the line, so that the warning is displayed on the console only once.

## 🔍 Description of Changes
I added a simple invocation of `nvidia-smi` to an already existing function which just used to check for the existence of the binary. If the invocation fails for whatever reason, the warning to the console is only printed once (because the result of the function is cached)

## 📋 Checklist

### DIFF
--- a/marimo/_server/api/endpoints/health.py
+++ b/marimo/_server/api/endpoints/health.py
@@ -232,11 +232,7 @@ async def usage(request: Request) -> JSONResponse:
     if _is_gpu_available():
         try:
             result = subprocess.run(  # noqa: ASYNC221
-                [
-                    "nvidia-smi",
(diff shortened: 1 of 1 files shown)


loginoff:
recheck

loginoff:

mscolnick (approved):
thank you!

Issue #3860, opened 2025-02-20
Add default_sort parameter to `mo.ui.table`


sarah-utilityapi:
### Description

I am hoping to be able to write the following code that would mean that a ui.table gets initially rendered with a sorted column.

```
mo.ui.table(data, default_sort='Title')
```

where 'Title' is the column name name of the.

By default it would be ascending.

We would need a mechanism to specify descending. Possible options

```
ascending=False  # like pandas
default_sort='-Title'  # like django
default_sort='ASC:Title' or default_sort='DESC:Title' 
```

If a user then subsequently sorts on a different column the default sort would be lost.

### Suggested solution

I've sketched out an API above. I don't know about the under-the-hood implementation.

mscolnick:
Why not sort the dataframe before passing it into the table? This way you'd have more flexibility with the sorting too

sarah-utilityapi:
When I was working with dataframes I was doing that. In this case I'm passing in a list of dictionaries that I've constructed with plain data and mo.ui components (such as making html link outputs.) In the light of morning, I'm realizing that I can sort this list, and will.

This is slightly lacking as a complete UX solution because it won't show to the user the sort that's been pre-applied. But it gets me out of my jam. Thanks!

mscolnick:
Got it, just wanted to understand the use-case (besides some logic), which is: **show the user how the data is currently sorted**. Thank you, this makes sense.

xeniumcode:
We have to add  functionality like ```mo.ui.table(data,descending=True) ```, if i'm understanding it right ?

hfactor13:
Is anyone still working on this issue? I was going to make a pull request, but wanted to double-check.

mscolnick:
@hfactor13 it is up for grabs! feel free to make the PR. let us know how we can help

hfactor13:
@sarah-utilityapi Hi Sarah, I am currently working on a PR for this issue and wanted some clarification on what kind of functionality you want to see. I managed to incorporate a `sort` keyword and added the logic to make it work, but I also noticed during my tests in Marimo that in the table itself, you can click on whether you want the column `ascending` or `descending`. If I am understanding correctly, you want to have a way to highlight the column that is being sorted. For example, if the column is ascending by default and you click on the column to make it descending, then highlight it; otherwise, remove the highlight when you clear the sort. Does that sound accurate? Also, do you still want the `sort` keyword functionality?

hfactor13:
[...]

PR #6205, opened 2025-08-30
Table sort enhancement

hfactor13:
## 📝 Summary

This fixes issue number #3860 by adding a default sort parameter and being able to pick whether it's ascending or descending.

## 🔍 Description of Changes

I added two keyword arguments called `default_sort` and `ascending` to the table constructor.

## 📋 Checklist

- [] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).

Comments:
Currently, I am troubleshooting why the `test_altair_chart.py` failed.
Screenshot 2025-08-30 at 3 31 14 PM

Screenshot 2025-08-30 at 2 18 54 PM
Screenshot 2025-08-30 at 2 19 10 PM

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -335,6 +335,8 @@ class table(
         max_columns (int, optional): Maximum number of columns to display. Defaults to the
             configured default_table_max_columns (50 by default). Set to None to show all columns.
         label (str, optional): A descriptive name for the table. Defaults to "".
+        default_sort (str, optional): Column name to sort by initially.
+        ascending (bool, optional): Sort ascending (default True). Set to False for descending.
     """
 
     _name: Final[str] = "marimo-table"
@@ -345,6 +347,8 @@ class table(
         *,
         page_size: Optional[int] = None,
         preload: bool = False,
+        default_sort: Optional[str] = None,
+        ascending: bool = True,
     ) -> table:
         """
         Create a table from a Polars LazyFrame.
@@ -392,6 +396,8 @@ class table(
             _internal_total_rows="too_many",
             _internal_lazy=True,
             _internal_preload=preload,
+            default_sort=default_sort,
+            ascending=ascending,
         )
 
     def __init__(
@@ -446,6 +452,8 @@ class table(
         _internal_total_rows: Optional[Union[int, Literal["too_many"]]] = None,
         _internal_lazy: bool = False,
         _internal_preload: bool = False,
+        default_sort: Optional[str] = None,
+        ascending: bool = True,
     ) -> None:
         if page_size is None:
             page_size = self.default_page_size
@@ -609,14 +617,23 @@ class table(
         field_types: Optional[FieldTypes] = None
         num_columns = 0
 
+        # Default sort support
+        initial_sort: Optional[SortArgs] = None
+        if default_sort is not None:
+            # Only sort if column exists
+            colnames = set(self._manager.get_column_names())
+            if default_sort in colnames:
+                initial_sort = SortArgs(by=default_sort, descending=not ascending)
+        
+
         if not _internal_lazy:
-            # Search first page
+            # Search first page, with initial sort if specified
             search_result = self._search(
                 SearchTableArgs(
                     page_size=page_size,
                     page_number=0,
                     query=None,
-                    sort=None,
+                    sort=initial_sort,
                     filters=None,
                 )
             )
--- a/tests/_plugins/ui/_impl/test_table.py
+++ b/tests/_plugins/ui/_impl/test_table.py
@@ -1319,7 +1319,6 @@ def test_download_as_for_unsupported_cell_selection() -> None:
         with pytest.raises(NotImplementedError):
             table._download_as(DownloadAsArgs(format="csv"))
 
-
 @pytest.mark.skipif(
     not DependencyManager.pandas.has() or not DependencyManager.polars.has(),
     reason="Pandas or Polars not installed",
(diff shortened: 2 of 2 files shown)


hfactor13:
I have read the CLA Document, and I hereby sign the CLA.

mscolnick:
Awesome @hfactor13! 

We don't need to support multi-sort yet, but would be nice to make it so we can later. Thoughts on maybe: 

`default_sort="ASC:column"`

Then we can make it a list of strings in future for multi-sort. 

Downside is maybe not obvious how to do ascending/descending without reading the docs.  

Thoughts?

hfactor13:
I think doing it that way might be less intuitive. While it is convenient to have one less keyword argument, my concern is that it could be a source of bugs in the future and like you said you would most likely have to review the documentation if you are new to using marimo. Another alternative could be to use a dictionary where the keys would be `column` and `ascending` followed by values for the column and True or False respectively. It's not the best alternative, but I think it's an option, but then again it runs into the issue of reviewing the docs. 🤔

Light2Dark:
hi @hfactor13 , I agree with Myles on using default_sort='ASC:column'. `ascending` is not explicit that it's for the default sort. You may need to rename it to `default_ascending` and `default_sort_column` which seems more verbose to me. And is more complicated when we add multiple columns sort (in the future).

default_sort
```python
mo.ui.table(data, default_sort='Title')        # ascending (default)
mo.ui.table(data, default_sort='ASC:Title')    # ascending (explicit)
mo.ui.table(data, default_sort='DESC:Title')   # descending (explicit)
```

Django also uses `-` sign, which I am not opposed to, curious to know your thoughts on this.
https://docs.djangoproject.com/en/5.2/ref/models/options/#ordering

The tests are unrelated (they are flakey to location & theme). Thanks!

marimo/_plugins/ui/_impl/table.py
```diff
@@ -609,14 +617,24 @@ def __init__(
         field_types: Optional[FieldTypes] = None
         num_columns = 0
 
+        # Default sort support
+        initial_sort: Optional[SortArgs] = None
+        if default_sort is not None:
```
Light2Dark: Maybe a LOGGER.warning when the default_sort column does not match any column in the data.

marimo/_plugins/ui/_impl/table.py
```diff
@@ -609,14 +617,24 @@ def __init__(
...
+            # Only sort if column exists
+            colnames = set(self._manager.get_column_names())
```
Light2Dark: I think we want to do this only in non-lazy mode (even getting column names). No worries if you don't want to make the change here, but we can follow up.

akshayka (commented):
Instead of "default_sort", can we call the argument "sort"? The "default" is redundant. For example, all of our UI elements take an initial value (which the user can change later), and we call this "value", not "default_value".

I am also in favor of collapsing to just one argument rather than two. It is confusing to have them decoupled.

Another option to consider: instead of adding this argument to the constructor, one could add a new method to table -- sort -- which could then take multiple arguments.

ui.table(df).sort(by="my_column", ascending=False)

which is similar to the Polars API. sort() would return a new table. I haven't thought this through fully, but API wise I prefer chaining over adding even more arguments to the constructor.

hfactor13:
@akshayka Based on your recommendations, this is what I was able to come up with. Let me know what you think.

Screenshot 2025-08-31 at 6 23 18 PM
Screenshot 2025-08-31 at 6 23 55 PM

marimo/_plugins/ui/_impl/table.py
```diff
@@ -719,6 +723,50 @@ def data(self) -> TableData:
         """
         return self._data
 
+    def sort(self, by: str, ascending: bool = True) -> table:
+        """Return a new table instance sorted by a column."""
+        colnames = set(self._manager.get_column_names())
+        if by not in colnames:
+            raise ValueError(f"Column '{by}' not found in table.")
+        sorted_manager = self._manager.sort_values(by, not ascending)
+        sorted_data = sorted_manager.data
```
mscolnick: i don't know if this actually achieves the issue linked. This does not provide much benefit over the user doing:

```
# before
mo.ui.table(data.sort(by="col"))
# after
mo.ui.table(data).sort(by="col")
```

i think what we want instead is to pass down `sort=` or `initial-sort=` to the frontend. similar to `freeze_columns_left`. this will then show in the UI the current default sort

hfactor13:
Here's what the sort functionality looks like now. I am not sure if I am being too restrictive by forcing the use of the `ASC` and `DESC` keywords, but I can change it so that it just does ascending by default.

Screenshot 2025-09-03 at 4 06 10 PM
Screenshot 2025-09-03 at 4 06 16 PM

mscolnick:
my assumption is that this should be shown in the UI (with the initial sort selected). i still don't know what benefit the default sort has over sorting the dataframe otherwise.

hfactor13:
I am not sure I follow. I did notice that in the marimo test notebook that I have, the tables themselves for each column already have ascending and descending functionality when you click on them. Maybe I can ask the original poster of the issue for more clarification, but I can see your stance as to why it could be redundant. How should we proceed?

mscolnick:
@hfactor13 i think asking in the original issue makes the most sense. we should clarify what their goal is with this feature

hfactor13:
I am thinking about closing out this pull request and possibly opening up a separate one (tentative). I haven't gotten much feedback on issue number #3860. To my understanding, and based on my own tinkering around with the ui table, the main thing that needs to be implemented for this issue is a highlighting mechanism whenever a sort is applied, and clearing the highlighting once the sort is cleared. I am not entirely sure what files would affect that. I am not too familiar with TypeScript, but I know more or less how it should behave. Any thoughts?

mscolnick:
@hfactor13 - i agree, we can close and maybe revisit. it does seem like the request requires changes to the frontend.

PR #6215, opened 2025-09-01
Update Gemini2.5 models 

koaning:
image

I noticed that the flash-2.5-preview model was down today while the other models from gemini worked just fine.

I also noticed that the model was removed from [this list](https://ai.google.dev/gemini-api/docs/models) and has seemingly been replaced by the non-preview flash model and a lite variant. So I figured I'd add those models to replace the one that does not work right now.

### DIFF
--- a/packages/llm-info/data/models.yml
+++ b/packages/llm-info/data/models.yml
@@ -123,7 +123,6 @@
   roles: [chat, edit]
   thinking: false
 
-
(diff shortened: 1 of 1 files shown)


koaning:
Reading [the docs here](https://ai.google.dev/gemini-api/docs/models#preview) it also seems like we should refrain from using preview models in general.

PR #6216, opened 2025-09-01
sql cells can't both define and refer to itself

### LINKED ISSUE #6209: mo.sql() hangs on multiple statements with table creation unless output variable matches table name
When a mo.sql() call contains multiple SQL statements that includes a `CREATE TABLE` statement, the cell only runs successfully if the output variable name matches the created table name. If the output variable name is different, the cell remains stuck in the “queued” state.

For example, this is broken:
```python
df = mo.sql(
    f"""
    CREATE OR REPLACE TABLE x AS (
        SELECT 1 AS value
    );

    SELECT * FROM x
    """
)
```

while this works:

```python
x = mo.sql(
    f"""
    CREATE OR REPLACE TABLE x AS (
        SELECT 1 AS value
    );

    SELECT * FROM x
    """
)
```

https://molab.marimo.io/notebooks/nb_y3Pk8wPsHDyQXLER5T8dpP

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.15.2",
  "editable": false,
  "OS": "Darwin",
  "OS Version": "24.6.0",
  "Processor": "arm",
  "Python Version": "3.13.3",
  "Binaries": {
    "Browser": "139.0.7258.155",
    "Node": "v22.12.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.2.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.2",
    "starlette": "0.47.3",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "anywidget": "0.9.18",
    "duckdb": "1.3.1",
    "ibis-framework": "10.8.0",
    "nbformat": "5.10.4",
    "openai": "1.102.0",
    "pandas": "2.3.2",
    "polars": "1.31.0",
    "pyarrow": "21.0.0",
    "loro": "1.6.0",
    "pytest": "8.4.1",
    "python-lsp-ruff": "2.2.2",
    "ruff": "0.12.11",
    "sqlglot": "27.10.0"
  },
  "Experimental Flags": {
    "inline_ai_tooltip": false,
    "chat_sidebar": true,
    "mcp_docs": false,
    "performant_table_charts": false,
    "multi_column": false,
    "rtc": false,
    "rtc_v2": false
  }
}
```

</details>

### Code to reproduce

https://molab.marimo.io/notebooks/nb_y3Pk8wPsHDyQXLER5T8dpP

Light2Dark:
## 📝 Summary

Fixes #6209 

## 📋 Checklist

### DIFF
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -287,6 +287,10 @@ class DirectedGraph:
                     _other_ids_defining_name, variable_name = (
                         self._find_sql_hierarchical_matches(sql_ref)
                     )
+                    # Ignore definitions from the current cell,
+                    # cells can't both define and reference the same name
+                    if cell_id in _other_ids_defining_name:
+                        _other_ids_defining_name.remove(cell_id)
                     other_ids_defining_name.update(_other_ids_defining_name)
 
                 # If other_ids_defining_name is empty, the user will get a
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -287,8 +287,8 @@ class DirectedGraph:
                     _other_ids_defining_name, variable_name = (
                         self._find_sql_hierarchical_matches(sql_ref)
                     )
-                    # Ignore definitions from the current cell,
-                    # cells can't both define and reference the same name
+                    # Ignore definitions from the current cell
+                    # as a reference in the defining cell isn't a cycle.
                     if cell_id in _other_ids_defining_name:
                         _other_ids_defining_name.remove(cell_id)
                     other_ids_defining_name.update(_other_ids_defining_name)
--- a/marimo/_runtime/dataflow.py
+++ b/marimo/_runtime/dataflow.py
@@ -287,10 +287,6 @@ class DirectedGraph:
                     _other_ids_defining_name, variable_name = (
                         self._find_sql_hierarchical_matches(sql_ref)
                     )
-                    # Ignore definitions from the current cell
-                    # as a reference in the defining cell isn't a cycle.
-                    if cell_id in _other_ids_defining_name:
-                        _other_ids_defining_name.remove(cell_id)
                     other_ids_defining_name.update(_other_ids_defining_name)
 
                 # If other_ids_defining_name is empty, the user will get a
--- a/tests/_runtime/test_dataflow.py
+++ b/tests/_runtime/test_dataflow.py
@@ -671,6 +671,25 @@ class TestSQL:
 
         assert graph.get_referring_cells("df", language="python") == set(["1"])
 
+    def test_sql_creation_and_select_from_same_cell(self):
+        graph = dataflow.DirectedGraph()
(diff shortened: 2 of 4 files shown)


mscolnick (approved):

marimo/_runtime/dataflow.py
```diff
@@ -287,6 +287,10 @@ def register_cell(self, cell_id: CellId_t, cell: CellImpl) -> None:
                     _other_ids_defining_name, variable_name = (
                         self._find_sql_hierarchical_matches(sql_ref)
                     )
+                    # Ignore definitions from the current cell,
+                    # cells can't both define and reference the same name
```
dmadisetti:
```suggestion
                    # as a reference in the defining cell isn't a cycle.
```

tests/_runtime/test_dataflow.py
```diff
@@ -671,6 +671,25 @@ def test_python_to_sql_ref(self):
 
         assert graph.get_referring_cells("df", language="python") == set(["1"])
 
+    def test_sql_creation_and_select_from_same_cell(self):
```
dmadisetti, line 674: Would keep this test, but also add a visitor test with the mo.ref and def example I showed. Should just be checking to see if we already defined the var. Would be careful on ordering too since:

```sql
drop table if not exists ... -- "references" table first
```

is a common pattern.

Light2Dark: you are right, thanks! I was working on a quick fix with limited time. I think I have a better change in.
can definitely add more tests and maybe optimize..although I don't see how off the bat.

marimo/_runtime/dataflow.py
```diff
@@ -287,6 +287,10 @@ def register_cell(self, cell_id: CellId_t, cell: CellImpl) -> None:
                     _other_ids_defining_name, variable_name = (
                         self._find_sql_hierarchical_matches(sql_ref)
                     )
+                    # Ignore definitions from the current cell
+                    # as a reference in the defining cell isn't a cycle.
+                    if cell_id in _other_ids_defining_name:
```
dmadisetti: I think you can actually remove this now?

dmadisetti (commented):
Nice! Seems optimal enough to me

marimo/_runtime/dataflow.py
```diff
@@ -287,6 +287,10 @@ def register_cell(self, cell_id: CellId_t, cell: CellImpl) -> None:
...
+                        _other_ids_defining_name.remove(cell_id)
```
dmadisetti:
```suggestion
```

dmadisetti (approved):

PR #6234, opened 2025-09-03
Migration custom serde to `msgspec`

manzt:
This PR migrates marimo's serialization infrastructure from Python dataclasses to [`msgspec`](https://github.com/jcrist/msgspec). The migration transforms all dataclass definitions to `msgspec.Struct` and replaces our custom `WebComponentEncoder` with msgspec's built-in encoder.

## Key Changes

The migration converts all message types that flow between frontend and backend from dataclasses to `msgspec.Struct`, including cell outputs, error types, operation messages, and request types. For discriminated unions like error types and operations, we use msgspec's tagged union feature with `tag_field` and `tag` parameters.

To handle marimo-specific types like `MIME` objects and numpy arrays, we implemented a single custom encoder hook that centralizes all special serialization logic previously scattered across `WebComponentEncoder` and various `to_dict` methods. For complex union types that `msgspec` cannot decode directly, we use Python's `TYPE_CHECKING` block as described in https://github.com/jcrist/msgspec/issues/722.

Existing tests pass with only minor updates for cosmetic serialization differences like compact JSON output.


marimo/_plugins/core/web_component.py
```diff
@@ -22,27 +27,26 @@
...
+
+else:
+    JSONType = Any
 
 S = TypeVar("S", bound=JSONType)
 
 
 def _build_attr(name: str, value: JSONType) -> str:
-    processed = escape(json.dumps(value, cls=WebComponentEncoder))
+    processed = escape(encoder.encode(value).decode("utf-8"))
```
mscolnick: could we expose some utils from this module:

`def to_json_str()`
`def to_dict()`

manzt: good idea. I added `encode_json_str`, `encode_json_bytes`, and `asdict`.

marimo/_runtime/requests.py
```diff
@@ -35,18 +35,42 @@
 SerializedCLIArgs = dict[str, ListOrValue[Primitive]]
 
 
+class UrlParts(msgspec.Struct):
+    """Mapping of parsed URL components."""
+
+    path: str
+    port: int | None
```
Light2Dark: Small thing, I've faced errors where the | operator did not work for Python 3.9. Is it okay here.

manzt: Should be ok with `from __future__ import annotations` at the top of the file I believe. Can switch to `Union` for consistency though.

PR #6237, opened 2025-09-03
openapi spec for msgspec

mscolnick:


frontend/src/utils/numbers.ts
```diff
@@ -1,7 +1,7 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
 export function prettyNumber(
-  value: number | string | undefined | null | boolean | unknown[],
+  value: number | string | undefined | null | boolean | unknown | unknown[],
```
manzt, line 4: same comment as the linting warning here, this could just be: `value: unknown`

marimo/_config/config.py
```diff
@@ -513,12 +513,9 @@ class ExperimentalConfig(TypedDict, total=False):
...
 
-if TYPE_CHECKING:
-    # Prefer to accept any dict since feature flags can change frequently
-    # But maintain type safety for known flags
-    ExperimentalConfigType = Union[dict[str, Any], ExperimentalConfig]
-else:
-    ExperimentalConfigType = dict
+# Prefer to accept any dict since feature flags can change frequently
+# But maintain type safety for known flags
+ExperimentalConfigType = dict[str, Any]
```
manzt, line 518: Comment is outdated if we want to get rid of the static checking for the typed dict

packages/openapi/convert_types.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+# Usage: ./convert_types.sh <path>
+# Replaces all occurrences of 'Record<string, never>' with 'Record<string, any>' in the given file or recursively in a directory.
```
manzt, line 4: Would replacing with `Record<string, unknown>` work? That would be my preference to capture the unknown nature of the type rather than copy-left effects of any.

manzt (approved):
Looks great, just some minor comments.

packages/openapi/convert_types.sh
mscolnick: we can keep track of followup improvements we want to make and either include them in this PR or followups. we can probably do `unknown`

PR #6238, opened 2025-09-04
preview: add `--server-startup-command`

mscolnick:
`--server-startup-command` flag to run a command on startup (non-blocking) and stream to any open sessions. if the command is finished, it will never be shown.

### DIFF
--- a/marimo/_server/session/session_view.py
+++ b/marimo/_server/session/session_view.py
@@ -14,6 +14,7 @@ from marimo._messaging.ops import (
     Interrupted,
     MessageOperation,
     SendUIElementMessage,
+    StartupLogs,
     UpdateCellCodes,
     UpdateCellIdsRequest,
     Variables,
@@ -95,6 +96,9 @@ class SessionView:
             WidgetModelId, list[SendUIElementMessage]
         ] = {}
 
+        # Startup logs for startup command - only one at a time
+        self.startup_logs: Optional[StartupLogs] = None
+
         # Auto-saving
         self.auto_export_state = AutoExportState()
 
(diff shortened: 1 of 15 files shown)


packages/openapi/src/api.ts
```diff
@@ -3371,6 +3371,13 @@ export interface components {
       | components["schemas"]["Banner"]
       | components["schemas"]["MissingPackageAlert"]
       | components["schemas"]["InstallingPackageAlert"]
+      | {
+          content: string;
+          /** @enum {string} */
+          name: "startup-logs";
+          /** @enum {string} */
+          status: "append" | "start" | "done";
+        }
```
mscolnick: planning to wait for @manzt's `msgspec` to merge to avoid a merge conflict, then will update this

marimo/_server/session/session_view.py
```diff
@@ -253,6 +256,13 @@ def add_operation(self, operation: MessageOperation) -> None:
             # TODO: cleanup/merge previous 'update' messages
             self.model_messages[operation.model_id] = messages
 
+        elif isinstance(operation, StartupLogs):
+            prev = self.startup_logs.content if self.startup_logs else ""
+            self.startup_logs = StartupLogs(
+                content=prev + operation.content,
```
manzt, line 262: any need to worry about this getting too long?

manzt (approved):
This looks great. Nice work

marimo/_server/session/session_view.py
mscolnick: i think its ok. we still probably want to ship the whole content. we do have limits on messages sent

PR #6240, opened 2025-09-04
feat: cell output for `@` context w/ attachments for media output

mscolnick:
This PR adds a new `@` context type for cell output. For text or json output, we just include that in the context body. For media types (image, video, charts, tables) we convert from DOM →  PNG and include the attachment.

Screenshot 2025-09-04 at 4 27 12 PM
Screenshot 2025-09-04 at 4 16 30 PM
Screenshot 2025-09-04 at 4 16 22 PM

### DIFF
--- a/frontend/src/components/chat/chat-panel.tsx
+++ b/frontend/src/components/chat/chat-panel.tsx
@@ -6,6 +6,7 @@ import { storePrompt } from "@marimo-team/codemirror-ai";
 import type { ReactCodeMirrorRef } from "@uiw/react-codemirror";
 import {
   DefaultChatTransport,
+  type FileUIPart,
   lastAssistantMessageIsCompleteWithToolCalls,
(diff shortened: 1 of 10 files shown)


frontend/src/components/chat/chat-panel.tsx
```diff
@@ -497,12 +497,13 @@ const ChatPanelBody = () => {
     // experimental_throttle: 100,
     api: runtimeManager.getAiURL("chat").toString(),
     headers: runtimeManager.headers(),
-    experimental_prepareRequestBody: (options) => {
-      const completionBody = getAICompletionBody({
+    experimental_prepareRequestBody: async (options) => {
```
mscolnick: this function is not allowed to be `async`. need to upgrade to v5.0 for this

Light2Dark (approved):

PR #6241, opened 2025-09-04
feat: Reference files as context with #

mscolnick:
This adds file search with `#`. Instead of adding it to the context as a resource, it attaches it as an attachment. 
I could enable this for `@` as well, but currently the resources expect static completions and not dynamic (like file search). This also does not handle directories.

Screenshot 2025-09-08 at 4 52 42 PM

This also:
* fixes reasoning parts that broke in the 5.0 ai sdk migration
* adds an `@` button to trigger context completion

### DIFF
--- a/tests/_server/files/test_os_file_system.py
+++ b/tests/_server/files/test_os_file_system.py
@@ -489,16 +489,15 @@ def test_search_directory_and_file_filters(
     assert len(file_results) > 0, "Should find files with no filter"
     assert len(dir_results) > 0, "Should find directories with no filter"
 
-    # Test with both file=False and directory=False should work like no filter
     results_no_filter = fs.search(query="test", path=str(test_dir))
-    results_both_false = fs.search(
+    results_both_true = fs.search(
         query="test",
         path=str(test_dir),
-        include_files=False,
-        include_directories=False,
+        include_files=True,
+        include_directories=True,
     )
-    assert len(results_no_filter) == len(results_both_false), (
-        "Both false should be same as no filter"
+    assert len(results_no_filter) == len(results_both_true), (
+        "Both true should be same as no filter"
     )
 
 


frontend/src/components/editor/alerts/startup-logs-alert.tsx
```diff
@@ -24,7 +25,7 @@ export const StartupLogsAlert: React.FC = () => {
 
     if (isDone) {
       // Dismiss after 5 seconds
-      timeout = window.setTimeout(() => handleClear(), 5_000);
+      timeout = window.setTimeout(() => handleClear(), 5000);
```
Light2Dark, line 28: add handleClear to useEffect dependency

marimo/_server/files/os_file_system.py
```diff
@@ -184,6 +189,140 @@ def update_file(self, path: str, contents: str) -> FileInfo:
...
+                        try:
+                            is_directory = entry.is_dir()
+                            entry_stat = entry.stat()
+
+                            # Apply directory/file filtering
+                            if (
+                                directory is not None
+                                and directory
+                                and not is_directory
+                            ):
```
Light2Dark:
```suggestion
                            if directory and not is_directory:
```

marimo/_server/files/os_file_system.py
```diff
@@ -184,6 +189,140 @@ def update_file(self, path: str, contents: str) -> FileInfo:
...
+                                # Skip non-directories if directory=True
+                                continue
+                            if file is not None and file and is_directory:
```
Light2Dark:
```suggestion
                            if file and is_directory:
```

marimo/_server/api/endpoints/file_explorer.py
```diff
@@ -249,3 +251,37 @@ async def open_file(
...
+    files = file_system.search(
+        query=body.query,
+        path=body.path,
+        directory=body.directory,
+        file=body.file,
+        depth=body.depth,
+        limit=body.limit,
+    )
+    return FileSearchResponse(
+        files=files, query=body.query, total_found=len(files)
```
Light2Dark, line 286: non-blocker, why do we return the query?

mscolnick: slightly easier for debugging and to invalidate previous requests, but no real reason

Light2Dark (approved):

PR #6245, opened 2025-09-04
fix 6020: output cell fails to wrap text in HTML export

### LINKED ISSUE #6020: Output cell fails to wrap text in HTML export
### Describe the bug

After running `marimo export html notebook.py -o notebook.html`, the HTML file fails to wrap long strings in output cells. In my case, it happened to all cells that output the object's  `__repr__()`. I don't know if it generalizes to any long string.

Image

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.17",
  "editable": false,
  "OS": "Linux",
  "OS Version": "6.14.0-27-generic",
  "Processor": "x86_64",
  "Python Version": "3.11.11",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.12.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.1.1",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.3.2",
    "ibis-framework": "10.5.0",
    "nbformat": "5.10.4",
    "openai": "1.99.5",
    "pandas": "2.2.3",
    "pyarrow": "18.1.0",
    "loro": "1.5.3",
    "pytest": "8.3.3",
    "python-lsp-ruff": "2.2.2",
    "python-lsp-server": "1.13.0",
    "ruff": "0.12.8",
    "sqlglot": "26.24.0"
  },
  "Experimental Flags": {
    "inline_ai_tooltip": true,
    "performant_table_charts": true,
    "chat_sidebar": true,
    "multi_column": true
  }
}
```

</details>

### Code to reproduce

_No response_

maxkoretskyi:
## 📝 Summary

Fixes #6020

## 🔍 Description of Changes

I've added `overflow:auto` to the `output-area` element. This fixes the setup shown in `Read Mode - Code Shown` story in the `one-column.stories.tsx` file:

image

and in the published mode, I reset overflow to visible again:

image

so that the scroll-bar is applicable to the entire notebook, and not the text inside the container.

Also, there was [a comment](https://github.com/marimo-team/marimo/blob/310e3a3c841604e896bdc5db6024657ef738512d/frontend/src/css/app/Cell.css#L408) previously:

>   TODO: Find a way to accommodate large elements, ideally in a way that
>     - doesn't break margin collapse. Setting overflow (eg, overflow: auto)
>     - breaks margin collapse. */

My understand is that it was referring to this case of a child block inside the cell container with a margin:

image

But it seems no longer applicable since output area already uses `display: flow-root;`, which creates block formatting context (BFC), so the margins won't collapse anyway. And secondly, it seems that now cells have borders in all modes except for the published mode, and borders as well prevent margin-collapsing.

### DIFF
--- a/frontend/src/css/app/Cell.css
+++ b/frontend/src/css/app/Cell.css
@@ -8,426 +8,447 @@
  * - disabled
  * */
 .marimo-cell {
-  position: relative;
-  border-radius: 10px;
-  max-width: inherit;
-  width: 100%;
-  border: 1px solid var(--gray-4);
+    position: relative;
+    border-radius: 10px;
+    max-width: inherit;
+    width: 100%;
+    border: 1px solid var(--gray-4);
 
-  @apply divide-y divide-[var(--gray-5)] shadow-sm-solid shadow-shade;
+    @apply divide-y divide-[var(--gray-5)] shadow-sm-solid shadow-shade;
 
-  &:focus-visible {
-    /* focus-visible outlines the entire cell body in black, but the cell's
-     * body is an irregular shape because of pseudo-elements that extend
-     * its hit-box / hover area. */
-    outline: none;
-  }
-
-  &:focus-within {
-    z-index: 20;
-  }
-
-  /* Hover z-index is higher than focus-within z-index
-   * This is because you may hover for tooltip docs while
-   * not focused in another cell.
-   */
-  &:hover {
-    z-index: 30;
-  }
-
-  /* Interactive */
-  &.interactive {
-    /* Only restrain output length in edit mode. */
-    .output-area,
-    .console-output-area {
-      max-height: 610px;
-      overflow: auto;
-    }
-
-    /* Special case for particular components */
-    .output-area:has(> .output > marimo-ui-element > marimo-table) {
-      max-height: none;
-      overflow: hidden;
-    }
-
-    & > :first-child {
-      border-top-left-radius: 9px;
-      border-top-right-radius: 9px;
-    }
-
-    & > :last-child {
-      border-bottom-left-radius: 9px;
-      border-bottom-right-radius: 9px;
-    }
-
-    &:hover {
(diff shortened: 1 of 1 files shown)


mscolnick (approved):

mscolnick:
amazing @maxkoretskyi! this looks right to me, and i think your understanding of the comment is correct and no longer needed

maxkoretskyi:
@mscolnick, great we will adjust if something pops up :).

I want to create a PR to remove this [confusing property](https://github.com/marimo-team/marimo/blob/5449b1c3b5ab2c20f3946be1d7cbbe87fd0c99b0/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx#L330), I've studied the code, and don't understand how it ever could get there, do you know of any setup?

In my findings, this is how rendering flow looks like

```
notebook run modes
	static
		- mode=read
		- runner=run_page
	nonstatic
		- mode:
			- edit
				runner: edit_page
					cells_renderer=CellArray (cell: EditableCellComponent | ReadonlyCellComponent | SetupCellComponent)
			- read
				runner: run_page
					cells_renderer=PluginCellRenderer (vertical-layout)
```

so, `vertical-layout.tsx` where the code is, will always only be rendered in the read mode, right?

maxkoretskyi:
I might be wrong in my assumption, this PR is mostly to confirm my understanding

PR #6249, opened 2025-09-04
Update pixi lock file

manzt:
Pixi was failing in CI due to an outdated lockfile. This commit refreshes the `pixi.lock` to resolve the issue.

We may want to consider pinning a specific version of Pixi in the future to avoid similar failures.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -111,6 +111,7 @@ environments:
       - pypi: https://files.pythonhosted.org/packages/c0/5a/9cac0c82afec3d09ccd97c8b6502d48f165f9124db81b4bcb90b4af974ee/jedi-0.19.2-py2.py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/a3/48/f97f0920c20bc522c34f1eaec4c62d7a03c95c37da2dec858f65c8cb4325/loro-1.5.0-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
       - pypi: https://files.pythonhosted.org/packages/51/3f/afe76f8e2246ffbc867440cbcf90525264df0e658f8a5ca1f872b3f6192a/markdown-3.8-py3-none-any.whl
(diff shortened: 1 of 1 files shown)


manzt:
Ah actually, this is related to #6237... going to change the base branch

schlich:
@manzt could be misunderstanding but i don't think pinning the Pixi version would help with an outdated lock file.  You'd want to add the `--frozen` argument to whichever command was triggering the failure -- see [Lockfile -- How to use a lockfile](https://pixi.sh/dev/workspace/lockfile/#how-to-use-a-lock-file).

I can take a look at the workflow runs and make a PR for this if it's indeed triggering a failure.  Not exactly sure what you were seeing in #6237, but it's also possible we'd (also) want to add the `--locked` flag earlier on in CI.  I think ideal scenario would be something like:

- add `pixi install/run/whatever --locked` early on in CI, possibly even as a pre-commit hook or using the pre-commit CI service.  This will help prevent other developers from pulling down misconfigured changes.
- add `pixi install/run/whatever --frozen` for final checks before deployment to make sure that CI is not using different pinned versions of dependencies than production is

does that sound about right?

schlich:
ah, okay, so i'm assuming [this workflow run](https://github.com/marimo-team/marimo/actions/runs/17472457714/job/49624024219#step:3:12) was the culprit. and it looks like  the `setup-pixi` action automatically runs `pixi install --locked` after installation, which is great!  I think the only thing further that can really be done is add a pre-commit hook, or adjust the workflows to surface this error before other checks that maybe should be downstream of the dependency installation steps.  I can put in a PR for the pre-commit hook at least.

manzt:
> Could be misunderstanding but i don't think pinning the Pixi version would help with an outdated lock file.

Yup, I was wrong. We added a new dependency in #6234, which is what was causing the `--locked` to fail (rightfully so!). However, I did notice that updating the lockfile added more than just the dependency (msgspec), some extra metadata I think.

PR #6257, opened 2025-09-05
Add multi-column sorting with stack-based behavior

lucharo:
## Summary
Implements multi-column sorting for marimo tables with stack-based behavior. Clicking a column sort moves it to the highest priority (end of sort array). Clicking the same direction again removes that sort. Visual indicators show sort priority (1, 2, 3...) in dropdown menus. "Clear sort" button adapts to "Clear all sorts" when multiple columns are sorted.

Backend uses `list[SortArgs]` where each `SortArgs` contains `by: ColumnName` and `descending: bool`. Frontend sends sort state as `[{by: string, descending: boolean}]`. Compatible with latest upstream table manager refactoring (IbisTableManager now extends NarwhalsTableManager).

🤖 Generated with [Claude Code](https://claude.ai/code)


marimo/_plugins/ui/_impl/table.py
```diff
@@ -135,12 +141,6 @@ class SearchTableResponse:
     cell_styles: Optional[CellStyles] = None
 
 
-@dataclass(frozen=True)
-class SortArgs:
-    by: ColumnName
-    descending: bool
```
Light2Dark, line 149: I think a cleaner API might just be list[SortArgs], you will also avoid zip'ing it up manually

marimo/_plugins/ui/_impl/tables/default_table.py
```diff
@@ -362,73 +362,103 @@ def get_sample_values(self, column: str) -> list[Any]:
         return self._as_table_manager().get_sample_values(column)
 
     def sort_values(
-        self, by: ColumnName, descending: bool
+        self, by: list[ColumnName], descending: list[bool]
```
Light2Dark: and here, list[tuple[ColumnName, bool]]

marimo/_plugins/ui/_impl/tables/default_table.py
```diff
@@ -362,73 +362,103 @@ def get_sample_values(self, column: str) -> list[Any]:
...
-                    # (False, None) will be for None values.
-                    # (True, x) will be for other values.
-                    # As True > False, other values come before None values
-                    if descending:
-                        return (
-                            sort_column[i] is not None,
-                            str(sort_column[i]),
+            # For column-oriented data (dict of lists)
+            data_dict = cast(dict[str, list[Any]], self.data)
+            indices = list(range(len(next(iter(data_dict.values())))))
```
Light2Dark: maybe we can simplify/optimize this, it's difficult to read.

lucharo:
Hey @Light2Dark, thanks for the quick review. I submitted this PR as a draft, as I still need some help with the front end that I can't quite get. I followed your comments and changed the sorting elements to tuples, which is easier to reason about as you suggested! 

I also need your feedback/help on how to best present the multisort. Initially, the multi-sort was enabled by clicking the Shift key, but then I thought that'd be too implicit and doesn't really resonate with the rest of marimo's design and might even clash with some users' keyboard shortcuts. So I've settled for including a checkbox for the currently selected sorted state (descending/ascending for any given column and including an explicit "Remove sort" button), but the design is really subpar. Front-end is not my thing. 

I'm hitting a union type issue I can't quite get my head around, see below for what this all looks like and the error I am getting (all in port 3000 after running `make dev`):
```
Error calling function search: Value '[{'by': 'Name', 'descending': False}]' does not fit any type of the union
```

Screenshot 2025-09-07 at 20 09 47

Light2Dark:
I tried out the branch, you may need to stop and run the make dev server again for it to catch the new API. I would remove the checkbox too, I don't think it's necessary. What might be nice is to have the Asc / Desc icon be clickable to remove the sort directly, instead of opening the menu. But we can do this in follow-ups.

frontend/src/components/data-table/header-items.tsx
```diff
@@ -163,22 +163,99 @@ export function renderCopyColumn<TData, TValue>(column: Column<TData, TValue>) {
...
-export function renderSorts<TData, TValue>(column: Column<TData, TValue>) {
+export function renderSorts<TData, TValue>(
+  column: Column<TData, TValue>,
+  table?: Table<TData>,
+) {
   if (!column.getCanSort()) {
     return null;
   }
 
+  // If table is available, use full multi-column sort functionality
```
Light2Dark: Hey Lucharo, I think this will be a nicer refactor, you can use the column.toggleSorting to achieve what you want instead of manually setting the table sort

```ts
export function renderSorts<TData, TValue>(
  column: Column<TData, TValue>,
  table?: Table<TData>,
) {
  if (!column.getCanSort()) {
    return null;
  }

  const sortDirection = column.getIsSorted();
  const sortingIndex = column.getSortIndex();

  const sortingState = table?.getState().sorting;
  const hasMultiSort = sortingState?.length && sortingState.length > 1;

  const renderSortIndex = () => {
    return (
      <span className="ml-auto text-xs font-medium">{sortingIndex + 1}</span>
    );
  };

  const renderClearSort = () => {
    if (!sortDirection) {
      return null;
    }

    if (!hasMultiSort) {
      // render clear sort for this column
      return (
        <DropdownMenuItem onClick={() => column.clearSorting()}>
          <ChevronsUpDown className="mo-dropdown-icon" />
          Clear sort
        </DropdownMenuItem>
      );
    }

    // render clear sort for all columns
    return (
      <DropdownMenuItem onClick={() => table?.resetSorting()}>
        <ChevronsUpDown className="mo-dropdown-icon" />
        Clear all sorts
      </DropdownMenuItem>
    );
  };

  const toggleSort = (direction: SortDirection) => {
    // Clear sort if clicking the same direction
    if (sortDirection === direction) {
      column.clearSorting();
    } else {
      // Toggle sort direction
      const descending = direction === "desc";
      column.toggleSorting(descending, true);
    }
  };

  return (
    <>
      <DropdownMenuItem
        onClick={() => toggleSort("asc")}
        className={sortDirection === "asc" ? "bg-accent" : ""}
      >
        <AscIcon className="mo-dropdown-icon" />
        Asc
        {sortDirection === "asc" && renderSortIndex()}
      </DropdownMenuItem>
      <DropdownMenuItem
        onClick={() => toggleSort("desc")}
        className={sortDirection === "desc" ? "bg-accent" : ""}
      >
        <DescIcon className="mo-dropdown-icon" />
        Desc
        {sortDirection === "desc" && renderSortIndex()}
      </DropdownMenuItem>
      {renderClearSort()}
      <DropdownMenuSeparator />
    </>
  );
}
```

marimo/_plugins/ui/_impl/tables/default_table.py
Light2Dark:
```python
            first_column = next(iter(data_dict.values()))
            num_rows = len(first_column)
            indices = list(range(num_rows))
```
I'm not sure if sorting for default_table is expensive. If not, then this looks fine. 
If you do see optimizations, then it would be nice to add.

Light2Dark (commented):
this looks great! small review

lucharo:
could you re-run this step @Light2Dark , I think it was a network error: https://github.com/marimo-team/marimo/actions/runs/18248342628/job/51959424513?pr=6257

Light2Dark:
re-running, @lucharo there are some failing tests

.claude/settings.local.json
```diff
@@ -0,0 +1,3 @@
+{
+  "outputStyle": "friendly-educational"
```
mscolnick: should this be committed? does it make sense to have this for everyone?

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -303,15 +303,18 @@ def _download_as(self, args: DownloadAsArgs) -> str:
...
 
         if query:
             result = result.search(query)
 
-        if sort and sort.by in result.get_column_names():
-            result = result.sort_values(sort.by, sort.descending)
+        if sort:
+            # Check that all columns exist
+            existing_columns = set(result.get_column_names())
+            if all(s.by in existing_columns for s in sort):
```
mscolnick: maybe we should just filter to the columns that do exist, then sort with those. otherwise the UI may be inconsistent (it says its sorting, but it is not)

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1175,8 +1179,11 @@ def _apply_filters_query_sort(
         if query:
             result = result.search(query)
 
-        if sort and sort.by in result.get_column_names():
-            result = result.sort_values(sort.by, sort.descending)
+        if sort:
+            # Check that all columns exist
+            existing_columns = set(result.get_column_names())
+            if all(s.by in existing_columns for s in sort):
```
mscolnick: same here, maybe just filter to valid/sortable columns

P_empty
mscolnick, line 1: delete file?

.claude/settings.local.json
lucharo: nope! That was a genuine mistake sorry!

Light2Dark (approved):
lgtm! thanks for the contribution

lucharo:
CI finally passed! Thanks @Light2Dark for the review and help throughout this PR! 🙏🏼

PR #6278, opened 2025-09-08
fix: ai endpoints from msgspec migration (more fixes)

mscolnick:
`ChatPart` is not a perfect discriminate union (because tool is `tool-{tool_name}` - blame ai-sdk), so on deserialization, we iterate one by one, finding the valid union type


Light2Dark (dismissed):
awesome!

Light2Dark (approved):

PR #6279, opened 2025-09-08
feat: stream pkg installation logs

mscolnick:
Stream logs during package installation back to the frontend

Screenshot 2025-09-05 at 5 38 57 PM
Screenshot 2025-09-05 at 5 38 50 PM

### DIFF
--- a/frontend/src/components/editor/package-alert.tsx
+++ b/frontend/src/components/editor/package-alert.tsx
@@ -199,7 +199,7 @@ export const PackageAlert: React.FC = () => {
     const { status, title, titleIcon, description } =
       getInstallationStatusElements(packageAlert.packages);
     if (status === "installed") {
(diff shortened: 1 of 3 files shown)


frontend/src/components/editor/package-alert.tsx
```diff
@@ -196,14 +199,14 @@ export const PackageAlert: React.FC = () => {
     const { status, title, titleIcon, description } =
       getInstallationStatusElements(packageAlert.packages);
     if (status === "installed") {
-      setTimeout(() => clearPackageAlert(packageAlert.id), 10_000);
+      // setTimeout(() => clearPackageAlert(packageAlert.id), 10_000);
```
Light2Dark: I think we may want this for run mode?

mscolnick: ok yea i just removed this for testing and forgot to undo it

PR #6290, opened 2025-09-09
Prefer `QueueType` annotation over concrete `queue.Queue`

manzt:
Follow up to #6262. Uses `QueueType` protocol over `queue.Queue` in a few other spots. Should help to avoid type-casting.


manzt:
Think the failing test is unrelated, going to merge.

PR #6303, opened 2025-09-10
Add experimental ZeroMQ module for multi-environment IPC

manzt:
Introduces a private `_zeromq` submodule that enables marimo-lsp to control marimo kernels via ZeroMQ-based inter-process communication. This allows the LSP to launch and manage marimo kernels in target Python environments (where only marimo is installed).

The module provides:
- `ZeroMqQueueManager` for bidirectional queue management over ZeroMQ
- `launch_kernel` entry point that proxies ZeroMQ messages to marimo's runtime

This is experimental and intentionally private (NOT intended for direct use). Users only need marimo installed in their target kernel environment for marimo-lsp integration to work.


mscolnick (approved):

marimo/_ipc/launch_kernel.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2025 Marimo. All rights reserved.
+"""Standalone kernel server entry point for IPC (using ZeroMQ)."""
+
+from __future__ import annotations
+
+import sys
+
+from marimo._ipc.queue_manager import QueueManager
```
dmadisetti, line 8: Should we do a dep check to yield a better error?

manzt: We could, but this really isn't a path users should ever hit on their own... The VS Code extension / LSP is responsible for launching the kernel.

PR #6308, opened 2025-09-10
Attempt to fix intermittent CLI test failures

manzt:
The CLI tests have been failing intermittently in CI with "Connection
refused" errors. It seems the `mcp_docs = true` in our `pyproject.toml`
is causing MCP servers to attempt initialization during tests.

This change disables `mcp_docs` in the test workflow before running CLI
tests.


manzt:
meh, doesn't seem to work...

PR #6265, opened 2025-09-05
fix(mcp-server): create dataclass wrapper for VariableValue to resolve Pydantic compatibility

bjoaquinc:
## 📝 Summary

During migration of the mcp-server from TypedDict to dataclass `VariableValue` caused a pydantic parsing error due to `mgspec.Struct` not being a pydantic-compatible type. I resolved this by creating a type called `CellVariableValue` that mirrors `VariableValue` but is pydantic-compatible.

Screenshot 2025-09-06 at 12 38 18 AM

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

PR #6309, opened 2025-09-10
refactor(tools): introduce ToolBase and centralize tool implementations

bjoaquinc:
## 📝 Summary

Introduced class based central tool management with inheritance and an MCP adapter function. Updated `marimo/_mcp/server/tools` to just handle registration.

## 🔍 Description of Changes

- Added ToolBase for general tool logic, adapters, and type handling
- Migrated tools from `marimo/_mcp/server/tools` to `marimo/_ai/tools/tools`
- Created dedicated types.py file in marimo/_ai/tools/types.py
- Added tests for `marimo/_ai/tools/tools`
- Changed tests in `marimo/_mcp/server/tools` to only assert registration

## 📋 Checklist

@mscolnick @Light2Dark


bjoaquinc:
Looks like the internal AI tool types/modules are being surfaced in the public API via import side effects. I don’t think we want that?

marimo/_ai/tools/utils/exceptions.py
```diff
@@ -0,0 +1,95 @@
...
+    """Structured representation of tool execution error details."""
+
+    message: str
+    code: str = "TOOL_ERROR"
+    status: int = 400
+    is_retryable: bool = False
+    suggested_fix: Optional[str] = None
+    meta: Optional[dict[str, Any]] = None
+
+    def to_dict(self) -> dict[str, Any]:
```
mscolnick: dataclasses have an `asdict` in the module. can we use that instead?

marimo/_ai/tools/utils/exceptions.py
```diff
@@ -0,0 +1,95 @@
...
+            "message": self.message,
+            "code": self.code,
+            "status": self.status,
+            "is_retryable": self.is_retryable,
+            "suggested_fix": self.suggested_fix,
+            "meta": self.meta or {},
+        }
+
+
+class ToolExecutionError(Exception):
```
mscolnick: can you combine these two classes?

```
@dataclass
class ToolExecutionError(Exception):
    tool_message: str
    code: str = "TOOL_ERROR"
    status: int = 400
    is_retryable: bool = False
    suggested_fix: Optional[str] = None
    meta: Optional[dict[str, Any]] = None
```

this might simplify a lot of the code

marimo/_ai/tools/base.py
```diff
@@ -0,0 +1,155 @@
...
+            raise TypeError(
+                f"Args must inherit from pydantic.BaseModel; got {self.Args!r}"
+            )
+        if not issubclass(self.Output, BaseModel):
+            raise TypeError(
+                f"Output must inherit from pydantic.BaseModel; got {self.Output!r}"
+            )
+
+    @abstractmethod
+    def __call__(self, args: ArgsT) -> OutT | Awaitable[OutT]:
```
mscolnick: i think it might be better to make create a specific handler instead of making `__call__` the abstract method.

```
    def handle(self, args: ArgsT) -> OutT | Awaitable[OutT]:
```

then we can push more boiler plate into the call:

```
    def __call__(self, args: ArgsT) -> OutT | Awaitable[OutT]:
	    try:
            Logger.info
             self.handle(args, ToolContext(app=self.app))
        except:
            Logger.error
```

marimo/_ai/tools/base.py
```diff
@@ -0,0 +1,155 @@
...
+        Adapters should always use this.
+        """
+        result = self.__call__(args)
+        if inspect.isawaitable(result):
+            return await result  # type: ignore[return-value]
+        return result  # type: ignore[return-value]
+
+    # adapters
+
+    def as_mcp_tool_fn(self) -> GenericMcpHandler[ArgsT, OutT]:
```
mscolnick: if we did:
1. `def handler` is the abstract method
2. `def __call__` is what you are calling `invoke` (we could both, `invoke` is a nice alias)

you might be able to just implement

`def __name__`() and `def __doc__` on this BaseTool and it might just be mpc compatible as-is

marimo/_ai/tools/types.py
```diff
@@ -0,0 +1,147 @@
+from __future__ import annotations
+
+import sys
+from enum import Enum
+from typing import TYPE_CHECKING, Any, Literal, Optional, Protocol, TypeVar
+
+from pydantic import BaseModel, Field
```
mscolnick: if possible, could we use `msgspec`? we have this as a dependency now. importing this file may cause build issues otherwise

marimo/_ai/tools/types.py
```diff
@@ -0,0 +1,147 @@
...
+    MARKDOWN = "markdown"
+    SQL = "sql"
+
+
+class GetLightweightCellMapArgs(BaseModel):
+    session_id: str
+    preview_lines: int = Field(default=3, ge=1, le=50)
+
+
+class LightweightCellInfo(BaseModel):
```
mscolnick: i think these types might be better in their respective tool, since they aren't used elsewhere

tests/_ai/tools/tools/test_cells.py
```diff
@@ -0,0 +1,170 @@
...
+    CellErrors,
+    CellRuntimeMetadata,
+    CellVariables,
+    CellVariableValue,
+)
+from marimo._messaging.cell_output import CellChannel
+from marimo._messaging.ops import VariableValue
+
+
+class MockCellOp:
```
mscolnick, line 20: these can all be `dataclasses` for less LOCs

tests/_ai/tools/test_base.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import annotations
+
+import inspect
+from typing import Any
+
+from pydantic import BaseModel
```
mscolnick: pydantic -> msgspec here as well

tests/_mcp/server/tools/test_registration_cells.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+pytest.importorskip("mcp", reason="MCP requires Python 3.10+")
+
+from mcp.server.fastmcp import FastMCP
```
mscolnick: this is not a core dep, so you will need to import it in the test and check if it is installed with `DependencyManager.mcp.has()` with `pytest.mark.skipif`

tests/_mcp/server/tools/test_registration_notebooks.py
```diff
@@ -0,0 +1,27 @@
...
+
+pytest.importorskip("mcp", reason="MCP requires Python 3.10+")
+
+from mcp.server.fastmcp import FastMCP
+from starlette.applications import Starlette
+
+from marimo._mcp.server.tools.notebooks import register_notebooks_tools
+
+
+def test_register_notebooks_tools_registers_callable(monkeypatch) -> None:
```
mscolnick: actually, i think this and the other test are useless. its just going to break each time we add new tools. i would just delete them

tests/_mcp/server/tools/test_registration_cells.py
mscolnick: nvm we can delete this file

marimo/_ai/tools/tools/cells.py
```diff
@@ -0,0 +1,367 @@
...
+                message=(
+                    "Refer to cells ordinally in the following format: @[cell:1]. "
+                    "Do _not_ use cell_id when discussing with users."
+                ),
+            )
+
+        except ToolExecutionError:
+            raise
+        except Exception as e:  # pragma: no cover - defensive guardrail
+            raise ToolExecutionError(
```
mscolnick: if possible, id bubble up these try/excepts that wrap the whole function into the "framework"/BaseTool

marimo/_ai/tools/tools/cells.py
```diff
@@ -0,0 +1,367 @@
...
+
+class GetLightweightCellMap(
+    ToolBase[GetLightweightCellMapArgs, GetLightweightCellMapOutput]
+):
+    """Get a lightweight map of cells showing the first few lines of each cell.
+
+    This tool provides an overview of notebook structure for initial navigation,
+    showing a preview of each cell's content without full code or outputs.
+
+    Args:
```
mscolnick: these args docs should live on the dataclass and not on this class

marimo/_ai/tools/tools/cells.py
```diff
@@ -0,0 +1,367 @@
...
+                    is_retryable=False,
+                )
+
+            app_state = AppStateBase.from_app(self.app)
+            session_id = args.session_id
+            cell_id = args.cell_id
+
+            # Access session manager and get the specific session
+            session_manager = app_state.session_manager
+            session_id_typed = SessionId(session_id)
```
mscolnick: you can likely just add `session_id: SessionId` to the args so you don't need to cast it

marimo/_ai/tools/tools/cells.py
```diff
@@ -0,0 +1,367 @@
...
+                session.app_file_manager.filename or "untitled_notebook.py"
+            )
+
+            cells: list[LightweightCellInfo] = []
+            for cell_data in cell_manager.cell_data():
+                code_lines = cell_data.code.split("\n")
+                preview = "\n".join(code_lines[: args.preview_lines])
+
+                # Determine cell type using compiled cell info when available
+                cell_type: SupportedCellType
```
mscolnick: this might be easier to read as a functin `def get_cell_type()`, then you can early return and avoid nested if/else and ternaries. you can probably get it with only one level of if statements and early returns

marimo/_ai/tools/base.py
mscolnick: i would also make the handler something like:
```
@dataclass
class ToolContext
       app: App | None

       @property
       def session(session_id: SessionId): Session
              .. do the validation and grab the session or raise an error

 def handle(self, args: ArgsT, ctx: ToolContext) -> OutT | Awaitable[OutT]
     ...
```

marimo/_ai/tools/base.py
```diff
@@ -0,0 +1,155 @@
...
+    TYPE_CHECKING,
+    Any,
+    Generic,
+    Optional,
+    cast,
+    get_args,
+    get_origin,
+)
+
+from pydantic import BaseModel
```
Light2Dark: just checking, if this is run in core mode (no openai installed), does it raise errors on run. Since pydantic isn't installed by default afaik.

marimo/_ai/tools/utils/exceptions.py
bjoaquinc: Done, updated it!

marimo/_ai/tools/base.py
bjoaquinc: Following Myles reco, I migrated away from using pydantic and instead used dataclasses for everything.

marimo/_ai/tools/utils/exceptions.py
bjoaquinc: Done! Yeah much cleaner.

marimo/_ai/tools/base.py
bjoaquinc: Yeah I reversed the order of __call__ and invoke. I think your approach makes more sense. R.e. the ToolContext ok that helps a lot, I'll update the ToolContext to be like that. I changed the context to include session_manager instead of the full app for portability. I also created these helper methods in ToolBase:

```python
    def _get_session_manager(self) -> SessionManager:
        if self.session_manager is None:
            raise ToolExecutionError("Session manager is not available", code="SESSION_MANAGER_NOT_AVAILABLE", is_retryable=False, suggested_fix="Try restarting the marimo server.")
        return self.session_manager

    def _get_session(self, session_id: str):
        """Return a session by id or raise a standardized tool error.

        Raises SESSION_NOT_FOUND with suggested fix and meta including the id.
        """
        session_manager = self._get_session_manager()
        typed_id = SessionId(session_id)
        if typed_id not in session_manager.sessions:
            raise ToolExecutionError(
                f"Session {session_id} not found",
                code="SESSION_NOT_FOUND",
                is_retryable=False,
                suggested_fix="Use get_active_notebooks to find valid session IDs",
                meta={"session_id": session_id},
            )
        return session_manager.sessions[typed_id]
```

bjoaquinc: Also did you want this to be in `base.py` or `types.py`:

```
@dataclass
class ToolContext
       app: App | None

       @property
       def session(session_id: SessionId): Session
              .. do the validation and grab the session or raise an error
```

marimo/_ai/tools/base.py
bjoaquinc: Oh cool! I'll try this

bjoaquinc: Hmmm... Tried this but it still breaks MCP’s Pydantic schema inference. Pydantic sees Coroutine[Any, Any, Any] and can’t build a concrete schema. Maybe keep my previous implementation?

marimo/_ai/tools/types.py
bjoaquinc: My concern with this is that all types defined here will be pushed into the MCP sdk for pydantic validation. So if we use msgspec.Struct we might run into this issue again:

#6265 

Maybe we should explore the suggestion of Trevor now that we have more usecases of Pydantic validation?

marimo/_ai/tools/types.py
bjoaquinc: I agree. I'll move it back on top of the tool and keep `types.py` for common types only

marimo/_ai/tools/base.py
mscolnick: probably base, given the logic and all

marimo/_ai/tools/base.py
mscolnick: yea, no worries. its still only written once and not over and over. find to keep as is

marimo/_ai/tools/types.py
mscolnick: yea dataclasses if fine then

mscolnick: we can also explore Trevor's suggestion, but maybe in a followup

marimo/_ai/tools/base.py
bjoaquinc: done! Also moved that logic to ToolContext. It was cleaner to keep ToolContext as the boundary for accessing app data

marimo/_ai/tools/types.py
bjoaquinc: done!

tests/_mcp/server/tools/test_registration_cells.py
bjoaquinc: done!

tests/_mcp/server/tools/test_registration_notebooks.py
bjoaquinc: done!

marimo/_ai/tools/tools/cells.py
bjoaquinc: done! I added standardizing all unknown errors to the __call__ function in ToolBase

marimo/_ai/tools/tools/cells.py
bjoaquinc: I need the full docstring though because this also serves as the **tool description** when its being registered to the MCP server. Maybe I should make that more explicit?

marimo/_ai/tools/tools/cells.py
bjoaquinc: done! Yeah sorry that was a weird work around

marimo/_ai/tools/tools/cells.py
bjoaquinc: done! also removed nested if/else statements

tests/_ai/tools/test_base.py
bjoaquinc: changed this to dataclass instead

tests/_ai/tools/tools/test_cells.py
bjoaquinc: done!

marimo/_ai/tools/tools/cells.py
mscolnick: no, this is ok then. didn't realize it was for that

PR #6332, opened 2025-09-12
Fallback to `__dict__`/`__slots___` in `enc_hook` for unknown objects

### LINKED ISSUE #6330: Loguru logging not working anymore
### Describe the bug

After ugrade to 0.5.13, logging in marimo with loguru is broken. 
 File "...\marimo\_messaging\msgspec_encoder.py", line 81, in enc_hook
        raise NotImplementedError(f"Objects of type {type(obj)} are not supported")
    NotImplementedError: Objects of type <class 'loguru._handler.Message'> are not supported

Before the upgrade, it worked well and the logging output was shown in the under the marimo cell. 

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.15.3",
  "editable": false,
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.11",
  "Binaries": {
    "Browser": "137.0.7151.120",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.5.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.2",
    "starlette": "0.47.3",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "pandas": "2.3.2",
    "pyarrow": "21.0.0",
    "loro": "1.5.3",
    "pytest": "8.4.2",
    "ruff": "0.13.0"
  },
  "Experimental Flags": {}
}```

</details>


### Code to reproduce

import marimo

__generated_with = "0.15.3"


@app.cell
def _():
    import marimo as mo
    from loguru import logger
    return (logger,)


@app.cell
def _(logger):
    logger.info("Test")
    return


@app.cell
def _():
    return


if __name__ == "__main__":
    app.run()

### PR DESCRIPTION
**Pull request by manzt**
Fixes #6330

Extends the `enc_hook` for msgspec with a fallback to `__dict__` / `__slots__` for unknown types. Also adds back a `test_json_encoder.py` which was removed becuase we deleted the `WebComponentEncoder`, but adapted with expected outputs with `msgspec`.

A few observations:

- Overall, encoding "just works" however there are some edge cases that were updated to msgspec-specific behavior, and it would be good to look through the cases individually.

- `enc_hook` is only called if a type is unknown to `msgspec`. Importantly, protocol-based serialization with `_mime_` only works if the class is NOT a `msgspec.Struct` or `dataclasses.dataclass`

- As a result, you cannot override the encoding of any "known" types to `msgspec` within the `enc_hook`. I had to update the timedelta test to use [ISO 8901 duration strings](https://en.wikipedia.org/wiki/ISO_8601#Durations) and `float(inf | -inf | nan)` all serialize to `null`. To my knowledge there is no way to intercept this. I added a test case just to make that constraint explicit.

### DIFF
--- a/tests/_plugins/core/test_json_encoder.py
+++ b/tests/_plugins/core/test_json_encoder.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import json
+import sys
 from collections import namedtuple
 from dataclasses import dataclass
 from typing import Any, Optional
@@ -337,6 +338,9 @@ def test_enum_encoding() -> None:
     assert encoded == "1"
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 11), reason="StrEnum not supported in Python 3.10"
+)
 def test_str_enum() -> None:
     from enum import StrEnum
 
@@ -410,7 +414,7 @@ def test_complex_nested_structure():
 
 
 def test_png_encoding() -> None:
-    purple_square = "b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x14\x00\x00\x00\x14\x08\x02\x00\x00\x00\x02\xeb\x8aZ\x00\x00\x00\tpHYs\x00\x00.#\x00\x00.#\x01x\xa5?v\x00\x00\x00\x1dIDAT8\xcbc\xac\x11\xa9g \x1701P\x00F5\x8fj\x1e\xd5<\xaa\x99r\xcd\x00m\xba\x017\xd3\x00\xdf\xcb\x00\x00\x00\x00IEND\xaeB`\x82'"  # noqa: E501
+    purple_square = b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x14\x00\x00\x00\x14\x08\x02\x00\x00\x00\x02\xeb\x8aZ\x00\x00\x00\tpHYs\x00\x00.#\x00\x00.#\x01x\xa5?v\x00\x00\x00\x1dIDAT8\xcbc\xac\x11\xa9g \x1701P\x00F5\x8fj\x1e\xd5<\xaa\x99r\xcd\x00m\xba\x017\xd3\x00\xdf\xcb\x00\x00\x00\x00IEND\xaeB`\x82"  # noqa: E501
     encoded = encode_json_str(purple_square)
     assert isinstance(encoded, str)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/_plugins/core/test_json_encoder.py [hunk 1]
```diff
@@ -0,0 +1,471 @@
...
+    def _mime_(self) -> tuple[str, str]:
+        return "text/plain", "data"
+
+    @property
+    def __dict__(self) -> dict:
+        return {}
+
+
+def test_mime_encoding() -> None:
+    mime_obj = MockMIMEObject()
```
manzt, line 115: If `MockMIMEObject` is a `dataclass`, this just returns `{}`, since the check for `_mime_` is never hit within the `enc_hook`.

tests/_plugins/core/test_json_encoder.py
```diff
@@ -0,0 +1,471 @@
...
+def test_null_encoding() -> None:
+    null = None
+    encoded = encode_json_str(null)
+    assert encoded == "null"
+
+
+def test_inf_encoding() -> None:
+    inf = float("inf")
+    encoded = encode_json_str(inf)
+    assert encoded == "null"  # msgspec encodes infinity as null
```
manzt, line 246: Was `"Infinity"` before... not sure how we can intercept

tests/_plugins/core/test_json_encoder.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+def test_nan_encoding() -> None:
+    nan = float("nan")
+    encoded = encode_json_str(nan)
+    assert encoded == "null"  # msgspec encodes NaN as null
```
manzt, line 252: Was `"NaN"` before, not sure how we can intercept.

tests/_plugins/core/test_json_encoder.py
```diff
@@ -0,0 +1,471 @@
...
+def test_str_enum() -> None:
+    from enum import StrEnum
+
+    class Color(StrEnum):
+        RED = "RED"
+        GREEN = "GREEN"
+        BLUE = "BLUE"
+
+    encoded = encode_json_str(Color.RED)
+    assert encoded == '"RED"'
```
manzt: Notice different behavior of `Enum` and `StrEnum` (handled natively by msgspec).

tests/_plugins/core/test_json_encoder.py
```diff
@@ -0,0 +1,471 @@
...
+    bytes_obj = b"hello"
+    encoded = encode_json_str(bytes_obj)
+    assert encoded == '"aGVsbG8="'
+
+
+def test_memoryview_encoding() -> None:
+    bytes_obj = b"hello"
+    memview = memoryview(bytes_obj)
+    encoded = encode_json_str(memview)
+    assert encoded == '"aGVsbG8="'
```
manzt, line 201: These were both `"hello"` before. msgspec always encodes to `base64`, not sure how to intercept.

tests/_plugins/core/test_json_encoder.py
```diff
@@ -0,0 +1,471 @@
...
+class MockDataclass:
+    a: int
+    b: str
+    items: Optional[list[Any]] = None
+    other_items: Optional[dict[str, Any]] = None
+
+
+class Button(MIME):
+    def _mime_(self) -> tuple[str, str]:
+        return "text/html", "<button>Click me!</button>"
```
manzt, line 154: This one is currently failing because `MIME` is a dataclass, so it gets serialized to `"{}"`

PR #6334, opened 2025-09-12
fix: json encode handle more human types, and more human readable for tables

mscolnick:
This adds a few more types to our `msgspec` encoding hook to handle more types.
It also adds a class `SuperJson` (which is not quite like what some "superjson" libraries do), but it does pretty-print datatypes so they are not lossy:

The key differences from the default msgspec encoder are:
  - It serializes the b'hello' to 'hello' instead of base64 encoded
  - It serializes the float('inf') to Infinity instead of null
  - It serializes the float('nan') to NaN instead of null
  - It serializes the timedelta to a human readable string instead of a ISO 8601 duration (e.g. "1 day, 2:03:00")

### DIFF
--- a/marimo/_messaging/msgspec_encoder.py
+++ b/marimo/_messaging/msgspec_encoder.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import collections
+import datetime
 import fractions
 from math import isnan
 from pathlib import PurePath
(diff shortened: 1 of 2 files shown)


marimo/_messaging/msgspec_encoder.py
```diff
@@ -105,7 +144,55 @@ def enc_hook(obj: Any) -> Any:
...
+
+    # Handle primitive types
+    if isinstance(obj, (int, bool)):
+        return obj
+
+    # Handle None
+    if obj is None:
+        return None
+
+    return str(obj)
```
manzt: so we no longer throw in this hook for unknown types. Would I make sense to use `repr` here?

mscolnick: yea i thought this would be better to, but then noticed datetime looks worse (not human-readable). maybe we can tweak this in a followup and handle datetime specifically 

image

mscolnick: nvm had a better solution

PR #6337, opened 2025-09-12
lint: add copy notice

dmadisetti:
## 📝 Summary

Add copy notice to new files


dmadisetti:
@bjoaquinc No biggie, but we have the script `./scripts/pycopyright.sh` which auto runs with `make py-check` (which runs all tools)

bjoaquinc:
> @bjoaquinc No biggie, but we have the script `./scripts/pycopyright.sh` which auto runs with `make py-check` (which runs all tools)

Got it thanks. I'll make sure to use make py-check in my next commits.

PR #6339, opened 2025-09-12
feat: pass in cli flags via `app.run(*cli_args)`

### LINKED ISSUE #6062: Allow passing arguments to App.run()
### Description

A notebook can be executed as a script via command line, with arguments passed as CLI. But it's sometimes useful to be able to run notebooks from within an existing python script (without using App.embed()). We can do this with my_notebook.app.run(), but don't have a convenient way of passing parameters (AFAIK). 

### Suggested solution

It would be useful if we could pass in regular python objects as arguments to the run() method, which are then accessible within the notebook. These arguments could be thought of as some kind of overrides. Regarding their access pattern and automatic dependency tracking, I'm not sure what the desirable behaviour would be, e.g., could they be exposed directly, or via some other api call?

- [ ] Yes

### Alternatives

A workaround is to introduce a separate module which is imported by the caller and notebook, but patched by the caller before the notebook is run. This is awkward and prone to buggy behavior, though.

### Additional context

_No response_

dmadisetti:
## 📝 Summary

fixes #6062

This PR adds the ability to pass CLI arguments directly to the App.run() method, enabling programmatic execution of marimo apps with custom parameters.

### Usage

```python
from my_notebook import app

# Run with custom CLI arguments
outputs, defs = app.run("--batch_size", "64", "--model", "transformer")
```

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -617,17 +617,20 @@ class App:
             with app.setup:
                 import pandas as pd
 
+
             @app.cell
             def config():
                 batch_size = 32
                 learning_rate = 0.01
                 return batch_size, learning_rate
 
+
             @app.cell
             def process_data(pd, batch_size, learning_rate):
                 data = pd.DataFrame({"x": [1, 2, 3]})
                 result = data * batch_size * learning_rate
                 return (result,)
+
 
             if __name__ == "__main__":
                 app.run()


tests/_ast/test_app.py
```diff
@@ -732,6 +732,59 @@ def __():
...
+            args = mo.cli_args()
+            return (args,)
+
+        # Test with various argument types
+        outputs, defs = app.run("--count", "42", "--name", "test", "--enabled", "true")
+
+        args = defs["args"]
+        assert args.get("count") == 42
+        assert args.get("name") == "test"
+        assert args.get("enabled") is True
```
manzt, line 751: Is this just forwarding down to how we already handle CLI args? Where does the type handling come from?

manzt (commented):
This looks good to me, especially since it’s just inheriting the CLI behavior. My main comment is about the function signature: I’d prefer a single `args=` argument (similar to `subprocess.run`) instead of collecting multiple positional arguments.

Using `args=` gives us more flexibility if we want to extend `run()` in the future. I know we could still add keyword-only args later with this API, but IMO it gets kind of awkward when there are lots of positionals with options as keyword only args.

So instead of:

```python
outputs, defs = app.run("--param1", "value1", "--param2", "value2")
```

I’d suggest:

```python
outputs, defs = app.run(args=["--param1", "value1", "--param2", "value2"])
```

tests/_ast/test_app.py
dmadisetti: https://github.com/marimo-team/marimo/blob/be3e52b30e564288f5ec56e57b603e5575ccb984/marimo/_cli/parse_args.py#L13

Which is honestly pretty brittle and should be refreshed, but it's probably good to have the logic in a single place.

dmadisetti:
Good call on the API, went ahead and updated it

mscolnick:
I think args should just be a dictionary, since then you cannot override args in nested app that are more complex types outside of strings (and number or booleans, but the downstream app is likely not handling that). 

Why not something like:
```
app.run(args={batch_size: 64, model="transformer"}
```

marimo/_runtime/app/script_runner.py
```diff
@@ -149,7 +150,7 @@ async def _run_asynchronous(
                             hook()
         return outputs, glbls
 
-    def run(self) -> RunOutput:
+    def run(self, cli_args: SerializedCLIArgs | None = None) -> RunOutput:
```
mscolnick, line 153: i don't think we should expose `SerializedCLIArgs` here and instead expose a `dict`

mscolnick:
The issue asks: 

> It would be useful if we could pass in regular python objects as arguments to the run() method,

which i don't think this actually solves (**regular python objects**)

dmadisetti:
> regular python objects

We already have this behavior for the cell level. e.g.

```python
from mynotebook import mycell

mycell.run(override=Whatever)
```

Passing in arbitrary values into run makes the API awkward on the imported notebook side

maybe something like:

```python
value = mo.script_args("value")
...
```

But this is just more surface area and prevents some notebooks from being standalone

mscolnick:
In hindsight, i think we can scope this out more. I wonder if this should be more like `cell.run()` that overrides the parameters, so you don't need to call `mo.args()` at all.

e.g. 

```
app = marimo.App()

@app.cell()
def _():
  limit = 10
  df.head(limit)
```

then running

```
app.run(limit=20)
```

will override the limit value in the notebook. this seems more inline with how the OP of the issue may want to use it, but i would suggest proposing a few APIs

dmadisetti:
Had a good conversation where we can just mirror the `cell.run` api. @mscolnick @manzt please see this api suggestion: https://github.com/marimo-team/marimo/blob/f3225b8959762e4b0ddf1fa508a8fff16fb49756/marimo/_ast/app.py#L600-L673

Going ahead and implementing this. I'll probably close this PR and rebase to keep the discussion more streamlined, but LMK if there are any major comments

PR #6360, opened 2025-09-15
improvement: better ACP commands, edit the notebook without --watch, slash-commands

mscolnick:
- better CLI commands to start the ACP: using [stdio-to-ws](https://github.com/marimo-team/stdio-to-ws)
- Allow agent to edit the notebook without running `--watch`
- Add slash-commands and agent modes


Light2Dark (commented):
should we encourage polars instead of pandas?

Light2Dark (approved):

Issue #6281, opened 2025-09-08
Could not find function given request: FunctionCallRequest(...)


maxkoretskyi:
### Describe the bug

The issue happens on UI elements that have associated server side part, eg `mo.ui.file_browser()` and with `auto_instantiate=false` in the config.

<img width="2351" height="1264" alt="Image" src="https://github.com/user-attachments/assets/f0842286-3d24-4527-aefe-c2b38e421982" />

The widget works OK after I run the cell.

The issue is easy to reproduce. 

Start server  `marimo edit --no-token yep.py`
Add a UI widget with associated server side part, eg `mo.ui.file_browser()`. 
Run widget.
Stop server.
Start serer again.
Refresh browser.

I'm willing to submit a PR if needed, just need a bit of guidance here.

### Will you submit a PR?

- [x] Yes

### Environment

{
  "marimo": "0.15.1",
  "editable": false,
  "OS": "Linux",
  "OS Version": "5.15.167.4-microsoft-standard-WSL2",
  "Processor": "x86_64",
  "Python Version": "3.10.4",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.17.1"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.1.1",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "typing-extensions": "4.14.1",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "openai": "1.102.0",
    "pandas": "2.3.2"
  },
  "Experimental Flags": {}
}

### Code to reproduce

```
import marimo

__generated_with = "0.15.2"
app = marimo.App(width="medium")


@app.cell
def _():
    import marimo as mo
    return (mo,)


@app.cell
def _(mo):
    mo.ui.file_browser()
    return


if __name__ == "__main__":
    app.run()

```

maxkoretskyi:
[...]

PR #6365, opened 2025-09-16
fix [6281]: Could not find function given request: FunctionCallReques…

maxkoretskyi:
## 📝 Summary

The error `Could not find function given request: FunctionCallRequest(...)` is shown for UI elements that have associated server side part, eg mo.ui.file_browser() and with auto_instantiate=false in the config. Fixes [6281](https://github.com/marimo-team/marimo/issues/6281).

## 🔍 Description of Changes

Check if the cell is not yet instantiated (haved run), and show user friendlier error message that tells them they need to re-run the cell. If there's no data (initial run), then don't show directory selector widget in the output, just the message:

image

The original FunctionCallRequest(...) error is caused by the fact, that cell with `auto_instantiate=false` loads the widget from the cache, which immediately calls plugin function (eg list_directories), but the runtime graph is empty and doesn't have any plugin functions registered b/c the cell hasn't run.

### DIFF
--- a/frontend/src/plugins/core/registerReactComponent.tsx
+++ b/frontend/src/plugins/core/registerReactComponent.tsx
@@ -1,4 +1,19 @@
 /* Copyright 2024 Marimo. All rights reserved. */
+import { notebookAtom } from "@/core/cells/cells.ts";
+import { findCellId } from "@/core/cells/ids.ts";
+import { isUninstantiated } from "@/core/cells/utils"
+import { createInputEvent, MarimoValueUpdateEvent } from "@/core/dom/events";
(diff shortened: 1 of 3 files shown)


maxkoretskyi:
@mscolnick do you think maybe it makes sense to hide console output as well for uninstantiated cells?

mscolnick (approved):

mscolnick:
@maxkoretskyi showing console outputs are fine (and desired)

maxkoretskyi:
@mscolnick ok, cool, but those logs are from the previous run (cached), users should understand that :)

PR #6380, opened 2025-09-16
internal: pass around single bytes of message for KernelMessage

mscolnick:


marimo/_pyodide/bootstrap.py
```diff
@@ -78,8 +81,9 @@ def create_session(
     """
 
     def write_kernel_message(op: KernelMessage) -> None:
-        op_name, data_bytes = op
-        text = f'{{"op": "{op_name}", "data": {data_bytes.decode("utf-8")}}}'
+        data_json_str = op.decode("utf-8")
+        op_name = deserialize_kernel_message(op).name
+        text = f'{{"op": "{op_name}", "data": {data_json_str}}}'
```
manzt, line 87: Seems a little unfortunate that we have to deserialize and serialize again. Ideally it would be something like:

```py
    def write_kernel_message(op: KernelMessage) -> None:
        message_callback(op.decode("utf-8"))
```

But I'm guessing this is still needed for the front end?

manzt (approved):
Looks good. It's much nicer to have the `NewType` type to pass around.

PR #6382, opened 2025-09-16
fix: handle nested msgspec parsing

mscolnick:
fix: handle nested msgspec parsing

### DIFF
--- a/marimo/_utils/parse_dataclass.py
+++ b/marimo/_utils/parse_dataclass.py
@@ -156,7 +156,7 @@ class _DataclassParser:
         if issubclass(cls, msgspec.Struct):
             return _parse_msgspec(
                 value, strict=not self.allow_unknown_keys, cls=cls
-            )  # type: ignore[no-any-return]
+            )  # type: ignore[return-value]
 
         try:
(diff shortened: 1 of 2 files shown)


marimo/_utils/parse_dataclass.py
```diff
@@ -194,6 +199,14 @@ def build_dataclass(self, values: dict[Any, Any], cls: type[T]) -> T:
         return cls(**transformed)
 
 
+def _parse_msgspec(value: Any, *, strict: bool, cls: type[T]) -> T:
+    # If it is a dict, it is already parsed and we can just build the dataclass.
+    if isinstance(value, dict):
+        return msgspec.convert(value, strict=strict, type=cls)
+
```
manzt, line 208: should we assert that value is either `bytes | str`?

manzt: Or alternatively, update the type signature:

```py
def _parse_msgspec(value: dict | str | bytes, strict: bool, cls: type[T]) -> T:
```

manzt (approved):
Nice changes

PR #6400, opened 2025-09-17
tests: tests for edge case file names

mscolnick:
some tests to validate weird filenames do work


dmadisetti (dismissed):
LGTM if they all pass. Hopefully catch future regressions

dmadisetti (dismissed):
Cool. Fair enough

dmadisetti (approved):

PR #6403, opened 2025-09-17
fix: shelex paths for lsp

dmadisetti:
## 📝 Summary

If a user creates a virtual environment in `/a path with spaces/` then the copilot lsp location may not be properly resolved. This adds string release to properly pass in the binary location.

Additionally, added the `location` key to environment to help debug in the future.

### DIFF
--- a/marimo/_server/lsp.py
+++ b/marimo/_server/lsp.py
@@ -326,10 +326,15 @@ class TyServer(BaseLspServer):
         return True
 
     def get_command(self) -> list[str]:
+        import shlex
+
         from ty.__main__ import find_ty_bin  # type: ignore
 
         lsp_bin = marimo_package_path() / "_lsp" / "index.cjs"
         log_file = _loggers.get_log_directory() / "ty-lsp.log"
+
+        # Properly quote the ty binary path to handle spaces and special characters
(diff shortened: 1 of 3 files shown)


marimo/_server/lsp.py
```diff
@@ -199,13 +201,16 @@ def get_command(self) -> list[str]:
...
+        copilot_command = f"node {shlex.quote(str(copilot_bin))} --stdio"
+
         return [
             "node",
             str(lsp_bin),
             "--port",
             str(self.port),
             "--lsp",
-            f"node {copilot_bin} --stdio",
+            copilot_command,
```
mscolnick, line 213: do we need to do this with `ty`?

tests/_server/test_lsp.py
```diff
@@ -131,6 +131,53 @@ def test_copilot_server():
     assert "GitHub Copilot" in alert.title
 
 
+def test_copilot_server_command_quotes_path():
```
mscolnick: you can do `tmp_path: Path` and avoid some of the setup (and cleanup) code

dmadisetti: Good call

PR #6412, opened 2025-09-18
i18n: use user's local for date inputs

mscolnick:


Light2Dark (commented):
should this provider wrap islands too?
in `web-components.tsx`

mscolnick:
> should this provider wrap islands too? in `web-components.tsx`

yep, it is there, right?

Light2Dark (approved):
oops, looks good then.

PR #6418, opened 2025-09-18
fix: pandas series with columns the same as index name

mscolnick:


Light2Dark:
let me know if you need a more performant fix, I was trying something in formatters where we only check for this when an error is raised.

PR #4245, opened 2025-03-25
Styling cells

nojaf:
## 📝 Summary

Hello, as asked on Discord, we are hoping to add some styling options to the cells of a data-table.

image

## 🔍 Description of Changes

On the Python side, I'm adding a new function that returns a dictionary-style object to describe the cell's style. This nested array of styles is then sent to the client and utilized by another TanStack plugin.

In our Discord conversation, we discussed foreground and background colors. I opted for a style dictionary, which provides more options. I'm not set on this idea, so please share your thoughts on what you think is best.

This is still a work in progress, so I would appreciate your guidance in the right direction.

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [ ] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

## 📜 Reviewers

@mscolnick

PR #6420, opened 2025-09-18
Hover over cell

nojaf:
## 📝 Summary

image

Similar to https://github.com/marimo-team/marimo/pull/4245, we would also like to have a function to provide a callback that can create hover text for the individual cells.

## 🔍 Description of Changes

This is another TanStack table plugin.
Needs some more polishing but you like to get a sense first if this would be accepted.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/renderers.tsx
+++ b/frontend/src/components/data-table/renderers.tsx
@@ -93,6 +93,25 @@ export const DataTableBody = <TData,>({
     handleCellsKeyDown,
   } = useCellRangeSelection({ table });
 
+  function applyHoverTemplate(
+    template: string,
+    cells: Array<Cell<TData, unknown>>,
+  ): string {
+    const variableRegex = /{{(\w+)}}/g;
+    // Map column id -> stringified value
+    const idToValue = new Map<string, string>();
+    for (const c of cells) {
+      const v = c.getValue();
+      // Prefer empty string for nulls to keep tooltip clean
+      const s = renderUnknownValue({ value: v, nullAsEmptyString: true });
+      idToValue.set(c.column.id, s);
+    }
+    return template.replace(variableRegex, (_substr, varName: string) => {
+      const val = idToValue.get(varName);
+      return val !== undefined ? val : `{{${varName}}}`;
+    });
+  }
+
   const renderCells = (cells: Array<Cell<TData, unknown>>) => {
     return cells.map((cell) => {
       const { className, style: pinningstyle } = getPinningStyles(cell.column);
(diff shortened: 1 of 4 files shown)


mscolnick:
@nojaf what is the use case? would it make sense to instead create a `hover_template` which is a templated string and we can do this on the frontend instead. 

something like what plotly does https://plotly.com/python/hover-text-and-formatting/#customizing-hover-text-with-a-hovertemplate

nojaf:
Hi, I asked inside and they mentioned two direct use cases:
- A hover-over header to explain the metric in more detail
- A hover-over cell to list some metadata

They use this in plotly a lot though usually provide hover_data=[Col1, Col2,..] and plotly does the formatting, probably calling hover_template in the background.

So, this would work for us.
Alas, I don't quite how this would work on a technical level.
Could you describe what needs to be done? Happy to adhere to your suggestion!

mscolnick:
@nojaf we can add a parameter `hover_template`, which could be `This car has {{num_wheels}} wheels and goes {{top_mph}}`, then on the frontend we just do a find/replace:

```ts
const variableRegex = /{{(\w+)}}/g;
const matches = [...prompt.matchAll(variableRegex)];
let result;
for (const match in matches) {
   // if match is in the row, replace it in the string
}  
```

since we pass the full fields anyways, we likely don't need the `hover_data` args, just `hover_template`

nojaf:
Hi @mscolnick, I tried something the way you described it.
Do you envision that each `hover_template` has access to all the columns of the row?
What if you want to show something specific to the cell? Should we have `{{row}}`, `{{column}}` and `{{value}}` as well? How do you envision this?

mscolnick:
yea, `hover_template` would have access to each row. and each `{{replace_me}` is a column name

mscolnick:
this replacement can happen on the frontend

frontend/src/components/data-table/renderers.tsx
```diff
@@ -101,6 +101,25 @@ export const DataTableBody = <TData,>({
         cell.getUserStyling?.() || {},
         pinningstyle,
       );
+      const hoverTemplate = table.getState().cellHoverTemplate || undefined;
```
nojaf: @mscolnick yeah this works.
But I wonder if we want to support showing something specific for each individual cell?

mscolnick: that seems like a weird user experience, because how are they going to expect what is in each popover without hovering over it. it would just be a surprise. with a template, they at least can infer what fields are being shown and changing

nojaf: Alright, makes sense. I'll stick to the column names within a row for now.

PR #6421, opened 2025-09-18
fix: remove body tags when rendering custom HTML

### LINKED ISSUE #6416: Marimo HTML wrapper makes marimo hang when using certain HTML tags like <body> or <html>
### Describe the bug

I've previously used marimo together with html templates to style my application. I used the html output of the ui components (the _.text_ property), and wrapped it into a custom HTML page which allowed me to import style sheets (css) and apply this in a component-like manner, to include borders, backgrounds etc.

Since 0.14.0 I am unable to do this. Instead, when outputting with mo.Html(...) marimo simply hangs and the browser window has to be shut down.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.0",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.9",
  "Binaries": {
    "Browser": "140.0.7339.128",
    "Node": "v20.18.1"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "2.5.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "starlette": "0.46.0",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "anywidget": "0.9.18",
    "pandas": "2.2.3",
    "loro": "1.6.0",
    "pytest": "8.3.5",
    "ruff": "0.9.9"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.14.0"


@app.cell
def _():
    import marimo as mo
    return (mo,)


@app.cell
def _(mo):
    text_input = mo.ui.text(label="Hello")
    return (text_input,)

@app.cell
def _(mo, text_input):
    html2 = mo.Html(
    f'<head><!--Here we could have a custom css style sheet that applies only to this html before 0.14.0--></head><body><div>{text_input.text}</div></body>')
    print(html2.text)
    return (html2,)


@app.cell
def _(html2):
    html2
    return


if __name__ == "__main__":
    app.run()
```

mscolnick:
Fixes #6416

### DIFF
--- a/frontend/src/plugins/core/RenderHTML.tsx
+++ b/frontend/src/plugins/core/RenderHTML.tsx
@@ -32,7 +32,7 @@ const removeWrappingBodyTags: TransformFn = (
     if (isValidElement(reactNode) && "props" in reactNode) {
       const props = reactNode.props as { children?: ReactNode };
       const children = props.children;
-      return <>{children}</>;
+      return <>{children}</>; // eslint-disable-line react/jsx-no-useless-fragment
     }
     return;
   }
@@ -47,7 +47,7 @@ const removeWrappingHtmlTags: TransformFn = (
     if (isValidElement(reactNode) && "props" in reactNode) {
       const props = reactNode.props as { children?: ReactNode };
       const children = props.children;
-      return <>{children}</>;
+      return <>{children}</>; // eslint-disable-line react/jsx-no-useless-fragment
     }
     return;
   }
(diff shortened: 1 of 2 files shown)


mrdobalina2k:
Will this still allow me to use custom html styles?

mscolnick:
@mrdobalina2k yes, it will. will just remove additional <body/<html elements since only one is allowed, but the children will be kept

PR #6432, opened 2025-09-18
improvement: add locale to display config

mscolnick:
Add the ability to configure the `locale` you want in marimo, under `config.display.locale` following [BCP47](https://www.ietf.org/rfc/bcp/bcp47.txt).

The default is not set which means we will use the user's browser locale. This config also means you can override it at the project level (pyproject.toml) and notebook level (inline-metadata)


frontend/src/components/chat/acp/__tests__/state.test.ts
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2025 Marimo. All rights reserved. */
+/* Copyright 2024 Marimo. All rights reserved. */
```
manzt, line 1: For follow up PR, but we should have a check that harmonizes the year for these copyright statements

frontend/src/core/config/config.ts
```diff
@@ -82,6 +82,10 @@ export const editorFontSizeAtom = atom<number>((get) => {
   return get(resolvedMarimoConfigAtom).display.code_editor_font_size;
 });
 
+export const localeAtom = atom<string | null | undefined>((get) => {
+  return get(resolvedMarimoConfigAtom).display.locale;
```
manzt, line 86: Could be nice to normalize to `null` or `undefined` here for the none case?

```suggestion
  return get(resolvedMarimoConfigAtom).display.locale ?? undefined;
```

frontend/src/core/config/config-schema.ts
```diff
@@ -144,6 +144,7 @@ export const UserConfigSchema = z
             }
             return width;
           }),
+        locale: z.string().nullable().optional(),
```
manzt, line 147: Not sure if it would be worth doing a transformation to a `TypedString` type if we don't want to actually validate but make sure it's coming from our config?

```ts
locale.string().nullable().optional().transform((v) => {
  if (v == null) return undefined; // normalize null | undefined
  return v as LocaleString;
})
```

frontend/src/core/i18n/__tests__/locale-provider.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+    );
+
+    const i18nProvider = getByTestId("i18n-provider");
+    expect(i18nProvider).toBeInTheDocument();
+    expect(i18nProvider.dataset.locale).toBe(testLocale);
+    expect(i18nProvider).toHaveTextContent("Test content");
+  });
+
+  it("should render I18nProvider with different locale values", () => {
+    const testCases = ["en-US", "fr-FR", "de-DE", "ja-JP"];
```
manzt, line 89: Do we want some test for behavior of an invalid locale config?

manzt (approved):
Looks great. Just a minor comment about what we expect for invalid locale string (I'm guessing whatever react-aria does).

frontend/src/core/config/config-schema.ts
mscolnick: i thought about making it a typed string. but i would need to wrap a lot of the react-aria components (or maybe i can do module augmentation), will give it a shot in a followup

frontend/src/core/i18n/__tests__/locale-provider.test.tsx
mscolnick: im not running this through react-aria specifically (its mocked). i can try it out though

PR #6466, opened 2025-09-22
Add hover text on table header

nojaf:
## 📝 Summary

We would like to provide some (static) hover text on table header columns.

image

## 🔍 Description of Changes

I went with an approach similar to `text_justify_columns`.
The use-case if to further explain what the table does to other team members.

## 📋 Checklist

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -328,6 +328,7 @@ class table(
         text_justify_columns (Dict[str, Literal["left", "center", "right"]], optional):
             Dictionary of column names to text justification options: left, center, right.
         wrapped_columns (List[str], optional): List of column names to wrap.
+        header_hover_text (Dict[str, str], optional): Mapping from column names to header hover text.
         label (str, optional): Markdown label for the element. Defaults to "".
         on_change (Callable[[Union[List[JSONType], Dict[str, List[JSONType]], IntoDataFrame, List[TableCell]]], None], optional):
             Optional callback to run when this element's value changes.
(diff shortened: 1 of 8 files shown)


mscolnick:
@nojaf i wonder if maybe this should be more descriptive like: `header_info` or `header_descriptions`. this way we can use it more in just hover? i would also propose putting it inside a little `<InfoIcon/>` to signify it has some info (because you may not cover all headers).

@Light2Dark what are your thoughts?

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1413,3 +1417,19 @@ def _validate_column_formatting(
             raise ValueError(
                 f"Column '{next(iter(invalid))}' not found in table."
             )
+
+
+def _validate_header_hover_text(
```
Light2Dark: could we add a test for this

Light2Dark:
I like the InfoIcon idea too and use TooltipContent instead of title. The accurate name for this param would be header_tooltip I think, but do you think we would use this in diff ways @mscolnick .

some inspiration from ag-grid: https://www.ag-grid.com/javascript-data-grid/tooltips/#top

nojaf:
image

Light2Dark:
hey @nojaf , would you mind fixing the test and lint errors. I will help to merge and clean up after that.

nojaf:
@Light2Dark I don't quite understand why tests like `FAILED tests/_utils/test_uv_tree.py::test_complex_project_tree_raw_snapshot - AssertionError` are no longer working.

Light2Dark:
thanks @nojaf , I'll have a look but I think unrelated. thanks for the work!

Issue #6388, opened 2025-09-17
GitHub Copilot connection fails


codydunne:
### Describe the bug

Trying to enable GitHub Copilot fails when the directory path contains spaces. 

E.g., cloning a repo as `C:\fs\code\cargo bike turning` then running marimo with `uv run marimo edit .\cargo-bike-turn-radius.py` creates this error after everything has been set up:

<img width="520" height="160" alt="Image" src="https://github.com/user-attachments/assets/383afebc-c975-4c03-890c-3b761c0cc7fd" />

However, after cloning the repo as  `C:\fs\code\cargo-bike-turning` (hyphens instead of spaces) and running the same command leads to copilot working!

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.15.5",
  "editable": false,
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 165 Stepping 5, GenuineIntel",
  "Python Version": "3.13.4",
  "Binaries": {
    "Browser": "129.0.6668.90",
    "Node": "v24.2.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.5.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.2",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "nbformat": "5.10.4",
    "openai": "1.107.3",
    "loro": "1.6.0"
  },
  "Experimental Flags": {}
}
```

</details>

mscolnick:
Thanks for the heads up. This should hopefully be an easy fix (hoping i can repro outside windows)

dmadisetti:
Hmm, likely windows specific: 

<img width="679" height="136" alt="Image" src="https://github.com/user-attachments/assets/d0d40aae-cdb8-4c5a-9142-7b98fb9e22af" />

But digging in some more. thanks for reporting

mscolnick:
[...]

PR #6470, opened 2025-09-22
fix: string quoting for marimo

dmadisetti:
## 📝 Summary

Incorporates some of https://github.com/smoofra/mslex/blob/master/mslex (Apache 2 with attribution) for correct string quoting on windows to fix copilot

Closes: #6388


dmadisetti:
Thanks @manzt !

PR #6479, opened 2025-09-22
feat: Add markdown to the cell level

dmadisetti:
## 📝 Summary

Extracts markdown on compilation level allowing for native rendering

TODOs:

- [ ] runtime hook
- [ ] tests

### DIFF
--- a/marimo/_ast/compiler.py
+++ b/marimo/_ast/compiler.py
@@ -152,6 +152,54 @@ def fix_source_position(node: Any, source_position: SourcePosition) -> Any:
     return node
 
 
+def const_string(args: list[ast.stmt]) -> str:
+    (inner,) = args
+    if hasattr(inner, "values"):
+        (inner,) = inner.values
+    return f"{inner.value}"  # type: ignore[attr-defined]
+
+
+def const_or_id(args: ast.stmt) -> str:
+    if hasattr(args, "value"):
+        return f"{args.value}"  # type: ignore[attr-defined]
+    return f"{args.id}"  # type: ignore[attr-defined]
+
+
+def extract_markdown(code: str) -> Optional[str]:
+    markdown_lines = [
+        line for line in code.strip().split("\n") if line.startswith("mo.md(")
+    ]
+    if len(markdown_lines) > 1:
+        return None
+
+    code = code.strip()
+    # Attribute Error handled by the outer try/except block.
+    # Wish there was a more compact to ignore ignore[attr-defined] for all.
+    try:
+        (body,) = ast.parse(code).body
+        if body.value.func.attr == "md":  # type: ignore[attr-defined]
+            value = body.value  # type: ignore[attr-defined]
+        else:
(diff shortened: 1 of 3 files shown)


marimo/_ast/compiler.py
```diff
@@ -152,6 +152,54 @@ def fix_source_position(node: Any, source_position: SourcePosition) -> Any:
...
+    if hasattr(args, "value"):
+        return f"{args.value}"  # type: ignore[attr-defined]
+    return f"{args.id}"  # type: ignore[attr-defined]
+
+
+def extract_markdown(code: str) -> Optional[str]:
+    markdown_lines = [
+        line for line in code.strip().split("\n") if line.startswith("mo.md(")
+    ]
+    if len(markdown_lines) > 1:
```
mscolnick: this could be slightly cheaper `some(line.startswith("mo.md(") for line in code.strip().split("\n")`

dmadisetti: Nice-  we don't actually need this check at all, ast parse handles this case.

mscolnick: but isn't `ast.parse` more expensive. should we have a cheap check first and then actual parsing?

mscolnick: the cheap one could even be `"mo.md(" in code`

dmadisetti: We have the have the ast already- so in those cases just directly use that, while allowing for the early stopping if extracted directly from code.

marimo/_runtime/runtime.py
```diff
@@ -2009,22 +2009,95 @@ async def instantiate(self, request: CreationRequest) -> None:
...
+
+        # Handle as default if no cells export 'mo'
+        if not exports_mo:
+            return
+
+        # Remove markdown cells from uninstantiated requests
+        for cell_id, content in markdown_cells.items():
+            # Since markdown cell, render and broadcast output
+            # Remove cell from outstanding requests
+            from marimo._output.md import md
```
mscolnick: maybe worth moving this outside the for loop

marimo/_runtime/runtime.py
```diff
@@ -2009,22 +2009,95 @@ async def instantiate(self, request: CreationRequest) -> None:
...
+                channel=CellChannel.OUTPUT,
+                mimetype=mimetype,
+                data=html_content,
+                cell_id=cell_id,
+                status="idle",
+            )
+
+            # Mark the cell as not stale (already "run")
+            CellOp.broadcast_stale(cell_id=cell_id, stale=False)
+            del execution_requests[cell_id]
```
mscolnick, line 2100: any way to break some of this up outside of the `Runner`? its gotten quite large. i know it doesn't really belong in hooks, but maybe a function in the same file

dmadisetti: This is a good point, but we still call some functions on the object

PR #6486, opened 2025-09-23
Support sticky headers on table 

nojaf:
## 📝 Summary

Sorry if this is more of an feature request than an actual PR.
We have long and wide tables in our organization and wish to improve scroll through this vertically. 
(We are using `pagination=False` to see all rows)

`freeze_columns` works well nicely for horizontal scrolling.
But when a table is long:
- The table takes the entire page, and the scrollbar is on the page level.
- The header is gone as soon as your scroll.

## 🔍 Description of Changes

It would be interesting if we could have something long these lines:

Kapture 2025-09-23 at 11 32 25

- Scrollbar is on the table
- Header is sticky

(table header is red for dramatic effect)

Could we have something that brings this experience?
If so, I'd be happy to work on this in whatever solution you deem best.

//cc @mscolnick @Light2Dark 

## 📋 Checklist

### DIFF
--- a/tests/_plugins/ui/_impl/test_table.py
+++ b/tests/_plugins/ui/_impl/test_table.py
@@ -1928,7 +1928,7 @@ def test_table_max_height():
 
     table_no_height = ui.table([1, 2, 3])
     # Keys may be absent when not configured
-    assert "max-height" not in table_no_height._component_args
+    assert table_no_height._component_args["max-height"] is None
 
 
 def test_calculate_top_k_rows():


Light2Dark:
I like the idea! Do you also want to pass in python params to control this? Eg. max_height and sticky_header

this is kinda interesting from [plotly](https://dash.plotly.com/datatable/height#setting-table-height-with-vertical-scroll):
```python
app.layout = dash_table.DataTable(
    data=df.to_dict('records'),
    columns=[{'id': c, 'name': c} for c in df.columns],
    style_table={'height': '300px', 'overflowY': 'auto'}  # <---
)

**Comment by nojaf** [comment]
Hi @Light2Dark, yes, perhaps some control makes sense.
However, a single style object won't fit the bill here.
It needs a change to the `thead` and wrapper `div` so I wonder how to expose this to the end-user. Expecting them to pass the right css objects puts a high bar at understanding how this internally works. So, I'm not quite sure about this one.

**Comment by mscolnick** [comment]
my preference would be to avoid css styles as the args to get this and it is a flag that handles it (e.g. `scrollable` or just `max_height` and its always scrollable)

**Comment by nojaf** [comment]
So, `max_height` as int? And do we derive `sticky` from that value or would you have another setting to control that?
Does `max_height` mean `pagination=False` or should that still be set by the user?

**Comment by mscolnick** [comment]
> So, max_height as int? 

Yep i think an `Optional[int]`

> And do we derive sticky from that value or would you have another setting to control that?

Yea i think if max-height is set, create a variable `hasMaxHeight` and then make it sticky

> Does max_height mean pagination=False or should that still be set by the user?

We can leave these separate and have the user config. give the flexibility of pagination 100 but still a max-height

File: frontend/src/components/ui/table.tsx [hunk 1]
```diff
@@ -3,30 +3,44 @@ import * as React from "react";
...
-  HTMLTableElement,
-  React.HTMLAttributes<HTMLTableElement>
->(({ className, ...props }, ref) => (
-  <div className="w-full overflow-auto flex-1">
-    <table
+interface TableProps extends React.HTMLAttributes<HTMLTableElement> {
+  maxHeight?: number;
+}
+
+const Table = React.forwardRef<HTMLTableElement, TableProps>(
```
Light2Dark: Is it possible to apply these styles in data-table rather than the base components? we can keep these clean

same for table-header

marimo/_plugins/ui/_impl/table.py
```diff
@@ -335,6 +339,9 @@ def style_cell(_rowId, _columnName, value):
         hover_template (str, optional): A template for the hover text of the table.
         max_columns (int, optional): Maximum number of columns to display. Defaults to the
             configured default_table_max_columns (50 by default). Set to None to show all columns.
+        max_height (int, optional): Maximum height of the table body in pixels. When set,
+            the table becomes vertically scrollable and the header may be made sticky
```
Light2Dark:
```suggestion
            the table becomes vertically scrollable and the header will be made sticky
```

frontend/src/components/data-table/data-table.tsx
```diff
@@ -47,6 +47,8 @@ import { getStableRowId } from "./utils";
 interface DataTableProps<TData> extends Partial<DownloadActionProps> {
   wrapperClassName?: string;
   className?: string;
+  maxHeight?: number;
+  isSticky?: boolean;
```
Light2Dark: maybe we don't need isSticky, since it's derived from maxHeight.

mscolnick (approved):

PR #6502, opened 2025-09-24
improvement: add completion info to SQL table completions

mscolnick:
This adds a completion info (unstyled for now) to the table completion.

Also
- utilities for building the `SQLNamespace`
- include default DB in name (not just the tables)

Screenshot 2025-09-25 at 11 20 42 AM

Screenshot 2025-09-25 at 11 19 48 AM
Screenshot 2025-09-25 at 11 19 04 AM
Screenshot 2025-09-25 at 11 19 00 AM

### DIFF
--- a/frontend/src/core/codemirror/language/__tests__/sql.test.ts
+++ b/frontend/src/core/codemirror/language/__tests__/sql.test.ts
@@ -789,7 +789,75 @@ describe("tablesCompletionSource", () => {
             "detail": "public",
             "info": [Function],
             "label": "public",
-            "type": "namespace",
+            "type": "schema",
+          },
+        },
+        "test_db": {
+          "children": {
+            "public": {
+              "children": {
+                "orders": {
+                  "children": [
+                    {
+                      "info": [Function],
+                      "label": "order_id",
+                      "type": "column",
+                    },
+                    {
+                      "info": [Function],
+                      "label": "user_id",
+                      "type": "column",
+                    },
+                    {
+                      "info": [Function],
+                      "label": "total",
+                      "type": "column",
+                    },
+                  ],
+                  "self": {
+                    "info": [Function],
+                    "label": "orders",
+                    "type": "table",
+                  },
+                },
+                "users": {
+                  "children": [
+                    {
+                      "info": [Function],
+                      "label": "id",
+                      "type": "column",
+                    },
+                    {
+                      "info": [Function],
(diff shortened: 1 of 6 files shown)


frontend/src/core/codemirror/language/__tests__/sql.test.ts
```diff
@@ -735,16 +735,130 @@ describe("tablesCompletionSource", () => {
...
-            "order_id",
-            "user_id",
-            "total",
-          ],
-          "users": [
-            "id",
-            "name",
-            "email",
-          ],
+          "children": {
```
mscolnick, line 738: these snapshots changed for two reason:

1. instead of just labels, it now includes a custom Completion with a `info` render
2. We also include default databases at the top (previously did not). that can be see in this commit: https://github.com/marimo-team/marimo/pull/6502/commits/780202d8c95d24448693121ee85385d1557abea3#diff-f79af3635ef377d486695e4101902aad0982c208b769b3851c9086e694f722f3

frontend/src/core/codemirror/language/languages/sql/completion-store.ts
```diff
@@ -33,81 +42,74 @@ class SQLCompletionStore {
...
+
+        for (const table of schema.tables) {
+          builder.addTable([schema.name], table);
+        }
       }
     }
 
     // Otherwise, we need to use the fully qualified name
-    for (const database of connection.databases) {
-      if (database.name === defaultDb?.name) {
```
mscolnick, line 92: this is the only functional change. we now include the default db which can help with autocomplete or if users want fully qualified names

Light2Dark:
some style edits, I think it's great overall!

CleanShot 2025-09-26 at 04 51 55

CleanShot 2025-09-26 at 04 55 58

I looked at overriding the tooltipRenderer to work with this, but may need some change to codemirror-sql, maybe make it async for something like this
```
            tooltipRenderers: {
              table: (data: NamespaceTooltipData) => {
                return await data.item.completion?.info().dom.outerHTML
              },
            }
```

Light2Dark (approved):

mscolnick:
> I looked at overriding the tooltipRenderer to work with this, but may need some change to codemirror-sql, maybe make it async for something like this

Yep, i already made these changes to codemirror-sql, so we can use them when we upgrade

Issue #4196, opened 2025-03-21
Can't enable autosave because of "Overridden by pyproject.toml [off]" even though there is no pyproject.toml


dentroai:
### Describe the bug

![Image](https://github.com/user-attachments/assets/37d7afc6-77b1-4a9e-bb1a-f4c1841d6831)

So I want to have autosave enabled. 
But it doesn't work. 
It tells me that a pyproject.toml which doesn't exist is overriding the config.
I also tried to create a pyproject.toml where I enable autosave in the folder where I run `marimo edit --watch`, as well as a .marimo.toml. Both doesn't help. 

`marimo config show` shows me the following:
```
  marimo config show 

🏠 User config from /home/itsmemario/.config/marimo/marimo.toml

[runtime]
std_stream_max_bytes = 1000000
output_max_bytes = 8000000
auto_instantiate = false
auto_reload = "off"
watcher_on_save = "lazy"
pythonpath = []
on_cell_change = "autorun"

[completion]
activate_on_typing = true
copilot = false

[formatting]
line_length = 100

[keymap]
preset = "default"

[keymap.overrides]

[experimental]
inline_ai_tooltip = true

[package_management]
manager = "uv"

[language_servers.pylsp]
enable_pyflakes = false
enable_ruff = true
enable_pydocstyle = false
enable_flake8 = false
enable_mypy = true
enabled = true
enable_pylint = false

[save]
autosave_delay = 1000
autosave = "after_delay"
format_on_save = false

[ai]
rules = ""

[ai.anthropic]
api_key = "********"

[ai.open_ai]
base_url = ""
model = "claude-3-7-sonnet-20250219"

[ai.google]
api_key = "********"

[server]
browser = "default"
follow_symlink = false

[display]
dataframes = "rich"
cell_output = "above"
default_width = "full"
theme = "dark"
code_editor_font_size = 14

[snippets]
include_default_snippets = true
custom_paths = []
```

I checked the folder `/home/itsmemario/.config/marimo` but there are no other files.

As a caveat, I use marimo in a subfolder of a python project. I.e. the virtualenv is part of the parent folder using uv and a requirements.txt in case that influences the situation which I doubt.

How can I enable autosave?

Marimo versions: 0.11.22 and 0.11.25 . 
Linux
python 3.12

### Environment

<details>

```
Replace this line with the output of marimo env. Leave the backticks in place.
```

</details>

dentroai:
[...]

Issue #4194, opened 2025-03-21
Can't trace config file when overwritten by pyproject.toml


KennethEnevoldsen:
So I am fairly new to Marimo so this might be an error on my end.

I have the following issue: I want autosave turned on but get a warning stating that it is overwritten:

<img width="452" alt="Image" src="https://github.com/user-attachments/assets/77c472b4-4383-4c8b-a45e-569bfe8d3163" />

First, I tried just deleting `pyproject.toml` file in the git repo. This did not change anything. 
Then I tried adding the following to the `pyproject.toml`

```toml
...

[marimo.tool.save]
autosave_delay = 1000
format_on_save = true
autosave = "after_delay"
```

this did not change the problem either.

Then I went out a folder and create a new notebook (assuming that it might look further down my directory):

```
cd ..
mkdir tmp
cd tmp
marimo new
```

Here, I did not have the issue.

This leaves me with a problem and a 'confusion': 

- problem: I can't figure out where the pyproject.toml is located. 
- confusion: Having newer specified marimo in a pyproject.toml before I do not understand why it would overwrite my settings

If I run `marimo config show` then I get:

```toml
...

[save]
autosave_delay = 1000
format_on_save = true
autosave = "after_delay"

...
```

A simple solution might be to allow me to click the pyproject.toml to go to the file or see the path.

mscolnick:
If you do `marimo config show`, it should show you the pyproject.toml it is pulling in. Can you paste your full output (and hide any sensitive info)?


Also your configuration in pyproject.toml should be: `[tool.marimo.save]`, not `[marimo.tool.save]`, following the [best practices for pyproject.toml config](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-your-pyproject-toml)

mscolnick:
@KennethEnevoldsen , i might have figured it out. Is it because you are running with `--watch`?

This wording is incorrect - it's not overridden by `pyproject.toml`, but it is overridden because `auto-saving` is disabled while in watch mode. (otherwise it will get conflicts with your external editor)

bulletmark:
I just had this problem. Spent a while then came to raise a bug and found this bug and #4196 which told me it was due to me running with `--watch`. I have been doing that lazily every time just in case I want to directly edit but seems there  is a disadvantage to doing that.

KennethEnevoldsen:
I am running it using the VS code extension which indeed does run it using `--watch`:

```
[...]

PR #6127, opened 2025-08-22
refactor: adjust cell matching for --watch

dmadisetti:
## 📝 Summary

The following assert in watch is occasionally hit:

https://github.com/marimo-team/marimo/blob/fcc3a6e1c8730b706194e146aaf5769bc82477d4/marimo/_ast/cell_manager.py#L484

This code provides a refactor that should make an inconsistent return result impossible, but also cleans up some of the logic and implements the hungarian algorithm for a more robust assignment problem solution.

For reference the new tests fail as follows with the previous code:

```bash
FAILED tests/_ast/test_cell_manager.py::TestCellMatching::test_similar_but_not_exact_matches_with_dupes - AssertionError: assert ['x', 'a', 'b', 'unused_a'] == ['c', 'a', 'b', 'x']
FAILED tests/_ast/test_cell_manager.py::TestCellMatching::test_left_inexact_matches_with_dupes - AssertionError: assert ['c', 'a', 'b', 'unused_a'] == ['c', 'a', 'b', 'x']
FAILED tests/_ast/test_cell_manager.py::TestCellMatching::test_right_inexact_matches_with_dupes - AssertionError: assert ['x', 'a', 'b'] == ['c', 'a', 'b']
FAILED tests/_ast/test_cell_manager.py::TestCellMatching::test_outer_inexact_matches - AssertionError: assert ['x', 'unused_a', 'b', 'a'] == ['c', 'x', 'b', 'a']
FAILED tests/_ast/test_cell_manager.py::TestCellMatching::test_outer_inexact_matches_with_dupes - AssertionError: assert ['z', 'y', 'x...nused_b', 'a'] == ['c', 'a', 'b', 'x', 'z', 'y']
FAILED tests/_ast/test_cell_manager.py::TestCellMatchingEdgeCases::test_multiple_identical_codes_in_next - AssertionError: assert ['a', 'x', 'b'] == ['a', 'y', 'b']
FAILED tests/_ast/test_cell_manager.py::TestCellMatchingEdgeCases::test_mixed_case_sensitivity - AssertionError: assert ['b', 'unused_x'] == ['a', 'b']
FAILED tests/_ast/test_cell_manager.py::TestCellMatchingEdgeCases::test_completely_different_codes_edge_case - AssertionError: assert ['b', 'unused_x'] == ['a', 'b']
[...]

PR #6535, opened 2025-09-26
don't override auto-save by when in --watch mode

### LINKED ISSUE #6534: Stop disabling autosave with --watch
### Description

Marimo blocks autosave when running with `--watch`, see

https://github.com/marimo-team/marimo/issues/4194
https://github.com/marimo-team/marimo/issues/4196
https://github.com/marimo-team/marimo/issues/6167

For our UC it would be helpful to have the option to enable both watch and autosave.

We use watch to detect external changes from git operations like branching. For changes coming from marimo like autosave, git will pick that up as an unstaged diff.

You could argue that it's on the other application (editor or otherwise) to handle potentially conflicting writes coming from your autosave functionality. Most editors are pretty good about warning when a file has changed externally.

Seems like something that should be possible, with the appropriate warnings and caveats.

### Suggested solution

Could be as easy as removing the config override here and turning the info into a warning.

https://github.com/marimo-team/marimo/blob/64225b3b4e88950315d7d37ccdf52e4973e3bb2a/marimo/_server/start.py#L199-L211

The GUI and the docs should probably warn that running watch and autosave together can cause problems.

- [x] Yes

### Alternatives

Is it possible to yoke the watch and autosave hooks so autosave only runs after a watch update?

### Additional context

_No response_

mscolnick:
Fixes #6534

Closes https://github.com/marimo-team/marimo/issues/4194
Closes https://github.com/marimo-team/marimo/issues/4196

This updates the behavior of watch mode in the server configuration to ensure that user-defined auto-save is kept, rather than being forcibly overridden. We had this originally enabled to avoid conflicts between the server's view of the file and external changes made from the filesystem. Since the cell matching has improved (https://github.com/marimo-team/marimo/pull/6127), I don't expect this to cause issues.

We can in a followup handle overwrites and show the user (similar to how vscode or other IDEs might handle this)

### DIFF
--- a/marimo/_server/start.py
+++ b/marimo/_server/start.py
@@ -197,7 +197,7 @@ def start(
         )
 
     if watch and config_reader.is_auto_save_enabled:
-        LOGGER.warning("Watch mode enabled and may interfere with auto-save.")
+        LOGGER.warning("Enabling watch mode may interfere with auto-save.")
 
     if GLOBAL_SETTINGS.MANAGE_SCRIPT_METADATA:
         config_reader = config_reader.with_overrides(
--- a/marimo/_server/start.py
+++ b/marimo/_server/start.py
@@ -196,19 +196,8 @@ def start(
             min_port=DEFAULT_PORT + 400,
         )
 
-    # If watch is true, disable auto-save and format-on-save,
-    # watch is enabled when they are editing in another editor
-    if watch:
-        config_reader = config_reader.with_overrides(
-            {
-                "save": {
-                    "autosave": "off",
-                    "format_on_save": False,
-                    "autosave_delay": 1000,
-                }
-            }
-        )
-        LOGGER.info("Watch mode enabled, auto-save is disabled")
+    if watch and config_reader.is_auto_save_enabled:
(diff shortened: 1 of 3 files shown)


marimo/_server/start.py
```diff
@@ -196,19 +196,8 @@ def start(
...
-                "save": {
-                    "autosave": "off",
-                    "format_on_save": False,
-                    "autosave_delay": 1000,
-                }
-            }
-        )
-        LOGGER.info("Watch mode enabled, auto-save is disabled")
+    if watch and config_reader.is_auto_save_enabled:
+        LOGGER.warning("Watch mode enabled and may interfere with auto-save.")
```
manzt:
```suggestion
        LOGGER.warning("Enabling watch mode may interfere with auto-save.")
```

If the future, is the idea to define the behavior of how these two options interplay and surface a useful alert?

manzt (dismissed):
Just a small comment

marimo/_server/start.py
mscolnick: i don't think thats needed

PR #6537, opened 2025-09-26
improvement: google vertex support, make google key optional

mscolnick:
This makes the Google API Key optional so you can use hit Google Vertex using environment variabes and application default credentials (ADC)


manzt (approved):
This looks great

PR #6538, opened 2025-09-26
fix: add more expressions for SQL refs

mscolnick:


tests/_ast/test_visitor.py
```diff
@@ -1435,7 +1435,7 @@ def test_sql_table_f_string() -> None:
             "drop table and reference it in same statement",
             "drop table schema.cars; select * from cars",
             set(),
-            {"mo", "cars"},
+            {"mo", "cars", "schema.cars"},
```
mscolnick, line 1438: @dmadisetti , @Light2Dark , does this change look right?

Light2Dark: i think it looks right.

Light2Dark (approved):

tests/_ast/test_visitor.py
Light2Dark: I see a lot more failing tests, will check.
edit: oh, only fails on minimal deps, maybe old duckdb Tokenizer.

mscolnick:
```
 module 'sqlglot.expressions' has no attribute 'Analyze'
```

im going to see if i can bump the minimum

PR #6539, opened 2025-09-26
fix: handle by mro types from upstream libs

mscolnick:


dmadisetti (approved):
What there an example of this beyond the test one?

PR #6548, opened 2025-09-26
fix: switch to depth-first traversal for inOrderIds method

mscolnick:
- Updated the inOrderIds method to use depth-first traversal instead of breadth-first, preserving logical document order.

### DIFF
--- a/marimo/_ast/load.py
+++ b/marimo/_ast/load.py
@@ -187,6 +187,13 @@ def get_notebook_status(filename: str) -> LoadResult:
     return LoadResult(status="valid", notebook=notebook, contents=contents)
 
 
+FAILED_LOAD_NOTEBOOK_MESSAGE = (
+    "Static loading of notebook failed; falling back to dynamic loading. "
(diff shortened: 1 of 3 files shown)


marimo/_ast/load.py
```diff
@@ -225,10 +232,5 @@ def load_app(filename: Optional[str]) -> Optional[App]:
         # Security advantages of static load are lost here, but reasonable
         # fallback for now.
         _app = _dynamic_load(filename)
-        LOGGER.warning(
-            "Static loading of notebook failed; "
-            "falling back to dynamic loading. "
-            "If you can, please report this issue to the marimo team — "
-            "https://github.com/marimo-team/marimo/issues/new?template=bug_report.yaml"
-        )
+        LOGGER.warning(FAILED_LOAD_NOTEBOOK_MESSAGE)
```
dmadisetti, line 235: Did this get triggered? We haven't had reports yet, and was considering removing it

dmadisetti (approved):
Nice

Issue #6513, opened 2025-09-24
Parent App does not know when embedded notebook is loading


aschreiber1:
### Describe the bug

When you embed a notebook into another notebook, when the embedded notebook is loading, the parent page does not get marked as loading and instead just hangs there

### Will you submit a PR?

- [ ] Yes

### Environment

NA

mscolnick:
@aschreiber1 do you have minimal code you can share? when you say it just hangs there, does it hang forever? or just doesn't showing the loading state?

aschreiber1:
Let me see if I can find some small code. But it doesn't hang forever, just normally in marimo when cells are running the page has a spinning thing in the top left corner that tells the user its doing work. That doesn't seem to translate to embedded notebooks. So if the embedded notebook does a 10s calculation, the parent page just sits there and doesn't give the user feedback on whats happening. You can still click around and such which would confuse users quite a lot

aschreiber1:
Here is an example. Notebook 1:
```python
import marimo

__generated_with = "0.16.1"
app = marimo.App(width="medium")


@app.cell
def _():
    import marimo as mo
    import time
    return mo, time


@app.cell
def _(mo):
    mo.md("""# Child Notebook - Slow Loading Component""")
    return


@app.cell
def _(mo):
    # Create a run button
    run_button = mo.ui.run_button()
    return (run_button,)


@app.cell
def _(mo, run_button, time):
    # Only run slow computation when button is clicked
    if run_button.value:
        time.sleep(10)  # 10 second delay
        out = mo.md("test")
    else:
        out = mo.md("not here")
    out
    return


@app.cell
def _(run_button):
    run_button
    return


if __name__ == "__main__":
    app.run()
```

notebook 2
```python
import marimo

__generated_with = "0.16.1"
app = marimo.App(width="medium")


@app.cell
def _():
    import marimo as mo
    return (mo,)


@app.cell
def _(mo):
    mo.md("""# Parent Notebook - Embedding Example""")
    return


@app.cell
def _(mo):
    mo.md("""## Loading child notebook (should show loading state)...""")
    return


@app.cell
def _():
    # Import the child notebook
    from child_notebook_slow import app as child_app
    return (child_app,)


@app.cell
async def _(child_app):
    # Execute the child notebook - this should show loading state
    result = await child_app.embed()
    return (result,)


@app.cell
def _(mo, result):
    # Display the child notebook's output
[...]

PR #6551, opened 2025-09-26
fix: issue runtime loading indication on button run

dmadisetti:
## 📝 Summary

fixes: #6513

Issue a loading state when UI components trigger server side work

### DIFF
--- a/frontend/src/core/kernel/RuntimeState.ts
+++ b/frontend/src/core/kernel/RuntimeState.ts
@@ -1,6 +1,8 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
 import { Logger } from "@/utils/Logger";
+import { uiElementLoadingAtom } from "../cells/cells";
+import type { UIElementId } from "../cells/ids";
 import {
   MarimoValueReadyEvent,
   type MarimoValueReadyEventType,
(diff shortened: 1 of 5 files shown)


frontend/src/core/kernel/RuntimeState.ts
```diff
@@ -81,17 +84,37 @@ export class RuntimeState {
...
+        .catch(
+          // This happens if the run failed to register (401, 403, network
+          // error, etc.) A run may fail if the kernel is restarted or the
+          // notebook is closed.
+          (error) => {
+            Logger.warn(error);
+            // Remove from loading state on error too
+            this.clearLoadingState(objectId as UIElementId);
+          },
+        );
```
mscolnick, line 110: nit: could do `finally`

frontend/src/core/kernel/RuntimeState.ts
```diff
@@ -81,17 +84,37 @@ export class RuntimeState {
...
         values: [value],
-      }).catch(
-        // This happens if the run failed to register (401, 403, network
-        // error, etc.) A run may fail if the kernel is restarted or the
-        // notebook is closed.
-        (error) => {
-          Logger.warn(error);
-        },
-      );
+      })
```
mscolnick, line 96: does this take a while? won't this just put the request on the queue and then finish? or it is blcoking for the queue to be free'd up?

dmadisetti: Currently doesn't work, just pushed up my iteration. Will come back to this after getting other prs in

PR #6556, opened 2025-09-27
Click filenames in stack trace to open locally in editor

### LINKED ISSUE #6390: Click filenames in stack trace to open locally in editor
### Description

When my Marimo cells raise exceptions and I see a stack-trace, it often involves code in my own modules:

```
  File "FILE", line LINE, in FUNC
```

It would be nice to go to the erroring line with a mouse-click on the trackback

### Suggested solution

The traceback filenames should become links/buttons that invoke the marimo process to invoke the user's editor of choice according to the `EDITOR` environment variable to the erroring file and line.

- [x] Yes

### Alternatives

An alternative suitable for the web would be for marimo to also include its own module file editor (like Jupyter does) and open the erroring module in it, but most users would probably prefer opening the source in their editor of choice.

### Additional context

_No response_

yairchu:
## 📝 Summary

Fixes #6390

## 🔍 Description of Changes

* When parsing tracebacks, in addition to marimo cells replacing the traceback with a link and button, now other files are also wrapped with a clickable div element, which invokes the user's `EDITOR` to open the file
* `marimo._server.files.OSFileSystem.open_in_editor` now supports opening a file in a specific line, which the traceback uses to open the relevant line from the trace

## 📋 Checklist

### On: I have added tests for the changes made

I'm unsure how to test this other than manually verifying it. I tried to see if there's a similar existing test for the existing functionality of cell links, so I could mimic that, but I didn't find such on

## Btw (devops)

I tried to follow the instructions according to `CONTRIBUTING.md`, but:

* I had to uninstall an older local version of node for things to work, otherwise node resolved to the oldest version I have which was too old?
* Couldn't get `pnpm` to work so had to use `make fe` for an >1min dev feedback cycle. (btw I'm a front-end noob so I might be missing basic things)


marimo/_server/models/files.py
```diff
@@ -32,6 +32,7 @@ class FileDetailsRequest(msgspec.Struct, rename="camel"):
 class FileOpenRequest(msgspec.Struct, rename="camel"):
     # The path of the file to open
     path: str
+    line_number: Optional[int] = None
```
Light2Dark, line 35: you will need to run `make fe-codegen` in root to update the API spec.

marimo/_server/files/os_file_system.py
```diff
@@ -328,9 +328,18 @@ def open_in_editor(self, path: str) -> bool:
             # otherwise it silently opens the terminal in the same window that is
             # running marimo.
             if editor and not _is_terminal_editor(editor):
+                args = [path]
```
Light2Dark: I think we can add tests for the arguments generation, maybe by moving it out to a separate function.

yairchu:

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -211,6 +211,21 @@ export const replaceTracebackFilenames = (domNode: DOMNode) => {
       </span>
     );
   }
+  if (info) {
```
Light2Dark: Could you share a screenshot for this. I'm wondering if this can be merged with the above rendering.

yairchu: Here's what it looks like:

Screenshot 2025-09-27 at 16 40 51

yairchu:
The previously failing CI tests pointed me to the additional tests I needed to maintain so can now say the changes have tests :)

Btw, `make test` failed for me in some tests but that also happens for me in `main` (so I don't think it's related to my changes)

frontend/src/utils/traceback.ts
```diff
@@ -29,7 +28,8 @@ export const elementContainsMarimoCellFile = (domNode: Element) => {
 };
 
 export interface TracebackInfo {
-  cellId: CellId;
+  src: string;
+  isCell: boolean;
   lineNumber: number;
 }
```
Light2Dark: The cellId is used in other parts of the codebase, we should keep it. Maybe this instead
```typescript
export interface TracebackInfo {
  filePath?: string;
  cellId?: CellId;
  lineNumber: number;
}
```

frontend/src/utils/traceback.ts
```diff
@@ -68,15 +68,16 @@ export function getTracebackInfo(domNode: DOMNode): TracebackInfo | null {
...
-        const cellId = /__marimo__cell_(\w+)_/.exec(
-          domNode.firstChild.nodeValue,
-        )?.[1];
         const lineNumber = Number.parseInt(
           lineSibling.firstChild.nodeValue || "0",
           10,
         );
-        if (cellId && lineNumber) {
-          return { cellId: cellId as CellId, lineNumber };
+        const isCell = domNode.firstChild.nodeValue?.includes("__marimo__");
```
Light2Dark: Return based on whether it's a file / cell
```ts
        let cellId: CellId | undefined;
        if (isCell) {
          cellId = /__marimo__cell_(\w+)_/.exec(
            domNode.firstChild.nodeValue,
          )?.[1] as CellId;
          if (!cellId) {
            return null;
          }
          return { cellId, lineNumber };
        }

        const filePath = /"(.+?)"/.exec(domNode.firstChild.nodeValue)?.[1];
        if (filePath && lineNumber) {
          return { filePath, lineNumber };
        }
```

Light2Dark: Also, if you could update the docstring for this function, I got this from chatgpt:
```ts
  // Example transformation:
  //
  //   File <span class="nb">"/tmp/marimo_<number>/__marimo__cell_<cell-id>.py"</span>
  //   , line <span class="n">1</span>...
  //
  //   becomes
  //
  //   { cellId: <CellID>, lineNumber: 1 }
  //
  //   or for files:
  //
  //   File <span class="nb">"/path/to/file.py"</span>
  //   , line <span class="n">42</span>...
  //
  //   becomes
  //
  //   { filePath: "/path/to/file.py", lineNumber: 42 }
```

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
```diff
@@ -211,6 +211,21 @@ export const replaceTracebackFilenames = (domNode: DOMNode) => {
...
+    return (
+      <div
+        className="inline-block cursor-pointer text-link hover:underline"
```
Light2Dark:
```suggestion
        className="inline-block cursor-pointer text-destructive hover:underline"
```

Light2Dark (commented):
hi @yairchu , I think it looks good overall.

By tests, I meant some Python ones to test the [args] generation.
Feel free to implement what you can, we can also help to add stuff and merge it in.

frontend/src/utils/traceback.ts
yairchu: Can do that. Am a type-script newbie but ideally we'd have a variant type. The `isCell` version mimics that more closely in that it doesn't allow to have both or neither options which the optional fields option allows.
But if you still prefer the optional fields version I'd implement that.

Light2Dark: That makes sense, sorry I was confused.
something like this sounds right then
```ts
export type TracebackInfo =
  | {
      kind: "file";
      filePath: string;
      lineNumber: number;
    }
  | {
      kind: "cell";
      cellId: CellId;
      lineNumber: number;
    };
```

yairchu: And are you sure that TracebackInfo.cellId was used in other parts of the code? If so I couldn't find them. Edit: found a place I missed

frontend/src/utils/traceback.ts
yairchu: Added the comment (for now with the `isCell` option)

frontend/src/components/editor/output/MarimoTracebackOutput.tsx
yairchu: Done

frontend/src/utils/traceback.ts
yairchu: Done

marimo/_server/files/os_file_system.py
yairchu: Extracted it to a separate function. Ideally it could even be in a separate package so various projects which open files in editors would share this logic. I took this snippet from my `srcview` package which iirc took it from `git-mediate`.
The main thing to test is that the editors really expect arguments this way but testing for that seems difficult.

yairchu:
I fixed the previous CI errors and I think this time this should be good!
Does the fact that it's at "Expected — Waiting for status to be reported" for several hours mean that this time it's not failing quick and it's just the time it takes the CI to run?

Light2Dark:
there is an error with the precommit, I created another PR for that. Thanks for the contribution!

yairchu:
Looks like this time the 3 failing CIs aren't this PR's fault:

* pre-commit already faulty as you mentioned
* cli test / Tests CLI on windows-latest 3.12: Failed in a timeout that seems totally unrelaters
* check-schemas (openapi, packages/openapi/api.yaml, OpenAPI specification, true): `Error: failed to load base spec from "main/packages/openapi/api.yaml"` - the yaml was auto-generated by `make fe-codegen`?

mscolnick:
@yairchu this looks great! the failing tests are likely bc you are on a fork and our GH workflows likely don't handle that well, they are just red-herring, but the main tests do pass

PR #6574, opened 2025-09-29
improvement: cache validation requests for sql linter

mscolnick:
Cache SQL validation requests that are shared from the linter and banner


frontend/src/core/codemirror/language/languages/sql/sql.ts
```diff
@@ -687,6 +687,7 @@ function sqlValidationExtension(): Extension {
       try {
         const result = await ValidateSQL.request({
           onlyParse: false,
```
mscolnick, line 689: @Light2Dark since `onlyParse` is true in one case but false in the other, the cache is not that effective. maybe they can both be `true`? or both `false` (for just duckdb)

frontend/src/__tests__/lru.test.ts
```diff
@@ -1,74 +0,0 @@
-/* Copyright 2024 Marimo. All rights reserved. */
```
mscolnick, line 1: duplicate file

frontend/src/core/codemirror/language/languages/sql/sql.ts
Light2Dark: yeah, both can be false if in validate sql mode for duckdb.

PR #6581, opened 2025-09-29
Windows Short 8.3 Path Names with kernel32.GetShortPathNameW

mscolnick:
Cherry-pick 6564 with fixed type-checking,


dmadisetti:
I think https://github.com/marimo-team/marimo/pull/6569 is a better solution since it removes the awkward quoting all together

kimmolinna:
@dmadisetti and @mscolnick  - #6569 is a far better solution. Your solution is working nice with the VS Code Extension too. Thanks.

PR #6608, opened 2025-09-30
improvement: use lazy-frames for multi-step processing, include more tests

mscolnick:
After moving to narwhals v2, we can support more dataframe types in certain codepaths and tests.
This also makes multi-step operations lazy to avoid incremental df states

### DIFF
--- a/marimo/_plugins/ui/_impl/charts/altair_transformer.py
+++ b/marimo/_plugins/ui/_impl/charts/altair_transformer.py
@@ -15,7 +15,7 @@ from marimo._plugins.ui._impl.tables.utils import (
     get_table_manager_or_none,
 )
 from marimo._utils.data_uri import build_data_url
-from marimo._utils.narwhals_utils import can_narwhalify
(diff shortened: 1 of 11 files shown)


marimo/_plugins/ui/_impl/charts/altair_transformer.py
```diff
@@ -154,7 +154,16 @@ def sanitize_nan_infs(data: Any) -> Any:
     """Sanitize NaN and Inf values in Dataframes for JSON serialization."""
     if can_narwhalify(data):
         narwhals_data = nw.from_native(data)
-        for col, dtype in narwhals_data.schema.items():
+        is_prev_lazy = isinstance(narwhals_data, nw.LazyFrame)
```
manzt, line 157: should we use the `is_narwhals_lazyframe` helper here too?

manzt (approved):

marimo/_plugins/ui/_impl/charts/altair_transformer.py
mscolnick: yep good call. just updated in a followup PR

PR #6612, opened 2025-09-30
improvement: cleanup ibis code to share from narwhals table manager

mscolnick:
Since narwhals.v2 supports `ibis`, we can share more code from the narwhals table manager. Migrating over histograms does not work.


Light2Dark (approved):
tested with some datasets, seems to work fine

PR #6648, opened 2025-10-02
fix: wasm with starlette issue

mscolnick:
Some of the refactors to `mpl` brought in starlette top-level which does not work for work in wasm


marimo/_plugins/stateless/mpl/_mpl.py
```diff
@@ -222,39 +221,38 @@ def _template(fig_id: str, port: int) -> str:
     }
 
 
-async def mpl_js(request: Request) -> Response:
```
dmadisetti, line 225: These are used in endpoint

dmadisetti (approved):

PR #6657, opened 2025-10-02
fix: Enable stub mechanism for caching un-pickable objects

### LINKED ISSUE #5542: Persistent cache keys that contain Pydantic models are not deterministically hashable
### Describe the bug

Root cause: https://github.com/pydantic/pydantic/issues/11603

Workaround: use a custom Pickler in `hash.py` like this (probably, I didn't test this):

```py
try:
    from pydantic import BaseModel
    from pydantic_core import PydanticUndefined

    PYDANTIC_INSTALLED = True
except ImportError:
    PYDANTIC_INSTALLED = False
    BaseModel = type("BaseModel", (), {})  # type: ignore[misc]
    PydanticUndefined = object()

class CustomPickler(Pickler):
    
    def reducer_override(self, obj: Any) -> Any:
        if not PYDANTIC_INSTALLED:
            return NotImplemented
        
        if isinstance(obj, BaseModel):
            private = getattr(obj, "__pydantic_private__", None)
            if private:
                private = {
                    k: v
                    for k, v in private.items()
                    if v is not PydanticUndefined
                }

            # Reconstruct Pydantic's __getstate__ logic, but with a sorted
            # __pydantic_fields_set__ for deterministic pickling.
            state = {
                "__dict__": obj.__dict__,
                "__pydantic_extra__": getattr(obj, "__pydantic_extra__", {}),
                "__pydantic_fields_set__": sorted(
                    getattr(obj, "__pydantic_fields_set__", set())
                ),
                "__pydantic_private__": private,
            }
            # Unsafe construction is used to bypass validation on unpickling,
            # which is the standard behavior for Pydantic models.
            return (obj.__class__.model_construct, (), state)
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

_No response_

dmadisetti:
## 📝 Summary

closes #5542 

- Introduces a `CustomStub` class which enables custom methods of hashing, saving, and restoring objects to and from cache.
- Implements PydanticStub to ensure deterministic pickling for pydantic base models

### DIFF
--- a/marimo/_save/stubs/__init__.py
+++ b/marimo/_save/stubs/__init__.py
@@ -42,7 +42,13 @@ def maybe_register_stub(value: Any) -> bool:
         return True
 
     # Walk MRO to find matching base class
-    for cls in value_type.__mro__:
+    try:
+        mro_list = value_type.mro()
+    except BaseException:
+        # Some exotic metaclasses or broken types may raise when calling mro
+        mro_list = [value_type]
+
+    for cls in mro_list:
         if not (hasattr(cls, "__module__") and hasattr(cls, "__name__")):
             continue
 


marimo/_save/stubs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        or newly registered), False otherwise.
+    """
+    value_type = type(value)
+
+    # Already registered in CUSTOM_STUBS
+    if value_type in CUSTOM_STUBS:
+        return True
+
+    # Walk MRO to find matching base class
+    for cls in value_type.__mro__:
```
mscolnick: i did this in formatters, and it has a chance of raising errors in odd edge maybe either catch the exception or comment if we just want to propoage it up

marimo/_save/stubs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+from marimo._save.stubs.pydantic_stub import PydanticStub
+from marimo._save.stubs.stubs import (
+    CUSTOM_STUBS,
+    CustomStub,
+    register_stub,
+)
+from marimo._save.stubs.ui_element_stub import UIElementStub
+
+# Track which class names we've already attempted to register
+_REGISTERED_NAMES: set[str] = set()
```
mscolnick, line 19: is this to terminate early? it doesn't look like its being used

mscolnick: nvm, i see it

mscolnick (approved):

marimo/_save/stubs/__init__.py
dmadisetti: Nice, looked at the formatters change, and a little too small to meaningfully DRY- but hopefully this saves us a bug report

mscolnick (approved):

PR #6659, opened 2025-10-02
improvement: mcp server ui, auth, and rules

mscolnick:
- cleanup MCP server ui in the CLI
- add auth to /mcp/route
- set defaults like remove server-token and add watch
- add Tool to fetch marimo rules


marimo/_server/print.py
```diff
@@ -163,14 +163,18 @@ def print_mcp_server(mcp_url: str, server_token: str | None) -> None:
...
-        f"{_utf8('🔗')} {green('Experimental MCP Server Configuration', bold=True)}"
+        f"{_utf8('🔗')} {green('Experimental MCP server configuration', bold=True)}"
     )
     print_tabbed(
-        f"{_utf8('➜')}  {green('MCP Server URL')}: {_colorized_url(mcp_url)}"
+        f"{_utf8('➜')}  {green('MCP server URL')}: {_colorized_url(mcp_url)}"
+    )
+    # Add to Claude code
+    print_tabbed(
+        f"{_utf8('➜')}  {green('Add to Claude Code')}: claude mcp add --transport http marimo {mcp_url}"
```
bjoaquinc, line 173: Should we also include `'add to Cursor'`? Since the most likely use cases for this are Claude Code and Cursor. It can be something like:

```python
print_tabbed(
    f"{_utf8('➜')} {green('Add to Cursor')}: Add the following to ~/.cursor/mcp.json:\n"
    f"""  {{
    "mcpServers": {{
      "marimo": {{
        "url": "{mcp_url}"
      }}
    }}
  }}"""
)
```

marimo/_ai/_tools/tools/rules.py
```diff
@@ -0,0 +1,63 @@
...
+MARIMO_RULES_URL = "https://docs.marimo.io/CLAUDE.md"
+
+
+@dataclass
+class GetMarimoRulesOutput(SuccessResult):
+    rules_content: Optional[str] = None
+    source_url: str = MARIMO_RULES_URL
+
+
+class GetMarimoRules(ToolBase[EmptyArgs, GetMarimoRulesOutput]):
```
bjoaquinc, line 25: This is really good! Maybe we should add this to the `next_steps=` suggestion for some of the other tools?
My suggestions would be:
- `GetNotebookErrors()` if there are any notebook specific or marimo specific errors
- `GetLightweightCellMap()` and `GetActiveNotebooks()` because these are entry points.
- `GetDatabaseTables()` because the rules contain some SQL-specific guidelines

bjoaquinc:
Also not sure if this is intentional or not but since all the other tools have tests, might be nice to include some tests too for `GetMarimoRules` in `tests/_ai/tools/tools/test_rules.py`? Esp since it uses an external http dep.

marimo/_mcp/server/main.py
```diff
@@ -53,7 +53,23 @@ def setup_mcp_server(app: "Starlette") -> "StreamableHTTPSessionManager":
     # Initialize streamable HTTP app
     mcp_app = mcp.streamable_http_app()
 
+    # Middleware to require edit scope
+    class RequiresEditMiddleware(BaseHTTPMiddleware):
```
bjoaquinc, line 57: `RequiresEditMiddleware()` doesn't seem to be mcp specific at all. Maybe it would be better to move this logic to `marimo/_server/api/middleware.py`?

marimo/_server/print.py
mscolnick: that seems a bit verbose for the CLI. we could look into making the CLI more interface in the future (e.g. `press i for more info`)

marimo/_ai/_tools/tools/rules.py
mscolnick: maybe we can do this in a followup. it hard to tell why an AI would ask for rules so im not sure if the next steps would veer it off track

marimo/_mcp/server/main.py
mscolnick: this is actually just for MCP. we already support auth on other endpoints with `@requires`

marimo/_ai/_tools/tools/rules.py
bjoaquinc: Ok sounds good. I'll keep this incase we decide to update it later. We can see first if the Chat LLMs either don't pull rules initially or pull it early on in the conversation then forget it.

marimo/_server/print.py
bjoaquinc: Yeah that makes sense

PR #6674, opened 2025-10-03
preview: enable cell level caching

dmadisetti:
## 📝 Summary

Experimental executor for cell level caching

Status:
 - Implemented (no known bugs after latest pass)
 - Slower than desired
 - Backburner until demand arrives

Needed Features:
 - Requires break glass on per notebook level (could be followup PR)
 - Missing "Time Saved" (could be followup PR)

Refactor:
 - Consolidate data on the context level (currently 3 fields particular to cell level caching)
 - Clean up serialization. Converted from protobuf- but some of the design decisions in the conversion are more messy than useful
 - Clean up tests
 - Clean up cache file locations


dmadisetti:
Good reference but closing

PR #6678, opened 2025-10-04
feat: add absolute line numbers for improved AI agent communication

lucharo:
## Summary / Use case

When working with AI coding assistants on marimo notebooks, it's much clearer to reference code by its actual line number in the source file (e.g., "modify line 42") rather than using relative cell references. This is particularly valuable in notebooks with many cells where cell-based references become difficult to track.

## Features Implemented

- **Absolute line number display**: CodeMirror extension that shows actual line numbers from the Python script file
- **Keyboard shortcut**: `Cmd+Shift+A` (Ctrl+Shift+A on Windows/Linux) to toggle absolute line numbers
- **UI toggle**: Settings → Display → "Absolute line numbers" checkbox
- **Visual indicator**: Small "A" badge appears in the top-left corner of cells when absolute mode is active
- **Configuration persistence**: Setting is saved in user config and persists across sessions

https://github.com/user-attachments/assets/98faa458-351b-4590-94c8-5b0c68562f25

Note I also added editable marimo install in the `pixi` section of `pyproject.toml` as performed in the #6679 PR

## test cases

- [ ] Does this work if you move cells around?
- [ ] Does this work for multi-column? 
- [ ] As you add new lines, do they get updated?
- [ ] updates when adding/deleting cells? 

## caveat

doesn't work in a temporary notebook when a file hasn't been assigned yet (e.g. when creating notebook via `marimo edit` rather than `marimo edit nb.py`)

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -9,23 +9,23 @@ environments:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
-      - conda: https://conda.anaconda.org/conda-forge/noarch/anyio-4.9.0-pyh29332c3_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/anyio-4.11.0-pyhcf101f3_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/backports-1.0-pyhd8ed1ab_5.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/backports.tarfile-1.2.0-pyhd8ed1ab_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-h4bc722e_7.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.4.26-hbd8a1cb_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/certifi-2025.4.26-pyhd8ed1ab_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cffi-1.17.1-py312h06ac9bb_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-hda65f42_8.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.8.3-hbd8a1cb_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/certifi-2025.8.3-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/cffi-2.0.0-py312h35888ee_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cfgv-3.3.1-pyhd8ed1ab_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/click-8.2.1-pyh707e725_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cryptography-45.0.3-py312hda17c39_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/click-8.3.0-pyh707e725_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/cryptography-46.0.2-py312hee9fe19_0.conda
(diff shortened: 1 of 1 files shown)


marimo/_ast/cell_manager.py
```diff
@@ -244,10 +250,19 @@ def register_ir_cell(
...
                 cell=None,
+                lineno=cell_def.lineno,
+                end_lineno=cell_def.end_lineno,
             )
             return
         cell._cell.configure(cell_config)
+        # Store line numbers in cell data after registering
         self._register_cell(cell, app=app)
+        # Update with line numbers
+        if cell._cell.cell_id in self._cell_data:
```
dmadisetti, line 261: So this will work when you first load the notebook. But updating the notebook (adding/deleting cells, adding new lines etc etc) and this won't work anymore.

It's a nice idea, but there's more to keep the back end in sync with the frontend

mscolnick:
Does this work if you move cells around? Does this work for multi-column? As you add new lines, do they get updated?

lucharo:
hey guys, thanks for the feedback, indeed in those scenarios it's not currently working, I added those as test cases to the PR description. 

It does work when adding new lines but indeed when you move cells around or add a new cell it doesn't work, I'll keep working on it

dmadisetti:
@lucharo it's a pretty hairy, I considered going down this path for adding debugging capabilities- but instead settled for https://github.com/marimo-team/marimo/pull/6311/files#diff-26c1c93dacf92c199621cd07b116adc86d6d6fc66e5447d5a94a6c51c51f05b5R1404 which just re-parses the notebook on execution every single time.

But if you're up for it, would be awesome

mscolnick:
re-parsing on execution sounds not ideal for performance. i understand this would be behind a config flag, but just something to consider both if you still plan to implement this and something worth showing in the UI

lucharo:
> @lucharo it's a pretty hairy, I considered going down this path for adding debugging capabilities- but instead settled for https://github.com/marimo-team/marimo/pull/6311/files#diff-26c1c93dacf92c199621cd07b116adc86d6d6fc66e5447d5a94a6c51c51f05b5R1404 which just re-parses the notebook on execution every single time.
> 
> But if you're up for it, would be awesome

which bit do you consider hairy? is it the keeping track of line count without reparsing the source script?

dmadisetti:
> which bit do you consider hairy? is it the keeping track of line count without reparsing the source script?

Yes, namely because what is in memory is not bound to be what's on disk (esp with autosave off). Toggling could send a request to save and then get line offsets. This would be a simplification over attempting to carry around line number and do constant accounting (which is the more complex, edge-case filled "hairy" issue)

lucharo:
> > which bit do you consider hairy? is it the keeping track of line count without reparsing the source script?
> 
> Yes, namely because what is in memory is not bound to be what's on disk (esp with autosave off). Toggling could send a request to save and then get line offsets. This would be a simplification over attempting to carry around line number and do constant accounting (which is the more complex, edge-case filled "hairy" issue)

yeah I thought so! I am thinking that this could be a feature that's only really available when `--watch` or auto-save is on as parse on save seems to be probably the sanest approach. This would make sense in the sense that having absolute line count would probably only makes sense when people are running marimo alongside an agent and likely have watch and auto save on

and clear on the keeping track of line number, that'd be indeed hairy, I understood correctly! I'll keep cracking on this feature working "on save" (if you guys like the feature enough)

mscolnick:
@lucharo - i think even with on-save, the line numbers will be outdated. personally, i think the complexity is not worth effort. 

taking a large step back, is there another way you can give Claude context of where is should focus that is not based on line numbers (e.g. cell name, cell contents)

lucharo:
> @lucharo - i think even with on-save, the line numbers will be outdated. personally, i think the complexity is not worth effort. 
> 
> 
> 
> taking a large step back, is there another way you can give Claude context of where is should focus that is not based on line numbers (e.g. cell name, cell contents)

Fair enough! I won't pursue it then.

I think cell name would work but by default cells are not named. It'd be great to have something that works out of the box. It could be cell number (again, toggleable) so that the human and the agent both have something to share perspectives over. I think agents would probably need to "count" the number of cells and might get it wrong sometimes. Cell content works fine but again might lose precision in longer periods, can copy and paste the exact function/snippet of code but that also makes for worse UX I think. 

I think the reason I suggested line count was because I see agents often use grep or sed or similar tools to scan files so line count works out of the box for referencing files. 

What are your thoughts?

mscolnick:
im not sure how large your notebooks are, but i find agents pretty good at consuming the whole file. not sure i have other thoughts on other ways give exact context, but open to explore

dmadisetti:
@lucharo we did have a feature request/ idea for auto-naming cells with an agent. I can't find the exact issue- but maybe that would help with your issue?

lucharo:
This line of work is too thorny and with how marimo is currently set up would be very expensive to set up (parsing source script at every save) so will abandone for other simpler ways of referencing code cells, thanks @dmadisetti and @mscolnick for the discussion!

PR #6727, opened 2025-10-10
fix: manually move copilot file extension from js to cjs

### LINKED ISSUE #6718: copilot lsp process crashes when containing package.json type is module
### Describe the bug

Related to https://github.com/marimo-team/marimo/issues/4263#issue-2949293229, but I think that issue is about the marimo lsp "launcher", which based on `ps -ef` is using a cjs file. This issue is with the copilot language server itself, which runs as `node /home/foo/.venv/lib/python3.12/site-packages/marimo/_lsp/copilot/language-server.js --stdio`

When there is a package.json with `"type": "module"` in the parent folder of the `.venv` running marimo, the copilot process crashes silently. The lsp logs just show dropped connections:

```txt
# ~/.cache/marimo/logs/github-copilot-lsp.log
[INFO] Parsed LSP command: node /home/foo/.venv/lib/python3.11/site-packages/marimo/_lsp/copilot/language-server.js --stdio
[INFO] New connection from ::ffff:127.0.0.1
[INFO] WebSocket closed with code 1005: 
```

When I run the language server directly I get the actual error:

```txt
❯ node /home/foo/.venv/lib/python3.11/site-packages/marimo/_lsp/copilot/language-server.js --version
file:///home/foo/.venv/lib/python3.11/site-packages/marimo/_lsp/copilot/language-server.js:23
require('./main').main();
^

ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/home/foo/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///home/foo/.venv/lib/python3.11/site-packages/marimo/_lsp/copilot/language-server.js:23:1
    at ModuleJob.run (node:internal/modules/esm/module_job:377:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:689:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v24.9.0
```

removing `"type": "module"` from my workspace package.json fixes it but the lsp should not be dependent on the containing folder.

Other errors also cause silent failures, such as a misformatted package.json. So ideally the fix isolates the lsp process and bubbles crashes up to the lsp or marimo logs.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.16.5",
  "editable": false,
  "location": "/home/foo/.venv/lib/python3.11/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.8.0-85-generic",
  "Processor": "x86_64",
  "Python Version": "3.11.13",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "140.0.7339.207",
    "Node": "v24.9.0"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.7.0",
    "packaging": "25.0",
    "psutil": "7.1.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.8.1"
  },
  "Experimental Flags": {}
}```

</details>


### Code to reproduce

```bash
mkdir tmp
cd tmp
# confirm that this does not have "type": "module"
npm init -y
uv init
uv add marimo
uv run marimo edit scratch.py
```

set up copilot and observe that it connects.

Now edit package.json and set `"type": "module"` then rerun `uv run marimo edit scratch.py`. Observe in the browser that you get a connection error after 5-10 seconds.

mscolnick:
Fixes #6718

Newer versions of node support `ejs` so they now try to parse `.js` files based on the package.json module type. This comes from a downstream dep but we manually move .js to .cjs

### DIFF
--- a/packages/lsp/move.js
+++ b/packages/lsp/move.js
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+const fs = require("node:fs");
+const path = require("node:path");
+
+// Copy node_modules/@github/copilot-language-server/dist/ to dist/
+const srcDir = path.join(
+  __dirname,
+  "node_modules/@github/copilot-language-server/dist",
+);
+const destDir = path.join(__dirname, "dist");
+
+// Recursively copy directory
+fs.cpSync(srcDir, destDir, { recursive: true, dereference: true });
+
+// Rename language-server.js to language-server.cjs
+const oldPath = path.join(destDir, "language-server.js");
+const newPath = path.join(destDir, "language-server.cjs");
+fs.renameSync(oldPath, newPath);
+
+// biome-ignore lint/suspicious/noConsole: build script
+console.log("Successfully copied and renamed language-server files");
--- a/packages/lsp/package.json
+++ b/packages/lsp/package.json
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "build:ts": "tsup index.ts --format cjs --minify",
-    "build:deps": "cp -LR node_modules/@github/copilot-language-server/dist/ dist/ && mv dist/language-server.js dist/language-server.cjs",
+    "build:deps": "node move.js",
     "typecheck": "tsgo",
     "build": "pnpm run build:ts && pnpm run build:deps",
     "test": "vitest"


packages/lsp/move.js
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env node
```
dmadisetti, line 1: Is there a benefit of this over `cp` and `mv`? If so can we add a comment?

PR #6749, opened 2025-10-12
fix: msgspec Python 3.14 compatibility

akshayka:
Python 3.14 introduced changes to annotations, which broke `msgspec` structs; unfortunately, `msgspec` is no longer actively maintained.

While we search for an alternative or long-term solution, a short-term workaround is to import annotations from `__future__`.

* issue: https://github.com/jcrist/msgspec/issues/847
* example workaround: https://github.com/return42/searxng/commit/3963c2371a3a33728771ebef0d173b3d47fc4e11

This PR adds the import to more modules than needed; not sure what the minimal set is.

### DIFF
--- a/marimo/__init__.py
+++ b/marimo/__init__.py
@@ -14,8 +14,6 @@ marimo is designed to be:
     5. fun
 """
 
-from __future__ import annotations
-
 __all__ = [
     # Core API
     "App",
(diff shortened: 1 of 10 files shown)


marimo/__init__.py
```diff
@@ -14,6 +14,8 @@
     5. fun
 """
 
+from __future__ import annotations
+
```
dmadisetti:
```suggestion
```

Failing API test

dmadisetti: NVM, more changes than that in all the inits. Maybe just worth updating the snapshot or test

PR #6777, opened 2025-10-14
Add CI workflow to test pre-built wheel installation

manzt:
This workflow verifies that marimo's wheel can be installed and run without requiring any build from source across Python 3.10-3.14. It uses `uv tool run --no-build` to ensure the wheel is truly standalone and doesn't depend on build tools being available at install time (only on ubuntu for now).

### DIFF
--- a/.github/workflows/test_no_build.yaml
+++ b/.github/workflows/test_no_build.yaml
@@ -7,10 +7,17 @@ on:
   push:
     branches:
       - main
+    paths:
+      - .github/workflows/test_no_build.yaml
+      - pyproject.toml
+      - marimo/_cli/**
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/test_no_build.yaml
```diff
@@ -0,0 +1,33 @@
...
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
```
mscolnick: should we filter to just certain paths? maybe:
```
.github/workflows/test_no_build.yaml
pyproject.toml
marimo/_cli
```

manzt: Done.

PR #6780, opened 2025-10-14
chore: remove importing directly from marimo

mscolnick:
This reduces the dependency graph that `pytest-changed` picks up


dmadisetti (approved):
A little surprised these are the only instances, but that's a good thing!

Issue #6787, opened 2025-10-15
Error when creating table with geometry column on in-memory DuckDB


hugoboia:
### Describe the bug

When using DuckDB in-memory, after loading the Spatial extension, running CREATE TABLE to make a table that has a geometry column, creates the table but shows the following logs:

```sql
_df = mo.sql(
    f"""
    INSTALL spatial;
    LOAD spatial;
    """
)
```

```sql
_df = mo.sql(
    f"""
    CREATE TABLE gdf AS
    SELECT * 
    FROM 'gdf.parquet'
    """
)
```

```
--- Logging error ---
Traceback (most recent call last):
  File "C:\Users\hmsb\AppData\Roaming\uv\python\cpython-3.13.6-windows-x86_64-none\Lib\logging\__init__.py", line 1154, in emit
    stream.write(msg + self.terminator)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
OSError: [WinError 1] Função incorreta
Call stack:
  File "<string>", line 1, in <module>
    from multiprocessing.spawn import spawn_main; spawn_main(parent_pid=16996, pipe_handle=1096)
  File "C:\Users\hmsb\AppData\Roaming\uv\python\cpython-3.13.6-windows-x86_64-none\Lib\multiprocessing\spawn.py", line 122, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "C:\Users\hmsb\AppData\Roaming\uv\python\cpython-3.13.6-windows-x86_64-none\Lib\multiprocessing\spawn.py", line 135, in _main
    return self._bootstrap(parent_sentinel)
  File "C:\Users\hmsb\AppData\Roaming\uv\python\cpython-3.13.6-windows-x86_64-none\Lib\multiprocessing\process.py", line 313, in _bootstrap
    self.run()
  File "C:\Users\hmsb\AppData\Roaming\uv\python\cpython-3.13.6-windows-x86_64-none\Lib\multiprocessing\process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\hmsb\AppData\Local\uv\cache\archive-v0\KW5E2onlDbMGojXznUPgG\Lib\site-packages\marimo\_runtime\runtime.py", line 3071, in launch_kernel
    asyncio.run(control_loop(kernel))
[...]

PR #6806, opened 2025-10-15
fix: duckdb interrupt

akshayka:
This change adds more graceful interrupts of DuckDB queries executed through `mo.sql()`.

Interrupting a connection with a `SIGINT` can apparently leave the connection in a bad state. This change instead interrupts the connection with its `interrupt()` method. To enable this, the `ExecutionContext` is augmented with an additional property in which the runtime can store the active duckdb connection.

This is difficult to test programmatically. Tested manually by interrupting the query in https://github.com/marimo-team/marimo/issues/6787. Basic unit tests are included to ensure the interrupt handler does the right thing.


marimo/_runtime/context/types.py
```diff
@@ -67,6 +69,16 @@ class ExecutionContext:
     local_cell_id: Optional[CellId_t] = None
     # output object set imperatively
     output: Optional[list[Html]] = None
+    duckdb_connection: duckdb.DuckDBPyConnection | None = None
+
+    @contextmanager
+    def with_connection(
+        self, connection: duckdb.DuckDBPyConnection
+    ) -> Iterator[None]:
+        old_conn = self.duckdb_connection
```
dmadisetti, line 78: Is there ever a case where this wouldn't be None? Maybe safer to set to None?

akshayka: Standard context manager practice is to save a new value and restore what was there previously, so I'd prefer to just do that here as well.

akshayka:
Failing in CI in a perplexing way ...

```
FAILED tests/_runtime/test_runtime.py::TestExecution::test_set_ui_element_value_with_cell_run[k] - KeyError: 'defs'
FAILED tests/_runtime/test_runtime.py::TestExecution::test_set_ui_element_value_with_cell_run[strict_kernel] - KeyError: 'defs'
FAILED tests/_runtime/test_runtime.py::TestExecution::test_set_ui_element_value_with_cell_run[lazy_kernel] - KeyError: 'defs'
```

PR #6808, opened 2025-10-16
chore: migrate out smart cell logic to its own package

mscolnick:
Internal refactor to help share code with vscode-extension for our smart cells


frontend/src/core/codemirror/language/__tests__/extension.test.ts
```diff
@@ -174,17 +174,25 @@ describe("switchLanguage", () => {
...
     switchLanguage(mockEditor, { language: "sql", keepCodeAsIs: false });
     expect(mockEditor.state.doc.toString()).toMatchInlineSnapshot(`
-      "print('Hello')
-      print('Goodbye')"
+      "mo.md(
+          r"""
+      print('Hello')
+      print('Goodbye')
+      """
+      )"
```
mscolnick, line 182: these changed because we used to try to do something smart that when SQL's transformIn when receiving markdown-wrapped code `mo.md(...)` we tried to extra that as well, but this was a bit leaky

Light2Dark: got it, I think we can add this later? Maybe keep one test a xfail

Light2Dark (approved):

frontend/src/core/codemirror/language/__tests__/extension.test.ts
mscolnick: i can create a ticket

PR #6831, opened 2025-10-17
fix: ci breakage in python 3.13.8

dmadisetti:
## 📝 Summary

Bug in python 3.13.8 incorrectly parses comment in the decorator
Should unblock ci, will follow up with more robust fix

### DIFF
--- a/tests/_ast/codegen_data/test_generate_filecontents_toplevel.py
+++ b/tests/_ast/codegen_data/test_generate_filecontents_toplevel.py
@@ -31,7 +31,6 @@ def _():
     return globe, shadow
 
 
-# Sanity check that base case works.
 @app.function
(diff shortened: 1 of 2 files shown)


dmadisetti:
I think maybe https://github.com/python/cpython/pull/106411/files
but will have to confirm

mscolnick (dismissed):

dmadisetti:
@mscolnick wait collection works, but the parsing itself is still breaking. I think we can patch this side

dmadisetti:
Nvm, just additional cases. The bug could break some user notebooks (top level function with comments above), but releases are maybe 2 - 3 months in between, and will hopefully be patched

mscolnick (dismissed):

PR #6845, opened 2025-10-20
prototypes to create new cells: in-between, + style, dropdown

Light2Dark:
## 📝 Summary

- subtler left bar to create new cells (create above opens a dropdown, create below is as normal)
- in between cells a few buttons appear

https://github.com/user-attachments/assets/203498da-e5b4-4512-a8f8-5c94695ebe69

## 📋 Checklist

### DIFF
--- a/frontend/src/components/editor/notebook-cell.tsx
+++ b/frontend/src/components/editor/notebook-cell.tsx
@@ -73,6 +73,7 @@ import { derefNotNull } from "../../utils/dereference";
 import { Functions } from "../../utils/functions";
 import { Logger } from "../../utils/Logger";
 import { renderShortcut } from "../shortcuts/renderShortcut";
+import { Button } from "../ui/button";
 import { CellStatusComponent } from "./cell/CellStatus";
 // import { CreateCellButton } from "./cell/CreateCellButton"; // OLD PROTOTYPE
 import {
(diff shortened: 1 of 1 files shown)


akshayka:
Oh nice! The `+` buttons look nice. I wonder if they should be positioned at the top and bottom of the entire cell, including outputs, instead of just the top and bottom of the code editor. I'm looking at the below frame in particular. cc @manzt 

image

manzt:
Really nice putting this together. I pulled it down and tried it out. Overall, I prefer the left sidebar `+` (over the 1 click actions) and agree with @akshayka that having the `+ ---- +` span the full cell (including outputs) would be nice to position the buttons closer to where the action takes place.

Probably just a impl detail, but i noticed that the add above and add below actions are slightly different. Top opens a popover to select a cell kind; bottom inserts a Python cell. I like the popover because we have other quick actions to add cells (keyboard shortcuts), and it gives us some future real estate should we ever want to expand what a user can insert (e.g., snippets, ai)

Light2Dark:
https://github.com/user-attachments/assets/43f8ea27-8baa-498b-af7a-851a9c54bf72

Thanks @manzt @akshayka , I made some improvements to the prototype:
- plus buttons span the entire cell
- remove vertical line (you can increase the opacity to see it, but i felt it was noisy)
- only 1 button for the in-between cell action with dropdown

i like both options now, we can discuss specific behaviours (on-click, right click, shift+click) later.

Items to resolve:
- which design
- on 1 click, do what. right click? shift+click?

PR #6849, opened 2025-10-20
improvement: auto_instantiate=false by default

dmadisetti:
## 📝 Summary

Sets `auto_instantiate` to be false by default, in addition to providing env hooks to enforce this.

### DIFF
--- a/marimo/_config/reader.py
+++ b/marimo/_config/reader.py
@@ -40,8 +40,7 @@ def sanitize_pyproject_dict(
             if key in current_level and isinstance(current_level[key], dict):
                 current_level = current_level[key]
             else:
-                current_level = None
-                break
+                return pyproject_dict
         if current_level and key_path[-1] in current_level:
             del current_level[key_path[-1]]
     return pyproject_dict


marimo/_config/manager.py
```diff
@@ -300,6 +302,38 @@ def _resolve_vimrc(
...
+        current[path[-1]] = value
+        return None
+
+    def get_config(self, *, hide_secrets: bool = True) -> PartialMarimoConfig:
+        """Get the configuration, as a partial configuration"""
+        project_config: PartialMarimoConfig = {}
+        # We could do this dynamically, but list explicitly for now to reduce
+        # surface area
+        self._maybe_override_from_env(
+            "_MARIMO_CONFIG_OVERLOAD_AUTO_INSTANTIATE",
```
mscolnick: Is this meant to be private? I would be ok if it's public if it follows a strict pattern (based on the paths in the coding)

mscolnick (commented):
we should also have a deny-list of what can be overridden from the ScriptMetadata (such as auto-instantiate)

marimo/_config/manager.py
dmadisetti: I think private makes sense for now- we can explore if we need more knobs, or if there's an ask for it.

Overload is consistent with the other injected envs, but we can just rename them all

mscolnick: sounds good. for a pattern, maybe `_MARIMO_CONFIG_OVERLOAD_RUNTIME_AUTO_INSTANTIATE`

mscolnick: `_MARIMO_CONFIG_OVERLOAD_{uppercase path to key}`

PR #6856, opened 2025-10-21
improve: package installation autorun

akshayka:
When a package is installed, we now run cells with non-static dependencies on the associated module.

For example, if a cell imports a module that in turn imports pandas (but fails with a ModuleNotFoundError on pandas), we now mark that cell for execution after successful installation of pandas.

https://github.com/user-attachments/assets/ff2cf34a-8900-4bb2-a64d-f8a0939144fe

### DIFF
--- a/marimo/_ast/cell.py
+++ b/marimo/_ast/cell.py
@@ -122,6 +122,7 @@ RunResultStatusType = Literal[
 @dataclasses.dataclass
 class RunResultStatus:
     state: Optional[RunResultStatusType] = None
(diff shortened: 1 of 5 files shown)


dmadisetti (approved):
Nice, looks pretty clean- nice tidy too. See the spots for a refactor

koaning:
Yeah, this is a super nice quality of life improvement. Works as expected on my end :)

koaning (approved):
LGTM

PR #6866, opened 2025-10-21
tidy: remove python 3.9 support

dmadisetti:
## 📝 Summary

Python 3.9's EOL is Nov 1st https://devguide.python.org/versions/

This PR removes explicit python 3.9 support


dmadisetti:
We can probably shelve this for subsequent release (closer to Nov)

Issue #6870, opened 2025-10-21
Connecting to a remote marimo instance interferes with local connection


akshayka:
### Describe the bug

1. On local machine, run `marimo edit --port 2718` and start a notebook
2. On a remote machine, run `marimo edit --headless --host 127.0.0.1 --port 2718`
3. On local, connect to the marimo server in (2) on port 2020 using port forwarding, then open a notebook:

`ssh -N -L 2720:127.0.0.1:2718 <REMOTE_USER>@<REMOTE_HOST>`

In my repro, this was with the gcloud CLI:

```bash
gcloud compute ssh <INSTANCE> --zone=<ZONE> --tunnel-through-iap -- -N -L 2720:127.0.0.1:2718
```

The remote notebook can be interacted with fine. But the locally running notebook editor is interrupted by a "Sign in" alert. If you sign in with the token as username and password, then the notebook that's running remotely gets interrupted by a "Sign in alert."

<img width="316" height="253" alt="Image" src="https://github.com/user-attachments/assets/a87abd36-4c7b-4edb-8cff-1e43e7b45e5c" />

### Will you submit a PR?

- [ ] Yes

### Environment

marimo 0.16.5 on both machines.

<details>

```
Replace this line with the output of marimo env. Leave the backticks in place.
```

</details>

dmadisetti:
Although we shouldn't put this on the user, does the following work?

`marimo edit --headless --host 127.0.0.1 --port 2718 --proxy=http://127.0.0.1:2720`

akshayka:
Yeah I think it will work, I was just looking into that. In the meantime, we can change our docs to recommend a port other than 2718 on remote (eg, 3718)

ShruAgarwal:
Hi @akshayka,
Please close this issue, as it has a merged PR #6871 :)

akshayka:
This is not completed, the associated PR is just a docs workaround

ffmiruz:
Possibly cookie clash due to: https://github.com/marimo-team/marimo/blob/543bea433c4f203952ff8075dd62eba7e241116d/marimo/_server/api/auth.py#L211-L217
Is appending some nonce value when constructing session_cookie an acceptable solution? 
Could be timestamp, could be random value.

PR #6871, opened 2025-10-21
docs: more prescriptive remote connection docs

akshayka:
Give guidance on how to connect with port forwarding to a remote server.

This is a workaround for #6870

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -434,17 +434,29 @@ You can use any Python package. marimo cells run arbitrary Python code.
 
 Use SSH port-forwarding to run marimo on a remote server
 and connect to it from a browser on your local machine. Make sure
-to pass the `--headless` flag when starting marimo on remote:
+to pass the `--headless` flag when starting marimo on remote; on the remote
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -434,18 +434,30 @@ You can use any Python package. marimo cells run arbitrary Python code.
...
 marimo edit --headless --host 0.0.0.0 --port 8080
 ```
 
+_On local, run:_
+
+```
+ssh -N -L 3718:127.0.0.1:8080 REMOTE_USER@REMOTE_HOST
+```
+
+Then open `localhost:3718` in your browser.
```
dmadisetti, line 459: I think we do an offset for LSP and other internal servers (starting from 400): marimo/_server/start.py

Not an issue, just a note, and double checked because I thought it might be 1000

dmadisetti (approved):
Cool!

I would do `1`2718 but comes down to personal preferences

docs/faq.md
akshayka: Nice note, thanks for the context

PR #6879, opened 2025-10-22
improvement: fallback to available storage when localStorage is not available

mscolnick:
When iframing marimo in a sandboxed iframe, depending on configuration `localStorage` may not be available so this adds some appropriate fallbacks

this also adds a lint rule (grit) to make sure we pass the optional storage into `atomWithStorage`


frontend/lint/atomWithStorageArgs.grit
manzt, line 1: 🔥

PR #6885, opened 2025-10-22
fix: respect sandbox confirmation

akshayka:
This fixes a bug in which we didn't respect sandbox confirmation.

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -109,7 +109,6 @@ def should_run_in_sandbox(
     # to start in sandbox mode by examining the notebook file and
     # prompting the user.
     if sandbox is None:
-
         sandbox = maybe_prompt_run_in_sandbox(name)
 
     # Validation: we don't yet support multi-notebook sandboxed servers.


marimo/_cli/cli.py
```diff
@@ -535,45 +537,9 @@ def edit(
...
-            click.style(
-                "Warning: Using sandbox with multi-notebook edit servers is dangerous.\n",
-                fg="yellow",
-            )
-            + "Notebook dependencies may not be respected, may not be written, and may be overwritten.\n"
-            + "Learn more: https://github.com/marimo-team/marimo/issues/5219l.\n",
-            err=True,
-        )
-
-    if sandbox is None:
```
akshayka, line 556: We had a faulty `if -> elif` cascade. Refactored this code and added tests.

marimo/_cli/sandbox.py
```diff
@@ -75,6 +75,62 @@ def maybe_prompt_run_in_sandbox(name: str | None) -> bool:
...
+            + "Notebook dependencies may not be respected, may not be written, and may be overwritten.\n"
+            + "Learn more: https://github.com/marimo-team/marimo/issues/5219l.\n",
+            err=True,
+        )
+
+    # When the sandbox flag is omitted we infer whether to
+    # to start in sandbox mode by examining the notebook file and
+    # prompting the user.
+    if sandbox is None:
+        from marimo._cli.sandbox import maybe_prompt_run_in_sandbox
```
dmadisetti: nit: you don't need to import
```suggestion
```

dmadisetti (changes requested):
Well at least we have a unit test now :)

dmadisetti (approved):
Nice

PR #6891, opened 2025-10-23
support column edits based on index, instead of id

Light2Dark:
## 📝 Summary

Better support multi-column edits by letting the AI pass column index instead of ID. This allows to add cells at the end of specific columns. The backend doesn't hold the same column id as the frontend, so this is better to do.

## 📋 Checklist


Light2Dark:
a bit tricky since the tools are registered on start. Maybe subscribe to an atom and change the tool instructions.

PR #6901, opened 2025-10-23
feat: add support for Codex agent

mscolnick:
Adds Codex to the available agents in the ACP sidebar and many other fixes:

* Add Codex command to the Agent dropdown
* Disable bottom buttons when the agent is running
* Dedupe todos/plans (in case they have status updates)
* some tools don't send a completion update, so we just always mark them completed if they are no longer the last block
* Collapse unchanged regions in the diff viewer
* Handle edge case when Claude ACP returns a different schema than the ACP protocol expects for diffs (`HACK: if the raw input is `abs_path`, `old_string`, `new_string` then handle it as if it is a diff`)
* Handle edge case were the title comes back `"undefined"` from the ACP

image

Screenshot 2025-10-27 at 4 25 16 PM

Screenshot 2025-10-27 at 4 24 14 PM

### DIFF
--- a/frontend/src/components/chat/acp/blocks.tsx
+++ b/frontend/src/components/chat/acp/blocks.tsx
@@ -115,7 +115,7 @@ export const ErrorBlock = (props: {
   if (error instanceof JsonRpcError) {
     const dataStr =
       typeof error.data === "string" ? error.data : JSON.stringify(error.data);
-    message = `${dataStr} (${error.code})`;
+    message = `${dataStr} (code: ${error.code})`;
   }
 
   return (


frontend/src/components/chat/acp/blocks.tsx
```diff
@@ -102,13 +104,20 @@ export const ErrorBlock = (props: {
...
 
   // Don't show WebSocket connection errors
   if (message.includes("WebSocket")) {
     return null;
   }
 
+  if (error instanceof JsonRpcError) {
+    const dataStr =
+      typeof error.data === "string" ? error.data : JSON.stringify(error.data);
+    message = `${dataStr} (${error.code})`;
```
Light2Dark:
```suggestion
    message = `${dataStr} (code: ${error.code})`;
```
no strong opinion on this

frontend/src/components/chat/acp/blocks.tsx
```diff
@@ -651,8 +668,12 @@ export const DiffBlocks = (props: {
...
   item: Pick<ToolCallUpdateNotificationEvent, "title" | "kind" | "locations">,
 ) {
-  const prefix =
-    item.title || Strings.startCase(item.kind || "") || "Tool call";
+  let title = item.title;
+  // Hack: sometimes title comes back: "undefined", so lets undo that
+  if (title === '"undefined"') {
+    title = undefined;
+  }
+  const prefix = title || Strings.startCase(item.kind || "") || "Tool call";
```
Light2Dark, line 676: should it be this?

```suggestion
  const prefix = title || (item.kind ? Strings.startCase(item.kind) : "Tool call");
```

mscolnick: it evaluates to the same

PR #6912, opened 2025-10-23
improvement: better tracebacks, padding, remove button for static files

mscolnick:
Some QOL improvements to static html exports of marimo notebooks.

- tracebacks are rendered as rich html (instead of plain text)
- remove debugger and other buttons when in static rendering
- add padding to the bottom

### DIFF
--- a/frontend/src/core/cells/__tests__/session.test.ts
+++ b/frontend/src/core/cells/__tests__/session.test.ts
@@ -210,8 +210,18 @@ describe("notebookStateFromSession", () => {
 
     it("handles console outputs in session cell", () => {
       const consoleOutputs = [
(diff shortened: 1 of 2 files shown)


marimo/_schemas/generated/session.yaml
```diff
@@ -126,6 +126,31 @@ components:
...
+          - image/png
+          - image/svg+xml
+          - image/tiff
+          - image/avif
+          - image/bmp
+          - image/gif
+          - image/jpeg
+          - video/mp4
+          - video/mpeg
+          - text/html
```
dmadisetti, line 147: Why not only text and error?

Issue #6872, opened 2025-10-21
templates query not quoted


thisismygitrepo:
### Describe the bug

When clicking (+) on table in database, a simple select statement populates a cell, which is nice, however, it fails to run because pgsql, timescaledb and many many other databases are case sensitive. Can you guys please surround the table name and the schema with quotes? that should solve it and every other db is happy with it too.
PS: this also fix a massive range of other problems related to nameing (e.g. if table has a space in its name or it has a special chacarcter etc).

Please make it like `SELECT * FROM "TABLE_NAME"."SCHEMA_NAME" LIMIT 100`.

<img width="1714" height="397" alt="Image" src="https://github.com/user-attachments/assets/d5631424-5451-48d2-a9c0-0d59a91be080" />

### Will you submit a PR?

- [ ] Yes

### Environment

󰕈 24.4.0 noble 󱆃 ➜ uvx marimo env
{
  "marimo": "0.17.0",
  "editable": false,
  "location": "/home/alex/.cache/uv/archive-v0/f96v7QZbERElKNL7VOj3u/lib/python3.14/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.6.87.2-microsoft-standard-WSL2",
  "Processor": "x86_64",
  "Python Version": "3.14.0",
  "Locale": "C",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.10.0"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.9.0",
    "packaging": "25.0",
    "psutil": "7.1.1",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.3",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.38.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {},
  "Experimental Flags": {}
}

Light2Dark:
Strange, is that the correct table name (streamStatus)? Why is the error showing lowercase. I think if we quoted it, postgres would enforce the case sensitivity.

cmiiw, unquoted names are case insensitive: https://stackoverflow.com/questions/20878932/are-postgresql-column-names-case-sensitive

Light2Dark:
[...]

PR #6915, opened 2025-10-24
generate different sql code based on dialect

### LINKED ISSUE #6875: Datasource "Add table to notebook" does not work for MS SQL
### Describe the bug

Marimo is using in the `Add table to notebook` function the SQL command `LIMIT`. This seems to be not available with MS SQL Servers.

I typically use 

```sql
SELECT TOP 100 * FROM SCHEMA.TABLENAME
```

instead of marimos default:
```sql
SELECT * FROM SCHEMA.TABLENAME LIMIT 100
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.17.0",
  "editable": false,
  "location": "OBFUSCATED/.venv/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.13.8",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.20.0"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.8.0",
    "packaging": "25.0",
    "psutil": "7.1.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "uvicorn": "0.37.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.4.1",
    "nbformat": "5.10.4",
    "openai": "2.3.0",
    "pandas": "2.3.3",
    "polars": "1.34.0",
    "pyarrow": "21.0.0",
    "loro": "1.8.1",
    "pytest": "8.4.2",
    "ruff": "0.14.0",
    "sqlglot": "27.27.0"
  },
  "Experimental Flags": {
    "sql_linter": true
  }
}
```

</details>

### Code to reproduce

Add a MS SQL server connection, using `sqlalchemy`:

```python
import sqlalchemy as sa
engine = sa.create_engine(
    "mssql+pyodbc:{URL}?driver=ODBC+Driver+17+for+SQL+Server"
)
```

Then navigate to the datasources in marimo, choose any table for the engine above and click the plus button to `Add table to notebook`.

This will add a sql query to the notebook:

```python
_df = mo.sql(
    f"""
    SELECT * FROM SCHEMA.TABLENAME LIMIT 100
    """,
    engine=engine
)
```

Light2Dark:
## 📝 Summary

Fixes #6875. Might also tackle #6872.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/datasources/utils.ts
+++ b/frontend/src/components/datasources/utils.ts
@@ -5,6 +5,7 @@ import { isKnownDialect } from "@/core/codemirror/language/languages/sql/utils";
 import type { SQLTableContext } from "@/core/datasets/data-source-connections";
 import { DUCKDB_ENGINE } from "@/core/datasets/engines";
 import type { DataTable, DataType } from "@/core/kernel/messages";
+import { logNever } from "@/utils/assertNever";
 import type { ColumnHeaderStatsKey } from "../data-table/types";
 
 // Some databases have no schemas, so we don't show it (eg. Clickhouse)
@@ -60,7 +61,26 @@ function getFormatter(dialect: string): SqlCodeFormatter {
         },
         formatSelectClause: defaultFormatter.formatSelectClause,
       };
+    case "postgresql":
+    case "postgres":
+    case "db2":
+    case "mysql":
+    case "sqlite":
+    case "duckdb":
+    case "mariadb":
+    case "cassandra":
+    case "noql":
+    case "athena":
+    case "hive":
+    case "redshift":
+    case "snowflake":
+    case "flink":
+    case "mongodb":
+    case "oracle":
+    case "oracledb":
+      return defaultFormatter;
     default:
+      logNever(dialect);
       return defaultFormatter;
   }
 }
@@ -109,10 +129,10 @@ export function sqlCode({
     );
 
     if (engine === DUCKDB_ENGINE) {
-      return `_df = mo.sql(f"""${selectClause}""")`;
+      return `_df = mo.sql(f"""\n${selectClause}\n""")`;
     }
 
-    return `_df = mo.sql(f"""${selectClause}""", engine=${engine})`;
+    return `_df = mo.sql(f"""\n${selectClause}\n""", engine=${engine})`;
   }
 
   return `_df = mo.sql(f'SELECT "${columnName}" FROM ${table.name} LIMIT 100')`;
--- a/frontend/src/core/codemirror/language/languages/sql/utils.ts
+++ b/frontend/src/core/codemirror/language/languages/sql/utils.ts
@@ -17,6 +17,8 @@ import {
   DuckDBDialect,
 } from "@marimo-team/codemirror-sql/dialects";
 import type { DataSourceConnection } from "@/core/kernel/messages";
+import { logNever } from "@/utils/assertNever";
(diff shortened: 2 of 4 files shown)


frontend/src/components/datasources/utils.ts
```diff
@@ -9,6 +9,46 @@ export function isSchemaless(schemaName: string) {
...
+  // TimescaleDB: Wrap each part of qualified name with double quotes
+  timescaledb: {
+    formatTableName: (tableName: string) => {
+      const parts = tableName.split(".");
+      return parts.map((part) => `"${part}"`).join(".");
+    },
+    formatSelectClause: defaultFormatter.formatSelectClause,
+  },
+  default: defaultFormatter,
+};
```
mscolnick: i wonder if we can instead use `guessDialect` and the information from there for quoting

frontend/src/components/datasources/utils.ts
```diff
@@ -39,11 +101,18 @@ export function sqlCode({
...
+      formattedTableName,
+    );
+
     if (engine === DUCKDB_ENGINE) {
-      return `_df = mo.sql(f"SELECT ${columnName} FROM ${tableName} LIMIT 100")`;
+      return `_df = mo.sql(f"""${selectClause}""")`;
     }
 
-    return `_df = mo.sql(f"SELECT ${columnName} FROM ${tableName} LIMIT 100", engine=${engine})`;
+    return `_df = mo.sql(f"""${selectClause}""", engine=${engine})`;
```
mscolnick: maybe new lines around the query in case it ends with a "

frontend/src/core/codemirror/language/languages/sql/utils.ts
```diff
@@ -18,13 +18,49 @@ import {
...
  * Guess the CodeMirror SQL dialect from the backend connection dialect.
+ * If unknown, return the standard SQL dialect.
  */
 export function guessDialect(
   connection: Pick<DataSourceConnection, "dialect">,
-): SQLDialect | undefined {
-  switch (connection.dialect) {
+): SQLDialect {
+  const dialect = connection.dialect;
+  if (!isKnownDialect(dialect)) {
```
mscolnick, line 61: would it better to just fall though the switch-case so this does not get out of sync?

mscolnick (commented):
do we need the early returns for knowndialect? It seems like that could be error prone if we forget to add a case to the switch

Light2Dark:
> do we need the early returns for knowndialect? It seems like that could be error prone if we forget to add a case to the switch

I can add a logNever and force us to return something. The reason to return early is to ensure the dialect is type-safe, so we must catch that dialect in the switch statements. It will throw error if spelling is off etc.

frontend/src/components/datasources/utils.ts
Light2Dark: thanks for this, it only works for bigquery as some dialects from codemirror don't have identifierQuotes.

PR #6922, opened 2025-10-24
feat: enable passing in token password from stdin

### LINKED ISSUE #6833: Don't expose token on command line
### Description

Running

```
marimo edit --token-password xyz
```

on a shared host exposes it to other users as command line is visible via `ps -ef`

### Suggested solution

Allow this to be specified as an environment variable.

In general, it probably makes sense to allow any command line parameter to be defaulted using either environment variables or marimo.toml.

- [ ] Yes

### Alternatives

1, Allow for a custom token provider that will return it on stdout:

token-provider.sh

```
echo my-token
````

then

```
marimo --token-provider-stdout token-provider.sh
```

2. Allow for a pluggable python module to provide the parameter overrides, along the lines of (though likely a typed `MarimoConfig` object in the similar fashion would be better);

marimo-config-provider.py

```
def marimo_config_provide():
  return {
    'token-password': 'my-token',
  }
```

then

```
marimo --config-provider marimo-config-provider.py
```

### Additional context

_No response_

dmadisetti:
## 📝 Summary

closes #6833

Enables passing in a token from a pipe.

`cat my_pass.txt | marimo edit --token-password-stdin`

Notably, you can just run

`marimo edit --token-password-stdin`

and type a password in too

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -568,7 +568,11 @@ def edit(
     from marimo._cli.sandbox import run_in_sandbox, should_run_in_sandbox
 
     # We support unix-style piping, e.g. cat notebook.py | marimo edit
-    if name is None and (stdin_contents := _get_stdin_contents()) is not None:
+    if (
+        not token_password_stdin
+        and name is None
+        and (stdin_contents := _get_stdin_contents()) is not None
+    ):
         temp_dir = tempfile.TemporaryDirectory()
         path = create_temp_notebook_file(
             "notebook.py", "py", stdin_contents, temp_dir


mscolnick:
@dmadisetti, the request was specifically `--token-provider-stdout`. why is stdin better? `cat my_pass.txt | marimo edit` is already used for piping in a notebook (i think)

dmadisetti:
stdin is the correct flag name- since it's the source. Nothing needs to be piped, you can type in the value yourself if you want.

I did confirm that `cat my_pass.txt | marimo edit` is possible- so maybe we want to think about handling that. Not sure how widely used it is in practice. It seems to me like the behavior is pretty explicit

mscolnick:
in your implementation `stdin` is the correct flag, but my question is why is that better than `--token-provider-stdout prints-token.sh` (as the originally issue requested)?

PR #6939, opened 2025-10-27
Document `AppMeta` and make it public

koaning:
This PR was made apon reflecting [this message on Discord](https://discord.com/channels/1059888774789730424/1179936876103225354/1432299089873211443). 

The user in this case was trying to figure out whether or not there was a present mode whenever you access the app meta information. The documentation doesn't actually confirm if this is the case. It doesn't list all the values that the run mode can represent. 

Hence, I figured adding the `AppMeta` class to our documentation. That way a user could at least confirm themselves if it is/not supported.

### DIFF
--- a/tests/snapshots/api.txt
+++ b/tests/snapshots/api.txt
@@ -1,4 +1,5 @@
 App
+AppMeta
 Cell
 Html
 MarimoIslandGenerator


docs/api/app.md
```diff
@@ -9,3 +9,5 @@
 ## AppMeta
 
 ::: marimo.app_meta
+
+::: marimo._runtime.app_meta.AppMeta
```
mscolnick: we don't normally document private classes, but since this is the return type of a public function, we should likely make this class public as well

mscolnick: this is what is looks like

image

koaning:
@akshayka I want to make this public before merging. As @mscolnick points out, it's ugly to document something from a hidden module.

koaning:
@mscolnick much better now, assuming we're cool with making it public like this. 

CleanShot 2025-10-27 at 17 41 56

mscolnick:
you will need to run `hatch run +py=3.12 test-optional:test tests/test_api.py::test_api` to update the snapshot of public APIs

koaning:
@mscolnick should be all good now.

PR #6950, opened 2025-10-27
fix: set initial mode for islands

mscolnick:
Islands started to break without setting the initial app mode.


dmadisetti (approved):
Ah, makes sense

PR #6974, opened 2025-10-28
feat: enable passing in token password from file

### LINKED ISSUE #6833: Don't expose token on command line

dmadisetti:
## 📝 Summary

Supercede's #6922 + fixes #6833

---

Allows the specification of a token password file:

```bash
marimo edit --token-password-file my_pass.txt
```

Notably also enables passing in a token from a pipe.

```bash
cat my_pass.txt | marimo edit --token-password-file -
```

Notably, you can just run

```bash
marimo edit --token-password-file -
```

and type a password in too


marimo/_cli/utils.py
```diff
@@ -28,3 +32,171 @@ def prompt_to_overwrite(path: Path) -> bool:
         )
 
     return True
+
+
+def resolve_token(
+    token: bool,
+    token_password: Optional[str],
```
mscolnick, line 40: nit: you could do `resolve_token(*, ..)` so you have to pass kwargs and don't mix up `token_password` and `token_password_file`

mscolnick (approved):
nice implementation and cleanup!

akshayka (dismissed):

marimo/_cli/utils.py
dmadisetti: Good comment

PR #7005, opened 2025-10-30
fix: simplify markdown

dmadisetti:
## 📝 Summary

Simplifies markdown logic from separating multiline and single line cases to always using multiline. This mitigates a breaking bug in 0.17.3 where long single lines had a spurious `"` resulting in an unparsable notebook.


mscolnick:
i'd prefer to not do this (led to previous bug) and instead keep this logic simple

mscolnick:
this changes (always choosing multi-line) is better now. I think it makes sense since not to have nuanced logic when it would go from single-line to multi-line. and as a result, no gotchas to the user for when that changes for various reasons like:
- a new line char
- ending or stating with a `"`
- over the line length

PR #7028, opened 2025-11-01
fix: safe annotation quotes

### LINKED ISSUE #7026: Marimo ast parser got confused parsing something like: `Final[Literal['int8']]` resulting in SyntaxError
### Describe the bug

I got this when annotating variables within marimo

```
File "...\...\bug_marimo_ast_parser.py", line 30
    DTYPE_FAULT: "Final[Literal["int8"]]",
                 ^^^^^^^^^^^^^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?
```

How did i get into this

- I was using edit mode on a directory to create notebooks, one of them contains annotation shown above
- All is good, until i close the notebook and reopen it.
- However, manual editing on the notebook script removes the bug

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
marimo v0.17.5
```

</details>

### Code to reproduce

Initial code:

```python
import marimo

__generated_with = "0.17.5"
app = marimo.App(width="columns", sql_output="polars")


@app.cell
def _():
    import marimo as mo
    return


@app.cell(hide_code=True)
def _():
    from   typing  import Final, Literal
    return Final, Literal


@app.cell
def _(Final, Literal):
    CHUNKS_SIZE: Final[tuple[int, int, int]] = (128, 128, 128)
    DTYPE_FAULT: Final[Literal['int8']] = "int8"
    DTYPE_SEIS : Final[Literal['float16']] = "float16"
    return CHUNKS_SIZE, DTYPE_FAULT, DTYPE_SEIS


if __name__ == "__main__":
    app.run()

```

Then I edit above notebook so it'd be like

```python

import marimo

__generated_with = "0.17.5"
app = marimo.App(width="columns", sql_output="polars")


@app.cell
def _():
    import marimo as mo
    return


@app.cell(hide_code=True)
def _():
    from   typing  import Final, Literal
    return Final, Literal


@app.cell
def _(Final, Literal):
    CHUNKS_SIZE: Final[tuple[int, int, int]] = (128, 128, 128)
    DTYPE_FAULT: Final[Literal['int8']] = "int8"
    DTYPE_SEIS : Final[Literal['float16']] = "float16"
    return CHUNKS_SIZE, DTYPE_FAULT, DTYPE_SEIS


@app.cell
def _(
    CHUNKS_SIZE: "Final[tuple[int, int, int]]",
    DTYPE_FAULT: "Final[Literal["int8"]]",
    DTYPE_SEIS: "Final[Literal["float16"]]",
):
    print(CHUNKS_SIZE, DTYPE_FAULT, DTYPE_SEIS)
    return


@app.cell
def _():
    return


if __name__ == "__main__":
    app.run()

```

dmadisetti:
## 📝 Summary

fixed #7026

If annotations without top-level resolution were quoted and containing quotes themselves, syntax breaking serialization occurred rendering a notebook un-openable.

This fixes this particular issue and additionally adds a defense guard such that invalid code should never be serialized.

### DIFF
--- a/frontend/src/components/data-table/column-summary/legacy-chart-spec.ts
+++ b/frontend/src/components/data-table/column-summary/legacy-chart-spec.ts
@@ -285,9 +285,9 @@ export function getDataSpecAndSourceName<T>(data: string | T[]): {
 
       // eslint-disable-next-line unicorn/prefer-ternary
       if (decoded.startsWith(ARROW_MAGIC_NUMBER)) {
-        // @ts-expect-error vega-typings does not include arrow format
         dataSpec = {
           values: byteStringToBinary(decoded),
+          // @ts-expect-error vega-typings does not include arrow format
           format: { type: "arrow" },
         };
       } else {


mscolnick (dismissed):

akshayka (dismissed):
LGTM. But does it make sense to add a unit test for the unparsable code path?

dmadisetti:
@akshayka I added a failure case by mocking. There's still an unhappy path where `unparsable` itself causes an issue- but at least we'll log an actionable error if that occurs

dmadisetti:
Think CI will be fixed by https://github.com/marimo-team/marimo/pull/7034

akshayka (approved):

PR #7053, opened 2025-11-03
feat: add Weights & Biases as an AI provider

mscolnick:
This adds Weights & Biases as an AI provider.

Screenshot 2025-11-03 at 4 04 01 PM
Screenshot 2025-11-03 at 4 03 54 PM

### DIFF
--- a/packages/llm-info/data/models.yml
+++ b/packages/llm-info/data/models.yml
@@ -422,56 +422,56 @@
 # Weights & Biases Models
 
 - name: Meta Llama 3.1 8B (W&B)
-  model: wandb/meta-llama/Llama-3.1-8B-Instruct
+  model: meta-llama/Llama-3.1-8B-Instruct
   description: Efficient conversational model optimized for responsive multilingual chatbot interactions
   providers: [wandb]
   roles: [chat, edit]
   thinking: false
 
 - name: OpenAI GPT OSS 120B
-  model: wandb/openai/gpt-oss-120b
+  model: openai/gpt-oss-120b
   description: Efficient Mixture-of-Experts model designed for high-reasoning, agentic and general-purpose use cases
   providers: [wandb]
   roles: [chat, edit]
   thinking: true
 
 - name: OpenAI GPT OSS 20B
-  model: wandb/openai/gpt-oss-20b
+  model: openai/gpt-oss-20b
   description: Lower latency Mixture-of-Experts model trained on OpenAI's Harmony response format with reasoning capabilities
   providers: [wandb]
   roles: [chat, edit]
   thinking: true
 
 - name: DeepSeek V3-0324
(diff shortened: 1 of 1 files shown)


akshayka (dismissed):
Docs LGTM, have not reviewed TS

packages/openapi/api.yaml
```diff
@@ -86,17 +86,41 @@ components:
...
-        \ the maximum number of tokens to use in AI completions\n    - `mode`: the\
-        \ mode to use for AI completions. Can be one of: `\"ask\"` or `\"manual\"\
-        `\n    - `inline_tooltip`: if `True`, enable inline AI tooltip suggestions\n\
-        \    - `models`: the models to use for AI completions\n    - `open_ai`: the\
-        \ OpenAI config\n    - `anthropic`: the Anthropic config\n    - `google`:\
-        \ the Google AI config\n    - `bedrock`: the Bedrock config\n    - `azure`:\
-        \ the Azure config\n    - `ollama`: the Ollama config\n    - `github`: the\
-        \ GitHub config\n    - `openrouter`: the OpenRouter config\n    - `open_ai_compatible`:\
-        \ the OpenAI-compatible config"
+      description: 'Configuration options for AI.
```
dmadisetti: @mscolnick downgrading to python 3.12 fixes this. Not a package issue

dmadisetti (commented):
Ran into the same schema issue. It's a python version difference. I'll open a PR to fix on 3.13, but in the meantime, you can just run with 3.12 if you want to minimize the diff

packages/openapi/api.yaml
mscolnick: sounds good, @dmadisetti i just pinned it to 3.12. no need to fix for 3.13 right now

Makefile
```diff
@@ -106,7 +106,7 @@ fe-typecheck:
 .PHONY: fe-codegen
 # 🔄 Generate frontend API
 fe-codegen:
-	uv run ./marimo development openapi > packages/openapi/api.yaml
+	uv run --python=3.12 ./marimo development openapi > packages/openapi/api.yaml
```
dmadisetti, line 109: fair

dmadisetti (dismissed):

packages/llm-info/data/models.yml
```diff
@@ -419,3 +419,61 @@
   roles: [chat]
   thinking: false
 
+# Weights & Biases Models
+
+- name: Meta Llama 3.1 8B (W&B)
+  model: wandb/meta-llama/Llama-3.1-8B-Instruct
```
Light2Dark: The model names should not have the `wandb` prefix, since it's already in the provider.

Else you end up with wandb/wandb/model_name

mscolnick: yep, was getting that error, thanks

PR #7071, opened 2025-11-04
Revert "don't override auto-save by when in --watch mode"

mscolnick:
Reverts marimo-team/marimo#6535

While it would be nice to not override auto-save. User's are still hitting conflicts when both are enabled. Until we find a better solution, we should not let them both be enabled.

### DIFF
--- a/marimo/_server/start.py
+++ b/marimo/_server/start.py
@@ -220,7 +220,7 @@ def start(
                 }
             }
         )
-        LOGGER.info("Watch mode enabled, auto-save is disabled")
+        LOGGER.warning("Watch mode enabled, auto-save is disabled")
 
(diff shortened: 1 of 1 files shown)


dmadisetti (dismissed):
info or warning?

marimo/_server/start.py
```diff
@@ -208,8 +208,19 @@ def start(
...
+        config_reader = config_reader.with_overrides(
+            {
+                "save": {
+                    "autosave": "off",
+                    "format_on_save": False,
+                    "autosave_delay": 1000,
+                }
+            }
+        )
+        LOGGER.info("Watch mode enabled, auto-save is disabled")
```
mscolnick:
```suggestion
        LOGGER.warning("Watch mode enabled, auto-save is disabled")
```

PR #7074, opened 2025-11-04
chore: consolidate hash_code, set usedforsecurity=false

mscolnick:
Consolidate logic used to create the code hash


manzt (dismissed):

marimo/_utils/code.py
```diff
@@ -5,4 +5,6 @@
 def hash_code(code: str) -> str:
     import hashlib
 
-    return hashlib.sha256(code.encode("utf-8")).hexdigest()
+    return hashlib.sha256(
```
dmadisetti:
```suggestion
    return hashlib.md5(
```

dmadisetti (dismissed):

PR #7075, opened 2025-11-04
fix: change typing for code_hash

dmadisetti:
## 📝 Summary


manzt (approved):

dmadisetti:
May cause more issues that it's worth

PR #7077, opened 2025-11-04
fix: don't sanitize when auto-instantiate is on

mscolnick:
If a user wants to have auto-run, we should also allow the JS to auto (i.e. no sanitizing). this can be configured by user config or pyproject.toml


mscolnick:
@dmadisetti , updated the docs, lmk what you think

dmadisetti (approved):

PR #7087, opened 2025-11-05
improvement: inline local images (virutal files) in the HTML export

mscolnick:
This traversed `<img src` elements and replaces virtual files `./@file/foo.png` with the file and data-uri, so these HTML files can be more self-contained

### DIFF
--- a/marimo/_server/export/exporter.py
+++ b/marimo/_server/export/exporter.py
@@ -150,7 +150,7 @@ class Exporter:
         # - display.cell_output
         config = deep_copy(DEFAULT_CONFIG)
         config["display"] = display_config
-        return config
+        return cast(MarimoConfig, config)
 
     def _inline_virtual_files(
         self, session_snapshot: NotebookSessionV1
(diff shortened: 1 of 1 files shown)


marimo/_runtime/virtual_file.py
```diff
@@ -221,6 +221,9 @@ def add(self, virtual_file: VirtualFile, context: RuntimeContext) -> None:
             return
 
         buffer = virtual_file.buffer
+        # Skip adding if buffer is empty
+        if len(buffer) == 0:
+            return
```
mscolnick, line 226: previous bug that would log an unnecessary error

marimo/_server/export/dom_traversal.py
```diff
@@ -0,0 +1,321 @@
...
+
+from marimo import _loggers
+from marimo._messaging.mimetypes import KnownMimeType
+from marimo._runtime.virtual_file import read_virtual_file
+from marimo._utils.data_uri import build_data_url
+
+LOGGER = _loggers.marimo_logger()
+
+
+class _HTMLAttributeReplacer(HTMLParser):
```
dmadisetti, line 18: Surprised you had to roll your own here

marimo/_server/export/exporter.py
```diff
@@ -68,8 +73,15 @@
 if TYPE_CHECKING:
     from nbformat.notebooknode import NotebookNode  # type: ignore
 
+VIRTUAL_FILE_ALLOWED_ATTRIBUTES = {"src"}
+VIRTUAL_FILE_ALLOWED_TAGS = {"img"}
```
dmadisetti, line 79: A previous version of this allowed video / audio tags as well. Is that not worth it too large? Might be worth putting a comment at least

mscolnick: yea i didn't want to shove those in here. i can add a comment

dmadisetti (dismissed):
It would be nice if we could delegate some of the sanitization stuff- but this is an OK stop gap otherwise

PR #7093, opened 2025-11-06
improvement: impove default DPI on matplotlib images

### LINKED ISSUE #3793: No automatic retina mode for matplotlib
### Describe the bug

When I create a plot with mathplotlib, the resolution of the resulting image is noticeably worse than the code and text around it. In Jupyter notebook, there is retina mode that up resolutions charts when on a high DPI monitor.

I could not find documentation on how to do this in Marimo. Obviously it would be nice if importing matplotlib made whatever changes need to be done happen automatically.

Image

### Environment

<details>
{
  "marimo": "0.11.4",
  "OS": "Darwin",
  "OS Version": "24.1.0",
  "Processor": "arm",
  "Python Version": "3.13.1",
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.26.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "ruff": "0.9.6",
    "starlette": "0.45.3",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "14.2"
  },
  "Optional Dependencies": {
    "pandas": "2.2.3"
  },
  "Experimental Flags": {}
}
</details>

### Code to reproduce

```
import marimo as mo
import pandas as pd
import numpy as np


x = np.random.rand(27)
y = 20 - np.linspace(2,20,27) * x
df = pd.DataFrame({'x':x, 'y':y})
plt.scatter(df.x, df.y)
```

mscolnick:
Fixes #3793

Implement automatic retina/HiDPI rendering for matplotlib plots to fix blurry images on high-resolution displays. This change makes marimo compatible with Jupyter's retina mode, rendering plots at 2x DPI with proper metadata for crisp display.

### DIFF
--- a/tests/_output/formatters/test_structures.py
+++ b/tests/_output/formatters/test_structures.py
@@ -49,7 +49,11 @@ async def test_matplotlib_special_case(
         formatter = executing_kernel.globals["formatter"]
         lines = executing_kernel.globals["lines"]
         assert formatter is not None
-        assert formatter(lines)[0].startswith("image")
+        mimetype = formatter(lines)[0]
+        assert (
+            mimetype.startswith("image")
+            or mimetype == "application/vnd.marimo+mimebundle"
+        )
 
 
 def test_format_structure_types() -> None:


marimo/_output/formatters/matplotlib_formatters.py
```diff
@@ -27,22 +27,69 @@ def register(self) -> None:
...
+            return width, height
+
         def mime_data_artist(artist: Artist) -> tuple[KnownMimeType, str]:
             buf = io.BytesIO()
-            artist.figure.savefig(buf, format="png", bbox_inches="tight")  # type: ignore
+            fig = artist.figure  # type: ignore
+
+            # Get current DPI and double it for retina display
+            original_dpi = fig.dpi
+            retina_dpi = original_dpi * 2
```
akshayka: Does this need to depend on whether the screen is a retina display?

mscolnick: we could, but Jupyter does not currently either. 

**Pros**: this is much simpler 
**Cons**: the downside is that we will send double-size kbs for not retina displays. it will be a better quality image, but i think it will look the same to the user. if the user downloads the image, it will still be better quality (twice the size) if shared

akshayka: If Jupyter doesn't either then this is fine, thanks for checking

PR #7094, opened 2025-11-06
fix: better virtual file matching

mscolnick:
This regex was a bit greedier and matched strings it should not have. This updates the config and updates the tests as well.

### DIFF
--- a/frontend/src/core/static/__tests__/virtual-file-tracker.test.ts
+++ b/frontend/src/core/static/__tests__/virtual-file-tracker.test.ts
@@ -65,8 +65,8 @@ describe("findVirtualFiles", () => {
   });
 
   it("should not match virtual files with mismatched quotes", () => {
-    const input = "'@file/938947-3587525-4XcHZISt.csv\"'";
-    const expected = new Set();
+    const input = "'/@file/938947-3587525-4XcHZISt.csv\"'";
+    const expected = new Set(["/@file/938947-3587525-4XcHZISt.csv"]);
     const result = findVirtualFiles(input);
     expect(result).toEqual(expected);
   });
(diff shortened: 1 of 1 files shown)


frontend/src/core/static/__tests__/virtual-file-tracker.test.ts
```diff
@@ -56,6 +56,41 @@ describe("findVirtualFiles", () => {
...
+
+  it("should match virtual files with HTML entity endings", () => {
+    const input = '"/@file/938947-3587525-4XcHZISt.csv&quot"';
+    const expected = new Set(["/@file/938947-3587525-4XcHZISt.csv"]);
+    const result = findVirtualFiles(input);
+    expect(result).toEqual(expected);
+  });
+
+  it("should not match virtual files with mismatched quotes", () => {
+    const input = "'@file/938947-3587525-4XcHZISt.csv\"'";
```
dmadisetti: This doesn't actually catch the behavior you want. It's because you don't have a **/**@file this fails

frontend/src/core/static/virtual-file-tracker.ts
```diff
@@ -4,7 +4,7 @@ import type { CellId } from "../cells/ids";
 import type { CellMessage } from "../kernel/messages";
 
 // Virtual files are of the form /@file/<file-name>.<extension>
-const VIRTUAL_FILE_REGEX = /\/@file\/([^\s/]+)\.([\dA-Za-z]+)/g;
+const VIRTUAL_FILE_REGEX = /\/@file\/([^\s/"'&]+)\.([\dA-Za-z]+)/g;
```
dmadisetti, line 7: None of the tests trigger this behavior, this invalidates cases like:

`/@file/file".csv`

Which used to match: https://regex101.com/r/xcC6Zx/1

mscolnick: i added a test, it works as expected (it does not pick up `/@file/file".csv`)

```
  it("should not match virtual files with quotes in the filename", () => {
    const input = '/@file/file".csv';
    const expected = new Set();
    const result = findVirtualFiles(input);
    expect(result).toEqual(expected);
  });
```

dmadisetti: I think that test is the only one that gets the new functionality. Other tests should pass without this change. Was there a particular pattern that we were trying to avoid?

mscolnick: yes, it was `/@file/file.csv&shouldnotcontinuematch` and `/@file/file.csv"shouldnotcontinuematch`

PR #7099, opened 2025-11-06
fix: track contents to avoid reloading on our own writes

mscolnick:
This PR track contents that we wrong to avoid reloading on our own writes when file watching.

### DIFF
--- a/marimo/_server/file_manager.py
+++ b/marimo/_server/file_manager.py
@@ -47,6 +47,8 @@ class AppFileManager:
         )
         self._default_sql_output: SqlOutputType | None = default_sql_output
         self.app = self._load_app(self.path)
+        # Track the last saved content to avoid reloading our own writes
+        self._last_saved_content: Optional[str] = None
 
     @staticmethod
     def from_app(app: InternalApp) -> AppFileManager:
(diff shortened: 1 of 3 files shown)


marimo/_server/file_manager.py
```diff
@@ -180,6 +183,8 @@ def _save_file(
 
         if persist:
             self._create_file(filename, contents)
+            # Record the last saved content to avoid reloading our own writes
+            self._last_saved_content = contents.strip()
```
dmadisetti, line 187: Could save just the hash

tests/_server/test_file_manager.py
```diff
@@ -793,3 +793,63 @@ def cell2():
...
+            filename=str(temp_file),
+            codes=["x = 1"],
+            names=["cell1"],
+            configs=[CellConfig()],
+            persist=True,
+        )
+    )
+
+    # Now the file should match the last save
+    assert manager.file_content_matches_last_save() is True
```
dmadisetti, line 836: Can you add another call mocking debug to make sure that we are skipping? Or maybe mock reload to make sure never called?

marimo/_server/file_manager.py
mscolnick: it saves a bit on space (not much), but requires hashing each time

tests/_server/test_file_manager.py
mscolnick: i have this already in a followup PR, will push it up in a bit

Issue #7060, opened 2025-11-04
ui_element_registry.resolve_lens throws error in notebook execution


bmerkle:
### Describe the bug

I got the following stacktrace when executing my marimo notebook;
I can add more details if necessary

        🧪 Experimental features (use with caution): rtc_v2
An exception was raised when finding cells that refer to a UIElement value, for bound name 7bd61d0f-8657-400f-95a6-dab93c3f67a9. This is a bug in marimo. Please copy the below traceback and paste it in an issue: https://github.com/marimo-team/marimo/issues
Traceback (most recent call last):
  File "C:\Users\merklbe1\AppData\Local\uv\cache\archive-v0\cNDvuhYL_pW_yOtejlimd\Lib\site-packages\marimo\_runtime\runtime.py", line 1783, in set_ui_element_value
    resolved_id, resolved_value = ui_element_registry.resolve_lens(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\merklbe1\AppData\Local\uv\cache\archive-v0\cNDvuhYL_pW_yOtejlimd\Lib\site-packages\marimo\_plugins\ui\_core\registry.py", line 149, in resolve_lens
    raise KeyError(f"UIElement with id {object_id} not found")
KeyError: 'UIElement with id de794bb6-d68d-4759-bff9-abf8eb9a2301MJUe-0 not found'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\merklbe1\AppData\Local\uv\cache\archive-v0\cNDvuhYL_pW_yOtejlimd\Lib\site-packages\marimo\_runtime\runtime.py", line 1940, in update_stateful_values
    ) - self.graph.get_defining_cells(name)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\merklbe1\AppData\Local\uv\cache\archive-v0\cNDvuhYL_pW_yOtejlimd\Lib\site-packages\marimo\_runtime\dataflow.py", line 86, in get_defining_cells
    return self.definitions[name]
           ~~~~~~~~~~~~~~~~^^^^^^
KeyError: '7bd61d0f-8657-400f-95a6-dab93c3f67a9'



### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.17.6",
  "editable": false,
[...]

PR #7102, opened 2025-11-06
improve: do not trigger UI element warnings for state updates

akshayka:
Right now the runtime eagerly updates states value when a state update
is registered. This should be unnecessary (but maybe was needed
for file watching element, which is when this code was introduced),
and it hits a code path for UI elements that triggers a very loud error.
Instead of fixing the underlying issue, this PR prevents the
code path from being triggered and prints a gentler warning.

Related: #7060

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -1470,6 +1470,16 @@ class Kernel:
         cell_id = ctx.execution_context.cell_id
         with self._state_lock:
             self.state_updates[state] = cell_id
+        # TODO(akshayka): State should not be updated eagerly like this,
+        # the runner already handles state execution. This is a hack
+        # until we figure out how to clean this up.
+        names = ctx.state_registry.bound_names(state)
+        if len(names) == 1 and not (name := list(names)[0]).isidentifier():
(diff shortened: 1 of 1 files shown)


marimo/_runtime/runtime.py
```diff
@@ -1470,6 +1470,16 @@ def register_state_update(self, state: State[Any]) -> None:
         cell_id = ctx.execution_context.cell_id
         with self._state_lock:
             self.state_updates[state] = cell_id
+        # TODO(akshayka): State should not be updated eagerly like this,
```
dmadisetti, line 1473: Context, this path is hit when you update state from the same cell:

Image

because state does not have a bound name it cannot be looked up, which means that some of this logic in the function doesn't apply- so early exiting is correct.

PR https://github.com/marimo-team/marimo/pull/5143
Also provided an early exit (but this is better since we update state_updates)

There's also an embedded app path where I think state is associated with the parent cell.

dmadisetti: Current behavior: 
image

akshayka: Yep, thank you!

PR #7103, opened 2025-11-06
fix: debouncing on mo.watch

dmadisetti:
## 📝 Summary

In the case of file watching, self cell reference should not be enabled since state updates are triggered via a poll watcher. Additionally solves another debouncing issue


akshayka (dismissed):

dmadisetti:
Closing this out since state fix seems to have stablized this behavior

PR #7104, opened 2025-11-06
fix: sandbox requirements from 'marimo edit <url>'

mscolnick:
Previously we would get warnings like: "Cannot read sandbox deps for file https://..". 

For remote files, we download this into a `/tmp` dir. This change moves the sandbox check until after downloading that so that we can properly read the sandbox details.

### DIFF
--- a/marimo/_server/api/endpoints/ai.py
+++ b/marimo/_server/api/endpoints/ai.py
@@ -399,8 +399,7 @@ async def mcp_status(
                     schema:
                         $ref: "#/components/schemas/MCPStatusResponse"
     """
-    app_state = AppState(request)
-
+    del request
     try:
         # Try to get MCP client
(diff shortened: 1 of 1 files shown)


marimo/_server/api/endpoints/ai.py
```diff
@@ -400,7 +400,6 @@ async def mcp_status(
                         $ref: "#/components/schemas/MCPStatusResponse"
     """
     app_state = AppState(request)
-    app_state.require_current_session()
```
dmadisetti, line 403: Did this just error out or unrelated change?

mscolnick: this was an error - was a drive-by cleanup

dmadisetti (approved):

PR #7141, opened 2025-11-10
feat: Add !pip metadata from ipynb convert and handle ! commands

dmadisetti:
## 📝 Summary

`! comand in jupyter`

is a valid method of running a subprocess. We now explicitly handle this in conversion to use `subprocess.call`. In explicit cases of `!pip` install, we leave a comment instead, and add the pip install to package metadata

### DIFF
--- a/marimo/_convert/ipynb.py
+++ b/marimo/_convert/ipynb.py
@@ -143,8 +143,6 @@ def transform_add_subprocess_import(
             stripped = line.strip()
             if stripped.startswith("import subprocess"):
                 return True
-            if stripped.startswith("from subprocess"):
-                return True
             return False
 
         # Slow check


marimo/_convert/ipynb.py
```diff
@@ -124,6 +124,45 @@ def is_in_import_line(line: str) -> bool:
...
+    def has_subprocess_import(cell: str) -> bool:
+        # Quick check
+        if "import subprocess" not in cell:
+            return False
+
+        def is_in_import_line(line: str) -> bool:
+            stripped = line.strip()
+            if stripped.startswith("import subprocess"):
+                return True
+            if stripped.startswith("from subprocess"):
```
mscolnick: this won't matter right? we would still want to `import subprocess` in this case

marimo/_convert/ipynb.py
```diff
@@ -348,20 +387,199 @@ def transform(cell: str) -> str:
...
+
+    Returns: (replacement_text, pip_packages, needs_subprocess)
     """
+    import shlex
 
-    def transform(cell: str) -> str:
-        if "!pip" in cell:
-            cell = cell.replace(
-                "!pip",
-                "# (use marimo's built-in package management features instead) !pip",  # noqa: E501
```
mscolnick, line 360: i wonder if we should have just preferred this, to running an actually process. this does not hook into the user's desired package manager

mscolnick: nvm, i see for packages we put in the package metadata

dmadisetti: Figured this was a good compromise, lmk if you disagree though

mscolnick: i think its ok. it will force users to use `uv` over `pip`  in order to use the package metadata, but they can always say No and continue with the auto-install

dmadisetti:
Great, going to follow up with a change in `sandbox workflow`

PR #7146, opened 2025-11-11
feat: app.embed(defs={'mine': value})

dmadisetti:
## 📝 Summary

Enables the override of values into embed, allowing for birectional communication of embedded and parent apps

### DIFF
--- a/tests/_runtime/test_trace.py
+++ b/tests/_runtime/test_trace.py
@@ -302,10 +302,17 @@ class TestEmbedTrace:
                         from tests._runtime.script_data import (
                             script_exception_with_output
                         )
+                    """
+                    ),
+                ),
+                ExecutionRequest(
+                    cell_id="1",
+                    code=textwrap.dedent(
+                        """
                         await script_exception_with_output.app.embed()
                     """
                     ),
-                )
+                ),
             ]
         )
 


marimo/_ast/app.py
```diff
@@ -756,7 +756,11 @@ async def _function_call(
         return await app_kernel_runner.function_call(request)
 
     @mddoc
-    async def embed(self) -> AppEmbedResult:
+    async def embed(
+        self,
+        defs: dict[str, Any] | None = None,
+        **kwargs: Any,
```
akshayka: I am not a fan of including `**kwargs` on all our APIs. It complicates readability and can lead to silent errors (an old version won't complain if an unknown arg is used, but the implementation won't do the right thing). We should introduce new arguments APIs judiciously, and be explicit on requiring users to upgrade when APIs are changed.

dmadisetti: I agree in general but I do have 2 counterpoints for this specific case:

a) it mirrors `app.run` which is nice from a consistency standpoint
b) a v1 gives us a chance to clean up apis

Checking if kwargs is not null might also address your point on silent errors though (throw a warning, though maybe just better to fail).

marimo/_ast/app.py
```diff
@@ -832,9 +846,16 @@ async def embed(self) -> AppEmbedResult:
             outputs: dict[CellId_t, Any]
             glbls: dict[str, Any]
             if not app_kernel_runner.outputs:
-                outputs, glbls = await app_kernel_runner.run(
-                    set(self._execution_order)
+                # Inject provided defs into the kernel's globals
+                if defs:
```
akshayka, line 874: Today `app.embed()` executes the app once and then serves cached outputs thereafter.

What happens if `app.embed()` is called with one defs relation, then called again with another defs relation? I think as written in this PR the user will get stale results, because `app_kernel_runner.outputs` will be non empty.

So, I think we need to retain some kind of cache that is keyed by a tuple of (key, value) pairs in defs.

tests/_ast/test_app.py
```diff
@@ -1150,6 +1150,108 @@ def __(x: int, y: int) -> None:
...
+
+        # Check first result - output then defs
+        assert "x is small" in result_initial.output.text
+        assert result_initial.defs["x"] == 10
+
+        # Check second result with first override - output then defs
+        assert "x is large" in result_override.output.text
+        assert "x is small" not in result_override.output.text
+        assert result_override.defs["x"] == 100
+
```
dmadisetti, line 1259: @akshayka this current fails. Is this the behavior you were thinking of, or a different issue?

akshayka: Yep, this is an example of the incorrect logic.

https://github.com/marimo-team/marimo/pull/7146#discussion_r2543269564

akshayka: I now remember why I didn't implement interpolation of inputs. It makes the kernel runner logic more complicated.

The kernel runner caches outputs to prevent unnecessary re-runs on UI element interactions, for example. UI element updates flow through the graph, only changing outputs for the affected cells. 

If the user is allowed to change inputs, we need to implement cache invalidation on `AppKernelRunner.outputs`. Totally possible, but just requires some work.

akshayka: @zilto would you be interested in helping carry this PR over the finish line? This comment thread describes most of what needs to be done.

zilto: I just spent 30min exploring the code. 

## Goal
If I understand, I need to figure out how to make the test pass in `test_app`. The dataflow logic was updated (and tests are passing), but the runner code caches things, skipping some dataflow nodes.

## Investigation
Those are the pieces of code that I believe are relevant
- `marimo/_runtime/runtime.py::Kernel._run_cells_internal()` creates the `cell_runner.Runner()` object
- `marimo/_rutime/runner/cell_runner.py::Runner` determines which cells are ran and executes the dataflow
- `marimo/_runtime/app/kernel_runner.py::AppKernelRunner` inside `__init__` defines a `cache_output()` hook for `Kernel` which writes to `AppKernelRunner.outputs`
  - Messing with this hook changes the test results

## Questions
Just to verify my understanding:
- `Runner` can be a `AppKernelRunner` or `ScriptRunner`
- required changes are limited to `AppKernelRunner`
- the `Kernel` hooks are difficult to print / debug because they happen in a separate thread. How can I inspect what's cached

dmadisetti: > Runner can be a AppKernelRunner or ScriptRunner
> required changes are limited to AppKernelRunner

The 2 tests I added capture the script (tests are passing) and kernel runner (this comment thread, tests are failing).
Yes, any changes will only touch the kernel runner, but we have a test in place for script runner regression

> the Kernel hooks are difficult to print / debug because they happen in a separate thread. How can I inspect what's cached

You can actually print by checking `k.stdout`/ `k.stderr`. A common pattern I have used is

```python
assert not k.stdout.messages, k.stdout
assert not k.stderr.messages, k.stderr
```

Which should give you a plain text log of what's happening inside the kernel.

---

There are 2 places for "`caching`":
1. We have a AppKernelRunnerRegistry
We could ensure that calls with embed are always given a unique KernelRunner (no caching issue in that case)

2. On the AppKernelRunner itself for outputs

---

I think AppKernelRunnerRegistry uniquely can address this.

---
Writing this up, I actually just pushed up what I think is a simple solution.

akshayka (changes requested):
@dmadisetti The change is not correct. Take a look at `AppKernelRunnerRegistry.get_runner` and you should see why cloning does not work.

This would also be too expensive. It clones and reexecutes the DAG on every UI interaction. Every UI element interaction reexecutes the cell that runs `app.embed()`, which is why caching is necessary.

Also take a look at this smoke test video for a way in which it does not work.

https://github.com/user-attachments/assets/583f0b3c-e520-48da-9a7d-26a3c2f19926

akshayka:
> My main question would be about performance. I see that `AppKernelRunnerRegistry.get_runner()` uses threading.

The runner doesn't use threads. Instead it is thread-safe, it uses the thread id to disambiguate between kernel runners. But you are right that cloning is not performant so another approach needs to be taken

dmadisetti:
Reverted my cloning change to give you more surface area to work @zilto. It is not correct, thanks for the smoke test @akshayka

akshayka:
@zilto do you have time (and interest) to pick this up? I'd be happy to provide guidance

zilto:
@akshayka I spent some time on it, but didn't make meaningful progress. I'm soon going on vacation and will back in January. If someone wants to pick it up, be my guest! Happy to review / comment

dmadisetti:
@akshayka thanks for getting this over the line- want to push through before commit conflicts crop up

PR #7165, opened 2025-11-13
fix: dont set empty cells as stale

mscolnick:
If you start a new notebook with auto-run off, then the first cell is stale, when it should not be until edited.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -2100,8 +2100,13 @@ class Kernel:
             self.reset_ui_initializers()
         else:
             self._uninstantiated_execution_requests = execution_requests
-            for cell_id in self._uninstantiated_execution_requests.keys():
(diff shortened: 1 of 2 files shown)


marimo/_runtime/runtime.py
```diff
@@ -2100,8 +2100,13 @@ async def instantiate(self, request: CreationRequest) -> None:
...
         else:
             self._uninstantiated_execution_requests = execution_requests
-            for cell_id in self._uninstantiated_execution_requests.keys():
-                CellOp.broadcast_stale(cell_id=cell_id, stale=True)
+            for (
+                cell_id,
+                cell_request,
+            ) in self._uninstantiated_execution_requests.items():
+                # Only mark non-empty cells as stale
+                if cell_request.code.strip():
```
dmadisetti, line 2108: Should we consider cells with only comments too?

mscolnick: i dont think that is necessary

PR #7170, opened 2025-11-13
create markdown cell from prompting 'Generate with AI'

Light2Dark:
## 📝 Summary

Prompts from 'Generate with AI' will create a markdown cell with the prompt.

https://github.com/user-attachments/assets/00ad92f1-746b-46c0-8a49-a23d34d8c508

## 📋 Checklist


frontend/src/components/editor/ai/add-cell-with-ai.tsx
```diff
@@ -450,3 +464,13 @@ export const PromptInput = ({
...
 };
+
+/** Converts a prompt to a markdown cell. Used to display the prompt in the UI. */
+export function prettifyPromptToMarkdown(prompt: string): string {
+  const footer = '<p align="center"><i>Generating with AI...</i></p>';
+  const wrappedPrompt = `\`\`\`md\n${prompt}\n\`\`\`\n${footer}`;
+  const [markdownCode] = LanguageAdapters.markdown.transformOut(wrappedPrompt, {
+    quotePrefix: "f",
+  });
+  return markdownCode;
```
mscolnick: i think just the prompt and no styling is cleaner. something that they can choose to commit to the notebook should they want some history of the prompt.


in future, we may want to add some hidden metadata that _could_ be use to regenerate the cells below

Light2Dark: yeah, I think that's a good start. thanks

mscolnick (approved):

akshayka (changes requested):
I like the idea of archiving the prompt. But I think most users may just end up deleting the markdown cell that is added.

Can we consider the following design alternative: prepend the prompt as a comment in the first cell generated by the assistant? To me this feels simpler and cleaner, since the prompt is more of a descriptive/archival comment of what the code does rather than a substantive comment. Here is an example.

image

In the future this comment could be extended with other metadata should we choose, in a way that would be awkward if it were a markdown cell.

manzt:
Do you think we can close out this PR? I think with marimo-pair / code mode, the concept of 1:1 prompt to cell is less relevant.

akshayka:
Yes, @manzt we can close it. It was a good experiment

PR #7175, opened 2025-11-14
feat: Contextually change stat color

kjgoodrick:
## 📝 Summary

With this PR it possible to manually swap the color of the triangle in the `mo.stat` UI element. Previously the color was always green for increases and always red for decreases, but sometimes a decrease is "good" and one may want to display a green downwards facing triangle. 

## 🔍 Description of Changes

The `stat` function now has a `color` argument that accepts the literals "positive" (green) or "negative" (red) to allow the user to set the color manually. If not specified the color will be as it is now, green for increase and red for decrease. 

## 📋 Checklist

- [X] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable). https://discord.com/channels/1059888774789730424/1059891311190229082/1438631478178353232

### DIFF
--- a/marimo/_plugins/stateless/stat.py
+++ b/marimo/_plugins/stateless/stat.py
@@ -29,12 +29,12 @@ def stat(
         direction: the direction of the statistic,
             either `increase` or `decrease`
         bordered: whether to display a border around the statistic
-        target_direction: the direction of the statistic 
-            corresponding to a positive or desirable outcome. Set to 
-            `increase` when higher values are better, or `decrease` 
+        target_direction: the direction of the statistic
+            corresponding to a positive or desirable outcome. Set to
+            `increase` when higher values are better, or `decrease`
             when lower values are better. By default the target
             direction is `increase`.
-            
+
 
     Returns:
         An `Html` object representing the statistic.


Light2Dark:
Thanks! Feel free to disagree, would it be more intuitive if the API is just
`reverse_color=True` or `invert_color`?

kjgoodrick:
Thanks for suggesting that, I do think it’s less confusing to use `reverse_color`. I will update the PR.

Light2Dark:
@mscolnick, fyi on the API change.

Light2Dark:
thank you! using var will raise pre-commit issues, we should use let / const instead. this snippet should help clean it up.
```typescript
  const fillColors = {
    increase: "var(--grass-8)",
    decrease: "var(--red-8)",
  };

  const strokeColors = {
    increase: "var(--grass-9)",
    decrease: "var(--red-9)",
  };

...
             <TriangleIcon
                className="w-4 h-4 mr-1 p-0.5"
                fill={reverseColor ? fillColors.decrease : fillColors.increase}
                stroke={
                  reverseColor ? strokeColors.decrease : strokeColors.increase
                }
              />
```

marimo/_plugins/stateless/stat.py
```diff
@@ -16,6 +16,7 @@ def stat(
     caption: Optional[str] = None,
     direction: Optional[Literal["increase", "decrease"]] = None,
     bordered: bool = False,
+    reverse_color: bool = False,
```
akshayka: This API is too imperative. We should prefer something more semantic or declarative.

For example, `goal: Literal["increase", "decrease"]`,

```python
mo.stat(value=10, direction="increase", goal="increase")
```

which leaves room for the "goal" to affect other behaviors in the future instead of just reversing  a color. This is more legible than `reverse_color=True` (Reverse color of what? Why do I want to reverse colors).

kjgoodrick: Thanks for this suggestion, after reading it I totally agree. I have updated the PR changing the parameter name to `target_direction` to make it clear that the target is associated with the direction. To keep the current behavior I set the default target to `"increase"`. This has also simplified the implementation, so great comment.

marimo/_plugins/stateless/stat.py
```diff
@@ -28,6 +29,8 @@ def stat(
         direction: the direction of the statistic,
             either `increase` or `decrease`
         bordered: whether to display a border around the statistic
+        target_direction: The target direction of the statistic,
+            either `increase` or `decrease`
```
akshayka: The comment for `target_direction` could benefit from being more verbose. Without more context it is difficult to understand the difference between `direction` and `target_direction`

akshayka (commented):
Thanks, this is better! One comment

PR #7178, opened 2025-11-14
chore: change 12px to text-xs for better compatible with vscode and other theming

mscolnick:
Rather than using 12px directly, we use the class name `text-xs` which allows for overriding the class


akshayka (dismissed):

mscolnick:
~i did this wrong. need to change text-sm to text-xs~ fixed

PR #7187, opened 2025-11-16
Add streaming support for chat: async generator bug and add documentation

adamwdraper:
Kapture 2025-11-18 at 09 08 50

## Summary

This PR adds real-time streaming support to `mo.ui.chat`, enabling ChatGPT-like experiences where responses appear word-by-word as they're generated.

### ✨ Streaming Support for Chat Models

Users can now stream responses from chat models in real-time:

**With built-in models** (OpenAI, etc.):
```python
chat = mo.ui.chat(
    mo.ai.llm.openai(
        "gpt-4o",
        stream=True,  # Enable streaming! Uses sync generator internally
    )
)
```

**With custom models** using sync generators:
```python
# Custom models can use EITHER sync or async generators
def my_streaming_model(messages, config):
    yield accumulated  # Sync generator - works!

async def my_async_model(messages, config):
    yield accumulated  # Async generator - also works!
```

### 📚 Documentation

- Added comprehensive "Streaming Responses" section to chat API docs
- Explains both built-in model streaming and custom async generator approach
- Includes links to working examples (`streaming_openai.py` and `streaming_custom.py`)

### 🎯 Examples

- `streaming_openai.py`: Demonstrates streaming with OpenAI models
- `streaming_custom.py`: Shows how to build custom streaming chatbots
- Fixed quote escaping in examples to prevent file corruption on auto-save

## Implementation Details

- Added `stream` parameter to `openai` ChatModel class
- Implemented async generator support in chat UI element
- Frontend receives incremental updates via `stream_chunk` messages
- Final message marked with `is_final: true` flag

## Testing

- Added `test_chat_streaming_sends_messages` to verify streaming behavior
- Manually tested both OpenAI and custom streaming examples
- Code passes lint and format checks

## User Experience

Users will see responses appear progressively in the chat interface, creating a more engaging and responsive experience similar to ChatGPT, Claude, and other modern AI chat interfaces.

### DIFF
--- a/examples/ai/chat/gemini.py
+++ b/examples/ai/chat/gemini.py
@@ -2,12 +2,13 @@
 # requires-python = ">=3.11"
 # dependencies = [
 #     "marimo",
+#     "google-genai>=1.20.0",
 # ]
 # ///
 
 import marimo
 
-__generated_with = "0.17.4"
+__generated_with = "0.17.8"
 app = marimo.App(width="medium")
 
 
@@ -52,7 +53,7 @@ def _(input_key, mo, os_key):
 def _(key, mo):
     chatbot = mo.ui.chat(
        mo.ai.llm.google(
-            "gemini-1.5-pro-latest",
+            "gemini-2.5-flash",
             system_message="You are a helpful assistant.",
             api_key=key,
             # stream=True,  # Uncomment to enable streaming responses


adamwdraper:

docs/api/inputs/chat.md
```diff
@@ -132,6 +132,69 @@ mo.ui.chat(
...
+
+async def streaming_model(messages, config):
+    """Stream responses word by word."""
+    response = "This response will appear word by word!"
+    words = response.split()
+    accumulated = ""
+    
+    for word in words:
+        accumulated += word + " "
+        yield accumulated
```
mscolnick, line 184: should this be just `yield word`? seems like the accumulation happens client-side

marimo/_plugins/ui/_impl/chat/chat.py
```diff
@@ -227,15 +227,38 @@ async def _send_prompt(self, args: SendMessageRequest) -> str:
...
-                #
-                # RPCs don't yet support bidirectional communication, so we
-                # just ignore all the initial responses; ideally we'd stream
-                # the response back to the frontend.
-                pass
+                # Convert the response to string for streaming
+                chunk_text = str(latest_response)
+
+                # Send incremental update to frontend
+                self._send_message(
```
mscolnick, line 227: could we DRY this up? to:

```
                self._send_message(
                    {
                        "type": "stream_chunk",
                        "message_id": message_id,
                        "content": accumulated_text,
                        "is_final": latest_response is not None,
                    },
                    buffers=None,
                )
```

marimo/_ai/llm/_impl.py
```diff
@@ -466,22 +560,54 @@ def __call__(
...
-                top_p=config.top_p,
-                frequency_penalty=config.frequency_penalty,
-                presence_penalty=config.presence_penalty,
-                stream=False,
-            )
-
-            return response.choices[0].message.content
+            if self.stream:
+                # Stream the response
+                response = litellm_completion(
```
mscolnick: i think we can write `response = litellm_completion(` once with ` stream=self.stream` and then just handle the response differently in the if/else

marimo/_ai/llm/_impl.py
```diff
@@ -306,22 +349,45 @@ def __call__(
...
 
-        content = response.text
-        return content or ""
+        if self.stream:
+            # Stream the response
+            accumulated = ""
+            response = client.models.generate_content_stream(
+                model=self.model,
+                contents=google_messages,
+                config={
```
mscolnick: could we DRY up this config and pull it out above

marimo/_ai/llm/_impl.py
```diff
@@ -55,6 +55,7 @@ class openai(ChatModel):
             If not provided, the API key will be retrieved
             from the OPENAI_API_KEY environment variable or the user's config.
         base_url: The base URL to use
+        stream: Whether to stream the response. Defaults to False.
```
akshayka: Can you provide more context in this docstring on what situations one can set this to `True`? Why does it default to `False`?

docs/api/inputs/chat.md
```diff
@@ -132,6 +132,69 @@ mo.ui.chat(
...
+        "gpt-4o",
+        system_message="You are a helpful assistant.",
+        stream=True,  # Enable streaming
+    ),
+    show_configuration_controls=True
+)
+chat
+```
+
+This works for all built-in models:
```
akshayka, line 158: If this works out of the box with built-in models, should streaming be enabled by default?

akshayka (commented):
Thanks for the PR! Two comments on the API change.

docs/api/inputs/chat.md
adamwdraper: The accumulation must happen server-side. Each yield sends the full accumulated text to the frontend, which displays it. If we yielded just `word`, the UI would only show individual words instead of the progressively building response. This is consistent with how all built-in models work (OpenAI, Anthropic, Google, etc.).

marimo/_plugins/ui/_impl/chat/chat.py
adamwdraper: Good catch! Simplified by eliminating the `chunk_text` variable and directly using `accumulated_text`. Fixed in d71c72574.

marimo/_ai/llm/_impl.py
adamwdraper: Done! Unified the completion call using `stream=self.stream`. The different response types are handled in the if/else branches. Fixed in d71c72574.

marimo/_ai/llm/_impl.py
adamwdraper: Agreed! Extracted the config dict to a single `generation_config` variable. Fixed in d71c72574.

marimo/_ai/llm/_impl.py
adamwdraper: Good point! Set to `True` when you want the response to appear word-by-word as it's generated (like ChatGPT), creating a more responsive UX. It defaults to `False` for backward compatibility and because not all custom models support async generators. I've updated the docstrings to clarify this in 0938d247e.

docs/api/inputs/chat.md
adamwdraper: Good question! I kept it `False` by default for backward compatibility - existing code shouldn't change behavior. But I can see the argument for `True` as default since it's a better UX and works with all built-in models. Happy to change it if you think that's the right call - what do you think?

adamwdraper:
## Additional UI Improvement: Chat Reset Icon Change

Changed the chat reset button icon from trash to rotate-clockwise to improve UX clarity.

**Why this change?**
The trash icon was confusing as it looked like it would delete the cell itself rather than reset the conversation. The rotate icon better represents the "reset/restart" action.

**Additional improvements:**
- Reset button is now disabled when there are no messages (consistent with send button behavior)
- Individual message delete buttons still use the trash icon appropriately

See commit: 4b987ea12

mscolnick:
@adamwdraper looks great! few build errors to fix, but otherwise good to merge afterwards.

we can followup as a team to figure out the correct default for `stream`

adamwdraper:
> @adamwdraper looks great! few build errors to fix, but otherwise good to merge afterwards.
> 
> we can followup as a team to figure out the correct default for `stream`

@mscolnick i tried to fix the build issues, but it seems like they are waiting for approval to run

docs/api/inputs/chat.md
akshayka: Since all built-in models support streaming, I'd say just remove the argument and always stream the response. I can't think of any case when a user wouldn't want to stream the response (are there any?)

PR #7199, opened 2025-11-18
Add `--force` flag to `marimo convert`

koaning:
When dealing with sphinx builds that need marimo to convert files this force flag sure would be nice to have around. 

This is the difference: 

```
> uv run marimo convert demo.py -o out.py                                     
Warning: The file 'out.py' already exists. Overwrite? [y/N]: 

> uv run marimo convert demo.py -o out.py --force
Converted notebook saved to out.py
```

I copied how [this test](https://github.com/marimo-team/marimo/blob/6a703fb5750b680a9b84bab9dbd874007cc7567f/tests/_cli/test_cli_export.py#L409) tests a similar feature and noticed that we don't generally expect a non-zero status code. It's because we check for [non interactive shells here](https://github.com/marimo-team/marimo/blob/6a703fb5750b680a9b84bab9dbd874007cc7567f/marimo/_cli/utils.py#L25) I think. 

Either way, feels like a useful flag, hence the PR.

### DIFF
--- a/tests/_cli/test_cli_convert.py
+++ b/tests/_cli/test_cli_convert.py
@@ -188,7 +188,6 @@ print('Hello from Markdown!')
         assert p.returncode == 0, p.stderr
         assert output_path.exists()
 
-        # Check that adding the force flag doesn't cause an error
         p = subprocess.run(
             [
                 "marimo",
@@ -203,7 +202,6 @@ print('Hello from Markdown!')
         )
         assert p.returncode == 0, p.stderr
 
-        # Check it does error without the extra flag
         p = subprocess.run(
             ["marimo", "convert", str(notebook_path), "-o", str(output_path)],
             capture_output=True,


mscolnick:
cc api change codeowners: @dmadisetti @akshayka

marimo/_cli/convert/commands.py
```diff
@@ -120,7 +131,7 @@ def convert(
 
     if output:
         output_path = Path(output)
-        if prompt_to_overwrite(output_path):
+        if force or prompt_to_overwrite(output_path):
```
mscolnick, line 134: Does "marimo -y" not work? Not sure if it's wired up but that should confirm any prompts

koaning: Oh. I guess I hadn't tried that, mainly because the CLI docs don't point that out. That does totally work!

Related question maybe, why does the `marimo export` command have a `--force` flag? Doesn't the `marimo -y` apply there too?

koaning: I could change this PR to a documentation PR so this is made clear there. Might be better?

mscolnick: we could add a `--force` here as well. it is more common and natural than the `marimo -y` (as you just experienced yourself). im open to adding a `--force` as well, but want to let the api codeowners (@dmadisetti @akshayka) chime in

dmadisetti: `uv run marimo convert demo.py > out.py` should also work. I don't think we need the extra flag

koaning: Fair enough. 

Still worth it to turn this PR into a docs PR that mentions the `-y` flag? It feels hard to discover that one right now.

dmadisetti: Definitely agree. as a note, `-y` provides auto accept for everything

koaning:
Made a new PR for the docs update [here](https://github.com/marimo-team/marimo/pull/7210)

PR #7202, opened 2025-11-18
add markdown renderer to popout columns

Light2Dark:
## 📝 Summary

Screenshot 2025-11-19 at 10 17 47 PM

also adds a copy button on the top right.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/columns.tsx
+++ b/frontend/src/components/data-table/columns.tsx
@@ -15,6 +15,7 @@ import { Maps } from "@/utils/maps";
 import { Objects } from "@/utils/objects";
 import { EmotionCacheProvider } from "../editor/output/EmotionCacheProvider";
 import { JsonOutput } from "../editor/output/JsonOutput";
+import { CopyClipboardIcon } from "../icons/copy-icon";
 import { Button } from "../ui/button";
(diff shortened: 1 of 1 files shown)


frontend/src/components/data-table/columns.tsx
```diff
@@ -531,7 +532,11 @@ export function renderCellValue<TData, TValue>({
         buttonText="X"
         wrapped={isWrapped}
       >
-        <UrlDetector parts={parts} />
+        {isMarkdown(stringValue) ? (
```
mscolnick: this is expensive to do on each string and likely won't have that many hits. i'd prefer to not do this. was there an issue or reason that sparked this?

Light2Dark: No, we don't need to do this, but it does offer a better view for markdown & images in the table

This logic should only trigger when the popup is clicked, so can be optimized (or already is but not obvious..)

mscolnick: `isMarkdown` seems to be called on call values regardless of being clicked.

Light2Dark: Wrapped in a component to make sure it's called on trigger. If there's still a perf concern, we can remove this.

PR #7205, opened 2025-11-18
fix: mermaid embedded in markdown

mscolnick:
Fix mermaid embedded in markdown. Bad boolean flipped. 

This also fixes the markdown formatting of `examples/misc/bayes_theorem.py`


frontend/src/plugins/core/sanitize.ts
```diff
@@ -80,7 +80,7 @@ export function sanitizeHtml(html: string) {
       tagNameCheck: /^(marimo-[A-Za-z][\w-]*|iconify-icon)$/,
       attributeNameCheck: /^[A-Za-z][\w-]*$/,
     },
-    SAFE_FOR_XML: html.includes("marimo-mermaid"),
+    SAFE_FOR_XML: !html.includes("marimo-mermaid"),
```
Light2Dark, line 85: do you want to add a comment? I am confused that it's supposed to be ! includes.

PR #7206, opened 2025-11-18
fix: hide table explorers in vscode

mscolnick:


frontend/src/plugins/impl/data-frames/DataFramePlugin.tsx
```diff
@@ -297,6 +297,7 @@ export const DataFrameComponent = memo(
           get_column_summaries={getColumnSummaries}
           showPageSizeSelector={(total_rows && total_rows > 5) || false}
           showColumnExplorer={false}
+          showRowExplorer={false}
```
Light2Dark: this is a change in existing behaviour,
I can make a follow-up, to pass the props from backend and use the same logic for tables.

I think it'd be nice to display these toggles for dataframes too.

mscolnick: i can make this `true`

mscolnick:
```suggestion
          showRowExplorer={true}
```

PR #7210, opened 2025-11-19
Update convert docsting to mention -y and -g global flags. 

koaning:
Replaces https://github.com/marimo-team/marimo/pull/7199. 

No need to add a `--force` flag, but would be nice to make the user more aware. This is a natural place to mention it.

### DIFF
--- a/marimo/_cli/convert/commands.py
+++ b/marimo/_cli/convert/commands.py
@@ -62,8 +62,8 @@ def convert(
 
         marimo convert script.py -o your_nb.py
 
-    You can also pass global flags to the command line app.
-    For example, you can add the `-q` flag to make the command quiet or the `-y` flag
+    You can also pass global flags to the main marimo command.
+    For example, use `-q` to suppress output or `-y`
     to automatically accept all prompts of the command.
 
         marimo -q -y convert script.py -o your_nb.py


marimo/_cli/convert/commands.py
```diff
@@ -62,6 +62,12 @@ def convert(
 
         marimo convert script.py -o your_nb.py
 
+    You can also pass global flags to the command line app.
+    For example, you can add the `-q` flag to make the command quiet or the `-y` flag
+    to automatically accept prompts any command.
```
dmadisetti:
```suggestion
    to automatically accept all prompts of the command.
```

marimo/_cli/convert/commands.py
```diff
@@ -62,6 +62,12 @@ def convert(
 
         marimo convert script.py -o your_nb.py
 
+    You can also pass global flags to the command line app.
+    For example, you can add the `-q` flag to make the command quiet or the `-y` flag
+    to automatically accept all prompts of the command.
```
akshayka:
```suggestion
    You can also pass global flags to the main marimo command.
    For example, use `-q` to suppress output or `-y`
    to automatically accept all prompts of the command.
```

akshayka (approved):

PR #7212, opened 2025-11-19
minimal create cell buttons

Light2Dark:
## 📝 Summary

- plain button styles
- buttons now span the entire height of the cell

This PR is just the design change. We can follow-up with opening the popup by default.

Before:

https://github.com/user-attachments/assets/1225885d-4ec8-4dc6-862a-210850ed246c

After: 

https://github.com/user-attachments/assets/01d07086-4613-4cab-8219-91eced3f6990

Screenshot 2025-11-19 at 11 39 38 PM

Screenshot 2025-11-19 at 11 40 12 PM

experimented with vertical line:
Screenshot 2025-11-19 at 11 42 04 PM

## 📋 Checklist

### DIFF
--- a/frontend/e2e-tests/py/bad_button.py
+++ b/frontend/e2e-tests/py/bad_button.py
@@ -2,27 +2,28 @@
 # [tool.marimo.runtime]
 # auto_instantiate = true
 # ///
+
 import marimo
 
-__generated_with = "0.0.1"
+__generated_with = "0.17.6"
 app = marimo.App()
 
 
 @app.cell
-def __():
+def _():
     import marimo as mo
-    return mo,
+    return (mo,)
 
 
 @app.cell
-def __(mo):
-    b = mo.ui.button(value=None, label='Bad button', on_click=lambda v: v + 1)
+def _(mo):
+    b = mo.ui.button(value=None, label="Bad button", on_click=lambda v: v + 1)
     b
-    return b,
+    return (b,)
 
 
 @app.cell
-def __(b):
+def _(b):
     b.value
     return
 


akshayka:
@Light2Dark let's also add three options with the pop-up:

Python
Markdown
SQL

An expanded menu is a larger design discussion, and these three options will be an improvement over the status quo

Light2Dark:
Screenshot 2025-11-20 at 3 05 11 AM

Light2Dark:
there's a minor issue where the hoverable area to trigger the buttons to popup needs to be expanded more to the bottom of a cell

akshayka:
> there's a minor issue where the hoverable area to trigger the buttons to popup needs to be expanded more to the bottom of a cell

Sorry what does this mean?

Light2Dark:
Ah sorry, added a video to demonstrate. For the second cell here, the buttons should appear when I hover close to the cell. Not a regression, but we should fix it.

https://github.com/user-attachments/assets/081cca7e-b820-47ee-8010-b372805fa5c8

Light2Dark:
looks good, will look at playwright tomorrow

https://github.com/user-attachments/assets/7165f1b1-38c9-486d-bf29-b0c9fa94d25f

Light2Dark:
> This is looking pretty good!
> 
> One change we should address before merging. @Light2Dark, in the pop-up menu, the setup-cell option should only show up on the top plus button for the first cell in the notebook. It doesn't make sense for the setup cell to be an option on any other plus button, since you cannot add a setup cell in the middle of a notebook.
> 
> Two options, either of which are okay:
> 
> 1. Don't include the setup cell in any pop-up menu.
> 2. Do the more complex thing I suggested and only have the setup cell be an option for the "create above" action for the first cell in the notebook, and nowhere else.

I think the downside is that the setup cell loses it's discoverability. It makes sense to me for it to be on the "create above" for all cells, and we have good UX to scroll to the top of the notebook when it's added

frontend/src/components/editor/notebook-cell.tsx
```diff
@@ -580,6 +580,7 @@ const EditableCellComponent = ({
             ref={cellContainerRef}
             {...cellDomProps(cellId, cellData.name)}
           >
+            <CellLeftSideActions cellId={cellId} actions={actions} />
```
mscolnick, line 583: @Light2Dark could this go under `tray`? does it break anything? currently putting it here messes with the css rounding of the editor

Light2Dark: yeah, this was to make sure the buttons span the entire cell height.
I think we can add border-radius on the cell to fix the root issue..

actually i guess there's something deeper misaligned.
Screenshot 2025-11-21 at 2 18 29 PM

PR #7215, opened 2025-11-19
improvement: drop 'pip install'  when used in the package manager input

mscolnick:
Just nice QOL if you are copy-pasting `pip install pkg` into the package installer


akshayka (approved):
Nice!

PR #7218, opened 2025-11-19
perf: module watcher perf improvements

mscolnick:
Various perf improvements such as: 
- avoiding unnecessary list copy
- `O(n^2) -> O(n)`
- some easy caching fixes (carefully looked it should not cause any stale issues)
- avoid try/except as that has overhead
- fast path checks and simplified logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ and deployable as apps.
 - 🧩 **reusable:** [import functions and classes](https://docs.marimo.io/guides/reusing_functions/) from one notebook to another
 - 🧪 **testable:** [run pytest](https://docs.marimo.io/guides/testing/) on notebooks
 - ⌨️ **a modern editor**: [GitHub Copilot](https://docs.marimo.io/guides/editor_features/ai_completion.html#github-copilot), [AI assistants](https://docs.marimo.io/guides/editor_features/ai_completion.html), vim keybindings, variable explorer, and [more](https://docs.marimo.io/guides/editor_features/index.html)
+- 🧑‍💻 **use your favorite editor**: run in [VS Code or Cursor](https://marketplace.visualstudio.com/items?itemName=marimo-team.vscode-marimo), or edit in neovim, Zed, [or any other text editor](https://docs.marimo.io/guides/editor_features/watching/)
 
 ```python
 pip install marimo && marimo tutorial intro
--- a/README_Japanese.md
+++ b/README_Japanese.md
@@ -3,7 +3,7 @@
 </p>
 
 <p align="center">
-  <em>再現性が高く、Git対応で、スクリプトやアプリとして展開できるリアクティブなPythonノートブック。</em>
+  <em>再現性が高く、Git に優しく、スクリプトやアプリとして展開できるリアクティブな Python ノートブック。</em>
 </p>
 
 <p align="center">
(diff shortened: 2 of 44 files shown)


marimo/_runtime/reload/autoreload.py
```diff
@@ -340,8 +339,9 @@ def update_class(old: object, new: object) -> None:
         except (AttributeError, TypeError):
             pass  # skip non-writable attributes
 
-    for key in list(new.__dict__.keys()):
-        if key not in list(old.__dict__.keys()):
```
mscolnick, line 344: this was `O(n^2)`

marimo/_runtime/reload/autoreload.py
```diff
@@ -432,14 +432,8 @@ def superreload(
...
 
     # collect old objects in the module
-    for name, obj in list(module.__dict__.items()):
-        if not append_obj(module, old_objects, name, obj):
-            continue
-        key = (module.__name__, name)
-        try:
-            old_objects.setdefault(key, []).append(weakref.ref(obj))
-        except TypeError:
-            pass
```
mscolnick, line 442: duplicate code already handled by `append_obj`

marimo/_runtime/reload/module_watcher.py
```diff
@@ -30,16 +29,20 @@ def is_submodule(src_name: str, target_name: str) -> bool:
...
-    if len(src_parts) > len(target_parts):
+    # Fast path: check if target starts with src
+    if not target_name.startswith(src_name):
         return False
-    return all(src_parts[i] == target_parts[i] for i in range(len(src_parts)))
+    # Exact match means it's the same module, which counts as submodule
+    if target_name == src_name:
+        return True
+    # Must be followed by a dot to be a true submodule (prevents matching "foo" with "foobar")
+    return target_name[len(src_name)] == "."
```
mscolnick, line 40: simplified logic

marimo/_runtime/reload/module_watcher.py
```diff
@@ -80,16 +79,32 @@ def _is_third_party_module(module: types.ModuleType) -> bool:
     filepath = safe_getattr(module, "__file__", None)
     if filepath is None:
         return False
-    return "site-packages" in pathlib.Path(filepath).parts
+    return "site-packages" in filepath
```
mscolnick: simplified logic, although its possible to have something named `site-packages-foo` and this would match. i think its ok

marimo/_runtime/reload/module_watcher.py
```diff
@@ -103,10 +118,17 @@ def _check_modules(
     # TODO(akshayka): could also exclude modules part of the standard library;
     # haven't found a reliable way to do this, however.
     excludes = _get_excluded_modules(sys_modules)
+
+    target_modules = set(m for m in modified_modules if m is not None)
+    target_filenames = set(
+        t.__file__ for t in target_modules if hasattr(t, "__file__")
+    )
```
mscolnick, line 126: pull out of the for loop to avoid doing this multiple times

marimo/_runtime/reload/module_watcher.py
akshayka: Nit: This will be negligible performance difference, I would keep the correct version

akshayka (dismissed):
Nice, one nit but otherwise lgtm. Did you notice a change in performance? Might be worth setting up a benchmarks in the future

mscolnick:
Odd passes locally but will dig into it

mscolnick:
Yea noticeably much fewer operations when I logged it manually - I didn't bench it since it quite difficult to mock the individual pieces

marimo/_runtime/reload/module_watcher.py
mscolnick: Yea I can revert this

akshayka:
The test that is failing is a timing component. You could try increasing the amount the test sleeps for, but it would be weird if it fixed it because that'd mean the operation is taking longer than before

PR #7222, opened 2025-11-20
api: remove stream argument for built-in AI models

akshayka:
This PR simplifies the API for the built-in chat models, removing the stream argument and instead always streaming responses (which is a much better experience than synchronous responses). All built-in models already support streaming, so this should just work.

(This does not change the API for custom models, only built-in models)

Removing the argument itself is not breaking change because streaming support (added in #7187) has not been released. However this does change the model return type, which is technically breaking (but we do not anticipate users calling models directly, and instead expect users to use models in `mo.ui.chat()`).

### DIFF
--- a/tests/_ai/llm/test_impl.py
+++ b/tests/_ai/llm/test_impl.py
@@ -27,14 +27,16 @@ def mock_openai_client():
         mock_client = MagicMock()
         mock_openai_class.return_value = mock_client
 
-        # Setup the response structure
-        mock_response = MagicMock()
+        # Setup the streaming response structure
+        mock_chunk = MagicMock()
         mock_choice = MagicMock()
-        mock_message = MagicMock()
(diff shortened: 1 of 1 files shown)


mscolnick (commented):
Changing to streaming is technically breaking since the return type has changed

This is fine if these are only used in chat, since we handle the breaking change. A small edge case could be if someone wrapped this not expecting a streaming response

akshayka:
> This is fine if these are only used in chat, since we handle the breaking change. A small edge case could be if someone wrapped this not expecting a streaming response

Thanks, this is a good callout. I didn't anticipate calling the models directly.

I am okay with making this breaking change if you/@dmadisetti are

akshayka:
Failures look unrelated. Hatch is not being installed for some reason.

PR #7227, opened 2025-11-20
fix: parsing in python-314

dmadisetti:
## 📝 Summary

Noticed these tests breaking on 3.14 with the deprecation of the `ast.Ellipsis` attribute. Easy fix, part of larger effort to stabilize 3.14

### DIFF
--- a/marimo/_ast/parse.py
+++ b/marimo/_ast/parse.py
@@ -188,9 +188,7 @@ class Extractor:
         if node.lineno - node.body[0].lineno == 0:
             # Quirk where the ellipse token seems to have a line index at
             # the end of the dots ...<
-            if sys.version_info < (3, 14) and isinstance(
-                getattr(node.body[0], "value", None), ast.Ellipsis
-            ):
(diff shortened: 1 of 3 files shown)


marimo/_ast/parse.py
```diff
@@ -187,7 +188,9 @@ def to_cell_def(
         if node.lineno - node.body[0].lineno == 0:
             # Quirk where the ellipse token seems to have a line index at
             # the end of the dots ...<
-            if isinstance(getattr(node.body[0], "value", None), ast.Ellipsis):
+            if sys.version_info < (3, 14) and isinstance(
+                getattr(node.body[0], "value", None), ast.Ellipsis
+            ):
```
akshayka: Don't we still need to handle ellipses in 3.14? I thought the `ast` API changed, but Ellipses still exist

dmadisetti: I thought there was a test, but thanks for pushing me to double check

PR #7229, opened 2025-11-20
style: adjust create cell button size, stroke

akshayka:
Also adjust markdown heading toggle to make appearance similar to create cell button, since they are shown together

image


Light2Dark:
thanks @akshayka , i'm adding this in the main PR.

PR #7248, opened 2025-11-21
feat: enhance cell log handling to include tracebacks

### LINKED ISSUE #7246: Traceback browser console logging
### Description

When using the `--redirect-console-to-browser` flag in run mode, it would be really helpful to actually see the tracebacks of exceptions in the browser console as well as what is currently shown, from playing around I think it should be a relatively simple change. 

I set up a very simple notebook that just throws an exception, e.g. one cell like this:
```py
def fn():
    raise ValueError("This is a value error")

fn()
```

When I run this notebook with the redirect flag on, what I see in the browser console is this:
```
[STDERR] [2:59:27 pm] (vblA) {"type":"internal","error_id":"a665e883-e4a3-4b44-a129-11ece22ec141","msg":"An internal error occurred: a665e883-e4a3-4b44-a129-11ece22ec141"}
[STDERR] [2:59:27 pm] (vblA) [31m[E 251121 14:59:27 ops:302](B[m (error_id=a665e883-e4a3-4b44-a129-11ece22ec141) This is a value error
```
Somewhat helpful - but could be better!

I had a play around with the codebase to see if I could figure out why the traceback is not logged, and discovered it's because `write_traceback` https://github.com/marimo-team/marimo/blob/main/marimo/_messaging/tracebacks.py#L24 is formatting the traceback nicely (presumably for the editor UI, though I'm not clear if anything else would use this) and sending it with a different mimetype `"application/vnd.marimo+traceback"`. If I change this method to always just return a string, I can get the traceback to log in the browser console exactly as desired:
```
[STDERR] [2:54:00 pm] (vblA) {"type":"internal","error_id":"58c86625-08f2-453f-9dd0-8e7f4713bdc2","msg":"An internal error occurred: 58c86625-08f2-453f-9dd0-8e7f4713bdc2"}
[STDERR] [2:54:00 pm] (vblA) Traceback (most recent call last):
  File "/home/oroberts/dev/python/general/marimo_notebooks_de/.pixi/envs/default/lib/python3.12/site-packages/marimo/_runtime/executor.py", line 139, in execute_cell
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/marimo_618098/__marimo__cell_vblA_.py", line 4, in <module>
    fn()
  File "/tmp/marimo_618098/__marimo__cell_vblA_.py", line 2, in fn
    raise ValueError("This is a value error")
ValueError: This is a value error
[31m[E 251121 14:54:00 ops:302](B[m (error_id=58c86625-08f2-453f-9dd0-8e7f4713bdc2) This is a value error
```

This is, however, an obviously bad solution because you then lose the traceback highlighting in the editor...

### Suggested solution

I'm really not sure of the best approach here but it _feels_ like it should be a small change!

Option 1 - Make a change here to also log exception mimetypes https://github.com/marimo-team/marimo/blob/f5eb28e5a549fcb65b491446a1b0de0333d2ee6b/frontend/src/core/cells/logs.ts#L24 - though you'd probably have to de-prettify the traceback somehow else you end up with something like:
```
[STDERR] [3:16:44 pm] (vblA) <span class="codehilite"><div class="highlight"><pre><span></span><span class="gt">Traceback (most recent call last):</span>
  File <span class="nb">&quot;/tmp/marimo_619907/__marimo__cell_vblA_.py&quot;</span>, line <span class="m">4</span>, in <span class="n">&lt;module&gt;</span>
<span class="w">    </span><span class="n">fn</span><span class="p">()</span>
  File <span class="nb">&quot;/tmp/marimo_619907/__marimo__cell_vblA_.py&quot;</span>, line <span class="m">2</span>, in <span class="n">fn</span>
<span class="w">    </span><span class="k">raise</span> <span class="ne">ValueError</span><span class="p">(</span><span class="s2">&quot;This is a value error&quot;</span><span class="p">)</span>
<span class="gr">ValueError</span>: <span class="n">This is a value error</span>
</pre></div>
</span>[31m[E 251121 15:16:44 ops:302](B[m (error_id=6d29c9a5-213c-47fe-97bf-2239f799b1f5) This is a value error
```

Option 2 - Maybe where `write_traceback` is called in `run` https://github.com/marimo-team/marimo/blob/f5eb28e5a549fcb65b491446a1b0de0333d2ee6b/marimo/_runtime/runner/cell_runner.py#L562 you could account for if you're in edit or run mode and only prettify the exception if you're in edit mode?

Happy to do the PR if it helps you guys but it feels like more of a decision on how to do it than a lot of code to be written.

- [x] Yes

### Alternatives

_No response_

### Additional context

_No response_

mscolnick:
Closes #7246

### DIFF
--- a/frontend/src/core/cells/__tests__/logs.test.ts
+++ b/frontend/src/core/cells/__tests__/logs.test.ts
@@ -7,7 +7,9 @@ import { formatLogTimestamp, getCellLogsForMessage } from "../logs";
 describe("getCellLogsForMessage", () => {
   beforeEach(() => {
     // Mock console.log to avoid cluttering test output
-    vi.spyOn(console, "log").mockImplementation(() => {});
+    vi.spyOn(console, "log").mockImplementation(() => {
+      // no-op
+    });
   });
 
   afterEach(() => {
@@ -243,7 +245,7 @@ describe("getCellLogsForMessage", () => {
       console: [
         {
           mimetype: "text/plain",
-          channel: "pdb" as any, // Non-logging channel
+          channel: "pdb" as unknown as "stdout", // Non-logging channel
           data: "Should be ignored",
           timestamp: 1234567890,
         },
@@ -262,7 +264,7 @@ describe("getCellLogsForMessage", () => {
   test("returns empty array when console is null", () => {
     const cellMessage: CellMessage = {
       cell_id: "cell-11",
-      console: null as any,
+      console: null as unknown as CellMessage["console"],
       output: null,
       status: "idle",
       stale_inputs: null,


frontend/src/core/cells/logs.ts
```diff
@@ -21,21 +22,31 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
   const consoleOutputs: OutputMessage[] = [cell.console].filter(Boolean).flat();
 
   for (const output of consoleOutputs) {
-    if (output.mimetype === "text/plain") {
-      invariant(typeof output.data === "string", "expected string");
+    // Handle text/plain, text/html, and traceback MIME types
+    const isTextPlain = output.mimetype === "text/plain";
+    const isHtml =
+      output.mimetype === "text/html" ||
+      output.mimetype === "application/vnd.marimo+traceback";
```
Light2Dark, line 29: do we need to include `"application/vnd.marimo+error"`?

Light2Dark (dismissed):

frontend/src/core/cells/logs.ts
mscolnick: as far as i remember `application/vnd.marimo+error` are never passed to `consoleOutputs`

PR #7256, opened 2025-11-21
improvement: fallback to --no-cache for uv installations that fail

mscolnick:
Adds automatic retry with `--no-cache` flag when `uv pip install` fails due to cache write permission errors.

  ## Problem

  In certain environments (e.g., Docker containers, restricted file systems), UV may fail to install packages with errors like:

  × Failed to download and build pyqtree==1.0.0
  ├─▶ Failed to write to the distribution cache
  ╰─▶ Operation not permitted (os error 1)

  This prevents package installation even though the installation would succeed if the cache were disabled.

  ## Solution

We now:
  1. Capture output during `uv pip install` commands
  2. Detect cache write permission errors in the output
  3. Automatically retry the installation with `--no-cache` flag when such errors occur

---

> [!NOTE]
> Adds output-capturing install flow for uv pip installs and retries with --no-cache when cache write permission errors are detected (no retry in project mode).
> 
> - **Runtime**:
>   - **`UvPackageManager`**:
>     - Add `_is_cache_write_error` and detect cache write permission failures.
>     - Change `_install` to stream/capture output and, on failure, retry `uv pip install` with `--no-cache` when applicable.
>     - Skip fallback in uv project mode; use existing flow.
> - **Tests**:
>   - Update non-project install tests to use `subprocess.Popen` with streaming.
>   - Add tests for cache-error fallback, non-cache error (no retry), and project mode (no fallback).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b30d743d344b7685216eadddd0c3ad11ac6da0cf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/marimo/_runtime/packages/pypi_package_manager.py
+++ b/marimo/_runtime/packages/pypi_package_manager.py
@@ -163,9 +163,15 @@ class UvPackageManager(PypiPackageManager):
         return find_uv_bin()
 
     def _is_cache_write_error(self, output_text: str) -> bool:
+        """Check if the output text indicates a cache write error.
+
+        This is somewhat fragile and could break with new uv output.
+        This was tested with uv ~0.9.7
+        """
+        output_text = output_text.lower()
         return (
-            "Failed to write to the distribution cache" in output_text
-            or "Operation not permitted" in output_text
+            "failed to write to the distribution cache" in output_text
+            or "operation not permitted" in output_text
         )
 
     def is_manager_installed(self) -> bool:


marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -162,6 +162,12 @@ class UvPackageManager(PypiPackageManager):
     def _uv_bin(self) -> str:
         return find_uv_bin()
 
+    def _is_cache_write_error(self, output_text: str) -> bool:
+        return (
```
dmadisetti, line 172: Can you leave a comment with the UV/pip version you verified this with? Seems like a potentially brittle check

dmadisetti (approved):
LGTM, worried it might be a little brittle, but we can always come back if updates break this

PR #7257, opened 2025-11-21
feat: Add `date_slider` input widget

### LINKED ISSUE #6316: slider option for date_range
### Description

I would like to have something like `mo.ui.range_slider` but that would show dates so I can visualize the dates linearly 

### Suggested solution

Could be an option from `mo.ui.date_range` itself or a way to change the label displayed for `mo.ui.range_slider`. For example, being able to change the value shown with `show_value` option to use POSIX timestamps in range_slider

- [x] Yes

### Alternatives

_No response_

### Additional context

_No response_

FBruzzesi:
## 📝 Summary

This is a proof of concept (maybe headstart?) to introduce a `date_slider` (edit: actually a `date_range_slider`) widget.

Fixes #6316

## 🔍 Description of Changes

- Adds `date_slider` ui widget
- Arguably too much refactor in the `dates.py` module
- The frontend is 95% AI generated, that's why I am opening this as a draft, and happy to let anyone take it over. My main objective here was to see how feasible it was to add this component.
- TODO: I would like to add the possibility to use a custom formatting that is applied to the dates before 

<details><summary>Screenshot of rendering</summary>
<p>

image

</p>
</details> 

## 📋 Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions). [Thread link](https://discord.com/channels/1059888774789730424/1441427087603925083/1441432814732972033).

### DIFF
--- a/docs/api/inputs/dates.md
+++ b/docs/api/inputs/dates.md
@@ -59,3 +59,27 @@
 
 ::: marimo.ui.date_range
   :members:
+
+## Date slider
+
+/// marimo-embed
+
+```python
+    @app.cell
+    def __():
+        date_slider = mo.ui.date_slider(
+            start="2025-01-01",
+            stop="2025-01-15",
+        )
+        return
+
+    @app.cell
+    def __():
+        mo.hstack([date_slider, mo.md(f"Has value: {date_slider.value}")])
+        return
+```
+
+///
+
+::: marimo.ui.date_slider
+  :members:
--- a/docs/api/inputs/index.md
+++ b/docs/api/inputs/index.md
@@ -16,6 +16,7 @@ powerful notebooks and apps. These elements are available in `marimo.ui`.
 | [`marimo.ui.date`][marimo.ui.date] | Date picker |
 | [`marimo.ui.datetime`][marimo.ui.datetime] | Date and time picker |
 | [`marimo.ui.date_range`][marimo.ui.date_range] | Date range picker |
+| [`marimo.ui.date_slider`][marimo.ui.date_slider] | Date slider |
 | [`marimo.ui.dictionary`][marimo.ui.dictionary] | Dictionary inputs |
 | [`marimo.ui.dropdown`][marimo.ui.dropdown] | Create dropdowns |
 | [`marimo.ui.file`][marimo.ui.file] | File uploads |
--- a/frontend/src/components/ui/date-slider.tsx
+++ b/frontend/src/components/ui/date-slider.tsx
@@ -1,0 +1,101 @@
+/* Copyright 2024 Marimo. All rights reserved. */
+
+import * as SliderPrimitive from "@radix-ui/react-slider";
+import * as React from "react";
+import { cn } from "@/utils/cn";
+import { useBoolean } from "../../hooks/useBoolean";
+import {
+  TooltipContent,
+  TooltipPortal,
+  TooltipProvider,
+  TooltipRoot,
+  TooltipTrigger,
+} from "./tooltip";
+
+const DateSlider = React.forwardRef<
+  React.ElementRef<typeof SliderPrimitive.Root>,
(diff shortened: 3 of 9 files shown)


FBruzzesi:
The automatic labeling to "documentation" seems off 🫠

mscolnick:
@akshayka @dmadisetti  - curious if you think `mo.ui.slider(` should support `date` and `datetime` in the params, or if we should keep them separate under `mo.ui.date_slider`

FBruzzesi:
@mscolnick for context this is a date _range_ slider (I guess I could rename it), meaning start to end.
But in both cases, it would make sense to have those to be generic over a few types. In general I think also custom ordered sequences would be interesting (at least to me)

Screenshot:

image

mscolnick:
I think if they are separate sliders, then it would be good rename to include `range`, but we can wait to see hear feedback if they should be merged. 

(I'm leaning towards merging them if it's possible and the api does not get overloaded with different params when a int vs date)

akshayka:
Neat PR!

I am leaning toward adding two new APIs:

1. `date_slider`
2. `date_range_slider`

These are the reasons I prefer adding new APIs:

1. The types are clear. `mo.ui.slider(...).value` is guaranteed to be numeric, so I can do arithmetic on it without having to check its type
2. I can imagine feature creep related to dates (formatting, timezones) that might not relate to numeric sliders.

That said ... maybe the type concern could be addressed with generics and `@overload`, which would allow us to keep a single API ([example](https://marimo.app/l/0cvz3a)). But that may be too complicated, and it would also be a breaking change because `mo.ui.slider` would become a function instead of a class.

(If we did use a single API, then I'd suggest we make the change across `slider` and `range_slider` at the same time, and also update `steps` to support `date`.)

akshayka:
>  In general I think also custom ordered sequences would be interesting (at least to me)

@FBruzzesi Can you describe your use case for custom ordered sequences? I am wondering if we just design for that solution instead of special casing dates

FBruzzesi:
> @FBruzzesi Can you describe your use case for custom ordered sequences? I am wondering if we just design for that solution instead of special casing dates

Hey @akshayka 👋🏼

First of all thanks for the patience in this PR, I am really tight these days. And thanks for considering a broader request.

Here a use case I have goes as follows.

I would like to create a range slider with a sequence that has a natural order. Specifically I would want to have a sequence of [isoweeks](https://en.wikipedia.org/wiki/ISO_week_date). I build a library [iso-week-date](https://github.com/FBruzzesi/iso-week-date) to simplify a bunch of operations in python (such as addition, subtraction and comparison), but in general notice that the iso weeks have a natural lexicographical order, e.g. "2025-W01" < "2025-W02").
With the library I can do: `IsoWeek.range(start, end, step)` and get a range of strings (or IsoWeek objects depending on another parameter).

If I were able to put these values into a slider, I can use such to filter a dataframe by some column (e.g. `frame.filter(pl.col("week").is_between(pl.lit(slider.value[0]), pl.lit(slider.value[1])))` to either plot or report metrics in different time periods - this is my actual use case.

yan-hic:
Bumping this. The screenshot looks great ! 
So much easier than browsing through the `date_range` calendars to select from/to dates across months or years.

2 additions I would find useful: 
- add default tuple value (like the `date_range` element)
- 'X' to reset the range to its default values above

To be complete, Panel has a nice https://panel.holoviz.org/reference/widgets/DateRangeSlider.html and  it works as-is in marimo but renders badly especially in dark mode:
```
import panel as pn
import datetime as dt

pn.extension()

date_slider = pn.widgets.DateRangeSlider(
    name='Date Range Slider',
    start=dt.datetime(2023, 1, 1),
    end=dt.datetime(2024, 1, 1),
    value=(dt.datetime(2023, 1, 1), dt.datetime(2023, 6, 1)),
    step=1
)

date_slider
```

PR #7282, opened 2025-11-24
Add Gemini v3

koaning:
This PR adds the new gemini model so folks can configure v3 easily.

### DIFF
--- a/packages/llm-info/data/models.yml
+++ b/packages/llm-info/data/models.yml
@@ -53,6 +53,13 @@
 
 # Google
 
+- name: Gemini 3.0 Pro
+  model: gemini-3-pro-preview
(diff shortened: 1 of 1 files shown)


packages/llm-info/data/models.yml
```diff
@@ -53,6 +53,13 @@
 
 # Google
 
+- name: Gemini 3.0 Pro
+  model: gemini-3-pro-preview
```
mscolnick, line 57: Is there nothing out of preview? At some point when they take them out of preview, this model name starts to fail

mscolnick: i couldn't find anything. we can put this in and then drop preview once its released

PR #7287, opened 2025-11-24
lint rule: check for branching expressions

dmadisetti:
## 📝 Summary

Adds common genai mistake where the last expression is expected to conditionally show if nested in an expression. Example:

```python
if condition:
    mo.md("Condition was true")
else:
    mo.md("Condition was false")
```

### DIFF
--- a/examples/third_party/cvxpy/signals/app.py
+++ b/examples/third_party/cvxpy/signals/app.py
@@ -250,7 +250,7 @@ def _(complib, radios, solved_ever):
         [complib.Components.TREND_LINE, complib.Components.PERIODIC]
     )
     if solved.now:
-        _ = solved_ever.set()
+        solved_ever.set()
     return (solved,)
 
 
--- a/tests/_lint/test_files/branch_expression.py
+++ b/tests/_lint/test_files/branch_expression.py
@@ -60,5 +60,63 @@ def _(x):
     return
 
 
+@app.cell
+def _(invalid):
+    # Negative case: print() should NOT be flagged
+    if invalid:
+        print("Debug message")
+    else:
+        print("Another debug message")
+    return
+
+
+@app.cell
+def _(invalid, mo):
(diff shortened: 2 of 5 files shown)


mscolnick:
it seems like there are a few false-positives. maybe we can special case `mo.md` or all `mo.ui` (except `mo.output`, `mo.stop`)

dmadisetti:
>  (except mo.output, mo.stop)

This is a good diea

tests/_lint/test_files/branch_expression.py
```diff
@@ -0,0 +1,64 @@
+import marimo
```
mscolnick, line 1: it might be worth capturing a valid use-cases as well in here:
- mo.output
- mo.stop
- sys.*

examples/third_party/cvxpy/signals/app.py
```diff
@@ -250,7 +250,7 @@ def __init__(self, sticky_bool):
         [complib.Components.TREND_LINE, complib.Components.PERIODIC]
     )
     if solved.now:
-        solved_ever.set()
+        _ = solved_ever.set()
```
mscolnick: should we leave this? i think this should be valid just as `solved_ever.set()`

dmadisetti: We could just limit this to only mo.md calls, and constant expressions

`solved_ever.set() -> Self`, which is ambiguous. There's also a user hint suggesting this behavior.

but this is somewhat consistent with pylint: https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/expression-not-assigned.html

mscolnick: we should just lint our function calls (`mo.*`) and leave pylint (or the user's preferred linter) to handle this. they can enable or disable this as they'd like

dmadisetti: Done!

PR #7299, opened 2025-11-25
add markdown rendering for tables in popout

Light2Dark:
## 📝 Summary

This was missed in the [earlier PR ](https://github.com/marimo-team/marimo/pull/7201) because it was a branch of another branch.

This doesn't support all flavours of markdown, so some styling is missed.

image

image

## 📋 Checklist


frontend/src/components/data-table/url-detector.tsx
```diff
@@ -82,6 +84,49 @@ export function parseContent(text: string): ContentPart[] {
...
+// Wrapper component so that we call isMarkdown only on trigger
+export const MarkdownUrlDetector = ({
+  content,
+  parts,
+}: {
+  content: string;
+  parts: ContentPart[];
+}) => {
+  if (isMarkdown(content)) {
+    return <MarkdownRenderer content={content} />;
```
mscolnick, line 123: this still runs on every cell, right?

frontend/src/components/data-table/url-detector.tsx
```diff
@@ -82,6 +84,49 @@ export function parseContent(text: string): ContentPart[] {
   });
 }
 
+export function isMarkdown(text: string): boolean {
+  const tokens = marked.lexer(text);
+
+  const commonMarkdownIndicators = [
```
mscolnick: id pull this out to a set and do `tokens.some(token => commonMarkdownIndicators.has(token))`

frontend/src/components/data-table/url-detector.tsx
```diff
@@ -1,12 +1,14 @@
...
 
+import { marked } from "marked";
 import { useState } from "react";
 import {
   Popover,
   PopoverContent,
   PopoverTrigger,
 } from "@/components/ui/popover";
 import { Events } from "@/utils/events";
+import { MarkdownRenderer } from "../chat/markdown-renderer";
```
mscolnick: we should move the markdown stuff to a common folder so data-table does not depend on the `chat/` folder

frontend/src/components/data-table/url-detector.tsx
mscolnick: we could make `<PopoutColumn>` take a `renderContent` instead of children and track when the popout is open and do `{isOpen ? renderContent() : null}`

Light2Dark: This only runs when the cell is clicked and the popout is opened. Radix's popover only renders the content on open.
But it has to be a component, previously I did not use a custom component and it would call on every cell.

Thanks for the other comments, I agree. And we can also merge this **after** release for safety.

mscolnick (approved):

PR #7304, opened 2025-11-25
Migrate back to `msgspec` from `msgspec_m`

manzt:
`msgspec_m` was a temporary fork of `msgspec` to work around an issue with python 3.14 compatibility. There should be regular releases of msgspec going forward.


pyproject.toml
```diff
@@ -48,7 +48,7 @@ dependencies = [
     "narwhals>=2.0.0",
     # for packaging.version; not sure what the lower bound is.
     "packaging",
-    "msgspec_m>=0.19.2",
+    "msgspec>=0.20.0",
```
manzt, line 51: Not sure if we need to pin this lower bound, 0.19 would likely be fine.

mscolnick: this should be fine. i highly doubt any other dep has pinned `<0.20.`

PR #7364, opened 2025-12-01
fix: remove NaN from number plugin

### LINKED ISSUE #7352: `mo.ui.number().form()` disappears after submitting empty value
### Describe the bug

When I use `mo.ui.number().form()` with an empty value and click **Submit**,  
the whole form UI disappears from the notebook.

I would expect the form to stay visible and just submit `None` (or some
empty value) instead, but currently the widget itself vanishes.

Image

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

{
  "marimo": "0.18.1",
  "editable": false,
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.11",
  "Locale": "ja_JP",
  "Binaries": {
    "Browser": "142.0.7444.176",
    "Node": "v22.12.0"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.3",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10",
    "narwhals": "2.12.0",
    "packaging": "25.0",
    "psutil": "7.1.3",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.17.2",
    "pyyaml": "6.0.3",
    "starlette": "0.50.0",
    "tomlkit": "0.13.3",
    "uvicorn": "0.38.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.10.0",
    "ruff": "0.14.7"
  },
  "Experimental Flags": {}
}

</details>

### Code to reproduce

```python
import marimo as mo

number = mo.ui.number(label="number")  # initial value is empty / None
form = number.form()

form
````

Steps:

1. Run the cell above.
2. Keep the number input empty (or delete the existing value).
3. Click **Submit**.
4. The form UI disappears.

mscolnick:
Fixes #7352, although I think the underlying issue/fix is #7363

I plan to follow up with more tests/updates to other places where this could crop up.

### DIFF
--- a/frontend/src/plugins/impl/NumberPlugin.tsx
+++ b/frontend/src/plugins/impl/NumberPlugin.tsx
@@ -67,7 +67,7 @@ const NumberComponent = (props: NumberComponentProps): JSX.Element => {
 
   const handleChange = (newValue: number) => {
     onChange(withoutNaN(newValue));
-  }
+  };
 
   return (
     <Labeled label={props.label} id={id} fullWidth={props.fullWidth}>
--- a/frontend/src/plugins/impl/__tests__/NumberPlugin.test.tsx
+++ b/frontend/src/plugins/impl/__tests__/NumberPlugin.test.tsx
@@ -469,7 +469,9 @@ describe("NumberPlugin", () => {
     });
 
     // Check that the NumberField has the full width class
-    const numberField = container.querySelector('[data-testid="marimo-plugin-number-input"]');
-    expect(numberField?.classList.contains('w-full')).toBe(true);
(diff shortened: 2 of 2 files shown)


Light2Dark:
Any thoughts on just doing #7363 fix & keep the tests + `value ?? NaN`? I think it's okay to have it as NaN

mscolnick:
@Light2Dark - the plugin state should be serializable, and `NaN` will just get serialized to `null`. i think its ok to do that here as well (being defensive twice, since this could come up again)

PR #7368, opened 2025-12-01
improvement: add file limit, warning, and more performant file detection

### LINKED ISSUE #7317: Workspace view - Request timed out
### Describe the bug

As my use of Marino notebooks grows in my project, the workspace view is struggling to keep up. I often get a "Request timed out: Loading workspace files took too long" error instead of displaying the files in the directory. Reloading the page often works to make it functional again.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
Windows 11
Marimo 0.18.0
```

</details>

### Code to reproduce

_No response_

mscolnick:
Fixes #7317

- add file limit when displaying the workspace (1000 files)
- show a warning when we go over the limit
- more performant file detection by only reading the first `500b` header

### DIFF
--- a/marimo/_server/file_router.py
+++ b/marimo/_server/file_router.py
@@ -23,6 +23,9 @@ if TYPE_CHECKING:
 
 LOGGER = _loggers.marimo_logger()
 
+# Maximum number of files to return in workspace scan
+MAX_FILES = 1000
+
 # Some unique identifier for a file
 MarimoFileKey = str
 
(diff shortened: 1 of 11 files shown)


marimo/_server/file_router.py
```diff
@@ -480,3 +514,56 @@ def validate_inside_directory(directory: Path, filepath: Path) -> None:
...
+
+
+# Count total marimo files (not directories)
+def count_files(file_list: list[FileInfo]) -> int:
+    count = 0
+    for item in file_list:
+        if not item.is_directory:
+            count += 1
+        if item.children:
+            count += count_files(item.children)
```
Light2Dark, line 568: can we early return? should help perf.
if > MAX_FILES, return "too_many"

mscolnick: we already return early and stop the recursion when fetching the files. this will really only be `MAX_FILES + a few`, where the additional few is coming from the additional children (e.g. if we had 999 and then added 7 more children)

PR #7372, opened 2025-12-02
Uv dev install

dmayilyan:
## 📝 Summary

For UV projects redirect dev dependencies from _Settings -> Optional Dependencies_ UI menu to be installed in dev dependency group. As mentioned menu is targeting development tools proposed change fits the logic of isolation.

## 🔍 Description of Changes

- [x] For UV projects add `--dev` flag when install request comes from _Optional Dependencies_.
- [x] For UV projects add `--dev` flag when uninstalling a package which is in **dev dependency group**, by passing tag content
- [x] Pass dev=True boolean when install request comes from _Optional Dependencies_
- [x] When uninstalling pass dev=True boolean when tag group is **dev**

## 📋 Checklist


dmayilyan:

mscolnick:
@dmayilyan thanks for the contribution! we will take a look after the release today

marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -58,7 +58,10 @@ class PipPackageManager(PypiPackageManager):
     name = "pip"
     docs_url = "https://pip.pypa.io/"
 
-    def install_command(self, package: str, *, upgrade: bool) -> list[str]:
+    @override
```
mscolnick: unfortunately `override` was only introduced in 3.12, so this likely will break earlier versions

frontend/src/components/app-config/optional-features.tsx
```diff
@@ -211,7 +211,10 @@ const InstallButton: React.FC<{
           return pkg.name;
         })
         .join(" ");
-      const response = await addPackage({ package: packageSpec });
+      const response = await addPackage({
+        package: packageSpec,
+        dev: true,
+      } as any);
```
mscolnick: if you run `make fe-codegen`, it will update the typings so you can remove the `any`

frontend/src/components/editor/chrome/panels/packages-panel.tsx
```diff
@@ -387,15 +387,22 @@ const UpgradeButton: React.FC<{
...
     try {
       setLoading(true);
-      const response = await removePackage({ package: packageName });
+      const isDev = tags?.some(
+        (tag) => tag.kind === "group" && tag.value === "dev",
+      );
+      const response = await removePackage({
+        package: packageName,
+        dev: isDev,
+      } as any);
```
mscolnick: if you run `make fe-codegen`, it will update the typings so you can remove the `any`

marimo/_runtime/packages/pypi_package_manager.py
dmayilyan: Good catch, I will try to avoid override (also typing_extensions based) altogether

mscolnick:
hey @dmayilyan, i just approved/ran the test. there are a few lint and test errors, although I think the bottom two might be broken on main currently. Ive pasted the commands and output. let me know if you need help

**`hatch run typecheck:check`**
```
marimo/_runtime/packages/package_manager.py:80: error: Incompatible return value type (got "Coroutine[Any, Any, bool]", expected "bool")  [return-value]
marimo/_runtime/packages/package_manager.py:80: note: Maybe you forgot to use "await"?
marimo/_runtime/packages/conda_package_manager.py:39: error: Incompatible return value type (got "Coroutine[Any, Any, bool]", expected "bool")  [return-value]
marimo/_runtime/packages/conda_package_manager.py:39: note: Maybe you forgot to use "await"?
marimo/_runtime/packages/pypi_package_manager.py:618: error: Incompatible return value type (got "Coroutine[Any, Any, bool]", expected "bool")  [return-value]
marimo/_runtime/packages/pypi_package_manager.py:618: note: Maybe you forgot to use "await"?
marimo/_runtime/packages/pypi_package_manager.py:656: error: Incompatible return value type (got "Coroutine[Any, Any, bool]", expected "bool")  [return-value]
marimo/_runtime/packages/pypi_package_manager.py:656: note: Maybe you forgot to use "await"?
```

**`hatch run +py=3.13 test-optional:test tests/_runtime/packages`**
```
FAILED tests/_runtime/packages/test_package_managers.py::test_pip_install_with_log_callback - assert <coroutine object test_pip_install_with_log_callback.<locals>.MockPipPackageManager.run at 0x11c624190> is True
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_failed_install_returns_false - assert not <coroutine object PackageManager.run at 0x11c5da0a0>
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_install - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_install_failure - assert <coroutine object PackageManager.run at 0x11af8fca0> is False
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_uv_install_in_project - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_uv_install_dev_dependency_in_project - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_runtime/packages/test_pypi_package_manager.py::test_uv_install_in_project_no_fallback - AssertionError: Expected 'run' to be called once. Called 0 times.
FAILED tests/_runtime/test_manage_script_metadata.py::test_manage_script_metadata_uv - assert '"marimo",' in '# /// script\n# requires-python = ">=3.13"\n# dependencies = [\n#     "marimo>=0.18.4",\n# ]\n# ///\n'
FAILED tests/_runtime/test_manage_script_metadata.py::test_manage_script_metadata_uv_deletion - assert '"marimo",' in '# /// script\n# requires-python = ">=3.13"\n# dependencies = [\n#     "marimo>=0.18.4",\n# ]\n# ///\n'
```

dmayilyan:
Indeed when I did a pull and rerun the tests, I see the errors you sent. I guess I missed a minor release in between.

dmayilyan:
@mscolnick Fixed the failing tests

dmayilyan:
Hello @mscolnick As I see, failing tests are not related to my code. Am I missing something? When ran locally these tests where failing as they do here. Is that an expected behavior? Thanks

mscolnick:
Hi sorry @dmayilyan the tests look good now. We have just been busy lately but @manzt will review this on Monday

dmayilyan:
Thanks @mscolnick

frontend/src/components/editor/chrome/panels/packages-panel.tsx
```diff
@@ -599,7 +606,11 @@ const DependencyTreeNode: React.FC<{
         {isTopLevel && (
           <div className="flex gap-1 invisible group-hover:visible">
             <UpgradeButton packageName={node.name} onSuccess={onSuccess} />
-            <RemoveButton packageName={node.name} onSuccess={onSuccess} />
+            <RemoveButton
+              packageName={node.name}
+              tags={node.tags as { kind: string; value: string }[]}
```
manzt: Ideally we should update our generated types for `node.tags` to be more type safe (instead of `Record<string,any>`). I can make a follow up PR.

frontend/src/components/app-config/optional-features.tsx
```diff
@@ -211,7 +211,7 @@ const InstallButton: React.FC<{
           return pkg.name;
         })
         .join(" ");
-      const response = await addPackage({ package: packageSpec });
+      const response = await addPackage({ package: packageSpec, dev: true });
```
manzt, line 214: Thinking through the implications of `{dev: true}` here...

I _think_ it's sensible for most optional features since they're editor-related (formatting, AI, LSP, etc.). My instinct is that we want to avoid writing metadata that would break:

```sh
uv run --no-dev my_notebook.py
```

The use case being that I have many of my "editor" features as optional deps, but then I can run my notebooks as scripts more efficiently.

The main case I can think of where this could be problematic is SQL cells - `duckdb` is a hard runtime dependency if the notebook actually uses SQL cells. The others should be safe as dev deps.

frontend/src/components/editor/chrome/panels/packages-panel.tsx
manzt:
```suggestion
              // FIXME: Backend types are wrong/outdated.
              // tags actually have the shape: Array<{ kind: string; value: string }>
              // @ts-expect-error — backend tag types do not match frontend expectations yet
              tags={node.tags}
```

Since the generated backend types here aren’t accurate (`Array<Record<string, any>`), I’d prefer avoiding a type cast since it won’t protect us if the shape changes in the future. Marking this as a TODO makes it clear we should fix the generated types in a follow-up PR.

frontend/src/components/app-config/optional-features.tsx
dmayilyan: Agree about duckdb. Besides, Altair can be an issue, as it may be someone's plotting tool in addition to being marimo's dev dependency.

There was an idea to have check boxes in the Optional Dependencies menu so that they can be selectively set to be installed in dev. That, on the other hand raises a new layer of conditions  _**a package manager supporting dev or not**_ and _**check marks being in the UI altogether/being inactive etc**_. I am not sure whether such a complex approach we would give us more value rather than visual/logical clutter to the product. What do you think?

dmayilyan: I was thinking about the sandbox issue you mentioned and In my perception that's how it should behave. In my view where there is an expectation of package name in the sidebar `--dev` should not be parsed. It's a bit stylistic choice, so it is hard to argue :). I would go with a checkbox (or dropdown list) near the Add button:
image

manzt: > I was thinking about the sandbox issue you mentioned and In my perception that's how it should behave.

Looking back, I agree with this. I thought we were passing down `--dev` somewhere explicitly, but we just forward the args there down to the command line. I wouldn't worry about the dev checkbox because it's the same issue we'd have to resolve with package managers that do not support a notion of "dev" dependencies.

manzt (approved):
This looks good to go! The unit test failing I believe is unrelated (#7563).

~It seems the api.yaml was manually edited. Could you regenerate it by running:~ `make fe-codegen` 
~This will regenerate the OpenAPI spec and TypeScript types from the Python models.~

The schema check failure is actually a pre-existing CI issue - `oasdiff` can't parse the `api.yaml` on `main`:

```
  Error: failed to load base spec from "main/packages/openapi/api.yaml"
```

But the changes here are backwards-compatible and should be fine to merge.

PR #7375, opened 2025-12-02
improvement: make the install/uninstall API endpoints non-blocking

mscolnick:
These endpoints get block other endpoints during installation. This is now fixed by wrapping the blocking commands in a `asyncio.to_thread`. 

Another option is to change `run` to a natural async function, but that requires changing more downstream sync functions in the kernel.

### DIFF
--- a/tests/_runtime/packages/test_package_managers.py
+++ b/tests/_runtime/packages/test_package_managers.py
@@ -35,7 +35,11 @@ def test_update_script_metadata() -> None:
     runs_calls: list[list[str]] = []
 
     class MockUvPackageManager(UvPackageManager):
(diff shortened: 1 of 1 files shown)


manzt (approved):
This seems like a reasonable/minimal change to me. It would make sense to make the package manager async itself, but I'd rather evaluate that if this doesn't sufficiently handle our current problem. We mostly want to offload the blocking nature of the install/uninstall endpoints.

mscolnick:
@manzt agreed. we would just need to convert `subprocess.popen` to the async version, but this causes some hairy downstream function color changes (sync -> async).

PR #7401, opened 2025-12-04
improvement: enhance lazy df utilities and ensure dataframe type preservation in transformations

### LINKED ISSUE #7399: mo.ui.dataframe changed behaviour for ibis tables
### Describe the bug

Those points might be introduced with #6772 
1. when working on an Ibis table with `ui=mo.ui.dataframe(ibis_table,limit=3)`, the set limit is only applied to the displayed table, but not to the underlying query. This was kindly introduced in [#2312](https://github.com/marimo-team/marimo/pull/2312)
2. The type of `ui.value` is not anymore an Ibis table, but a pyarrow table. So the lazy behavior of Ibis is gone.

I am not into the narwhals, so can not really help in fixing.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
"marimo": "0.18.2",
```

</details>

### Code to reproduce

```Python
import marimo as mo
import pytest
import ibis

# Connect to DuckDB via Ibis
con = ibis.duckdb.connect()

# Enable logging inside DuckDB to capture queries
con.raw_sql("CALL enable_logging();")

# Sample data to populate a test table
samle_data = {"a": [1, 2, 3, 4, 5]}

# Limit used for the UI dataframe preview
ui_limit = 3

# Create a table in DuckDB from the sample data
ibis_table = con.create_table("test_table", samle_data)

# Create a marimo UI dataframe with a preview limit
ui = mo.ui.dataframe(
    ibis_table, limit=ui_limit
)

# Retrieve DuckDB logs to inspect queries executed by the UI/dataframe operations
df_duckdb_logs = con.sql("SELECT * FROM duckdb_logs;").to_pandas()

# Extract the message from the second-to-last log entry (expected to contain the LIMIT clause)
dataframe_query = df_duckdb_logs.iloc[-2].message

def test_ibis_dataframe_limit():
    assert f'LIMIT {ui_limit}' in dataframe_query
def test_ibis_dataframe_type():
    assert type(ibis_table) == type(ui.value)
def test_ui_value():
    assert 'LIMIT' not in ibis.to_sql(ui.value)
```

mscolnick:
Fixes #7399
When using `mo.ui.dataframe` with lazy dataframes (Ibis tables, Polars LazyFrames), the lazy behavior was being lost. The dataframe would be inadvertently collected/materialized, converting lazy Ibis tables to PyArrow tables and Polars LazyFrames to eager DataFrames. This PR refactors the lazy/eager tracking logic into a reusable utility function that properly preserves the original dataframe type.

This PR adds two utilities `make_lazy` and `collect_and_preserve_type` to help with this nuanced logic and then updates usage of this throughout (and some tests).

### DIFF
--- a/tests/_utils/test_narwhals_utils.py
+++ b/tests/_utils/test_narwhals_utils.py
@@ -211,6 +211,7 @@ def test_make_lazy_with_all_dataframes(df: IntoDataFrame) -> None:
     assert type(result_from_collected) is type(df)
 
 
+@pytest.mark.skipif(not HAS_DEPS, reason="optional dependencies not installed")
 def test_undo_preserves_lazy_or_eager_type():
     """Test that the undo callback returns the correct type (eager or lazy) matching the original input."""
     import polars as pl


marimo/_utils/narwhals_utils.py
```diff
@@ -239,3 +243,108 @@ def is_narwhals_dataframe(df: Any) -> TypeIs[nw.DataFrame[Any]]:
...
+    original_backend: nw.Implementation,
+) -> nw.LazyFrame[Any]:
+    if is_narwhals_lazyframe(df):
+        return df
+    else:
+        try:
+            # Try to convert to the original backend, otherwise fallback to default
+            return df.lazy(backend=original_backend)
+        except ValueError:
+            return df.lazy()
```
Light2Dark, line 274:
```suggestion
        except ValueError as e:
            LOGGER.debug("Unable to convert to original backend, converting to default", e)
            return df.lazy()
```

marimo/_utils/narwhals_utils.py
```diff
@@ -239,3 +243,108 @@ def is_narwhals_dataframe(df: Any) -> TypeIs[nw.DataFrame[Any]]:
...
+            LOGGER.warning(
+                "Expected a narwhals DataFrame, got %s", type(result)
+            )
+            return result.lazy()
+
+        try:
+            # Try to convert to the original backend, otherwise fallback to default
+            return result.lazy(backend=original_backend)
+        except ValueError:
+            return result.lazy()
```
Light2Dark: this looks similar to `_to_lazyframe` function

marimo/_utils/narwhals_utils.py
mscolnick: this is common / expected. i can add a note that this is an intentional code path.

for more context: we always pass the implementation, but if its not a lazy impl, then it will throw a ValueError (which is about 50/50 of the implementations)

marimo/_utils/narwhals_utils.py
mscolnick: ah yes, thanks. i made that to DRY this up

Light2Dark (approved):

szst11:
@mscolnick @Light2Dark 
really cool, how fast you made that change. I am impressed!

Light2Dark (approved):

Light2Dark:
> @mscolnick @Light2Dark really cool, how fast you made that change. I am impressed!

glad to hear it! Myles is a beast :)

PR #7406, opened 2025-12-04
improvement: improve scroll position handling when toggling between edit and app mode

mscolnick:
Improve logic when toggling the from edit to present mode (and back) so that the scroll position is adjusted to look at the same output area.


frontend/src/core/layout/useTogglePresenting.ts
```diff
@@ -2,47 +2,102 @@
 
 import { useSetAtom } from "jotai";
 import { useCallback } from "react";
+import { Logger } from "@/utils/Logger";
 import { type CellId, HTMLCellId } from "../cells/ids";
 import { toggleAppMode, viewStateAtom } from "../mode";
 
+const OUTPUT_AREA_CLASS = "output-area";
```
Light2Dark: there's a few places where "output-area" is used as a classname, should this replace those strings too?
I think it makes sense to standardize

Issue #7431, opened 2025-12-08
[ACP] Codex agent fails with 'cwd is not absolute' when notebook is unsaved


lucharo:
## Describe the bug

When using the Codex ACP agent with an unsaved notebook, the agent fails with a "cwd is not absolute" error. Interestingly, Claude ACP works fine in the same scenario.

The `getCwd()` function returns an empty string when no filename is set:

https://github.com/marimo-team/marimo/blob/main/frontend/src/components/chat/acp/agent-panel.tsx#L652-L658

```typescript
function getCwd() {
  const filename = store.get(filenameAtom);
  if (!filename) {
    return "";
  }
  return Paths.dirname(filename);
}
```

It's unclear why Claude handles the empty `cwd` gracefully while Codex does not.

## Environment

This affects the ACP (Agent Client Protocol) integration in the frontend.

## Will you submit a PR?

- [ ] Yes

Light2Dark:
Fixed in #7470 . Thanks!

PR #7432, opened 2025-12-08
feat(acp): add dynamic model selection for ACP agents

lucharo:
## Summary

- Capture available models from agent's `newSession`/`loadSession` response
- Add `ModelSelector` component that displays agent-provided models dynamically
- Call `agent.setSessionModel()` when user changes model selection
- Add `selectedModel` field to `AgentSession` state

Previously, model selection would have required hardcoding model lists per agent. This implementation uses the models returned by the ACP protocol, so it works with any agent that supports the `models` field in their session response.

I couldn't verify whether it works for Codex and for Gemini or OpenCode because I don't have OpenCode installed and Codex doesn't seem to work in dev mode so I couldn't get it to work which is related to issue #7431. My Gemini either hanged or failed because of an `npx` version incompatibility 

## Test plan

- [x] Connect to Claude ACP agent and verify model selector appears with available models (Default, Opus, Haiku)

Screenshot 2025-12-08 at 11 15 11

```txt
[Client → Server]: {
  jsonrpc: '2.0',
  id: 6,
  method: 'session/set_model',
  params: {
    sessionId: 'a1900398-de87-4d8d-a215-3d85c5c55ea5',
    modelId: 'haiku'
  }
}
```

Screenshot 2025-12-08 at 11 16 32

### DIFF
--- a/frontend/src/components/chat/acp/__tests__/state.test.ts
+++ b/frontend/src/components/chat/acp/__tests__/state.test.ts
@@ -87,6 +87,7 @@ describe("state utility functions", () => {
               "createdAt": 1735689600000,
               "externalAgentSessionId": null,
               "lastUsedAt": 1735689600000,
+              "selectedModel": null,
               "title": "New claude session",
             },
           ],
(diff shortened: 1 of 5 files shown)


Light2Dark:
I think this is a nice add. The model selection UI can be closer to the other AI model selector (without the card). Happy to follow up.

mscolnick (approved):

PR #7440, opened 2025-12-09
ux: single-click to create Python cells

### LINKED ISSUE #7371: Please roll back #7235 or add an option to make "+" always add a Python cell
As I understand, #7235 introduced new cell creation buttons (see video in the pull request), so now adding a Python cell is extremely complicated:

1. Click "+". A dropdown will appear... In 99.99% of cases when I click "+" I intend to add a Python cell.
2. Choose "Python cell" in the menu. Thankfully, this option appears right under the cursor, so I don't need to move the mouse.

Alternatively, use Cmd+Click to "auto insert a [Python] cell". Why make this so complicated? It just slows me down. I'm _always_ going to be Cmd+clicking, why add another key to press?

In my opinion, adding a Python cell should be the default, the easiest option, because (unless I'm missing something) marimo is a notebook environment _for Python_: everything else (SQL cells, Markdown, LLMs) is optional and not the main course.

One option is to reverse the current setting:
- Clicking "+" _always_ creates a Python cell. Accessing the main features should require as few clicks as possible.
- Cmd-Clicking brings up the menu that lets you choose what kind of cell to create. Accessing additional options should require additional buttons.

akshayka:
The create cell buttons now default to creating Python cells, since marimo is predominantly a Python notebook. Cmd+click opens a menu for cell types.

Fixes #7371

### DIFF
--- a/frontend/e2e-tests/helper.ts
+++ b/frontend/e2e-tests/helper.ts
@@ -19,13 +19,12 @@ export async function createCellBelow(opts: {
     page.getByTestId("create-cell-button").locator(":visible"),
   ).toHaveCount(2);
 
-  // Clicking the first button opens a dropdown menu with the cell types
+  // Clicking the first button creates a new cell below
   await page
     .getByTestId("create-cell-button")
     .locator(":visible")
     .last()
     .click();
-  await page.getByText("Python cell").click();
   // Type into the currently focused cell
   if (content) {
     await page.locator("*:focus").type(content);


frontend/src/components/editor/cell/CreateCellButton.tsx
```diff
@@ -123,7 +124,7 @@ export const CreateCellButton = ({
         >
           <Tooltip content={finalTooltipContent}>
             <PlusIcon
-              strokeWidth={4}
+              strokeWidth={3}
```
manzt, line 128: I was going to suggest this change separately :)

Light2Dark (approved):

PR #7441, opened 2025-12-09
Unify handling of widget state/buffers/bufferPaths

mscolnick:
This keeps the `state/buffer/bufferPaths` in sync when sending to/from the backend for the widget. This changes some typings and adds a few more tests. This is an improvement, but not the final state as we can hopefully simplify further. This does fix a few errors when sending back binary data.

Prior work done by @manzt in https://github.com/marimo-team/marimo/pull/7441/commits/9426b131008d827f207f85f53274d7f3ee71dac3

### DIFF
--- a/frontend/src/utils/__tests__/data-views.test.ts
+++ b/frontend/src/utils/__tests__/data-views.test.ts
@@ -232,7 +232,7 @@ describe("Immutability Tests", () => {
     const result = serializeBuffersToBase64(input);
 
     // Mutate result state
-    (result.state.nested as any).value = "changed";
+    (result.state.nested as Record<string, unknown>).value = "changed";
 
     // Original nested object should be unchanged
     expect(nested.value).toBe("nested");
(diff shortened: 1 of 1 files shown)


frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx
```diff
@@ -228,15 +267,9 @@ const LoadedSlot = ({
...
 
+  // value is already decoded from wire format
   const model = useRef<Model<T>>(
-    new Model(
-      // Merge the initial value with the current value
-      // since we only send partial updates to the backend
-      { ...data.initialValue, ...value },
-      setValue,
-      functions.send_to_widget,
-      getDirtyFields(value, data.initialValue),
```
Light2Dark, line 238: is `getDirtyFields` intentionally removed? i see the function exists but is no longer used outside of tests.

marimo/_plugins/ui/_impl/from_anywidget.py
```diff
@@ -18,6 +19,47 @@
...
+) -> dict[str, Any]:
+    """Decode wire format { state, bufferPaths, buffers } to plain state with bytes."""
+    if "state" not in wire or "bufferPaths" not in wire:
+        return wire  # Not wire format, return as-is
+
+    state = wire.get("state", {})
+    buffer_paths = wire.get("bufferPaths", [])
+    buffers_base64: list[str] = wire.get("buffers", [])
+
+    if buffer_paths and buffers_base64:
```
Light2Dark, line 48: is it possible for one to be true and the other false? wonder if we should add some logs to check, either here or frontend

frontend/src/plugins/impl/anywidget/AnyWidgetPlugin.tsx
mscolnick: yep it was

marimo/_plugins/ui/_impl/from_anywidget.py
mscolnick: added a log

PR #7446, opened 2025-12-10
fix: handle comprehension shadowing when parsing required references for variables

dmadisetti:
This change fixes a bug in our caching mechanism's parsing of references required for each variable. List comprehension was not catching shadowed references.

This showed up by causing an issue in threaded caching.

### DIFF
--- a/tests/_ast/test_visitor.py
+++ b/tests/_ast/test_visitor.py
@@ -350,8 +350,6 @@ def test_nested_function_param_in_comprehension_not_required_ref() -> None:
 
     assert v.defs == {"helper"}
     assert v.refs == set()  # No external refs!
-    # extension is a PARAMETER, not an external dependency
-    # Compare to test_globals_in_functions: foo(a...) where a is not in required_refs
     assert v.variable_data == {
         "helper": [VariableData(kind="function", required_refs=set())]
     }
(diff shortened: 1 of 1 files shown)


marimo/_ast/visitor.py
```diff
@@ -1006,12 +1006,20 @@ def visit_Name(self, node: ast.Name) -> ast.Name:
...
+        # non-module ancestor block. This prevents function parameters from
+        # being incorrectly marked as refs when used in nested scopes like
+        # list comprehensions, while still capturing module-level references.
         if (
             isinstance(node.ctx, ast.Load)
             and self._is_defined(node.id)
             and node.id not in self.ref_stack[-1]
             and (
-                node.id not in self.block_stack[-1].defs
+                # Check blocks[1:] - skip module block so module-level vars
```
akshayka, line 1018: Want to add a test to `test_visitor` to make sure module-level vars are still tracked as refs?

akshayka (approved):

tests/_ast/test_visitor.py
```diff
@@ -305,6 +305,102 @@ def test_nested_comprehension_generator_with_named_expr() -> None:
...
+    assert v.defs == {"helper"}
+    assert v.refs == set()  # No external refs!
+    # extension is a PARAMETER, not an external dependency
+    # Compare to test_globals_in_functions: foo(a...) where a is not in required_refs
+    assert v.variable_data == {
+        "helper": [VariableData(kind="function", required_refs=set())]
+    }
+
+
+def test_param_in_comprehension_has_required_ref() -> None:
```
dmadisetti, line 358: Out of the tests added, this is the only one that passes without the change (which is expected)- but nice to sanity check for regression

PR #7461, opened 2025-12-10
Always collapse runtime settings into a single dropdown

manzt:
Collapse the runtime settings ("on startup", "on cell change", "on module change") into a single dropdown. Previously this was only collapsed on small screens.

The icon shows yellow (active) when any reactivity setting is enabled, or gray when all are disabled. Click to open the dropdown and see which specific settings are active.


fff.py
mscolnick, line 1: remove file?

manzt: Oops, thanks

akshayka (approved):

foo.py
akshayka, line 1: Delete this file :)

manzt: Thank you :)

akshayka (approved):
A couple of issues to address in follow-up PRs (sidebar dragging requires two clicks when sidebar not in focus, tracing panel inline chart by default when in footer). Nice work!

akshayka (dismissed):

PR #7463, opened 2025-12-10
Move kernel health status into developer panel header

manzt:
The kernel health indicator has moved from the footer into the developer panel header, next to the tabs. This keeps the footer minimal while still making health status accessible when debugging.


manzt:
These changes came already in #7460

PR #7488, opened 2025-12-12
feat: cgroup-aware machine stats

Ddfulton:
## 📝 Summary

When running marimo in containers with resource limits (e.g., docker run --memory=16g), the machine stats footer shows host resources instead of container resources. This PR auto-detects containerized environments with cgroup limits and displays container resources instead.

Before: Shows host memory (512GB) even in a 16GB container
image

After: Shows actual container limits with "container memory" label
image

This is something we've been doing at Carolina Cloud for a while and have found very useful. Not sure whether it belongs in the main source code but figured I'd share to see what the community thinks.

Also did this for CPU % by using a method similar to `psutil` (computing average usage over a time period).

## 🔍 Description of Changes

Backend:
- Added cgroup detection functions (has_cgroup_limits(), get_container_resources()) supporting both v1 and v2
- Fixed missing cgroup v1 CPU detection bug
- Modified /api/usage endpoint to return container stats when limits detected, with is_container flag
Frontend:
- Updated memory stats to show "container memory" vs "computer memory" based on is_container field
Compatibility:
- Only activates on Linux with cgroups (Docker, Kubernetes, Podman)
- Safely falls back to host stats on macOS/Windows or containers without limits

## 📋 Checklist

### DIFF
--- a/.github/workflows/test_cli.yaml
+++ b/.github/workflows/test_cli.yaml
@@ -139,8 +139,7 @@ jobs:
           uv run --no-project pytest -v tests/_cli/test_cli* --maxfail=2
 
   test_examples:
-    needs: [changes, build_wheel]
-    if: ${{ needs.changes.outputs.cli == 'true' }}
+    needs: [changes]
     name: Tests examples and smoke_tests
     runs-on: ubuntu-latest
     timeout-minutes: 3 # 2024-01-18 avg: 0.3m max: 0.5m
@@ -155,16 +154,6 @@ jobs:
         with:
           enable-cache: true
           python-version: 3.12
-
-      - name: Download wheel
-        uses: actions/download-artifact@v4
-        with:
-          name: testwheel
-
-      - name: Install Python deps
-        run: |
-          uv add pytest pytest-asyncio
-          uv pip install marimo*whl
 
       - name: Check example quality with marimo check
         shell: bash
--- a/docs/guides/testing/pytest.md
+++ b/docs/guides/testing/pytest.md
@@ -4,10 +4,10 @@
 
 By default, marimo discovers and executes tests inside your notebook.
 When the optional `pytest` dependency is present, marimo runs `pytest` on cells that
-consist exclusively of test code - i.e. functions whose names start with `test_` or
-classes whose names start with `Test`. If a cell mixes in anything else (helper
-functions, constants, variables, imports, etc.), that cell is skipped by the test
-runner (we recommend you move helpers to another cell).
+consist exclusively of test code - i.e. functions whose names start with `test_`,
+classes whose names start with `Test`, or functions decorated with `@pytest.fixture`.
+If a cell mixes in anything else (helper functions, constants, variables, imports, etc.),
+that cell is skipped by the test runner (we recommend you move helpers to another cell).
 
 For example,
 
@@ -155,3 +155,86 @@ test_notebook.py:17: AssertionError
 FAILED test_notebook.py::test_fails - AssertionError: This test fails
 ========================= 1 failed, 3 passed in 0.82s ==========================
 ```
+
+## Using Pytest Fixtures
+
+
+marimo supports pytest fixtures, with one limitation: fixtures defined in one cell cannot be used in another cell, unless the fixtures were defined in the setup cell.
+For this reason, we recommend defining (or importing) fixtures in your notebook's setup cell, or defining fixtures in a pytest `conftest.py` file.
+
+### Examples
+
+**Fixtures defined in the setup cell**:
+
+```python
+# test_notebook.py
+import marimo
+app = marimo.App()
+
+with app.setup:
+    from fixtures import db_connection, sample_data
+
+@app.cell
+def _(sample_data):
+    def test_data_loaded(sample_data):
+        assert len(sample_data) > 0
+```
+
+**Fixtures in the same cell as tests**:
+
+```python
+@app.cell
+def _():
+    import pytest
+    return pytest
+
+
+@app.cell
+def _(pytest):
+    @pytest.fixture
+    def temp_file():
+        import tempfile
+        with tempfile.NamedTemporaryFile() as f:
+            yield f
+
+    def test_writes_to_file(temp_file):
+        temp_file.write(b"hello")
+        temp_file.seek(0)
+        assert temp_file.read() == b"hello"
+```
+
+**Class fixtures**:
+
+```python
+@app.cell
+def _():
+    import pytest
+    return pytest
+
+
+@app.cell
+def _(pytest):
+    class TestDatabase:
+        @pytest.fixture(scope="class")
+        def connection(self):
+            return create_connection()
+
+        def test_query(self, connection):
+            result = connection.query("SELECT 1")
+            assert result == 1
+```
+
+**`conftest.py` fixtures** work as expected - pytest discovers them automatically.
+
+!!! info "Fixture Limitations"
+
+    Fixtures defined in one cell **cannot** be used by tests in a different cell.
+    This is because pytest collects tests **statically** by parsing the notebook
+    file without executing it. During collection, pytest can see module-level
+    fixtures (from `conftest.py` or imported modules) and fixtures defined in the
+    same scope as the test, but it cannot see fixtures defined in other cells.
+
+    **Why?** Running the entire notebook just for fixture discovery would be
+    expensive, and static analysis cannot determine which fixtures will be
+    available after cell execution since cell order is determined at runtime by
+    marimo's dependency graph.
--- a/marimo/_ast/cell_manager.py
+++ b/marimo/_ast/cell_manager.py
@@ -165,7 +165,7 @@ class CellManager:
     ) -> Cell:
         """Registers cells when called through a context block."""
         cell = context_cell_factory(
-            cell_id=CellId_t(SETUP_CELL_NAME),
+            cell_id=self.setup_cell_id,
             # NB. carry along the frame from the initial call.
             frame=frame,
         )
@@ -221,7 +221,7 @@ class CellManager:
         self, cell_def: CellDef, app: InternalApp | None = None
     ) -> None:
         if isinstance(cell_def, SetupCell):
-            cell_id = CellId_t(SETUP_CELL_NAME)
+            cell_id = self.setup_cell_id
         else:
             cell_id = self.create_cell_id()
         filename = app.filename if app is not None else None
@@ -265,7 +265,7 @@ class CellManager:
         # If this is the first cell, and its name is setup, assume that it's
         # the setup cell.
         if len(self._cell_data) == 0 and name == SETUP_CELL_NAME:
-            cell_id = CellId_t(SETUP_CELL_NAME)
+            cell_id = self.setup_cell_id
         else:
             cell_id = self.create_cell_id()
 
@@ -518,3 +518,7 @@ class CellManager:
     @property
     def seen_ids(self) -> set[CellId_t]:
         return self._cell_id_generator.seen_ids
+
+    @property
+    def setup_cell_id(self) -> CellId_t:
+        return CellId_t(self.prefix + SETUP_CELL_NAME)
--- a/marimo/_ast/compiler.py
+++ b/marimo/_ast/compiler.py
@@ -24,6 +24,7 @@ from marimo._ast.cell import (
     SourcePosition,
 )
 from marimo._ast.names import SETUP_CELL_NAME, TOPLEVEL_CELL_PREFIX
+from marimo._ast.pytest import has_fixture_decorator
 from marimo._ast.transformers import ContainedExtractWithBlock
 from marimo._ast.variables import is_local
 from marimo._ast.visitor import ImportData, Name, ScopedVisitor
@@ -101,7 +102,7 @@ def ends_with_semicolon(code: str) -> bool:
 
 
 def contains_only_tests(tree: ast.Module) -> bool:
-    """Returns True if the module contains only test functions."""
+    """Returns True if the module contains only test functions or fixtures."""
     scope = tree.body
     for node in scope:
         if isinstance(node, ast.Return):
@@ -110,7 +111,9 @@ def contains_only_tests(tree: ast.Module) -> bool:
             node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)
         ):
             return False
-        if not node.name.lower().startswith("test"):
+        if not node.name.lower().startswith(
+            "test"
+        ) and not has_fixture_decorator(node):
             return False
     return bool(scope)
 
--- a/marimo/_ast/pytest.py
+++ b/marimo/_ast/pytest.py
@@ -54,14 +54,17 @@ def build_stub_fn(
         allowed = [arg for arg in args.keys()]
     name = func_body.name
 
-    # Typing checks for mypy
+    # Typing checks for mypy - template structure is known
     fn = copy.deepcopy(PYTEST_BASE)
     body = fn.body[0]
-    assert isinstance(body, ast.FunctionDef)
+    if not isinstance(body, ast.FunctionDef):
+        raise ValueError("Invalid pytest scaffold template structure")
     returned = body.body[-1]
-    assert isinstance(returned, ast.Return)
+    if not isinstance(returned, ast.Return):
+        raise ValueError("Invalid pytest scaffold template structure")
     call = returned.value
-    assert isinstance(call, ast.Call)
+    if not isinstance(call, ast.Call):
+        raise ValueError("Invalid pytest scaffold template structure")
 
     # Using _pytest_scaffold as a template, the resultant function will look
     # like:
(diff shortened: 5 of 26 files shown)


Ddfulton:

frontend/src/components/editor/chrome/wrapper/footer-items/machine-stats.tsx
```diff
@@ -55,8 +55,10 @@ const MemoryUsageBar: React.FC<{
   kernel: UsageResponse["kernel"];
   server: UsageResponse["server"];
 }> = ({ memory, kernel, server }) => {
-  const { percent, total, available } = memory;
+  const { percent, total, available, is_container } = memory;
+  const isContainer = is_container === true;
   const roundedPercent = Math.round(percent);
+  const memoryLabel = isContainer ? "container memory" : "computer memory";
```
Light2Dark: nit
```suggestion
  const roundedPercent = Math.round(percent);
  const memoryLabel = is_container ? "container memory" : "computer memory";
```

Light2Dark (commented):
I think it's a great improvement :). The frontend looks good to me. 

Curious if you have tried this on the backend to use psutil? https://github.com/giampaolo/psutil/issues/1011. api docs: https://psutil.readthedocs.io/en/latest/index.html#psutil.PROCFS_PATH

mscolnick:
this is a good find @Light2Dark . @Ddfulton using https://psutil.readthedocs.io/en/latest/index.html#psutil.PROCFS_PATH should hopefully clean up some code, if you are open to make the change

Ddfulton:
@Light2Dark @mscolnick Thanks for taking a look! I gave `psutil.PROCFS_PATH` a try, but ran into an issue: `psutil` already reads from the container's `/proc` by default, and `/proc/meminfo` shows the **host's** memory, not the container's limits.

I think that [thread](https://github.com/giampaolo/psutil/issues/1011) was actually about the reverse — folks wanting to get **host** stats from inside a container by mounting the host's `/proc` to the container. We need the opposite: **container** limits from inside the container.

When `psutil.virtual_memory()` runs inside a container, it reads `/proc/meminfo` which reflects the host's memory:

```
# Inside container with --memory=1g
psutil.virtual_memory().total     # → 512GB (host, from /proc/meminfo)
open("/sys/fs/cgroup/memory.max") # → 1GB (container limit)
```

I looked at how Docker/Kubernetes tooling handles this and found they read cgroup files directly ([example from cAdvisor](https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L73)). The cgroup paths are part of the stable Linux kernel ABI (v1 since 2008, v2 since 2016), but I agree the file-reading approach is more verbose than ideal.

To make it more robust, I've added:
- `f.close()`, per Copilot's suggestion, for all file operations
- Smoke tests for `has_cgroup_limits()` and `get_container_resources()`
- Rented an Alma linux machine (which has cgroups v1) and tested it there

Light2Dark:
> @Light2Dark @mscolnick Thanks for taking a look! I gave `psutil.PROCFS_PATH` a try, but ran into an issue: `psutil` already reads from the container's `/proc` by default, and `/proc/meminfo` shows the **host's** memory, not the container's limits.

I think the suggestion from the thread is to change `PROCFS_PATH` to something else that contains the container information
```python
import psutil
psutil.PROCFS_PATH = “/path/to/container/procfs”
```

However, looking at some of the threads, it seems there is no standard to do ^, and the host & container are not independent so it would not give the container information. Cmiiw. It's a rabbit hole of an issue lol. If so, I think your solution makes sense..

Separately, I do see some users reporting that they want the reverse...which means it's working correctly? Not sure how it's working for them and not here.

Ddfulton:
With you 100%. I think that's exactly right — a container-specific `/proc` basically doesn't exist which I too initially found hard to believe but I guess that's the nature of how the kernel manages containers? Over my head for sure.

Not familiar with that Schema Breaking Changes Check above and not clear that it's from this PR — let me know if there's anything I can do to resolve it!

marimo/_utils/health.py
```diff
@@ -183,3 +184,188 @@ def communicate_with_timeout(
...
+    Returns:
+        (has_memory_limit, has_cpu_limit): Tuple of booleans indicating
+        whether memory and CPU limits are set.
+    """
+    has_memory = False
+    has_cpu = False
+
+    try:
+        # Check cgroup v2 (modern containers)
+        if os.path.exists("/sys/fs/cgroup/memory.max"):
```
Light2Dark: It'd be better if you can pull out these paths into variables so it's clearer what they are, eg.
```python
cgroup_memory_location = "/sys/fs/cgroup/memory.max"
cgroup_cpu_location = "/sys/fs/cgroup/cpu.max"
```

Light2Dark: looks like they are the same paths in the function below, we can make it a constant at the top of the file

marimo/_utils/health.py
```diff
@@ -183,3 +184,188 @@ def communicate_with_timeout(
...
+    resources: dict[str, Any] = {}
+
+    # Get memory stats if limited
+    if has_memory_limit:
+        try:
+            # Try cgroup v2 first
+            if os.path.exists("/sys/fs/cgroup/memory.max"):
+                with open("/sys/fs/cgroup/memory.max", encoding="utf-8") as f:
+                    memory_max = f.read().strip()
+                    f.close()
```
Light2Dark: Using a context manager, you don't need to manually close the file.

marimo/_server/api/endpoints/health.py
```diff
@@ -193,7 +196,32 @@ async def usage(request: Request) -> JSONResponse:
...
 
-    memory = psutil.virtual_memory()
+    # Check if running in a container with resource limits
+    container_resources = get_container_resources()
+
+    if container_resources and "memory" in container_resources:
+        # Use container memory stats
+        container_mem = container_resources["memory"]
+        memory_stats = {
+            "total": container_mem["total"],
```
Light2Dark: This isn't safe to access, might throw a key error. Could use a TypedDict for typesafety.
```python
class MemoryStats(TypedDict)
  total: int
  available: int
  ...

def get_container_resources() -> MemoryStats:
   ...
```

marimo/_utils/health.py
```diff
@@ -183,3 +184,188 @@ def communicate_with_timeout(
...
+                    "total": total,
+                    "used": used,
+                    "available": available,
+                    "percent": percent,
+                }
+        except (FileNotFoundError, PermissionError, ValueError) as e:
+            LOGGER.debug(f"Error reading container memory stats: {e}")
+
+    # Get CPU stats if limited
+    if has_cpu_limit:
```
dmadisetti: I think you should break these into functions `get_cpu_limit`, 'get_mem_limit`

marimo/_utils/health.py
```diff
@@ -183,3 +184,188 @@ def communicate_with_timeout(
...
         return "", "Error: Process timed out"
+
+
+def has_cgroup_limits() -> tuple[bool, bool]:
+    """
+    Check if cgroup resource limits are explicitly set.
+
+    This is the standard way to detect container resource limits on Linux.
+    These cgroup paths are part of the kernel ABI and are stable across versions.
+    See: https://www.kernel.org/doc/Documentation/cgroup-v2.txt
```
dmadisetti: nice docstring

marimo/_utils/health.py
```diff
@@ -183,3 +184,188 @@ def communicate_with_timeout(
...
+         https://www.kernel.org/doc/Documentation/cgroup-v1/
+
+    Returns:
```
dmadisetti, line 240: `has_cpu_limit()`, `has_mem_limit()` might make this a bit cleaner

Ddfulton:
@Light2Dark 
>cmiiw, It seems the 2 functions you created do similar jobs like reading the same paths. Instead, is it possible to create 2 
functions

Did you mean one function? I merged them into one function each for memory and cpu: `get_cgroup_cpu_limit` and `get_cgroup_mem_stats`. First iteration of this PR was still not cgroup-aware for CPU, this one is.

>It looks like your method to find if containerized is to check for the cgroup location, which is correct I think. I see other implementations do a os.path.exists("/.dockerenv") but that only works for docker

Right. A quick is_docker check would be a good opportunity to exit this flow early, but you're right it only works for docker. There's also linux containers, kata containers (which we are use at Carolina Cloud) and podman (which I have never used). Might even be more. The more I researched this the more I realized we don't actually care whether it's in a container, we really only care if there's a cgroup limit which is just highly correlated with being in a container. 

>looks like they are the same paths in the function below, we can make it a constant at the top of the file

✅ 

>Using a context manager, you don't need to manually close the file.

✅ 

>This isn't safe to access, might throw a key error. Could use a TypedDict for typesafety.

✅ 

@dmadisetti 
>I think you should break these into functions get_cpu_limit, 'get_mem_limit`

Broke into `get_cgroup_mem_stats` and `get_cgroup_cpu_percent`.

>has_cpu_limit(), has_mem_limit() might make this a bit cleaner

I tried that but realized that to check whether the cgroup limit exists or not required reading the same files that contain the actual cgroup limits itself. In other words, reading the limit is basically the same thing as checking for whether the limit exists. Ultimately just included these in `get_cgroup_mem_stats()` and `get_cgroup_cpu_percent()`. Another reason was that in python usually the word "get" means there's a possibility of returning None, ie `dictionary.get('key')`. This way in the usage endpoint code we can do this:

```
    if cgroup_mem_stats := get_cgroup_mem_stats():
        memory_stats = {
            "has_cgroup_mem_limit": True,
            **cgroup_mem_stats,
        }
    else:
        # use psutil

    cpu_percent = get_cgroup_cpu_percent()
    if cpu_percent is None:
        # interval=None is nonblocking; first call returns meaningless value
        # subsequent calls return delta since last call
        cpu_percent = psutil.cpu_percent(interval=None)
```

They were reading a lot of the same files and I was only calling the `_has_mem_limit` and `_has_cpu_limit` function in the `get_cgroup_...` funcs themselves.

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+
+        last_usage, last_time = _last_cgroup_cpu_sample
+        _last_cgroup_cpu_sample = (current_usage, current_time)
+
+        delta_time = current_time - last_time
+        if delta_time <= 0:
+            return 0.0
+
+        delta_usage_usec = current_usage - last_usage
+        # percent = (usage_usec / (time_sec * 1_000_000 * cores)) * 100
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+        percent = (
+            delta_usage_usec / (delta_time * 1_000_000 * allocated_cores)
+        ) * 100
```
Light2Dark, line 379:
```suggestion
        delta_time_microseconds = delta_time * 1_000_000
        percent = (
             delta_usage_microseconds / (delta_time_in_microseconds * allocated_cores)
        ) * 100
```

marimo/_utils/health.py
```diff
@@ -2,16 +2,43 @@
...
+from typing import Optional, TypedDict
 
 from marimo import _loggers
 
 LOGGER = _loggers.marimo_logger()
 
 TIMEOUT = 10  # seconds
 
+# Module-level state for cgroup CPU percent calculation (like psutil does)
+_last_cgroup_cpu_sample: Optional[tuple[int, float]] = (
```
Light2Dark: okay to make this capitalized like other parts in the codebase
```suggestion
_LAST_CGROUP_CPU_SAMPLE: Optional[tuple[int, float]] = (
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+    """
+    global _last_cgroup_cpu_sample
+
+    # Early return if no CPU limit is configured
+    if not _has_cgroup_cpu_limit():
+        return None
+
+    try:
+        # Read current usage (microseconds)
+        current_usage: Optional[int] = None
```
Light2Dark:
```suggestion
        current_usage_microseconds : Optional[int] = None
```

marimo/_utils/health.py
```diff
@@ -2,16 +2,43 @@
...
+
+class MemoryStats(TypedDict):
+    total: int
+    used: int
+    available: int
+    percent: float
+    free: int
+
+
+# Constants for cgroup file locations
```
Light2Dark: I think you had a reference implementation / codebase for this logic, that would be a good comment to add (if you have)
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
         process.kill()
         return "", "Error: Process timed out"
+
+
+def _has_cgroup_cpu_limit() -> bool:
+    """
+    Returns True/False whether the container has a CPU limit set.
+    This function checks for both cgroups v1 and v2.
+    """
+    # cgroups v2
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+        {
+            'total': 2147483648,      # bytes
+            'used': 1073741824,      # bytes
+            'available': 1073741824, # bytes
+            'free': 1073741824,      # bytes
+            'percent': 50.0,         # percentage
+        }
+    """
+    try:
+        # Try cgroup v2 first
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+                available = total - used
+                percent = (used / total) * 100 if total > 0 else 0
+                return MemoryStats(
+                    total=total,
+                    used=used,
+                    available=available,
+                    percent=percent,
+                    free=available,  # free == available for cgroup memory
+                )
+        # Fallback to cgroup v1
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+        LOGGER.debug(f"Error reading container memory stats: {e}")
+
+    return None
+
+
+def _get_cgroup_allocated_cores() -> Optional[float]:
+    """Get the number of CPU cores allocated to this cgroup (quota / period)."""
+    try:
+        if os.path.exists(CGROUP_V2_CPU_MAX_FILE):
+            # cgroups v2
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+            with open(CGROUP_V2_CPU_MAX_FILE, encoding="utf-8") as f:
+                parts = f.read().strip().split()
+            if len(parts) == 2 and parts[0] != "max":
+                return int(parts[0]) / int(parts[1])
+        elif os.path.exists(CGROUP_V1_CPU_CFS_QUOTA_US_FILE):
+            # cgroups v1
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+    # Early return if no CPU limit is configured
+    if not _has_cgroup_cpu_limit():
+        return None
+
+    try:
+        # Read current usage (microseconds)
+        current_usage: Optional[int] = None
+
+        if os.path.exists(CGROUP_V2_CPU_STAT_FILE):
+            # cgroups v2
```
Light2Dark:
```suggestion
```

marimo/_utils/health.py
```diff
@@ -183,3 +210,173 @@ def communicate_with_timeout(
...
+            with open(CGROUP_V2_CPU_STAT_FILE, encoding="utf-8") as f:
+                for line in f:
+                    if line.startswith("usage_usec"):
+                        current_usage = int(line.split()[1])
+                        break
+
+        elif os.path.exists(CGROUP_V1_CPU_USAGE_FILE):
+            # cgroups v1
```
Light2Dark:
```suggestion
```

packages/openapi/src/api.ts
```diff
@@ -2663,6 +2663,7 @@ export interface paths {
               memory: {
                 available: number;
                 free: number;
+                has_cgroup_mem_limit: boolean;
```
Light2Dark: if this was just `containerized` would that be accurate?

Light2Dark (approved):
tq! 🚀. you have some failing imports (see CI)

Ddfulton:
Fixed the test imports. Bear with me, looking into the sql tutorial failure 🤦

Ddfulton:
Re CI: Tutorial check and and `tests/_server/api/endpoints/test_terminal.py` failed. I lack the broader project context to fix these and it's not clear to me if they are related to my changes, any thoughts @Light2Dark?

mscolnick:
@Ddfulton yes there are unrelated failing tests

dmadisetti (approved):
Test looks unrelated (I'll look into it)

Thanks for your perseverance and happy New Year!

Light2Dark:
Echoing Dylan—thanks for this! And sorry for all the reviews; the PR looks great now.

Ddfulton:
No problem at all and thanks for working with me on this one. Happy to contribute as a lot of our users use Marimo in containers. Please don't hesitate to ping me derek@carolinacloud.io if you ever get any bugs with this.

PR #7490, opened 2025-12-12
tests: correct imports to actual file instead of accidental exports

mscolnick:
I have a script (which I may put in CI in a followup) that detects accidentally importing symbols that are exported from others files (not intentionally, e.g. not `__all__`).

This should reduce some of the module dep graph (hopefully reducing some extra tests picked up in the changeset)


dmadisetti (approved):
Test failure looks unrelated

PR #7494, opened 2025-12-13
fix: app embed parametrization

akshayka:
Invalidate `AppKernelRunner`'s output cache when definitions change.

This change also updates `embed()` to raise a runtime error when attempting to call `embed()` in the same cell that defines the app, similar to UI elements.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -809,12 +809,19 @@ class App:
 
             ```python
             one = app.clone()
+            ```
+
+            ```python
             r1 = await one.embed()
             ```
 
             ```python
             two = app.clone()
+            ```
+
+            ```python
             r2 = await two.embed()
+            ```
 
         Args:
             defs (dict[str, Any]):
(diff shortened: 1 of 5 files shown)


akshayka:
Failing tests unrelated

marimo/_ast/app.py
```diff
@@ -830,18 +837,42 @@ async def embed(
...
+            )
+
         if running_in_notebook():
-            # TODO(akshayka): raise a RuntimeError if called in the cell
-            # that defined the name bound to this App, if any
+            ctx = get_context()
+            bound_variable: str | None = None
+            for k, v in ctx.globals.items():
+                if v is self:
+                    bound_variable = k
```
dmadisetti, line 857:
```suggestion
                    bound_variable = k
                    break
```

or remove bound variable and indent

marimo/_runtime/app/kernel_runner.py
```diff
@@ -113,6 +114,13 @@ def cache_output(
     def outputs(self) -> dict[CellId_t, Any]:
         return self._outputs
 
+    def are_outputs_cached(self, defs: dict[str, Any] | None) -> bool:
+        # The equality check is brittle but hashing isn't great either ...
+        return (defs == self._previously_seen_defs) and len(self.outputs) > 0
```
dmadisetti, line 119: Very brittle but distinguishes between `no` args and some args. Esp with a wrapper like mowidget, this should be fine

dmadisetti (approved):

PR #7507, opened 2025-12-15
feat: copy as markdown in the dataframe table

mscolnick:
Add a "Copy as Markdown" to the table to easily paste into markdown docs.

### DIFF
--- a/frontend/src/utils/__tests__/json-parser.test.ts
+++ b/frontend/src/utils/__tests__/json-parser.test.ts
@@ -290,3 +290,39 @@ it("can convert json to markdown - handles different data types", () => {
     | {"nested":"value"} |"
   `);
 });
+
+it("can convert json to markdown - handles existing markdown links", () => {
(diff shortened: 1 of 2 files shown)


frontend/src/utils/__tests__/json-parser.test.ts
```diff
@@ -131,3 +135,158 @@ it("handles NaN values in TSV", () => {
...
+it("can convert json to markdown - with URLs", () => {
+  expect(
+    jsonToMarkdown([
+      { name: "Google", url: "https://google.com" },
+      { name: "GitHub", url: "https://github.com" },
+    ]),
+  ).toMatchInlineSnapshot(`
+    "| name | url |
+    |---|---|
+    | Google | [https://google.com](https://google.com) |
```
manzt, line 182: Maybe want to include a test of the expected behavior when there is a markdown link already? I would assume we generally wouldn't expect those inputs, but it would result in a nested link.

manzt (approved):
Nice feature!

PR #7514, opened 2025-12-15
poc: reorderable sidebar

mscolnick:
poc to reorder and hide/remove elements from the sidebar

### DIFF
--- a/frontend/src/components/ui/reorderable-list.tsx
+++ b/frontend/src/components/ui/reorderable-list.tsx
@@ -1,9 +1,8 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
 import type React from "react";
-import { useEffect, useMemo, useState } from "react";
+import { useMemo } from "react";
 import { ListBox, ListBoxItem, useDragAndDrop } from "react-aria-components";
-import { useListData } from "react-stately";
 import {
   ContextMenu,
   ContextMenuCheckboxItem,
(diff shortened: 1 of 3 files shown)


frontend/src/components/ui/reorderable-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+  //   if (isEqual(list.items, value)) {
+  //     return;
+  //   }
+
+  //   list.setSelectedKeys(new Set());
+  //   list.remove(...list.items.map((item) => item.id));
+  //   for (const item of value) {
+  //     list.append(item);
+  //   }
+  // }, [value, list]);
```
manzt: delete?

frontend/src/components/ui/reorderable-list.tsx
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * <ReorderableList
+ *   value={items}
+ *   setValue={setItems}
+ *   renderItem={(item) => <div>{item.name}</div>}
+ *   ariaLabel="My reorderable list"
+ * />
+ * ```
+ */
+export function ReorderableList<T extends { id: string | number }>({
```
manzt, line 70: I'm wondering if we could consider simplifying this by not using `useListData` at all? Since we already have external state via `value`/`setValue`, the internal list state feels like it's adding complexity. Especially the `needsSync` boolean and deferred `useEffect` to work around the timing issues.

Something like:

```tsx
export function ReorderableList<T extends { id: string | number }>({
  value,
  setValue,
  renderItem,
  availableItems,
  getItemLabel = (item) => String(item.id),
  minItems = 1,
  ariaLabel = "Reorderable list",
  className,
}: ReorderableListProps<T>) {
  const { dragAndDropHooks } = useDragAndDrop<T>({
    getItems: (keys) => [...keys].map((key) => ({ "text/plain": String(key) })),
    onReorder(e) {
      const keySet = new Set(e.keys);
      const draggedItems = value.filter((item) => keySet.has(item.id));
      const remaining = value.filter((item) => !keySet.has(item.id));

      const targetIndex = remaining.findIndex((item) => item.id === e.target.key);
      const insertIndex = e.target.dropPosition === "before" ? targetIndex : targetIndex + 1;

      setValue([
        ...remaining.slice(0, insertIndex),
        ...draggedItems,
        ...remaining.slice(insertIndex),
      ]);
    },
  });

  const handleToggleItem = (item: T, isChecked: boolean) => {
    if (isChecked) {
      setValue([...value, item]);
    } else if (value.length > minItems) {
      setValue(value.filter((v) => v.id !== item.id));
    }
  };

  // ... rest stays the same, but use `value` instead of `list.items` in ListBox
```

Could trop `react-stately` as well.

manzt (commented):
Nice, just a small comment in implementation of `ReorderableList`.

frontend/src/components/ui/reorderable-list.tsx
manzt: @mscolnick what do you think? I can take it from here rebase merge into my original PR, and update all the other branches.

PR #7525, opened 2025-12-16
fix: detect !pip and give better error message

dmadisetti:
## 📝 Summary

Intercept syntax errors to provide more context if it seems as if there may be some improper jupyter-like usage.

Handles:
- magics `%`
- installations `! pip`
- General bang commands `!ls`

image

### DIFF
--- a/tests/_runtime/test_runtime.py
+++ b/tests/_runtime/test_runtime.py
@@ -949,6 +949,49 @@ except NameError:
         # Python reports line numbers starting from 1
         assert k.errors["0"][0].lineno >= 1
 
+    async def test_syntax_error_hint_pip_install(
+        self, any_kernel: Kernel
+    ) -> None:
+        """Test that !pip commands show package manager hint"""
+        k = any_kernel
+        await k.run(
+            [
+                ExecutionRequest(cell_id="0", code="!pip install pandas"),
+            ]
+        )
+        assert set(k.errors.keys()) == {"0"}
+        assert isinstance(k.errors["0"][0], MarimoSyntaxError)
+        assert "package manager panel" in k.errors["0"][0].msg
+
+    async def test_syntax_error_hint_shell_command(
+        self, any_kernel: Kernel
+    ) -> None:
+        """Test that shell commands (not pip) show os.subprocess hint"""
+        k = any_kernel
+        await k.run(
+            [
+                ExecutionRequest(cell_id="0", code="!ls -la"),
+            ]
+        )
+        assert set(k.errors.keys()) == {"0"}
+        assert isinstance(k.errors["0"][0], MarimoSyntaxError)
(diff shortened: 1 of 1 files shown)


marimo/_runtime/runtime.py
```diff
@@ -881,8 +881,25 @@ def _try_compiling_cell(
...
                     error = MarimoImportStarError(msg=str(e), lineno=lineno)
                 else:
+
+                    def _get_syntax_hints(broken_line: str) -> str:
+                        if broken_line.startswith("!"):
+                            if "pip" in broken_line:
+                                return (
+                                    "\nHint: Use the package manager to add a "
+                                    "dependency to the notebook"
+                                )
```
akshayka: Can we do this in the frontend, and make it so that clicking "package manager" opens the package manager panel? Or at least opens docs?

marimo/_runtime/runtime.py
```diff
@@ -881,8 +881,25 @@ def _try_compiling_cell(
...
+                            return "\nHint: Use os.subprocess to run commands"
+                        elif broken_line.startswith("%"):
+                            return (
+                                "\nHint: marimo doesn't support Jupyter magics, "
+                                "use a decorator pattern instead"
+                            )
```
akshayka, line 894: I'm not sure what a decorator pattern is. Instead, I would write:

marimo notebooks are pure Python and don't support IPython's magic commands. See [our documentation](LINK) for workarounds.

You can use links.marimo.app to create the link to https://docs.marimo.io/guides/coming_from/jupyter/#magic-commands

marimo/_runtime/runtime.py
dmadisetti: So right now, typing `!pip` opens the panel and creates a toast, but we can have it work on eval too

marimo/_runtime/runtime.py
dmadisetti: Changed

dmadisetti:
I was going to move the errors client side, but realized this captures VS code too

marimo/_runtime/runtime.py
```diff
@@ -881,8 +881,24 @@ def _try_compiling_cell(
...
+
+                    def _get_syntax_hints(broken_line: str) -> str:
+                        if broken_line.startswith("!"):
+                            if "pip" in broken_line:
+                                return "\nTo install packages, use the package manager panel."
+                            return "\nTo run shell commands, use os.subprocess(...)"
+                        elif broken_line.startswith("%"):
+                            return (
+                                "\nIPython magic commands (starting with %) are not supported."
+                                "\nSee: http://links.marimo.app/from-jupyter-magics"
```
akshayka:
```suggestion
                                "\nSee: https://links.marimo.app/from-jupyter-magics"
```

https?

PR #7527, opened 2025-12-17
refactor: split up DirectedGraph into smaller components

mscolnick:
This is a refactor of the DirectedGraph class into smaller components: GraphTopology, CycleTracker, DefinitionRegistry, EdgeComputer, etc.
These changes aim to improve code maintainability and clarity in the dataflow graph structure.

Only some tests were added since we still test the DirectedGraph class.

### DIFF
--- a/marimo/_runtime/dataflow/runner.py
+++ b/marimo/_runtime/dataflow/runner.py
@@ -9,7 +9,7 @@ from marimo._runtime.executor import ExecutionConfig, get_executor
 
 if TYPE_CHECKING:
     from marimo._ast.cell import CellImpl
-    from marimo._runtime.dataflow.coordinator import DirectedGraph
+    from marimo._runtime.dataflow.graph import DirectedGraph
     from marimo._types.ids import CellId_t
 
 


marimo/_runtime/dataflow/runner.py
```diff
@@ -0,0 +1,154 @@
...
+
+from marimo._runtime.executor import ExecutionConfig, get_executor
+
+if TYPE_CHECKING:
+    from marimo._ast.cell import CellImpl
+    from marimo._runtime.dataflow.coordinator import DirectedGraph
+    from marimo._types.ids import CellId_t
+
+
+class Runner:
```
dmadisetti, line 16: Runner should probably be removed. It's only used in an `app.run` context (which IMO should be moved over to embedded ScriptRunner)

marimo/_runtime/dataflow/graph.py
```diff
@@ -0,0 +1,441 @@
+# Copyright 2024 Marimo. All rights reserved.
```
dmadisetti, line 1: why `coordinator.py` Thoughts on just `graph`?

marimo/_runtime/dataflow/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+                in interrupted_states
+            )
+
+        return children_ids
+
+    return import_block_relatives
+
+
+__all__ = [
+    "DirectedGraph",
```
dmadisetti: Dupe

marimo/_runtime/dataflow/topology.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def get_path(self, source: CellId_t, dst: CellId_t) -> list[Edge]:
+        """Get a path from `source` to `dst`, if any.
+
+        Returns an empty list if source == dst or if no path exists.
+        """
+        if source == dst:
+            return []
+
+        # BFS to find path
```
dmadisetti, line 137: basically an interview question

marimo/_runtime/dataflow/graph.py
```diff
@@ -0,0 +1,441 @@
...
+    cycle_tracker: CycleTracker = field(default_factory=CycleTracker)
+
+    # Cells that define the same name
+    #
+    # siblings[cell_id] is a set of cell ids, one for each cell that shares a
+    # definition with cell_id.
+    #
+    # If this dict is non-empty, then the marimo program contains multiply
+    # defined names (and is therefore in an error state)
+    siblings: dict[CellId_t, set[CellId_t]] = field(default_factory=dict)
```
dmadisetti, line 58: siblings should be a graphtopology property, i think it's misplaced here

marimo/_runtime/dataflow/graph.py
```diff
@@ -0,0 +1,441 @@
...
+    - Delegate to specialists
+    """
+
+    topology: MutableGraphTopology = field(
+        default_factory=MutableGraphTopology
+    )
+    definition_registry: DefinitionRegistry = field(
+        default_factory=DefinitionRegistry
+    )
+    cycle_tracker: CycleTracker = field(default_factory=CycleTracker)
```
dmadisetti, line 48: cycle tracker almost feels like it should be a private class to topology. It has state, but can't really be decoupled from a given topology

marimo/_runtime/dataflow/edges.py
```diff
@@ -0,0 +1,269 @@
...
+    from marimo._runtime.dataflow.definitions import DefinitionRegistry
+    from marimo._runtime.dataflow.topology import GraphTopology
+    from marimo._types.ids import CellId_t
+
+from marimo._ast.sql_visitor import SQLTypes
+
+LOGGER = _loggers.marimo_logger()
+
+
+class EdgeComputer:
```
dmadisetti, line 23: In implementation used as basically a namespace. I think it's worth just exposing as just functions opposed to a class

marimo/_runtime/dataflow/runner.py
dmadisetti: But fine for now, just a refactor I've been thinking of

marimo/_runtime/dataflow/graph.py
mscolnick: yea having state in this feels like a bit of a smell. i may handle this in a followup and think how this fits in. i want the topology to be pretty slim though

marimo/_runtime/dataflow/graph.py
mscolnick: this seems not used at all. ill remove it in a followup

marimo/_runtime/dataflow/edges.py
mscolnick: the refactor made the writing of the class args a bit more verbose (passing them all the way through multiple levels), so i plan to leave this as-is, but i can revisit

dmadisetti: It could live on DirectedGraph. `get_transitive_references` falls into this wheelhouse too

PR #7533, opened 2025-12-17
refactor: convert edges class to util functions, remove unused siblings

mscolnick:
* Convert `EdgeComputer` to just utility functions
* Remove siblings as it was not used at all

### DIFF
--- a/marimo/_runtime/dataflow/graph.py
+++ b/marimo/_runtime/dataflow/graph.py
@@ -47,15 +47,6 @@ class DirectedGraph(GraphTopology):
     )
     cycle_tracker: CycleTracker = field(default_factory=CycleTracker)
 
-    # Cells that define the same name
-    #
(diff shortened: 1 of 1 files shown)


marimo/_runtime/dataflow/graph.py
```diff
@@ -54,7 +54,6 @@ class DirectedGraph(GraphTopology):
     #
     # If this dict is non-empty, then the marimo program contains multiply
     # defined names (and is therefore in an error state)
-    # TODO: remove this field (might be unused)
     siblings: dict[CellId_t, set[CellId_t]] = field(default_factory=dict)
```
dmadisetti: Is this good to go? still have this property

mscolnick: yep, thanks i forgot to remove it

PR #7534, opened 2025-12-17
refactor: split up SessionManager

mscolnick:
This PR split up the SessionManager and introduces a system for handling session lifecycle events (`SessionEventListener`) for filewatch and recent notebooks. Also cleans up session resumption strategies based on read/edit mode. 

The changes decouple concerns, improve maintainability, and enable easier extension for new behaviors around session management.

**Session Lifecycle Event System:**

* Added a new event system for session lifecycle events, including a `SessionEventBus` and `SessionEventListener`, allowing components to subscribe to and react to session creation, closure, and resumption without tight coupling.

**File Change Handling and Watcher Integration:**

Changes to decouple the watching and on-change to make it more testable and integrate with the session lifecyle.

**Session Resume Strategies:**

* Implemented `resume_strategies.py` to encapsulate mode-specific logic for resuming sessions, with separate strategies for edit and run modes.

**Reusable Event Listeners:**

* Added `listeners.py` with a `RecentsTrackerListener` that updates the recent files list on session creation and resumption. This is future looking as we add more.

### DIFF
--- a/marimo/_server/notebook/file_manager.py
+++ b/marimo/_server/notebook/file_manager.py
@@ -99,8 +99,10 @@ class AppFileManager:
             Set of cell IDs that were added, deleted, or modified
         """
         prev_cell_manager = self.app.cell_manager
-        self.app = self._load_app(self.path)
-        self.app.cell_manager.sort_cell_ids_by_similarity(prev_cell_manager)
+        new_app = self._load_app(self.path)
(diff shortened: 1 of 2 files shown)


dmadisetti (dismissed):
Bones look great, a little wary because file watching has been tricky. Would recommend for staging if we had it

PR #7538, opened 2025-12-17
Dm/dataflow

dmadisetti:


marimo/_runtime/dataflow/graph.py
```diff
@@ -412,4 +534,24 @@ def definitions(self) -> Mapping[Name, set[CellId_t]]:
     @property
     def cycles(self) -> set[tuple[tuple[CellId_t, CellId_t], ...]]:
         """Get the cycles set."""
-        return self.cycle_tracker.cycles
+        return self.topology.cycles
+
+
+# Used over _MarimoGraph for backwards compatibility
+class DirectedGraph(_MarimoGraph):
```
akshayka, line 541: This feels excessive to me. I would recommend renaming `_MarimoGraph` to just be `DirectedGraph` as it was before

marimo/_runtime/dataflow/dag.py
```diff
@@ -0,0 +1,152 @@
...
+
+from marimo._runtime.dataflow.topology import MutableGraphTopology
+from marimo._runtime.dataflow.types import Edge
+
+if TYPE_CHECKING:
+    from marimo._types.ids import CellId_t
+
+
+@dataclass
+class MutableDirectedGraph(MutableGraphTopology):
```
mscolnick, line 17: prefer composition over inheritance

marimo/_runtime/dataflow/dag.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Could be abstracted further with CellId_t -> Node
+    # and "disabled", "stale" -> colourings
+    # with distinctions between DG/ coloured DG.
+    # However, this is sufficient for our purposes
+
+    _cycles: set[tuple[Edge, ...]] = field(default_factory=set)
+
+    @property
+    def cycles(self) -> set[tuple[Edge, ...]]:
```
mscolnick, line 33: this `set` should be immutable/readonly

marimo/_runtime/dataflow/graph.py
```diff
@@ -5,47 +5,79 @@
...
+    # For SQL hierarchical references, find the matching variable name
+    for _, matching_variable_name in sql_matches:
+        if matching_variable_name in other_cell.variable_data:
+            return matching_variable_name
+
+    return name
+
+
 @dataclass(frozen=True)
-class DirectedGraph(GraphTopology):
```
manzt, line 32: Just noting that we import `DirectedGraph` in marimo-lsp. Should be ok since we pin marimo upper bound in our rolling releases now, but wanted to flag.

dmadisetti:
@mscolnick pointed out that the mutation of cells really doesn't fall under pure data structure.

I do think this is a cleaner pattern, but to make the distinction I wanted (just graph primitives v "marimo-informed") would require a larger refactor I don't think justifies even the time

PR #7539, opened 2025-12-17
Add empty markdown cell placeholder

manzt:
Empty hidden markdown cells now show "Double-click (or enter) to edit" instead of being invisible. The placeholder only appears after the cell has been run.

This was trickier than anticipated. The raw `cellData.code` always contains the `mo.md(...)` wrapper, so checking for empty code doesn't work. Instead, detection reads the editor's inner content directly (the transformed markdown that the user actually edits).

https://github.com/user-attachments/assets/ae9b4b29-382f-43cf-9827-6a047e2118b7

### DIFF
--- a/frontend/src/components/editor/notebook-cell.tsx
+++ b/frontend/src/components/editor/notebook-cell.tsx
@@ -487,6 +487,8 @@ const EditableCellComponent = ({
     isEmptyMarkdownContent &&
     !needsRun && (
       <div
+        role="button"
+        aria-label="Double-click to edit markdown"
         className="relative cursor-pointer px-3 py-2"
         onDoubleClick={showHiddenCodeIfMarkdown}
         onKeyDown={(e) => {


akshayka (approved):
Beautiful!

frontend/src/components/editor/notebook-cell.tsx
```diff
@@ -478,7 +483,26 @@ const EditableCellComponent = ({
...
+        className="relative cursor-pointer px-3 py-2"
+        onDoubleClick={showHiddenCodeIfMarkdown}
+        onKeyDown={(e) => {
+          if (e.key === "Enter") {
+            showHiddenCodeIfMarkdown();
+          }
+        }}
+        tabIndex={0}
+      >
+        <span className="text-[var(--slate-8)] text-sm">
```
Light2Dark, line 501:
```suggestion
        <span className="text-(--slate-8) text-sm">
```

frontend/src/components/editor/notebook-cell.tsx
```diff
@@ -478,7 +483,26 @@ const EditableCellComponent = ({
...
+    !needsRun && (
+      <div
+        className="relative cursor-pointer px-3 py-2"
+        onDoubleClick={showHiddenCodeIfMarkdown}
+        onKeyDown={(e) => {
+          if (e.key === "Enter") {
+            showHiddenCodeIfMarkdown();
+          }
+        }}
+        tabIndex={0}
```
Light2Dark, line 499: Using tabIndex this raises a biome warning on a11y. Would it be better to  use a button
```ts
      <Button
        variant="ghost"
        size="sm"
        className="relative hover:bg-transparent border-none hover:shadow-none"
        onDoubleClick={showHiddenCodeIfMarkdown}
        onKeyDown={(e) => {
          if (e.key === "Enter") {
            showHiddenCodeIfMarkdown();
          }
        }}
      >
        <span className="text-(--slate-8)">
          Double-click (or enter) to edit
        </span>
      </Button>
```

frontend/src/components/editor/notebook-cell.tsx
Light2Dark: actually, i would darken it a touch `--slate-9` for it to be more visible, but im okay either way.

manzt: i chose `--slate-8` to match the ghost text of "Start coding...". Should we consider darkening that too?

frontend/src/components/editor/notebook-cell.tsx
manzt: I'm not totally sure, but I don't think a button would be a good alternative here (unless I'm missing something) - it would activate on single-click and Space, but this is intentionally double-click only. Maybe need to ignore the warning?

Light2Dark: It wouldn't activate on enter/space, since you've configured it onDoubleclick, there's no onClick handler.

An alternative is to add a role attribute & aria-label

frontend/src/components/editor/notebook-cell.tsx
Light2Dark: ah, okay, fine to keep it then, thanks!

frontend/src/components/editor/notebook-cell.tsx
manzt: Ah I see, thanks for clarifying. I added role attribute and aria-label.

PR #7542, opened 2025-12-18
internal: session extensions

mscolnick:
This refactors the `Session` class into extensions that attach/detach and can optionally listen to session events. This is the start of the refactor, but by the end, it should essentially just be an event bus. This really reduces the size of the session (and conditional logic) making it more composable and re-usable for different use cases (marimo-lsp) and different condition (read/edit, autoinstantiate/not)

### DIFF
--- a/tests/_server/test_server_utils.py
+++ b/tests/_server/test_server_utils.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 from marimo._server.api.utils import parse_title
 
 
-# The function to be tested
 def test_parse_title() -> None:
     assert parse_title(None) == "marimo"
     assert parse_title("example_file.txt") == "example file"


marimo/_server/sessions/events.py
```diff
@@ -6,28 +6,48 @@
...
 
+if TYPE_CHECKING:
+    from marimo._server.sessions.session import Session
 
-class SessionEventListener(Protocol):
-    """Protocol for session event listeners."""
+LOGGER = _loggers.marimo_logger()
+
+
+class SessionEventListener:
```
dmadisetti, line 20: Is this no longer a protocol?

marimo/_server/sessions/extensions.py
```diff
@@ -0,0 +1,221 @@
...
+            codes=codes, marimo_version=__version__, cell_ids=cell_ids
+        )
+        session.session_view = self.session_cache_manager.read_session_view(
+            key
+        )
+
+        # Start the background task to write the session view to disk
+        self.session_cache_manager.start()
+
+    def on_detach(self) -> None:
```
Light2Dark, line 165: maybe poor understanding, but on detach, wouldn't we unsubscribe from event_bus?

marimo/_server/sessions/events.py
mscolnick: yea i shouldve mentioned. this is so that implementors don't need to implement all the fields

marimo/_server/sessions/extensions.py
mscolnick: yes nice catch, we should. the event bus gets destroyed when this detaches anyways, but good to be safe.

dmadisetti (approved):

marimo/_server/sessions/extensions.py
mscolnick: going to do in a followup

PR #7555, opened 2025-12-19
internal: internal api for template rendering

mscolnick:
Internal API for rendering templates. Not for public use as API may change


dmadisetti:
I would put you and @akshayka as code owners

PR #7556, opened 2025-12-19
feat: CLI flag to configure whether a notebook is trusted

akshayka:
This PR adds a new flag to `marimo edit` and `marimo run`: `--trusted`. This flag controls whether marimo is run in a Docker container.

* `marimo edit --trusted ...`: does not run in a container
* `marimo edit --untrusted ...`: runs in a Docker container
* `marimo edit <NOTEBOOK>`: prompts whether to run in Docker if `NOTEBOOK` is determined to be hosted remotely; this behavior is the same as it was before this change.

This enables users to run trusted notebooks that are hosted remotely on the host machine, without having to respond interactive prompts. For example, `uvx marimo edit --trusted --sandbox URL`

### DIFF
--- a/marimo/_cli/cli.py
+++ b/marimo/_cli/cli.py
@@ -340,12 +340,12 @@ requirements. These and other issues are described in
 https://github.com/marimo-team/marimo/issues/5219.""",
 )
 @click.option(
-    "--trusted/--no-trusted",
+    "--trusted/--untrusted",
     is_flag=True,
     default=None,
     show_default=False,
     type=bool,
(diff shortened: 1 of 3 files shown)


marimo/_cli/cli.py
```diff
@@ -339,6 +339,14 @@ def _get_stdin_contents() -> str | None:
 requirements. These and other issues are described in
 https://github.com/marimo-team/marimo/issues/5219.""",
 )
+@click.option(
+    "--trusted/--no-trusted",
```
dmadisetti: bugs me a little. I guess `--not-trusted` breaks our pattern

dmadisetti (approved):

PR #7569, opened 2025-12-19
refactor: cleanup markdown serializers

mscolnick:
This refactors some of the serializers and converts to:
1. pass down the IR instead of full Apps when possible
2. dont pass down duplicate filename (since it may already exist on the notebook)
3. simplify APIs and tests

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -1031,6 +1031,7 @@ class InternalApp:
             app=AppInstantiation(
                 options=self._app._config.asdict(),
             ),
+            filename=self._app._filename,
         )
 
     def to_py(self) -> str:
--- a/marimo/_server/export/exporter.py
+++ b/marimo/_server/export/exporter.py
@@ -272,6 +272,7 @@ class Exporter:
     def export_as_ipynb(
         self,
         app: InternalApp,
+        *,
         sort_mode: Literal["top-down", "topological"],
         session_view: Optional[SessionView] = None,
     ) -> str:
--- a/marimo/_server/notebook/file_manager.py
+++ b/marimo/_server/notebook/file_manager.py
@@ -174,7 +174,11 @@ class AppFileManager:
 
         # Get the header in case it was modified by the user (e.g. package installation)
         handler = get_format_handler(path)
-        header = handler.extract_header(previous_path or path)
+        header: Optional[str] = None
+        if previous_path and previous_path.exists():
+            header = handler.extract_header(previous_path)
+        elif path.exists():
+            header = handler.extract_header(path)
 
         if header:
             notebook = NotebookSerializationV1(
@@ -266,6 +270,7 @@ class AppFileManager:
 
         previous_filename = self._filename
         self._filename = new_path
+        self.app._app._filename = str(new_path)
 
         self._save_file(
             new_path,
--- a/tests/_server/export/test_exporter.py
+++ b/tests/_server/export/test_exporter.py
@@ -40,10 +40,7 @@ def test_export_ipynb_empty():
     internal_app = InternalApp(app)
     exporter = Exporter()
 
-    content, filename = exporter.export_as_ipynb(
-        internal_app, None, sort_mode="top-down"
-    )
-    assert filename == "notebook.ipynb"
+    content = exporter.export_as_ipynb(internal_app, sort_mode="top-down")
     snapshot("empty_notebook.ipynb.txt", content)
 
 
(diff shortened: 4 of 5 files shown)


marimo/_ast/load.py
```diff
@@ -132,44 +133,11 @@ def load_notebook_ir(
             app._unparsable_cell(cell.code, **cell.options)
             continue
         app._cell_manager.register_ir_cell(cell, InternalApp(app))
+    if notebook.header and notebook.header.value:
+        app._header = notebook.header.value
     return app
 
 
-def load_notebook(filename: str) -> Optional[NotebookSerialization]:
```
mscolnick, line 138: this was dead code

marimo/_convert/markdown/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+from marimo._version import __version__
+
+if TYPE_CHECKING:
+    from marimo._ast.cell import CellImpl
+    from marimo._ast.visitor import Language
+
+LOGGER = _loggers.marimo_logger()
+
+
+def convert_from_ir_to_markdown(
```
mscolnick, line 25: similar logic, just moved from exporter

tests/_convert/snapshots/unsafe-doc.md.txt
```diff
@@ -55,9 +55,9 @@ mo.md(r"""
 <!-- Improperly nested code block -->
 
 ````python {.marimo unparsable="true"}
-"""
+\"\"\"
 ```python {.marimo}
-print("Hello, World!")
+print(\"Hello, World!\")
```
mscolnick: this oddly changed and i could not figure out why. this gets parsed as an unparsable cell and quotes escaped

dmadisetti: This is pretty key though because it could break the whole notebook. Test was meant as a bit of a stress test. Will see if I can find what's up

tests/_convert/snapshots/unsafe-doc-old.md.txt
dmadisetti, line 1: Woops, think this snuck in because I was doing a compare when I did the forced md change

tests/_convert/test_markdown_conversion.py
```diff
@@ -364,96 +344,6 @@ def __():
     )
```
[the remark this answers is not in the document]
dmadisetti: Looks fine

marimo/_convert/markdown/__init__.py
dmadisetti: Similar or moved? I think we should match logic for now, at least while that test is failing, then revisit

dmadisetti (commented):
Didn't do a line diff on `convert_from_ir_to_markdown`

Could definitely be cleaned up- but maybe in a followup PR so we don't regress

marimo/_convert/markdown/__init__.py
mscolnick: moved, with a change to dedupe notebook name. nothing changed that should affect that quoting. My guess is the bug has always existed but it's now going through a different path

tests/_convert/test_markdown_conversion.py
```diff
@@ -42,25 +39,8 @@ def sanitized_version(output: str) -> str:
...
-            f.write(py)
-            f.seek(0)
-            output = export_as_md(MarimoPath(tempfile_name)).contents
-    finally:
-        os.remove(tempfile_name)
-
-    title = format_filename_title(tempfile_name)
-    output = re.sub(rf"'?{title}'?", "Test Notebook", output)
+def convert_from_py_to_md(py: str) -> str:
+    output = MarimoConvert.from_py(py).to_markdown(filename="Test Notebook")
```
dmadisetti, line 43: The bug could just be `from_py` ? Does the test pass if this function is reverted?

mscolnick: no, because it fails on `main` too

dmadisetti: the replicating test doesn't go through this path though, but implements the change directly

mscolnick: oh yea that is correct. the bug is in `from_py`. but previously we went through App (instead of the IR), which masked the bug

dmadisetti (approved):

tests/_convert/snapshots/unsafe-doc.md.txt
dmadisetti: Solved in other PR

PR #7583, opened 2025-12-22
fix: unparsable cell is not idempotent

mscolnick:
Fixes cases where unparsable cell was not idempotent (writing and reading would produce a different result). This also fixes some cases where there was incorrect indentation on the unparsable cell.


dmadisetti:
Thanks for isolating

PR #7585, opened 2025-12-22
refactor: pull out notification utils

mscolnick:
Refactor notification utils to separate logic and schemas

### DIFF
--- a/tests/_server/test_sessions.py
+++ b/tests/_server/test_sessions.py
@@ -26,8 +26,8 @@ from marimo._config.manager import (
 from marimo._messaging.ops import (
     MessageOperation,
     UpdateCellCodes,
-    deserialize_kernel_message,
 )
+from marimo._messaging.serde import deserialize_kernel_message
 from marimo._messaging.types import KernelMessage
 from marimo._runtime.requests import (
     AppMetadata,


marimo/_messaging/notification_utils.py
```diff
@@ -0,0 +1,271 @@
...
+from marimo._messaging.types import Stream
+from marimo._runtime.context import get_context
+from marimo._runtime.context.types import ContextNotInitializedError
+from marimo._runtime.context.utils import get_mode
+from marimo._types.ids import CellId_t
+
+LOGGER = loggers.marimo_logger()
+
+
+def broadcast_op(op: msgspec.Struct, stream: Optional[Stream] = None) -> None:
```
manzt, line 40: Wonder if we could use our `BaseStruct` here?

marimo/_messaging/console_output_worker.py
```diff
@@ -37,16 +37,20 @@ def _write_console_output(
     data: str,
     mimetype: ConsoleMimeType,
 ) -> None:
+    from marimo._messaging.notification_utils import broadcast_op
```
manzt, line 40: I'm guessing a name change coming for this suffix (maybe in a separate PR?) now that we are using notification terminology? notify?

marimo/_runtime/runtime.py
```diff
@@ -1315,30 +1318,32 @@ def mutate_graph(
...
-                    used_by=list(
-                        self.graph.get_referring_cells(
-                            variable, language="python"
-                        )
-                    ),
-                )
-                for variable, declared_by in self.graph.definitions.items()
-            ]
-        ).broadcast()
+        broadcast_op(
```
manzt, line 1331: Is there any rule of thumb for when we have a designated helper on `CellNotificationUtils` (compared to raw usage of `broadcast_op`)?

marimo/_messaging/console_output_worker.py
mscolnick: yep exactly. will do that in another PR

marimo/_runtime/runtime.py
mscolnick: not really. i just wanted to keep the migration simple and keep things simple. `CellOps` had their own factory class so i kept it that way

marimo/_messaging/notification_utils.py
mscolnick: oh yea good call. i can do that in a followuip

PR #7593, opened 2025-12-23
refactor: rename kernel Requests to Command

mscolnick:
This renames and makes consistent the kernel **Commands**. Previously they were called `X_Request` which was similar but kinda different from the REST Requests. Having this indirection/separate will be better long term.

Some Commands match exactly request to they are aliased like:

```python
class ListSecretKeysRequest(ListSecretKeysCommand, tag=False):
    pass

# this does not work since we want to get rid of the `tag` field
ListSecretKeysRequest = ListSecretKeysCommand
```

This also cleans up some bugs within the pyodide session and improves the typings.

### DIFF
--- a/marimo/_server/api/utils.py
+++ b/marimo/_server/api/utils.py
@@ -7,7 +7,13 @@ import sys
 import webbrowser
 from pathlib import Path
 from shutil import which
-from typing import TYPE_CHECKING, Optional, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Optional,
+    Protocol,
+    TypeVar,
+    runtime_checkable,
+)
 
 from marimo._runtime.commands import CommandMessage
 from marimo._server.models.models import SuccessResponse
@@ -27,6 +33,13 @@ async def parse_request(
     )
 
 
+@runtime_checkable
+class RequestAsCommand(Protocol):
+    """Protocol for requests that can be converted to commands."""
+
+    def as_command(self) -> CommandMessage: ...
+
+
 async def dispatch_control_request(
     request: Request,
     cls: type[CommandMessage] | CommandMessage,
@@ -41,8 +54,8 @@ async def dispatch_control_request(
         body = await parse_request(request, cls)
     else:
         body = cls
-    if hasattr(body, "as_command"):
-        body = body.as_command()  # type: ignore
+    if isinstance(body, RequestAsCommand):
+        body = body.as_command()
     app_state.require_current_session().put_control_request(
         body,
         from_consumer_id=ConsumerId(app_state.require_current_session_id()),
--- a/tests/_runtime/test_commands.py
+++ b/tests/_runtime/test_commands.py
@@ -1,0 +1,10 @@
+# Copyright 2024 Marimo. All rights reserved.
+from __future__ import annotations
+
+from marimo._runtime.commands import kebab_case
+
+
+def test_kebab_case() -> None:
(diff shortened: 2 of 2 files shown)


marimo/_runtime/commands.py
```diff
@@ -31,7 +32,24 @@
...
+    if name.endswith("Command"):
+        name = name[:-7]  # Remove 'Command' (7 characters)
+    if not name:
+        return name
+    # Insert hyphens before uppercase letters that follow lowercase letters
+    # or before an uppercase letter followed by a lowercase letter (handling acronyms)
+    # This groups consecutive uppercase letters together (e.g., "SQL" -> "sql")
+    s1 = re.sub("(.)([A-Z][a-z]+)", r"\1-\2", name)
+    s2 = re.sub("([a-z0-9])([A-Z])", r"\1-\2", s1)
+    return s2.lower()
```
mscolnick, line 45: this has a runtime cost (very small). i could just inline the tags instead

frontend/src/core/wasm/bridge.ts
```diff
@@ -247,13 +248,19 @@ export class PyodideBridge implements RunRequests, EditRequests {
   sendRun: EditRequests["sendRun"] = async (request) => {
     await this.rpc.proxy.request.loadPackages(request.codes.join("\n"));
 
-    await this.putControlRequest(request);
+    await this.putControlRequest({
```
mscolnick, line 251: typings greatly improved in this file

marimo/_runtime/commands.py
```diff
@@ -31,7 +32,24 @@
     from starlette.requests import HTTPConnection
 
 
-CompletionRequestId = str
+def kebab_case(name: str) -> str:
```
Light2Dark, line 35: add tests?

marimo/_server/api/utils.py
```diff
@@ -41,6 +41,8 @@ async def dispatch_control_request(
         body = await parse_request(request, cls)
     else:
         body = cls
+    if hasattr(body, "as_command"):
```
Light2Dark: do you think we should add this in `models.py`
```python
from typing import runtime_checkable

@runtime_checkable
class RequestAsCommand(Protocol):
    """Protocol for requests that can be converted to commands."""

    def as_command(self) -> CommandMessage: ...
```
```python
if isinstance(body, RequestAsCommand):
   ...
```
just to make sure this doesn't get changed easily, a bit more typesafety.

marimo/_server/models/models.py
```diff
@@ -7,31 +7,184 @@
...
+    UpdateUIElementCommand,
+    UpdateUserConfigCommand,
+    UpdateWidgetModelCommand,
+    ValidateSQLCommand,
 )
 from marimo._types.ids import CellId_t, UIElementId
 
 
-class UpdateComponentValuesRequest(msgspec.Struct, rename="camel"):
+class ListSecretKeysRequest(ListSecretKeysCommand, tag=False):
```
Light2Dark, line 36: I tried the idea of inheriting from an abstract class (since they all have the as_command method), but it's more complicated than I expected.

Light2Dark (dismissed):

marimo/_server/api/utils.py
mscolnick: good idea, updating

marimo/_runtime/commands.py
mscolnick: yep, good call

PR #7596, opened 2025-12-23
refactor: pull out variable logic from notification schemas

mscolnick:
Pull out more logic from the notifications schemas.

### DIFF
--- a/tests/_server/session/test_session_view.py
+++ b/tests/_server/session/test_session_view.py
@@ -30,9 +30,9 @@ from marimo._messaging.notification import (
     VariablesNotification,
     VariableValue,
     VariableValuesNotification,
-    serialize_kernel_message,
 )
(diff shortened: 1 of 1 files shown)


marimo/_messaging/notification.py
```diff
@@ -118,27 +112,6 @@ class FunctionCallResultNotification(Notification, tag="function-call-result"):
     return_value: JSONType
     status: HumanReadableStatus
 
-    def serialize(self) -> KernelMessage:
```
mscolnick, line 121: dead code

mscolnick: cc @manzt - i think we removed the class level `serialize` on the base `Op` class (now notification) so this may have been regressed. i havent seen issues of unable to deserialize function requests, but in case there is, its likely missing this `catch`

PR #7603, opened 2025-12-23
improvement: run auto-instaniate on the backend for run mode

mscolnick:
This PR makes the `/instantiate` call happen on the backend for run mode. The backend has all the information to start the session and does not need to have the frontend call it. Doing this allows us to:
1. require `edit` perms to call `/instantiate`, and also:
2. optionally accept `codes: dict[CellId, str]` to instantiate. 

This lets us lazily call instantiate so we can allow edits or add/remove cells, and then lazily start a kernel by using the new local state.

### DIFF
--- a/tests/_server/api/endpoints/test_execution.py
+++ b/tests/_server/api/endpoints/test_execution.py
@@ -259,9 +259,7 @@ class TestExecutionRoutes_RunMode:
                 "autoRun": True,
             },
         )
-        assert response.status_code == 200, response.text
-        assert response.headers["content-type"] == "application/json"
-        assert "success" in response.json()
+        assert response.status_code == 401, response.text
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


marimo/_server/models/models.py
```diff
@@ -186,6 +186,11 @@ def __post_init__(self) -> None:
 
 class InstantiateNotebookRequest(UpdateUIElementValuesRequest):
     auto_run: bool = True
+    # Optional: cell codes to use instead of the codes from the file.
+    # This is used when the frontend has local edits that should be
+    # used instead of the file codes (e.g., pre-connect editing).
```
akshayka, line 191: nice!

akshayka (dismissed):

dmadisetti (dismissed):

PR #7608, opened 2025-12-26
Also allow `Ctrl` as modifier key on macOS

manzt:
On macOS, users coming from Jupyter and Colab may expect `Ctrl-Enter` to run cells, since those tools use `Ctrl` as the modifier key regardless of platform. Previously, only `Cmd-Enter` worked in marimo on macOS.

There are two separate systems handling keyboard shortcuts in marimo:

1. Event handlers that use `parseShortcut()` to match keyboard events against shortcut strings. These check event properties like `metaKey` and `ctrlKey` directly.

2. ~CodeMirror's keymap system, which has its own string-based key parsing. When we pass `Cmd-Enter` to CodeMirror, it only matches the Command key.~

The previous implementation added `withCtrlEquivalents()` which duplicated all Cmd-based keybindings to also accept Ctrl on macOS. This was too broad and risked conflicts with CodeMirror's built-in Ctrl shortcuts (Ctrl-D for selection, Ctrl-K for line operations, etc.).

The fix narrows the scope: only run-related cell actions (`cell.run` and `cell.runAndNewAbove`) get `Ctrl` equivalents. These are the shortcuts that Jupyter and Colab users expect to work with Ctrl-Enter. Other cell keymaps like format, hide code, and focus navigation remain `Cmd`-only.

### DIFF
--- a/frontend/src/core/hotkeys/shortcuts.ts
+++ b/frontend/src/core/hotkeys/shortcuts.ts
@@ -1,5 +1,6 @@
 /* Copyright 2026 Marimo. All rights reserved. */
 
+import type { KeyBinding } from "@codemirror/view";
 import { Logger } from "@/utils/Logger";
 import { NOT_SET } from "./hotkeys";
 
@@ -162,7 +163,7 @@ export function resolvePlatform(): Platform {
  * Note: If the binding already contains Ctrl (e.g., Cmd-Ctrl-Enter),
  * no duplication is done to avoid producing invalid Ctrl-Ctrl-key combos.
  */
-export function duplicateWithCtrlModifier<T extends { key?: string }>(
+export function duplicateWithCtrlModifier<T extends KeyBinding>(
   binding: T,
 ): T[] {
   // Skip if not macOS, not a Cmd binding, or already has Ctrl


mscolnick:
> On macOS, users coming from Jupyter and Colab may expect Ctrl-Enter to run cells, since those tools use Ctrl as the modifier key regardless of platform. Previously, only Cmd-Enter worked in marimo on macOS.

Does it make sense to just map this keyboard shortcut specifically? Instead of duplicating all `Cmd` to `Ctrl`? This feels like there will be other undesired behavior as a result.

This is a good starting point to see if there are potential conflicts: https://github.com/search?q=org%3Acodemirror+language%3ATypeScript+Ctrl&type=code

manzt:
> Does it make sense to just map this keyboard shortcut specifically? 

I believe there are more mappings beyond running cells (e.g., actions like split cell), but I see the conflict now with others beyond that (builtin to codemirror). Maybe the right idea would be to expand all our [cell-specific keymaps](https://github.com/marimo-team/marimo/blob/main/frontend/src/core/codemirror/cells/extensions.ts#L27-L27)? cc: @akshayka

akshayka:
I like the idea of scoping to cell-specific if that does not introduce conflicts

Light2Dark (approved):

manzt:
This should be ready to go i think.

PR #7617, opened 2025-12-28
feat(debugger): add command history navigation with arrow keys

### LINKED ISSUE #7604: Support pdb up and down arrows for command history navigation
### Description

In the marimo pdb, the up and down arrow keys don’t seem to navigate through previously entered commands. In standard `pdb`, up and down let you cycle through command history, which is a helpful quality-of-life feature during debugging.

I might be missing something but I couldn't find anything on it, and without it, repeating or slightly modifying past commands requires retyping them manually.

### Suggested solution

Add command history navigation to the marimo debugger using the up and down arrow keys, similar to standard `pdb`.

**Expected behavior:**

* `up-arrow` shows previous commands
* `down-arrow` moves forward in the history

This would make iterative debugging faster and bring marimo’s debugger experience closer to standard Python tools.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

majiayu000:
## Summary
Add up/down arrow key support in pdb debugger input for navigating through previously entered commands.

## Changes
- ArrowUp navigates to previous commands
- ArrowDown navigates forward in history
- Preserves current input when starting to navigate
- Avoids adding duplicate consecutive commands

## Test plan
- [x] Added unit tests for debugger controls

Closes #7604

### DIFF
--- a/frontend/src/components/editor/output/console/ConsoleOutput.tsx
+++ b/frontend/src/components/editor/output/console/ConsoleOutput.tsx
@@ -11,6 +11,7 @@ import type { CellId } from "@/core/cells/ids";
 import { isInternalCellName } from "@/core/cells/names";
 import type { WithResponse } from "@/core/cells/types";
 import type { OutputMessage } from "@/core/kernel/messages";
+import { useInputHistory } from "@/hooks/useInputHistory";
 import { useSelectAllContent } from "@/hooks/useSelectAllContent";
 import { cn } from "@/utils/cn";
 import { invariant } from "@/utils/invariant";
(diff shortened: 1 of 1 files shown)


mscolnick (commented):
This looks great, would it be possible to extra the logic into an isolated hook like `useInputHistory`. and this hook can return the `history` and callbacks

majiayu000:

majiayu000:
Done! Extracted the logic into `useInputHistory` hook with tests.

mscolnick (dismissed):
nice, thank you!

frontend/src/components/debugger/debugger-code.tsx
```diff
@@ -19,7 +19,7 @@ import { Button } from "../ui/button";
 import { Tooltip } from "../ui/tooltip";
 import "./debugger-code.css";
 import { useKeydownOnElement } from "@/hooks/useHotkey";
-import { Functions } from "@/utils/functions";
+import { useInputHistory } from "@/hooks/useInputHistory";
```
mscolnick, line 22: this file is actually not used (its fine to keep this code here though when we use it again).

instead, you will want to modify `<StdInput/>`. you can at least use the same hook that you created

majiayu000:
Thanks for the feedback\! I've applied the `useInputHistory` hook to the `StdInput` component in `ConsoleOutput.tsx` which is the actual stdin input used for pdb debugging.

PR #7623, opened 2025-12-29
feat: quarto specific code block

### LINKED ISSUE #5394: Editing Quarto markdown files
Marimo team,

Exceptional project!  I have been exploring this: https://docs.marimo.io/guides/editor_features/watching/#as-markdown, which I find incredibly useful.

I like to use VS code, with the Quarto extension, which gives the ability have VS code execute code in a python kernel, and gives intellisense.  If I have a file called marimo_interoperability.qmd like this:

````
```{python}
import datetime as dt
import marimo as mo
```
 
```{python}
print("Hi Marimo")
```
 
```{python}
print(dt.date.today())
```
 
```{python}
mo.ui.dropdown(options=["a", "b", "c"], value="a", label="choose one")
```
````

I can then also edit this in marimo with command ```marimo edit --watch marimo_interoperability.qmd```.  I also really like the marimo editor, and what is super cool here is that I can also edit the file in the marimo editor and it syncs to the raw .qmd file as well.  However, when I edit and save in the marimo editor, the fencing style of ```{python}``` changes to ```python {.marimo}```, and then the Quarto extension in VS code no longer can provide intellisense, and VS code can't execute the .qmd file.  

Is that expected behavior? Is there an option to toggle to preserve the ```{python}``` fencing?

### PR DESCRIPTION
**Pull request by dmadisetti**
## 📝 Summary

closes #5394

Use Quarto-compatible fence syntax `{marimo .python}` for .qmd files while preserving existing behavior for .md files.

Fence Format Reference

  | Format              | Executable? | claimsLanguage() | Use Case                                 |
  |---------------------|-------------|------------------|------------------------------------------|
  | ```{marimo}         | ✅          | ✅               | Posit recommended                        |
  | ```{marimo-python}  | ✅          | ✅               | Alternative                              |
  | ```{marimo .python} | ✅          | ✅               | This PR - preserves language info        |
  | ```{python.marimo}  | ✅          | ❌               | Requires explicit engine                 |
  | ```python {.marimo} | ✅          | ❌               | Current .md format (GitHub highlighting) |
  | ```{.python.marimo} | ❌          | ❌               | Fallback .md format                      |

I think we'd like `{python.marimo}` as the preferred option, but this PR goes with `{marimo .python}` which is also fine

### CODE REVIEW

**Comment by dmadisetti** [comment]
cc @gordonwoodhull

File: marimo/_server/notebook/serializer.py [hunk 1]
```diff
@@ -62,7 +62,7 @@ class MarkdownNotebookSerializer(NotebookSerializer):
 
     def serialize(self, notebook: NotebookSerializationV1) -> str:
         """Serialize notebook to Markdown format."""
-        return convert_from_ir_to_markdown(notebook)
+        return convert_from_ir_to_markdown(notebook, notebook.filename)
```
mscolnick: this can be reverted right?

marimo/_server/export/__init__.py
```diff
@@ -77,7 +77,7 @@ def export_as_script(path: MarimoPath) -> ExportResult:
 def export_as_md(path: MarimoPath) -> ExportResult:
     ir = _as_ir(path)
```
mscolnick, line 78: should `_as_ir` be putting the filename in the `NotebookSerializationV1` object, instead of passing it down again to `to_markdown`

marimo/_server/notebook/serializer.py
dmadisetti: Yes, the regex is loose enough to get both cases without passing in filename information

marimo/_server/export/__init__.py
dmadisetti: Woops. Yeah fall backs capture this. Good catch- cleaned up.

dmadisetti:
Failing tests look unrelated

PR #7624, opened 2025-12-29
feat: [tools.marimo.env] to specify the desired env

dmadisetti:
## 📝 Summary

Let's user specify:

```toml
[tool.marimo.env]
conda = "conda_name"
# OR
python = "<path>"
```

Which will relaunch with the given environmental variables

With `marimo edit --sandbox` working I think this path will be tweaked, but proof of concept works for now.

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -39,15 +39,8 @@ def should_use_external_env(name: str | None) -> str | None:
     Returns:
         Absolute path to Python interpreter, or None if no external env configured.
     """
-    from marimo._cli.external_env import (
-        is_external_env_active,
-        resolve_python_path,
-    )
+    from marimo._cli.external_env import is_same_python, resolve_python_path
     from marimo._config.manager import get_default_config_manager
-
-    # Prevent recursion - already running with external env
-    if is_external_env_active():
-        return None
 
     # Check per-notebook config first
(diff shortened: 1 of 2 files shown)


marimo/_config/config.py
```diff
@@ -97,6 +97,28 @@ class KeymapConfig(TypedDict):
...
+    """Configuration for external Python environments.
+
+    Allows specifying an existing Python environment to use instead of
+    marimo's sandbox mode. This is useful for JupyterHub, conda environments,
+    or corporate managed Python installations.
+
+    **Keys.**
+
+    - `use_active`: if `True`, use the currently active environment
+      (from VIRTUAL_ENV or CONDA_PREFIX environment variables)
```
akshayka, line 112: Not needed as discussed

marimo/_cli/sandbox.py
```diff
@@ -27,6 +27,65 @@
...
+            env_config = pyproject.env_config
+            if env_config:
+                python_path = resolve_python_path(env_config)
+                if python_path and not is_same_python(python_path):
+                    return python_path
+        except Exception as e:
+            LOGGER.debug(f"Failed to read env config from notebook: {e}")
+
+    # Check project-level config ([tool.marimo.env] in pyproject.toml)
+    try:
```
dmadisetti, line 58: Maybe doesn't make sense

PR #7628, opened 2025-12-29
refactor: cleanup app-router

mscolnick:
Reduce the size of the `AppRouter` class, make it more testable, add more tests.

### DIFF
--- a/marimo/_server/files/directory_scanner.py
+++ b/marimo/_server/files/directory_scanner.py
@@ -120,6 +120,8 @@ class DirectoryScanner:
             if max_execution_time is not None
             else self.MAX_EXECUTION_TIME
         )
+        # Stores partial results in case of timeout
+        self.partial_results: list[FileInfo] = []
 
     @property
     def allowed_extensions(self) -> tuple[str, ...]:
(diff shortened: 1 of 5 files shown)


marimo/_server/files/directory_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+def is_marimo_app(full_path: str) -> bool:
+    """
+    Detect whether a file is a marimo app.
+
+    Rules:
+    - Markdown (`.md`/`.qmd`) files are marimo apps if the first 512 bytes
+      contain `marimo-version:`.
+    - Python (`.py`) files are marimo apps if the header (first 512 bytes)
+      contains both `marimo.App` and `import marimo`.
+    - If the header contains `# /// script`, read the full file and check for
```
dmadisetti, line 26: if the 512 trick is a DDOS prevention isn't this just an easy trick around that?
Could just bump this up a fair amount in this case.

marimo/_server/app_defaults.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import TYPE_CHECKING, Optional
+
+from marimo._config.config import ExportType, SqlOutputType, WidthType
+
+if TYPE_CHECKING:
+    from marimo._config.manager import MarimoConfigManager
+
+
+@dataclass
+class AppDefaults:
```
dmadisetti, line 14: Chance to couple with `marimo/_ast/app_config.py`? Agree we shouldn't move around a full AppConfig object though

marimo/_server/files/directory_scanner.py
dmadisetti: Oh, I see for perf. May as well only check an MB or something

mscolnick: this is only a refactor, not changing any logic

marimo/_server/app_defaults.py
mscolnick: this is only for the server though, i don't think "defaults" belong in `_ast`

PR #7634, opened 2025-12-29
fix: set only partial updates for config changes

mscolnick:


mscolnick:
`saveAppConfig` doesn't have nested objects so not sure it is needed

PR #7644, opened 2025-12-30
deps: wider duckdb support, wider types check

mscolnick:
This fixes some utils to support duckdb `>1.4.0`

### DIFF
--- a/marimo/_sql/utils.py
+++ b/marimo/_sql/utils.py
@@ -66,7 +66,7 @@ def try_convert_to_polars(
     query: str,
     connection: ConnectionOrCursor,
     lazy: bool,
-) -> tuple[Optional[pl.DataFrame | pl.LazyFrame], Optional[str]]:
+) -> tuple[Optional[pl.DataFrame | pl.LazyFrame], Optional[Exception]]:
     """Try to convert the query to a polars dataframe.
 
     Returns:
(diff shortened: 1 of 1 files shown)


akshayka (approved):

akshayka:
CLI test failing otherwise looks good

marimo/_tutorials/sql.py
```diff
@@ -196,7 +196,7 @@ def generate_random_numbers(mean, std_dev, num_samples):
         import pandas as pd
 
         df = pd.DataFrame(_data)
-    return
+    return (df,)
```
mscolnick, line 199: @dmadisetti any idea where this drift comes from? maybe sqlglot not installed previously? or maybe sqlglot is not installed anymore?

Light2Dark (approved):

marimo/_tutorials/sql.py
dmadisetti: Maybe it just hasn't been refreshed since @Light2Dark and I touched this? Very strange though

mscolnick: i think it has to do with sqlglot being installed vs not. but im not sure which is correct.

mscolnick: im going to merge to unblock main

PR #7645, opened 2025-12-30
fix: map matplotlib-pyplot -> matplotlib

mscolnick:


marimo/_runtime/packages/module_name_to_pypi_name.py
```diff
@@ -533,6 +533,7 @@ def module_name_to_pypi_name() -> dict[str, str]:
         "markdown": "Markdown",
         "marks": "pytest-marks",
         "markupsafe": "MarkupSafe",
+        "matplotlib-pyplot": "matplotlib",
```
akshayka, line 536: What's this for? `import matplotlib.pyplot as plt` already correctly prompts to install `matplotlib`. 

Image

mscolnick: i think it was failing in webassembly, but could be another code path. ill take a deeper look

PR #7648, opened 2025-12-31
fix: add cell prefix to setup cell registration

dmadisetti:
## 📝 Summary

When an app with setup cell embeds an app with a setup cell- name collision breaks things. This is because embedded setup cells did not use a cell-id prefix causing a "cell-op" to be send to the frontend that was never satisfied- blocking execution.

### DIFF
--- a/marimo/_ast/cell_manager.py
+++ b/marimo/_ast/cell_manager.py
@@ -165,7 +165,7 @@ class CellManager:
     ) -> Cell:
         """Registers cells when called through a context block."""
         cell = context_cell_factory(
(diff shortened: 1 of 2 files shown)


dmadisetti:
Tests are failing, will re-request on fix

mscolnick:
@dmadisetti one if failing, i think from https://github.com/marimo-team/marimo/pull/7325, but i raised a question in https://github.com/marimo-team/marimo/pull/7644#discussion_r2654336861. i think the fail might be correct?

akshayka (approved):

akshayka:
Failing tests look very unrelated

PR #7657, opened 2026-01-03
Default to running sandboxed notebooks in isolated venvs when the sandbox flag is unspecified

akshayka:
This PR updates the CLI to automatically run notebooks in package sandboxes when the sandbox flag is left unspecified and the notebook at hand has script metadata, removing the interactive prompt.

A hint message is printed communicating to the user that they can opt-out with `--no-sandbox`.

Here is a sample of what is printed when the user types `marimo edit my_sandboxed_notebook.py`:

image


akshayka:
After discussion with the team, closing this in favor of requiring users to be explicit, since users may not want packages installed on their machine without their consent

PR #7664, opened 2026-01-05
Add ADBC DB-API engine for SQL connections

### LINKED ISSUE #7549: Arrow Database Connectivity (ABDC) driver-based sql connections
### Description

I want to use Marimo's data source connections with engines that aren't currently supported.  In particular, I want to use it with analytics-focused query engines, ideally maintaining good performance even when retrieving large result sets.

### Suggested solution

Supporting ADBC driver-based connections, so any db/query engine with an adbc driver could be used as a SQL connection in Marimo.  

From the [ADBC docs](https://arrow.apache.org/adbc/current/):
>ADBC is a set of APIs and libraries for Arrow-native access to databases. Execute SQL and [Substrait](https://substrait.io/) queries, query database catalogs, and more, all using Arrow data to eliminate unnecessary data copies, speed up access, and make it more convenient to build analytical applications.

Aside from expanding supported engines in Marimo, ADBC-based connections could also potentially improve performance if used as an alternative to some existing sqlalchemy-based connections.

- [x] Yes

### Alternatives

_No response_

### Additional context

I haven't yet looked deeply at how SQL connections are implemented in Marimo today.  If there is aligned interest in this feature, I would appreciate any guidance/potential footguns to consider from any maintainers familiar with the implementation.

corleyma:
Closes #7549.

### What
- Adds `AdbcDBAPIEngine` (`marimo/_sql/engines/adbc.py`) to support Arrow ADBC **DB-API wrapper** connections as first-class SQL engines.
- Uses `cursor.fetch_arrow_table()` for Arrow-native result fetching and converts via existing `convert_to_output`.
- Implements catalog discovery via `adbc_get_objects()` / `adbc_get_table_schema()` and respects `include_schemas` / `include_tables` / `include_table_details` via `depth` + filters.
- Adds dialect inference from `adbc_get_info()` (cached), attempting to derive dialect from vendor name.

### Tests
- `hatch run lint`
- `hatch run format --check`
- `hatch run typecheck:check`
- `hatch run typos -w`
- `hatch run +py=3.12 test:test tests/_sql/test_adbc.py`
- `hatch run +py=3.12 test-optional:test tests/_sql/test_adbc.py`

### Notes
- Integration tests for the SQLite ADBC driver are gated behind `pytest.importorskip` and run in `test-optional` (adds `adbc_driver_manager` + `adbc_driver_sqlite` there).

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1133,7 +1133,7 @@ packages:
 - pypi: ./
   name: marimo
   version: 0.18.4
-  sha256: 64534d3c58bd4c81fc84e6f6b7afb039e1d1eee5b8e3ae047b9421d3be87363a
+  sha256: 00d04a9c700e085afe77f6f44d2455742a7247499a2022ef89fe8f9c04858e5d
   requires_dist:
   - click>=8.0,<9
   - jedi>=0.18.0
@@ -1166,7 +1166,6 @@ packages:
   - mcp>=1.0.0 ; extra == 'mcp'
   - pydantic>2 ; extra == 'mcp'
   requires_python: '>=3.10'
-  editable: true
 - pypi: https://files.pythonhosted.org/packages/70/81/54e3ce63502cd085a0c556652a4e1b919c45a446bd1e5300e10c44c8c521/markdown-3.10-py3-none-any.whl
   name: markdown
   version: '3.10'


corleyma:

corleyma:
recheck

marimo/_sql/sql.py
```diff
@@ -40,14 +40,17 @@ def sql(
...
+    DB-API wrappers provided by ADBC drivers.
 
     The result of the query is displayed in the UI if output is True.
 
     Args:
         query: The SQL query to execute.
         output: Whether to display the result in the UI. Defaults to True.
-        engine: Optional SQL engine to use. Can be a SQLAlchemy, DuckDB, Clickhouse, Redshift, Ibis, or DBAPI 2.0 compatible engine.
+        engine: Optional SQL engine to use. Can be a SQLAlchemy, DuckDB, Clickhouse,
+            Redshift, Ibis, or DB-API 2.0 compatible connection (including ADBC DB-API wrappers).
```
Light2Dark:
```suggestion
            Redshift, Ibis, or DB-API 2.0 compatible connection (including ADBC drivers).
```

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,621 @@
...
+    return "string"
+
+
+class AdbcConnectionCatalog:
+    """Catalog implementation backed by ADBC DB-API wrapper extensions."""
+
+    def __init__(
+        self,
+        *,
+        adbc_connection: Any,
```
Light2Dark: can the adbc_connection be of type `AdbcDbApiConnection`? Should give you the typesafety for subsequent use

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,621 @@
...
+        objects_pylist = cast(
+            list[dict[str, Any]],
+            self._adbc_connection.adbc_get_objects(
+                depth="tables",
+                catalog_filter=database or None,
+                db_schema_filter=schema or None,
+            )
+            .read_all()
+            .to_pylist(),
+        )
```
Light2Dark: can we check if it's in the intended format (list[dict]) instead of casting right away?

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,621 @@
...
+                    "execute",
+                    "fetch_arrow_table",
+                    "close",
+                )
+                return all(
+                    callable(getattr(cursor, method, None))
+                    for method in required_cursor_methods
+                )
+            finally:
+                cursor.close()
```
Light2Dark: not obvious to me, but does this return False if try errors?

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,621 @@
...
+
+    def fetch_arrow_table(self) -> Any: ...
+
+    def close(self) -> None: ...
+
+
+class AdbcDbApiConnection(Protocol):
+    """ADBC DB-API wrapper connection.
+
+    The ADBC Python cookbook documents these methods and their return types.
```
Light2Dark: could you add a link to the cookbook here?

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,621 @@
...
+    def adbc_get_info(self) -> dict[str, Any]: ...
+
+
+def _resolve_table_type(table_type: str) -> Literal["table", "view"]:
+    if "view" in table_type.lower():
+        return "view"
+    return "table"
+
+
+_ADBC_DIALECT_SUBSTRING_RULES: tuple[tuple[str, str], ...] = (
```
Light2Dark: I'd make this a dict, so we can do fast lookup instead of iterating through each tuple

tests/_sql/test_adbc.py
```diff
@@ -0,0 +1,729 @@
...
+        self.adbc_get_info_calls += 1
+        return {"vendor_name": "PostgreSQL"}
+
+
+def test_adbc_info_to_dialect_maps_known_vendor() -> None:
+    assert (
+        _adbc_info_to_dialect(info={"vendor_name": "PostgreSQL"})
+        == "postgresql"
+    )
+    assert (
```
Light2Dark: I don't think we need to test all the vendors, it's just a lookup. Just a few would do.

tests/_sql/test_adbc.py
```diff
@@ -0,0 +1,729 @@
...
+    engine = AdbcDBAPIEngine(conn, engine_name=VariableName("adbc_conn"))
+    databases = engine.get_databases(
+        include_schemas=True,
+        include_tables=True,
+        include_table_details=False,
+    )
+    assert conn.last_get_objects_kwargs is not None
+    assert conn.last_get_objects_kwargs["depth"] == "tables"
+
+    assert len(databases) == 1
```
Light2Dark: could you change the individual assert tests to a complete object test
assert databases == [...]

marimo/_sql/get_engines.py
```diff
@@ -39,6 +40,7 @@
     ClickhouseServer,
     PyIcebergEngine,
     RedshiftEngine,
+    AdbcDBAPIEngine,
```
Light2Dark, line 43: I think you can add a test in `test_get_engines.py`

marimo/_sql/engines/adbc.py
corleyma: This constant is an ordered list of substring-match rules, not a hash lookup. Ordering is correctness-critical for overlapping needles (e.g., "oracledb" contains "oracle", "db2i" contains "db2"), so we keep an explicitly ordered structure to encode precedence deterministically.

Admittedly I didn't read the frontend codebase that carefully, but based on my browsing when writing this:
-  db2 vs db2i: the frontend treats them as distinct dialects for SQL formatting (sql-formatter uses db2 vs db2i).
-  oracle vs oracledb: the frontend recognizes both and collapses them to the same formatter/parser behavior (both map to plsql), but the UI display differs (oracle gets a friendly “Oracle” label; oracledb falls back to the raw string).

marimo/_sql/engines/adbc.py
corleyma: There is a definite signature for this method: https://arrow.apache.org/adbc/current/python/api/adbc_driver_manager.html#adbc_driver_manager.dbapi.Connection.adbc_get_objects 

if the object is truly an adbc connection, the type should be guaranteed.  So, is the intent here to guard against things that looked like an ADBC connection but really aren't?  If so, what's the right thing to do here if there is a type mismatch?

marimo/_sql/engines/adbc.py
Light2Dark: yeah, ideally I'd just pass the raw dialect to the frontend, without normalizing it. Since the frontend normalizes it. Is that possible? And you can add the extra cases in the frontend (excluding unecessary conversions like duckdb -> duckdb)

marimo/_sql/engines/adbc.py
Light2Dark: Can the connection be typed, and this method be typed.

marimo/_sql/engines/adbc.py
corleyma: Well, the metadata we get from adbc info is the _vendor_, which is ~approximately a dialect, but these values are kind of unbounded and are up to whoever implements the driver.  That said, I can introspect the existing abdc drivers (most of which are [here](https://github.com/apache/arrow-adbc) in the foundry) and enumerate the current values for vendor in the wild and update the frontend to deal with those...

personally, I don't love leaking these details across to the frontend but it is kind of consistent with how it's been done in the repo so far, so I'll match that.

Note also: some drivers are more generic, e.g.,  adbc-flight-sql, which can be used with any db exposing a Flight SQL interface, sometimes very different vendors with different dialects.  The abdc_info will have the vendor but I can't introspect the values just looking at the driver implementation, so that's something that will just have to be built out over time.

marimo/_sql/engines/adbc.py
corleyma: Try there isn't really going to error (everything there is safe getattr) but the finally is there to be sure we always close.  Still, I cleaned up this logic a bit.

frontend/src/core/codemirror/language/languages/sql/utils.ts
```diff
@@ -49,6 +49,9 @@ const KNOWN_DIALECTS_ARRAY = [
   "oracledb",
   "singlestoredb",
   "timescaledb",
+  "databricks",
```
corleyma, line 52: all of these values were derived by introspection of existing adbc driver implementations in the driver foundry for what is returned as vendor by adbc_get_info; this was the net that was missing in frontend known dialect set.

pyproject.toml
```diff
@@ -117,6 +117,8 @@ dev = [
     # For SQL
     "duckdb>=1.0.0",
     "sqlglot[rs]>=26.2.0",
+    # Stubs for optional dataframe backends
+    "pyarrow-stubs",
```
corleyma, line 121: hopefully this is non-controversial? I added this dep here and in the typecheck; here for better contributor devex, there for the benefits during typechecking.

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,553 @@
...
+    ) -> pa.RecordBatchReader: ...
+
+    def adbc_get_table_schema(
+        self, table_name: str, *, db_schema_filter: str | None = None
+    ) -> pa.Schema: ...
+
+    def adbc_get_info(self) -> dict[str | int, Any]: ...
+
+
+def _resolve_table_type(table_type: str) -> Literal["table", "view"]:
```
Light2Dark:
```suggestion
def _resolve_table_type(table_type: str) -> DataTableType:
```
there's a type somewhere that you can import

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,553 @@
...
+    def get_table_details(
+        self, *, table_name: str, schema_name: str, database_name: str
+    ) -> Optional[DataTable]:
+        _ = database_name
+        try:
+            adbc_get_table_schema = getattr(
+                self._adbc_connection, "adbc_get_table_schema", None
+            )
+            if not callable(adbc_get_table_schema):
+                return None
```
Light2Dark: can we treat this the same as adbc_get_objects, just call `self._adbc_connection.adbc_get_table_schema`

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,553 @@
...
+            engine_name=self._engine_name,
+        )
+
+    def get_default_database(self) -> Optional[str]:
+        catalog = self._catalog()
+        return None if catalog is None else catalog.get_default_database()
+
+    def get_default_schema(self) -> Optional[str]:
+        catalog = self._catalog()
+        return None if catalog is None else catalog.get_default_schema()
```
Light2Dark: should we have an `__init__` function to set this catalog once and use it throughout

marimo/_sql/engines/adbc.py
```diff
@@ -0,0 +1,553 @@
...
+            auto_discover_columns=False,
+        )
+
+    def _catalog(self) -> Optional[AdbcConnectionCatalog]:
+        if not callable(getattr(self._connection, "adbc_get_objects", None)):
+            return None
+        if not callable(
+            getattr(self._connection, "adbc_get_table_schema", None)
+        ):
+            return None
```
Light2Dark: It looks like this method must exist for this to be created (`is_compatible` check). Can we call it directly?

Light2Dark (changes requested):

frontend/src/core/codemirror/language/languages/sql/utils.ts
Light2Dark: nice

Light2Dark:
requested copilot to review, it may be noisy so feel free to ignore

corleyma:
> requested copilot to review, it may be noisy so feel free to ignore

I responded to your latest feedback and did a pass over the copilot feedback; mostly wrong but there was one improvement to import structure in tests that I implemented.

Light2Dark (approved):
awesome, tq!

mscolnick:
@corleyma it looks like there is a test that is failing maybe from the new dep imports.
```
FAILED tests/_plugins/ui/_impl/test_table.py::test_download_as[df0] - ValueError: Unsupported format: csv
```

it may not be related, but this isnt one of our usual flakey test. the test might have been poorly written, so feel free to adapt or reduce the scope of the test if needed.

corleyma:
@mscolnick Since the test is parameterized to test download functionality of any eager DF format that has the relevant dependency installed, I just extended the helper in the test to also handle rountripping pyarrow Table dfs now pyarrow is installed in the test env.

mscolnick:
thank you for the contribution!

PR #7669, opened 2026-01-05
Fix sidebar drag-and-drop requiring click before drag

manzt:
The sidebar panel reordering only worked if you first clicked an icon, _then_ attempted to drag. Coming from elsewhere in the UI and immediately trying to click+drag would fail silently.

The root cause was a nested `<button>` inside react-aria's `ListBoxItem` which captures pointer events before drag can initiate. Click handling now goes through `ListBox`'s `onAction` callback, and `SidebarItem` renders as a `<div>` when used inside `ReorderableList`.

### DIFF
--- a/frontend/src/components/ui/reorderable-list.tsx
+++ b/frontend/src/components/ui/reorderable-list.tsx
@@ -3,6 +3,7 @@
 import type React from "react";
 import { useMemo } from "react";
 import { ListBox, ListBoxItem, useDragAndDrop } from "react-aria-components";
+import { Logger } from "@/utils/Logger";
 import {
   ContextMenu,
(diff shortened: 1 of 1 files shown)


frontend/src/components/ui/reorderable-list.tsx
```diff
@@ -112,13 +118,23 @@ export const ReorderableList = <T extends { id: string | number }>({
     }
   };
 
+  const handleAction = (key: React.Key) => {
+    if (onAction) {
+      const item = value.find((i) => i.id === key);
+      if (item) {
+        onAction(item);
+      }
+    }
```
mscolnick, line 125: could put a warning here, im not sure if we never expect None

PR #7674, opened 2026-01-05
feature: generate additional comments during jupyer2marimo notebook converter (for modified or redefined variables)

bmerkle:
- enhance variable renaming with comments during notebook conversion
- comment for modified or redefined variable is attached for better understanding

## 📝 Summary

Enhanced variable renaming with contextual comments during Jupyter → marimo conversion.
When variables are renamed, modified, or redefined as part of the conversion process, explanatory comments are automatically attached. This improves traceability, preserves intent, and makes the resulting marimo notebook easier to understand and review—especially for non-trivial refactorings introduced during conversion.

## 🔍 Description of Changes

- for modified or redefined variable an additional python comment is generated

## 📋 Checklist

### DIFF
--- a/tests/_convert/test_ipynb.py
+++ b/tests/_convert/test_ipynb.py
@@ -899,14 +899,14 @@ def test_transform_duplicate_definition_kwarg():
         ]
     )
     result = transform_duplicate_definitions(sources)
-    mod_x = _rename_comment("x", RenameType.MODIFIED)
+    redef_x = _rename_comment("x", RenameType.REDEFINED)
     assert_sources_equal(
         result,
         [
             "x = 0",
             "x",
             f"""
-            x_1 = 1{mod_x}
+            x_1 = 1{redef_x}
             def f(x=x_1):
                 return x
             """,
--- a/tests/_convert/test_ipynb.py
+++ b/tests/_convert/test_ipynb.py
@@ -8,6 +8,7 @@ import pytest
 from marimo._convert.ipynb import (
     CellsTransform,
     CodeCell,
+    RenameType,
     Transform,
     convert_from_ipynb_to_notebook_ir,
     transform_add_marimo_import,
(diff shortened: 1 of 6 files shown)


bmerkle:

tests/_convert/test_ipynb.py
```diff
@@ -878,13 +899,14 @@ def f(x=x):
...
+    mod_x = _rename_comment("x", RenameType.MODIFIED)
     assert_sources_equal(
         result,
         [
             "x = 0",
             "x",
-            """
-            x_1 = 1
+            f"""
+            x_1 = 1{mod_x}
```
dmadisetti: Shouldn't this be renamed?

dmadisetti (commented):
Thanks for this.
Noticed a slight discrepancy in the tests.

If this is for reconstruction, then we need to account for in place operations

```python
# ie
x += 1
# vs
x = x + 1
```

tests/_convert/test_ipynb.py
bmerkle: yes you are right, it is also a redefinition case. Thanks for spotting it.

bmerkle:
@dmadisetti is the PR ok from your side ?

dmadisetti:
Thanks for checking in @bmerkle

I think this might add a lot of noise without a ton of value to the user. The benefit I see from doing something like this is that we could potentially exactly reconstruct the ipynb output in addition to letting the user know there's been a change.

But! This is implementation is lacking since it still doesn't distinguish in-place operations.
Maybe we show the converted code exactly and don't need to make the redefined distinction? e.g.

```python
# originally: x += 1
x_1 = x + 1  # No double stacking comments!
```

Thoughts? @akshayka may also have ideas- but hold off on implementation and we can discuss. Thanks for taking the initiative though!

bmerkle:
@dmadisetti thanks for the feedback. I will hold off implementation until we have agreed on the best approach. @akshayka 
I like the idea with the original code as comment above.

The idea was to make it more transparent to the user, where the code came from during the conversion.
We might offer this feature also only as opt-in or opt-out, so somebody who already knows the conversion rules is not cluttered with additional comments. But for debugging or new users i think is is absolutely a good thing to have traces...

we can also chat at discord if that is more efficient

bmerkle:
@dmadisetti @akshayka how should i procced here ? I am happy to modify the PR towards your suggestion 😃

mscolnick:
@bmerkle im going to move this to draft for now, but feel free to close yourself. if you'd like, you can create a GH issue to track the request so we can discuss and track Upvotes

PR #7678, opened 2026-01-05
fix: address windows flakiness by reducing introspection calls

dmadisetti:
## 📝 Summary

Pytest flakiness failures are related to timeout errors in windows. This reduces the number of frame calls needed (and hopefully) fixes the windows issues observed.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1133,7 +1133,7 @@ packages:
 - pypi: ./
   name: marimo
   version: 0.18.4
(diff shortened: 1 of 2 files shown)


mscolnick:
@dmadisetti those ci failures look related to the change in pixi maybe?

dmadisetti:
Ah dang. I though pixi lock lagged. I'm getting `✔ Lock-file was already up-to-date` on main, so maybe a version mismatch

mscolnick:
it does not lag as far as i am aware, and yes seems like version mismatch

PR #7682, opened 2026-01-06
Lighten shadows

akshayka:
This PR lightens shadows throughout the notebook to reduce distraction:

* Cells in a default status do not have shadows
* All shadow changes on cell hover are removed

Additionally:

* Cell movement on hover is removed, eliminating the "pop-out" effect
* A stale (yellow) border is added to cells that need to be run, to make their status more visible
* The "Add Cell" component at the bottom of the Cell Array is changed to use solid shadows, for consistency
* The shadow on the footer bar is removed

### DIFF
--- a/frontend/src/css/app/Cell.css
+++ b/frontend/src/css/app/Cell.css
@@ -159,6 +159,10 @@
     border: 1px solid var(--red-4);
   }
 
+  &.has-error:not(.needs-run):focus-within {
+    @apply shadow-md-solid shadow-error/10;
+  }
+
   &.error-outline,
   &.error-outline:focus-within {
     box-shadow: 8px 8px 0 0 color-mix(in srgb, var(--error), transparent 80%);
@@ -170,7 +174,7 @@
   &.needs-run {
     /* TODO(akshayka): Can give this an outline to make more visible. */
     border-color: var(--stale);
-    @apply divide-stale shadow-md-solid shadow-stale;
+    @apply divide-stale shadow-sm-solid shadow-stale;
 
     &:focus-within {
       @apply divide-stale shadow-md-solid shadow-stale;


frontend/src/css/app/Cell.css
```diff
@@ -205,20 +169,14 @@
 
   &.needs-run {
     /* TODO(akshayka): Can give this an outline to make more visible. */
-    outline: none;
+    border-color: var(--stale);
     @apply divide-stale shadow-md-solid shadow-stale;
```
Light2Dark: i would change this to `shadow-sm-solid`, so there's a more visible difference when it's focused

frontend/src/css/globals.css
```diff
@@ -182,4 +182,10 @@
 
   /* biome-ignore format: definition needs to be oneline or breaks variants */
   --shadow-2xl-solid: 10px 12px 0px 0px var(--base-shadow-darker), 0 0px 8px 0px hsl(0deg 0% 90% / 50%);
+
+  /* Solid shadows with lighter shade color */
+  /* biome-ignore format: definition needs to be oneline or breaks variants */
+  --shadow-sm-solid-shade: 2px 2px 0px 0px var(--base-shadow), 0px 0px 2px 0px hsl(0deg 0% 50% / 20%);
+  /* biome-ignore format: definition needs to be oneline or breaks variants */
+  --shadow-md-solid-shade: 4px 4px 0px 0px var(--base-shadow), 0 0px 2px 0px hsl(0deg 0% 60% / 50%);
```
mscolnick, line 190: it would be nice to reduce the variant of shadows but if thats not possible, thats ok

akshayka: I think we could probably get rid of some of the other ones but I didn't do a pass. This is what we use for the Cell shadows

frontend/src/css/app/Cell.css
akshayka: done, thanks!

PR #7697, opened 2026-01-06
Add cross-list drag-drop for sidebar and developer panel

manzt:
Previously, sidebar panels and developer panel tabs were separate systems with distinct types and no way to move items between them. 

These changes unify both into a single `PanelType` with a `defaultSection` property, stored in a new `panelLayoutAtom` that tracks which panels belong to each section. Panels can now be dragged between the sidebar and developer panel, with context menus dynamically showing only panels not already in the other section.

The `ReorderableList` component gains cross-list drag support via a `crossListDrag` prop that groups lists by `dragType`.

### DIFF
--- a/frontend/src/components/editor/chrome/types.ts
+++ b/frontend/src/components/editor/chrome/types.ts
@@ -54,8 +54,6 @@ export interface PanelDescriptor {
   tooltip: string;
   /** If true, the panel is completely unavailable */
   hidden?: boolean;
-  /** If true, the panel is available but not shown by default */
-  defaultHidden?: boolean;
   /** Which section this panel belongs to by default */
   defaultSection: PanelSection;
 }
@@ -99,8 +97,7 @@ export const PANELS: PanelDescriptor[] = [
     Icon: SquareDashedBottomCodeIcon,
     label: "Snippets",
     tooltip: "Snippets",
-    defaultSection: "sidebar",
-    defaultHidden: true,
+    defaultSection: "developer-panel",
   },
   {
     type: "outline",
--- a/frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
+++ b/frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
@@ -22,7 +22,6 @@ import { ReorderableList } from "@/components/ui/reorderable-list";
 import { Tabs, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { LazyMount } from "@/components/utils/lazy-mount";
 import { cellErrorCount } from "@/core/cells/cells";
-import { snippetsEnabledAtom } from "@/core/config/config";
 import { getFeatureFlag } from "@/core/config/feature-flag";
 import { cn } from "@/utils/cn";
 import { ErrorBoundary } from "../../boundary/ErrorBoundary";
@@ -81,7 +80,6 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
   const { aiPanelTab, setAiPanelTab } = useAiPanelTab();
   const errorCount = useAtomValue(cellErrorCount);
   const [panelLayout, setPanelLayout] = useAtom(panelLayoutAtom);
-  const snippetsEnabled = useAtomValue(snippetsEnabledAtom);
 
   // Convert current developer panel items to PanelDescriptors
   const devPanelItems = useMemo(() => {
@@ -133,13 +131,9 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
       if (sidebarIds.has(p.type)) {
         return false;
       }
-      // Show defaultHidden panels only if enabled via config
-      if (p.defaultHidden && p.type === "snippets" && !snippetsEnabled) {
-        return false;
-      }
       return true;
     });
-  }, [panelLayout.sidebar, snippetsEnabled]);
+  }, [panelLayout.sidebar]);
 
   // sync sidebar
   useEffect(() => {
--- a/frontend/src/components/editor/chrome/state.ts
+++ b/frontend/src/components/editor/chrome/state.ts
@@ -27,11 +27,10 @@ export interface PanelLayout {
 
 const DEFAULT_PANEL_LAYOUT: PanelLayout = {
   sidebar: PANELS.filter(
-    (p) => !p.hidden && !p.defaultHidden && p.defaultSection === "sidebar",
+    (p) => !p.hidden && p.defaultSection === "sidebar",
   ).map((p) => p.type),
   developerPanel: PANELS.filter(
-    (p) =>
-      !p.hidden && !p.defaultHidden && p.defaultSection === "developer-panel",
+    (p) => !p.hidden && p.defaultSection === "developer-panel",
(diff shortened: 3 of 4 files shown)


akshayka (commented):
After dragging the errors panel to the sidebar, I managed to get my sidebar into a state where it looks like two panels are selected even though they aren't

Image

frontend/src/components/editor/chrome/types.ts
```diff
@@ -30,160 +34,131 @@ export type PanelType =
...
     label: "Terminal",
+    defaultSection: "developer-panel",
   },
-  // TODO(akshayka): The cache panel should not be default shown,
-  // even when it's out of feature flag. (User config to turn it on.)
   {
     type: "cache",
     Icon: DatabaseZapIcon,
     label: "Cache",
+    defaultSection: "developer-panel",
```
mscolnick, line 172: `defaultSection` empty so its not included?

frontend/src/components/editor/chrome/types.ts
```diff
@@ -30,160 +34,131 @@ export type PanelType =
...
     type: "errors",
     Icon: XCircleIcon,
     label: "Errors",
+    defaultSection: "developer-panel",
   },
   {
     type: "scratchpad",
     Icon: NotebookPenIcon,
     label: "Scratchpad",
+    defaultSection: "developer-panel",
```
mscolnick, line 136: @akshayka do you still want this default hidden?

frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
```diff
@@ -212,6 +279,18 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
             {selectedPanel === "documentation" && <LazyDocumentationPanel />}
             {selectedPanel === "snippets" && <LazySnippetsPanel />}
             {selectedPanel === "ai" && renderAiPanel()}
+            {selectedPanel === "errors" && <LazyErrorsPanel />}
```
mscolnick, line 276: maybe possible to make this programatic instead?. i guess with `renderAiPanel` and `terminal`, those might need to be excluded.

frontend/src/components/editor/chrome/types.ts
akshayka: I personally like having the scractpad in the developer panel

frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
```diff
@@ -325,6 +415,60 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
         </div>
         {/* Panel content */}
         <div className="flex-1 overflow-hidden">
+          {selectedDeveloperPanelTab === "files" && (
+            <LazyMount isOpen={isDeveloperPanelOpen}>
+              <Suspense fallback={<div />}>
```
mscolnick: we could hoist the `Suspense` out of here. i think `LazyMount` was only needed for the Terminal (to avoid a spurious connection), if i recall. (sorry, I should've documented that.

frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
```diff
@@ -68,11 +75,71 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
...
   const terminalRef = React.useRef<ImperativePanelHandle>(null);
   const { aiPanelTab, setAiPanelTab } = useAiPanelTab();
   const errorCount = useAtomValue(cellErrorCount);
+  const [panelLayout, setPanelLayout] = useAtom(panelLayoutAtom);
+  const snippetsEnabled = useAtomValue(snippetsEnabledAtom);
+
+  // Convert current developer panel items to PanelDescriptors
+  const devPanelItems = useMemo(() => {
+    return panelLayout.developerPanel
+      .map((id) => PANELS.find((p) => p.type === id))
```
mscolnick: we might want to create a `PANELS_BY_ID`, and then can do 

```ts
panelLayout.developerPanel.flatMap(id => PANELS_BY_ID.get(id))

panelLayout.developerPanel.map(id => PANELS_BY_ID.get(id)).filter(Boolean)
```

frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
```diff
@@ -68,11 +75,71 @@ export const AppChrome: React.FC<PropsWithChildren> = ({ children }) => {
...
+        sidebar: prev.sidebar.filter((id) => id !== item.type),
+      }));
+
+      // If the moved item was selected in sidebar, select the first remaining item
+      if (selectedPanel === item.type) {
+        const remainingSidebar = panelLayout.sidebar.filter(
+          (id) => id !== item.type,
+        );
+        if (remainingSidebar.length > 0) {
+          openApplication(remainingSidebar[0]);
```
mscolnick, line 113: is it possible to have no remaining items and you cannot drag it back to the list? maybe if the list has a min-width?

frontend/src/components/editor/chrome/types.ts
```diff
@@ -30,160 +34,131 @@ export type PanelType =
...
   },
   {
-    id: "snippets",
+    type: "snippets",
     Icon: SquareDashedBottomCodeIcon,
-    tooltip: "Snippets",
-    position: "sidebar",
+    label: "Snippets",
+    defaultSection: "sidebar",
     defaultHidden: true,
```
mscolnick: thinking out loud. i wonder if `defaultHidden` should be when `defaultSection` is `null`? or does that make adding new sidebar items harder?

frontend/src/components/editor/chrome/types.ts
mscolnick: it looks like its hidden now (`defaultHidden: true`).

maybe for simplicity in the logic, we drop `defaultHidden` and can add it back or some other logic when there is a "first" default hidden panel

frontend/src/components/editor/chrome/types.ts
manzt: I think we can make it a requirement that `defaultSelection` is always defined?

frontend/src/components/editor/chrome/types.ts
manzt: I think `defaultHidden` is only true for snippets (not scratchpad)

mscolnick: oh sorry, i was looking at the wrong one. ok then this is fine if we do have a some tab that is hidden by default

frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
manzt: Ah, I already have `PANEL_MAP` ... reusing.

frontend/src/components/editor/chrome/wrapper/app-chrome.tsx
manzt: I updated the component to have a special "empty" element which still allows right click and drag and drop if no items are shown.

mscolnick (approved):

PR #7698, opened 2026-01-06
fix: specific handling for lazy requests

mscolnick:
Each request now will either: `"throwError" | "dropRequest" | "startConnection" | "waitForConnectionOpen"`. This is a bit cumbersome to manage but i think getting it right will have a good UX. For example, we can let the user type and make edits, but not start the kernel. Once the user hits "run", then we can actually start the kernel.

### DIFF
--- a/frontend/src/core/network/requests-lazy.ts
+++ b/frontend/src/core/network/requests-lazy.ts
@@ -1,6 +1,6 @@
 /* Copyright 2024 Marimo. All rights reserved. */
 
-import { NoKernelConnected } from "@/utils/errors";
+import { NoKernelConnectedError } from "@/utils/errors";
 import { Objects } from "@/utils/objects";
 import { memoizeLastValue } from "@/utils/once";
 import { waitForKernelToBeInstantiated } from "../kernel/state";
@@ -140,7 +140,7 @@ export function createLazyRequests(
           return Promise.resolve();
 
         case "throwError":
-          throw new NoKernelConnected();
+          throw new NoKernelConnectedError();
 
         case "waitForConnectionOpen":
           // Wait for connection but don't start it
--- a/frontend/src/core/network/requests-toasting.tsx
+++ b/frontend/src/core/network/requests-toasting.tsx
@@ -5,7 +5,7 @@ import { useAtomValue } from "jotai";
 import { Spinner } from "@/components/icons/spinner";
 import { Button } from "@/components/ui/button";
 import { toast } from "@/components/ui/use-toast";
-import { NoKernelConnected, prettyError } from "@/utils/errors";
+import { NoKernelConnectedError, prettyError } from "@/utils/errors";
 import { Logger } from "@/utils/Logger";
 import { useConnectToRuntime } from "../runtime/config";
 import { store } from "../state/jotai";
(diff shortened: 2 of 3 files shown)


frontend/src/components/editor/chrome/wrapper/footer-items/runtime-settings.tsx
```diff
@@ -81,149 +81,136 @@ export const RuntimeSettings: React.FC<RuntimeSettingsProps> = ({
...
-              data-testid="footer-runtime-settings"
-            >
-              <div className="flex items-center gap-1">
-                {allReactivityDisabled ? (
-                  <ZapOffIcon size={16} className="text-muted-foreground" />
-                ) : (
-                  <ZapIcon size={16} className="text-amber-500" />
-                )}
-                <ChevronDownIcon size={14} />
+    <DropdownMenu>
```
mscolnick, line 84: just removed a wrapping div

dmadisetti:
Breaks `uv run marimo edit example/` (home) for me 
image

mscolnick:
@dmadisetti, thanks fixed

frontend/src/core/network/requests-lazy.ts
```diff
@@ -1,79 +1,175 @@
...
+          // Wait for connection but don't start it
+          await waitForConnectionOpen();
+          await waitForKernelToBeInstantiated();
+          return request(...args);
+
+        case "startConnection":
+          // Start connection and wait for it to be open
+          await initOnce(runtimeManager);
+          await waitForConnectionOpen();
+          if (key !== "sendInstantiate") {
```
manzt, line 155: Maybe worth adding a comment explaining why `sendInstantiate` is exempt (it's the one that creates the instantiation?)

frontend/src/core/network/requests-lazy.ts
```diff
@@ -1,79 +1,175 @@
...
-        await waitForKernelToBeInstantiated();
+      const runtimeManager = getRuntimeManager();
+
+      if (!runtimeManager.isLazy) {
+        return request(...args);
+      }
+
+      switch (action) {
+        case "dropRequest":
+          // Silently drop the request
```
manzt, line 139: Maybe consider logging?

frontend/src/core/network/requests-lazy.ts
```diff
@@ -1,79 +1,175 @@
...
-  "sendInterrupt",
-  "sendPdb",
-  "sendRun",
-  "sendRunScratchpad",
-]);
+// We have various requests that act different when called and not connected to a Kernel. They are:
+// - throw an Error
+// - drop the request
+// - start the connection
+// - wait for the connection to be open
```
manzt, line 19:
```suggestion
// We have various requests that act differently when called and not connected to a Kernel:
//
// - throwError: Throws NoKernelConnectedError, caught by requests-toasting.tsx
//   and shown as a toast with a "Connect" button. Use for operations that
//   shouldn't silently fail but also shouldn't auto-start the kernel.
//
// - dropRequest: Silently returns undefined. Only for requests where failure is
//   expected and doesn't matter (e.g., background polling).
//
// - startConnection: Initializes the runtime and waits for connection before
//   executing. Use for user-initiated actions that should "just work" and
//   kick off the kernel if needed (e.g., clicking Run).
//
// - waitForConnectionOpen: Waits for an existing connection but won't start one.
//   Use for operations that depend on a running kernel but shouldn't be the
//   trigger to start it (e.g., saving, interrupting).
```

manzt (commented):
Sorry for the post-merge review. Nice implementation, just a couple clarifying comments.

frontend/src/core/network/requests-lazy.ts
mscolnick: thanks, great comments, followed up with all them here: https://github.com/marimo-team/marimo/pull/7710

PR #7692, opened 2026-01-06
add softer shadows on hovering a cell

Light2Dark:
## 📝 Summary

Adds a hover shadow over a cell
- hovering over a cell is more distinguishable, 
- a softer transition to thicker shadow when cell is focused.

https://github.com/user-attachments/assets/5cf2d8a5-3615-4a69-8faf-c7e17da8103f

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [ ] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

PR #7700, opened 2026-01-06
style: shadow adjustments

m-rgba:
## 📝 Summary

| Before | After |
| -------- | ------- |
| image | image |

https://github.com/user-attachments/assets/2b809d35-eff3-4733-993f-58b06dd3bbca

- Adds hover state (darker border) to default / has-error / needs-run cells.
- Uses 2px border for inactive state for emphasized cells (has-error / needs-run) - instead of 1px / solid shadow.
  - Reverts to 1px when focused.
- Adds `tabIndex` to error output (so that clicking outputs triggers focus-within vs our outline).

### DIFF
--- a/frontend/src/css/app/Cell.css
+++ b/frontend/src/css/app/Cell.css
@@ -159,15 +159,15 @@
   /* Error styles for Cell, less precedence than needs run */
   &.has-error:not(.needs-run),
   &.error-outline {
-    @apply border-error/20 border-2 divide-error/20;
+    @apply border-error/20 border-1 divide-error/20 outline-1 outline-error/20;
   }
 
   &.has-error:not(.needs-run):hover {
-    @apply border-error/30;
+    @apply border-error/30 outline-error/30;
   }
 
   &.has-error:not(.needs-run):focus-within {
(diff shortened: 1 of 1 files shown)


m-rgba:

mscolnick:
@m-rgba 1px to 2px border swapping is going to cause a layout shift, which we would like to avoid

m-rgba:
> @m-rgba 1px to 2px border swapping is going to cause a layout shift, which we would like to avoid

@mscolnick gotcha, that's fair - we could look at something like an outline pattern to have a similar effect without the shift on focus. Updated with an example of that.

https://github.com/user-attachments/assets/8da55cac-e760-4405-b135-c5d11ada3ddc

manzt (approved):
Thanks for iterating on this! The outline approach looks good and addresses the layout shift concern.

We're aligned with the direction here. The hover state improvement and cleaner shadow treatment are nice additions. We had some parallel work happening on our side (#7692) which unfortunately led to some wasted effort. These changes are definitely in the direction we want, but in the future it would be great to align on goals before diving into implementation so we can avoid stepping on each other's toes.

PR #7640, opened 2025-12-30
draft: proof of concept for ipc kernels

dmadisetti:
## 📝 Summary

Proof of concept regarding the feasibility of IPC controlled kernels. I expect to essentially rewrite this
Didn't want to go too deep with Lazy kernels also likely changing this code.

- removes dangerous sandbox
- spawns ipc process for edit mode

Considerations on revisiting:
 - sandbox edge cases
 - run mode
 - performance hit with processes (maybe only run in sandbox mode, retain original code)
 - stdout / stderr streaming during sandbox setup

PR #7702, opened 2026-01-06
feat: sandboxed home

### LINKED ISSUE #2598: make sandbox compatibel with edit server (multiple notebooks)

dmadisetti:
## 📝 Summary

Adds `--sandbox` to `marimo edit --sandbox directory/`. Leveraging pep-723 where ever it can.

closes #2598 rebases and supercedes #7640

 - removes dangerous sandbox
 - spawns ipc process _only_ for `marimo edit --sandbox` (home mode)
 - uses lazy front end

UX changes on kernel launch failure:

image

### DIFF
--- a/marimo/_session/notebook/file_manager.py
+++ b/marimo/_session/notebook/file_manager.py
@@ -193,7 +193,9 @@ class AppFileManager:
             if GLOBAL_SETTINGS.MANAGE_SCRIPT_METADATA:
                 from marimo._utils.scripts import write_pyproject_to_script
 
-                header = write_pyproject_to_script({"dependencies": ["marimo"]})
+                header = write_pyproject_to_script(
+                    {"dependencies": ["marimo"]}
+                )
 
         # Rewrap with header if relevant and set filename.
         notebook = NotebookSerializationV1(


akshayka (commented):
Before releasing this feature, we should make sure the UX issues you alluded to in the PR description, in particular when the frontend handles the case in which sandbox notebooks fail to start (e.g., `uv sync` fails):

* the error should be communicated to the user;
* we should either fail fast (show the error and stop trying to connect to the kernel), or provide a way for the user to fix the error and try again (edit script metadata, then click a button to try creating the venv again)

If you prefer to work in multiple PRs, `sandbox home` could be feature flagged until the FE handles the failure case

marimo/_cli/cli.py
```diff
@@ -508,14 +494,22 @@ def edit(
...
+    home_sandbox = is_home_sandbox_mode(sandbox_mode, name)
+    if home_sandbox:
+        # Check for pyzmq dependency
+        from marimo._dependencies.dependencies import DependencyManager
+
+        if not DependencyManager.has("zmq"):
+            raise click.UsageError(
+                "pyzmq is required for sandbox home mode.\n"
+                "Install it with: pip install 'marimo[sandbox]'\n"
+                "Or: pip install pyzmq"
```
akshayka: In documentation and errors, I like to avoid the word "mode" and instead be descriptive.

```suggestion
                "pyzmq is required when running the marimo edit server on a directory with --sandbox.\n"
                "Install it with: pip install 'marimo[sandbox]'\n"
                "Or: pip install pyzmq"
```

marimo/_cli/sandbox.py
```diff
@@ -399,3 +393,165 @@ def handler(sig: int, frame: Any) -> None:
     signal.signal(signal.SIGINT, handler)
 
     return process.wait()
+
+
+# Dependencies required for IPC kernel communication (ZeroMQ-based)
+IPC_KERNEL_DEPS: list[str] = ["pyzmq", "msgspec"]
```
mscolnick: we already require `msgspec`, is that still needed?

marimo/_session/managers/ipc.py
```diff
@@ -0,0 +1,302 @@
...
+        if self._process.poll() is None:
+            self._process.terminate()
+            try:
+                self._process.wait(timeout=5)
+            except subprocess.TimeoutExpired:
+                self._process.kill()
+
+        # Clean up sandbox directory
+        cleanup_sandbox_dir(self._sandbox_dir)
+        self._sandbox_dir = None
```
akshayka: nit: delete redundant comments

```suggestion
        self.queue_manager.put_control_request(commands.StopKernelCommand())
        self.queue_manager.close_queues()

        # Terminate process if still alive
        if self._process.poll() is None:
            self._process.terminate()
            try:
                self._process.wait(timeout=5)
            except subprocess.TimeoutExpired:
                self._process.kill()

        cleanup_sandbox_dir(self._sandbox_dir)
        self._sandbox_dir = None
```

marimo/_session/managers/kernel.py
```diff
@@ -255,6 +162,8 @@ def pid(self) -> int | None:
 
     @property
     def profile_path(self) -> str | None:
+        from uuid import uuid4
```
mscolnick: unused?

marimo/_session/managers/kernel.py
```diff
@@ -1,138 +1,43 @@
...
 
-    def put_input(self, text: str) -> None:
-        """Put an input request in the input queue."""
-        self.input_queue.put(text)
+class KernelManagerImpl(KernelManager):
+    """Kernel manager using multiprocessing Process or threading Thread.
 
+    Uses Process for edit mode (allows SIGINT interrupts) and Thread for
+    run mode (lower memory overhead).
+    """
```
mscolnick, line 40: niiice

marimo/_session/managers/ipc.py
```diff
@@ -0,0 +1,302 @@
...
+LOGGER = _loggers.marimo_logger()
+
+
+class KernelStartupError(Exception):
+    """Raised when kernel subprocess fails to start."""
+
+    pass
+
+
+class IPCQueueManagerImpl(QueueManager):
```
mscolnick, line 45: might be more to share with the `QueueManagerImpl` (or instantiate this from `QueueManagerImpl`. e.g. `IPCQueueManagerImpl.from_ipc(queues)`

marimo/_server/api/endpoints/assets.py
```diff
@@ -144,6 +145,27 @@ async def index(request: Request) -> HTMLResponse:
...
+        # Pre-compute notebook snapshot for faster initial render
+        # Only in home sandbox mode where each notebook gets its own IPC kernel
+        notebook_snapshot = None
+        if (
+            app_state.session_manager.home_sandbox_mode
+            and app_manager.filename
+        ):
+            from marimo._convert.converters import MarimoConvert
+
+            filepath = Path(app_manager.filename)
```
mscolnick: could use our `AsyncPath` util

pyproject.toml
```diff
@@ -87,10 +87,17 @@ sql = [
...
+sandbox = [
+    "pyzmq>=27.1.0",           # IPC communication for sandbox kernels
+    "uv>=0.9.21",              # for sandbox management
+]
+
 # List of deps that are recommended for most users
 # in order to unlock all features in marimo
 recommended = [
     "marimo[sql]",
+    "marimo[sandbox]",          # For sandbox home mode
```
akshayka:
```suggestion
    "marimo[sandbox]",          # For `marimo edit --sandbox DIRECTORY`
```

akshayka (commented):
Shutting down the server on the frontend should also shutdown running notebooks, but that does not appear to be happening right now

marimo/_server/api/endpoints/assets.py
dmadisetti: I thought we had a better way. Thanks for the reminder

frontend/src/core/websocket/useMarimoKernelConnection.tsx
```diff
@@ -134,6 +135,11 @@ export function useMarimoKernelConnection(opts: {
       case "interrupted":
         return;
 
+      case "kernel-startup-error":
```
mscolnick, line 138: should we just re-use the generic "banner" notification? less code and more consistent.

marimo/_cli/cli.py
```diff
@@ -508,15 +494,32 @@ def edit(
...
-        # TODO: consider adding recommended as well
         run_in_sandbox(sys.argv[1:], name=name, additional_features=["lsp"])
         return
 
+    # Home sandbox mode: use IPC kernels with per-notebook sandboxed venvs
+    if home_sandbox:
+        # Check for pyzmq dependency
+        from marimo._dependencies.dependencies import DependencyManager
+
+        if not DependencyManager.has("zmq"):
```
mscolnick: `DependencyManager.zmq.has()`?

marimo/_messaging/notification.py
```diff
@@ -246,6 +246,13 @@ class BannerNotification(Notification, tag="banner"):
     action: Optional[Literal["restart"]] = None
 
 
+class KernelStartupErrorNotification(Notification, tag="kernel-startup-error"):
```
mscolnick, line 249: we have `StartupLog` and `Banner` (which can be used for errors). i wonder if that might be better to piggy back off

marimo/_server/api/endpoints/ws_endpoint.py
```diff
@@ -425,6 +432,23 @@ async def _close_already_connected(self) -> None:
...
+        """Send full error as message, then close the WebSocket."""
+        if self.websocket.application_state is WebSocketState.CONNECTED:
+            # Send full error as a message before closing
+            # Format must match what ws_message_loop sends:
+            # {"op": "xxx", "data": {...}}
+            msg = serialize_kernel_message(
+                KernelStartupErrorNotification(error=error_message)
+            )
+            text = f'{{"op": "kernel-startup-error", "data": {msg.decode("utf-8")}}}'
+            await self.websocket.send_text(text)
```
mscolnick, line 443: `self.session.notify( KernelStartupErrorNotification(error=error_message))`

mscolnick: oh maybe you don't have the session? this bit feels fragile

marimo/_cli/cli.py
```diff
@@ -545,6 +548,8 @@ def edit(
         server_startup_command=server_startup_command,
         asset_url=asset_url,
         timeout=timeout,
+        sandbox_mode=sandbox_mode,
+        home_sandbox_mode=home_sandbox,
```
mscolnick: 1. are these mutually exclusive? should we combine into a single flag?
2. `home_sandbox_mode` feels like it may become a misnomer quite soon. should it be `sandbox_mode: "none" | "single" | "multi"`?

frontend/src/core/websocket/useMarimoKernelConnection.tsx
dmadisetti: I do think it needs to be more visually - "this will not work"

frontend/src/components/editor/KernelStartupErrorModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    <AlertDialog open={true} onOpenChange={(open) => !open && handleClose()}>
+      <AlertDialogContent className="max-w-2xl">
+        <AlertDialogHeader>
+          <AlertDialogTitle className="flex items-center gap-2 text-destructive">
+            <XCircleIcon className="h-5 w-5" />
+            Kernel Startup Failed
+          </AlertDialogTitle>
+          <AlertDialogDescription>
+            The kernel failed to start. This usually happens when there's an
+            issue with the sandbox environment or missing dependencies.
```
akshayka:
```suggestion
            The kernel failed to start. This usually happens when the package manager
            can't install your notebook's dependencies. 
```

marimo/_server/api/endpoints/ws_endpoint.py
dmadisetti: Good point. Changing this up, an advantage to having a session is that we could potentially stream the errors/ startup before closing. Will keep that for a future PR because this is quiet long, but will do the non-blocking kernel launch

akshayka (commented):
Might not be related to this change, but I created a new notebook and `marimo` was not added to its script metadata:

```python
# /// script 
# requires-python = ">=3.13"
# dependencies = [
#     "matplotlib==3.10.8",
# ]
# ///
```

Additionally, I am finding a bad interaction with format on save, see video:

https://github.com/user-attachments/assets/31a8ac9a-8d83-4b5f-8606-ddd24bde43ca

marimo/_server/api/endpoints/ws_endpoint.py
dmadisetti: I did try to refactor this such that we:
 - async attempted startup kernel
 - started session and the web socket loop
 - push notifications from kernel startup directly to the queue

This also has the advantage of being able to stream messages during startup
image

In theory. In practice the installation modal popped up after it was done (not sure what I was missing there), I had a couple race condition issues where kernel closed out before WS was ready.

There's also not really a clear mechanism to turn off the session from inside the kernel, and started going down the route of a process interrupt- I didn't want to rely on the FE for teardown since it was leaving orphaned sessions. Then remembered if we just fail before we set up our scaffolding, then we don't have to do all this- so I reverted and just cleaned up.

Maybe we can revisit, but it was looking to be like a fair amount of more work on an already long PR. Maybe there's a simpler way but seemed marginal for just a slight code smell

dmadisetti:
marimo not being added a a bug particular to this (but we should fix). Opening the motebook again and it works as expected.

For the ruff error, I'm just going to throw an error and revisit if we need to. Ruff is installed in the sandbox, but the wrong sandbox! Installation path goes through the kernel

mscolnick (approved):

dmadisetti:
Sync'd with Akshay offline- gave his approval !

PR #7715, opened 2026-01-07
style: small fixes for cell shadows and hover

akshayka:
* Add hover border for disabled cells
* Dark-mode hover styling on base cell
* Remove tab-index on outputs, which conflicted with entering command-mode when clicking on outputs

### DIFF
--- a/frontend/src/css/app/Cell.css
+++ b/frontend/src/css/app/Cell.css
@@ -294,7 +294,7 @@
   }
 
   &.needs-run:hover {
-      @apply border-action-foreground/30 outline-action-foreground/30;
+    @apply border-action-foreground/30 outline-action-foreground/30;
   }
 
   &.has-error:hover {


frontend/src/css/app/Cell.css
```diff
@@ -285,6 +289,10 @@
 .dark .marimo-cell {
   @apply border-border;
 
+  &:hover {
```
Light2Dark, line 292: this applies to needs-run as well, maybe

```suggestion
  &:hover:not(.needs-run) {
```

Light2Dark: but this breaks has-error hover for dark mode. Might need a new rule just for that

akshayka: Added rules for dark needs-run and has-error hovers, thanks!

PR #7570, opened 2025-12-19
refactor: SessionConsumer as a SessionExtension

mscolnick:
This is further refactoring to make SessionConsumer a type of extension (plus some).

PR #7722, opened 2026-01-07
fix: relative path fixes for home mode

dmadisetti:
## 📝 Summary

Solves some ux issues with `marimo edit subdir/`
1. Starting a new notebook will default to cwd
2. You can still save to cwd (but refreshing caused an auth error)
3. Recent notebooks are shown in cwd
4. File tree shows cwd

### DIFF
--- a/tests/_server/test_file_manager_absolute_path.py
+++ b/tests/_server/test_file_manager_absolute_path.py
@@ -4,12 +4,15 @@
 from __future__ import annotations
 
 import os
+import sys
 from pathlib import Path
 
 import pytest
 
 from marimo._server.file_router import AppFileRouter
 from marimo._utils.http import HTTPException, HTTPStatus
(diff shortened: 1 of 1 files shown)


marimo/_server/recents.py
```diff
@@ -86,19 +86,21 @@ def get_recents(self) -> list[MarimoFile]:
         )
         files: list[MarimoFile] = []
 
-        cwd = pathlib.Path.cwd()
+        base_dir = (directory or pathlib.Path.cwd()).resolve()
```
mscolnick: are you sure you want resolve here? should this just absolute path?

dmadisetti: Ran git bisect and found a regression in #7570 that broken "recent files" when verifying this. Will followup with a fix

PR #7723, opened 2026-01-07
Fix duplicate HTML5 backend error in file explorer

manzt:
The file explorer uses `react-arborist` which internally uses `react-dnd`. When multiple `Tree` components mount or when other `react-dnd` contexts exist, this causes the "Cannot have two HTML5 backends at the same time" error since react-dnd only allows a single HTML5Backend instance.

These changes add a shared DnD provider wrapper that creates a single `DragDropManager` instance scoped to the file explorer panel. The Tree component receives this manager via the `dndManager` prop, preventing it from creating its own conflicting backend. The wrapper also scopes the backend to a specific root element to avoid conflicts with any other drag-and-drop functionality elsewhere in the app.


frontend/package.json
```diff
@@ -144,6 +144,8 @@
     "react-aria": "^3.44.0",
     "react-aria-components": "^1.13.0",
     "react-codemirror-merge": "4.25.4",
+    "react-dnd": "^16.0.1",
+    "react-dnd-html5-backend": "^16.0.1",
```
mscolnick, line 149: react-aborist was on 14. im down for 16, but does this still work? I guess if the types work

manzt: Yeah the types seem to be fine and we dedupe in vite?

PR #7726, opened 2026-01-07
fix: disable shared memory in wasm

mscolnick:
We need to use InMemoryStorage (which may not even get used at all since virtual_files are disabled).

### DIFF
--- a/marimo/_runtime/context/kernel_context.py
+++ b/marimo/_runtime/context/kernel_context.py
@@ -19,7 +19,6 @@ from marimo._runtime.dataflow import DirectedGraph
 from marimo._runtime.functions import FunctionRegistry
 from marimo._runtime.params import CLIArgs, QueryParams
 from marimo._session.model import SessionMode
-from marimo._utils.platform import is_pyodide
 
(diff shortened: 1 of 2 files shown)


dmadisetti:
I think we disable virtual files all together in pyodide. We could potentially enable them?

mscolnick:
@dmadisetti we don't have a webserver for wasm so we cant serve virtual files

marimo/_runtime/context/kernel_context.py
```diff
@@ -159,9 +160,10 @@ def create_kernel_context(
 
     # Use shared memory in edit mode,
     # in-memory storage in run mode (same process)
+    # Pyodide doesn't support shared memory, so we use in-memory storage.
     storage = (
         SharedMemoryStorage()
-        if mode == SessionMode.EDIT
+        if mode == SessionMode.EDIT and not is_pyodide()
```
dmadisetti: Right, this change is basically a Noop (aside from the arg change), since the storage isn't applicable

I'd remove this since it implies storage will work

PR #7727, opened 2026-01-07
fix: hide terminal if not available

mscolnick:
Terminal doesn't need to be shown for wasm or if not supported (some windows machines)


dmadisetti (approved):
Nice

PR #7742, opened 2026-01-08
Move minimap into "Dependencies" panel with tabbed UI

manzt:
The minimap and dependency graph viewer had overlapping functionality as separate components. This consolidates them into the dependencies panel with a [Minimap][Graph] tab switcher, following the same pattern as the AI panel's [Chat][Agents] tabs. It also removes the toggle from the footer.

The minimap tab shows the custom SVG-based cell visualization that was previously in the floating footer popover. The graph tab shows the React Flow tree views with vertical/horizontal layout options. The React Flow minimap view was removed since it duplicated the custom minimap functionality. The hotkey (`global.toggleMinimap`) now opens the dependencies panel with the minimap tab selected.

https://github.com/user-attachments/assets/b2d6f874-0915-46f0-a005-90f5c4b1f076


koaning:
Love it!

PR #7750, opened 2026-01-08
fix: add step part to ChatPart

mscolnick:


Light2Dark:
I think we can change to debug, im tracing if there needs to be a better solution

PR #7753, opened 2026-01-08
Fix terminal not appearing in developer panel

manzt:
`isPanelHidden` was called at module load time when computing `DEFAULT_PANEL_LAYOUT`, before capabilities were received from the backend. This meant terminal was filtered out and never added to the stored layout.

These changes check static `hidden` property at module init, defer capability checks to render time.

### DIFF
--- a/frontend/src/components/editor/chrome/types.ts
+++ b/frontend/src/components/editor/chrome/types.ts
@@ -18,7 +18,6 @@ import {
   VariableIcon,
   XCircleIcon,
 } from "lucide-react";
-import { hasCapability } from "@/core/config/capabilities";
 import { getFeatureFlag } from "@/core/config/feature-flag";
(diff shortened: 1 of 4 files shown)


frontend/src/components/editor/chrome/types.ts
```diff
@@ -188,12 +187,15 @@ export const PANEL_MAP = new Map<PanelType, PanelDescriptor>(
  * Check if a panel should be hidden based on its `hidden` property
  * and `requiredCapability`.
  */
-export function isPanelHidden(panel: PanelDescriptor): boolean {
+export function isPanelHidden(
+  panel: PanelDescriptor,
+  capabilities: Capabilities,
```
manzt, line 192: I changed this signature so we can provide capabilities atom within the component.

PR #7771, opened 2026-01-10
add custom AI providers in config

### LINKED ISSUE #7243: Allow adding named custom AI providers, so that adding a custom model is tied to the specific provider.
Using custom models with custom providers does not feel clear from the UI. One suggestion is to allow users to name the custom provider. Then adding a custom model can be tied to the custom provider added.

### Discussed in https://github.com/marimo-team/marimo/discussions/7015

<sup>Originally posted by **zhh210** October 31, 2025</sup>
Having trouble understanding the usage and what is the difference between AI Providers and AI Models tab. There are quite a few overlaps between the two but I don't see which 'Model' is associated with the 'OpenAI-Compatible Provider':
image
image
Also if I were to configure my AI assistant (AI Features tab) and choosing a "Model", is it referring to the AI Providers tab or the AI Models tab? (Seems to be referring to AI Models tab but in that case from the error). I've configured an OpenAI-Compatible provider from the AI Providers tab but chat keeps giving an error of 
`OpenAI API key not configured. Go to Settings > AI to configure.`

Also the xx AI Model doesn't use xx AI Provider as when I specify base_url and token on OpenAI provider and use OpenAI model it gives `Error code: 404 - {'detail': 'Not Found'}`

The AI Models tab does allow adding a model but you'll need to specify the 'Provider' which only shows the list from the AI Models tab but no 'OpenAI-Compatible' option.
</div>

Light2Dark:
## 📝 Summary

Fixes #7243.

image

image

image

## 📋 Checklist

### DIFF
--- a/docs/guides/configuration/llm_providers.md
+++ b/docs/guides/configuration/llm_providers.md
@@ -381,6 +381,7 @@ Add multiple OpenAI-compatible providers through the Settings UI. Each custom pr
 1. Open marimo's Settings panel
 2. Navigate to **AI** → **AI Providers**
 3. Scroll to **Custom Providers** and click **Add Provider**
+4. Navigate to the **AI Models** tab and select your custom provider.
 
 <div align="center">
 <video autoplay muted loop playsinline width="640px" height="480px">
@@ -388,14 +389,6 @@ Add multiple OpenAI-compatible providers through the Settings UI. Each custom pr
   Your browser does not support the video tag.
 </video>
 </div>
-
-Once added, use your custom provider with the prefix you chose:
-
-```toml title="marimo.toml"
-[ai.models]
-chat_model = "groq/llama-3.1-70b-versatile"
-edit_model = "together/meta-llama/Llama-3-70b-chat-hf"
-```
 
 **Configuration via marimo.toml**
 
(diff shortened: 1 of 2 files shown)


docs/guides/configuration/llm_providers.md
```diff
@@ -365,53 +366,60 @@ autocomplete_model = "ollama/codellama" # or another model from `ollama ls`
...
+3. Scroll to **Custom Providers** and click **Add Provider**
+
+<div align="center">
+<video autoplay muted loop playsinline width="640px" height="480px">
+  <source src="/_static/docs-add-custom-provider.mp4" type="video/mp4">
+  Your browser does not support the video tag.
+</video>
+</div>
+
+Once added, use your custom provider with the prefix you chose:
```
akshayka: Should users specify their models directly in the `marimo.toml` file, or in the settings UI? As written it seems like the recommendation is the `marimo.toml` since that is what follows this sentence

akshayka (commented):
This is really nice! One clarification may be needed in the docs (see my comment), otherwise looks good to me

docs/guides/configuration/llm_providers.md
Light2Dark: thanks! fixed

Issue #7774, opened 2026-01-10
Flaky test: test_running_notebooks_handles_files_outside_directory times out on Windows


Ravinlathigra:
## Description

The test `test_running_notebooks_handles_files_outside_directory` in `tests/_server/api/endpoints/test_home.py` intermittently times out on Windows CI (specifically `windows-latest / Py 3.10 / core deps`).

## Error

```
Timeout (0:00:30)!
...
File "D:\a\marimo\marimo\tests\_server\api\endpoints\test_home.py", line 289, in test_running_notebooks_handles_files_outside_directory
    with client.websocket_connect(
  File "...\starlette\testclient.py", line 123, in __exit__
    return self.exit_stack.__exit__(*args)
  ...
  File "...\starlette\testclient.py", line 174, in close
    self.send({"type": "websocket.disconnect", "code": code, "reason": reason})
```

## Root Cause

The test hangs during websocket cleanup/disconnect. The stack trace shows threads stuck in:
- `watch_file` (`marimo/_runtime/watch/_file.py:30`)
- `watch_directory` (`marimo/_runtime/watch/_directory.py:54`)

The file watchers use daemon threads that loop with `time.sleep()` and check `should_exit.is_set()`. Cleanup relies on `__del__` in `PathState` to call `_should_exit.set()`. On Windows, the garbage collector may not run `__del__` promptly during test teardown, leaving watcher threads running and blocking websocket disconnect.

## Impact

This flaky test blocks unrelated PRs from merging. For example, PR #7770 failed CI twice due to this test despite changes being unrelated to `test_home.py`.

## Environment

- OS: Windows (windows-latest)
- Python: 3.10
- Test matrix: core deps

## Suggested Fixes

1. **Add explicit cleanup in session teardown** - Call `PathState._should_exit.set()` explicitly during session cleanup rather than relying on `__del__` garbage collection
2. **Add explicit cleanup in test fixtures** - Ensure file watcher threads are explicitly stopped in test teardown
3. **Skip on Windows (workaround)** - Mark as `pytest.mark.skipif(sys.platform == 'win32')` if Windows-specific behavior isn't critical - not ideal as it reduces coverage
4. **Investigate `__del__` reliability on Windows** - The root issue is `__del__` not being called reliably during test cleanup on Windows

Light2Dark:
Sorry about that, we've been trying to address some of these flakey tests but we don't use Windows machines.

[...]

PR #7775, opened 2026-01-10
fix: shutdown before websocket disconnect in test

Ravinlathigra:
## 📝 Summary

  Partially addresses #7774

  Marks `test_running_notebooks_handles_files_outside_directory` as `xfail` on Windows to unblock CI while the root cause is investigated.

  ## 🔍 Description of Changes

  **What this PR does:**
  Adds `@pytest.mark.xfail(sys.platform == "win32", strict=False)` to the flaky test.

  **Why `xfail` instead of a real fix:**

  The root cause is complex and not fully understood. Investigation revealed:

  1. The test times out during websocket disconnect cleanup on Windows
  2. Stack traces show threads stuck in `marimo/_runtime/watch/_file.py` and `_directory.py` (the `mo.watch` module)
  3. These watcher threads use `time.sleep()` polling loops and rely on `__del__` to signal exit via `threading.Event`
  4. On Windows, `__del__` may not be called reliably during test teardown, leaving threads running

  **Why a proper fix is non-trivial:**

  - The watcher threads are daemon threads that should die with the process, but on Windows they can block cleanup
  - The threads may originate from previous tests (test isolation issue) rather than this specific test
  - The `mo.watch` module's cleanup relies on garbage collection (`__del__`), which behaves differently on Windows
  - A proper fix would require changes to production code (`marimo/_runtime/watch/_path.py`) to use explicit cleanup rather than `__del__`

  **What `xfail` does:**
  - If the test fails → reports as `XFAIL` (expected failure) - CI passes
  - If the test passes → reports as `XPASS` (unexpected pass) - CI passes
  - The test still runs, maintaining coverage

  ## 📋 Checklist

### DIFF
--- a/tests/_server/api/endpoints/test_home.py
+++ b/tests/_server/api/endpoints/test_home.py
@@ -275,6 +275,11 @@ def test_shutdown_session_returns_relative_paths(client: TestClient) -> None:
             # Note: session is already shut down, so we don't restore filename
 
 
+@pytest.mark.xfail(
+    sys.platform == "win32",
+    reason="Flaky on Windows - websocket cleanup may hang due to background threads. See #7774",
+    strict=False,
+)
 def test_running_notebooks_handles_files_outside_directory(
     client: TestClient,
 ) -> None:


dmadisetti (approved):
Thank you! Sorry for our flaky CI. Trying to get back consistently green!

PR #7782, opened 2026-01-11
Fix lint issues and set CI to error if there are more warnings

Light2Dark:
## 📝 Summary

- Fixes generic lint errors, 
- Sets CI to error if warnings are above threshold, so we can stop bleed more anti-patterns
- I did my best to grep existing IDs and see if they are used anywhere

## 📋 Checklist


frontend/src/components/editor/header/filename-input.tsx
```diff
@@ -56,7 +59,10 @@ export const FilenameInput = ({
 
   const onBlur = (evt: React.FocusEvent<HTMLInputElement>) => {
     // If we are coming from a click event from inside the popover, don't blur
-    if (evt.relatedTarget?.closest(".filename-input")) {
+    if (
+      evt.relatedTarget instanceof HTMLElement &&
+      evt.relatedTarget.closest(`[data-id="${FILENAME_INPUT_DATA_ID}"]`)
```
mscolnick, line 64: this is `data-id` which is not on this element. and we should not use `data-testid` as a proper locator. instead we can just use `id`

frontend/src/stories/textarea.stories.tsx
```diff
@@ -22,26 +23,32 @@ export const Disabled = {
...
-    <div className="grid w-full gap-1.5">
-      <Label htmlFor="message-2">Your Message</Label>
-      <Textarea placeholder="Type your message here." id="message-2" />
-      <p className="text-sm text-muted-foreground">
-        Your message will be copied to the support team.
-      </p>
-    </div>
-  ),
+  render: function Render() {
+    const id = useId();
```
mscolnick, line 41: there is a lot of churn in the stories. can we just disable these rules in `**/*.stories.tsx`

PR #7783, opened 2026-01-12
Fix: wait for thread completion in test client cleanup

akshayka:
This PR fixes a race condition in tests, identified in #7779, in which kernel threads would race against the test client fixture cleanup, with kernel threads incorrectly overwriting the module restored by the fixture. This PR consolidates client cleanup in the fixture teardown, and makes sure to join kernel threads before restoring the main module.


akshayka:
The test failure in `test_auto_instantiate` is really perplexing to me. I have marked as a draft because I don't know what could be causing that to fail

PR #7813, opened 2026-01-13
fix: more defensive edge cases when adding packages to inline metadata

mscolnick:
Previously if you installed deps like `crewai=0.177.0` (with a version) or `crewai_tools` (with an underscore) they were not added to the inline metadata since they were not in the version map. This fixes that

### DIFF
--- a/tests/_runtime/packages/test_package_managers.py
+++ b/tests/_runtime/packages/test_package_managers.py
@@ -15,6 +15,7 @@ from marimo._runtime.packages.pypi_package_manager import (
     PipPackageManager,
     RyePackageManager,
     UvPackageManager,
+    VersionMap,
 )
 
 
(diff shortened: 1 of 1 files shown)


marimo/_runtime/packages/pypi_package_manager.py
```diff
@@ -347,18 +393,15 @@ def _is_direct_reference(package: str) -> bool:
             return False
 
         def _is_installed(package: str) -> bool:
-            without_brackets = package.split("[")[0]
-            return without_brackets.lower() in version_map
+            return version_map.has(package)
```
dmadisetti, line 413: Nice

mscolnick:
i actually need to make sure `resolve_with_version` keeps the extras when resolved

PR #7821, opened 2026-01-14
feat: add [tool.marimo.venv].path to specify a specific venv

dmadisetti:
Introduces `[tool.marimo.venv].path` configuration option that allows users to specify a custom virtual environment for sandbox multi mode. This enables using predefined environments (conda, poetry, custom venvs) instead of ephemeral sandboxes.

## Configuration

In your notebook's script metadata or `pyproject.toml`:

```toml
[tool.marimo.venv]
path = "path/to/venv"      # relative or absolute path
writable = true            # optional: allow marimo to install deps (default: false)
```

## Behavior

| `editable` | marimo installed? | Action |
|:-----------|:------------------|:-------|
| `true` | - | Installs `marimo` + IPC deps (`pyzmq`) into the venv |
| `false` | Yes | Uses venv as-is, warns if version mismatch |
| `false` | No | Injects `PYTHONPATH` for marimo (requires matching Python version) |

Additionally, has some handling / logic for incompatible venvs:

image

## Error Handling

- **Invalid venv path**: Clear error message if venv doesn't exist or has no Python interpreter
- **Python version mismatch**: When `editable=false` and marimo isn't installed, checks that venv Python matches server Python (required for binary deps like `msgspec`, `pyzmq`)
- **Version mismatch warning**: Logs warning if venv's marimo version differs from current

Supercedes #7624

### DIFF
--- a/marimo/_session/managers/ipc.py
+++ b/marimo/_session/managers/ipc.py
@@ -225,7 +225,7 @@ class IPCKernelManagerImpl(KernelManager):
                         f"Options:\n"
                         f"  1. Set writable=true in [tool.marimo.venv] to allow marimo to install deps\n"
                         f"  2. Install marimo in your venv: uv pip install marimo --python {venv_python}\n"
-                        f"  3. Remove [tool.marimo.venv].venv to use an ephemeral sandbox instead"
+                        f"  3. Remove [tool.marimo.venv].path to use an ephemeral sandbox instead"
                     )
 
                 # Inject PYTHONPATH for marimo and dependencies from the


marimo/_config/config.py
```diff
@@ -97,6 +97,26 @@ class KeymapConfig(TypedDict):
...
+
+    Allows specifying an existing virtualenv to use instead of creating
+    ephemeral sandboxes per notebook. Only applies in home sandbox mode.
+
+    **Keys.**
+
+    - `venv`: path to a virtualenv directory (absolute or relative to
+      pyproject.toml)
+    - `editable`: if true, marimo will manage script metadata (inline
+      dependencies). Defaults to false.
```
akshayka: * If `editable` is False, will the marimo editor's package manager still install packages into `venv`?
* Is the terminology "editable" standard here, or is there another term we should use? It makes me think of an "editable installation" of a package, which I think is different from what we mean here

marimo/_session/managers/ipc.py
```diff
@@ -162,28 +175,81 @@ def start_kernel(self) -> None:
...
+                    install_marimo_into_venv(venv_python)
+                except Exception as e:
+                    raise KernelStartupError(
+                        f"Failed to install marimo into configured venv.\n\n{e}"
+                    ) from e
+            elif not has_marimo_installed(venv_python):
+                # Check Python version compatibility for binary deps
+                if not check_python_version_compatibility(venv_python):
+                    raise KernelStartupError(
+                        f"Configured venv uses a different Python version than marimo.\n"
```
akshayka, line 223: It sounds like ephemeral sandboxes can use different Python versions than the server process, but configured venv's can't. Is that correct, and if so why the asymmetry?

marimo/_session/managers/ipc.py
```diff
@@ -162,28 +175,81 @@ def start_kernel(self) -> None:
...
+            echo(
+                f"Running kernel in sandbox: {muted(venv_python)}",
+                err=True,
+            )
+
+        cmd = [venv_python, "-m", "marimo._ipc.launch_kernel"]
+        if editable:
+            # Setting this attempts to make auto-installations work even if
+            # other normally detected criteria are not true.
+            # IPC by itself does no seem to trigger them.
```
akshayka: In addition to this typo, can you explain what you mean here more? What do you mean by "other normally detected criteria"?

```suggestion
            # Setting this attempts to make auto-installations work even if
            # other normally detected criteria are not true.
            # IPC by itself does not seem to trigger them.
```

marimo/_session/managers/ipc.py
```diff
@@ -162,28 +175,81 @@ def start_kernel(self) -> None:
...
+                    raise KernelStartupError(
+                        f"Configured venv uses a different Python version than marimo.\n"
+                        f"Binary dependencies (pyzmq, msgspec) aren't cross-version compatible.\n\n"
+                        f"Options:\n"
+                        f"  1. Set editable=true in [tool.marimo.env] to allow marimo to install deps\n"
+                        f"  2. Install marimo in your venv: uv pip install marimo --python {venv_python}\n"
+                        f"  3. Remove [tool.marimo.env].venv to use an ephemeral sandbox instead"
+                    )
+
+                # Inject PYTHONPATH for marimo and dependencies
```
akshayka: Can you expand on this comment? I think these next few lines IIUC are the main logic to use the configured venv. But one basic quesiton I have is why do we need to set the `PYTHONPATH` when using a configured venv, but don't need to set it when using ephemeral sandboxes?

marimo/_cli/venv.py
```diff
@@ -0,0 +1,217 @@
...
+if TYPE_CHECKING:
+    from marimo._config.config import EnvConfig
+
+LOGGER = _loggers.marimo_logger()
+
+# Dependencies required for IPC kernel communication (ZeroMQ-based)
+IPC_KERNEL_DEPS: list[str] = ["pyzmq"]
+
+
+def get_marimo_dir() -> Path:
```
akshayka: Add a unit test for this, since it's implementation is a little brittle?

marimo/_config/config.py
dmadisetti: Editable is definitely confusing, because an "editable installation" generally an installation pointing to editable source.

Maybe `read-only`? Or `writable`?

The fact this is [tool.marimo.env] may also be a bit confusing too. `[tool.marimo.venv].path` seems much more scoped. Leaning this way (and `writable`) after a bit more thought

marimo/_cli/venv.py
```diff
@@ -0,0 +1,217 @@
...
+    in a configured venv that has marimo available via path injection.
+
+    Returns:
+        Colon-separated (or semicolon on Windows) path string for PYTHONPATH.
+    """
+    paths: list[str] = []
+
+    # Add marimo's parent directory (where marimo package can be imported from)
+    # NB. If running in edit mode this may be local directory.
+    marimo_parent = str(get_marimo_dir())
```
mscolnick: we also have `marimo_package_path` which might be more correct?

marimo/_session/_venv.py
```diff
@@ -0,0 +1,209 @@
...
+    if venv_version != __version__:
+        LOGGER.warning(
+            f"marimo version mismatch: venv has {venv_version}, "
+            f"current is {__version__}"
+        )
+
+    return True
+
+
+def check_python_version_compatibility(venv_python: str) -> bool:
```
mscolnick, line 175: could we add a basic test for this?

marimo/_session/_venv.py
```diff
@@ -0,0 +1,209 @@
...
+                module_path_str = str(module_path)
+                if module_path_str not in paths:
+                    paths.append(module_path_str)
+        except ImportError:
+            pass
+
+    return os.pathsep.join(paths)
+
+
+def has_marimo_installed(venv_python: str) -> bool:
```
mscolnick, line 141: could we add a basic test for this?

marimo/_session/_venv.py
```diff
@@ -0,0 +1,209 @@
...
+        text=True,
+    )
+    if result.returncode != 0:
+        return False
+
+    venv_version = result.stdout.strip()
+    if venv_version != __version__:
+        LOGGER.warning(
+            f"marimo version mismatch: venv has {venv_version}, "
+            f"current is {__version__}"
```
mscolnick: what happens if we hit a version mismatch?

marimo/_config/config.py
mscolnick: `[tool.marimo.venv].path` seems better to me as well

marimo/_session/_venv.py
dmadisetti: Trigger a warning, but still try to launch. We can just fail- but figured this was sufficent

marimo/_session/managers/ipc.py
dmadisetti: Let me know if the new comments address this

marimo/_cli/venv.py
dmadisetti: Removed this call from the path

marimo/_session/_venv.py
dmadisetti: Done!

marimo/_session/managers/ipc.py
```diff
@@ -162,28 +175,96 @@ def start_kernel(self) -> None:
...
+                    # - The venv does not contain marimo nor zmq
+                    # As such there is nothing we can do, as we can't get marimo
+                    # into the runtime without installing it somewhere else.
+                    raise KernelStartupError(
+                        f"Configured venv uses a different Python version than marimo.\n"
+                        f"Binary dependencies (pyzmq, msgspec) aren't cross-version compatible.\n\n"
+                        f"Options:\n"
+                        f"  1. Set writable=true in [tool.marimo.venv] to allow marimo to install deps\n"
+                        f"  2. Install marimo in your venv: uv pip install marimo --python {venv_python}\n"
+                        f"  3. Remove [tool.marimo.venv].venv to use an ephemeral sandbox instead"
```
mscolnick: `[tool.marimo.venv].path`

dmadisetti:
```suggestion
                        f"  3. Remove [tool.marimo.venv].path to use an ephemeral sandbox instead"
```

PR #7824, opened 2026-01-14
fix: deflake test_imported_app_has_prefixed_setup_cell

akshayka:
This change fixes a flaky test that checked the context (runtime context or not) in which an imported app object was created.

Use `importlib.reload()` to force the notebook module to be re-executed in the kernel context. This fixes flakiness caused by Python's module caching — if the module was previously imported without a kernel context, the app would have unprefixed cell IDs.

### DIFF
--- a/.github/workflows/test_be.yaml
+++ b/.github/workflows/test_be.yaml
@@ -102,7 +102,7 @@ jobs:
     if: ${{ needs.changes.outputs.backend == 'true' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'test-all')) }}
     name: ${{ matrix.os }} / Py ${{ matrix.python-version }} / ${{ matrix.dependencies }} deps
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 20
+    timeout-minutes: 25
     defaults:
       run:
         shell: bash


akshayka:
Do we need to increase the time limit for py 3.13 core optional? https://github.com/marimo-team/marimo/actions/runs/20982753702/job/60311581139?pr=7824

dmadisetti:
Hmm, it shouldn't be running all those tests:

Found 1 affected test files:
  - tests/_ast/test_app.py

akshayka:
@dmadisetti it looks like we include everything for 3.13 optional deps: 

```
173           # Set INCLUDE_UNCHANGED: true if test-all label, python 3.13 + optional deps, or main branch  ci: support test-all label (#6776) • 10-14-2025 12:32:44 • Myles Scolnick • <03cb04d>
  1           if [[ "${{ contains(github.event.pull_request.labels.*.name, 'test-all') }}" == "true" ]] || \
  2              [[ "${{ matrix.python-version }}" == "3.13" && "${{ matrix.dependencies }}" == "core,optional" ]] || \
  3              [[ "${{ github.ref }}" == "refs/heads/main" ]]; then
  ```
  
 I bumped the timeout to 25 minutes since we pretty often see backend timeouts it seems

mscolnick (approved):

PR #7828, opened 2026-01-14
Move marimo import to top in convert command

koaning:
This PR moves the `import marimo as mo` to the top cell when we convert from Jupyter to marimo. Having the cell at the bottom is bothersome because it may not run at all if the original notebook contains errors.


akshayka (dismissed):

akshayka:
@koaning you will need to run the tests locally and commit the updated snapshots

mscolnick (dismissed):

koaning:
@akshayka ping, I think it's all fixed now.

mscolnick (approved):

PR #7876, opened 2026-01-16
fix: normalize git urls in conversion from jupyter to marimo deps

mscolnick:
normalize git urls in conversion from jupyter to marimo deps

### DIFF
--- a/marimo/_convert/ipynb/to_ir.py
+++ b/marimo/_convert/ipynb/to_ir.py
@@ -415,16 +415,10 @@ def _normalize_git_url_package(package: str) -> str:
     Returns the package as-is if it's not a git URL.
     """
     # Check if this looks like a git URL or VCS URL
-    # Common patterns: git+https://, git+ssh://, git+git://, https://github.com/...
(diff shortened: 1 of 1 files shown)


marimo/_convert/ipynb/to_ir.py
```diff
@@ -403,6 +403,64 @@ class ExclamationMarkResult:
...
+    Converts:
+        git+https://github.com/user/repo.git
+    To:
+        repo @ git+https://github.com/user/repo.git
+
+    Returns the package as-is if it's not a git URL.
+    """
+    # Check if this looks like a git URL or VCS URL
+    # Common patterns: git+https://, git+ssh://, git+git://, https://github.com/...
+    is_git_url = package.startswith("git+") or (
```
dmadisetti, line 419: Opposed to hardcode I'd `re.match(r"git\+.*/([^\.]+).git$")`

dmadisetti: or `urlparse(package.split("+")).pop()` I don't think we should confine to github/ gitlab etc.

dmadisetti (approved):

PR #7879, opened 2026-01-17
fix: preserve trailing comments in pure functions, setup cell

akshayka:
Before this change, trailing comments in pure functions were stripped on load. That meant that a function

```python
@app.function
def f():
  # BEGIN YOUR CODE
  ...
  # END YOUR CODE
```

was incorrectly parsed as

```python
def f():
  # BEGIN YOUR CODE
  ...
```

### DIFF
--- a/tests/_ast/test_parse.py
+++ b/tests/_ast/test_parse.py
@@ -242,7 +242,7 @@ class TestTrailingComments:
     @staticmethod
     def test_function_with_trailing_comment() -> None:
         """@app.function should preserve trailing comments at body indentation."""
-        code = '''import marimo
+        code = """import marimo
 
 __generated_with = "0.0.0"
 app = marimo.App()
@@ -257,7 +257,7 @@ def f():
 
 if __name__ == "__main__":
     app.run()
-'''
+"""
         notebook = parse_notebook(code)
         assert len(notebook.cells) == 1
         assert "# end" in notebook.cells[0].code
@@ -265,7 +265,7 @@ if __name__ == "__main__":
     @staticmethod
     def test_function_with_trailing_comment_and_empty_line() -> None:
         """@app.function should preserve trailing comments at body indentation."""
-        code = '''import marimo
+        code = """import marimo
 
 __generated_with = "0.0.0"
 app = marimo.App()
@@ -281,7 +281,7 @@ def f():
 
 if __name__ == "__main__":
     app.run()
-'''
+"""
         notebook = parse_notebook(code)
         assert len(notebook.cells) == 1
         assert "# end" in notebook.cells[0].code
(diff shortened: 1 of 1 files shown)


marimo/_ast/parse.py
```diff
@@ -363,11 +363,54 @@ def to_cell(
             }
             cell_type = cell_types.get(attribute, None)
             if cell_type is not None:
-                code = self.extract_from_code(node)
-                violations.extend(code.violations)
+                code_result = self.extract_from_code(node)
+                violations.extend(code_result.violations)
+                extracted_code = code_result.unwrap()
+
+                # Capture trailing comments after the AST's end position.
```
dmadisetti: I think this would be better to put in `extract_from_code` itself since we use it multiple places

tests/_ast/test_parse.py
```diff
@@ -236,6 +236,179 @@ def test_unparsable_cell_with_non_string_creates_violation() -> None:
...
+
+__generated_with = "0.0.0"
+app = marimo.App()
+
+
+@app.class_definition
+class MyClass:
+    def method(self):
+        pass
+    # trailing class comment
```
dmadisetti, line 347: Good check

marimo/_ast/parse.py
dmadisetti: Yeah, `setup_cell` may have the same issue

tests/_ast/test_parse.py
```diff
@@ -236,6 +236,179 @@ def test_unparsable_cell_with_non_string_creates_violation() -> None:
...
+if __name__ == "__main__":
+    app.run()
+"""
+        notebook = parse_notebook(code)
+        assert len(notebook.cells) == 1
+        assert "# comment 1" in notebook.cells[0].code
+        assert "# comment 2" in notebook.cells[0].code
+
+    @staticmethod
+    def test_function_without_trailing_comment() -> None:
```
dmadisetti, line 382: Think we have other tests for this, but we do need one for `setup_cell`

dmadisetti: Checking comments after return might be unexpected though (especially if we do change `extract_from_code`)

marimo/_ast/parse.py
akshayka: @dmadisetti is there a reason we strip `pass` in the setup cell? I found this surprising in testing, it seems incorrect

https://github.com/marimo-team/marimo/blob/fc3d0fdd2bfcdd5d91b4ae8ad359a6634065e161/marimo/_ast/parse.py#L316-L317

akshayka: Okay I understand actually. It's an edge case that users won't hit most likely, fine to keep

tests/_ast/test_parse.py
akshayka: Added a setup cell test

mscolnick (approved):

PR #7881, opened 2026-01-17
Fix SQL f-string placeholder causing parsing errors in interval expressions

### LINKED ISSUE #7717: Spurious SQL parsing error logged when f-string variable is used in interval expression
### Describe the bug

I encountered a quasi-bug today that wasted a couple hours of my time today because I thought my script was raising an error when it wasn't.

Here is Claude's summary of the problem:

When using an f-string variable inside a SQL interval expression in mo.sql(), a confusing error is logged to stderr during static analysis, even though the query executes successfully.

  The error occurs because normalize_sql_f_string() in _ast/sql_visitor.py:77-98 replaces f-string variables with the literal string "null" for dependency analysis. This creates invalid SQL like interval 'null days' which fails to parse when passed to find_sql_refs() in _ast/visitor.py:771.

  The error is caught and logged at ERROR level (visitor.py:779-786), but execution continues normally with the correctly interpolated values. However, the error message is alarming and makes it appear something has failed when it hasn't.

  Possible fixes:
  - Downgrade the log level from LOGGER.error to LOGGER.debug for this known pattern
  - Use a smarter placeholder than "null" that won't create invalid SQL syntax (e.g., 0 for numeric contexts)
  - Suppress the error entirely when it's a known limitation of static f-string analysis

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
    "marimo": "0.18.4",
    "editable": false,
    "location": "/Users/rob/tmp/marimo_bug/.venv/lib/python3.12/site-packages/marimo",
    "OS": "Darwin",
    "OS Version": "24.6.0",
    "Processor": "arm",
    "Python Version": "3.12.8",
    "Locale": "en_US",
    "Binaries": {
      "Browser": "143.0.7499.192",
      "Node": "v23.6.1"
    },
    "Dependencies": {
      "click": "8.3.1",
      "docutils": "0.22.4",
      "itsdangerous": "2.2.0",
      "jedi": "0.19.2",
      "markdown": "3.10",
      "narwhals": "2.15.0",
      "packaging": "25.0",
      "psutil": "7.2.1",
      "pygments": "2.19.2",
      "pyyaml": "6.0.3",
      "starlette": "0.48.0",
      "tomlkit": "0.13.3",
      "uvicorn": "0.40.0",
      "websockets": "15.0.1"
    },
    "Optional Dependencies": {
      "duckdb": "1.4.3",
      "ibis-framework": "11.0.0",
      "loro": "1.10.3",
      "mcp": "1.25.0",
      "openai": "2.14.0",
      "pandas": "2.3.3",
      "polars": "1.36.1",
      "pyarrow": "22.0.0",
      "sqlglot": "28.5.0"
    },
    "Experimental Flags": {
      "sql_linter": true,
      "rtc_v2": false,
      "chat_sidebar": true,
      "inline_ai_tooltip": true,
      "external_agents": true,
      "lsp": true
    }
  }
```

</details>

### Code to reproduce

```python

# THIS MUST BE INVOKED WITH `python this_script.py` to see the error message
import marimo
app = marimo.App()

@app.cell
def _():
    days_ago = 7
    return (days_ago,)

@app.cell
def _(days_ago):
    import marimo as mo
    df = mo.sql(f"""
        with some_table as (
            select '2025-01-01'::timestamp as timestamp, 'STRING' as message_type
        ),
        cte as (
            select timestamp as ts
            from some_table
            where timestamp > current_date - interval '{days_ago} days'
                and message_type = 'STRING'
        )
        select * from cte
    """)
    return (df,)

if __name__ == "__main__":
    app.run()

```

Run with with `python <script_name>`

VedantMadane:
﻿## Summary

Fix spurious SQL parsing errors when f-string variables are used in interval expressions.

When using f-strings in `mo.sql()` like `interval '{days_ago} days'`, the static analysis would replace the variable with "null", creating `interval 'null days'` which is invalid SQL. This caused alarming ERROR log messages even though the query executed successfully.

## Changes

- Changed the f-string placeholder from "null" to "1" in `normalize_sql_f_string()`
- "1" is valid in more SQL contexts (e.g., `interval '1 days'` parses correctly)
- Updated tests to reflect the new placeholder
- Added regression test for interval expression case

## Test plan

- [x] Added `test_normalize_sql_f_string_with_interval` regression test
- [x] Updated existing tests to expect "1" instead of "null"

Fixes #7717

### DIFF
--- a/tests/_ast/test_visitor.py
+++ b/tests/_ast/test_visitor.py
@@ -1204,6 +1204,7 @@ def test_unbound_local_not_deleted_ref() -> None:
 
 
 HAS_DEPS = DependencyManager.duckdb.has()
+HAS_SQLGLOT = DependencyManager.sqlglot.has()
 
 
 @pytest.mark.skipif(not HAS_DEPS, reason="Requires duckdb")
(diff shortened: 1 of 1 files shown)


mscolnick:
> "1" is valid in more SQL contexts (e.g., interval '1 days' parses correctly)

Is this true? Could we add a parameterized test with various contexts which this gets replaced. im sure we can't get 100%, but would be good to document the coverage for it. 

e..g. 
```
# WHERE clause 
SELECT * FROM table WHERE {var} = 10
SELECT * FROM table WHERE col = {var}
SELECT * FROM table WHERE {col_name} = {value}
SELECT * FROM table WHERE {var} > 100
SELECT * FROM table WHERE {var} BETWEEN 10 AND 20
SELECT * FROM table WHERE {var} IN (1, 2, 3)
SELECT * FROM table WHERE {var} LIKE '%search%'
SELECT * FROM table WHERE {var} IS NULL
SELECT * FROM table WHERE {var} IS NOT NULL

# Table name
SELECT * FROM {table_name} WHERE col = 10
SELECT * FROM {schema}.{table} WHERE col = 10
SELECT * FROM {database}.{schema}.{table} WHERE col = 10

# JOIN clauses:
SELECT * FROM table1 JOIN {table2} ON table1.id = {table2}.id
SELECT * FROM {table1} JOIN {table2} ON {join_col} = {join_col}
SELECT * FROM table1 {join_type} JOIN table2 ON {condition}

# ORDER BY and LIMIT:
SELECT * FROM table ORDER BY {column}
SELECT * FROM table ORDER BY {column} {direction}
SELECT * FROM table LIMIT {limit}
SELECT * FROM table LIMIT {limit} OFFSET {offset}
SELECT * FROM {table} ORDER BY {col} DESC LIMIT {n}

# Misc:
SELECT * FROM table WHERE {col} = {val} OR {col2} = {val2}
SELECT * FROM table WHERE ({condition1}) AND ({condition2})
SELECT * FROM table WHERE {col} IN ({value_list})
SELECT * FROM table WHERE {date_col} > {date_value}
```

VedantMadane:
Added a new pytest.mark.parametrize parameterized coverage test in tests/_ast/test_visitor.py for normalize_sql_f_string() across several common SQL contexts (WHERE comparisons, BETWEEN, IN, LIKE with quoted pattern, LIMIT/OFFSET, VALUES). For the subset that should produce valid SQL after normalization, the test also runs a best effort sqlglot parse (when sqlglot is installed) to document the intended coverage.

mscolnick:
thank you @VedantMadane for the improvement!

Issue #7849, opened 2026-01-15
Add cached outputs to run mode


jpopesculian:
### Description

For long running notebook cells, having the cached outputs through the CachingExtension is useful. For example, if it takes 10 seconds to load a datatable, having the datatable view available as a static element before it loads and becomes interactive is a good user experience. Although I think the changes in https://github.com/marimo-team/marimo/pull/7737 are correct (run mode should not cause side effects where possible), it does remove that cached view when loading. I would like to bring back a read-only version of that cache that just populates the initial session

### Suggested solution

1. Add an extra setting `cache_outputs`. I find the control of cache to be confusing based on `auto_instantiate`
2. Set the caching extension to be `CachingExtension(enabled=cache_outputs, mode="write" if mode == SessionMode.EDIT else "read")`
3. Optionally add some way to run notebooks and write to the cache with the marimo CLI so you could something like `marimo cache notebook.py && marimo run notebook.py`

### Are you willing to submit a PR?

- [x] Yes

jameskermode:
I've been exploring a related pattern with [marimo-precompute](https://github.com/jameskermode/marimo-precompute): precomputing cell results offline and loading them instantly in both marimo run and WASM.

In the current implementation, `with persistent_cache(name="my_calc"):` writes cache files to `public/__marimo_precompute__/` on first run. On subsequent marimo run invocations, the cache hits and the block is skipped, giving the "instant cached outputs" behaviour this issue requests.

For WASM deployment, I additionally bundle a manifest.json and use `await prefetch_all()` to load the cache into Pyodide's VFS, with hash-patching to handle cross-environment bytecode differences.

The `marimo cache notebook.py && marimo run notebook.py` workflow suggested here aligns well with this: the precompute step is essentially `marimo export html --no-sandbox` (which executes all cells and writes the cache) followed by `marimo run`.

Sushit-prog:
Hey! I'd like to work on this.

@jameskermode,  really interesting work with `marimo-precompute`. The 
`marimo export html --no-sandbox` → `marimo run` workflow you describe is 
essentially the `marimo cache notebook.py && marimo run notebook.py` pattern 
the issue suggests, and the fact that it already works externally is strong 
validation that this is the right approach.

I noticed PR #7895 attempted an in-runtime fix but was reverted in #8281. 
My read is that mutating cache behavior inside the session lifecycle is risky
your approach sidesteps that entirely by precomputing *before* the session 
starts, which is much safer.

[...]

PR #7895, opened 2026-01-19
improvement: enable reading session cache in App Mode (configurable)

jpopesculian:
Implements https://github.com/marimo-team/marimo/issues/7849

## Changes

1. Bug fix: `cell_ids` was not initialized on the `SessionView` so cache exports would sometimes be exported out of order. The fix was to emit an `UpdateCellIdsNotification` on notebook creation
2. Add a new `serve_cached_sessions_in_apps` runtime configuration item to use the cache in a read-only manner when running as an application

### DIFF
--- a/marimo/_session/session.py
+++ b/marimo/_session/session.py
@@ -29,6 +29,7 @@ from marimo._runtime.commands import (
 from marimo._session.consumer import SessionConsumer
 from marimo._session.events import SessionEventBus
 from marimo._session.extensions.extensions import (
+    CacheMode,
     CachingExtension,
     HeartbeatExtension,
     LoggingExtension,
@@ -141,12 +142,22 @@ class SessionImpl(Session):
                 redirect_console_to_browser=redirect_console_to_browser,
             )
 
+        if mode == SessionMode.EDIT:
+            cache_enabled = not auto_instantiate
+            cache_mode = CacheMode.READ_WRITE
+        else:
+            cache_enabled = config_manager.get_config()["runtime"].get(
+                "serve_cached_sessions_in_apps", False
+            )
+            cache_mode = CacheMode.READ
+
         extensions = [
             *(extensions or []),
             LoggingExtension(),
             HeartbeatExtension(),
             CachingExtension(
-                enabled=not auto_instantiate and mode == SessionMode.EDIT
+                enabled=cache_enabled,
+                mode=cache_mode,
             ),
             NotificationListenerExtension(
                 kernel_manager=kernel_manager, queue_manager=queue_manager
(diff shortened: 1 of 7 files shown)


frontend/src/components/app-config/user-config-form.tsx
```diff
@@ -1069,6 +1069,37 @@ export const UserConfigForm: React.FC = () => {
...
+                    </FormControl>
+                    <FormMessage />
+                    <IsOverridden
+                      userConfig={config}
+                      name="runtime.cache_outputs"
+                    />
+                  </FormItem>
+
+                  <FormDescription>
+                    Whether to cache cell outputs to disk and restore them on
```
dmadisetti: This is true by default when auto instantiation is off I'm not sure we need the additional distinction

marimo/_runtime/runtime.py
```diff
@@ -2111,6 +2112,12 @@ async def instantiate(self, request: CreateNotebookCommand) -> None:
             LOGGER.info("App is already instantiated, skipping instantiation.")
             return
 
+        cell_ids = [er.cell_id for er in request.execution_requests]
```
dmadisetti: Is this right? Not certain the cellida must be in order. If it is correct can you leave a comment why?

mscolnick:
We don't want the config for edit-mode and app-mode to be the same (there are cases you want one without the other).

How about `runtime.serve_cached_sessions_in_apps`
```python
class RuntimeConfig:
   ...
   serve_cached_sessions_in_apps: NotRequired[bool]
```

this way we 1) don't need to set a default and 2) we can fallback to `False`. this will also result in less of a diff footprint.

also we don't need to show this in the UI if this is just deployment configuration (which I assume is your use-case)

marimo/_runtime/runtime.py
jpopesculian: I've updated this to take an explicit cell_ids for clarity. I think because dict keys are ordered in python as far as my knowledge goes, this would be essentially the same, but it seems somewhat fragile, so better to be explicit

tests/_pyodide/test_pyodide_session.py
```diff
@@ -351,7 +351,7 @@ async def test_pyodide_session_put_input(
         # Notebook operations
         (
             '{"type": "create-notebook", "executionRequests": [{"cellId": "cell-1", "code": "print(1)", "type": "execute-cell"}], '
-            '"setUiElementValueRequest": {"objectIds": [], "values": [], "type": "update-ui-element"}, '
+            '"cellIds": [], "setUiElementValueRequest": {"objectIds": [], "values": [], "type": "update-ui-element"}, '
```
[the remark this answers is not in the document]
jpopesculian: Fixed

marimo/_session/session.py
```diff
@@ -141,12 +142,22 @@ def create(
...
             )
 
+        if mode == SessionMode.EDIT:
+            cache_enabled = not auto_instantiate
+            cache_mode = CacheMode.READ_WRITE
+        else:
+            cache_enabled = config_manager.get_config()["runtime"].get(
+                "serve_cached_sessions_in_apps", False
+            )
+            cache_mode = CacheMode.READ
```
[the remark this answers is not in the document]
jpopesculian: Updated PR description

PR #7905, opened 2026-01-20
ci: label 'dependencies'

mscolnick:


.github/labeler.yml
```diff
@@ -4,6 +4,9 @@ documentation:
 bash-focus:
   - changed-files:
       - any-glob-to-any-file: ['packages/openapi/api.yaml', 'frontend/index.html']
+dependencies:
+  - changed-files:
+      - any-glob-to-any-file: ['.pre-commit-config.yaml', 'pnpm-lock.yaml']
```
Light2Dark, line 9:
```suggestion
      - any-glob-to-any-file: ['.pre-commit-config.yaml', 'pnpm-lock.yaml', 'pyproject.toml']
```

mscolnick: that could be noisy so i didnt want to add that

PR #7906, opened 2026-01-20
improvement: use vega 6 mimetype, capture screenshots on pdf export

mscolnick:
This makes sure we use the correct vega/vegalite mime type (now 6) and well as updates some logic to include also include vega charts when capturing screenshots.

### DIFF
--- a/.github/workflows/test_cli.yaml
+++ b/.github/workflows/test_cli.yaml
@@ -151,13 +151,9 @@ jobs:
       - name: Test CLI
         shell: bash
         run: |
-          export MARIMO_WHEEL_PATH=$(ls $PWD/marimo*.whl)
-          uv run \
-            --group=test-optional \
-            --group=test \
-            --group=dev \
-            --with="${MARIMO_WHEEL_PATH}" \
-            pytest -v tests/_cli/test_cli* --maxfail=2
+          uv venv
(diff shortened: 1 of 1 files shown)


frontend/src/core/export/hooks.ts
```diff
@@ -85,6 +85,15 @@ export function useAutoExport() {
...
 const richCellsToOutputAtom = atom<Record<CellId, unknown>>({});
 
+// MIME types to capture screenshots for
+const MIME_TYPES_TO_CAPTURE_SCREENSHOTS = new Set([
+  "text/html",
+  "application/vnd.vegalite.v5+json",
+  "application/vnd.vega.v5+json",
+  "application/vnd.vegalite.v6+json",
+  "application/vnd.vega.v6+json",
+]);
```
mscolnick, line 95: cc @Light2Dark for export-to-pdf

manzt (dismissed):
LGTM. Are the vega specs kept in the repr mimebundle for the ipynb exports? I know that is separate from these changes, but some renders will support that directly on top of PNG.

mscolnick:
@manzt yes, we augment the outputs with the PNG (instead of overwrite)

manzt:
Great.

Light2Dark (dismissed):

PR #7910, opened 2026-01-20
internal: add marimo._internal exports

mscolnick:
This exports functions/classes that are meant to be re-used in marimo-lsp under a `marimo._internal` which will help signify that this is being re-used.


tests/_internal/snapshots/internal_api.txt
```diff
@@ -1,3 +1,135 @@
...
+  StartupLogsNotification
+  UIElementMessageNotification
+  UpdateCellCodesNotification
+  UpdateCellIdsNotification
+  ValidateSQLResultNotification
+  VariableDeclarationNotification
+  VariableValue
+  VariableValuesNotification
+  VariablesNotification
+packages
```
dmadisetti, line 102: `construct_uv_flags`, `PyProjectReader`
Are also used by quarto, but are really more utility functions

mscolnick: thanks for flagging. im not sure those are ones i want to export right now. i want to look if there is a higher level thing to export

PR #7921, opened 2026-01-21
fix: fail CLI edit if shared memory is not available

akshayka:
marimo edit today requires shared memory; this change updates the CLI to exit with an informative error if we detect that shared memory is not available

This fixes an issue in which running marimo on minimal sandboxes that don't have /dev/shm led to a confusing error state, with the frontend repeatedly failing to connect.

### DIFF
--- a/tests/_utils/test_platform.py
+++ b/tests/_utils/test_platform.py
@@ -37,7 +37,9 @@ class TestIsPyodide:
         assert not is_pyodide()
 
     def test_is_pyodide_when_loaded(self) -> None:
-        with patch("marimo._utils.platform.sys.modules", {"pyodide": object()}):
+        with patch(
+            "marimo._utils.platform.sys.modules", {"pyodide": object()}
+        ):
             assert is_pyodide()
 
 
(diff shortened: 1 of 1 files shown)


marimo/_cli/cli.py
```diff
@@ -525,6 +525,23 @@ def edit(
...
+    shm_available, shm_error = check_shared_memory_available()
+    if not shm_available:
+        _loggers.marimo_logger().error(
+            f"marimo failed to start: marimo edit requires shared memory support for multiprocessing.\n\n"
+            f"{shm_error}\n\n"
+            "Possible solutions:\n"
+            "  - If running in Docker, ensure /dev/shm is mounted with sufficient size\n"
+            "    (e.g., --shm-size=256m or -v /dev/shm:/dev/shm)\n"
+            "  - If /dev/shm is full, clear unused shared memory segments\n"
+            "  - Use 'marimo run' instead if you only need to view notebooks"
```
mscolnick, line 541: another option is to warn and then disable `virtualfiles` (we do support disabling virtual files and it will inline data)

akshayka: Oh interesting. I think this is fine for now, but if someone opens an issue we can consider

akshayka:
Failing test looks unrelated (this PR does not have API changes)

PR #7940, opened 2026-01-22
improvement: downgrade dataframe in non-interactive environments

mscolnick:
This prints plain HTML dataframes when we are exporting to non-interactive environments (e.g. ipynb, pdf)

### DIFF
--- a/marimo/_plugins/ui/_impl/dataframes/dataframe.py
+++ b/marimo/_plugins/ui/_impl/dataframes/dataframe.py
@@ -229,7 +229,7 @@ class dataframe(UIElement[dict[str, Any], DataFrameType]):
             ),
         )
 
-    # Override _mime_ to return an a plain HTML representation in non-interactive environments
+    # Override _mime_ to return a plain HTML representation in non-interactive environments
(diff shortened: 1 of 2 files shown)


marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -226,6 +229,17 @@ def __init__(
...
         )
 
+    # Override _mime_ to return a plain HTML representation in non-interactive environments
+    def _mime_(self) -> tuple[KnownMimeType, str]:
+        if is_non_interactive():
+            # Generates a plain HTML representation of the table data,
+            # useful for rendering in the GitHub viewer.
+            repr_html = getattr(self._data, "_repr_html_", None)
+            if repr_html is not None and callable(repr_html):
+                return ("text/html", cast(str, repr_html()))
```
Light2Dark, line 239: should we move this logic to `hypertext.py`?
the logic is also repeated in `table.py`

mscolnick: i think the duplication is fine since, its about 2 lines and they are both tested

PR #7954, opened 2026-01-23
fix: --sandbox flag should add script metadata when none exists

akshayka:
When using `marimo edit --sandbox notebook.py` on a file without PEP 723 script metadata, the marimo dependency and Python version were not being added. This caused `uv run notebook.py` to fail because marimo wasn't declared as a dependency.

Now `_ensure_marimo_in_script_metadata()` will:

- Create the script metadata block if it doesn't exist
- Add `requires-python` with the current Python version
- Add marimo as a dependency

Also ensures `requires-python` is added when marimo is already present but the Python version is missing.

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -360,10 +360,10 @@ def _ensure_python_version_in_script_metadata(name: str) -> None:
 
 
 def _ensure_marimo_in_script_metadata(name: str | None) -> None:
-    """Ensure marimo and python version are in the script metadata.
+    """Ensure marimo is in the script metadata.
 
     If the file has no PEP 723 script metadata or marimo is not listed
-    as a dependency, add marimo and the Python version.
+    as a dependency, add marimo using uv.
     """
     # Only applicable to `.py` files.
     if name is None or not name.endswith(".py"):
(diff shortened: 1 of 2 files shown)


marimo/_cli/sandbox.py
```diff
@@ -320,25 +320,72 @@ def construct_uv_command(
...
+        # No script metadata exists
+        return
+
+    if "requires-python" in project:
+        # Already has Python version
+        return
+
+    python_version = (
+        f">={platform.python_version_tuple()[0]}"
+        f".{platform.python_version_tuple()[1]}"
```
mscolnick: nit: put this out `platform.python_version_tuple()`

marimo/_session/notebook/file_manager.py
```diff
@@ -191,10 +191,19 @@ def _save_file(
             from marimo._config.settings import GLOBAL_SETTINGS
 
             if GLOBAL_SETTINGS.MANAGE_SCRIPT_METADATA:
+                import platform
+
                 from marimo._utils.scripts import write_pyproject_to_script
 
+                python_version = (
+                    f">={platform.python_version_tuple()[0]}"
+                    f".{platform.python_version_tuple()[1]}"
```
mscolnick: maybe DRY up with the other: `def python_version_string(prefix=">=")`

PR #7980, opened 2026-01-26
Avoid ValueError when comparing embed defs with NumPy arrays

### LINKED ISSUE #7969: ValueError with numpy arrays in embed defs
### Describe the bug

When the `defs` given to an embed are numpy arrays, a `ValueError` may pop up when given an array with different shape than before. This also hints another bug where a new def value may not propagate.

* Edit/run the provided example `embed_np_a.py`
* Change selection in the dropdown
* Expected: updated result
* Actual: exception -

```
ValueError
The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
See the console area for a traceback.
```

This probably happens because numpy overridden `==` to compare elements in arrays rather than the whole array. Marimo should probably do some kind of actual comparison which results with a boolean rather than `==` which is overridden by popular libraries.

### Will you submit a PR?

- [x] Yes

### Environment

<details>

```
{
  "marimo": "0.19.6",
  "editable": true,
  "location": "/Users/yairchu/dev/marimo/marimo",
  "OS": "Darwin",
  "OS Version": "25.2.0",
  "Processor": "arm",
  "Python Version": "3.13.1",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "140.0.7339.134",
    "Node": "v22.21.1"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.7.0",
    "packaging": "25.0",
    "psutil": "7.1.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "duckdb": "1.4.1",
    "loro": "1.8.1",
    "openai": "2.15.0",
    "pyarrow": "21.0.0",
    "ruff": "0.14.0",
    "sqlglot": "27.26.0"
  },
  "Experimental Flags": {
    "chat_modes": true,
    "cache_panel": true,
    "multi_column": true
  }
}
```

</details>

### Code to reproduce

As this is an embed issue, two short notebooks are provided, the external `embed_np_a.py`:

```Python
import marimo

__generated_with = "0.19.6"


@app.cell
def _():
    import marimo as mo
    import numpy as np
    return mo, np


@app.cell
def _():
    import embed_np_b
    return (embed_np_b,)


@app.cell
def _(mo, np):
    dropdown = mo.ui.dropdown(
        {"ones(1)": np.ones(1), "zeros(2)": np.zeros(2)}, value="ones(1)"
    )
    dropdown
    return (dropdown,)


@app.cell
async def _(dropdown, embed_np_b):
    embed = await embed_np_b.app.embed(defs={"arr": dropdown.value})
    embed.output
    return


if __name__ == "__main__":
    app.run()
```

And the internal `embed_np_b.py`:

```Python
import marimo

__generated_with = "0.19.6"


@app.cell
def _():
    import marimo as mo
    import numpy as np
    return mo, np


@app.cell
def _(np):
    arr = np.array([[1, 2, 3]])
    return (arr,)


@app.cell
def _(arr, mo):
    mo.md(f"`{arr.shape = }`")
    return


if __name__ == "__main__":
    app.run()
```

MukeshK17:
### Summary
Fixes a bug where `App.embed()` could raise a `ValueError` when embed `defs` contain NumPy arrays.  
The issue occurred during cache checks when comparing `defs` using `==`, which is ambiguous for NumPy arrays.

---

### Problem
Changing embed `defs` containing NumPy arrays (e.g. arrays with different shapes) could raise:

`ValueError: The truth value of an array with more than one element is ambiguous`

This originated in `AppKernelRunner.are_outputs_cached`.

---

### Solution
- Added a safe helper (`_defs_equal`) for comparing embed `defs`
- Uses `numpy.array_equal` for NumPy arrays
- Falls back safely for ambiguous comparisons
- Updated cache logic to use this helper

---

### Tests
Added regression tests ensuring:
- Different NumPy arrays invalidate the cache
- Equal NumPy arrays reuse the cache

These tests fail on `main` and pass with this change.

---

Fixes #7969

### CLA

### DIFF
--- a/marimo/_runtime/app/kernel_runner.py
+++ b/marimo/_runtime/app/kernel_runner.py
@@ -45,19 +45,16 @@ def _defs_equal(a: dict[str, Any] | None, b: dict[str, Any] | None) -> bool:
 
         try:
             if DependencyManager.numpy.imported():
-                import numpy as _np
-            else:
-                _np = None
-            if (
-                _np is not None
-                and isinstance(va, _np.ndarray)
-                and isinstance(vb, _np.ndarray)
-            ):
-                if not _np.array_equal(va, vb):
-                    return False
-            else:
-                if va != vb:
-                    return False
+                import numpy as np
+
+                if isinstance(va, np.ndarray) and isinstance(vb, np.ndarray):
+                    if not np.array_equal(va, vb):
+                        return False
+                    continue
+
+            if va != vb:
+                return False
+
         except Exception:
             # Any ambiguous or unsafe comparison => treat as changed
             return False


MukeshK17:

tests/_runtime/runner/test_kernel_runner.py
```diff
@@ -0,0 +1,44 @@
...
+    runner = AppKernelRunner.__new__(AppKernelRunner)
+    runner._previously_seen_defs = None
+    runner._outputs = {cast(CellId_t, "cell-1"): "dummy"}
+    return runner
+
+
+def test_numpy_defs_do_not_crash_and_invalidate_cache():
+    runner = _make_runner()
+
+    import numpy as np
```
mscolnick, line 21: we will need to skip these tests if numpy is not installed. 

you can do `@pytest.mark.requires("numpy")` on the test function

marimo/_runtime/app/kernel_runner.py
```diff
@@ -25,6 +25,47 @@
...
+    for key in a:
+        va, vb = a[key], b[key]
+
+        # identical object
+        if va is vb:
+            continue
+
+        try:
+            # numpy-safe comparison
+            try:
```
mscolnick: can you do `if DependencyManager.numpy.imported()` before this so we don't import numpy if it hasnt been imported (otherwise this could be a slower operation

tests/_runtime/runner/test_kernel_runner.py
MukeshK17: Thanks for the feedback! I’ve updated the code as requested.

marimo/_runtime/app/kernel_runner.py
MukeshK17: I’ve updated the implementation to use DependencyManager.numpy as suggested.

mscolnick:
@MukeshK17 there is a small typecheck error

`marimo/_runtime/app/kernel_runner.py:50: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]`

MukeshK17:
Fixed the reported typecheck issue.

PR #7997, opened 2026-01-27
feat: export notebooks as PDFs via CLI

### LINKED ISSUE #896: Export Marimo notebook to PDF
### Description

Hello,

I think PDF export is missing

Kind regards

### Suggested solution

Following #892 a PDF export feature will be great.
It will be even greater if it is accessible both through CLI and from UI
Embedding code into PDF to me able to regenerate notebook (into https://marimo.app/ or in our own Marimo install) could be a nice feature to also have.

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #3415: pdf output with code
### Description

I am hoping to be able to make pdf output with code, while using `mo.mpl.interactive` to display plots. Currently when I use the menu to do pdf output, it tells me to press `Crtil+.` to switch to app mode before making the pdf output. All my plots output fine, but the code is hidden.

### Suggested solution

I can think of 2
1. allow pdf output from notebook mode (I know there were correctness issues, so those would need to be fixed).
2. allow app mode to show code, just not editable. (Is this already possible?)

### Alternative

_No response_

### Additional context

My use case is that I want a hard copy of "I did this analysis this way and got this result" that is immutable. I prefer pdf to html, as some features don't work in html output. Eg the "dataframe viewer" doesn't seem to work in html output since it no longer has access to the dataframe. I just want a copy of what was on the screen (and would have been if I scrolled) when I said to export a copy, and pdf seems best matched to that.

peter-gy:
## 📝 Summary

Adds `marimo export pdf [OPTIONS] NAME [ARGS]...` to CLI with options for output inclusion, webpdf / latex rendering mode, file watching and sandboxed notebook execution.

Closes #896, closes #3415.

## 🔍 Description of Changes

-  Reused the existing `nbformat` & `nbconvert` PDF export pipeline.
- Added support for `--include-outputs/--no-include-outputs`, `--webpdf`, `--watch`, and `--sandbox` for DX
- Introduced `watch_and_export_bytes` (kept separate from `watch_and_export` to avoid complicating the existing text export helper with bytes/stdout/encoding concerns).
- Added a thin smoke test (`test_export_pdf_missing_dependencies`) to ensure missing deps produce a helpful error.
- Purposefully no e2e PDF generation test here to avoid heavy system deps (Chromium / TeX / Pandoc)
  - plan is to cover that in `marimo-team/marimo-integration-ci`

## 📋 Checklist

### DIFF
--- a/tests/_cli/test_cli_export.py
+++ b/tests/_cli/test_cli_export.py
@@ -842,3 +842,21 @@ class TestExportIpynb:
     def test_cli_export_html_sandbox_no_prompt(temp_marimo_file: str) -> None:
         p = _run_export("html", temp_marimo_file, "--no-sandbox")
         _assert_success(p)
+
+
+class TestExportPDF:
+    @pytest.mark.skipif(
+        DependencyManager.nbformat.has() and DependencyManager.nbconvert.has(),
+        reason="This test expects PDF export deps to be missing.",
+    )
+    def test_export_pdf_missing_dependencies(
+        self, temp_marimo_file: str
+    ) -> None:
+        output_file = temp_marimo_file.replace(".py", ".pdf")
+        p = _run_export(
+            "pdf", temp_marimo_file, "--output", output_file, "--no-sandbox"
+        )
+        _assert_failure(p)
+        stderr = p.stderr.decode()
+        assert "nbconvert" in stderr
+        assert "pip install" in stderr


marimo/_cli/export/commands.py
```diff
@@ -106,6 +108,92 @@ async def start() -> None:
     asyncio_run(start())
 
 
+def watch_and_export_bytes(
```
mscolnick, line 111: we might be able to kill `watch_and_export` and consolidate to one function (and use `bytes`. it looks like nothing may use `ExportResult.download_filename`

we can do this in a followup though if you prefer

peter-gy: Happy to do it in a followup.

PR #7998, opened 2026-01-27
fix: patch html-to-image to avoid duplicate iframe in screenshot

mscolnick:
patch html-to-image to avoid duplicate iframe in screenshot


Light2Dark:
maybe could lock the dep as well, but I don't think it will be updated

PR #8000, opened 2026-01-27
refactor: consolidate `watch_and_export` implementations

peter-gy:
## 📝 Summary

Consolidate `watch_and_export_bytes` and `watch_and_export` implementations & represent `ExportResult` contents as bytes while still allowing string-based access.

Follow-up of #7997

## 🔍 Description of Changes

- Main refactor is making `ExportResult` represent results as bytes instead of strings to support cases like PDF export
- Implemented "contents is bytes, text is string" convention familiar from `httpx`, `requests`, etc.
- killed `watch_and_export_bytes` instead of `watch_and_export` so that `export_callback` can keep returning `ExportResult` objects to (1) avoid duplicating encode / decode logic, (2) have a place for cross-cutting helpers like `.text` / `.from_text` and (3) keep the door open for additional metadata like e.g. warnings or execution stats

## 📋 Checklist

### DIFF
--- a/marimo/_server/export/__init__.py
+++ b/marimo/_server/export/__init__.py
@@ -52,16 +52,6 @@ class ExportResult:
     download_filename: str
     did_error: bool
 
-    @classmethod
-    def from_text(
-        cls, *, text: str, download_filename: str, did_error: bool
-    ) -> ExportResult:
-        return cls(
-            contents=text,
-            download_filename=download_filename,
-            did_error=did_error,
-        )
-
     @cached_property
     def bytez(self) -> bytes:
         """Return UTF-8 encoded bytes (cached)."""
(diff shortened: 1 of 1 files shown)


peter-gy:
@mscolnick I deviated from https://github.com/marimo-team/marimo/pull/7997#discussion_r2732763357 in this refactor and kept `ExportResult` instead of using plain `bytes` as otherwise we would find ourselves in `.encode` / `.decode` duplication hell + no easy way to associate further metadata with results down the line. what do you think?

marimo/_server/export/__init__.py
```diff
@@ -47,10 +47,25 @@
 
 @dataclass
 class ExportResult:
-    contents: str
+    contents: bytes
```
mscolnick: there might be a small inefficiency over converting from text to bytes and back. maybe we could do:

`contents: bytes | str`

with a `@property` (or `@cached_property`) for `bytes` and `text` that will check the `contents` type and convert or not convert

peter-gy: Great point.

To make sure we don't shadow the built-in `bytes` via a `def bytes(self) ...`: how about `contents: bytes | str` and either `as_bytes()` / `as_text()` or via `cached_property` `data_bytes` / `data_text` ?

or just `bytez`? (:

mscolnick: yes, great call

PR #8006, opened 2026-01-27
poc: modern-screenshot comparison

mscolnick:


mscolnick:
leaving this here in case you wanted to tweak params (cc @Light2Dark)

PR #8008, opened 2026-01-27
improvement: add progress bar to PDF download

mscolnick:
Add a progress bad to the PDF export.

When we are snapshotting images, it will increment. Otherwise it will just show a loading progress bar.


Light2Dark:
going to merge, to test. looks good here.

PR #8011, opened 2026-01-28
improvement: lsp status icon and refresh

mscolnick:
Expose the status icon and allow users to click to refresh if there are any issues.

Screenshot 2026-01-27 at 7 40 36 PM

### DIFF
--- a/marimo/_server/models/lsp.py
+++ b/marimo/_server/models/lsp.py
@@ -6,7 +6,7 @@ from typing import Literal, NewType, Optional
 import msgspec
 
 # Type-safe server identifier
-ServerId = NewType("ServerId", str)
+LspServerId = NewType("LspServerId", str)
 
 # Status enum for LSP server health
 LspServerStatus = Literal[
@@ -29,7 +29,7 @@ class LspServerHealth(msgspec.Struct, rename="camel"):
     - unresponsive: process alive but not responding to pings
     """
 
-    server_id: ServerId
+    server_id: LspServerId
     status: LspServerStatus
     port: int
     last_ping_ms: Optional[float] = None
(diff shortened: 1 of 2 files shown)


marimo/_server/models/lsp.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
+from typing import Literal, Optional
+
+import msgspec
+
+
+class LspServerHealth(msgspec.Struct, rename="camel"):
```
manzt: These booleans (`is_running`, `is_responsive`, `has_failed`) have some overlap. VS Code's LSP client uses a `status` enum, which we could similarly:

```python
class LspServerHealth(msgspec.Struct, rename="camel"):
    """Health status for a single LSP server.

    Status meanings:
    - starting: process launched, initializing
    - running: healthy and responsive to pings
    - stopped: not running (never started or cleanly stopped)
    - crashed: exited with non-zero code
    - unresponsive: process alive but not responding to pings
    """

    server_id: str
    status: Literal["starting", "running", "stopped", "crashed", "unresponsive"]
    port: int
    last_ping_ms: Optional[float] = None
    error: Optional[str] = None  # Why it crashed/stopped
```


We could also consider `started_at`:

```py
    started_at: Optional[float] = None  # Unix timestamp
```

which gives us uptime for free (frontend can show "running for 5m" or detect restart loops).

marimo/_server/models/lsp.py
```diff
@@ -0,0 +1,39 @@
...
+    """Aggregated health response for all LSP servers."""
+
+    status: Literal["healthy", "degraded", "unhealthy"]
+    servers: list[LspServerHealth]
+
+
+class LspRestartRequest(msgspec.Struct, rename="camel"):
+    """Request to restart LSP servers."""
+
+    server_ids: Optional[list[str]] = None  # None = restart failed servers
```
manzt: Is it worth making a `ServerId` `NewType` or at least a `TypeAlias` scoped to this module?

```py
ServerId = typing.NewType("ServerId", str)
```

msgspec handles these fine.

PR #8053, opened 2026-01-29
feat(sql): respect display.dataframes config setting

chrisflath:
## Summary

When `display.dataframes` is set to `"plain"`, `mo.sql()` now uses `plain()` to format output instead of the rich interactive table viewer.

This makes SQL cell output consistent with regular DataFrame display behavior when the user has configured plain table output.

## Changes

- Added check for `include_opinionated()` in `marimo/_sql/sql.py` before creating table output
- When `include_opinionated()` returns `False` (plain mode), uses `plain(df)` instead of `table.table()`

## Testing

Added tests in `tests/_sql/test_sql.py`:

- `test_sql_plain_output_when_not_opinionated`: Verifies that `Plain` object is used when `include_opinionated()` returns `False`
- `test_sql_rich_output_when_opinionated`: Verifies that `table` is used when `include_opinionated()` returns `True`

Both tests mock `include_opinionated()` and verify the correct output type is passed to `replace()`.

### DIFF
--- a/tests/_sql/test_sql.py
+++ b/tests/_sql/test_sql.py
@@ -634,39 +634,30 @@ Expected 'FROM' keyword"""
 class TestDisplayConfigBehavior:
     """Test that sql() respects the display.dataframes config setting."""
 
+    @patch("marimo._sql.sql.replace")
+    @patch("marimo._output.formatters.df_formatters.include_opinionated")
     @pytest.mark.skipif(
         not HAS_POLARS or not HAS_DUCKDB, reason="polars and duckdb required"
     )
-    @patch("marimo._sql.sql.replace")
-    def test_sql_plain_output_when_not_opinionated(self, mock_replace):
+    def test_sql_plain_output_when_not_opinionated(
+        self, mock_include_opinionated, mock_replace
+    ):
         """Test that SQL uses plain() when include_opinionated returns False."""
-        import duckdb
+        mock_include_opinionated.return_value = False
         import polars as pl
 
         from marimo._output.formatting import Plain
 
-        # Create a test table
-        duckdb.sql(
-            "CREATE OR REPLACE TABLE test_plain AS SELECT * FROM range(5)"
-        )
+        # Test query
+        result = sql("SELECT 1")
+        assert isinstance(result, pl.DataFrame)
 
-        with patch(
-            "marimo._output.formatters.df_formatters.include_opinionated",
-            return_value=False,
-        ):
-            # Test query
-            result = sql("SELECT * FROM test_plain")
-            assert isinstance(result, pl.DataFrame)
+        # Should call replace with Plain object
+        mock_replace.assert_called_once()
+        call_args = mock_replace.call_args[0][0]
 
-            # Should call replace with Plain object
-            mock_replace.assert_called_once()
-            call_args = mock_replace.call_args[0][0]
-
-            # The call should be a Plain object (not a table)
-            assert isinstance(call_args, Plain)
-
-        # Clean up
-        duckdb.sql("DROP TABLE test_plain")
+        # The call should be a Plain object (not a table)
+        assert isinstance(call_args, Plain)
 
     @pytest.mark.skipif(
         not HAS_POLARS or not HAS_DUCKDB, reason="polars and duckdb required"


marimo/_sql/sql.py
```diff
@@ -127,13 +127,29 @@ def sql(
...
             t = plain_text(text_output)
+        elif not include_opinionated():
+            # Respect display.dataframes config - show plain HTML when set to "plain"
+            from marimo._output.hypertext import Html
+
+            display_df = df
+            if SQLAlchemyEngine.is_cursor_result(df):
+                display_df = SQLAlchemyEngine.get_cursor_metadata(df)
+            elif DBAPIEngine.is_dbapi_cursor(df):
+                display_df = DBAPIEngine.get_cursor_metadata(df)
```
mscolnick: there is some duplicate code from below:

```
            if SQLAlchemyEngine.is_cursor_result(df):
                display_df = SQLAlchemyEngine.get_cursor_metadata(df)
            elif DBAPIEngine.is_dbapi_cursor(df):
                display_df = DBAPIEngine.get_cursor_metadata(df)
```

you might be able to do to `t = plain(display_df)` (plain will run back through our formatters, but skip making them pretty dataframes)

chrisflath:

marimo/_sql/sql.py
chrisflath: Good suggestion! A minimal elif instead of the long manual handling. The pull request has been updated with a simplified implementation.

marimo/_sql/sql.py
```diff
@@ -127,13 +127,19 @@ def sql(
...
         from marimo._plugins.stateless.plain_text import plain_text
         from marimo._plugins.ui._impl import table
 
         if isinstance(sql_engine, DuckDBEngine) and is_explain_query(query):
             # For EXPLAIN queries in DuckDB, display plain output to preserve box drawings
             text_output = extract_explain_content(df)
             t = plain_text(text_output)
+        elif not include_opinionated():
+            # Respect display.dataframes config - use plain formatting
+            from marimo._output.formatting import plain
```
Light2Dark: I think it'd be better to move this import to below `if output` for consistency. In general I'd prefer moving all imports to top of the file, but for lazy loading and consistency, could move it below output imo

Light2Dark (commented):
Would you be able to add some tests? Probably under `test_sql.py`, may need to mock different return values of `include_opinionated()`

marimo/_sql/sql.py
chrisflath: sure

chrisflath:
> Would you be able to add some tests? Probably under `test_sql.py`, may need to mock different return values of `include_opinionated()`

Forcepushed tests.

tests/_sql/test_sql.py
```diff
@@ -629,3 +629,75 @@ def test_strip_explain_from_error_message(self):
...
+    )
+    @pytest.mark.skipif(
+        not HAS_POLARS or not HAS_DUCKDB, reason="polars and duckdb required"
+    )
+    def test_sql_plain_output_when_not_opinionated(self, mock_replace):
+        """Test that SQL uses plain() when include_opinionated returns False."""
+        import duckdb
+        import polars as pl
+
+        from marimo._output.formatting import Plain
```
Light2Dark, line 649: better to move this import to the top of file. It's in a test file so it's fine.

tests/_sql/test_sql.py
```diff
@@ -629,3 +629,75 @@ def test_strip_explain_from_error_message(self):
...
+
+        # Create a test table
+        duckdb.sql(
```
Light2Dark: Is it possible to just do `sql("SELECT 1")`. We don't need to create & drop a new table for this.

tests/_sql/test_sql.py
```diff
@@ -629,3 +629,75 @@ def test_strip_explain_from_error_message(self):
                       ^
 Expected 'FROM' keyword"""
         assert strip_explain_from_error_message(multiline_error) == expected
+
+
+class TestDisplayConfigBehavior:
+    """Test that sql() respects the display.dataframes config setting."""
+
+    @pytest.mark.usefixtures("_mock_include_opinionated")
```
Light2Dark: Hmm, this test seems to fail for me. The fixture cannot be found (not sure if/how CI passes)
You can patch it in the code itself for simplicity

```python
    @patch("marimo._sql.sql.replace")
    @patch("marimo._output.formatters.df_formatters.include_opinionated")
    @pytest.mark.skipif(
        not HAS_POLARS or not HAS_DUCKDB, reason="polars and duckdb required"
    )
    def test_sql_plain_output_when_not_opinionated(
        self, mock_include_opinionated, mock_replace
    ):
        """Test that SQL uses plain() when include_opinionated returns False."""
        mock_include_opinionated.return_value = False
        import polars as pl

        from marimo._output.formatting import Plain

        # Test query
        result = sql("SELECT 1")
        assert isinstance(result, pl.DataFrame)

        # Should call replace with Plain object
        mock_replace.assert_called_once()
        call_args = mock_replace.call_args[0][0]

        # The call should be a Plain object (not a table)
        assert isinstance(call_args, Plain)
```

chrisflath: Was going in circles in trying to fix this (and frankly with every round I think was having a harder time) - so maybe you can glance over my last try.

tests/_sql/test_sql.py
```diff
@@ -629,3 +629,64 @@ def test_strip_explain_from_error_message(self):
...
+
+            # Should call replace with table (not Plain)
+            mock_replace.assert_called_once()
+            call_args = mock_replace.call_args[0][0]
+
+            # The call should be a table object
+            assert isinstance(call_args, ui.table)
+
+        # Clean up
+        duckdb.sql("DROP TABLE test_rich")
```
Light2Dark, line 692: this one too, we don't need to create a table & cleanup after. just `select 1` should suffice.

tests/_sql/test_sql.py
Light2Dark: maybe this was missed ^ @chrisflath . If you face circular import errors, then its fine to leave it here

Light2Dark (commented):
thanks @chrisflath! it looks close to ready, just test fixes.

PR #8059, opened 2026-01-29
feat: add art-openpipe fallback package

dmadisetti:
## 📝 Summary

`ART` is a RL tuning package provided by OpenPipe https://github.com/OpenPipe/ART
`art` is an ascii package with a quarter as many Github stars: https://pypi.org/project/art/

Falling back to `ART` for `import art` is much more likely for our users

cc @danielbolus


danielbolus:
@akshayka yeah, our bad on the naming front but I like the base molab container if that's possible and not too much of a lift/tech bloat for you guys.

PR #8062, opened 2026-01-30
Handle marimo tex components when exporting to PDF

Light2Dark:
## 📝 Summary

Fix LaTeX rendering in ipynb/PDF exports by converting marimo's internal `<marimo-tex>` HTML format back to standard `$...$ / $$...$$` delimiters.

`\[...\]` → `$$...$$` conversion is also implemented but disabled. However, it seems nbconvert/jupyter doesn't handle them https://github.com/jupyter/nbconvert/issues/477.

Example cell-output
```
[{'output_type': 'display_data', 'metadata': {}, 
  'data': {'text/markdown': '<span class="markdown prose dark:prose-invert contents">
  <span class="paragraph">The exponential function 
  <marimo-tex class="arithmatex">||(f(x) = e^x||)</marimo-tex> can be represented as</span></span>'}
}]
```

## 🔍 Description of Changes

We wrap LaTeX in custom <marimo-tex> tags with ||(...) delimiters (via pymdownx.arithmatex). The frontend renders these with KaTeX, but nbconvert doesn't understand them, causing broken math in exports.

### Not included
- mhchem support (\ce{H2O}) - I think would require custom nbconvert LaTeX template. https://stackoverflow.com/questions/30203791/converting-ipython-notebook-mhchem-markdown-to-pdf

## 📋 Checklist

### DIFF
--- a/tests/_sql/test_sql_parse.py
+++ b/tests/_sql/test_sql_parse.py
@@ -185,7 +185,7 @@ class TestDuckDBInvalidQueries:
             ("SELECT (", ["syntax", "error"]),
             ("SELECT * FROM", ["syntax", "error"]),
             ("SELECT ,", ["syntax", "error"]),
-            ("SELECT * FROM table", ["syntax", "error"]),
+            ("SELEC * FROM table", ["syntax", "error"]),
             ("SELECT * FROM table WHERE", ["syntax", "error"]),
             ("SELECT * FROM table ORDER", ["syntax", "error"]),
             ("INSERT INTO", ["syntax", "error"]),
--- a/marimo/_save/hash.py
+++ b/marimo/_save/hash.py
@@ -78,7 +78,7 @@ def hash_module(
 
     def process(code_obj: CodeType) -> None:
         # Recursively hash the constants that are also code objects
-        for const in code_obj.co_constants:
+        for const in code_obj.co_consts:
             if isinstance(const, types.CodeType):
                 process(const)
             else:


marimo/_smoke_tests/latex/latex.py
```diff
@@ -2,7 +2,7 @@
 
 import marimo
 
-__generated_with = "0.15.5"
+__generated_with = "0.19.6"
```
dmadisetti, line 5: For a sanity check, can you convert this to our markdown format too?

dmadisetti:
Yeah, I see a natural ask being to include `.sty` and `.cls` files for things like mhchem support and other libraries. But maybe we can cross that bridge when we get there

Light2Dark:
> Nice. But we're not converting `\[` notation? A little confused, if jupyter doesn't support it, then we should convert right?

They don't support rendering this syntax afaik. And nbconvert doesn't support parsing this syntax either hence raises an error. So, we are altering the code we send to nbconvert to make it work. From my understanding, `\[` is basically escaping a bracket and hence treated like normal markdown. This issue https://github.com/jupyter/nbconvert/issues/477 and consequent PR attempts to add more parsing for nbconvert but was rejected.

image

I don't really know what the state is now, maybe use jupyter with this extension? https://github.com/jfbercher/jupyter_latex_envs

Light2Dark:
@dmadisetti I think I'm okay to support it. My worry was false positives but the regex looks right. Also seems like it's the recommended way to write LaTeX now.

tests/_sql/test_sql_parse.py
```diff
@@ -185,7 +185,7 @@ class TestDuckDBInvalidQueries:
             ("SELECT (", ["syntax", "error"]),
             ("SELECT * FROM", ["syntax", "error"]),
             ("SELECT ,", ["syntax", "error"]),
-            ("SELEC * FROM table", ["syntax", "error"]),
```
mscolnick, line 188: i assume this syntax error was intentional. but typo is re-writing it? we can create a different syntax error

Light2Dark: addressed in the pre-commit PR, but yeah, that's also possible lol

PR #8077, opened 2026-01-31
tests: upgrade pytest plugins, upperbound sqlglot

mscolnick:
* upgrade pytest plugins
* upperbound sqlglot since it breaks some tests with ibis now (https://github.com/tobymao/sqlglot/issues/6908)

### DIFF
--- a/tests/_runtime/test_pytest_runtime.py
+++ b/tests/_runtime/test_pytest_runtime.py
@@ -1,8 +1,14 @@
+from __future__ import annotations
+
 import asyncio
 import os
+import sys
 from pathlib import Path
 
+import pytest
 
+
+@pytest.mark.skipif(sys.platform == "win32", reason="Fails on Windows CI")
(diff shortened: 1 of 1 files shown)


marimo/_ast/pytest.py
```diff
@@ -224,23 +224,43 @@ def _make_hook(
...
-        res = run()
-        if isinstance(res, Awaitable):
-            import asyncio
+    if is_async:
 
-            loop = asyncio.new_event_loop()
-            _, cell_defs = loop.run_until_complete(res)
-        else:
-            _, cell_defs = res
+        async def _async_hook(*args: Any, **kwargs: Any) -> Any:
```
dmadisetti, line 233: I can follow up with some unit tests. I think this can also be DRY'd too

mscolnick: there is one more async break i am working on

PR #8078, opened 2026-01-31
fix: enable weave tracing on chat widget

dmadisetti:
## 📝 Summary

Currently weave tracing fails on the following code:

```python
chat_client = OpenAI()

@weave.op()
def handle_chat(messages, config):
    """Wrapper that ensures string return for marimo."""
    user_msg = messages[-1].content
    chat_messages = [{"role": "user", "content": user_msg}]

    response = chat_client.chat.completions.create(
        model="gpt-4o-mini",
        messages=[get_system_prompt()] + chat_messages,
        tools=get_robot_tools(),
    )

    msg = response.choices[0].message
    return msg.content

chat = mo.ui.chat(
    handle_chat
)
chat
```

This is because `weave.op` returns a `BoxedStr(str)` which the causes pydantic to fail. As such, we force a conversion to string even if `isinstance(chunk, str)`.

### DIFF
--- a/marimo/_dependencies/dependencies.py
+++ b/marimo/_dependencies/dependencies.py
@@ -252,6 +252,7 @@ class DependencyManager:
     )
     pydantic = Dependency("pydantic")
     zmq = Dependency("zmq")  # pyzmq for sandbox IPC kernels
+    weave = Dependency("weave")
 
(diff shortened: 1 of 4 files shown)


dmadisetti:
Optional deps look like they'll be fixed with #8077 
Going to merge for pixi lock

PR #8109, opened 2026-02-03
fix: handle objects with `__getattr__` returning non-callable values in serialization

### LINKED ISSUE #8096: Objects implementing `getattr` are interpreted as having special methods for serialization.
### Describe the bug

I've been working with rdflib, and it provides a `Namespace` class to make creation of RDF URIs simpler. It works as follows:
```python
import rdflib
ns = rdflib.Namespace("http://www.w3.org/2001/XMLSchema#")
ns.dateTime == rdflib.URIRef("http://www.w3.org/2001/XMLSchema#dateTime")
```

However, in a marimo notebook, the Namespace class can't be serialized due to it looking like it has the `_marimo_serialize_` [special function](https://github.com/marimo-team/marimo/blob/c374384799bec177908a14a0799b3d02fc96fffd/marimo/_messaging/msgspec_encoder.py#L28). `hasattr(ns, "_marimo_serialize_")` returns true, and accessing the attribute returns a URIRef that is not callable. The same will happen with the `_mime_` function.

I think a possible fix for this concrete error would be checking for `callable()` on the returned value before calling it (I'll try and make a PR).

### Will you submit a PR?

- [x] Yes

### Environment

```
{
  "marimo": "0.19.7",
  "editable": false,
  "location": "/home/jeeger/projects/work/50hertz/throwaway/validation-poc/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.18.5+deb14-amd64",
  "Processor": "",
  "Python Version": "3.13.11",
  "Locale": "de_DE",
  "Binaries": {
    "Browser": "144.0.7559.59",
    "Node": "v22.22.0"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.1",
    "narwhals": "2.16.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "basedpyright": "1.37.3",
    "loro": "1.10.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

mscolnick:
### Summary

Objects that implement `__getattr__` (like `rdflib.Namespace`) caused serialization errors because `hasattr()` returns `True` for any attribute name, but the returned value may not be callable.

Added `getcallable()` utility function that safely retrieves callable attributes, returning `None` if the attribute doesn't exist or isn't callable

Closes #8096

### DIFF
--- a/marimo/_utils/methods.py
+++ b/marimo/_utils/methods.py
@@ -23,7 +23,7 @@ def getcallable(obj: object, name: str) -> Callable[..., Any] | None:
     This safely handles objects that implement __getattr__ and return
     non-callable values for any attribute name.
     """
-    if (attr := getattr(obj, name)) is not None:
+    if (attr := getattr(obj, name, None)) is not None:
         if callable(attr):
             return cast(Callable[..., Any], attr)
     return None


mscolnick:
test was fixed and merged to main

PR #8110, opened 2026-02-03
enhancement: anywidget dynamic imports in static notebooks

mscolnick:
Adds support for import() of virtual files in static notebooks by resolving /@file/ URLs to blob URLs.

Unlike fetch(), dynamic import() cannot be patched globally. This change enables anywidget JS modules embedded in exported HTML notebooks to load correctly.

### DIFF
--- a/frontend/src/utils/blob.ts
+++ b/frontend/src/utils/blob.ts
@@ -14,32 +14,19 @@ export function serializeBlob<T>(blob: Blob): Promise<DataURLString> {
   });
 }
 
-export function deserializeBlob(serializedBlob: DataURLString): Promise<Blob> {
-  return new Promise((resolve, reject) => {
-    try {
-      // Extract the base64 data from the data URL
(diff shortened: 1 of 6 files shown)


frontend/src/utils/blob.ts
```diff
@@ -14,32 +14,19 @@ export function serializeBlob<T>(blob: Blob): Promise<DataURLString> {
...
+  const [prefix, base64Data] = serializedBlob.split(",", 2);
+  const mimeType = /^data:(.+);base64$/.exec(prefix)?.[1];
+  // Decode the base64 string
+  const binaryString = atob(base64Data);
+  // Convert the binary string to an array buffer
+  const len = binaryString.length;
+  const bytes = new Uint8Array(len);
+  for (let i = 0; i < len; i++) {
+    bytes[i] = binaryString.charCodeAt(i);
   }
```
manzt, line 28: could maybe use our base64 utilities here.

PR #8111, opened 2026-02-03
fix: gracefully handle timeouts/input from uv when adding script metadata

mscolnick:
This pull request improves the robustness of the `_ensure_marimo_in_script_metadata` function in `marimo/_cli/sandbox.py` by preventing potential hangs and handling timeouts when running subprocesses.

Reliability improvements for subprocess execution:

* Added `stdin=subprocess.DEVNULL` to prevent hanging on Windows when `uv` might wait for input.
* Set a `timeout=30` for the subprocess call and added handling for `TimeoutExpired` exceptions to log a warning if the operation times out.

### DIFF
--- a/marimo/_cli/sandbox.py
+++ b/marimo/_cli/sandbox.py
@@ -391,10 +391,16 @@ def _ensure_marimo_in_script_metadata(name: str | None) -> None:
             check=True,
             capture_output=True,
             text=True,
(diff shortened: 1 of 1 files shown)


mscolnick:
> Was this from an issue that we should close out?

@dmadisetti its from the flakey windows tests, not an issue

PR #8113, opened 2026-02-04
fix: remove some deprecation warnings

mscolnick:
Remove various deprecation warnings that come up in CI.

### DIFF
--- a/marimo/_ast/parse.py
+++ b/marimo/_ast/parse.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 import ast
 import io
-import sys
 import token as token_types
 import warnings
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


marimo/_ast/parse.py
```diff
@@ -1000,9 +999,9 @@ def is_body_cell(node: Node) -> bool:
 def _is_ellipsis(node: Optional[Node]) -> bool:
     if node is None:
         return False
-    if sys.version_info < (3, 14):
-        return isinstance(node, ast.Ellipsis)
-    return isinstance(node, ast.Constant) and node.value == ...
+    # ast.Ellipsis is deprecated in 3.12+ and removed in 3.14
```
dmadisetti, line 1002: Don't we support 3.10 though?

mscolnick: yes, and this still works in 3.10 as far as i can tell. `ast.Ellipsis` was deprecated in 3.12, but `isinstance(node, ast.Constant) and node.value is ...` is still equivalent

dmadisetti:
```python
uv run --python=3.10 python -c "import ast; node=ast.parse('...');print(isinstance(node, ast.Constant) and node.value is ...)"
```

gave me `False`

dmadisetti: Nvm, it's

```python
uv run --python=3.10 python -c "import ast; node=ast.parse('...').body[0].value;print(isinstance(node, ast.Constant) and node.value is ..., ast.dump(node))"
```

PR #8124, opened 2026-02-04
fix: decouple anywidget model and view

### LINKED ISSUE #7686: Undefined anywidget state when switching between edit and app modes (0.18.4)
### Describe the bug

When rendering an anywidget, it seems like some parts of its state are set to undefined when switching between edit mode and app mode (in either direction). We noticed this with our own custom anywidgets when upgrading from marimo 0.18.3 to 0.18.4. I'm including a minimal example using `drawdata.ScatterWidget` below that seems to have the same behaviour:

1. **Load widget in edit mode**
Image

2. **Draw some sample data in edit mode**

Image

3. **Switch to app mode — everything still works functionally, but the brush size state is displayed as undefined**
Image

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.18.4",
  "editable": false,
  "location": "~/.cache/uv/archive-v0/0KB-X7fx0HxBHIpXbmuD8/lib/python3.13/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "21.6.0",
  "Processor": "arm",
  "Python Version": "3.13.5",
  "Locale": "en_GB",
  "Binaries": {
    "Browser": "143.0.7499.170",
    "Node": "v20.15.0"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10",
    "narwhals": "2.15.0",
    "packaging": "25.0",
    "psutil": "7.2.1",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.50.0",
    "tomlkit": "0.13.3",
    "uvicorn": "0.40.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.10.3"
  },
  "Experimental Flags": {
    "external_agents": true,
    "chat_modes": true
  }
}
```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "anywidget==0.9.21",
#     "drawdata==0.3.8",
#     "marimo",
#     "traitlets==5.14.3",
# ]
# ///

import marimo

__generated_with = "0.18.4"
app = marimo.App(width="medium", sql_output="native")


@app.cell
def _():
    import marimo as mo
    from drawdata import ScatterWidget
    return ScatterWidget, mo


@app.cell
def _(ScatterWidget, mo):
    widget = mo.ui.anywidget(ScatterWidget())
    widget
    return


if __name__ == "__main__":
    app.run()
```

mscolnick:
_Code code-written with @manzt_ 

These changes decouple the AnyWidget model and view. Previously, models and views were created together. The plugin received full widget state as its value, created the model at render time, and widget messages were piggybacked on the `send-ui-element-message` notification with runtime type-checks to distinguish them from normal UI element messages.

This meant models couldn't exist before their views rendered, and `initialize()` was incorrectly called on every mount rather than once per model lifetime (violating the AFM spec).

Now, models are sent before rendering the anywidget view via a dedicated `model-lifecycle` WebSocket notification (open/update/custom/close). The view receives just a `model_id` and grabs the pre-existing model from a global model manager. This properly separates the model lifecycle from the view lifecycle, matching the AFM spec where `initialize()` runs once per model and `render()` runs once per view.

These changes also include cleanups and removal of old code paths:

- Typed model lifecycle protocol using msgspec tagged unions on the backend and zod schemas on the frontend, replacing untyped message dicts and TypedDicts that were hard to keep in sync
- Each message type owns its serialization (`into_comm_payload_content()`), removing manual dict construction in the comm manager
- Removed `encode_to_wire`/`decode_from_wire`, `_prev_state` tracking, `send_to_widget` RPC, wire format handling in the plugin, and `isMessageWidgetState` filtering
- Few perf improvements: native `Uint8Array.fromBase64`/`.toBase64` when available (~7-25x faster), removed `structuredClone` in wire format decoding, eliminated double encode/decode round-trip through the UIElement value system

Closes #7686

### DIFF
--- a/tests/_server/api/endpoints/test_execution.py
+++ b/tests/_server/api/endpoints/test_execution.py
@@ -97,10 +97,11 @@ class TestExecutionRoutes_EditMode:
             json={
                 "modelId": "model-1",
                 "message": {
+                    "method": "update",
                     "state": {"key": "value"},
                     "bufferPaths": [["a"], ["b"]],
                 },
-                "buffers": ["buffer1", "buffer2"],
+                "buffers": ["YnVmZmVyMQ==", "YnVmZmVyMg=="],
             },
         )
         assert response.status_code == 200, response.text
@@ -296,10 +297,11 @@ class TestExecutionRoutes_RunMode:
             json={
                 "modelId": "model-1",
                 "message": {
+                    "method": "update",
                     "state": {"key": "value"},
                     "bufferPaths": [["a"], ["b"]],
                 },
-                "buffers": ["buffer1", "buffer2"],
+                "buffers": ["YnVmZmVyMQ==", "YnVmZmVyMg=="],
             },
         )
         assert response.status_code == 200, response.text


frontend/src/plugins/impl/anywidget/model.ts
```diff
@@ -111,20 +137,25 @@ export class Model<T extends Record<string, any>> implements AnyModel<T> {
...
+      .sendModelValue({
+        modelId: this.modelId,
+        message: {
+          state: state,
+          bufferPaths: bufferPaths,
+          method: "custom",
+          content: content,
+        },
+        buffers: buffers,
+      })
```
mscolnick: this is off, likely needs fixing

marimo/_plugins/ui/_impl/comm.py
```diff
@@ -32,26 +32,42 @@ def receive_comm_message(
...
         msg: TypedModelMessagePayload = {
             "content": {
                 "data": {
                     "state": message.state,
-                    "method": "update",
+                    "method": message.method,
                     "buffer_paths": message.buffer_paths,
+                    "content": message.content,
                 }
             },
```
mscolnick: this is awkwardly nested and would be nice to simplify

marimo/_plugins/ui/_impl/from_anywidget.py
```diff
@@ -168,141 +200,67 @@ def __init__(self, widget: AnyWidget):
...
-            return cast(T, encode_to_wire(merged))
-
-        LOGGER.warning(
-            f"Expected anywidget value to be a dict, got {type(value)}"
-        )
-        self._prev_state = value
-        return cast(T, value)
+        # Otherwise, it's a state update from the frontend
+        # Update the widget's state
+        self.widget.set_state(value)
```
mscolnick, line 217: is this getting hit and is this causing issues?

mscolnick: this doesn't get called as far as i can tell

PR #8134, opened 2026-02-05
fix: avoid unnecessary throwing error in delete cell

mscolnick:
Deleting a cell would then expand its children, but this would throw an error that was later caught. This is just some cleanup to avoid that

### DIFF
--- a/frontend/src/utils/id-tree.tsx
+++ b/frontend/src/utils/id-tree.tsx
@@ -522,11 +522,10 @@ export class CollapsibleTree<T> {
     if (found.length === 0) {
       return this;
     }
-    let result: CollapsibleTree<T> = this;
-    for (const node of found) {
-      result = result.expand(node);
-    }
-    return result;
+    return found.reduce<CollapsibleTree<T>>(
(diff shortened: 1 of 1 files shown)


frontend/src/utils/id-tree.tsx
```diff
@@ -524,15 +522,10 @@ export class CollapsibleTree<T> {
...
     for (const node of found) {
-      try {
-        result = result.expand(node);
-      } catch {
-        // Don't care if its the last node and its not expanded
-      }
+      result = result.expand(node);
     }
-
     return result;
```
Light2Dark:
```suggestion
    return found.reduce<CollapsibleTree<T>>(
      (result, node) => result.expand(node),
      this,
    );
```

Light2Dark: could also add this `// eslint-disable-next-line unicorn/no-this-assignment, @typescript-eslint/no-this-alias`.

mscolnick: oh i fixed this, but havent pushed up yet actually. thanks

PR #8136, opened 2026-02-05
fix: search functionality for DataFrames with index columns

### LINKED ISSUE #7945: Search functionality returns no results for df.value_counts()
### Describe the bug

Image
Image
Image

I am encountering an issue where searching within the output of df.value_counts() consistently returns no results. This persists even when searching for single characters or exact matches (e.g., 'D' or 'Dave').

The attached screenshots demonstrate the behavior:

1. The initial df.value_counts() output.
2. An empty result set when searching for the exact string 'Dave'.
3. An empty result set when searching for the single letter 'D'.

### Will you submit a PR?

- [x] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

_No response_

mscolnick:
## Summary

Fixed an issue where table search functionality returned no results for DataFrames with non-trivial indexes (e.g., from `df.value_counts()`).

The root cause was that `PandasTableManager` inherited its `search` method from `NarwhalsTableManager`, which only searched through regular columns, not index columns. While the `to_json_str` method correctly reset the index to display index values in the table UI, the search was performed before this transformation.

Closes #7945

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/pandas_table.py
+++ b/marimo/_plugins/ui/_impl/tables/pandas_table.py
@@ -42,6 +42,34 @@ def _trivial_range_index(index: pd.Index) -> bool:
     return isinstance(index, pd.RangeIndex) and index.name is None
 
 
+def _resolve_index_column_conflicts(df: pd.DataFrame) -> pd.DataFrame:
+    """Rename index names that conflict with column names by appending '_index'.
+
+    Avoids 'ValueError: cannot insert x, already exists' on reset_index().
+    Modifies the DataFrame in-place and returns it.
+    """
+    import pandas as pd
+
+    index_names = df.index.names
+    conflicting_names = set(index_names) & set(df.columns)
+    if not conflicting_names:
+        return df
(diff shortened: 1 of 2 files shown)


Light2Dark:
this breaks unnamed indexes, I think we'd need to set the names back again

image

Screenshot 2026-02-06 at 2 17 08 AM

maybe this would work
```python
if self._has_non_trivial_index():
    index = self._original_data.index
    num_levels = index.nlevels
    original_names = list(index.names)

    df_with_index = self._original_data.reset_index()
    # reset_index() puts index columns at the start
    index_columns = list(df_with_index.columns[:num_levels])

    manager = PandasTableManager(df_with_index)
    result = super(PandasTableManager, manager).search(query)
    native_df: pd.DataFrame = nw.to_native(result.data)

    # Restore the original index structure
    native_df = native_df.set_index(index_columns)
    native_df.index.names = original_names
    return PandasTableManager(native_df)
```

Light2Dark:
i can put up a PR

PR #8137, opened 2026-02-05
Fix: Use cell name to identify setup cells when id is null

mscolnick:
**Problem**

Setup cells don't appear when initializing from __MARIMO_MOUNT_CONFIG__ because the
  cell data has "id": null with "name": "setup", but the code only checked the id
field and generated random IDs instead of using SETUP_CELL_ID.

**Solution**

Added a getCellId() helper that checks both id and name fields:
- If cell.id exists, use it
- If cell.name === "setup", use SETUP_CELL_ID
- Otherwise, generate a new random ID

### DIFF
--- a/frontend/src/core/cells/session.ts
+++ b/frontend/src/core/cells/session.ts
@@ -6,7 +6,7 @@ import { MultiColumn } from "@/utils/id-tree";
 import { Logger } from "@/utils/Logger";
 import { parseOutline } from "../dom/outline";
 import type { NotebookState } from "./cells";
-import { CellId } from "./ids";
+import { CellId, SETUP_CELL_ID } from "./ids";
 import {
   type CellData,
   type CellRuntimeState,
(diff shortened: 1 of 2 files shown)


frontend/src/core/cells/session.ts
```diff
@@ -20,6 +20,22 @@ const EMPTY_STRING = "";
...
+ * If the cell has a name of "setup", use the special SETUP_CELL_ID.
+ * Otherwise, generate a new random ID.
+ */
+function getCellId(cell: { id?: string | null; name?: string | null }): CellId {
+  if (cell.id) {
+    return cell.id as CellId;
+  }
+  if (cell.name === SETUP_CELL_ID) {
+    return SETUP_CELL_ID;
+  }
```
Light2Dark, line 35: move this before the first check? For future-proofing in case it does have `id`

mscolnick: in case it does have an `id`, the `id` is the source of truth and needs to match the backend

PR #8141, opened 2026-02-05
cli: format help message with colors, bold

akshayka:
This PR updates the marimo CLI help message to use colors for command and option names, and bold headings. It generally mirrors the formatting of popular Rust CLIs such as `cargo` and `uv`. Additionally, this PR migrates `cli.print` to use `click.style()` instead of raw ANSI codes.

**Before.**
image

**After.**
image


marimo/_cli/print.py
```diff
@@ -6,6 +6,8 @@
 import sys
 from typing import Any, cast
 
+import click
```
mscolnick: unfortunately `marimo/_cli/print.py` has been used in a few places outside the CLI, so this will break web-assembly. we can either lazy import or fallback

Image

mscolnick: or remove these dependencies

akshayka: Thanks, did a refactor which should fix this

PR #8146, opened 2026-02-05
improvement: support pandas 3.0 str col

### LINKED ISSUE #8093: Pandas v3 issues
> [Highlights of pandas 3.0](https://pandas.pydata.org/community/blog/pandas-3.0.html#highlights-of-pandas-30)
> 
> ...
> Dedicated string data type by default: string columns are now inferred as the new str dtype instead of object, providing better performance and type safety

The first side effect of this that I noticed in marimo tables is the inability to filter by values when a column is of the new str dtype.

There is likely more that propagate beyond marimo - such as [DuckDB SQL](https://github.com/duckdb/duckdb-python/issues/291) on dataframes with new types.

mscolnick:
Closes #8093 

Pandas 3.0 added a new `str` column so this adds support for that when mapping to their unified value type.

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/pandas_table.py
+++ b/marimo/_plugins/ui/_impl/tables/pandas_table.py
@@ -364,9 +364,7 @@ class PandasTableManagerFactory(TableManagerFactory):
                     return ("string", dtype)
                 if lower_dtype == "category":
                     return ("string", dtype)
-                if lower_dtype == "string":
-                    return ("string", dtype)
-                if lower_dtype == "str":
+                if lower_dtype == "string" or lower_dtype == "str":
                     return ("string", dtype)
                 if lower_dtype.startswith("complex"):
                     return ("unknown", dtype)


marimo/_plugins/ui/_impl/tables/pandas_table.py
```diff
@@ -366,6 +366,8 @@ def _map_dtype_to_field_type(
                     return ("string", dtype)
                 if lower_dtype == "string":
```
Light2Dark:
```suggestion
                if lower_dtype == "string" or lower_dtype == "str":
```

marimo/_plugins/ui/_impl/tables/pandas_table.py
```diff
@@ -366,6 +366,8 @@ def _map_dtype_to_field_type(
...
                     return ("string", dtype)
+                if lower_dtype == "str":
+                    return ("string", dtype)
```
Light2Dark:
```suggestion
```

PR #8152, opened 2026-02-05
refactor: move hooks into a container for less arg drilling

mscolnick:
Move kernel hooks into an encapsulated class to avoid prop-drilling and keep code consistent / clean. This maybe would've helped avoid https://github.com/marimo-team/marimo/pull/8138

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -1380,7 +1380,7 @@ class Kernel:
         Returns set of cells that need to be re-run due to state updates.
         """
 
-        # Some hooks that are closures and require the kernel instance
+        # Some hooks are leaky and require the kernel
         # Free cell state ahead of running to relieve memory pressure
         #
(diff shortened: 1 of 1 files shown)


marimo/_runtime/runtime.py
```diff
@@ -3243,6 +3196,18 @@ def _enqueue_control_request(req: CommandMessage) -> None:
...
             set_ui_element_queue.put_nowait(req)
 
+    # Create hooks with mode-specific configuration
+    from marimo._runtime.runner.hooks_post_execution import (
+        attempt_pytest,
+        render_toplevel_defs,
+    )
+
+    hooks = create_default_hooks()
+    if user_config["runtime"].get("reactive_tests", False):
```
dmadisetti, line 3206: `and is_edit_mode`

mscolnick: i can do that in a different PR, just to keep this a "refactor"

PR #8156, opened 2026-02-06
Separate anywidget Model from widget binding lifecycle

manzt:
The Model class previously owned three independent concerns: widget definition loading (ESM import), model state/events/comm, and the binding between a widget definition and a model instance.

This coupling meant every widget instance ran its own ESM import with no cross-instance deduplication, and Model had to manage AbortControllers and widget definition caching that weren't really its responsibility.

This refactor extracts the widget lifecycle into a new `widget-binding.ts` module with three focused abstractions:

- `WidgetDefRegistry` deduplicates ESM imports by content hash so multiple instances of the same widget share a single import.
- `WidgetBinding` owns the initialize/render lifecycle per model, handling hot-reload teardown and view cleanup via AbortSignal.
- `BindingManager` maps model IDs to bindings and ensures cleanup on close messages.

Model is now purely about state, events, and comm. it no longer imports AnyWidget types or manages AbortControllers.

### DIFF
--- a/frontend/src/plugins/impl/anywidget/model.ts
+++ b/frontend/src/plugins/impl/anywidget/model.ts
@@ -91,6 +91,9 @@ class ModelManager {
   }
 
   delete(key: WidgetModelId): void {
+    Logger.debug(
+      `[ModelManager] Deleting model=${key}, aborting lifecycle signal`,
+    );
     this.#entries.get(key)?.controller.abort();
     this.#entries.delete(key);
   }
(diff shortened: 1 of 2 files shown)


frontend/src/plugins/impl/anywidget/model.ts
```diff
@@ -411,25 +341,32 @@ export async function handleWidgetMessage(
...
-      modelManager.set(modelId, model);
+      modelManager.create(
+        modelId,
+        (signal) =>
+          new Model(
+            stateWithBuffers,
+            {
+              async sendUpdate(changeData) {
+                const { state, buffers, bufferPaths } =
+                  serializeBuffersToBase64(changeData);
```
mscolnick, line 360: could we put the `signal` is `aborted` terminal

manzt: Good suggestion. Done.

manzt:
Should be good to go

PR #8157, opened 2026-02-06
Add uv version to `marimo env` output

manzt:
The `marimo env` command reports binary versions for debugging, but was missing `uv` which is increasingly central to Python package management and sandbox workflows. This adds it to the Binaries section, using the existing `find_uv_bin()` helper to respect the `UV` environment variable.


marimo/_cli/envinfo.py
```diff
@@ -98,6 +99,7 @@ def get_system_info() -> dict[str, Union[str, bool, dict[str, Any]]]:
         # back-filled in frontend
         "Browser": get_chrome_version() or "--",
         "Node": get_node_version() or "--",
+        "UV": get_uv_version() or "--",
```
akshayka:
```suggestion
        "uv": get_uv_version() or "--",
```

? :)

PR #8161, opened 2026-02-06
fix: use mount_config.code and filename as fallback in WASM mode

### LINKED ISSUE #8026: `notebookFileStore` does not read `__MARIMO_MOUNT_CONFIG__.code`
### Describe the bug

When the html page for wasm contains no `<marimo-code>`, the fallback is the url arguments, then the local storage.

https://github.com/marimo-team/marimo/blob/718f80320a3bb2b316181dc5515e8d81683ed1fb/frontend/src/core/wasm/store.ts#L114

It does not take into account `mount_config.code`, which makes embedding marimo inside other contexts impossible (see also #7949). In my use-case, I fallback to the default marimo notebook, wihtout having any clue.

I think there might be another related problem:
On https://marimo.app, the "create permalink" button does not work. If I click on it and open the link in a private window, I get a different code (either a previous version or a notebook in my local storage). Maybe I am supposed to save beforehand, but when I click save I get `Failed to handle request: sendSave Error: Failed to fetch: `. If the error is unrelated, I will open a new issue.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
N/A (wasm)
```

</details>

### Code to reproduce

N/A

ffmiruz:
When embedding marimo in WASM mode without a marimo-code element, the mount config's code and filename options are now used as fallbacks.

Code fallback priority:
1. mount_config.code
2. marimo-code element
3. URL params
4. local storage
5. remote default
6. empty notebook

Filename fallback priority:
1. mount_config.filename
2. URL param (?filename=...)

Fixes #8026

### DIFF
--- a/frontend/src/core/wasm/store.ts
+++ b/frontend/src/core/wasm/store.ts
@@ -4,9 +4,9 @@ import {
   compressToEncodedURIComponent,
   decompressFromEncodedURIComponent,
 } from "lz-string";
-import { TypedLocalStorage } from "@/utils/storage/typed";
-import { store } from "@/core/state/jotai";
 import { codeAtom } from "@/core/saving/file-state";
+import { store } from "@/core/state/jotai";
+import { TypedLocalStorage } from "@/utils/storage/typed";
 import { PyodideRouter } from "./router";
 
 export interface FileStore {
--- a/frontend/src/core/wasm/__tests__/store.test.ts
+++ b/frontend/src/core/wasm/__tests__/store.test.ts
@@ -1,5 +1,7 @@
 /* Copyright 2026 Marimo. All rights reserved. */
 import { describe, expect, it, vi } from "vitest";
+import { codeAtom } from "../../saving/file-state";
+import { store } from "../../state/jotai";
 import {
   CompositeFileStore,
   domElementFileStore,
@@ -7,8 +9,6 @@ import {
   mountConfigFileStore,
   notebookFileStore,
 } from "../store";
-import { store } from "../../state/jotai";
-import { codeAtom } from "../../saving/file-state";
 
 describe("localStorageFileStore", () => {
   it("calls set with correct contents on saveFile", () => {
--- a/frontend/src/core/wasm/bridge.ts
+++ b/frontend/src/core/wasm/bridge.ts
@@ -9,7 +9,6 @@ import { Logger } from "@/utils/Logger";
 import { reloadSafe } from "@/utils/reload-safe";
 import { generateUUID } from "@/utils/uuid";
 import { notebookIsRunningAtom } from "../cells/cells";
-import { filenameAtom } from "../saving/file-state";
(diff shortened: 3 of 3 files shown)


ffmiruz:

frontend/src/core/wasm/store.ts
```diff
@@ -67,6 +67,22 @@ const remoteDefaultFileStore: FileStore = {
...
+export function setMountConfigCode(code: string | null) {
+  mountConfigCode = code;
+}
+
+export const mountConfigFileStore: FileStore = {
+  saveFile(_contents: string) {
+    // Read-only: mount config code is set externally
+  },
+  readFile() {
+    return mountConfigCode;
```
mscolnick: Maybe we can just read directly from the store here? And then can remove the `setMountConfig` function?

ffmiruz: Done. Redo to not hold a copy of config code somewhere else.

frontend/src/core/wasm/store.ts
```diff
@@ -118,8 +129,10 @@ export const notebookFileStore = new CompositeFileStore([
 ]);
 
 export const fallbackFileStore = new CompositeFileStore([
-  // Prefer then local storage, then remote default, then empty
+  // Prefer local storage (persists user edits), then mount config code,
+  // then remote default, then empty
   localStorageFileStore,
+  mountConfigFileStore,
```
mscolnick: actually taking a step back, shouldn't this get picked up by `domElementFileStore` in `notebookFileStore`?

mscolnick: oh, maybe mount config is being called without using `<marimo-code/>`?

do we want this to take precedence over local-storage? then maybe this should be in `notebookFileStore` (not a fallback)?

mscolnick: it probably should go: 

1. `mountConfig` (because you make load a page once but mount a few times via naivation
2. `domElement` (one per page)
3. `url`

thoughts?

ffmiruz: Hmm how i see it is, mount code is "init". And then if domELement is set, there is intention to override it. Same goes for url.
About localStorage, yea I'm a bit fuzzy. My thinking is, localStorage is to survive reload but if it always behind mount code then it is never useful, no? idk

But yea, mount code, domELement, url in some order inside notebookFileStore is probably the solution here.

mscolnick: @ffmiruz, ping me when you make the change. i think we are on the same page

ffmiruz: Can you confirm the order

mscolnick: mount code, domELement, url in  notebookFileStore, seems correct to me.

ffmiruz: Done

PR #8163, opened 2026-02-06
Merge model messages into a single replay snapshot

manzt:
When a browser reconnects, `SessionView` replays _all_ buffered widget model messages. Previously it stored the full message history per model (open, update, update, ...) which grew without bound and replayed redundant intermediate states.

This replaces the append-only list with a `ModelReplayState` that aggregates open/update messages into one snapshot per model. Updates merge state dicts and replace stale buffer paths, close removes the entry entirely, and custom messages are skipped since they're ephemeral. On replay only a single synthetic ModelOpen is emitted per live model, keeping the reconnection payload minimal.

We could similarly consider reducing ui_element_messages to a single message per element (similar to `ModelReplayState`) instead of keeping the full history. However, chat "stream_chunk" messages require full replay. Maybe we could filter those out? Either way, out of scope for this PR.

### DIFF
--- a/marimo/_session/state/session_view.py
+++ b/marimo/_session/state/session_view.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 import time
 from dataclasses import dataclass
-from typing import Any, Literal, Optional, cast
+from typing import Any, Literal, Optional, Union, cast
 
 from marimo import _loggers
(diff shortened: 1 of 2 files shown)


mscolnick (approved):
nice, great cleanup

PR #8166, opened 2026-02-06
Persist anywidget model state in static HTML exports

manzt:
Anywidgets don't render in static HTML exports because there's no kernel to send the WebSocket "open" message that delivers model state. The `MODEL_MANAGER.get()` call times out after 10s and the widget never appears.

These changes embed all live model states as a JSON blob in the existing `__MARIMO_STATIC__` global. On frontend startup, before React renders, we iterate over the embedded states and feed synthetic "open" messages through the existing `handleWidgetMessage` path. In static mode the comm uses no-op functions so `save_changes()` and `send()` silently do nothing, but client-side interactions (map panning, slider dragging) still work because `model.set()` updates local state and fires events.

The Python side reuses the new `ModelReplayState` (#8163) from our `SessionView` to export the latest static state for the front end.


frontend/src/core/static/types.ts
```diff
@@ -1,8 +1,15 @@
...
-import type { DataURLString } from "@/utils/json/base64";
+import type { Base64String, DataURLString } from "@/utils/json/base64";
 
 export type StaticVirtualFiles = Record<string, DataURLString>;
 
+export interface StaticModelState {
+  state: Record<string, unknown>;
+  buffer_paths: (string | number)[][];
+  buffers: Base64String[];
+}
```
manzt: Should maybe export this type or reuse the open notification type to ensure no drift if those types change.

mscolnick (approved):
very nice!

frontend/src/core/static/types.ts
```diff
@@ -1,8 +1,15 @@
 /* Copyright 2026 Marimo. All rights reserved. */
-import type { DataURLString } from "@/utils/json/base64";
+import type { Base64String, DataURLString } from "@/utils/json/base64";
 
 export type StaticVirtualFiles = Record<string, DataURLString>;
 
+export interface StaticModelState {
```
mscolnick: we do export `ModelUpdate` (just can omit method), or keep it in there for future proof

manzt: Ok so I changed the serialized data type to `modelNoficiations: Array<ModelLifecycleNotification>` instead of `modelStates: Record<string, Omit<ModelOpen, "method">`.

There wasn't much on the Python side that was holding us to keeping that interface consistent, and this way we just "replay" the notifications.

tests/_cli/test_cli.py
```diff
@@ -1382,8 +1382,8 @@ def test_cli_with_custom_pyproject_config_no_file(tmp_path: Path) -> None:
     def assert_custom_config(contents: bytes | None) -> None:
         assert contents is not None
         # Verify that the custom config is applied
-        assert b'"line_length": 111' in contents
```
manzt, line 1385: Not totally sure what we should do here. The failing tests run marimo in `--sandbox` mode, which re-launches via uv run `--isolated` and installs marimo from PyPI.. not the local build. So the sandboxed server still uses `json.dumps` (spaces after :), but the snapshot updat expects the new compact msgspec format (no spaces). These will stay broken until a release includes the `json_script` change.

mscolnick: Can you do an `or` for now and we can figure out how to decouple some cli tests with --sandbox

mscolnick (approved):

tests/_server/templates/snapshots/export2.txt
```diff
@@ -59,7 +59,7 @@
...
...etion": {"activate_on_typing": true, "copilot": false, "signature_hint_on_typing": false}, "diagnostics": {"sql_linter": true}, "display": {"cell_output": "below", "code_editor_font_size": 14, "dataframes": "rich", "default_table_max_columns": 50, "default_table_page_size": 10, "default_width": "medium", "reference_highlighting": true, "theme": "light"}, "formatting": {"line_length": 79}, "keymap": {"overrides": {}, "preset": "default"}, "language_servers": {"pylsp": {"enable_flake8": false, "enable_mypy": true, "enable_pydocstyle": false, "enable_pyflakes": false, "enable_pylint": false, "enable_ruff": true, "enabled": false}}, "mcp": {"mcpServers": {}, "presets": []}, "package_management": {"manager": "uv"}, "runtime": {"auto_instantiate": false, "auto_reload": "off", "default_csv_encoding": "utf-8", "default_sql_output": "auto", "on_cell_change": "autorun", "output_max_bytes": 8000000, "reactive_tests": true, "std_stream_max_bytes": 1000000, "watcher_on_save": "lazy"}, "save": {"autosave": "after_delay", "autosave_delay": 1000, "format_on_save": false}, "server": {"browser": "default", "follow_symlink": false}, "snippets": {"custom_paths": [], "include_default_snippets": true}},
```
manzt: I think these are changing from `"manager": "uv"`

PR #8168, opened 2026-02-06
Batch anywidget model updates like UI element updates

manzt:
The anywidget refactors in #8156, #8159, and #8163 separated model lifecycle from widget binding and moved to a dedicated ModelCommand for frontend-to-backend communication.

However, unlike `UpdateUIElementCommand` which goes through `SetUIElementRequestManager`'s drain-and-merge path, the new `ModelCommand` is processed immediately. Each rapid model update (e.g. dragging a map widget) triggered an individual cell re-execution.

These changes puts `ModelCommand` on the same shared queue as `UpdateUIElementCommand` so both go through the same batching pipeline. When multiple model updates arrive in quick succession, they are now drained and merged (last-write-wins per model ID on state keys), matching the existing UI element behavior.

The handler for model messages also now enqueues the resulting `UpdateUIElementCommand` back through the control queue instead of calling `set_ui_element_value` directly, so the downstream cell re-execution also benefits from batching.

### DIFF
--- a/packages/openapi/api.yaml
+++ b/packages/openapi/api.yaml
@@ -3048,7 +3048,8 @@ components:
       description: "Widget model message command.\n\n    Handles widget model communication\
         \ between frontend and backend.\n\n    Attributes:\n        model_id: Widget\
         \ model identifier.\n        message: Model message (update or custom).\n\
-        \        buffers: Base64-encoded binary buffers."
+        \        buffers: Base64-encoded binary buffers.\n        token: Unique identifier\
+        \ for deduplication across dual queues."
       properties:
         buffers:
           items:
@@ -3065,6 +3066,8 @@ components:
               update: '#/components/schemas/ModelUpdateMessage'
             propertyName: method
         modelId:
+          type: string
+        token:
           type: string
         type:
           enum:
@@ -3182,6 +3185,8 @@ components:
               update: '#/components/schemas/ModelUpdateMessage'
             propertyName: method
         modelId:
+          type: string
+        token:
           type: string
       required:
       - modelId
@@ -4839,7 +4844,7 @@ components:
       type: object
 info:
   title: marimo API
-  version: 0.19.7
+  version: 0.19.9
 openapi: 3.1.0
 paths:
   /@file/{filename_and_length}:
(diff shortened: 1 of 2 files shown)


marimo/_runtime/runtime.py
```diff
@@ -2281,14 +2281,17 @@ async def handle_receive_model_message(
...
+            # This prevents rapid model updates from bombarding the kernel
+            # with individual cell re-executions.
             if ui_element_id and state:
-                await self.set_ui_element_value(
+                self.enqueue_control_request(
                     UpdateUIElementCommand.from_ids_and_values(
                         [(UIElementId(ui_element_id), state)]
                     )
                 )
-                broadcast_notification(CompletedRunNotification())
```
mscolnick, line 2275: do we still need this completed run notification?

mscolnick: nevermind, i see its processed in handle_set_ui_element_value

marimo/_session/types.py
```diff
@@ -35,7 +35,9 @@ class QueueManager(Protocol):
     """Protocol for queue management."""
 
     control_queue: QueueType[commands.CommandMessage]
-    set_ui_element_queue: QueueType[commands.UpdateUIElementCommand]
+    set_ui_element_queue: QueueType[
+        commands.UpdateUIElementCommand | commands.ModelCommand
```
mscolnick: maybe we can share more types of `commands.UpdateUIElementCommand | commands.ModelCommand` (could live in commands.py)

marimo/_runtime/utils/set_ui_element_request_manager.py
```diff
@@ -1,62 +1,141 @@
...
-                merged[ui_id] = value
-        last_request = requests[-1]
-        return UpdateUIElementCommand(
-            object_ids=list(merged.keys()),
-            values=list(merged.values()),
-            token=last_request.token,
-            request=last_request.request,
-        )
+    @staticmethod
+    def _merge(
```
mscolnick: this does change the order of the list based on batches, but i _think_ that is ok. maybe a callout?

definitely worth writing tests for this logic. probably overkill to modularize it but could create a `MergeFunction[T] = Callable[list[T], list[T]]`. im not sure if we will have more in future.

Issue #7777, opened 2026-01-10
Justify stops working in hstack/vstack grid when column widths are set


mmlynarik:
### Describe the bug

I would like to create a grid of two columns and five rows, with second column occupying cca 20% of total width. However, when I use combination of vstack and hstack, setting both `justify` for `vstack` and `width` for `hstack`, the layout breaks down (justify is not longer active):

First, start with only justify. This is okay, but it still lacks the widths proportions set.
```
text = "QGEN-DE: Sicherheitsinformationen Teile der Anlage, die im abgeschalteten Zustand der Anlage unter Spannung bleiben können SAP-Nummer 211919"
q = mo.vstack([mo.md(text)] * 5)
s = mo.vstack([mo.md(str(0.5))] * 5, align="end", justify="space-between")
mo.hstack([q, s])
```

<img width="1095" height="326" alt="Image" src="https://github.com/user-attachments/assets/b78e8169-6800-449c-9720-63d363afb4e7" />

However, when I try to add widths so that the chart makes sense, the justify around is gone, the numbers are no longer nicely separated from each other:

```
text = "QGEN-DE: Sicherheitsinformationen Teile der Anlage, die im abgeschalteten Zustand der Anlage unter Spannung bleiben können SAP-Nummer 211919"
q = mo.vstack([mo.md(text)] * 5)
s = mo.vstack([mo.md(str(0.5))] * 5, align="end", justify="space-between")
mo.hstack([q, s], widths=[5, 1])
```

<img width="1095" height="326" alt="Image" src="https://github.com/user-attachments/assets/110f3242-9c1b-4481-9016-067f7367766f" />

Is this a bug, or am I doing something wrong or could you please help me fix the problem?

Thank you for otherwise super cool tool!

### Environment

<details>

```
Replace this line with the output. Leave the backticks in place.
```

</details>

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
Replace this line with the output of marimo env. Leave the backticks in place.
```

</details>

Light2Dark:
It looks like we are not applying correct styles when the width is set in https://github.com/marimo-team/marimo/blob/444e708af64b81067d355c22e69e7a1c47291011/marimo/_plugins/stateless/flex.py#L54-L60

mmlynarik:
Thanks for the reply. Do you think this issue requires a complex fix or should it be relatively easy to do?

Light2Dark:
[...]

Issue #7184, opened 2025-11-15
Front End Interactivity - Cursor still indicating left panel is resizable when left panel is collapsed


kelreeeeey:
### The bug

It is pretty much the title, here's a video showing what i meant.

https://github.com/user-attachments/assets/aec98e53-ac6d-475f-95f0-ace5d9ca6342

Additional context:

- I'm using `column` layout
- The browser is Brave

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
marimo v0.17.8
python 3.11
```

</details>

mscolnick:
Likely a one-line css change. Thanks for filing the issue, we can look into it.

kelreeeeey:
You're welcome!!!

Issue #6852, opened 2025-10-20
Sidebar collapse on flex items squishes text


Deufel:
### Describe the bug

using something like this in a marimo notebook; 

```python
mo.sidebar( [ mo.outline() ] )
```

<img width="460" height="513" alt="Image" src="https://github.com/user-attachments/assets/9ba824b0-d979-4e7e-8586-7018ef72d763" />

but when collapsed it squishes the text to an awkward size... 

<img width="105" height="734" alt="Image" src="https://github.com/user-attachments/assets/10084844-351a-4992-9224-2504c17a7378" />

I understand that is probably not exactly how that sidebar component was intended to be used but just a bit of a weird behavior here. 

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.17.0",
  "editable": false,
  "location": "/home/mike/.local/share/uv/tools/marimo/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.16.8-asahi-1-2-ARCH",
  "Processor": "",
  "Python Version": "3.13.7",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.8.0"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.9.0",
    "packaging": "25.0",
    "psutil": "7.1.1",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.3",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.38.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.8.1",
    "pytest": "8.4.2"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

```python
mo.sidebar( [ mo.outline() ] )
```

Issue #6680, opened 2025-10-05
Packages side bar blocked by dependencies error popup


MrAndrew:
### Describe the bug

Issue: 
The packages side bar menu is unusable from the keyboard in Safari when there is a dependency error popup displayed.

System:
Macbook on Mac silicon using Safari web browser.

How to reproduce:
Open up a notebook using `marimo edit` with the default browser set to Safari when the notebook has dependencies that are not currently installed. This should display the red popup in the right top listing the missing dependencies.

Then, open up the packages side menu bar, where you can add/install said missing dependencies. It will display correctly, but the search bar to enter the package name to install using pip will be unresponsive. 

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
Replace this line with the output of marimo env. Leave the backticks in place.
```

</details>

mscolnick:
@MrAndrew can you share a screenshot? Im not sure i totally understand the issue

MrAndrew:
@mscolnick 

Missing packages: 

<img width="605" height="241" alt="Image" src="https://github.com/user-attachments/assets/01779197-8aa9-4408-9de8-04584fd16f53" />

Manage Packages side menu where the edit text is not selectable or interactive: 

<img width="615" height="297" alt="Image" src="https://github.com/user-attachments/assets/dbfaf304-3006-4e2b-9520-850a754a3d74" />

Now it is editable once I close the missing packages error message popup:

<img width="590" height="239" alt="Image" src="https://github.com/user-attachments/assets/b838fbb0-1ab5-403f-b9fc-ed0f2f64bd41" />

mscolnick:
oh i see now, thank you! we can fix this

MrAndrew:
@mscolnick awesome thanks! I would have offered a PR but I wasn't sure where to began with this codebase to figure things out. I hope it's an easy fix for you!

Light2Dark:
I can only reproduce in safari (maybe z-stacking issues), but here's the [codepath](https://github.com/marimo-team/marimo/blob/36647e2e5de476de4f17430e3fa31e99564d081e/frontend/src/components/editor/package-alert.tsx#L98)

PR #8205, opened 2026-02-08
fix: CSS and layout fixes across frontend components

AhmadYasser1:
## Summary
Targeted CSS and layout fixes across frontend components.

- Prevent margin collapse between consecutive cell outputs (#3063)
- Fix package alert z-index blocking sidebar interactions (#6680)
- Hide sidebar content when collapsed instead of squishing (#6852)
- Prevent resize cursor on collapsed panels (#7184)
- Preserve justify on nested stacks when widths/heights are set (#7777)
- Prevent tooltips from being clipped inside containers (#3237)

## Test Plan
- Verified existing tests pass for all changed components
- Each fix is scoped to the specific issue's reproduction case

### DIFF
--- a/frontend/src/components/editor/cell/toolbar.tsx
+++ b/frontend/src/components/editor/cell/toolbar.tsx
@@ -59,11 +59,7 @@ export const ToolbarItem: React.FC<ToolbarItemProps> = ({
 
   if (tooltip) {
     return (
-      <Tooltip
-        content={tooltip}
-        side="top"
-        delayDuration={200}
-      >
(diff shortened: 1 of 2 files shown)


mscolnick:
hey @AhmadYasser1, thanks for all the contributions. we'd prefer they are broken up into one PR per issue (instead of batching them), this way comments on one won't block others.

AhmadYasser1:
Hey @mscolnick, thanks for the heads up. I’ll split these into one PR per issue and close this one.

PR #8214, opened 2026-02-08
feat: add disabled parameter to mo.ui.chat

### LINKED ISSUE #3823: Support disabling mo.ui.chat input
### Description

I was hoping to make my chatbot disabled until the user entered an api key in another input but didn't see any parameter to disable the input.

### Suggested solution

Support disabled parameter on the Chat input.

When set to true it can create a non interactive overlay over the input.

Html now supports inert attribute to disable interactions inside an element. Inert elements can be styled to achieve the disabled styling.

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert

### Alternative

Currently using mo.stop to not show the cell at all. 

### Additional context

Here's the notebook where I want to use this: 

https://aszenz.github.io/kashmiri-chatbot/

By not showing the input, I have a feeling people won't understand how the chatbot looks like.

AhmadYasser1:
## Summary
Adds a `disabled` parameter to `mo.ui.chat` that prevents user interaction when set to `True`. Uses the HTML `inert` attribute on the form container so all child inputs, buttons, and links are automatically non-interactive.

Closes #3823

## Test Plan
- Added Python test for disabled parameter serialization
- Visual: `opacity-50 cursor-not-allowed` styling when disabled

### DIFF
--- a/frontend/src/plugins/impl/chat/ChatPlugin.tsx
+++ b/frontend/src/plugins/impl/chat/ChatPlugin.tsx
@@ -47,6 +47,7 @@ export const ChatPlugin = createPlugin<{ messages: UIMessage[] }>(
       maxHeight: z.number().optional(),
       config: configSchema,
       allowAttachments: z.union([z.boolean(), z.string().array()]),
+      disabled: z.boolean().default(false),
     }),
   )
   .withFunctions<PluginFunctions>({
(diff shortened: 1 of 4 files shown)


mscolnick:
This looks good, here is a screenshot
image

mscolnick (approved):

PR #8219, opened 2026-02-08
fix: preserve justify on nested stacks when widths/heights are set

### LINKED ISSUE #7777: Justify stops working in hstack/vstack grid when column widths are set

AhmadYasser1:
## Summary
Children of `hstack`/`vstack` with `flex: 1` didn't stretch because the wrapper div for width/height constraints wasn't a flex container. Makes the wrapper a flex container with `min-width: 0` / `min-height: 0` to prevent overflow.

Closes #7777

## Test Plan
- Updated existing flex tests to verify flex properties are preserved

### DIFF
--- a/marimo/_plugins/stateless/flex.py
+++ b/marimo/_plugins/stateless/flex.py
@@ -57,7 +57,14 @@ def _flex(
         child_flex = child_flexes[idx]
         if child_flex is None:
             return ""
-        return create_style({"flex": f"{child_flex}"})
+        return create_style(
+            {
+                "flex": f"{child_flex}",
+                "display": "flex",
+                "min-width": "0",
+                "min-height": "0",
+            }
+        )
 
     # If there are no child flexes, don't wrap them in an additional <div>
     if child_flexes is None:
--- a/tests/_plugins/stateless/test_flex.py
+++ b/tests/_plugins/stateless/test_flex.py
@@ -14,7 +14,7 @@ def test_vstack() -> None:
     result = vstack(["item1", "item2"], justify="center", heights=[1, 2])
     assert (
         result.text
-        == "<div style='display: flex;flex: 1;flex-direction: column;justify-content: center;align-items: normal;flex-wrap: nowrap;gap: 0.5rem'><div style='flex: 1'><span>item1</span></div><div style='flex: 2'><span>item2</span></div></div>"  # noqa: E501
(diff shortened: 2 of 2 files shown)


mscolnick:
Wow amazing, this actually fixes it. TIL. thank you!

image

mmlynarik:
Thanks for solving my issue!

ms-pablo-ruiz:
I think this actually brings a regression, the pic shows vstask and looks fine ,but what about hstack? I think this broke the stretching on vstacks when width is provided.

ms-pablo-ruiz:
@AhmadYasser1 , this PR introduced a regression, this made the displayed items inside the flex box a flex container even if they are not nested. 

It does fix the specific case described, but breaks if there is no stack nested, I would encourage you to fix it or revert it.

AhmadYasser1:
@ms-pablo-ruiz thanks for letting me know! @Light2Dark please let me know if there are any edits i need to be making

ms-pablo-ruiz:
@AhmadYasser1, @Light2Dark  already amended and fixed the regresion

PR #8223, opened 2026-02-08
fix: normalize Html subclasses in as_html for correct f-string formatting

### LINKED ISSUE #7931: `as_html` not working as expected for markdown
### Describe the bug

In the documentation about html elements ([here](https://docs.marimo.io/api/html/#marimo.as_html)), it is stated that `mo.as_html` converts any python value to an Html object you can use inside markdown or html.

But it does not seem to be the case when use in combination with `mo.md`. If you create a markdown element with `mo.md`, convert it to html with `mo.as_html` and try to use it inside a html template, the raw markdown content is rendered.

Note that this may be a limitation of the current implementation of the Html system, but the docs heavily imply that the example I gave is supposed to work. If it cannot be fixed, the docs should give an example of how to embed markdown into an html template.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.19.4",
  "editable": false,
  "location": "/var/mnt/data/Proj/marimo-website/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.18.5-200.fc43.x86_64",
  "Processor": "",
  "Python Version": "3.13.9",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.11.1"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10",
    "narwhals": "2.15.0",
    "packaging": "25.0",
    "psutil": "7.2.1",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "loro": "1.10.3",
    "openai": "2.15.0",
    "polars": "1.37.1",
    "ruff": "0.14.13",
    "ty": "0.0.12",
    "watchdog": "6.0.0"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

Copy-paste:

```py
x = mo.as_html(mo.md("# Title"))
mo.Html(
    f"""
    {x}
    """
)
```

AhmadYasser1:
## Summary
`f"{mo.as_html(mo.md('# Title'))}"` returned raw markdown instead of rendered HTML because `mo.md()` returns an `_md` subclass that overrides `__format__()`. Normalizes subclasses to the base `Html` type so f-string formatting works correctly.

Closes #7931

## Test Plan
- Added test verifying `as_html(mo.md(...))` returns base `Html` type

### DIFF
--- a/marimo/_output/formatting.py
+++ b/marimo/_output/formatting.py
@@ -304,6 +304,10 @@ def as_html(value: object) -> Html:
         ```
     """
     if isinstance(value, Html):
+        # Normalize Html subclasses (like _md) to plain Html so that
+        # __format__ returns processed HTML instead of raw source text.
+        if type(value) is not Html:
+            return Html(value.text)
         return value
 
     formatter = get_formatter(value)
--- a/tests/_output/formatters/test_formatters.py
+++ b/tests/_output/formatters/test_formatters.py
@@ -689,6 +689,21 @@ def test_as_html_with_html_object() -> None:
     assert result.text == "<h1>Hello</h1>"
 
 
+def test_as_html_with_md_object_formats_as_html() -> None:
+    """Test as_html normalizes md objects so __format__ returns HTML."""
+    from marimo._output.md import md
+
+    md_obj = md("# Title")
+    result = as_html(md_obj)
+
+    # Should be a plain Html, not the _md subclass
+    assert type(result).__name__ == "Html"
+    # When used in an f-string, should produce HTML, not raw markdown
+    formatted = f"{result}"
+    assert "# Title" not in formatted
+    assert "<" in formatted  # Should contain HTML tags
+
+
 def test_as_html_with_repr_html() -> None:
     """Test as_html with objects that have _repr_html_ method."""
     register_formatters()


mscolnick:
@Light2Dark i know you were looking at this before. what do you think of this approach?

Light2Dark:
I think this should be an improvement over the original behaviour. Tested it with the smoke tests

However, nested markdown and nested md in Html still doesn't work cleanly. You'd need to call `.text` or wrap with as_html

eg. https://github.com/marimo-team/marimo/issues/7931#issuecomment-3796977734
```python
solution = mo.md("**the solution**")

mo.md(rf"""
<details>
    <summary>
    Solution
    </summary>
    {solution}
</details>
""")
```

PR #8225, opened 2026-02-08
fix: prevent tooltips from being clipped inside containers

### LINKED ISSUE #3237: tooltip clipped inside container
### Describe the bug

Following the example on tooltips: https://docs.marimo.io/api/markdown/?h=tooltip#tooltips

I found that the tooltip becomes clipped inside the container. I need to manually add padding to the cell containing the tooltipped element.

This becomes a problem, in particular when using grid layout, where components can be placed freely. I'm not sure it's a bug, or simply a not very well implemented feature, but it is documented.

### Environment

{
  "marimo": "0.10.5",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.2",
  "Binaries": {
    "Browser": "129.0.6668.60",
    "Node": "v20.14.0"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.20.1",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.1",
    "markdown": "3.6",
    "narwhals": "1.14.0",
    "packaging": "24.0",
    "psutil": "5.9.8",
    "pygments": "2.17.2",
    "pyyaml": "6.0.1",
    "ruff": "0.5.2",
    "starlette": "0.37.2",
    "tomlkit": "0.12.4",
    "uvicorn": "0.29.0",
    "websockets": "11.0.3"
  },
  "Optional Dependencies": {
    "pandas": "2.2.2"
  }
}

### Code to reproduce

```python
import marimo

__generated_with = "0.10.5"
app = marimo.App(width="medium", layout_file="layouts/test_grid.grid.json")


@app.cell
def _():
    import marimo as mo
    return (mo,)


@app.cell
def _(mo):
    mo.ui.button(
        label='<div data-tooltip="This is a tooltip">Hover over me</div>'
    )
    return


if __name__ == "__main__":
    app.run()

AhmadYasser1:
## Summary
Tooltips were cut off when rendered inside a container with `overflow: hidden` because they were rendered inline instead of in a portal. Removes `usePortal={false}` so tooltips render in a portal attached to `document.body`.

Closes #3237

## Test Plan
- Visual: tooltips are no longer clipped by overflow containers

### DIFF
--- a/frontend/src/components/editor/cell/toolbar.tsx
+++ b/frontend/src/components/editor/cell/toolbar.tsx
@@ -59,11 +59,7 @@ export const ToolbarItem: React.FC<ToolbarItemProps> = ({
 
   if (tooltip) {
     return (
-      <Tooltip
-        content={tooltip}
-        side="top"
-        delayDuration={200}
-      >
+      <Tooltip content={tooltip} side="top" delayDuration={200}>
         {content}
       </Tooltip>
     );
--- a/frontend/src/components/editor/code/readonly-python-code.tsx
+++ b/frontend/src/components/editor/code/readonly-python-code.tsx
@@ -101,7 +101,7 @@ const CopyButton = (props: { text: string }) => {
   });
 
   return (
-    <Tooltip content="Copy code" >
+    <Tooltip content="Copy code">
       <Button onClick={copy} size="xs" className="py-0" variant="secondary">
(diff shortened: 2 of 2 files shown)


AhmadYasser1:
ah you're right, i was looking at the wrong tooltips. closing this.

PR #8227, opened 2026-02-08
feat: use variable name as download filename in dataframe viewer

### LINKED ISSUE #8095: dataframe download from rich dataframe viewer automatically names download file by variable name
### Description

I use marimo's rich dataframe viewer every day to validate and share validations with my team. I absolutely love the viewer and especially appreciate the ability to easily download a frame.

Image

I think assuming the name of the dataframe variable here (if possible) would be an incredible *nice touch*.

### Suggested solution

For example,
```
some_frame = pl.Dataframe()
```

When the above frame is downloaded through the rich viewer, instead of the default `download` or `download (1)`, `download (2)` I end up seeing in the filename, we see `some_frame`.

It is worth calling out that the DATA SOURCES feature on the side bar in the UI already appears to have both the variable name and frame content ;) .

Image

- [x] Yes

### Alternatives

_No response_

### Additional context

_No response_

AhmadYasser1:
## Summary
Downloaded files from the dataframe viewer used a generic `download.csv` name. Now threads the variable name from the plugin through to the download action so the file is named after the variable (e.g. `my_dataframe.csv`).

Closes #8095

## Test Plan
- Added test for download filename propagation

### DIFF
--- a/marimo/_utils/variable_name.py
+++ b/marimo/_utils/variable_name.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import inspect
-import logging
 from typing import Any
 
 from marimo import _loggers


Light2Dark (changes requested):
The donwloadFileName is not passed from the backend. Is that a follow-up?

AhmadYasser1:
Good catch — I've now added the backend side. The table's `__init__` infers the variable name via `inspect.currentframe()` (same pattern as `DataFramePlugin`) and passes it as `download-file-name` in the component args. Also added a test for it.

AhmadYasser1:
pushed fixes for the review feedback:

- moved the variable name inference before `add_selection_column()` so the identity check works correctly on the original data
- added `finally: del frame / del frame_back` to avoid reference cycles
- swapped the inline regex for the shared `Filenames.withoutExtension()` utility
- added `LOGGER.debug(...)` instead of bare `pass`

Light2Dark (commented):
.

marimo/_plugins/ui/_impl/table.py
```diff
@@ -502,6 +503,30 @@ def __init__(
...
+                        download_file_name = var_name
+                        break
+        except Exception:
+            LOGGER.debug(
+                "Failed to infer download file name from caller frame.",
+                exc_info=True,
+            )
+        finally:
+            del frame
+            del frame_back
```
Light2Dark: i would maybe refator this to a utility function, since `dataframe.py` uses it too

AhmadYasser1: good call, extracted it into `marimo/_utils/variable_name.py` — both `table.py` and `dataframe.py` now use `infer_variable_name(value, fallback)` instead of duplicating the frame-walking logic.

marimo/_utils/variable_name.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+import inspect
+import logging
+from typing import Any
+
+LOGGER = logging.getLogger(__name__)
```
Light2Dark: I don't know why I can't suggest code anymore, sigh GitHub UI
```python
from marimo import _loggers

```

Light2Dark (commented):
lgtm, just a small change for correctness

AhmadYasser1:
Thanks @mscolnick and @Light2Dark ! Appreciate the help and feedback :)

PR #8234, opened 2026-02-09
all mo.sql to use rf"""..."" instead of f"""..."""

### LINKED ISSUE #8179: Auto-convert Windows backslashes to forward slashes in SQL cell file paths
### Description

Pasting a Windows path (e.g., from Explorer's "Copy as path") into a SQL cell causes unicode escape errors. 

### Suggested solution

Marimo could automatically convert backslashes to forward slashes in SQL cell strings, or use raw strings by default for SQL cells ie. when converting to python: 
```python
mo.sql(rf"""...""")
```

- [ ] Yes

### Alternatives

Currently users must manually replace \ with / after pasting. Tedious for long paths...

### Additional context

_No response_

Light2Dark:
## 📝 Summary

Closes #8179. This supports path strings in queries because the f-string doesn't treat \ characters properly and will error out.

Taken from issue ^:
The rf prefix prevents:

5 types of parsing errors (\U, \u, \x, \N, invalid octal)
7 types of silent corruption (\t, \n, \r, \b, \f, \v, \a)

In my tests, there is no noticeable regression in queries. Complex queries like regex should be more correct. \x for example shouldn't be treated like a special character in a `mo.sql` query, it should be treated literally.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/chat/acp/__tests__/__snapshots__/prompt.test.ts.snap
+++ b/frontend/src/components/chat/acp/__tests__/__snapshots__/prompt.test.ts.snap
@@ -76,7 +76,7 @@ exports[`getAgentPrompt > should generate complete agent prompt with default fil
   </ui_elements>
 
   <sql>
-  - When writing duckdb, prefer using marimo's SQL cells, which start with df = mo.sql(f"""<your query>""") for DuckDB, or df = mo.sql(f"""<your query>""", engine=engine) for other SQL engines.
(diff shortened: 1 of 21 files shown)


mscolnick:
this needs to be backwards compatible, we can't just change to `r` strings as it likely will create issues with people who already have put in their own `\`. 

we can consider tagging as a breaking change if backwards compat is hard to maintain.

PR #8247, opened 2026-02-09
refactor: scope down exposed state from the runner, add explicit context

mscolnick:
Right now each hook got access to the runner. This is a bit leaky and makes it harder to test individual hook logic (we don't at all today). This PR refactors to create more scoped down context

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -777,18 +777,15 @@ def merge_config(
     # Patches for backward compatibility
     if "runtime" in merged:
         if (
-            merged["runtime"].get("auto_reload")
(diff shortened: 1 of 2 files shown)


marimo/_runtime/runner/hooks_post_execution.py
```diff
@@ -114,33 +108,20 @@ def _set_run_result_status(
         cell.set_run_result_status("success")
 
 
-def _should_broadcast_data() -> bool:
```
mscolnick, line 117: this was also called once per post_execution instead of called 3 times (before)

PR #8251, opened 2026-02-10
perf: better datastructures for kernel Runner

mscolnick:
Performance improvements to the marimo runtime hot paths, reducing algorithmic complexity in the cell execution loop, graph mutation, and module resolution.

- cells_to_run deque: Replace list.pop(0) (O(n) per pop) with collections.deque.popleft() (O(1)) in Runner and ScriptRunner, eliminating O(n²) total cost across a full run
- CancelledCells dataclass: Encapsulate cancellation tracking into a dedicated class with O(1) membership checks via a flat set, replacing O(k) scans over dict values on every iteration of the run loop
- remove_node targeted removal: Only iterate a node's direct neighbors instead of scanning all nodes in the topology, reducing deletion cost from O(total_cells) to O(neighbors)
- Pre-compute all_temporaries: Compute the union of all cell temporaries once per run and pass via PostExecutionHookContext, eliminating O(cells²) recomputation per run
- register_state_update early exit: Break after finding the first matching ref for a state update instead of continuing to scan all cells

### DIFF
--- a/marimo/_runtime/runner/hook_context.py
+++ b/marimo/_runtime/runner/hook_context.py
@@ -9,7 +9,7 @@ from marimo._messaging.errors import Error
 from marimo._types.ids import CellId_t
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping, Sequence
+    from collections.abc import Iterator, Mapping, Sequence
     from contextlib import AbstractContextManager
 
     from marimo._runtime.context.types import ExecutionContext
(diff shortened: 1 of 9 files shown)


marimo/_runtime/runner/hook_context.py
```diff
@@ -42,22 +74,18 @@ class PostExecutionHookContext:
     execution_context: ExecutionContextManager | None
     # Dict, because errors get mutated (formatted) by hooks.
     exceptions: dict[CellId_t, ExceptionOrError]
-    cells_cancelled: Mapping[CellId_t, set[CellId_t]]
+    cancelled_cells: CancelledCells
+    # Pre-computed union of all cell temporaries
+    all_temporaries: frozenset[str]
```
dmadisetti, line 79: This seems a little brittle and far from source. Maybe set in `__post_init__`? or cached property?

mscolnick: this is run at the start of execution and then never used again. it is just hoisted from the hook to not be run each time

dmadisetti: Derivative attributes that can be decoupled in the `dataclass` still seems like a code smell to me. A `.from(Runner)` would also be better

PR #8283, opened 2026-02-12
Support `__buffer__` objects in widget comm buffers

manzt:
Libraries like obstore expose a `Bytes` type that implements Python's buffer protocol but isn't a subclass of `bytes`, `memoryview`, or `bytearray`. When these objects ended up in widget comm buffers (e.g. from lonboard map tiles), msgspec didn't recognise them as native binary types and fell through to `enc_hook`, which produced invalid output instead of base64.

This adds `_ensure_bytes()` in the comm layer to coerce any buffer-protocol object to plain `bytes` before building model messages.

Needed to support raster tiles in `lonboard`:

### DIFF
--- a/marimo/_plugins/ui/_impl/comm.py
+++ b/marimo/_plugins/ui/_impl/comm.py
@@ -72,19 +72,18 @@ BufferType = Optional[list[Buffer]]
 
 
 def _ensure_bytes(buf: object) -> bytes:
-    """Coerce a buffer to a type msgspec can base64-encode.
+    """Coerce a buffer to plain ``bytes`` for msgspec serialization.
 
     msgspec natively handles ``bytes``, ``memoryview``, and ``bytearray``.
     Some libraries (e.g. obstore) use custom types that hold binary data
     but aren't subclasses of these, so msgspec can't serialize them directly.
+
+    ``bytes()`` handles memoryview/bytearray on all Python versions, and
+    on Python 3.12+ also handles any object implementing ``__buffer__``.
     """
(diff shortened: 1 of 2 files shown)


marimo/_plugins/ui/_impl/comm.py
```diff
@@ -67,17 +67,34 @@ def receive_comm_message(
 MsgCallback = Callable[[Msg], None]
 DataType = Optional[dict[str, Any]]
 MetadataType = Optional[dict[str, Any]]
-BufferType = Optional[list[bytes]]
+Buffer = bytes | memoryview | bytearray
```
kylebarron, line 70: Why not use `collections.abc.Buffer`? or to support multiple Python versions: https://github.com/developmentseed/obspec/blob/6c74dc0e10bff82c08c530af37dca96cf25c39fe/src/obspec/_put.py#L17-L20

marimo/_plugins/ui/_impl/comm.py
```diff
@@ -67,17 +67,34 @@ def receive_comm_message(
...
+BufferType = Optional[list[Buffer]]
+
+
+def _ensure_bytes(buf: object) -> bytes:
+    """Coerce a buffer to plain ``bytes`` for msgspec serialization.
```
kylebarron, line 75: Seems like that's the core issue, right? msgspec should handle buffers

manzt: Yeah exactly, we need upstream support but need some compatibility now.

PR #8299, opened 2026-02-13
Storage inspector experimental

Light2Dark:
## 📝 Summary

Adds storage inspector feature under a feature flag.

## 📋 Checklist


marimo/_data/_external_storage/models.py
```diff
@@ -82,6 +97,23 @@ async def get_entry(self, path: str) -> StorageEntry:
...
 
+    async def download_file(self, path: str) -> DownloadResult:
+        """Download the file at the given path with extracted metadata.
+
+        Calls download() and extracts the filename and extension from the path.
+        Subclasses can override this to customize filename extraction
+        (e.g., using content-disposition headers).
+        """
+        file_bytes = await self.download(path)
+        filename = path.rsplit("/", 1)[-1] or "download"
```
mscolnick, line 109: could the path by "\"?

marimo/_runtime/runtime.py
```diff
@@ -2642,6 +2656,123 @@ async def preview_datasource_connection(
...
+                    entries=[],
+                    namespace=request.namespace,
+                    prefix=request.prefix,
+                    error=error,
+                ),
+            )
+            return
+
+        try:
+            entries = backend.list_entries(
```
mscolnick: could this be async?

marimo/_messaging/notification.py
```diff
@@ -589,6 +592,46 @@ class StorageNamespacesNotification(Notification, tag="storage-namespaces"):
...
+    Attributes:
+        request_id: Request ID this responds to.
+        url: Virtual file URL to download from.
+        filename: Suggested filename for the download.
+        error: Error message if the download failed.
+    """
+
+    name: ClassVar[str] = "storage-download-ready"
+    request_id: RequestId
+    url: Optional[str] = None
```
mscolnick, line 630: it might be better to just pass the bytes and a mimetype, and construct a data-url on the frontend. have you thought through the tradeoffs? the pro i see is that you don't need a dependency on the VFile system

marimo/_server/api/router.py
```diff
@@ -63,6 +66,11 @@ def build_routes(base_url: str = "") -> list[BaseRoute]:
         datasources_router, prefix="/api/datasources", name="datasources"
     )
     app_router.include_router(sql_router, prefix="/api/sql", name="sql")
+    app_router.include_router(
+        external_storage_router,
+        prefix="/api/external-storage",
```
mscolnick: thoughts on just `/api/storage`? just a cleaner route and maybe lets use move the filesystem there as well

marimo/_messaging/notification.py
Light2Dark: Yeah, my reasoning was we don't want to pass back a large payload through the websocket channel, similar to downloading table data 🤔. Let the frontend/browser handle downloading it. I don't think it's a strong reason..

mscolnick: Ok sounds good. It might fail since not tied to a cell but I can't remember

marimo/_runtime/runtime.py
Light2Dark: good catch! missed this

marimo/_data/_external_storage/models.py
Light2Dark: it doesn't fail, downloads the root. Added a test case

marimo/_messaging/notification.py
mscolnick, line 630: i dont know if the file in sharememory will be cleaned up, but something to monitor. because taking out of GA, we may want to check and instead just pass the raw bytes
cc @dmadisetti

marimo/_runtime/virtual_file/virtual_file.py
```diff
@@ -79,6 +79,39 @@ def from_external_url(url: str) -> VirtualFile:
...
+        try:
+            ctx = get_context()
+        except ContextNotInitializedError:
+            return return_data_url()
+
+        if not ctx.virtual_files_supported:
+            return return_data_url()
+
+        vfile = VirtualFile(filename=vfile_name, buffer=buffer)
+        ctx.virtual_file_registry.add(vfile, ctx)
```
mscolnick, line 112: we may never remove this, so could lead to a leak

marimo/_data/_external_storage/models.py
```diff
@@ -82,11 +98,49 @@ async def get_entry(self, path: str) -> StorageEntry:
...
+        _, sep, suffix = filename.rpartition(".")
+        ext = suffix if sep and suffix else "bin"
+        return DownloadResult(
+            file_bytes=file_bytes,
```
mscolnick, line 113: another approach would would relieve some memory pressure in the marimo server and handle larger files (but also wont work for all cases), is creating a a signed url to hand back to the frontend:

https://developmentseed.org/obstore/latest/api/sign/#obstore.sign

(something to consider for a followup)

Light2Dark: Yeah.., maybe you are right to not rely on vfiles, and return bytes as a fallback.

PR #8307, opened 2026-02-13
Add rich HTML formatter for PyTorch `nn.Module`

manzt:
Adds a formatter that renders `nn.Module` instances as rich, collapsible HTML trees.

The formatter uses PyTorch's subpackage structure to map known modules to four display categories (Weight, Activation, Normalization, Regularization). Custom modules that follow PyTorch conventions work automatically, and unknown/containers default to "Other".

Keyword-heavy layers (Conv2d with 8+ args) get an expandable row that shows the full args truncated with CSS ellipsis, expandable to a line-by-line view. Positional args (like channel dimensions) stay always-visible next to the type pill for quick scanning. Frozen layers are visually dimmed and show "(frozen)" next to their param count, while partially-frozen containers show the trainable subset.

A hover legend in the footer explains the category color coding using mini swatch replicas of the actual type pills.

Hovering legend:

### DIFF
--- a/marimo/_output/formatters/pytorch_formatters.py
+++ b/marimo/_output/formatters/pytorch_formatters.py
@@ -509,7 +509,7 @@ def _walk(mod: torch.nn.Module, name: str = "") -> str:
                 body_parts.append(
                     '<div class="nn-t-expand-sep">'
                     '<span class="nn-t-expand-sep-label">tensor</span>'
-                    '</div>'
+                    "</div>"
                 )
             body_parts.append(
                 f'<span class="nn-t-key">dtype</span> {dtype_s}'


dmadisetti:
😍 

Checked in darkmode too:

image

koaning:
~Could we also show the number of trainable and non-trainable weights? This is relevant when you're keen to do some finetuning and you want to make sure that you're only taking a subset of all weights to train the bigger system.~

Nevermind, found it! 

CleanShot 2026-02-13 at 22 51 50@2x

Maybe the dtype of the layer too? Maybe what device (cpu/gpu) the element is on? 

The general feeling is that it is really sweet :)

manzt:
> Maybe the dtype of the layer too? Maybe what device (cpu/gpu) the element is on?

I added a "Tensor" section to the details section for each node. It's a little buried, but I'd like to understand how important that is to see at a glance before changing the information hierarchy or encoding differently. This information isn't included in the standard repr (since it just shows the static module parameters and these can be dynamic).

akshayka (approved):

Issue #8272, opened 2026-02-11
Notebook content deletion with sandboxed home


mark-alec:
### Describe the bug

With the latest release of marimo (0.19.10) - opening a notebook results in all cells being deleted! (was recoverable with real work using `git restore <FILE>` .

Does not appear to be a problem with regular sandbox or using a sandboxed home with 0.19.9.

Tested using `uvx --with 'marimo[sandbox]' marimo edit --sandbox ./`, and `uvx --with 'marimo[sandbox]==0.19.9' marimo edit --sandbox ./`

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.19.10",
  "editable": false,
  "location": "/Users/markkowarsky/.cache/uv/archive-v0/N56IgdlG6ZXEf8_NVfuEx/lib/python3.12/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "24.6.0",
  "Processor": "arm",
  "Python Version": "3.12.9",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "0.9.28 (Homebrew 2026-01-29)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.16.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.20.1",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3"
  },
  "Experimental Flags": {}
}
```

</details>

mark-alec:
Easy to reproduce. Just create a new notebook using the sandboxed home. Add any content (e.g. print("hello"). Shut down the notebook, then reopen it - after a brief moment the contents will be cleared.

Restarting the kernel from the hamburger menu has the same effect.

mscolnick:
Thanks for the reproduction and sorry you hit this issue. We will look into this.

PR #8309, opened 2026-02-13
fix: notebook cell_ids for MarimoConvert.to_notebook_v1()

jpopesculian:
In response to #8272

The problem was that notebooks created with `NotebookConvert.to_notebook_v1` were not adding cell ids so the mismatch caused the notebooks to clear the cells. Solution is to use the `CellIdGenerator`. I've added some tests to make sure the ids generated match the kernel

### DIFF
--- a/marimo/_convert/notebook.py
+++ b/marimo/_convert/notebook.py
@@ -1,6 +1,8 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


dmadisetti:
Thanks for revisiting. Would you mind merging this with #8310 ? I think they both require each other.

however, I'm a little skeptical on ordering. I think your initial fix was fine _if_ it was only confined to app mode. The issue was improper session blowing away the notebook content in edit mode

PR #8310, opened 2026-02-13
fix: initialize cell_ids on notebook creation

jpopesculian:
In response to https://github.com/marimo-team/marimo/issues/8272

The problem was that notebooks created with NotebookConvert.to_notebook_v1 were not adding cell ids so the mismatch caused the notebooks to clear the cells. Solution is to use the CellIdGenerator. I've added some tests to make sure the ids generated match the kernel

And reimplements the fix reverted by #8281

Bug fix: `cell_ids` was not initialized on the `SessionView` so cache
exports would sometimes be exported out of order. The fix was to emit an
`UpdateCellIdsNotification` on notebook creation

### DIFF
--- a/marimo/_pyodide/bootstrap.py
+++ b/marimo/_pyodide/bootstrap.py
@@ -54,6 +54,7 @@ def instantiate(
     session.put_control_request(
         CreateNotebookCommand(
             execution_requests=execution_requests,
+            cell_ids=tuple(app.cell_manager.cell_ids()),
             set_ui_element_value_request=UpdateUIElementCommand(
                 object_ids=[], values=[], request=None
             ),
--- a/marimo/_runtime/commands.py
+++ b/marimo/_runtime/commands.py
@@ -488,16 +488,16 @@ class CreateNotebookCommand(Command):
 
     Attributes:
         execution_requests: ExecuteCellCommand for each notebook cell.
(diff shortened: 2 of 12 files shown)


dmadisetti:
@jpopesculian following up on thoughts on squashing these together for #8309 (which should just be scoped to run)

jpopesculian:
Hi @dmadisetti sorry for the slow response :snail: I've rebased on top of #8309 and closed that one. I'm not sure I understand the request to limit the scope to app mode. The actual reason for this fix was that when serializing the notebook sessions in edit mode, because the cell_ids were not present, the serialization would sometimes happen out of order, and therefore would create an invalid session due to https://github.com/marimo-team/marimo/blob/main/marimo/_session/state/serialize.py#L91

jpopesculian:
@dmadisetti any update on this one?

dmadisetti:
> use the cell_ids were not present

I think this is fine though, a mismatch could be destructive, but I'll write some more unit tests to make sure we just don't populate the cache.

Sorry this fell off my radar too! BE CI seems flaky, and that may be on us. Pre-commit is normally pretty reliable and easy to pass for future PRs

jpopesculian:
@dmadisetti fixed the pre-commit and rebased on main

jpopesculian:
ah sorry i missed that PR! that's my bad!

PR #8311, opened 2026-02-13
feat: add the ability to control the cache

jpopesculian:
Reimplements the feature reverted by #8281
Goes best with #8310 to avoid invalid session caches

Add a new `serve_cached_sessions_in_apps` runtime configuration item
to use the cache in a read-only manner when running as an application

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -152,6 +152,8 @@ class RuntimeConfig(TypedDict):
         affected cells as stale, `"autorun"` automatically runs affected cells.
     - `output_max_bytes`: the maximum size in bytes of cell outputs; larger
         values may affect frontend performance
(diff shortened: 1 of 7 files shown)


dmadisetti:
Happy with this once https://github.com/marimo-team/marimo/pull/8310 passes. If you don't mind grabbing the pre-commit failure that would be great!

jpopesculian:
@dmadisetti my bad! my vim configuration was handling ruff weird. should be fixed. don't know about the rest of the tests

stubbi:
@dmadisetti is it good to go and be merged?

dmadisetti:
@stubbi thanks for the ping. Merged in #8310

Just manually tested and continually got cache misses- but looking into it right now

PR #8326, opened 2026-02-16
DataTable: add summary statistics for selected cells

### LINKED ISSUE #6015: Excel-like metrics with data table selection
### Description

Hi, now that we can select multiple cells to copy and paste their values. It would be nice to have a quick glimpse at the sum, average, count of the selection in a similar fashion that Excel is doing.

Thanks,

Adrien

### Suggested solution

Bottom display of metrics at the bottom (average, sum, count, n_unique?)

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

andrewsoonqn:
## 📝 Summary

Closes #6015

Introduce summary statistics for selected cells to the data table component. When users select two or more cells, the UI now displays the count, sum, and average of the selected _numeric_ values. 

image

Demo:

https://github.com/user-attachments/assets/de79ec0b-fa2c-4448-9e34-a8e782cf8099

Works also with other components where DataTable is used, e.g., dataframes.

https://github.com/user-attachments/assets/9fdc581c-949e-4e1b-adcb-c51c288dcfb8

## 🔍 Description of Changes

Decisions made:
* Non-numeric data is included in Count only, and not included in calculation of Sum and Average
	* Empty cells, infinity, NaN, object, boolean, empty or practically empty (e.g. "   ") strings are considered non-numeric 
	* Numeric strings are considered numeric, e.g. the entry below shows up as a number in the table
```
    {"name": "Alice", "age": 30, "city": "New York", "balance": "100"},
```
image

* The checkbox column to select rows is not included in any stats
* Sum and Average are shown only when the selection contains at least one numeric value.
* The `CellSelectionStats` component only shows when 2 or more cells are selected
* Have a layout shift and add a new bottom row (like what search does when clicked)
* Calculated stats are rounded to max 8 decimal points

Future considerations:
* I have not implemented a way to disable this display. One suggestion is that we can add a context menu that can be used to check and uncheck stats to toggle their visibility. 
* Ability to copy stats by clicking on them.
* More stats. 

### Implementation Details
* Add a new `CellSelectionStats` component that displays the count, sum, and average of selected cells when two or more are selected.
* Integrate `CellSelectionStats` into the `DataTableInternal` component, position is below the table. Wrap it in the `CellSelectionProvider` context.
* Add a `getNumericValuesFromSelectedCells` utility function in `utils.ts` to extract numeric values from selected cells, filtering out non-numeric and special columns.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/data-table.tsx
+++ b/frontend/src/components/data-table/data-table.tsx
@@ -324,10 +324,7 @@ const DataTableInternal = <TData,>({
             />
           </Table>
         </div>
-        <CellSelectionStats
-          table={table}
-          className="px-2 pt-1 mb-0.5 ml-auto"
-        />
+        <CellSelectionStats table={table} className="px-2 pt-1 ml-auto" />
       </CellSelectionProvider>
       <TableActions
         enableSearch={enableSearch}


Light2Dark:
This looks really cool! Some of the team may be out (& I'm unwell), but happy to review soon.

frontend/src/components/data-table/range-focus/cell-selection-stats.tsx
```diff
@@ -0,0 +1,80 @@
...
+  if (dataCellCount < 2) {
+    return null;
+  }
+
+  const numericValues = getNumericValuesFromSelectedCells(table, selectedCells);
+
+  return (
+    <div
+      className={cn(
+        "flex items-center justify-end gap-3 px-2 py-1 text-xs text-muted-foreground shrink-0 ml-auto",
```
Light2Dark: too much of a gap imo
```suggestion
        "flex items-center justify-end gap-3 px-2 pt-1 text-xs text-muted-foreground shrink-0 ml-auto",
```

frontend/src/components/data-table/range-focus/cell-selection-stats.tsx
```diff
@@ -0,0 +1,80 @@
...
+const StatSpan = (statName: string, statValue: number) => {
+  return (
+    <span>
+      {statName}: {statValue}
+    </span>
+  );
+};
+
+const CountStat = ({ selectedCells }: { selectedCells: SelectedCells }) => {
+  const count = countDataCellsInSelection(selectedCells);
```
Light2Dark: it looks like we don't need to count again for this stat, we already counted at the top.

frontend/src/components/data-table/range-focus/utils.ts
```diff
@@ -43,6 +43,69 @@ export function getTabSeparatedValues(values: string[][]) {
...
+}
+
+/**
+ * Extract numeric values from the selected cells. Only finite numbers and
+ * non-empty numeric strings (e.g. "42", "3.14", "0") are included. Skips select
+ * checkbox column, missing cells, and all other types (boolean, null, etc.).
+ */
+export function getNumericValuesFromSelectedCells<TData>(
+  table: Table<TData>,
+  selectedCellIds: Set<string>,
```
Light2Dark, line 64: maybe we can pass the valid cells instead of selected cells? since we are filtering them at the top already (removed select_col_id)

frontend/src/components/data-table/range-focus/utils.ts
```diff
@@ -43,6 +43,69 @@ export function getTabSeparatedValues(values: string[][]) {
...
+    if (cellId.includes(SELECT_COLUMN_ID)) {
+      continue;
+    }
+    const rowId = cellId.split("_")[0];
+    const row = table.getRow(rowId);
+    if (!row) {
+      continue;
+    }
+
+    const tableCell = row.getAllCells().find((c) => c.id === cellId);
```
Light2Dark: I think this may be slightly unoptimized (I see it in old code too, and maybe we can replace it). The code below uses row.getValue instead of iterating through all the cells in a row.
```tsx
export function getNumericValuesFromSelectedCells<TData>(
  table: Table<TData>,
  selectedCellIds: Set<string>,
): number[] {
  const numericValues: number[] = [];
  for (const cellId of selectedCellIds) {
    if (cellId.includes(SELECT_COLUMN_ID)) {
      continue;
    }
    const { rowId, columnId } = getRowAndColumnId(cellId);
    const row = table.getRow(rowId);
    if (!row) {
      continue;
    }

    const value = row.getValue(columnId);
    let num: number;
    if (typeof value === "number") {
      num = value;
    } else if (typeof value === "string") {
      if (value.trim() === "") {
        continue;
      }
      num = Number(value);
    } else {
      continue;
    }

    if (Number.isFinite(num)) {
      numericValues.push(num);
    }
  }
  return numericValues;
}

// at the end of the file
function getRowAndColumnId(cellId: string) {
  const sepIdx = cellId.indexOf("_");
  const rowId = cellId.slice(0, sepIdx);
  const columnId = cellId.slice(sepIdx + 1);
  return { rowId, columnId };
}
```

frontend/src/components/data-table/range-focus/cell-selection-stats.tsx
andrewsoonqn: Alright, will fix

frontend/src/components/data-table/range-focus/utils.ts
andrewsoonqn: Ok, i'll take a look

frontend/src/components/data-table/range-focus/utils.ts
andrewsoonqn: I'll fix this along with the old code as well?

Light2Dark: Yeah, that would be good, you may need to update the tests as well.

frontend/src/components/data-table/range-focus/utils.ts
andrewsoonqn: @Light2Dark I'm abit unclear on what you mean here that we are already filtering the cells. Current flow is `const selectedCells = useAtomValue(selectedCellsAtom);` which gets unfiltered cells, then passes it to `getNumericValuesFromSelectedCells` which does the filtering of `SELECT_COLUMN_ID`.

Right now `getNumericValuesFromSelectedCells` is like an alternative for the existing `getSelectedCells`, except it only includes numeric values. So the inputs are similar to `getSelectedCells` taking in a set of selected cells' ids (unfiltered for `SELECT_COLUMN_ID` yet)

andrewsoonqn:
> curious if you tried the other approaches (remove pagination / put in the middle at the bottom of the table)? How did it feel?

@Light2Dark 

- Feels like putting in the middle will make the bottom a bit crowded.
- Replacing pagination will make the user unable to navigate to other pages while having cells selected. This might be a minor inconvenience for some workflows. User needs to tap and unselect before they can switch pages. 
- Personal opinion is that it looks better as is. It feels more visually connected to the table, which makes sense given it's displaying stats about the table's data.
- Feels like having it expand like this makes it more extensible in the future (more space)

frontend/src/components/data-table/range-focus/utils.ts
andrewsoonqn: If we use this approach, it seems like there is no easy way to distinguish between missing cells and cells with value `undefined`. `row.getValue(columnId)` might return `undefined` both when the cell itself has value `undefined`. The tanstack table row also does not expose any methods like `contains(columnId)`. 

Do you have any suggestions? Was wondering whether a cell's value being `undefined` is possible in the first place, or will only empty cells have value of `undefined`?

Light2Dark: Yeah, I think it's impossible to be undefined in the first place. I assume it would be undefined if columnId is incorrect, which shouldn't happen. I believe the test checks for this impossible scenario?

frontend/src/components/data-table/range-focus/utils.ts
Light2Dark: The cells are being filtered in `const dataCellCount = countDataCellsInSelection(selectedCells);`.  But I think it's fine, sorry, just leave it in for now. We can optimize later

Light2Dark (approved):
I think it looks good! Thank you :) The test failure seems unrelated.

frontend/src/components/data-table/range-focus/utils.ts
andrewsoonqn: @Light2Dark Sorry havent got a chance to look at this yet. But last I checked, there was still some tests that failed in `src/components/data-table/range-focus/__tests__/utils.test.ts` related to null and undefined values. Will probably check those out by tomorrow.

Light2Dark: ah no worries, have fixed it for you. There were other CI failures as well so I jumped in.

PR #8342, opened 2026-02-17
feat: mo.ui.matplotlib()

akshayka:
This PR adds a new public API, `mo.ui.matplotlib(axis: plt.Axes, *, debounce: bool=False.)`, which adds reactive selection to a matplotlib Axes. This API is designed for scatter plots and scatter plot selections, but should be extensible enough to support other plot types and interactions (such as span selections)/geometries in the future.

This feature was inspired by and based on @koaning's `Wigglystuff` `ChartSelect` anywidget for matplotlib. 

## Selections

Two types of selections are supported: box and lasso. Box is the default selection, with <kbd>Shift</kbd>+click triggering lasso selection.

## Python usage

```python
import marimo as mo
import numpy as np

x = np.arange(5)
y = x**2
plt.scatter(x=x, y=y)
fig = mo.ui.matplotlib(plt.gca())
fig
```

```python
# Filter data using the selection
mask = fig.value.get_mask(x, y)
selected_x, selected_y = x[mask], y[mask]
```

If `debounce=True` is passed to the constructor, data is only sent back on mouse up.

## `value` type

The element's `value` is a frozen dataclass containing information about the interaction and selection. Each dataclass exposes a `get_mask(x: np.typing.ArrayLike, y: np.typing.ArrayLike)` method which returns a boolean mask for indexing into the scattered data.

An empty selection returns an `EmptySelection` sentinel object which is `False-y`, so users can write code like

```python
if fig.value:
  # do something with selection
  ...
```

`EmptySelection.get_mask(x, y)` does return an all-`False` mask, so users can index into the original data without checking if the selection is empty.

The dataclass type is not returned as part of the public API. We could extend the class of interactions/geometries handled by adding new classes in the future. 

## Smoke test

This PR includes a smoke test that exercises linear and log scale axes

## Media

https://github.com/user-attachments/assets/e085f4d0-d17b-4d44-854c-7c3f2f156b4d


marimo/_plugins/ui/_impl/mpl.py
```diff
@@ -0,0 +1,271 @@
...
+        """
+        import numpy as np
+
+        return np.zeros(len(np.asarray(x)), dtype=bool)
+
+    def __bool__(self) -> bool:
+        return False
+
+
+MatplotlibSelection = Union[BoxSelection, LassoSelection, EmptySelection]
```
manzt: Might be overkill, but we could make `MatplotlibSelection` a `typing.Protocol` if we anticipate other selection types in the future:

```python
class MatplotlibSelection(Protocol):
    def get_mask(self, x: ArrayLike, y: ArrayLike) -> NDArray[np.bool_]: ...
```

Also, I'm not totally sure on my numpy typing, but would it be richer type information to use `np.typing.NDArray[np.bool_]` instead of `np.ndarray` for the `get_mask` return types?

marimo/_plugins/ui/_impl/mpl.py
```diff
@@ -0,0 +1,271 @@
...
+    Attributes:
+        value: The selected data, with `get_mask(x, y)` returning a
+            mask array corresponding to the selection.
+
+    Args:
+        axes: A matplotlib `Axes` object. The full figure is rendered,
+            but selections map to this axes' coordinate space.
+        debounce: If `True`, the selection is only sent to Python on
+            mouse-up. If `False` (the default), it streams while dragging.
+        on_change: Optional callback invoked when the selection changes.
```
manzt: I "get" this  API but couldn't find an example of it being used. Is there a use case you had in mind? My suggestion would be to be open to extending `mo.ui.matplotlib` with this later unless we already have a motivation example/use case.

frontend/src/plugins/impl/matplotlib/MatplotlibPlugin.tsx
```diff
@@ -0,0 +1,708 @@
...
+      xBounds,
+      yBounds,
+      xScale,
+      yScale,
+      selectionColor,
+      selectionOpacity,
+      strokeWidth,
+      debounce,
+      setValue,
+    };
```
manzt: Not exactly blocking, but it really seems like we're fighting React in this implementation.

All the `useCallback(() => ..., [])` + `propsRef` + eslint-disables exist to prevent React from re-rendering during prop changes. But since it's just a canvas, we don't need React for rendering at all... and we have to remember to avoid taking a dependency in those callbacks / disabling linting rules to keep React "off."

We could "reactify" this and let React handle the rendering, but my suggestion would be to just move the logic into an imperative class and use React only to manage the lifecycle:

```tsx
const MatplotlibComponent = (props: MatplotlibComponentProps) => {
  const ref = useRef<HTMLDivElement>(null);
  const instance = useRef<MatplotlibCanvas | null>(null);

  useEffect(() => {
    instance.current = new MatplotlibCanvas(ref.current!, props, props.setValue);
    return () => instance.current?.destroy();
  }, []);

  // Runs every render. Syncs fresh props without unmounting
  useEffect(() => {
    instance.current?.update(props, props.setValue);
  });

  return <div ref={ref} />;
};
```

The class would look something like:

```ts
class MatplotlibCanvas {
  #canvas: HTMLCanvasElement;
  #props: Props;
  #state: InteractionState = { mode: "idle", boxStart: null, ... };

  constructor(container: HTMLElement, props: Props, onChange: (v: T) => void) {
    this.#props = props;
    this.#canvas = document.createElement("canvas");
    container.appendChild(this.#canvas);
    this.#canvas.addEventListener("pointerdown", this.#onPointerDown);
    this.#canvas.addEventListener("pointermove", this.#onPointerMove);
    this.#canvas.addEventListener("pointerup", this.#onPointerUp);
  }

  // Auto-bound, stable identity, reads this.#props (always fresh)
  #onPointerDown = (e: PointerEvent) => {
    this.#canvas.setPointerCapture(e.pointerId);
    const pt = this.#clampToAxes(this.#getCanvasPoint(e));
    // ... just this.#props and this.#state, no stale closures
  };

  update(props: Props, onChange: (v: T) => void) {
    this.#props = props;
    this.#draw();
  }

  destroy() {
    this.#canvas.removeEventListener("pointerdown", this.#onPointerDown);
    // ...
  }
}
```

manzt (commented):
Really love the API. Just a few comments.

frontend/src/plugins/impl/matplotlib/MatplotlibPlugin.tsx
manzt: I made https://github.com/marimo-team/marimo/pull/8358 against this branch if we want to go this direction.

marimo/_plugins/ui/_impl/mpl.py
```diff
@@ -0,0 +1,271 @@
...
+            args={
+                "chart-base64": _figure_to_base64(figure),
+                "x-bounds": list(axes.get_xlim()),
+                "y-bounds": list(axes.get_ylim()),
+                "axes-pixel-bounds": axes_pixel_bounds,
+                "width": fig_width_px,
+                "height": fig_height_px,
+                "debounce": debounce,
+                "x-scale": axes.get_xscale(),
+                "y-scale": axes.get_yscale(),
```
manzt: matplotlib's built-in scale types from `get_xscale()` are: "linear", "log", "symlog", "logit", "asinh", and "function" (plus any custom registered scales)... 

On the frontend we only handle "log" and fallback to linear... how should we handle that?

manzt: see https://github.com/marimo-team/marimo/pull/8363

marimo/_plugins/ui/_impl/mpl.py
akshayka: Removed

marimo/_plugins/ui/_impl/mpl.py
akshayka: Done on both accounts. Good catch on the return type, thank you

docs/guides/working_with_data/plotting.md
```diff
@@ -4,29 +4,90 @@ marimo supports most major plotting libraries, including Matplotlib, Seaborn,
...
+```python
+import matplotlib.pyplot as plt
+import marimo as mo
+import numpy as np
+
+x = np.random.randn(500)
+y = np.random.randn(500)
+plt.scatter(x, y)
+# Wrap the Axes in mo.ui.matplotlib to make them reactive ⚡
+fig = mo.ui.matplotlib(plt.gca())
```
dmadisetti: gca() returns an axes https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.gca.html

we might want to think about calling this `ax`

docs/guides/working_with_data/plotting.md
```diff
@@ -4,29 +4,90 @@ marimo supports most major plotting libraries, including Matplotlib, Seaborn,
...
+|------|------|------------|
+| `EmptySelection` | Nothing selected (falsy) | — |
+| `BoxSelection` | Box drag | `x_min`, `x_max`, `y_min`, `y_max` |
+| `LassoSelection` | <kbd>Shift</kbd>+drag | `vertices` (tuple of `(x, y)` pairs) |
+
+All three have a `get_mask(x, y)` method that returns a boolean NumPy array,
+so you can always write:
+
+```python
+mask = fig.value.get_mask(x, y)
```
dmadisetti: Cool! Seen the demos and they look awesome. I guess the limitation of `get_mask` vs `.value` in plotly and altair is that there's no way of extracting the rasterized values from the plot itself?

An alternative API is just to wrap the axes object:

```python
ax = mo.ui.matplotlib(plt.gca())
ax.plot(x, y) # Plot directly on the axes object, pass through
ax.set_xlabel('X Value')
ax.set_ylabel('Y Value')
ax.set_title('Sine Wave Plot')
ax.grid(True)
ax

# ax.value == df of selected like plotly, altair
```

this does fall apart for multiple axes in a figure- but an extension of the api could just be mirroring `subplots` https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplots.html

so 

```python
fig, (ax0, ax1) = mo.ui.matplotlib(1, 2, sharey=True)

ax0.plot(x, y) # Plot directly on the axes object
...
fig # Would show both interactive axes
```

This would also limit the risk of someone passing in something like a `plt.figure().add_subplot(projection='3d')`

but this maybe seems a bit over the top, and also bifurcates from the plotly/ altair wrappers in different ways.

docs/guides/working_with_data/plotting.md
akshayka: good call, done

docs/guides/working_with_data/plotting.md
akshayka: I'm concerned about the API surface area for something like this, and I think I prefer something more minimal and explicit. It is an interesting idea though

dmadisetti:
Failing test is relevant but lgtm otherwise

PR #8343, opened 2026-02-17
Add interactive doc lookup for `data-marimo-doc` in cell output

manzt:
Any HTML formatter can now annotate elements with a `data-marimo-doc` attribute to get hover-to-docs behavior in the Documentation panel:

```html
    <div data-marimo-doc="torch.nn.Linear">
      <span class="nn-t-type">Linear</span> ...
    </div>
```

When the user hovers over an element carrying this attribute, the frontend fires a doc lookup request through the existing autocomplete system (no new API endpoints). Jedi resolves the qualified name and the result populates the Documentation panel, the same one used for live docs when the cursor is on a symbol in the editor.

The first consumer is the PyTorch `nn.Module` formatter, where hovering a layer row shows the docstring for that module class. 

But the mechanism is generic: any formatter that emits HTML with `data-marimo-doc="some.qualified.Name"` gets the same behavior for free.

https://github.com/user-attachments/assets/59297de6-93c5-4b7a-a724-0a4687c3d6b0

### DIFF
--- a/frontend/src/core/documentation/DocHoverTarget.tsx
+++ b/frontend/src/core/documentation/DocHoverTarget.tsx
@@ -15,7 +15,7 @@ export const DocHoverTarget: React.FC<{
   return (
     <span
       onMouseEnter={handleMouseEnter}
-      onMouseLeave={handleMouseEnter.cancel}
+      onMouseLeave={() => handleMouseEnter.cancel()}
     >
       {children}
     </span>


frontend/src/plugins/impl/anywidget/model.ts
```diff
@@ -397,11 +397,10 @@ export async function handleWidgetMessage(
       return;
     }
 
-    case "close": {
+    case "close":
       BINDING_MANAGER.destroy(modelId);
       modelManager.delete(modelId); // aborts the model's signal, clearing listeners
       return;
-    }
```
akshayka, line 404: Was this intentional? Or lint/format?

akshayka (approved):

frontend/src/plugins/impl/anywidget/model.ts
manzt: hmm this came from running the linter `pnpm lint`. I thought it would fix linting CI... but that still seems to be failing. I'll revert and figure it out.

PR #8345, opened 2026-02-17
Mask `getpass.getpass()` input in notebook UI

### LINKED ISSUE #8338: Usage of dots instead of plain text in getpass.getpass()
### Description

In the process of creating a marimo notebook, I came across the problem that password input is echoed in plain text. 

Image

I think this is well known, since it's also shown as a warning.
Would it be possible to get an input field, which obfuscates the entered text like in a jupyter notebook or in VSCode with the normal dots?

### Suggested solution

Don't echo the entered text, but rather just dots.

- [ ] Yes

### Alternatives

Perhaps it would be even better to not give any feedback in the input box.

### Additional context

I was looking for any issue related to this, but I couldn't find one also not in the closed ones.

manzt:
Fixes #8338

Override getpass.getpass to route through marimo's stdin with a "text/password" mimetype so the frontend renders a native <input type="password"> instead of showing plaintext.

> [!NOTE]
> These changes require updating the API types (codegen).

I used `mimetype="text/password"` on the existing `ConsoleMsg` to signal password mode to the frontend. The alternative is adding a `password: bool` field to the message, but that introduces an invalid state (`password=True` with a non-text mimetype). The mimetype approach avoids that.

`text/password` isn't a real IANA mimetype — it's purely internal to our WebSocket protocol.

### DIFF
--- a/marimo/_messaging/mimetypes.py
+++ b/marimo/_messaging/mimetypes.py
@@ -4,8 +4,12 @@ from __future__ import annotations
 from typing import Any, Literal, Union
 
 # It is convenient to write mimetypes as strings,
-# but can lead to typos. This literal type
-# helps us avoid typos.
+# but can lead to typos. This literal type helps
+# us avoid typos.
+#
+# While nonstandard MIME types are often given an
+# `x-` prefix our non IAIA types to not follow
+# this convention (e.g., text/password, text/latex).
 KnownMimeType = Literal[
     "application/json",
     "application/vnd.marimo+error",


frontend/src/components/editor/output/console/ConsoleOutput.tsx
```diff
@@ -167,13 +167,15 @@ const ConsoleOutputInternal = (props: Props): React.ReactNode => {
             );
 
             const originalIdx = consoleOutputs.length - idx - 1;
+            const isPassword = output.mimetype === "text/password";
```
dmadisetti, line 170: Just an open question on mimetypes, are we ok handling non-standard ones?

 - `text/vnd.marimo.password`
 - `text/x-password`

Are a bit more conventional. But open and want to hear your thoughts

manzt: Yeah, sorry updated the PR description with a comment on this just a moment ago. I'd be for `text/vnd.marimo.password` or `text/x-password`.

manzt: Probably a preference for `text/x-password`.

dmadisetti: `text/markdown` is official, but `text/latex` is not. Since it's just internal and there's precedent, `text/password` seems totally fine to me and consistent. Maybe we just leave a comment in source though so we don't revisit

Sorry for the churn

PR #8315, opened 2026-02-14
Normalize math rendering in docs panel and LSP tooltips

LiquidGunay:
## 📝 Summary
This PR fixes math rendering consistency for documentation shown in both the docs explorer and LSP tooltips by normalizing additional math syntaxes in the backend markdown path and frontend LSP docs path.

## 🔍 Description of Changes
- Extended backend markdown preprocessing in DisplayMathPreprocessor to normalize:
  - .. math::
  - :math:\...``
  - :math:<code>...</code>
  - \(...\)
  - \[...\]

while preserving existing $$...$$ spacing behavior.
- Kept backend docstring conversion path thin in completion flow so math docs reliably go through markdown rendering.
- Added frontend LSP normalization utility and applied it to:
  -   hover
  -   completion item docs
  -   completion resolve docs
  -   signature help docs

- Preserved fenced code and inline code literals (no unwanted math conversion inside code examples).
- Added concise comments for regex patterns to clarify exactly what each pattern matches.
### Tests
- Python:
  - test_md.py
  - test_complete.py
- Frontend:
  - normalize-markdown-math.test.ts
  - notebook-lsp.test.ts

### Screenshot
math-rendering-after
math-rendering-after_1

## 📋 Checklist
- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] Tests have been added for the changes made.
- [ ] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Pull request title is a good summary of the changes - it will be used in the [release notes](https://github.com/marimo-team/marimo/releases).

PR #8348, opened 2026-02-17
fix: display math rendering

akshayka:
Fix `$$...$$` display math rendering: do not require surrounding blank lines

The pymdownx.arithmatex block processor requires `$$...$$` to be in its
own markdown block (between blank lines) to recognize it as display
math. Without blank lines, it falls back to inline delimiters or fails.

Add a `DisplayMathPreprocessor` that inserts blank lines around `$$...$$`
blocks before arithmatex processes them, so users don't need to
manually add blank lines. The preprocessor runs at priority 24 (after
superfences at 25) to avoid modifying `$$` inside code fences.

This is something I've hit many times translating Jupyter notebooks to marimo notebooks.

### DIFF
--- a/tests/_output/test_md.py
+++ b/tests/_output/test_md.py
@@ -686,9 +686,7 @@ def test_md_display_math_without_blank_lines() -> None:
     assert "||(" not in result_end
 
     # Inline $...$ should NOT be affected
-    result_inline = _md(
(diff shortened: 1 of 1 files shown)


dmadisetti (commented):
https://github.com/marimo-team/marimo/pull/8315 might be a tiny bit more comprehensive, but this is much cleaner

PR #8351, opened 2026-02-17
fix: merge new panels into saved localStorage layout on upgrade

### LINKED ISSUE #8350: Saved panel layout in localStorage prevents new panels from appearing on upgrade
### Describe the bug

When marimo introduces a new panel (e.g., Terminal in the developer panel), existing users who have a previously saved `marimo:panel-layout` in `localStorage` will never see the new panel. The saved layout is loaded verbatim via `atomWithStorage` with no merge against the current set of registered panels, so any panel added after the layout was first persisted is permanently invisible.

This affects any long-lived deployment where users upgrade marimo in place — the browser retains the stale layout across server upgrades since `localStorage` is scoped to the origin, not the server image.

### Root cause

In [`frontend/src/components/editor/chrome/state.ts`](https://github.com/marimo-team/marimo/blob/main/frontend/src/components/editor/chrome/state.ts), the panel layout atom is initialized as:

```ts
panelLayoutAtom = atomWithStorage<PanelLayout>(
  "marimo:panel-layout",
  DEFAULT_PANEL_LAYOUT,
  jotaiJsonStorage,
  { getOnInit: true }
)
```

`DEFAULT_PANEL_LAYOUT` is built from the `PANELS` array in [`types.ts`](https://github.com/marimo-team/marimo/blob/main/frontend/src/components/editor/chrome/types.ts), which includes all current panels (including Terminal). However, `atomWithStorage` returns the saved value from `localStorage` when it exists, and the default is only used when no saved value is present.

`resolvePanelLocation()` determines where a panel renders:

```ts
if (layout.sidebar.includes(panelType)) return "sidebar"
if (layout.developerPanel.includes(panelType)) return "developer-panel"
return null
```

If a panel type is not in either saved array, it returns `null` and the panel is never rendered — regardless of server-reported capabilities or the `PANELS` definition.

There is no reconciliation step that diffs the saved layout against the current `PANELS` registry and appends missing entries.

### Workaround

```js
// Add missing panel
const layout = JSON.parse(localStorage.getItem('marimo:panel-layout'));
layout.developerPanel.push('terminal');
localStorage.setItem('marimo:panel-layout', JSON.stringify(layout));
location.reload();

// Or reset to defaults
localStorage.removeItem('marimo:panel-layout');
```

### Suggested fix

When hydrating `panelLayoutAtom`, diff the saved layout against the current set of non-hidden panels from `PANELS`. Append any panel present in `DEFAULT_PANEL_LAYOUT` but absent from the saved layout to its `defaultSection` array. This preserves user ordering while ensuring new panels are discoverable on upgrade. Something like:

```ts
function mergeWithDefaults(saved: PanelLayout, defaults: PanelLayout): PanelLayout {
  const allSaved = new Set([...saved.sidebar, ...saved.developerPanel]);
  return {
    sidebar: [...saved.sidebar, ...defaults.sidebar.filter(p => !allSaved.has(p))],
    developerPanel: [...saved.developerPanel, ...defaults.developerPanel.filter(p => !allSaved.has(p))],
  };
}
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
marimo: 0.19.11
Python: 3.12.3
OS: Linux x86_64 (also reproduced on macOS with Firefox 147 and Chrome)
```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "marimo",
# ]
# ///
#
# Reproduction does not require running a notebook.
# This is a browser localStorage issue reproducible with the following steps:
#
# 1. Run: marimo edit
# 2. Confirm Terminal tab is visible in the developer panel (bottom bar)
# 3. Open browser console and simulate a pre-terminal-era saved layout:
#
#    localStorage.setItem('marimo:panel-layout', JSON.stringify({
#      sidebar: ["files","variables","packages","ai","outline","documentation","dependencies"],
#      developerPanel: ["errors","scratchpad","tracing","secrets","snippets","logs"]
#    }));
#
# 4. Refresh the page
# 5. Terminal tab is gone — even though the server reports capabilities.terminal = true
# 6. Fix: localStorage.removeItem('marimo:panel-layout') and refresh
#
# This affects any user who saved a panel layout before Terminal was added
# and then upgraded marimo. Common in persistent environments like
# SageMaker Studio where the browser origin stays the same across image upgrades.
```

jacobcbeaudin:
## Problem

When marimo adds a new panel (e.g. Terminal in the developer panel), existing users with a previously saved `marimo:panel-layout` in `localStorage` never see it. The saved layout is loaded verbatim via `atomWithStorage` with no merge against the current set of registered panels, so any panel added after the layout was first persisted is permanently invisible.

This affects long-lived deployments where users upgrade marimo in place. The browser retains the stale layout across server upgrades since `localStorage` is scoped to the origin, not the server image. We hit this on SageMaker Studio where the browser origin persists across container image upgrades.

### Root cause

In `state.ts`, `panelLayoutAtom` uses `jotaiJsonStorage` which returns the saved value from `localStorage` as-is. `DEFAULT_PANEL_LAYOUT` is only used when no saved value exists:

```ts
export const panelLayoutAtom = atomWithStorage<PanelLayout>(
  "marimo:panel-layout",
  DEFAULT_PANEL_LAYOUT,
  jotaiJsonStorage,       // returns raw saved value, no merge
  { getOnInit: true },
);
```

`resolvePanelLocation()` returns `null` for any panel absent from both saved arrays, so the panel is never rendered regardless of server capabilities or the `PANELS` definition.

## Solution

Replace `jotaiJsonStorage` with a custom storage adapter (using the existing `adaptForLocalStorage` utility) that merges saved layouts with `DEFAULT_PANEL_LAYOUT` on read. Panels present in defaults but absent from the saved layout are appended to their default section. Panels the user moved between sections are left in place.

```ts
function mergePanelLayout(saved: PanelLayout): PanelLayout {
  const allSaved = new Set([...saved.sidebar, ...saved.developerPanel]);
  return {
    sidebar: [
      ...saved.sidebar,
      ...DEFAULT_PANEL_LAYOUT.sidebar.filter((p) => !allSaved.has(p)),
    ],
    developerPanel: [
      ...saved.developerPanel,
      ...DEFAULT_PANEL_LAYOUT.developerPanel.filter((p) => !allSaved.has(p)),
    ],
  };
}
```

### Changes

- **`frontend/src/components/editor/chrome/state.ts`**: Add `mergePanelLayout()` and wire it through `adaptForLocalStorage` as the storage adapter for `panelLayoutAtom`.
- **`frontend/src/components/editor/chrome/__tests__/state.test.ts`** (new): 8 tests covering the merge function and an integration test exercising the full storage read path with a mock `localStorage`.

### What this preserves

- Custom panel ordering
- Panels moved between sidebar and developer panel (checked via `allSaved` across both sections)
- Existing behavior when no saved layout exists (falls through to `DEFAULT_PANEL_LAYOUT`)
- Existing behavior when the saved layout is already current (no-op)

## Reproduction

```js
// 1. Run: marimo edit
// 2. Simulate a pre-terminal-era saved layout in the browser console:
localStorage.setItem('marimo:panel-layout', JSON.stringify({
  sidebar: ["files","variables","packages","ai","outline","documentation","dependencies"],
  developerPanel: ["errors","scratchpad","tracing","secrets","snippets","logs"]
}));
// 3. Refresh. Terminal tab is gone.
// 4. With this fix applied, refresh. Terminal tab appears at the end of the developer panel.
```

## Test plan

- [x] Unit tests for `mergePanelLayout`: missing sidebar panels, missing developer panels, no duplication, cross-section moves, empty saved layout and exact stale layout scenario
- [x] Integration test: mock `localStorage` with stale layout through `adaptForLocalStorage` + `mergePanelLayout`, assert "terminal" present
- [x] Integration test: empty `localStorage` returns `initialValue` default
- [x] `pnpm biome check` clean
- [x] `pnpm typecheck` no errors in changed files
- [x] CI

Closes #8350

### DIFF
--- a/frontend/src/components/editor/chrome/__tests__/state.test.ts
+++ b/frontend/src/components/editor/chrome/__tests__/state.test.ts
@@ -30,11 +30,26 @@ describe("mergePanelLayout", () => {
         "logs",
       ],
     };
-    const result = mergePanelLayout(saved);
-    expect(result.developerPanel).toContain("terminal");
-    // Existing user ordering is preserved at the front
-    expect(result.developerPanel.slice(0, 6)).toEqual(saved.developerPanel);
-    expect(result.sidebar).toEqual(saved.sidebar);
+    expect(mergePanelLayout(saved)).toEqual({
+      sidebar: [
+        "files",
+        "variables",
+        "packages",
+        "ai",
+        "outline",
+        "documentation",
+        "dependencies",
+      ],
+      developerPanel: [
+        "errors",
+        "scratchpad",
+        "tracing",
+        "secrets",
+        "snippets",
+        "logs",
+        "terminal",
+      ],
+    });
   });
 
   it("appends new sidebar panels missing from saved layout", () => {
@@ -50,15 +65,26 @@ describe("mergePanelLayout", () => {
         "snippets",
       ],
     };
-    const result = mergePanelLayout(saved);
-    expect(result.sidebar[0]).toBe("files");
-    expect(result.sidebar[1]).toBe("variables");
-    expect(result.sidebar).toContain("packages");
-    expect(result.sidebar).toContain("ai");
-    expect(result.sidebar).toContain("outline");
-    expect(result.sidebar).toContain("documentation");
-    expect(result.sidebar).toContain("dependencies");
-    expect(result.developerPanel).toEqual(saved.developerPanel);
+    expect(mergePanelLayout(saved)).toEqual({
+      sidebar: [
+        "files",
+        "variables",
+        "packages",
+        "ai",
+        "outline",
+        "documentation",
+        "dependencies",
+      ],
+      developerPanel: [
+        "errors",
+        "scratchpad",
+        "tracing",
+        "secrets",
+        "logs",
+        "terminal",
+        "snippets",
+      ],
+    });
   });
 
   it("appends new developer panel entries missing from saved layout", () => {
@@ -74,12 +100,26 @@ describe("mergePanelLayout", () => {
       ],
       developerPanel: ["errors", "scratchpad", "tracing", "logs"],
     };
-    const result = mergePanelLayout(saved);
-    expect(result.sidebar).toEqual(saved.sidebar);
-    expect(result.developerPanel).toContain("terminal");
-    expect(result.developerPanel).toContain("snippets");
-    expect(result.developerPanel.indexOf("errors")).toBe(0);
-    expect(result.developerPanel.indexOf("scratchpad")).toBe(1);
+    expect(mergePanelLayout(saved)).toEqual({
+      sidebar: [
+        "files",
+        "variables",
+        "packages",
+        "ai",
+        "outline",
+        "documentation",
+        "dependencies",
+      ],
+      developerPanel: [
+        "errors",
+        "scratchpad",
+        "tracing",
+        "logs",
+        "secrets",
+        "terminal",
+        "snippets",
+      ],
+    });
   });
 
   it("does not duplicate panels already present", () => {
@@ -103,9 +143,26 @@ describe("mergePanelLayout", () => {
         "snippets",
       ],
     };
-    const result = mergePanelLayout(saved);
-    expect(result.sidebar).toEqual(saved.sidebar);
-    expect(result.developerPanel).toEqual(saved.developerPanel);
+    expect(mergePanelLayout(saved)).toEqual({
+      sidebar: [
+        "files",
+        "variables",
+        "packages",
+        "ai",
+        "outline",
+        "documentation",
+        "dependencies",
+      ],
+      developerPanel: [
+        "errors",
+        "scratchpad",
+        "tracing",
+        "secrets",
+        "logs",
+        "terminal",
+        "snippets",
+      ],
+    });
   });
 
   it("preserves panels the user moved between sections", () => {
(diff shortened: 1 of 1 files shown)


jacobcbeaudin:

jacobcbeaudin:
recheck

frontend/src/components/editor/chrome/state.ts
```diff
@@ -35,10 +35,34 @@ const DEFAULT_PANEL_LAYOUT: PanelLayout = {
...
+    sidebar: [
+      ...saved.sidebar,
+      ...DEFAULT_PANEL_LAYOUT.sidebar.filter((p) => !allSaved.has(p)),
+    ],
+    developerPanel: [
+      ...saved.developerPanel,
+      ...DEFAULT_PANEL_LAYOUT.developerPanel.filter((p) => !allSaved.has(p)),
+    ],
+  };
+}
```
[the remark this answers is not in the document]
jacobcbeaudin: Intentionally left unfiltered. Stale IDs are safely ignored at render time. Both `sidebarItems` and `devPanelItems` use `PANEL_MAP.get(id)` and discard unknown types via `flatMap`. They also get naturally cleaned up when a user reorders a panel, since the write-back only persists rendered items. Adding explicit filtering would silently drop IDs, which would lose the user's saved position if a panel were temporarily removed and re-added. Not worth the trade-off for a few inert strings in a per-browser JSON blob.

frontend/src/components/editor/chrome/__tests__/state.test.ts
```diff
@@ -0,0 +1,220 @@
...
+        "scratchpad",
+        "tracing",
+        "secrets",
+        "logs",
+        "terminal",
+        "snippets",
+      ],
+    };
+    const result = mergePanelLayout(saved);
+    expect(result.sidebar[0]).toBe("files");
```
Light2Dark: instead of individually testing the attributes, can we do a complete inline snapshot test (or just a normal object assertion)
```
expect(result).toMatchInlineSnapshot
```
and the tests below too

jacobcbeaudin: Done! Updated all tests to use complete toEqual assertions. Thank you for the review.

manzt (approved):
Nice fix! I was initially going to suggest we just bust the cache by bumping the storage key (e.g. "marimo:panel-layout-v2"), but I think this merge approach is more robust.

Issue #8306, opened 2026-02-13
Marimo can deadlock in run mode when `redirect_console_to_browser=True` in create_asgi_app


wlach:
### Describe the bug

Ran into this problem when using marimo in server mode serving a bunch of concurrent notebooks in run mode.

When there are concurrent sessions in run mode (with `redirect_console_to_browser` enabled), all writing to standard output, there can be data races where fd 2 points to a pipe that has already closed. This can lead to the whole server becoming unresponsive:

https://github.com/marimo-team/marimo/blob/36381ea9b9bada047a7b8153f5090a58685b8f1c/marimo/_messaging/streams.py#L194

Judging from this comment, I believe this redirection was unintentional:

https://github.com/marimo-team/marimo/blob/36381ea9b9bada047a7b8153f5090a58685b8f1c/marimo/_runtime/runtime.py#L3209

We *can* still redirect just python logging output to standard out (which is usually what we want), and I patched marimo to do that. However, it seems when that bug is fixed  I can't successfully start a new session when my thread is executing. I think (?) this might have to do with the behaviour of threading vs. the async event loop, but I haven't investigated thoroughly enough to be sure:

https://github.com/marimo-team/marimo/blob/36381ea9b9bada047a7b8153f5090a58685b8f1c/marimo/_runtime/runtime.py#L704

As a note I used an AI model (Claude Opus 4.6) heavily when debugging this, however the above words and thoughts are my own.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.19.11",
  "editable": false,
  "location": "/Users/wlach/src/marimo-server-deadlock-reproduction/.venv/lib/python3.11/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "24.6.0",
  "Processor": "arm",
  "Python Version": "3.11.13",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "144.0.7559.133",
    "Node": "v20.19.5",
    "uv": "0.9.14 (Homebrew 2025-12-01)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.16.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.20.1",
[...]

Issue #7984, opened 2026-01-26
Allow tracebacks to be exposed to users in deployed apps


tdsmith:
### Description

We’re interested in deploying Marimo apps to internal users using the ASGI app interface. If a cell throws an unhandled exception, we would like a traceback to be visible to the user on the app. Having an artifact that users can paste into Slack so that we can help them debug is valuable.

[This trivial app](https://gist.github.com/tdsmith/2def3ef6d173818bbc32f44a88931127) (run with `pixi run python serve.py`) demonstrates the current behaviour of unhandled exceptions. The cell where the error occurs renders a blank result. The first time the exception is encountered, a banner appears in the bottom right corner of the screen indicating that an internal error occurred. I don’t think there’s any visible indication the second time the error occurs.

The message is printed to the server's console, which is not accessible to the user.

### Suggested solution

Optionally render an error message containing a traceback as the result of the cell, instead of an empty value, if the cell has an unhandled exception.

### Are you willing to submit a PR?

- [ ] Yes

### Alternatives

Previously: https://github.com/marimo-team/marimo/issues/5250 was marked as fixed by https://github.com/marimo-team/marimo/pull/7249, which allows console messages to be published in the browser console, but not on the page itself. Our users understand how to copy tracebacks but are not generally familiar with browser devtools, so we'd really like to render the errors to the page.

### Additional context

Brief conversation on Discord: https://discord.com/channels/1059888774789730424/1465434155847389225/1465434155847389225

manzt:
Thanks for the detailed write-up. Better error visibility in run mode is a reasonable need, especially for internal deployments.

We already have `redirect_console_to_browser` on `create_asgi_app()` which sends console output (including tracebacks) to the browser... so there's some existing similar API here.

Off the top of my head, something like a display config option:

```
display.run_mode_errors: "off" | "console" | "inline"
```

Where:
- `"off"` - errors go to server logs only (default)
- `"console"` - errors show in browser console/output area (currently `redirect_console_to_browser`)
- `"inline"` - errors render directly in the cell output

It would be awkward to have separate configurations, so we'd probably want some backwards-compatible migration from `redirect_console_to_browser`. Still some details to figure out, but open to feedback on the direction.

tdsmith:
Thanks for reading. I agree that would be responsive to our needs!

wlach:
Very much agreed, this would be very useful for us.

[...]

PR #8353, opened 2026-02-18
fix: thread-safe redirection of console outputs in run mode

akshayka:
In run mode, multiple session threads share the same process,
and our manipulations of file descriptors were not designed to be
thread-safe. When `--redirect-console-to-browser` was supplied to `marimo run`, this was breaking
the marimo process, noted in #8306.

This change adds a proxy in run mode that is installed once as
`sys.stdout`/`sys.stderr` that dispatches writes to per-thread streams.

During cell execution, we register thread-local streams instead of
replacing `sys.stdout` globally or using file descriptor-level redirection.

This change also fixes a bug/ in which each client would receive the
console outputs belonging to all other clients.

NB: With this change, in run mode, output written directly to the
standard output and standard error file descriptors (for example, by a C-extension)
will no longer be redirected to the browser. This tradeoff should be acceptable,
since at least we won'thang the marimo process.

### DIFF
--- a/marimo/_messaging/thread_local_streams.py
+++ b/marimo/_messaging/thread_local_streams.py
@@ -25,6 +25,8 @@ if TYPE_CHECKING:
 
 _proxies_installed = False
 _install_lock = threading.Lock()
+_original_stdout: TextIO | None = None
+_original_stderr: TextIO | None = None
 
 
 class ThreadLocalStreamProxy(io.TextIOBase):
@@ -41,6 +43,11 @@ class ThreadLocalStreamProxy(io.TextIOBase):
         self._original = original
         self._local = threading.local()
         self._name = name
+        # Expose the underlying binary buffer so that code writing to
+        # sys.stdout.buffer (e.g. package installation logging) keeps working.
+        self.buffer: io.BufferedIOBase | None = getattr(
+            original, "buffer", None
+        )
 
     # -- per-thread registration -------------------------------------------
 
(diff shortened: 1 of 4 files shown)


akshayka:
@wlach I *think* this fixes the bug you reported (thanks so much for finding the root cause).

wlach:
> @wlach I _think_ this fixes the bug you reported (thanks so much for finding the root cause).

@akshayka neat! did you try it with my reproduction repo? I tried making a fix using similar premises as yours but it didn't quite work as I expected (it seems like I couldn't make new connections while an existing cell was running in another thread). Anyway inside here there's a good torture test which reproduces the original bug:

https://github.com/wlach/marimo-server-deadlock-reproduction

akshayka:
> > @wlach I _think_ this fixes the bug you reported (thanks so much for finding the root cause).
> 
> @akshayka neat! did you try it with my reproduction repo? I tried making a fix using similar premises as yours but it didn't quite work as I expected (it seems like I couldn't make new connections while an existing cell was running in another thread). Anyway inside here there's a good torture test which reproduces the original bug:
> 
> https://github.com/wlach/marimo-server-deadlock-reproduction

@wlach sorry I missed this earlier. I tried just now and can confirm that:

(1) Without the fix, after enough refreshes the server eventually refuses to start new sessions (kernel not found)
(2) With the fix, the server remains responsive even after tons of fast refreshes

Are you okay with the following tradeoff that this change introduces? In run mode `print` and `sys.stdout.write` will continue to be forwarded to the frontend, but `os.write(1, "my string")` won't be.

> NB: With this change, in run mode, output written directly to the
standard output and standard error file descriptors (for example, by a C-extension)
will no longer be redirected to the browser. This tradeoff should be acceptable,
since at least we won'thang the marimo process.

wlach:
> Are you okay with the following tradeoff that this change introduces? In run mode print and sys.stdout.write will continue to be forwarded to the frontend, but os.write(1, "my string") won't be.

Yup! I don't think there's any other reasonable way of doing it that avoids deadlocks. 

I think actually I would prefer to turn this feature off altogether for our purposes: the main thing we want is to be able to get tracebacks in running code as described in https://github.com/marimo-team/marimo/issues/7984

I actually have a prototype for that in https://github.com/wlach/marimo/tree/marimo-error-run-mode-modal -- if you're aligned with the solution I propose I can file a PR.

akshayka:
> I actually have a prototype for that in https://github.com/wlach/marimo/tree/marimo-error-run-mode-modal -- if you're aligned with the solution I propose I can file a PR.

@wlach I tried it out and I'm generally aligned, especially if this makes your experience better. Please open a PR, we will be happy to review it. Thanks!

PR #8354, opened 2026-02-18
feat: mo.ui.matrix()

akshayka:
This PR adds two new public APIs:

1. `mo.ui.matrix()`, which adds a reactive matrix component inspired by @koaning's wigglystuff matrix input
2. `mo.ui.vector()`, which is analogous to `matrix` but for vectors

## Signature remarks

* An initial value for the matrix is required. This value can be a Python list of lists, with each list a row of the matrix. It can also be a NumPy array.
* Minimum and maximum values can be provided, as well as a step size. These can be scalars or `ArrayLike` objects.
* The matrix can be constrained to be symmetric with a keyword argument
* The matrix can be disabled, either entirely or on an element wise basis with a boolean array mask. The latter provides a convenient way to constrain a matrix to be (for example) upper triangular, diagonal, or banded.
* The precision at which to display the entries is chosen intelligently on the user's behalf, but can be overriden

## Python usage

With Python built-in types:

```python
mat = mo.ui.matrix([[1, 0], [0, 1]])
```

```python
matrix.value
```

With NumPy:

```python
mat = mo.ui.matrix(np.eye(2))
```

```python
np.asarray(matrix.value)
```

The value, bounds, step, and disabled arguments can optionally be NumPy
arrays, interpreted elementwise.

```python
mat = mo.ui.matrix(
    np.zeros((3, 3)),
    min_value=np.full((3, 3), -10.0),
    max_value=np.full((3, 3), 10.0),
    step=np.full((3, 3), 0.5),
)
```

## `value` type

The element's `value` is a list of scalar (int or float) lists for `matrix`, and a flat list for `vector`. This is readily converted to a NumPy array by the user.

## Implementation

`mo.ui.vector` reuses the frontend implementation of `ui.matrix`, promoting its input to a 2D array 

## Media

image

https://github.com/user-attachments/assets/c5bf1ff5-3cfc-4cd0-9254-42a5364f4941

### DIFF
--- a/marimo/_plugins/ui/_impl/matrix.py
+++ b/marimo/_plugins/ui/_impl/matrix.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 from typing import (
     TYPE_CHECKING,
     Any,
-    Callable,
     Final,
 )
 
--- a/marimo/_plugins/ui/_impl/vector.py
+++ b/marimo/_plugins/ui/_impl/vector.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 from typing import (
     TYPE_CHECKING,
     Any,
-    Callable,
     Final,
 )
 


koaning:
Not 100% sure, but it might also be fun to play around with `__add__`-like dunder methods so that a user could also do things like `widget @ matrix`? We can also choose to ignore this for now and add that later if we want. It feels like it could be fun, but it is also easy to overdo operators in Python.

koaning:
With regards to the name, do we really want to call it `matrix`? Technically we may also want to declare 1D arrays here? So maybe `mo.ui.ndarray`? Not 100% sure.

manzt:
> So maybe mo.ui.ndarray? Not 100% sure.

To me, this naming would suggest I could pass in n-dimensional arrays (well beyond 1D and 2D).

marimo/_plugins/ui/_impl/matrix.py
```diff
@@ -0,0 +1,305 @@
...
+            downstream computations. Defaults to False.
+        label (str, optional): Markdown/LaTeX label for the element.
+            Defaults to "".
+        on_change (Callable | None, optional): Optional callback to run
+            when this element's value changes.
+    """
+
+    _name: Final[str] = "marimo-matrix"
+
+    def __init__(
```
manzt, line 339: Would it be overkill to consider using [`@overload`](https://typing.python.org/en/latest/spec/overload.html) on `__init__` for some nice ergnomics/typing behavior? e.g. `.value` returns `np.ndarray` when constructed from an ndarray, `list[list[Numeric]]` when constructed from a list:

```python
_T = TypeVar("_T")

class matrix(UIElement[list[list[Numeric]], _T]):
    @overload
    def __init__(self: matrix[np.ndarray], value: np.ndarray, ...) -> None: ...
    @overload
    def __init__(self: matrix[list[list[Numeric]]], value: list[list[Numeric]], ...) -> None: ...

    def _convert_value(self, value: list[list[Numeric]]) -> _T:
        if self._return_numpy:
            return np.asarray(value)
        return value
```

marimo/_plugins/ui/_impl/matrix.py
```diff
@@ -0,0 +1,305 @@
...
+        column_labels (list[str] | None, optional): Labels for each column.
+            Defaults to None.
+        debounce (bool, optional): If True, value updates are only sent
+            to the backend on mouse-up (pointer release) instead of on
+            every drag movement. Useful when the matrix drives expensive
+            downstream computations. Defaults to False.
+        label (str, optional): Markdown/LaTeX label for the element.
+            Defaults to "".
+        on_change (Callable | None, optional): Optional callback to run
+            when this element's value changes.
```
manzt: Same question as `mo.ui.matplotlib`, is this just a convention across all our UI elements we keep for consistency?

frontend/src/plugins/impl/MatrixPlugin.tsx
```diff
@@ -0,0 +1,280 @@
...
+    col: number;
+  } | null>(null);
+
+  // Local display value – always tracks the latest visual state.
+  // When debounce is true we update this locally during drag and only
+  // call setValue on pointer-up.
+  const [internalValue, setInternalValue] = useState(value);
+  useEffect(() => {
+    setInternalValue(value);
+  }, [value]);
```
manzt, line 95: nit: `useEffect` sync causes an extra render. I _think_ we could just read the prop directly when not dragging instead. Not a big deal though.

```tsx
const [draft, setDraft] = useState(value);
const internalValue = activeCell != null ? draft : value;
```

And then `setDraft` in various handlers.

akshayka:
> Not 100% sure, but it might also be fun to play around with __add__-like dunder methods so that a user could also do things like widget @ matrix? We can also choose to ignore this for now and add that later if we want. It feels like it could be fun, but it is also easy to overdo operators in Python.

Maybe in the future? I see the appeal. But historically we've chosen not to implement overrides (e.g., for `mo.ui.slider`) in favor of being explicit.

> With regards to the name, do we really want to call it matrix?

I can add a convenience `mo.ui.vector()` that reuses the frontend plugin under the hood

manzt (approved):

marimo/_plugins/ui/_impl/matrix.py
akshayka: We have this as a convention across all our UI elements, yea. I never use it, and actually prefer that people don't use it (I prefer folks opt for reactivity instead). But I believe some people call state setters or other side effects in the `on_change`. Do you still think we should omit it?

marimo/_plugins/ui/_impl/matrix.py
akshayka: It's not overkill on the implementation. But I wonder about the ergonomics. Is it confusing for `value`'s type at runtime to depend on what was passed in?

akshayka: If we did support returning NumPy out, then we might also need to support returning as a `torch.Tensor`, `jax.Array`, `tensorflow.Tensor`, since these are all arraylike objects that should work with the `matrix` constructor. I am mildly worried this may add too much complexity.

akshayka:
@dmadisetti @manzt I've added `mo.ui.vector` for 1D arrays, and have updated the PR description to describe the new API.

marimo/_plugins/ui/_impl/matrix.py
manzt: i honestly had the same thought. We'd need some kind of narwhals abstraction which i'm not sure there is for arrays. 

this current api seems the most forward thinking.

marimo/_plugins/ui/_impl/matrix.py
manzt: I'm fine with keeping for convention. Maybe something to consider for the chopping block in a major release.

manzt: Thinking about this a bit more... I'd be for omitting it (unless consistency is most important). I think there is precedent (`mo.ui.anywidget` just takes a widget instance), and adding might just lead to discoverability of a path which we don't consider a happy path (and make it harder to deprecate in the future). I'd consider adding in the future upon the most minimal request from our users.

akshayka: Removed

marimo/_plugins/ui/_impl/matrix.py
```diff
@@ -0,0 +1,399 @@
...
+        )
+
+        super().__init__(
+            component_name=matrix._name,
+            initial_value=data,
+            label=label,
+            args=args,
+            on_change=None,
+        )
+
```
dmadisetti, line 395: think defining an `__array__`/`__array_interface__` would be ideal (for our own apis as well like mo.image) https://numpy.org/doc/stable/user/basics.interoperability.html#the-array-interface-protocol

marimo/_plugins/ui/_impl/matrix.py
dmadisetti: See my comment below, there's already an interop standard

akshayka:
@dmadisetti for inter-op, we would still need to hardcode the original value's data type, and hardcode its constructor when returning the matrix via the value attribute, no? That's not hard for numpy, torch, jax, but it feels cumbersome and not very maintainable long term.

PR #8355, opened 2026-02-18
storage inspector frontend

Light2Dark:
## 📝 Summary

- Adds a feature flag to enable storage inspector
- Implements frontend features for storage inspector including:
     - Downloading a file
     - Searching a file (must have been expanded, only client-side search). This is consistent with cloudflare / aws where they do client-side searching rather than server-side search which can be expensive.
     - UI display for file types, file size, last modified
- Refactors UI components across session-panel for reuse here.

image

https://github.com/user-attachments/assets/96bedd6a-ee0e-44cb-967e-8d71e1a42d67

**Future improvements**
- ~Add 'copy' feature, copy the full URL~. Maybe 'copy Python snippet' to read locally in Python
- Preview file (open file in external browser / sidepanel / expanded)
- ~Refresh connection~

## 📋 Checklist

### DIFF
--- a/frontend/src/components/editor/chrome/panels/file-explorer-panel.tsx
+++ b/frontend/src/components/editor/chrome/panels/file-explorer-panel.tsx
@@ -91,10 +91,17 @@ const FileExplorerPanel: React.FC = () => {
     [setState],
   );
 
-  // Set a max-height since both sections are not resizable at the moment
-  const storageMaxHeight = Math.round(panelHeight * 0.5);
-  // File tree height: panel height minus both trigger heights
-  const fileTreeHeight = Math.max(200, panelHeight - TRIGGER_HEIGHT * 2);
+  const availableContent = panelHeight - TRIGGER_HEIGHT * 2;
+  const storageIsOpen = openSections.includes("remote-storage");
+  const bothOpen = storageIsOpen && openSections.includes("files");
+
+  const storageMaxHeight = bothOpen
+    ? Math.round(availableContent * 0.4)
+    : availableContent;
+  const fileTreeHeight = Math.max(
+    200,
+    bothOpen ? availableContent - storageMaxHeight : availableContent,
+  );
 
   const storageInspectorEnabled = getFeatureFlag("storage_inspector");
   if (!storageInspectorEnabled) {


frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -0,0 +1,570 @@
...
+    day: "numeric",
+    hour: "2-digit",
+    minute: "2-digit",
+  });
+}
+
+/** Extract display name from a full path (e.g., "folder/subfolder/" -> "subfolder") */
+function displayName(path: string): string {
+  // Remove trailing slash
+  const trimmed = path.endsWith("/") ? path.slice(0, -1) : path;
```
manzt, line 71: paths from these APIs are always `/`?

frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -0,0 +1,570 @@
...
+  depth,
+  locale,
+  searchValue,
+}) => {
+  const { entriesByPath } = useStorage();
+  const { setEntries } = useStorageActions();
+  const pathKey = storagePathKey(namespace, prefix);
+  const cached = entriesByPath.get(pathKey);
+
+  const { isPending, error } = useAsyncData(async () => {
```
manzt: This fetch logic (`useAsyncData` + `ListStorageEntries`) and the state (`storageAtom` + `setEntries`) sort of split across two layers, with this UI component manually orchestrating between them.

Could we encapsulate this into a `useStorageEntries` hook so the component just consumes `{ entries, isPending, error, refetch }`?

Something like:

```ts
// in core/storage/state.ts
export function useStorageEntries(
  namespace: string,
  prefix: string | null | undefined,
) {
  const { entriesByPath } = useStorage();
  const { setEntries, clearEntriesForPath } = useStorageActions();
  const cached = entriesByPath.get(storagePathKey(namespace, prefix));

  const { isPending, error, refetch } = useAsyncData(async () => {
    if (cached) return;
    const result = await ListStorageEntries.request({
      namespace,
      prefix: prefix ?? "",
      limit: DEFAULT_FETCH_LIMIT,
    });
    setEntries({ namespace, prefix, entries: result.entries });
  }, [namespace, prefix, cached === undefined]);

  return {
    entries: cached ?? [],
    isPending: isPending && !cached,
    error: cached ? undefined : error,
    refetch,
  };
}
```

Then the UI becomes just:
```ts
const { entries, isPending, error } = useStorageEntries(namespace, prefix);
```

frontend/src/core/storage/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  namespaces: StorageNamespace[];
+  /** Lazy-loaded entries keyed by "namespace::prefix" */
+  entriesByPath: ReadonlyMap<StoragePathKey, StorageEntry[]>;
+}
+
+export function storageUrl(
+  protocol: string,
+  rootPath: string,
+  entryPath: string,
+): string {
```
manzt: would it be overkill to return a [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL)?

Light2Dark: Doesn't change much, but I've changed it. I assume you're referring to the type, it is a valid url.

frontend/src/components/storage/storage-inspector.tsx
Light2Dark: For fsspec yeah, returns entire path from root. While obstore returns path from current directory.

Light2Dark:
> Very cool! A few small comments, otherwise looks good

Changed to remote storage, thanks for the name! Agree with not being descriptive enough, I want to do that in a follow-up, since I will add docs too.

frontend/src/components/storage/storage-inspector.tsx
Light2Dark: ty!

PR #8358, opened 2026-02-18
Extract MatplotlibPlugin to imperative renderer

manzt:
Implements a version of the suggestion in [review](https://github.com/marimo-team/marimo/pull/8342#discussion_r2820118258). Since all interaction state already lives in refs and React does no rendering work (it's all canvas), separating the DOM/canvas logic into its own class gives a cleaner boundary between React lifecycle and imperative drawing code.

These chnages extract into an imperative `MatplotlibRenderer` class without React. The class uses `AbortSignal` for cleanup (all `addEventListener` calls pass `{ signal }` for automatic removal) and a generation counter for stale image load guards. The React component becomes a thin lifecycle shell that creates the renderer on mount, syncs props via `update()` each render, and aborts the controller on unmount.

### DIFF
--- a/frontend/src/plugins/impl/matplotlib/MatplotlibPlugin.tsx
+++ b/frontend/src/plugins/impl/matplotlib/MatplotlibPlugin.tsx
@@ -24,8 +24,8 @@ export class MatplotlibPlugin implements IPlugin<T, Data> {
     selectionOpacity: z.number().default(0.15),
     strokeWidth: z.number().default(2),
     debounce: z.boolean(),
-    xScale: z.string().default("linear"),
-    yScale: z.string().default("linear"),
+    xScale: z.enum(["linear", "log"]).default("linear"),
+    yScale: z.enum(["linear", "log"]).default("linear"),
   });
 
   render(props: IPluginProps<T, Data>): JSX.Element {
--- a/frontend/src/plugins/impl/matplotlib/matplotlib-renderer.ts
+++ b/frontend/src/plugins/impl/matplotlib/matplotlib-renderer.ts
@@ -2,6 +2,8 @@
 
 import type { Setter } from "@/plugins/types";
 
+export type ScaleType = "linear" | "log";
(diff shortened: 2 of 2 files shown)


frontend/src/plugins/impl/matplotlib/MatplotlibPlugin.tsx
```diff
@@ -61,648 +39,30 @@ export class MatplotlibPlugin implements IPlugin<T, Data> {
...
-    }, []);
-
-    // Clean up rAF on unmount
-    useEffect(() => {
-      return () => cancelAnimationFrame(interactionRef.current.rafId);
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, []);
+  useEffect(() => {
+    instance.current?.update(props);
+  });
```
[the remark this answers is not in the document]
manzt: This is intentional... i'll add a comment

frontend/src/plugins/impl/matplotlib/matplotlib-renderer.ts
```diff
@@ -0,0 +1,631 @@
...
+      state.height !== prev.height
+    ) {
+      this.#drawCanvas();
+    }
+  }
+
+  #loadImage(chartBase64: string): void {
+    this.#currentChartBase64 = chartBase64;
+    this.#imageGeneration++;
+    const generation = this.#imageGeneration;
```
Light2Dark, line 256: sorry, just cursory review
is this intentional? assigning after increment? looks like the check below will always be true

Light2Dark: probably is, to avoid faulty on-loads

Light2Dark (approved):
possible to add tests? looks quite well composed

manzt:
> possible to add tests? looks quite well composed

I added some unit tests for the scaling funcs

PR #8362, opened 2026-02-18
feat: allow excluding code blocks when exporting to PDF

### LINKED ISSUE #8319: Option to suppress inputs for `marimo export pdf`
### Description

I would love an option for `marimo export pdf` to produce output similar to `marimo run` followed by print to PDF, namely only including markdown cells and the _output_ of python cells.

### Suggested solution

A command line flag or environment variable (or both) would do the trick

- [ ] Yes

### Alternatives

I tried 
```bash
uvx --with nbconvert --with jupyter --from jupyter-core jupyter nbconvert --to pdf --execute --stdout $NOTEBOOK \
  --TemplateExporter.exclude_input=True
```
as suggested in [the docs](https://docs.marimo.io/guides/exporting/#export-pdf-from-the-command-line) but math equations in markdown came out unrendered

### Additional context

_No response_

peter-gy:
## 📝 Summary

Enables PDF export without including input code blocks.

Closes #8319

## 🔍 Description of Changes

We collect `--include-inputs/--no-include-inputs` and pass it to the `PDFExporter` instance as option.

## Result

left: notebook exported with `--include-inputs` (default behavior)
right: notebook exported with `--no-include-inputs`

**WebPDF Export**

Unknown

**LaTeX Export**

Unknown-1

### DIFF
--- a/docs/guides/exporting.md
+++ b/docs/guides/exporting.md
@@ -131,6 +131,14 @@ enables exporting to PDF and other formats with Pandoc. See this [publishing](./
 
 ### Export PDF from the command line
 
+You can export directly with marimo:
+
+```bash
+marimo export pdf notebook.py -o notebook.pdf --no-include-inputs
+```
+
+The `--no-include-inputs` flag hides Python code cells while keeping markdown and cell outputs in the generated PDF.
+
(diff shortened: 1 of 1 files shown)


Light2Dark (approved):
nice!

PR #8365, opened 2026-02-18
Use `<table>` for `mo.ui.matrix` with right-aligned numbers

manzt:
The flex-based layout couldn't right-align numbers within columns because each cell sized to its content independently. Switching to an HTML [`table`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/table) gives automatic column-width equalization, so `text-right` on each `<td>` properly aligns numbers like numpy output.  It is also semantically more appropriate here since a matrix is tabular data.

Also adds `tabular-nums` for consistent digit widths.

| before | after |
   |-------------|---------|
   |  |  |

https://github.com/user-attachments/assets/e08b905e-73a8-4f1e-8f78-f7b7b67b4acd

### DIFF
--- a/frontend/src/plugins/impl/MatrixPlugin.tsx
+++ b/frontend/src/plugins/impl/MatrixPlugin.tsx
@@ -193,10 +193,13 @@ const MatrixComponent = ({
   const hasRowLabels = rowLabels != null && rowLabels.length > 0;
   const hasColumnLabels = columnLabels != null && columnLabels.length > 0;
 
+  const numRows = internalValue.length;
+  const numCols = internalValue[0]?.length ?? 0;
+
   return (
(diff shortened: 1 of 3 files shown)


manzt:
With labels:

akshayka (approved):
This looks great to me (though frontend test is failing). EDIT:  I do prefer the labels on the outside of the matrix though

manzt:
I'm happy to merge base branch, and then fast follow with this in the morning (fixing test and moving labels).

manzt:
> I do prefer the labels on the outside of the matrix though

Just pushed up:

https://github.com/user-attachments/assets/2e692ad8-290d-42d7-b0ab-b5f6cddd8132

akshayka (approved):

PR #8374, opened 2026-02-19
fix: Only trigger cell when ALL its dependencies are unblocked

snooze92:
## 📝 Summary

I noticed that when one cell depended on two branches that stopped, the dependant cell correctly stopped initially. But one of the two branches getting "unstopped" unfortunately triggered the dependant cell to run too soon, which could lead to issues such as NameError.

This is an attempt at fixing the issue.

## 🔍 Description of Changes

It seems that every new trigger uses clean state, where the whole transitive closure of the triggered cell gets triggered without the rest of the tree being evaluated at all. With this change, I look for any cell that is about to be run for blocked missing refs and I cancel the cell and its transitive closure whenever I find such blocked missing refs.

Unfortunately, I have to assume that a cell found in "exception" state with a None exception was a stop, and re-wrap the None in a MarimoStopError to get the correct "ancestor stopped" message in the frontend. That seems like a separate cleanup that would be good for Marimo: either make MarimoStopError extend Exception rather than BaseException, or let Cell.exception store BaseExceptions.

In CancelledCells, we might now call add with the same `raising_cell` arguments but different `descendants`, in the case where two different cells depend on the same blocked cell. Things workout fine without a change, but it seems more correct to append all descendants in this case for completeness.

Below is a simple notebook to manually reproduce/verify:

```python
import marimo

__generated_with = "0.19.11"

@app.cell
def _():
    import marimo as mo

    return (mo,)

@app.cell
def _(mo):
    a = mo.ui.text(value="ha!", label="A:").form()
    a
    return (a,)

@app.cell
def _(a, mo):
    mo.stop(not a.value, "A not submitted")

    aa = a.value.upper()
    return (aa,)

@app.cell
def _(mo):
    b = mo.ui.text(value="beh...", label="B:").form()
    b
    return (b,)

@app.cell
def _(b, mo):
    mo.stop(not b.value, "B not submitted")

    bb = b.value.upper()
    return (bb,)

@app.cell
def _(aa, bb):
    aa + bb
    return

@app.cell
def _(aa, bb):
    bb + aa
    return

if __name__ == "__main__":
    app.run()
```

Before this fix, submitting either form before the other runs the last two cells and fails with a NameError because either `aa` or `bb` is not defined...

A side effect of this fix is that when explicitly trying to run a cell that depends on a stopped ancestor, instead of getting a NameError we now get a clear "stopped ancestor" message.

## 📋 Checklist

### DIFF
--- a/marimo/_runtime/runner/cell_runner.py
+++ b/marimo/_runtime/runner/cell_runner.py
@@ -665,6 +665,24 @@ class Runner:
                 blamed_cell = var_cell_id
         return ref, blamed_cell
 
+    def _find_first_blocked_missing_ref(
+        self, cell_id: CellId_t
+    ) -> Optional[CellId_t]:
+        """Return the first out-of-run ancestor still in a stopped/errored state."""
+        cell = self.graph.cells[cell_id]
+        for ref in cell.refs:
+            if ref in self.glbls:
+                # The reference is already available
+                continue
+            for defining_cell_id in self.graph.get_defining_cells(ref):
+                if defining_cell_id in self._run_position:
+                    # The defining cell is part of this run
+                    continue
+                defining_cell = self.graph.cells[defining_cell_id]
+                if defining_cell.run_result_status == "exception":
+                    return defining_cell_id
+        return None
+
     async def run_all(self) -> None:
         from marimo._runtime.runner.hook_context import (
             OnFinishHookContext,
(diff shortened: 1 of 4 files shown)


snooze92:
I don't seem able to add a label myself.

I see `tests._server.api.endpoints.test_resume_session.test_resume_session_with_watch` failed in the CI, but it works locally and I notice `@pytest.mark.flaky(reruns=3)` in the code for that test.

PR #8375, opened 2026-02-19
feat: CLI error UX improvements

peter-gy:
## 📝 Summary

Improves CLI error UX so error output is easier to scan, easier to act on, and more consistent across entry points.

Parser errors now do not include full help dump, typo paths suggest likely commands/flags, and dependency failures show pkg manager aware install commands with clear fallback guidance. Runtime/setup failures that were previously surfaced as usage errors were also normalized to runtime-style CLI errors.

## 🔍 Description of Changes

I used https://github.com/marimo-team/marimo/compare/ms/cli-errors?expand=1 as the starting direction for CLI error normalization and carried over the core idea of moving non-usage failures to structured runtime/dependency errors through shared abstractions.

From there, this PR extends the scope by:

- compact parser UX so parse failures show: `error`, `usage`, and `--help` hint, instead of full command help.
- typo suggestions for unknown commands/subcommands and short-flag mistakes.
- declarative, manager-aware install/follow-up command generation and wires it through dependency-heavy CLI paths (`export`, `thumbnail`, `convert`, `sandbox`, `mcp`, `upgrade`).

### Top-level typo suggestion

```bash
# Before
$ marimo xport
Error: No such command 'xport'.
...full help output...

# After
$ marimo xport
error: unrecognized command 'xport'

  tip: a similar command exists: 'export'

Usage: marimo [OPTIONS] COMMAND [ARGS]...
For more information, try '--help'.
```

### Nested command typo suggestion

```bash
# Before
$ marimo export hxml
Error: No such command 'hxml'.
...full export help output...

# After
$ marimo export hxml
error: unrecognized command 'hxml'

  tip: a similar command exists: 'html'

Usage: marimo export [OPTIONS] COMMAND [ARGS]...
For more information, try '--help'.
```

### Flag typo handling

```bash
# Before
$ marimo edit --hedlss
Error: No such option: --hedlss (Possible options: --headless, --help)
...full edit help output...

# After
$ marimo edit --hedlss
error: unexpected argument '--hedlss' found

  tip: some similar arguments exist: '--headless', '--help'

Usage: marimo edit [OPTIONS] [NAME] [ARGS]...
For more information, try '--help'.
```

### Runtime vs usage error semantics

```bash
# Before
$ SHELL= marimo shell-completion
...usage-style failure + help dump...
# exit 2

# After
$ SHELL= marimo shell-completion
Error: Could not determine shell. Please set $SHELL environment variable.
# exit 1
```

### Dependency hints (manager-aware + fallback)

```bash
# Before (example: export pdf missing deps)
Error: ... required ...
  Tip: Install with:
    pip install ...

# After
Error: ... required ...

  Tip: Install with:
    <manager-aware command>

  Or with pip:
    python -m pip install ...

  Tip: Alternative:
    marimo export pdf ... --sandbox
```

### MCP / sandbox dependency paths

```bash
# Before
single hard-coded install guidance, often usage-error style

# After
structured dependency error with:
- manager-aware install command
- pip fallback
- consistent runtime/dependency error semantics
```

### DIFF
--- a/marimo/_cli/upgrade.py
+++ b/marimo/_cli/upgrade.py
@@ -29,9 +29,12 @@ def print_latest_version(current_version: str, state: MarimoCLIState) -> None:
     echo(orange(message))
     upgrade_commands = get_upgrade_commands("marimo")
     if upgrade_commands:
-        echo(f"Run {green(upgrade_commands[0])} to upgrade.")
-        if len(upgrade_commands) > 1:
-            echo(f"Or run {green(upgrade_commands[1])}.")
+        if len(upgrade_commands) == 1:
+            echo(f"Run {green(upgrade_commands[0])} to upgrade.")
+        else:
+            primary_command = green(upgrade_commands[0])
+            fallback_command = green(upgrade_commands[1])
+            echo(f"Run {primary_command} or {fallback_command} to upgrade.")
 
     if state.notices:
         echo()
--- a/tests/_cli/test_upgrade.py
+++ b/tests/_cli/test_upgrade.py
@@ -64,6 +64,18 @@ def test_print_latest_version_uses_manager_aware_commands(
         "python -m pip install --upgrade marimo" in call[0][0]
         for call in mock_echo.call_args_list
     )
+    assert any(
+        (
+            "Run pixi upgrade marimo or "
+            "python -m pip install --upgrade marimo to upgrade."
+        )
+        in call[0][0]
+        for call in mock_echo.call_args_list
+    )
+    assert not any(
+        call.args and str(call.args[0]).startswith("Or run ")
+        for call in mock_echo.call_args_list
+    )
 
 
 @patch("marimo._cli.upgrade._fetch_data_from_url")


peter-gy:
@akshayka while this is not an API change, it is a very close-to-the-user modification, so your feedback would be appreciated!

dmadisetti:
Nice ! 
image

dmadisetti:
🤔  @peter-gy I think we need to rename that test. Not sure if you are online so I'll do it

marimo/_cli/suggestions.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _lower_map(candidates: Iterable[str]) -> dict[str, str]:
+    lowered: dict[str, str] = {}
+    for candidate in candidates:
+        lowered.setdefault(candidate.lower(), candidate)
+    return lowered
+
+
+def suggest_commands(token: str, candidates: Iterable[str]) -> list[str]:
```
dmadisetti, line 24: Ah, I was wondering if it was a built in. I guess fine, but would be nice to maybe break out `levenshtein_distance` into a util or something if needed elsewhere. Did you have an implementation source? I think in fe we also call this edit_distance

peter-gy:
> 🤔 @peter-gy I think we need to rename that test. Not sure if you are online so I'll do it

thanks!

marimo/_cli/suggestions.py
peter-gy: > Ah, I was wondering if it was a built in. I guess fine, but would be nice to maybe break out levenshtein_distance into a util or something if needed elsewhere.

Refactored it into a dedicated util and renamed to `edit_distance` for consistency

> Did you have an implementation source?

I let gpt-5.3-codex-high generate it

marimo/_cli/help_formatter.py
```diff
@@ -9,13 +9,22 @@
...
 
 
+def _split_option_token(token: str) -> tuple[str, str]:
+    """Split option-like tokens without depending on click's private parser."""
+    first = token[:1]
+    if first.isalnum():
+        return "", token
+    if token[1:2] == first:
+        return token[:2], token[2:]
+    return first, token[1:]
```
peter-gy, line 25: older supported Click versions (e.g. 8.1.x) don’t expose `click.parser._split_opt`, so importing it crashes CLI module import. this is a compat fix to address the issue.

marimo/_cli/suggestions.py
dmadisetti: Well tests pass !

dmadisetti: Also, I think it's fine to keep scoped in the file it's only used one place. Was more thinking aloud. But nice

peter-gy:
Thanks for the input @akshayka.

> 1. When is the upgrade message shown? Is it deliberate for it to only show on error?

Good catch. I changed the order so `check_for_updates(...)` runs only after `edit` preflight succeeds (sandbox deps + shared-memory checks), right before `start(...)`.  So on setup failures like missing `pyzmq`, we no longer print the upgrade banner first.

> 2. Error legibility.

Great point, just made the `Error` prefix bold red.

> 3. pyzmq suggested install.

Mm, yes, makes sense. Got rid of redundant `pyzmq` suggestion.

akshayka (commented):
@peter-gy thanks for the changes, the messages feel much better now.

One nit:

Instead of splitting the upgrade messages over two sentences, i.e. instead of

```
Run pixi upgrade marimo to upgrade.
Or run python -m pip install --upgrade marimo.
```

could we instead have a single sentence

```
Run pixi upgrade marimo or python -m pip install --upgrade marimo to upgrade.
```

peter-gy:
> @peter-gy thanks for the changes, the messages feel much better now.
> 
> One nit:
> 
> Instead of splitting the upgrade messages over two sentences, i.e. instead of
> 
> ```
> Run pixi upgrade marimo to upgrade.
> Or run python -m pip install --upgrade marimo.
> ```
> 
> could we instead have a single sentence
> 
> ```
> Run pixi upgrade marimo or python -m pip install --upgrade marimo to upgrade.
> ```

The idea of the list-like design was to let each command sit on its own copy-pasteable line, similarly to the suggestions we display for missing dependencies.
I can see the value of the single sentence approach as well though.

PR #8376, opened 2026-02-19
Add option to display error tracebacks inline and in modal

### LINKED ISSUE #7984: Allow tracebacks to be exposed to users in deployed apps

wlach:
## 📝 Summary

Closes #7984

## 🔍 Description of Changes

Add an optional mode (disabled by default) to display error tracebacks both inline and in a modal when in run (app) mode. This makes it easier for users to report errors in internal environments.

## 📋 Checklist

### DIFF
--- a/frontend/src/core/cells/logs.ts
+++ b/frontend/src/core/cells/logs.ts
@@ -78,7 +78,9 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
 
     // Find exception errors and check for traceback
     const exceptionErrors = cell.output.data.filter(
-      (error) => "type" in error && error.type === "exception" || error.type === "internal",
+      (error) =>
+        ("type" in error && error.type === "exception") ||
+        error.type === "internal",
     );
 
     if (exceptionErrors.length > 0 && !didAlreadyToastError) {
@@ -94,13 +96,15 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
         const handleClick = () => {
           store.set(tracebackModalAtom, {
             traceback: errorWithTraceback.traceback,
-            errorMessage: errorWithTraceback.msg || "An internal error occurred",
+            errorMessage:
+              errorWithTraceback.msg || "An internal error occurred",
           });
         };
 
         toast({
           title: "An internal error occurred",
-          description: errorWithTraceback.msg || "Click 'View' to see traceback",
+          description:
+            errorWithTraceback.msg || "Click 'View' to see traceback",
           variant: "danger",
           action: React.createElement(
             ToastAction,
--- a/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
+++ b/frontend/src/components/editor/renderers/vertical-layout/vertical-layout.tsx
@@ -121,9 +121,7 @@ const VerticalLayoutRenderer: React.FC<VerticalLayoutProps> = ({
         staleInputs={cell.staleInputs}
         name={cell.name}
         kiosk={kioskMode}
-        showErrorTracebacks={
-          userConfig.runtime.show_tracebacks ?? false
-        }
+        showErrorTracebacks={userConfig.runtime.show_tracebacks ?? false}
       />
     );
   };
@@ -439,7 +437,8 @@ const VerticalCell = memo(
       outputIsError &&
       Array.isArray(output?.data) &&
       output.data.some(
-        (e: MarimoError) => e.type === "exception" && "traceback" in e && e.traceback,
+        (e: MarimoError) =>
+          e.type === "exception" && "traceback" in e && e.traceback,
       );
     const hidden =
       (errored || interrupted || stopped || outputIsError) && !hasTraceback;


wlach:
@akshayka @manzt per discussion, would love some feedback. Currently lacking docs, happy to try my hand at adding some.

marimo/_cli/cli.py
```diff
@@ -1002,6 +1002,13 @@ def _collect_marimo_files(paths: list[str]) -> _CollectedRunFiles:
     hidden=True,
     help="Custom asset URL for loading static resources. Can include {version} placeholder.",
 )
+@click.option(
+    "--show-error-tracebacks/--no-show-error-tracebacks",
```
akshayka:
```suggestion
    "--show-tracebacks/--no-show-tracebacks",
```

marimo/_config/config.py
```diff
@@ -169,6 +169,9 @@ class RuntimeConfig(TypedDict):
        The default is None.
     - `default_csv_encoding`: the default encoding for CSV exports.
         The default is `"utf-8"`.
+    - `show_error_tracebacks`: if `True`, show detailed error tracebacks in run mode.
+        When enabled, exceptions will display a clickable toast that opens a modal with the full traceback.
+        The default is `True`.
```
akshayka: Does this need to be in config? This isn't really a runtime configuration in my mind. The console redirection is also not in config

frontend/src/core/cells/logs.ts
```diff
@@ -72,17 +76,48 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
       });
     });
 
-    const shouldToast = cell.output.data.some(
-      (error) => error.type === "internal",
+    // Find exception errors and check for traceback
+    const exceptionErrors = cell.output.data.filter(
+      (error: any) => error.type === "exception" || error.type === "internal",
     );
```
manzt, line 84: I believe we can remove the explicit `any` with type narrowing:

```suggestion
    const exceptionErrors = cell.output.data.filter(
      (error) =>
        "type" in error &&
        (error.type === "exception" || error.type === "internal"),
    );
```

frontend/src/core/cells/logs.ts
```diff
@@ -72,17 +76,48 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
...
-      toast({
-        title: "An internal error occurred",
-        description: "See console for details.",
-        className:
-          "text-xs text-background bg-(--red-10) py-2 pl-3 *:flex *:gap-3",
-      });
+
+      // Find first error with a traceback
+      const errorWithTraceback = exceptionErrors.find(
+        (error: any) => error.traceback,
```
manzt: I think with changes above this can be removed (type is inferred).

frontend/src/core/cells/logs.ts
```diff
@@ -72,17 +76,48 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
...
+      );
```
manzt, line 92: I think with changes above this can be removed (type is inferred).

```suggestion
      const errorWithTraceback = exceptionErrors.find(
        (error) => error.traceback,
      );
```

wlach:
Thanks all! Very reasonable requests for changes, I'll try to find some time to polish this up next week.

marimo/_config/config.py
wlach: Happy to take this out.

wlach: Actually looking again I think it would be difficult for it not to be runtime config as that seems to be the only easy way to propagate the state from the server to the client. Lmk if I'm missing something

wlach: Note also that the comment was wrong, the default should be `False`

frontend/src/core/cells/logs.ts
wlach: Appears so!

tests/_messaging/test_tracebacks.py
```diff
@@ -129,6 +129,43 @@ def test_write_traceback_html_for_regular_http_request(self) -> None:
         finally:
             HTTP_REQUEST_CTX.reset(token)
 
+    def test_write_traceback_suppressed_in_run_mode(self) -> None:
```
wlach, line 132: Adding some tests here to verify both that the traceback *is* propagated via websocket when the setting is there and it is *not* propagated via websocket if it isn't (the default).

Not sure if this is the best way of doing this, feedback welcome.

wlach:
@mscolnick @akshayka @manzt Sorry, I got a little over my skis with this one. I attempted to reply to all the review feedback and I'm fairly confident the PR is *functionally* correct, however I've run into some issues with CI that I'm not sure how to overcome (I'm getting errors that don't seem to have anything to do with these changes both in CI and locally) and I also don't feel I understand marimo's internal architecture well enough to be fully confident that my changes to the internals are done in the best possible way. 

FWIW I've been testing locally with this notebook:

```py
import marimo

@app.cell
def _():
    import marimo as mo
    return


@app.cell
def _():
    raise RuntimeError("LOL")
    return


if __name__ == "__main__":
    app.run()
```

Verified that:

1. In regular mode (`marimo -d run exceptional_notebook.py`) a trace uuid is printed to the console (without details) and an error is emitted to standard out.
2. In traceback mode (`marimo -d run --show-tracebacks exceptional_notebook.py`) it shows the traceback in a modal per the spec here https://github.com/marimo-team/marimo/issues/7984#issuecomment-3900431217.
3. With just `--redirect-console-to-browser` (`marimo -d run --redirect-console-to-browser exceptional_notebook.py`) it will show the traceback in the browser console but not show the new modal UI to copy the traceback
4. With both `--redirect-console-to-browser` *and* `--show-tracebacks` (`marimo -d run --redirect-console-to-browser --show-tracebacks exceptional_notebook.py`) it will both show the trackback in the console and the new UI.

Definitely happy to iterate further on this as time allows, but also fine if you want to take it from here (maintainers should be able to write to my branch).

dmadisetti:
@wlach thanks for your patience, I know this one has been stalled for a bit. For the most part looks good to me. Just some thoughts for discussion:

1. We only toast the first error seen 
2. The spacing between errors is a little harsh

image

Aside from that I think it looks great. A step up over just console logs I think for immediate debugging

dmadisetti:
I'd actually be OK with these as a followup. Happy to get this in as is

wlach:
> I'd actually be OK with these as a followup. Happy to get this in as is

Cool, that's obviously fine with me (if you think the implementation is sound). I'm not sure what to do about the CI error though. Looking at the logs it seems to be unrelated?

```
Running in a sandbox: /opt/hostedtoolcache/uv/0.11.3/x86_64/uv run --isolated --no-project --compile-bytecode --with-requirements /tmp/tmp4c_gsnxm.txt --refresh --python 3.12.3 marimo new -p 2724 --headless --no-token
  × No solution found when resolving `--with` dependencies:
  ╰─▶ Because there is no version of marimo[lsp]==0.22.0 and you require
      marimo[lsp]==0.22.0, we can conclude that your requirements are
      unsatisfiable.
```

dmadisetti:
Test looks failure looks unrelated, but thanks for checking

PR #8379, opened 2026-02-19
feat: support vectors in ui.matrix

akshayka:
This change updates the (unreleased) API for ui.matrix, allowing it to take column vectors. When the input is 1 dimensional, the element's value is also one dimensional. This change also removes the (unreleased) `ui.vector` symbol, in favor of a single `ui.matrix` element.

Despite the misnomer having an element that supports both 1 and 2d inputs is more convenient than making the user choose between two different APIs.


marimo/_plugins/ui/_impl/matrix.py
```diff
@@ -338,12 +426,18 @@ class matrix(UIElement[list[list[Numeric]], list[list[Numeric]]]):
...
+        min_value: (
+            Numeric | list[list[Numeric]] | list[Numeric] | ArrayLike | None
+        ) = None,
+        max_value: (
+            Numeric | list[list[Numeric]] | list[Numeric] | ArrayLike | None
+        ) = None,
+        step: (
+            Numeric | list[list[Numeric]] | list[Numeric] | ArrayLike
+        ) = 1.0,
+        disabled: bool | list[list[bool]] | list[bool] | ArrayLike = False,
```
dmadisetti, line 440: Right, the typing is pretty annoying, but smaller api will be a plus I think

PR #8413, opened 2026-02-21
fix: Copy `ExecutionContext.output` to `marimo.Thread`

coastalwhite:
## 📝 Summary

Closes #8402

## 🔍 Description of Changes

The issue is that `ProgressBar` uses the `mo.output.flush()` which relies on the existing `ExecutionContext.output`. For `mo.Thread`, these are left empty. This patch copies over the `output` into the threads `ExecutionContext`.

This is a one-line change to copy over the output.

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [ ] Tests have been added for the changes made.
- [ ] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Pull request title is a good summary of the changes - it will be used in the [release notes](https://github.com/marimo-team/marimo/releases).

PR #8418, opened 2026-02-21
fix: make threads share their accumulated output

akshayka:
This PR makes `mo.Thread`s share their creating cell's accumulated output list. This improves the experience of imperatively writing output with threads in two ways:

1. Multiple threads can now write to the output, without overwriting each other's outputs
2. Thread outputs are no longer automatically cleared when the spawning cell completes

To accommodate this, this PR abstracts away the underlying list of imperative outputs, and instead introduces a thread-safe container that manages access to the outputs data structure.

This change also makes `mo.status.progress_bar` thread-safe.

Follow up to #8413

### DIFF
--- a/marimo/_runtime/output/_output.py
+++ b/marimo/_runtime/output/_output.py
@@ -66,7 +66,7 @@ def replace_at_index(value: object, idx: int) -> None:
     except ContextNotInitializedError:
         return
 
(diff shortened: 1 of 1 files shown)


akshayka:
@coastalwhite this should generally improve your experience of writing to output from multiple threads (including with progress bars)

image

PR #8426, opened 2026-02-22
fix: exclude kernel memory from server memory stats to prevent double-counting

### LINKED ISSUE #8320: Machine stats card on bare metal host (ie not in container environment) showing higher marimo server and kernel usage than computer memory
### Describe the bug

Image

Unless I am missing something I think it doesn't make sense for total computer memory to be lower than that of the marimo server + kernel.

This should theoretically be independent of https://github.com/marimo-team/marimo/pull/7488 since this is not a container environment.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
ccloud@mars:~/medicaid$ uv run marimo env
{
  "marimo": "0.19.11",
  "editable": false,
  "location": "/home/ccloud/medicaid/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.8.0-94-generic",
  "Processor": "x86_64",
  "Python Version": "3.13.12",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "0.10.2"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.16.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "polars": "1.38.1"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
import polars as pl

# This file is pretty big, maybe not big enough to fill up the whole 600+GiB of kernel ram but should reproduce it
df = pl.read_parquet('https://console.carolinacloud.io/static/datasets/medicaid-provider-spending.parquet')

# This was just part of my code since the rest relies on a bunch of local data files so let me know if this doesn't actually reproduce the issue and figure something out
```

bxff:
## Summary

Fixes a bug where the usage endpoint reports `server + kernel > total system memory`.

**Problem**: `server_memory` was calculated by summing the main process + all recursive children via `main_process.children(recursive=True)`. Since the kernel process is a child of the server, its memory (and its children's memory) was included in both the `server` and `kernel` fields of the response.

**Fix**: Collect kernel PIDs first, then skip them when summing server memory.

Fixes #8320

## Test Plan

- All 6 existing tests in `test_health.py` pass
- The fix is logically straightforward: reorder the two code blocks and add a `kernel_pids` set filter

### DIFF
--- a/marimo/_server/api/endpoints/health.py
+++ b/marimo/_server/api/endpoints/health.py
@@ -221,31 +221,36 @@ async def usage(request: Request) -> JSONResponse:
         # subsequent calls return delta since last call
         cpu = psutil.cpu_percent(interval=None)
 
-    # Server memory (and children)
-    main_process = psutil.Process()
-    server_memory = main_process.memory_info().rss
-    children = main_process.children(recursive=True)
-    for child in children:
-        try:
-            server_memory += child.memory_info().rss
-        except psutil.NoSuchProcess:
-            pass
-
-    # Kernel memory
+    # Collect kernel PIDs first so we can exclude them from server memory
     kernel_memory: Optional[int] = None
+    kernel_pids: set[int] = set()
     session = AppState(request).get_current_session()
     try:
         if session and (pid := session.kernel_pid()) is not None:
             kernel_process = psutil.Process(pid)
+            kernel_pids.add(kernel_process.pid)
             kernel_memory = kernel_process.memory_info().rss
             kernel_children = kernel_process.children(recursive=True)
             for child in kernel_children:
+                kernel_pids.add(child.pid)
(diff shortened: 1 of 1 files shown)


bxff:

PR #8430, opened 2026-02-23
fix: plotly-resampler (and other FigureWidget subclasses) not working in marimo by routing `FigureWidget` through the anywidget formatter and syncing widget state via `_repr_mimebundle_()`.

### LINKED ISSUE #4091: plotly-resampler doesn't work with marimo
### Describe the bug

It seems to work with jupyter but fails with marimo, even when the mode is set to 'widget' (which should output an ipython widget i think?).

I'm not sure where the incompatibility lies, there's also a corresponding issue opened in plotly-resampler https://github.com/predict-idlab/plotly-resampler/issues/331

### Environment

<details>

```
{
  "marimo": "0.11.19",
  "OS": "Linux",
  "OS Version": "5.10.234-205.895.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.12.9",
  "Binaries": {
    "Browser": "--",
    "Node": "v18.20.5"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.30.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "ruff": "0.10.0",
    "starlette": "0.46.1",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "anywidget": "0.9.15",
    "pandas": "2.2.3",
    "polars": "1.24.0",
    "pyarrow": "18.1.0"
  },
  "Experimental Flags": {}
}```

</details>


### Code to reproduce

import plotly.express as px
from plotly_resampler import register_plotly_resampler
register_plotly_resampler(mode='widget') # 'auto' doesn't work either
px.line(df, ..., ...) # Plot some high frequency data

### PR DESCRIPTION
**Pull request by mscolnick**
Fix plotly-resampler (and other FigureWidget subclasses) not working in marimo by routing `FigureWidget` through the anywidget formatter and syncing widget state via `_repr_mimebundle_()`.

**Two problems, two fixes:**

1. **Wrong formatter:** `PlotlyFormatter` registered a static HTML formatter for `FigureWidget`. Since `FigureWidget` appears before `anywidget.AnyWidget` in the MRO, the static formatter always won — rendering FigureWidget as a non-interactive `marimo-plotly` component. The widget comm was never initialized, so interactive features like plotly-resampler's dynamic resampling never worked. **Fix:** Remove the `FigureWidget` registration from `PlotlyFormatter` so it falls through to the `AnyWidgetFormatter`.

2. **Stale widget state:** Plotly's `FigureWidget` maintains a split internal model — figure data lives in `_data`/`_layout_obj`, but widget traits (`_widget_data`, `_widget_layout`) are only synced during `_repr_mimebundle_()`. Without this call, the widget comm sends stale/empty state to the frontend. This is critical for plotly-resampler, which populates downsampled data during this sync. **Fix:** Call `_repr_mimebundle_()` in `from_anywidget()` before creating the wrapper, ensuring widget traits reflect the current figure state.

Closes #4091

### DIFF
--- a/marimo/_plugins/ui/_impl/from_anywidget.py
+++ b/marimo/_plugins/ui/_impl/from_anywidget.py
@@ -15,6 +15,7 @@ from typing import (
 
 import marimo._output.data.data as mo_data
 from marimo import _loggers
+from marimo._messaging.mimetypes import KnownMimeType
 from marimo._output.rich_help import mddoc
 from marimo._plugins.ui._core.ui_element import InitializationArgs, UIElement
 from marimo._plugins.ui._impl.comm import MarimoComm
@@ -83,15 +84,6 @@ def from_anywidget(widget: AnyWidget) -> UIElement[Any, Any]:
     """Create a UIElement from an AnyWidget."""
     el = _cache.get(widget)
     if el is None:
-        # Sync widget state before creating the wrapper.
-        # Some widgets (e.g. plotly FigureWidget, plotly-resampler) only sync
-        # their internal data to widget traits during _repr_mimebundle_().
-        # Without this, the comm's initial state may be stale/empty.
-        #
-        # NOTE: If you are a widget author and need to sync state from your widget,
-        # do not do this in the repr_mimebundle method.
-        # This is not a supported pattern and may break in the future.
-        _sync_widget_state(widget)
         el = anywidget(widget)
         _cache.add(widget, el)  # type: ignore[no-untyped-call, unused-ignore, assignment]  # noqa: E501
     return el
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: marimo/_plugins/ui/_impl/from_anywidget.py [hunk 1]
```diff
@@ -82,11 +83,38 @@ def from_anywidget(widget: AnyWidget) -> UIElement[Any, Any]:
     """Create a UIElement from an AnyWidget."""
     el = _cache.get(widget)
     if el is None:
+        # Sync widget state before creating the wrapper.
+        # Some widgets (e.g. plotly FigureWidget, plotly-resampler) only sync
+        # their internal data to widget traits during _repr_mimebundle_().
+        # Without this, the comm's initial state may be stale/empty.
+        _sync_widget_state(widget)
```
mscolnick: @manzt, thoughts on special casing this for plotly? or i wonder if this is other things we may see/hit with other anywidgets. 

calling `_repr_mimebundle_` should be cheap anyways

manzt: I think this is a fine fix, but want to call out that it might be a partial solution.

Calling `_repr_mimebundle_()` eagerly at creation time works for plotly's trait-syncing case, but I could imagine a widget that does something meaningful at display time, e.g.:

```python
widget = mo.ui.anywidget(FooWidget(...))  # _repr_mimebundle_ fires here now
widget.update_something()
widget.update_something()
widget  # but it should really fire here
```

anywidget will always use `_repr_mimebundle_` because that's the protocol/convention for rendering.

PR #8431, opened 2026-02-23
fix: more sql prepared statements and quoting

mscolnick:
- Fix unsafe SQL identifier construction across ClickHouse, Redshift, Ibis, and DuckDB engines by replacing raw string interpolation with parameterized queries and proper identifier quoting
- Add shared `sql_quoting` module with dialect-aware identifier quoting (double-quote for DuckDB/Redshift/Postgres, backtick for ClickHouse/MySQL) and a parser for quoted fully-qualified table names
- Add `execute_duckdb_sql` helper that supports parameterized queries (`$1`, `$2`, ...) while preserving the kernel globals context that `wrapped_sql` provides

### DIFF
--- a/marimo/_sql/sql_quoting.py
+++ b/marimo/_sql/sql_quoting.py
@@ -56,6 +56,15 @@ def parse_fully_qualified_table_name(
     Raises ValueError for malformed input (unterminated quotes, stray quotes,
     wrong number of parts).
     """
+    # Fast path for simple unquoted identifiers (no quotes)
+    if '"' not in fully_qualified_table_name:
+        parts = fully_qualified_table_name.split(".")
+        if len(parts) != 3:
+            raise ValueError(
+                f"Invalid fully qualified table name: {fully_qualified_table_name}"
+            )
+        return parts[0], parts[1], parts[2]
+
     # Each identifier is either:
     #   - a quoted identifier: "..." with escaped "" inside
     #   - an unquoted identifier: no dots or quotes


marimo/_sql/sql_quoting.py
```diff
@@ -0,0 +1,63 @@
...
+        escaped = identifier.replace('"', '""')
+        return f'"{escaped}"'
+    elif dialect in ("clickhouse", "mysql"):
+        # Backtick style: escape embedded ` as ``
+        escaped = identifier.replace("`", "``")
+        return f"`{escaped}`"
+    else:
+        # Default to double-quote (ANSI SQL standard)
+        escaped = identifier.replace('"', '""')
+        return f'"{escaped}"'
```
Light2Dark: I think maybe we don't want to quote by default. It can fail/treated as literals for certain databasess

marimo/_sql/sql_quoting.py
```diff
@@ -0,0 +1,63 @@
...
+        dialect: The SQL dialect. Supported: "duckdb", "redshift", "clickhouse".
+
+    Returns:
+        The properly quoted identifier string.
+    """
+    if dialect in ("duckdb", "redshift", "postgresql", "postgres"):
+        # Double-quote style: escape embedded " as ""
+        escaped = identifier.replace('"', '""')
+        return f'"{escaped}"'
+    elif dialect in ("clickhouse", "mysql"):
```
Light2Dark: should be able to include bigquery: https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/lexical
quote with backticks

marimo/_data/sql_summaries.py
```diff
@@ -138,30 +139,35 @@ def get_column_type(
     # First, get the column info and data type
     if "." in fully_qualified_table_name:
         # Fully qualified table name
-        db_name, schema_name, table_name = _parse_fully_qualified_table_name(
+        db_name, schema_name, table_name = parse_fully_qualified_table_name(
             fully_qualified_table_name
         )
```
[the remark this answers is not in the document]
mscolnick: this is maybe an existing bug that we can fix in a followup

marimo/_sql/sql_quoting.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Raises ValueError for malformed input (unterminated quotes, stray quotes,
+    wrong number of parts).
+    """
+    # Each identifier is either:
+    #   - a quoted identifier: "..." with escaped "" inside
+    #   - an unquoted identifier: no dots or quotes
+    _ident = r'(?:"([^"]*(?:""[^"]*)*)"|([^."]+))'
+    pattern = re.compile(rf"^{_ident}\.{_ident}\.{_ident}$")
+    match = pattern.fullmatch(fully_qualified_table_name)
```
Light2Dark, line 73: do you think we should handle the simple case for optimization? since im not sure a quoted identifier will get passed to this function.
something like

```python
if  '"' not in fully_qualified_name:
    parts = fully_qualified_table_name.split(".")
    if len(parts) != 3:
        raise ValueError(
            f"Invalid fully qualified table name: {fully_qualified_table_name}"
        )
    return parts[0], parts[1], parts[2]
```

mscolnick: yea. thats a fair callout. easier to read as well

PR #8433, opened 2026-02-23
fix: forward cell names and configs on external file reload (#8421)

### LINKED ISSUE #8421: Cell metadata (hide_code, names) lost on external file edit with --watch
## Describe the bug

When editing a `.mo.py` notebook externally (e.g., via Claude Code or any text editor) while `marimo edit --watch` is running, **cell-level metadata like `hide_code=True` and cell function names are silently dropped**. The file watcher detects the change and reloads the code, but cell configs and names are not forwarded to the frontend. The next time marimo saves from the UI, it overwrites the file without the externally-added decorator arguments.

### Steps to reproduce

1. Run `marimo edit --watch notebook.mo.py`
2. In a separate terminal/editor, change `@app.cell` to `@app.cell(hide_code=True)` on a cell, or rename a cell function
3. Marimo detects the change and reloads
4. The `hide_code` state is **not** reflected in the UI
5. Make any edit in the marimo UI and save — the file is rewritten **without** `hide_code=True`

**Expected**: External edits to cell decorator arguments and cell function names should be picked up and reflected in the frontend.

**Actual**: Only cell code is synced; cell configs and names are silently discarded.

### Root cause

The backend correctly parses configs during `AppFileManager.reload()`. However, `EditModeReloadStrategy.handle_reload()` in `file_change_handler.py` only reads and sends `codes()` and `cell_ids()`. The notifications sent to the frontend (`UpdateCellIdsNotification`, `UpdateCellCodesNotification`) don't include cell configs or names, so the frontend keeps its stale in-memory state. On the next save, the frontend's stale state overwrites the file.

### Workaround

Close the marimo session before editing externally, then reopen.

## Will you submit a PR?

- [x] Yes

## Environment

<details>

```
{
  "marimo": "0.20.1",
  "OS": "Linux",
  "OS Version": "6.6.87.2-microsoft-standard-WSL2",
  "Python Version": "3.13.5"
}
```

</details>

## Code to reproduce

```python
import marimo

__generated_with = "0.20.1"
app = marimo.App()

# This cell has hide_code=True added externally — it will be dropped on reload
@app.cell(hide_code=True)
def my_named_cell():
    mo.md("## This cell should be hidden and named")
    return

@app.cell
def _():
    mo.md("This cell is normal")
    return

if __name__ == "__main__":
    app.run()
```

mscolnick:
When editing a `.mo.py` notebook externally while `marimo edit --watch` is
running, cell metadata like `hide_code=True` and cell function names were
silently dropped. The file watcher detected changes and reloaded code, but
only cell codes and IDs were sent to the frontend — names and configs were
discarded. On the next UI save, the frontend's stale state would overwrite
the file without the externally-added metadata.

## Changes

**Backend:**
- Add `names` and `configs` fields to `UpdateCellCodesNotification` (with
  defaults for backward compatibility)
- Update `EditModeReloadStrategy.handle_reload()` to extract and forward
  cell names and configs in both lazy and autorun paths
- Update `AppFileManager.reload()` to detect config-only and name-only
  changes (previously only compared code)

**Frontend:**
- Update `setCellCodes` reducer to accept and apply optional `names` and
  `configs` from the notification
- Pass `names` and `configs` from the WebSocket message handler

Closes #8421

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ fe-typecheck:
 .PHONY: fe-codegen
 # 🔄 Generate frontend API
 fe-codegen:
-	uv run ./marimo development openapi > packages/openapi/api.yaml
+	uv run --python=3.12 ./marimo development openapi > packages/openapi/api.yaml
 	pnpm run codegen
 	pnpm format packages/openapi/
 
--- a/packages/openapi/api.yaml
+++ b/packages/openapi/api.yaml
@@ -94,43 +94,19 @@ components:
       title: AiCompletionRequest
       type: object
     AiConfig:
-      description: 'Configuration options for AI.
-
-
-        **Keys.**
-
-
-        - `rules`: custom rules to include in all AI completion prompts
-
-        - `max_tokens`: the maximum number of tokens to use in AI completions
-
-        - `mode`: the mode to use for AI completions. Can be one of: `"ask"` or `"manual"`
-
-        - `inline_tooltip`: if `True`, enable inline AI tooltip suggestions
-
-        - `models`: the models to use for AI completions
-
-        - `open_ai`: the OpenAI config
-
-        - `anthropic`: the Anthropic config
-
-        - `google`: the Google AI config
-
-        - `bedrock`: the Bedrock config
-
-        - `azure`: the Azure config
-
-        - `ollama`: the Ollama config
-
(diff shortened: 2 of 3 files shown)


dmadisetti:
I think we're OK to go ahead with this PR and follow with a more comprehensive solution

PR #8437, opened 2026-02-23
Fix SQL cell deps lost when duckdb fails to parse normalized SQL

### LINKED ISSUE #8436: SQL cell dependencies prevent auto-update upon widget state change
### Describe the bug

I found a small bug when trying to join two polars dataframes in a SQL cell, using a dropdown value as the join key, and an optional filter in one of the dataframes (see code example). When referencing `join_dropdown` in the join key, the cell seems to lose sight of the dependencies on `df1_filtered` and `df2`, which prevents the cell from auto-updating upon toggling the `filter` switch (this can be seen by the lack of `df1_filtered` and `df2` in the arguments of the marimo cell function). This dependency bug disappears if I hard-code the join key (see the final cell of the example). I wonder is this is some sort of SQL cell parsing bug?

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.2",
  "OS": "Darwin",
  "OS Version": "25.2.0",
  "Processor": "arm",
  "Python Version": "3.13.7",
  "Locale": "C/C",
  "Binaries": {
    "Browser": "145.0.7632.77",
    "Node": "v22.20.0",
    "uv": "0.8.15 (8473ecba1 2025-09-03)"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.12.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16",
    "pyyaml": "6.0.2",
    "starlette": "0.47.1",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "anywidget": "0.9.18",
    "duckdb": "1.4.4",
    "ibis-framework": "11.0.0",
    "loro": "1.10.3",
    "mcp": "1.22.0",
    "nbformat": "5.10.4",
    "openai": "2.15.0",
    "pandas": "2.3.1",
    "polars": "1.31.0",
    "pyarrow": "20.0.0",
    "ruff": "0.12.2",
    "sqlglot": "27.0.0",
    "vegafusion": "2.0.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.20.2"
app = marimo.App(width="medium", sql_output="polars")

with app.setup:
    import marimo as mo
    import polars as pl


@app.cell
def _():
    df1 = pl.DataFrame({
        "colA": [1.0, 2.0, 3.0, 4.0, 5.0],
        "colB": [10, 20, 30, 40, 50],
        "colC": [100, 200, 300, 400, 500],
    })

    df2 = pl.DataFrame({
        "colA": [1.0, 2.0, 3.0, 4.0, 5.0],
        "colB": [10, 20, 30, 40, 50],
        "colD": [1000, 2000, 3000, 4000, 5000],
    })


    join_dropdown = mo.ui.dropdown(options=["colA", "colB"], value="colA", label="Join key")
    filter = mo.ui.switch(value=False, label="Filter")
    mo.hstack([join_dropdown, filter], justify="start")
    return df2, filter, join_dropdown


@app.cell
def _(filter):
    filter_condition = "WHERE colB < 35" if filter.value else None

    df1_filtered = mo.sql(
        f"""
        SELECT
            *
        FROM df1
        {filter_condition}
        """
    )
    return (df1_filtered,)


@app.cell(hide_code=True)
def _():
    mo.md(r"""
    Below example performs a join with the key being taken from `join_dropdown.value`. As you can see the filter toggle doesn't auto-update this cell when it should.
    """)
    return


@app.cell
def _(join_dropdown):
    _df = mo.sql(
        f"""
        SELECT
            {join_dropdown.value},
            colC,
            colD
        FROM
            df1_filtered
            JOIN df2 USING ({join_dropdown.value})
        """
    )
    return


@app.cell(hide_code=True)
def _():
    mo.md(r"""
    Below example performs a join with the key being hard-coded to `colA`. This now auto-updates upon toggling the filter.
    """)
    return


@app.cell
def _(df1_filtered, df2):
    _df = mo.sql(
        f"""
        SELECT
            colA,
            colC,
            colD
        FROM
            df1_filtered
            JOIN df2 USING (colA)
        """
    )
    return

if __name__ == "__main__":
    app.run()
```

manzt:
Fixes #8436

SQL cells use duckdb to split multi-statement queries before extracting
refs with sqlglot. When duckdb couldn't parse the SQL, we bailed out
entirely and could lose dependency information. Now when duckdb parsing
fails, we fall through to sqlglot on the full SQL string instead of
returning early, which can still salvage table refs.

### DIFF
--- a/marimo/_ast/sql_visitor.py
+++ b/marimo/_ast/sql_visitor.py
@@ -526,9 +526,13 @@ def find_sql_refs(sql_statement: str) -> set[SQLRef]:
             table=table_name, schema=schema_name, catalog=catalog_name
         )
 
-    # May raise a ParseError
-    with _loggers.suppress_warnings_logs("sqlglot"):
-        expression_list = parse(sql_statement, dialect="duckdb")
+    from sqlglot.errors import ParseError
+
+    try:
+        with _loggers.suppress_warnings_logs("sqlglot"):
+            expression_list = parse(sql_statement, dialect="duckdb")
+    except ParseError:
+        return set()
 
     refs: set[SQLRef] = set()
 
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -655,17 +655,6 @@ class ScopedVisitor(ast.NodeVisitor):
             ):
                 import duckdb  # type: ignore[import-not-found,import-untyped,unused-ignore] # noqa: E501
 
-                # Import ParseError outside try block so we can except it
-                # Use a Union approach to handle both cases
-                ParseErrorType: type[Exception]
-                if DependencyManager.sqlglot.has():
-                    from sqlglot.errors import ParseError as SQLGLOTParseError
-
-                    ParseErrorType = SQLGLOTParseError
-                else:
-                    # Fallback when sqlglot is not available
-                    ParseErrorType = Exception
-
                 # TODO: Handle other SQL languages
                 # TODO: Get the engine so we can differentiate tables in diff engines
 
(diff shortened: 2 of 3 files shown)


mscolnick (dismissed):
thanks for the fix. this is definitely more correct

manzt:
> thanks for the fix. this is definitely more correct

Some failing tests... looking into it

PR #8441, opened 2026-02-24
storage download with signed url and vfile cleanup

Light2Dark:
## 📝 Summary

Adds a download_signed_url method which is preferred when downloading.

Alternatively, on creation of vfiles, we schedule a cleanup after 60s to delete the file from the registry. I tried to add a threading.Timer approach so we don't have to schedule this in the caller, but doing so would mean we need to handle race conditions.

Small frontend fix to set the error if any.

## 📋 Checklist

### DIFF
--- a/marimo/_data/_external_storage/models.py
+++ b/marimo/_data/_external_storage/models.py
@@ -13,7 +13,7 @@ from marimo._utils.assert_never import log_never
 KNOWN_STORAGE_TYPES = Literal[
     "s3", "gcs", "azure", "http", "file", "in-memory"
 ]
-SIGNED_URL_EXPIRATION = 3600
+SIGNED_URL_EXPIRATION = 60
 
 
 # Note: We may want to consolidate with FileInfo from _server/models/files.py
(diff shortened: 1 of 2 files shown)


marimo/_data/_external_storage/models.py
```diff
@@ -13,6 +13,7 @@
 KNOWN_STORAGE_TYPES = Literal[
     "s3", "gcs", "azure", "http", "file", "in-memory"
 ]
+SIGNED_URL_EXPIRATION = 3600
```
mscolnick: an hour seems like a long time. maybe we do 1 minute? doesn't the download happen right away?

marimo/_data/_external_storage/storage.py
```diff
@@ -249,6 +270,20 @@ def _read() -> str | bytes:
             return file.encode("utf-8")
         return file
 
+    async def sign_download_url(
+        self, path: str, expiration: int = 3600
```
mscolnick: `int = SIGNED_URL_EXPIRATION`

marimo/_data/_external_storage/storage.py
```diff
@@ -91,6 +93,25 @@ async def download(self, path: str) -> bytes:
...
+            return None
+        try:
+            return await sign_async(
+                self.store,
+                "GET",
+                path,
+                expires_in=timedelta(seconds=expiration),
+            )
+        except Exception:
+            LOGGER.debug("Failed to sign URL for %s", path)
```
mscolnick: these might be better as `LOGGER.info`

PR #8446, opened 2026-02-24
ci: install hatch with `uv` to avoid timeouts

mscolnick:


mscolnick:
this took way to long to install (~1min)

PR #8454, opened 2026-02-25
fix: optimize table search for wide dataframes

### LINKED ISSUE #8449: Table Search Functions fail or crash on wide dataframes
### Describe the bug

The Search function of the table viewer is very useful, but it fails when the table is very wide.

Specifically, if the table is less than 399 columns wide, the Search function works correctly.
Larger than 400 columns, there is no result, until you get to very wide tables - I can get consistent kernel crashes at 2000 columns wide.

Here is a demonstration of the bug.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

[wide_dataframe.py](https://github.com/user-attachments/files/25524867/wide_dataframe.py)

stakeswky:
## Summary

Fixes #8449 — table search crashes/fails on wide dataframes (400+ columns, kernel crash at 2000+).

## Root Cause

The `search()` method in both `NarwhalsTableManager` and `PolarsTableManager` built the filter expression by iteratively chaining `|` operators:

```python
or_expr = expressions[0]
for expr in expressions[1:]:
    or_expr = or_expr | expr
```

This creates a deeply nested binary expression tree — one level of nesting per column. At ~400 columns the expression tree exceeds recursion/stack limits, causing silent failures or kernel crashes.

## Fix

Replace the iterative OR chaining with `any_horizontal()`, which evaluates all expressions in a single flat call:

- `narwhals_table.py`: `nw.any_horizontal(*expressions, ignore_nulls=True)`
- `polars_table.py`: `pl.any_horizontal(*expressions)`

This is both more efficient and avoids the deep nesting entirely.

## Testing

All existing tests pass (`test_narwhals.py`, `test_polars_table.py`, `test_pandas_table.py`). Verified manually with 500-column DataFrames.

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/narwhals_table.py
+++ b/marimo/_plugins/ui/_impl/tables/narwhals_table.py
@@ -329,11 +329,9 @@ class NarwhalsTableManager(
         if not expressions:
             return NarwhalsTableManager(self.data.filter(nw.lit(False)))
 
-        or_expr = expressions[0]
-        for expr in expressions[1:]:
-            or_expr = or_expr | expr
-
-        filtered = self.data.filter(or_expr)
+        filtered = self.data.filter(
+            nw.any_horizontal(*expressions, ignore_nulls=True)
+        )
         return NarwhalsTableManager(filtered)
 
     def get_stats(self, column: str) -> ColumnStats:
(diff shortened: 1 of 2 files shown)


Light2Dark:
Hi! Thanks for the PR, I made a similar change earlier with tests https://github.com/marimo-team/marimo/pull/8450. I think we can merge that in instead unless you have other additions?

PR #8097, opened 2026-02-03
feat: rich notebook previews through OpenGraph metadata

peter-gy:
## 📝 Summary

Adds notebook-level OpenGraph metadata (PEP 723 + [Next.js-inspired](https://nextjs.org/docs/app/api-reference/functions/generate-metadata) optional generator hook) with thumbnail generation plumbing. As a first use case, wired it up to improve gallery cards in `marimo run <dir>`.

Follow-up of #8056.

## 🔍 Description of Changes

- Introduced `OpenGraphMetadata` for notebooks via PEP 723: `[tool.marimo.opengraph]` with `title`, `description`, `image`, and optional `generator`
  - `image` supports either an HTTPS URL or a notebook-relative path
- Added Next.js-style merge semantics for generators: static PEP 723 fields are the base; generator can override only the fields it returns
- Added a canonical thumbnail endpoint: `GET /api/home/thumbnail?file=...`
  - redirects to HTTPS images
  - serves notebook-relative images only from the notebook’s `__marimo__/` directory
  - falls back to a branded placeholder SVG when no image exists
- Inject OG tags into notebook pages (`og:title`, `og:description`, `og:image`)
- Gallery cards now prefer `FileInfo.opengraph` (title/description/image) and fall back to the existing title-casing behavior
- Added `marimo tools thumbnails generate ...` to write thumbnails to `__marimo__/assets/<stem>/opengraph.png`
  - default is `--no-execute` (fast; no cell outputs)
  - opt-in `--execute` to include outputs
  - opt-in `--sandbox` (only applies with `--execute`)

## 📸 Thumbnails

> in the examples below `marimo-team/learn` is a local clone of https://github.com/marimo-team/learn

### Auto-generated thumbnails without `playwright`

No file gets written to disk. The `/thumbnail` API endpoint calls `DEFAULT_OPENGRAPH_PLACEHOLDER_IMAGE_GENERATOR` to construct an SVG dynamically and serves it.

Screenshot 2026-02-03 at 14 36 25

### Auto-generated thumbnails with `playwright`, but without notebook execution

Renders code blocks without outputs, similar to Observable, as suggested by @manzt.

```
 marimo tools thumbnails generate marimo-team/learn
 marimo run marimo-team/learn
```

Screenshot 2026-02-03 at 14 40 50

### Auto-generated thumbnails with `playwright`, with sandboxed notebook execution for cell outputs

[...]

PR #8458, opened 2026-02-25
feat: render web elements in PDF export

peter-gy:
## 📝 Summary

Allows calling `marimo export pdf` with `--rasterize-outputs` flag, with which users can include a rasterized representation of their interactive components (marimo webcomponents, anywidgets, etc.) in both `--webpdf` and `--no-webpdf` mode (web vs. LaTeX export).

## 🔍 Description of Changes

The implementation consists of two main components.

First, we need to detect and rasterize eligible components. To do so, we dynamically spin up either a static html server (similar to marimo-team/marimo#8097 for thumbnail capture) or a headless, live marimo server if a python runtime is needed to hydrate a component.

Second, once we have access to these components as PNGs, we inject them into the `nbformat.NotebookNode` before passing it to `nbconvert` and let `nbconvert` handle the conversion natively, as if cell outputs with image mimetype were there all along.

## Results

I used this notebook for local tests: [demo.py](https://github.com/user-attachments/files/25542836/demo.py) (https://molab.marimo.io/notebooks/nb_r4LQZFDhURekMpjsjvu7ym).

image

image

image

This feature pairs nicely with the `--no-include-inputs` flag from marimo-team/marimo#8362, enabling cleaner exports focusing purely on the rendering results and not the input code blocks:

Screenshot 2026-02-25 at 11 35 57

### DIFF
--- a/docs/guides/exporting.md
+++ b/docs/guides/exporting.md
@@ -152,13 +152,32 @@ Rasterization runs only when outputs are included (the default `--include-output
 
 !!! warning "Choose the raster server mode carefully"
 
-    marimo gives you explicit control over raster capture mode.
-    Use `--raster-server=live` when the first meaningful render of a widget or component requires an active Python connection.
+    marimo gives you control over how output is captured.
+    Use `--raster-server=live` when a widget needs Python to finish rendering.
     Otherwise, prefer the default `--raster-server=static`.
 
-    ```bash
-    --raster-server=live
-    ```
+The notebook below is a concrete case where `--raster-server=live` helps.
+
+/// marimo-embed-file
+    size: xlarge
+    mode: edit
+    filepath: examples/outputs/live_raster.py
+///
+
+This widget starts at `Initializing...` and then updates to
+`count is ... from ... host` after it receives data from Python.
+Static capture can freeze the initial placeholder; live capture gets the
+updated output.
+
+```bash
+# Static mode captures only the initial "Initializing..." placeholder
+marimo export pdf examples/outputs/live_raster.py \
+  -o live-raster-static.pdf --raster-server=static --no-sandbox --no-include-inputs
+
+# Live mode captures the updated widget output
+marimo export pdf examples/outputs/live_raster.py \
+  -o live-raster-live.pdf --raster-server=live --no-sandbox --no-include-inputs
+```
 
 !!! note "Rasterization dependencies"
 
--- a/examples/outputs/live_raster.py
+++ b/examples/outputs/live_raster.py
@@ -1,0 +1,95 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = [
+#     "anywidget==0.9.21",
+#     "marimo>=0.20.2",
+#     "traitlets==5.14.3",
+# ]
+# ///
+import marimo
+
+__generated_with = "unknown"
+app = marimo.App(width="medium")
+
+
+@app.cell(hide_code=True)
+def _(CounterWidget):
+    CounterWidget(count=42)
+    return
+
+
+@app.cell
+def _(anywidget, os, traitlets):
+    class CounterWidget(anywidget.AnyWidget):
+        _esm = """
+        export default async () => {
+          let hostName = null;
+
+          return {
+            initialize({ model }) {
+              // This message gets handled by _handle_custom_msg on the Python side
+              model.send({ event: "requestHostName" });
+            },
+
+            render({ model, el }) {
+              let count = () => model.get("count");
+              let btn = document.createElement("button");
+              btn.classList.add("counter-button");
+              btn.innerHTML = `Initializing...`;
+
+              // Set proper HTML content once message arrives from Python connection
+              model.on("msg:custom", (msg, buffers) => {
+                hostName = msg.response;
+                btn.innerHTML = `count is ${count()} from ${hostName} host`;
+              });
+
+              btn.addEventListener("click", () => {
+                model.set("count", count() + 1);
+                model.save_changes();
+              });
+
+              model.on("change:count", () => {
+                btn.innerHTML =
+                  hostName
+                    ? `count is ${count()} from ${hostName} host`
+                    : `Initializing...`;
+              });
+
+              el.appendChild(btn);
+            },
+          };
+        };
+        """
+        _css = """
+        .counter-button {
+          background: #387262;
+          border: 0;
+          border-radius: 10px;
+          padding: 10px 50px;
+          color: white;
+        }
+        """
+        count = traitlets.Int(0).tag(sync=True)
+
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
+            self.on_msg(self._handle_custom_msg)
+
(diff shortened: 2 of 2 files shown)


Light2Dark:
Could we have rasterized outputs by default? Feels like a good addition

docs/guides/exporting.md
```diff
@@ -134,11 +134,28 @@ enables exporting to PDF and other formats with Pandoc. See this [publishing](./
...
+
+Rasterization runs only when outputs are included (the default `--include-outputs` mode). Use `--raster-scale` (range `1.0` to `4.0`, default `4.0`) to trade export speed/file size for sharper captured output.
+
+!!! note "Rasterization dependencies"
+
+    Rasterized output capture requires Playwright and Chromium:
+
+    ```bash
+    python -m pip install playwright
+    python -m playwright install chromium
```
Light2Dark:
```suggestion
    uv pip install playwright
    playwright install chromium
```

marimo/_server/export/_live_notebook_server.py
```diff
@@ -0,0 +1,165 @@
...
+
+from marimo import _loggers
+from marimo._utils.net import find_free_port
+
+LOGGER = _loggers.marimo_logger()
+
+_LIVE_SERVER_START_TIMEOUT_S = 90.0
+_LIVE_SERVER_POLL_INTERVAL_S = 0.2
+_LIVE_SERVER_SHUTDOWN_TIMEOUT_S = 5.0
+_LIVE_SERVER_DEFAULT_PORT = 2718
```
Light2Dark: small thing, should we do 2719/2720? Since the default server runs on 2718.

marimo/_server/export/_html_asset_server.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc: BaseException | None,
+        tb: TracebackType | None,
+    ) -> None:
+        del exc_type
+        del exc
+        del tb
```
Light2Dark: Sorry not familiar with context manager code. Should we do something with these exceptions? Log/return False?

peter-gy:
> Could we have rasterized outputs by default? Feels like a good addition

I was hesitant about this and ended up designing for rasterization opt-in rather than opt out because (1) it makes exports take more time, (2) need playwright as additional dependency by default and (3) would introduce silent behavior change: running the same `marimo export pdf ...` before and after this change would have different dependency requirements and content-wise different outputs.

Do you think it would worth it it go with `--rasterize-outputs` by default nonetheless?

marimo/_server/export/_html_asset_server.py
peter-gy: Great point. We don’t actually suppress exceptions here as `__exit__` returns None (same as `False` for context managers), so exceptions from inside the `with` block still propagate. I updated this for clarity by removing the `del` pattern.

Light2Dark:
> I was hesitant about this and ended up designing for rasterization opt-in rather than opt out because (1) it makes exports take more time, (2) need playwright as additional dependency by default and (3) would introduce silent behavior change: running the same `marimo export pdf ...` before and after this change would have different dependency requirements and content-wise different outputs.
> 
> Do you think it would worth it it go with `--rasterize-outputs` by default nonetheless?

I personally prefer it, but Akshay should have better opinions. Do you have a rough estimate of the time taken? I think the dependency change is fine (maybe if we don't detect it, we skip this process).

peter-gy:
> > I was hesitant about this and ended up designing for rasterization opt-in rather than opt out because (1) it makes exports take more time, (2) need playwright as additional dependency by default and (3) would introduce silent behavior change: running the same `marimo export pdf ...` before and after this change would have different dependency requirements and content-wise different outputs.
> > Do you think it would worth it it go with `--rasterize-outputs` by default nonetheless?
> 
> I personally prefer it, but Akshay should have better opinions. Do you have a rough estimate of the time taken? I think the dependency change is fine (maybe if we don't detect it, we skip this process).

The time the export takes depends heavily on the notebook itself, as it gets executed in a headless environment so that the screenshots can be captured. E.g. the EmbeddingAtlas-based example can take 1-2 minutes for the dataset download, embedding model download, actual embedding of 10k+ data points. On a related note, some notebooks have side effects (e.g. API calls, file writes, etc.) and as a developer of such a notebook already using `marimo export pdf ...` in an existing workflow, I would be unpleasantly surprised if from one release to another, executing the same command triggers these side effects without my explicit approval (without my opt-in).

Light2Dark:
That makes a lot of sense, I agree. Side thoughts: I felt a static html server was sufficient for rendering many marimo components, only need a live server for some specific ones 🤔, like how we hydrate outputs from session.json

peter-gy:
> That makes a lot of sense, I agree. Side thoughts: I felt a static html server was sufficient for rendering many marimo components, only need a live server for some specific ones 🤔, like how we hydrate outputs from session.json

Indeed, based on what I saw from local tests, live server is only needed when output needs runtime round-trips during initial render (e.g., widget initialization needing backend comm before final paint).

That's why in `_pdf_raster.py` I implemented a (probably naive) discovery logic deciding whether the target requires static of live server. I'm not very familiar with the session.json construction yet, so I'll look into that part of the codebase to see if I can tighten the logic and opt more often for cheap static capture instead of spinning up a live server.

peter-gy:
All these make sense. I enabled rasterization by default now.

> `marimo export pdf` already runs the notebook, no?

Yes, I should have phrased things more precisely. The nuance I meant is that previously there were no cases where a dedicated live server got launched for an export in addition, so now, if there is a notebook which has some components where static server suffices but some others requiring a live server, the notebook would actually get executed an additional time.

To give users more clarity and control, I exposed a CLI option to set if they want the `live` vs. `static` rasterizer server to be used (we default to `static` to keep thing faster)

Light2Dark:
I think the code looks fine, there are a few cli commands being added. It makes sense to me, but highlighting if we think this is too many for users.
- rasterize_outputs
- raster_scale
- raster_server

Light2Dark (approved):

peter-gy:
> In the documentation, can you provide a concrete example for when a user might prefer a live server?

I embedded a concrete example into the docs to demo when a live server is required to capture the "proper" widget state.

akshayka (approved):
Great work! Clean code and a big feature unlock.

I noticed one thing, and it's possible nothing can be done about it But interpolated (f-string) Markdown does not get rendered as Markdown. This is still an improvement since at least the embedded HTML objects get rendered. But just thought I would mention, just in case.

Image

peter-gy:
> Great work! Clean code and a big feature unlock.
> 
> 
> 
> I noticed one thing, and it's possible nothing can be done about it But interpolated (f-string) Markdown does not get rendered as Markdown. This is still an improvement since at least the embedded HTML objects get rendered. But just thought I would mention, just in case.
> 
> 
> 
> Image

Thank you for the feedback!

I can look into this in a follow up. I also noticed various edge cases where we can improve things, just didn't want to bloat the scope of this PR.

PR #8466, opened 2026-02-25
Show smart previews for SQL and markdown cells in minimap

### LINKED ISSUE #8457: Show first content line of unnamed MD and SQL cells in dependency minimap
### Description

In the minimap unnamed md and sql cells show up as:

`mo.md(f"""` and `mo.sql(f"""` respectively.

### Suggested solution

If the minimap would show the first line of code instead, it would the start of the sql / md text, or one could add a meaningful comment at the front of the content.

`select * from cars` or `-- get cars list for the table` would be more helpful information.

- [ ] Yes

### Alternatives

Named SQL cells show the name, which is probably best, but not all cells need to be named, some are just experiments during development, and naming them seems like overkill. 

Even more so for most markdown cells.

### Additional context

_No response_

manzt:
Fixes #8457

The dependency minimap previously showed the raw first line of every cell, which meant SQL and markdown cells displayed unhelpful boilerplate like `mo.md(f"""` or `_df = mo.sql(f"""`. These changes extract the inner content (using smart-cells helpers) to show the first meaningful line instead.

We could consider some alternative preview (e.g., a summary comment or visual indicator of cell type), but my experiments led to a lot of visual noise/clutter and if found the stripping simple & effective.

Screenshot 2026-02-25 at 2 01 49 PM

### DIFF
--- a/frontend/src/components/dependency-graph/utils/cell-preview.ts
+++ b/frontend/src/components/dependency-graph/utils/cell-preview.ts
@@ -1,0 +1,46 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+
+import { MarkdownParser, SQLParser } from "@marimo-team/smart-cells";
+
+export interface CellPreview {
+  text: string | undefined;
+  type: "python" | "markdown" | "sql";
+}
+
+const markdownParser = new MarkdownParser();
+const sqlParser = new SQLParser();
+
+function firstNonEmptyLine(content: string): string | undefined {
+  for (const line of content.split("\n")) {
+    const trimmed = line.trim();
+    if (trimmed.length > 0) {
+      return trimmed;
+    }
+  }
+  return undefined;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


frontend/src/components/dependency-graph/utils/cell-preview.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { MarkdownParser, SQLParser } from "@marimo-team/smart-cells";
+
+export interface CellPreview {
+  text: string | undefined;
+  type: "python" | "markdown" | "sql";
+}
+
+const markdownParser = new MarkdownParser();
+const sqlParser = new SQLParser();
```
Light2Dark, line 11: one alternative, but I think your solution is better, is to pull LanguageAdapters.markdown.transformIn
but I think the sql one has side effects.

Light2Dark:
CI failure is unrelated

PR #8474, opened 2026-02-26
fix: detect symlinked files inside assets directory for follow_symlink warning

mscolnick:
The previous check only detected when the assets/ directory itself was a
symlink. With pdm/uv, the directory is real but files inside are symlinks,
causing silent 404s. Now samples files within the directory too.

### DIFF
--- a/marimo/_server/api/endpoints/assets.py
+++ b/marimo/_server/api/endpoints/assets.py
@@ -63,7 +63,7 @@ def _has_symlinks(directory: Path) -> bool:
         for i, child in enumerate(directory.iterdir()):
             if child.is_symlink():
                 return True
-            if i >= 5:
+            if i >= 1:
                 break
     except OSError:
         pass


mscolnick:
cc @drewp

marimo/_server/api/endpoints/assets.py
```diff
@@ -53,9 +53,28 @@
...
+def _has_symlinks(directory: Path) -> bool:
+    """Check if a directory is a symlink or contains symlinked files."""
+    if directory.is_symlink():
+        return True
+    try:
+        # Check a small sample of files for symlinks
+        for i, child in enumerate(directory.iterdir()):
+            if child.is_symlink():
+                return True
+            if i >= 1:
```
dmadisetti, line 66: 1 directory is not a very large sample? would maybe move to kwarg if we need to change it in the future

mscolnick: this is 2 items. i don't want to read a bunch of files on startup

mscolnick: and only provides a warning has not functional changes

dmadisetti (approved):

PR #8483, opened 2026-02-26
fix: preserve cell config (hidden/disabled) on undo delete

### LINKED ISSUE #8330: Undo delete disabled/hidden cell does not restore its original state
### Describe the bug

Deleting a hidden or disabled cell and undoing it will restore the cell, but it will not restore the hidden/disabled state.

### Will you submit a PR?

- [ ] Yes

### Environment

Marimo 0.19.11 on molab.

### Code to reproduce

_No response_

mscolnick:
Fixes #8330. When a hidden or disabled cell was deleted and then the deletion was undone, the cell was restored but its `hide_code` and `disabled` config was lost — it would revert to the default (visible, enabled). The root cause was that the deletion history entry did not store the cell's config, so `undoDeleteCell` called `createCell()` without it, resulting in default values.

### DIFF
--- a/frontend/src/core/cells/cells.ts
+++ b/frontend/src/core/cells/cells.ts
@@ -591,7 +591,7 @@ const {
 
     // release the granular atom(s) created for this cell
     releaseCellAtoms(cellId);
-    const prevData = state.cellData[cellId] || { name: "", config: {} };
+    const prevData = state.cellData[cellId];
     return {
       ...state,
       cellIds: state.cellIds.deleteById(cellId),


frontend/src/components/editor/chrome/panels/outline/floating-outline.tsx
```diff
@@ -30,7 +30,7 @@ export const FloatingOutline: React.FC = () => {
       onMouseEnter={() => setIsHovered(true)}
       onMouseLeave={() => setIsHovered(false)}
       className={cn(
-        "fixed top-[25vh] right-8 z-10000",
+        "fixed top-[25vh] right-8 z-10000 print:hidden",
```
mscolnick, line 33: drive by

frontend/src/core/cells/cells.ts
```diff
@@ -590,17 +591,19 @@ const {
 
     // release the granular atom(s) created for this cell
     releaseCellAtoms(cellId);
+    const prevData = state.cellData[cellId] || { name: "", config: {} };
```
Light2Dark: it would throw an error anyway if cellId is not present. And we have accessed it prior

```suggestion
    const prevData = state.cellData[cellId];
```

Light2Dark (approved):

PR #8491, opened 2026-02-27
Pass cwd to agents from backend

Light2Dark:
## 📝 Summary

Previously, the agent would not know the absolute filepath, this led to errors when an agent is instantiated outside of the current directory. This fix ensures current working directory is passed from the backend on-mount, and subsequently passed to the agent.

Screenshot 2026-03-03 at 11 31 58 PM

## 📋 Checklist


mscolnick:
it's unfortunate to have to pass filename and filepath. can these be merged? I thought filename was absolute? should we force it? 

Another option is to call the /files api which gives you the root iirc

Light2Dark:
> it's unfortunate to have to pass filename and filepath. can these be merged? I thought filename was absolute? should we force it?
> 
> Another option is to call the /files api which gives you the root iirc

It's a bit tricky, we started passing the relative paths instead of absolute path for filename (where possible).  https://github.com/marimo-team/marimo/pull/7722 Not sure if I want to touch this.

The other option, I think we could use files_details endpoint from `file_explorer`, but this will fetch contents of the file as well. Maybe can create another light endpoint to get absolute path

mscolnick:
@Light2Dark got it, this is just for home mode? maybe we could just make those absolute URLs. @dmadisetti do you see any issue with that? that does help with consistency, but maybe folks used "home mode" in order to avoid absolute paths.

separately, right now for Agents we use `basedir(filename)` as the CWD. we don't need to do this now, but maybe it would be better to actually use the `CWD`, in which the better long-term approach would be to pass that down instead of `filepath`.

manzt (approved):

manzt:
I think I like the idea of `cwd`

dmadisetti (approved):
No issues passing the data, assuming our permissions work

dmadisetti:
The windows failure might be real (file path issues)

Light2Dark:
I pushed a fix but not showing up here. I think its a test issue, it's escaping \ with \ so shows up as `\\`

PR #7863, opened 2026-01-15
feat: add `--session-ttl` flag to marimo edit

AnirudhDagar:
## 📝 Summary

Fix #7834

## 🔍 Description of Changes
This PR adds the `--session-ttl` flag to `marimo edit` (extending the functionality from the run mode), enabling automatic cleanup of inactive sessions on multi-user edit servers.

When set, sessions are automatically closed after the specified number of seconds
following a websocket disconnect.

Usage:

```bash
marimo edit --session-ttl 300  # Close sessions 5 minutes after disconnect
```

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] Tests have been added for the changes made.
- [x] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Pull request title is a good summary of the changes - it will be used in the [release notes](https://github.com/marimo-team/marimo/releases).

PR #8495, opened 2026-02-27
fix: memory leak in ASGI session cleanup

mscolnick:
Session cleanup for ASGI apps was broken after #7863 refactored session TTL handling. The `_on_disconnect` handler checked `manager.ttl_seconds` which is None by default for `create_asgi_app()`, causing sessions to never be closed on WebSocket disconnect.

This also forces a non-deafult for `create_asgi_app`'s session_ttl parameter, for another layer of safety.


dmadisetti (dismissed):

dmadisetti:
Teardown may still be doing strange things, tests seem especially flaky- I've run them 4 times

PR #8503, opened 2026-02-27
Disable "Open in Vega Editor" link in `ChartBuilder`

### LINKED ISSUE #8501: configure ChartBuilder to remove link "open in vega editor"
### Description

I'm hoping to configure the ChartBuilder somehow to disable the "Open in Vega Editor" link in  the chart created with ChartBuilder.

### Suggested solution

Vega exposes [options for vega-embed](https://vega.github.io/vega-embed/#options).
In my case i would like to overwrite the default setting at one point to remove the "Open in Vega Editor" link

```
var spec = { /* ... your vega-lite JSON here ... */ };

// Embed the chart and configure the actions menu
vegaEmbed('#vis', spec, {
  actions: {
    export: true,   // "Save as SVG/PNG"
    source: true,   // "View Source"
    editor: false   // "Open in Vega Editor"    <--
  }
});
```

**Suggestion 1:**
A solution could be a configuration option in Notebook Settings, maybe in easiest case with custom Files.

One could leave a `custom-chartbuilder-options.json` according to [options for vega-embed](https://vega.github.io/vega-embed/#options), which will be used as embedOptions in the [VegaPlugin](https://github.com/marimo-team/marimo/blob/12b4f4e1f6c3cc9f3bc57f48e79c2b9e29309f8e/frontend/src/plugins/impl/vega/VegaPlugin.tsx#L27)

**Suggestion 2:**
When using altair for a single chart it's possible to set this configuration i.e.
```py
import altair as alt

chart = alt.Chart(plot_df).mark_bar().encode(x='City', y='Region').properties(
    usermeta={'embedOptions': {'actions': {'editor': False, 'source': False}}}
)
chart
```
A way to set this 'properties' for the chartbuilder object in a cell would be nice.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

manzt:
Fixes #8501

The ChartBuilder generates charts as Vega-Lite specs with the intended workflow being: build a chart visually, then copy the generated Altair code into a cell. Opening the raw Vega JSON in an external editor doesn't fit this flow.

### DIFF
--- a/frontend/src/components/data-table/charts/lazy-chart.tsx
+++ b/frontend/src/components/data-table/charts/lazy-chart.tsx
@@ -37,7 +37,7 @@ export const LazyChart: React.FC<{
               export: true,
               source: false,
               compiled: false,
-              editor: true,
+              editor: false,
             },
             mode: "vega",
             renderer: "canvas",


manzt:
> it may not solve the original issue completely, since the generated code still contains this.

@Light2Dark isn't the generated code altair code? I think our VegaPlugin omits the action by default.

PR #8508, opened 2026-02-27
fix: clear QueueDistributor consumers on stop to prevent memory leaks

mscolnick:
- Rename public fields to private (_consumers, _queue, _thread)
- Set _stop flag in stop() before sending sentinel
- Clear consumer list when loop exits to release captured references
  (e.g. session closures that keep large data alive)


marimo/_utils/distributor.py
```diff
@@ -114,43 +114,50 @@ def flush(self) -> None:
...
-        self.thread = threading.Thread(target=self._loop, daemon=True)
-        self.thread.start()
-        return self.thread
+        self._thread = threading.Thread(target=self._loop, daemon=True)
+        self._thread.start()
+        return self._thread
 
     def stop(self) -> None:
-        self.queue.put_nowait(None)
+        self._stop = True
```
dmadisetti, line 171: This leave an RC that could leave values + `None` on the queue. Maybe that's fine? @copilot do we ever start the distributor again? Because we could stop from the previous case

dmadisetti: Oh that's annoying. I thought you'd just respond in a comment

marimo/_utils/distributor.py
```diff
@@ -114,43 +114,81 @@ def flush(self) -> None:
...
+            self._consumers.clear()
+
     def start(self) -> threading.Thread:
-        self.thread = threading.Thread(target=self._loop, daemon=True)
-        self.thread.start()
-        return self.thread
+        # Reset state so the distributor can be restarted on the same queue.
+        # Drain any leftover None sentinels from a previous stop().
+        self._stop = False
+        self._drain_sentinels()
```
dmadisetti, line 165: Double checked, currently _we do not_ restart the queue. So we could just clear it and move on. This is fine though

marimo/_utils/distributor.py
```diff
@@ -114,43 +114,81 @@ def flush(self) -> None:
...
+
+        Only None values are removed; real messages are put back.
+        """
+        recovered: list[T] = []
+        try:
+            while True:
+                item = self._queue.get_nowait()
+                if item is not None:
+                    recovered.append(item)
+        except Exception:
```
dmadisetti:
```suggestion
        except queue.Empty:
```

PR #8510, opened 2026-02-27
Replace hatch env layer with native uv commands

manzt:
Many of the APIs are similar and I'm curious what this would look like in CI. Converts hatch environments to PEP 735 dependency groups and replaces `hatch run` with `uv run --group` across Makefile, scripts, workflows, and docs. Keeps `hatch publish` for releases.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,56 +20,26 @@ contribution._ Get in touch at
 or [on Discord](https://marimo.io/discord?ref=contributing).
 
 ## Setup
+
 _Note: We recommend that Windows developers use [WSL](https://learn.microsoft.com/en-us/windows/wsl/install) and clone the marimo repository [into the WSL environment and not the Windows mount](https://learn.microsoft.com/en-us/windows/wsl/filesystems)._
 
-Install [pixi](https://github.com/prefix-dev/pixi) to manage your development environment. The following command uses `pixi` to launch a development shell with all dependencies installed.
+### Prerequisites
 
-> [!NOTE]
->
-> As an alternative to installing `pixi`, you can try developing in [Gitpod](https://gitpod.io/#https://github.com/marimo-team/marimo).
-> Note that developing in Gitpod is not officially supported by the marimo team.
+- [uv](https://docs.astral.sh/uv/getting-started/installation/) (Python package manager)
+- [Node.js](https://nodejs.org/) 20+
+- [pnpm](https://pnpm.io/installation) 9+
+
+### Getting started
+
+```bash
+make fe && make py
+make dev
+```
+
+This will build the frontend, install Python dependencies in editable mode, and launch the dev server (backend on port 2718, frontend on port 3000).
 
 > [!TIP]
-> `pixi` installs and manages both the Python and Node toolchains, then drops you into a ready shell with one command. If you already have Node 20+/pnpm 9+/uv installed, you can skip pixi and use `uv sync` directly. Typical flows:
-> - `pixi shell` → `make fe && make py` → `make dev`
-> - `uv sync` → `make fe && make py` → `make dev`
-
-```bash
-pixi shell
-```
-
-If you have the right non-python dependencies installed via other methods (e.g. homebrew) you can simply sync your `marimo` development
-environment with `uv sync`.
-
-Now you can install the environment frontend and Python dependencies.
-
-```bash
-make fe && make py
-```
-
-After doing this, you can instantiate your marimo development environment by running the following command.
-
-```bash
-make dev
-```
-
-This will launch two processes, the backend one in port 2718 and the front end one in port 3000.
-
-In summary you will need to run:
-
-```bash
-pixi shell
-make fe && make py
-make dev
-```
-
-or if not using `pixi`:
-
-```bash
-uv sync
-make fe && make py
-make dev
-```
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -269,42 +244,8 @@ dependencies = [
...
-# This is used externally from the narwhals repo to run our tests from their repo.
-# This should include any tests that may use narwhals.
-# It is ok if we test more than narwhals here, but we should not test less.
-test-narwhals = """
-    pytest{env:HATCH_TEST_ARGS:} \
-        tests/_data/ \
-        tests/_plugins/ui/_impl/ \
-        tests/_utils/test_narwhals_utils.py
-"""
-
```
manzt, line 299: moved into Makefile

development_docs/adding_lint_rules.md
```diff
@@ -274,13 +274,13 @@ This will create:
 
 ```bash
 # Run lint tests
-uv run hatch run test:test tests/_lint
+uv run --python 3.12 --group test pytest tests/_lint
```
mscolnick: i wonder if these should be make commands: `make py-test tests/_lint`

mscolnick: nvm, make wont take args

mscolnick: we can probably drop the `--python 3.12` and just leave it to `.python-version`, right?

manzt: > we can probably drop the --python 3.12 and just leave it to .python-version, right?

Yeah for the "default" version... we fan out the matrix in CI

pyproject.toml
mscolnick: this will break narwhals, so you will need to make a PR to their repo: https://github.com/narwhals-dev/narwhals/blob/ef7e90ba1f51f0d88dc2c1a34513cfbc2dac6eb0/.github/workflows/downstream_tests.yml#L125

pyproject.toml
```diff
@@ -116,8 +116,9 @@ mcp = [
 
 [dependency-groups]
 dev = [
-    "hatch>=1.16.3",
     "black~=23.12.1",
```
manzt: curious, is black used for anything? Maybe a follow up or separate but we could probably remove black (and it's config) for marimo.

akshayka: I don't think it's used anymore

pyproject.toml
```diff
@@ -184,7 +186,7 @@ test-optional = [
     "duckdb>=1.0.0",
     "ibis-framework[duckdb]>=10.3.0",
     # For testing mo.image
-    "pillow~=10.4.0",
+    "pillow>=10.4.0",
```
akshayka: Why change this?

development_docs/adding_lint_rules.md
akshayka: This could also just be `./scripts/pytest.sh tests/_lint`

pyproject.toml
manzt: Removed

pyproject.toml
manzt: It seems there are not longer wheels for some platforms and [CI failed](https://github.com/marimo-team/marimo/actions/runs/22520951583/job/65245638099?pr=8510) to build:


```
   Building pillow==10.4.0
  × Failed to build `pillow==10.4.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `backend.build_wheel` failed (exit status: 1)
```



pillow is now on 12.1.1. We could setup environment to support the build... or bump the semver `pillow~=12.1.1` if we want to support minor.

pyproject.toml
manzt: https://github.com/narwhals-dev/narwhals/pull/3487

mscolnick (dismissed):
looks good

mscolnick (approved):

CONTRIBUTING.md
```diff
@@ -22,24 +22,24 @@ or [on Discord](https://marimo.io/discord?ref=contributing).
...
+Install [pixi](https://github.com/prefix-dev/pixi) to manage your development environment. The following command uses `pixi` to launch a development shell with all dependencies installed.
 
 > [!NOTE]
 >
 > As an alternative to installing `pixi`, you can try developing in [Gitpod](https://gitpod.io/#https://github.com/marimo-team/marimo).
 > Note that developing in Gitpod is not officially supported by the marimo team.
 
 > [!TIP]
-> New to both `pixi` and `hatch`? Pick `pixi`. It installs and manages both the Python and Node toolchains, then drops you into a ready shell with one command. Choose `hatch` only if you already maintain Node 20+/pnpm 9+ yourself and just want a Python environment manager. Typical flows:
+> `pixi` installs and manages both the Python and Node toolchains, then drops you into a ready shell with one command. If you already have Node 20+/pnpm 9+/uv installed, you can skip pixi and use `uv sync` directly. Typical flows:
```
dmadisetti: nit: but I would mainly lead with this, we already get pixi reports that can be avoided. Not sure if anyone internally uses this flow

.github/workflows/release-prod.yml
```diff
@@ -39,8 +39,11 @@ jobs:
       - name: 🥚 Install Hatch
         uses: pypa/hatch@install
```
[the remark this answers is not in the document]
mscolnick: cc @manzt

PR #8513, opened 2026-02-28
improve storage protocol detection

Light2Dark:
## 📝 Summary

Checks the endpoint URL to see if a more specific protocol can be found

image

## 📋 Checklist

### DIFF
--- a/marimo/_data/_external_storage/storage.py
+++ b/marimo/_data/_external_storage/storage.py
@@ -377,6 +377,7 @@ _PROTOCOL_MAP: dict[str, KNOWN_STORAGE_TYPES] = {
 
 # Specific provider patterns checked before generic ones (e.g. S3),
 # since S3-compatible services may also have "s3" in their URL.
+# The order of the patterns is important, the first pattern that matches will be used.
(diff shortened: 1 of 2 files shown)


marimo/_data/_external_storage/storage.py
```diff
@@ -335,18 +358,49 @@ def is_compatible(var: Any) -> bool:
...
+    ("s3", "s3"),
+    ("amazonaws", "s3"),
+]
+
+
+def detect_protocol_from_url(url: str) -> CLOUD_STORAGE_TYPES | None:
+    """Detect the storage provider from an endpoint URL."""
+    url = url.strip().lower()
+    for pattern, protocol in _URL_PATTERNS:
+        if pattern in url:
```
manzt, line 400: This substring matching seems fragile. A URL like `s3.cloudflare.com` would match "s3" if the order changed, right? Maybe we consider matching against the hostname only, or using more specific patterns? Or document the expected behavior in tests.

Light2Dark: I've added to tests and a comment that the order matters. I agree, but I don't have much data to infer what the urls would look like 🤔, so opted for the simpler approach.

PR #8517, opened 2026-02-28
Add Cmd/Alt+Shift+L to toggle between Python and SQL cells

### LINKED ISSUE #8386: Feature Request: Allow Converting Existing Cells Between Python and SQL (Like Markdown)
### Description

**Hi team 👋**

_First of all, really enjoying using marimo_ — especially the SQL cell support with syntax highlighting and execution integration.

I’d like to propose a usability improvement around SQL cells.

## Current Behavior

Image

Image

* To create a SQL cell, I must manually choose it from the UI using the mouse.
* When creating a SQL cell, it auto-inserts `SELECT * FROM` by default.
* Once a cell is created as Python, there is no simple way to convert it to SQL.
* Likewise, SQL cells cannot be converted back to Python.

In contrast, Markdown cells can be converted easily.

---

## Problem

When working interactively:

* I often start typing in a Python cell and later realize it should be SQL.
* Or vice versa.
* Switching requires deleting and recreating the cell.
* This breaks flow and adds friction.

Also, if I'm working heavily with SQL, it would be helpful to configure SQL as the default cell type (similar to how some editors allow choosing a default language mode).

---

### 1️⃣ Cell Type Conversion

Add the ability to:

* Convert an existing Python cell → SQL cell
* Convert an existing SQL cell → Python cell

Similar to how Markdown conversion works.

This could be:

* a keyboard shortcut
* Or a right-click option

---

### 2️⃣ Configurable Default Cell Type

Add a setting to allow:

* Default cell type on Shift+Enter
* Example options:

  * Python (default)
  * SQL
  * Last used type

This would greatly improve workflow for SQL-heavy notebooks.

---

## Optional Improvement

When converting:

* Avoid auto-inserting `SELECT * FROM`, i think every time it is not needed
---

## Why This Matters

* Improves workflow speed
* Reduces mouse dependency
* Makes SQL-heavy usage smoother
* Aligns with how modern notebook tools handle language modes

---

## Example Use Case

When building analytical notebooks:

* Rapidly alternating between:

  * SQL data extraction
  * Python transformation/visualization
* Being able to toggle cell types instantly would significantly reduce friction.

---

## Summary

Adding cell type conversion and configurable default cell type would:

* Improve UX
* Reduce repetitive actions
* Support advanced SQL-heavy users

Happy to provide more details or help test if needed 🙂

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

Light2Dark:
## Summary
- Adds `Cmd/Alt+Shift+L` keyboard shortcut to toggle cells between Python and SQL (mirroring `Cmd/Ctrl+Shift+M` for Markdown). 
- Alt for Windows, Cmd for Mac (because Mod-shift-L conflicts with another key on windows)
- Merges the separate "Convert to SQL" and "Toggle as Python" menu actions into a single context-aware item that shows "View as SQL" or "View as Python" depending on the current cell language

Closes #8386

### DIFF
--- a/frontend/src/core/hotkeys/hotkeys.ts
+++ b/frontend/src/core/hotkeys/hotkeys.ts
@@ -141,7 +141,10 @@ const DEFAULT_HOT_KEY = {
   "cell.viewAsSQL": {
     name: "Toggle SQL",
     group: "Editing",
-    key: "Mod-Shift-l",
+    key: {
+      windows: "Alt-Shift-l",
+      main: "Mod-Shift-l",
+    },
   },
   "cell.complete": {
     name: "Code completion",


frontend/src/components/editor/actions/useCellActionButton.tsx
```diff
@@ -235,35 +238,41 @@ export function useCellActionButtons({ cell, closePopover }: Props) {
...
-      },
-      {
-        icon: <PythonIcon />,
-        label: "Toggle as Python",
-        handle: () => {
-          const editorView = getEditorView();
-          if (!editorView) {
-            return;
-          }
-          maybeAddMarimoImport({ autoInstantiate, createNewCell: createCell });
```
Light2Dark, line 262: removed this, since it's just a Python cell, we don't need a marimo import

Light2Dark:
> I think `Ctrl+Shift+L` conflicts with ["Fold all cells"](https://github.com/marimo-team/marimo/blob/main/frontend/src/core/hotkeys/hotkeys.ts#L298-L304) on Windows/Linux. No conflict on macOS since fold uses `Ctrl+Cmd+L` there.

That's unfortunate, and thank you for highlighting. I've opted for Alt-shift-L for Windows after looking at the vscode keybindings. However, I don't have a great opinion on this. https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf

Issue #5831, opened 2025-07-30
Error downloading slides as PDF


juanmpga:
### Describe the bug

When I present as "Slides" (a feature that I absolutely love!) and then download the pdf there seems to be some error in the arrangement of the slides and they do not appear correctly in the document.

* Here is what I get: [mre.pdf](https://github.com/user-attachments/files/21511670/mre.pdf)
* And here is what I would expect: [google-slides.pdf](https://github.com/user-attachments/files/21511679/google-slides.pdf)

I'm considering migrating from google slides to marimo because I feel the workflow is much more convenient (in terms of handling code, images, formulas), but this is something that's making me hesitant (the slides are hard to share). 

I put the label bug because it feels like an error, but I'm also aware that this functionality is on the periphery of what marimo notebooks really are, so perhaps it's not expected to work perfectly and you consider it a feature request (feel free to relabel). In that line I don't expect this to be fixed soon and I'm already grateful that there's the slides functionality to begin with.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.13",
  "editable": false,
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 154 Stepping 4, GenuineIntel",
  "Python Version": "3.13.2",
  "Binaries": {
    "Browser": "138.0.7204.170",
    "Node": "v20.14.0"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.0.1",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "typing-extensions": "missing",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.5.3"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "marimo",
# ]
# ///

import marimo

__generated_with = "0.14.13"
app = marimo.App(width="medium", layout_file="layouts/test.slides.json")


@app.cell
def _():
    import marimo as mo
[...]

PR #8523, opened 2026-03-01
feat: export slides as PDF

peter-gy:
Extends @mscolnick's POC #8435 based on the discussion under #5831 to better support exporting notebooks as slides both via CLI and the UI.

## 🔍 Description of Changes

- Introduces PDF export presets via `marimo export pdf notebook.py -o notebook.pdf --as=document|slides` API
- Leverages the output rasterization logic from #8458 so that slides can render interactive components

## UX Preview

Users can choose the PDF export preset via the UI with a `Recommended` badge suggesting the suitable preset inferred from the notebook's layout.

**View as Slides configured --> Slides preset recommended**

Screenshot 2026-02-28 at 10 06 30

**View as Vertical configured --> Document preset recommended**

Screenshot 2026-02-28 at 10 06 10

**Live Flow**

https://github.com/user-attachments/assets/2e10485a-8f50-43ee-8c86-1209696ffbb7

**Slide with exported interactive widget**

With #8458 the slides can also include interactive elements like [Quak](https://github.com/manzt/quak) dataframe viewer.

Screenshot 2026-03-01 at 18 11 30

### DIFF
--- a/marimo/_server/export/_pdf_raster.py
+++ b/marimo/_server/export/_pdf_raster.py
@@ -587,7 +587,7 @@ async def _capture_pngs_from_page(
     )
 
     async with async_playwright() as playwright:
-        browser = await playwright.chromium.launch(headless=False)
+        browser = await playwright.chromium.launch()
         context = await browser.new_context(
             viewport={
                 "width": _VIEWPORT_WIDTH,


marimo/_server/export/_pdf_raster.py
```diff
@@ -613,6 +655,7 @@ async def _capture_pngs_from_page(
                 continue
 
             try:
+                await page.evaluate(GO_TO_NEXT_SLIDE)
```
mscolnick, line 658: should this be conditional? i see in the `GO_TO_NEXT_SLIDE` there is a condition, but idk if that ends up changing

frontend/src/components/editor/actions/useNotebookActions.tsx
```diff
@@ -141,13 +134,47 @@ export function useNotebookActions() {
...
     isServerSidePdfExportEnabled || viewState.mode === "present";
+  const isSlidesLayout = selectedLayout === "slides";
 
   const renderCheckboxElement = (checked: boolean) => (
     <div className="w-8 flex justify-end">
       {checked && <CheckIcon size={14} />}
     </div>
   );
 
+  const renderRecommendedElement = (recommended: boolean) => {
```
mscolnick, line 145: this is nice, but not sure if its overkill. maybe someone with better design input can weight in. 

cc @Light2Dark @manzt this shows a little badge

Image

mscolnick: i guess i can see the useful when the user did not design their notebook for slides, otherwise they probably will get a mediocre output

Light2Dark: Maybe just remove it? And then in slides mode, slides layout could be first on the dropdown

marimo/_server/export/_pdf_raster.py
peter-gy: I deferred the conditional to the JS side because the capturing flow is highly dependent on the fact that Swiper is used on the frontend to render slides. So if that ends up changing on the frontend (e.g. because we migrate off of Swiper), this module needs to be revised anyways. So to me, a conditional check here on the Python side for `--as=slide` preset would just create the illusion of robustness at the expense of LOC++.

frontend/src/components/editor/actions/useNotebookActions.tsx
peter-gy: My goal here was to indicate to the user that we inferred the optimal export preset based on their layout. But yes, maybe this is way too visually prominent.

> slides layout could be first on the dropdown

I would be also fine with indicating this through the order of options, but I can imagine that some users would find it too implicit, maybe even report it as a bug due to perceived UI non-determinism.

mscolnick: i personally don't like changing the order. that does feel weird and would could be classified as a bug

mscolnick: lets keep this for now, and then can modify later if needed

PR #8525, opened 2026-03-02
storage-inspector: When parent dir is searched, child files should be open too

Light2Dark:
Previously a search would only match exact files, so if a directory matched, children files would not show.

This fixes that
image

## 📋 Checklist

### DIFF
--- a/frontend/src/components/storage/storage-inspector.tsx
+++ b/frontend/src/components/storage/storage-inspector.tsx
@@ -140,7 +140,6 @@ const StorageEntryChildren: React.FC<{
   depth: number;
   locale: string;
   searchValue: string;
-  parentMatched: boolean;
   onOpenFile: (info: OpenFileInfo) => void;
 }> = ({
   namespace,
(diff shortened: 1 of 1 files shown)


frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -240,6 +243,12 @@ const StorageEntryRow: React.FC<{
   const name = displayName(entry.path);
   const hasSearch = !!searchValue.trim();
 
+  const selfMatches =
```
manzt, line 243: I wonder if instead of threading `parentMatched` through both components, we could just clear `searchValue` for children of a matched directory:

```tsx
const selfMatches =
  isDir && hasSearch && name.toLowerCase().includes(searchValue.trim().toLowerCase());

<StorageEntryChildren
  ...
  searchValue={selfMatches ? "" : searchValue}
/>
```

Shouldn't `filterEntries` already return everything when search is empty? Only difference is subdirectories under a matched parent won't auto-expand.

Light2Dark: that's a much simpler fix, thank you!

PR #8528, opened 2026-03-02
add storage/database connections UI

Light2Dark:
## 📝 Summary

Implements a UI to create obstore and fsspec connections. This unifies with the 'Add database form'. Supported providers at the moment are
- s3 (tested)
- gcs (semi-tested)
- azure
- google drive (tested)

We should probably standardize icons in the future

image

image

image

image

image

image

## 📋 Checklist

### DIFF
--- a/frontend/src/components/editor/connections/components.tsx
+++ b/frontend/src/components/editor/connections/components.tsx
@@ -2,7 +2,7 @@
 
 import { zodResolver } from "@hookform/resolvers/zod";
 import React from "react";
-import { type FieldValues, useForm } from "react-hook-form";
+import { type DefaultValues, type FieldValues, useForm } from "react-hook-form";
 import type { z } from "zod";
 import { type FormRenderer, ZodForm } from "@/components/forms/form";
 import { getDefaults } from "@/components/forms/form-utils";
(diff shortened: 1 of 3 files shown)


frontend/src/components/editor/connections/database/add-database-form.tsx
```diff
@@ -0,0 +1,302 @@
...
+    libraries: ConnectionLibrary[];
+    preferred: ConnectionLibrary;
+  };
+}
+
+const DATABASES = [
+  {
+    name: "PostgreSQL",
+    schema: PostgresConnectionSchema,
+    color: "#336791",
```
manzt, line 49: Outside the scope of this PR but I find all the colors for the database and store tiles to be a bit noisy/distracting. It's also sometimes hard to find the right color (e.g., Google Cloud is white on green?).

I wonder if we could just do the logos with a shared palette.

frontend/src/components/editor/connections/components.tsx
```diff
@@ -0,0 +1,177 @@
...
+  displayNames: Record<string, string>;
+  libraryLabel?: string;
+  generateCode: (values: T, library: string) => string;
+  onSubmit: () => void;
+  onBack: () => void;
+}) => {
+  const form = useForm<T>({
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unnecessary-type-assertion
+    defaultValues: getDefaults(schema) as any,
+    resolver: zodResolver(schema as unknown as z.ZodType<unknown, T>),
```
manzt: Maybe just a comment here why this is ok? It seems like typescript is just failing to narrow the type. The key seems to be that we want to keep a mapping of `T` to `z.ZodType<T>`.

manzt (commented):
I like this addition and definitely will help with both discovery of storage UI as well as ergonomics of using obstore/fsspec.

frontend/src/components/editor/connections/database/add-database-form.tsx
Light2Dark: I agree, one issue is we have both coloured logos and black and white ones. I think we should standardize.
Some icons only have coloured ones (we should probably just grayscale it)

PR #8533, opened 2026-03-02
feat: add `marimo export session` CLI tool

peter-gy:
This PR enables executing notebooks and storing their session view via `marimo export session ...` CLI. It is a utility similar to `marimo export thumbnail`, introduced in #8097, #8154.

According to a brief async discussion with @mscolnick, as a follow up we'll want to ensure that notebooks are not executed over and over again across different CLI commands (e.g. `marimo export thumbnail ...`, `marimo export pdf ...` also generate session JSON via `run_app_until_completion`). In preparation for that, I added the `"--overwrite/--no-overwrite"` option to this utility, so that if a `<notebook>.json` session file already exists for `<notebook>`, we can opt to skip re-execution.

### DIFF
--- a/tests/_cli/test_cli_export.py
+++ b/tests/_cli/test_cli_export.py
@@ -1177,6 +1177,7 @@ class TestExportSession:
         ]
 
     def test_export_session_directory_default_skips_up_to_date(
+        self,
         tmp_path: Path,
         temp_marimo_file: str,
     ) -> None:


marimo/_cli/export/session.py
```diff
@@ -0,0 +1,316 @@
...
+    type=bool,
+    help=_sandbox_message,
+)
+@click.option(
+    "--overwrite/--no-overwrite",
+    default=True,
+    help="Overwrite existing session snapshots.",
+)
+@click.option(
+    "--continue-on-error/--no-continue-on-error",
```
peter-gy, line 368: I was torn about the naming here. Something more intent-driven like `"--continue-on-error/--fail-on-error"` would be a decent alternative, but then we'd break the `"--<action>/--no-<action>"` consistency we have in the CLI.

I'd love to hear your thoughts on this.

marimo/_cli/export/session.py
```diff
@@ -0,0 +1,316 @@
...
+if TYPE_CHECKING:
+    from marimo._cli.sandbox import SandboxMode
+
+_sandbox_message = (
+    "Run the command in an isolated virtual environment using "
+    "`uv run --isolated`. Requires `uv`."
+)
+
+
+def _is_multi_target(paths: list[Path]) -> bool:
```
mscolnick: nit: this is duplicated now

marimo/_cli/export/session.py
```diff
@@ -0,0 +1,316 @@
...
+        "args": list(notebook_args),
+    }
+    return await asyncio.to_thread(
+        _export_session_snapshot_in_subprocess,
+        venv_python,
+        payload,
+    )
+
+
+def _export_session_snapshot_in_subprocess(
```
mscolnick, line 101: why do this in a subprocess? clean global state? performance?

marimo/_cli/export/session.py
```diff
@@ -0,0 +1,316 @@
...
+                    continue
+
+                echo(f"Running {notebook.short_name}...")
+                venv_python = (
+                    sandbox_pool.get_python(str(notebook.path))
+                    if sandbox_pool is not None
+                    else None
+                )
+
+                session_snapshot, did_error = await _export_session_snapshot(
```
mscolnick: could do this in parallel (with some max chunk) and running in a [task group](https://docs.python.org/3/library/asyncio-task.html#task-groups)

marimo/_cli/export/session.py
peter-gy: good catch, thanks

marimo/_cli/export/session.py
peter-gy: it is for global state isolation and for correct per-notebook sandboxes

marimo/_cli/export/session.py
peter-gy: yes, I was just not sure about the `--(no)-continue-on-error` semantics in a parallel setting, so went for naive sequential for v1.

I'd implement a best-effort fail-fast under concurrency, so I'd cancel pending tasks and let already running tasks finish.

peter-gy: @mscolnick for max chunk would you expose a CLI option with a sensible default value like `4`, or let's just hard code it as a constant for now?

mscolnick: preference for sensible defaults, and only expose configuration until asked for. this reduces the configuration cognitive load on the user and allows us to avoid more API surface area

mscolnick: > I'd implement a best-effort fail-fast under concurrency, so I'd cancel pending tasks and let already running tasks finish.

I agree with this approach

marimo/_cli/export/session.py
akshayka: Let's keep `--no-continue-on-error`

docs/guides/exporting.md
```diff
@@ -122,6 +122,57 @@ their dependency graph.
...
+marimo export session folder/
+```
+
+Pass CLI args through to notebooks with `--`:
+
+```bash
+marimo export session notebook.py -- --foo 123
+```
+
+By default, marimo overwrites existing session snapshots. To keep existing
```
akshayka: Have you considered the following alternative, in which the CLI would by default only overwrite "stale" sessions:

"By default, marimo will only overwrite a session file if the notebook's code has changed."

Then introduce a flag that allow marimo to force overwriting of all notebooks, regardless of staleness. This is the workflow that I would personally want for `marimo-team/gallery-examples`.

akshayka (commented):
One comment on the overwriting behavior

marimo/_cli/export/session.py
akshayka: I would actually not do parallelism for now. I'd keep it simple and have serial, and let users configure their own concurrency. We can always give users a four line recipe for running in parallel on a directory.

peter-gy: just finished implementing `asyncio.TaskGroup`-based parallelism, just to realize that it is not compatible with Python 3.10, so we'd also need to implement a semaphore + `asyncio.wait(...)` approach for compat.

so staying serial would keep things simpler for sure.

docs/guides/exporting.md
peter-gy: this would make a lot of sense

marimo/_cli/export/session.py
akshayka: Let's keep it serial for now

docs/guides/exporting.md
akshayka: Great, let's do that. I have some code in the gallery-examples repo that implements a staleness check: 

https://github.com/marimo-team/gallery-examples/blob/3f588ca4f77914b3137429819ba672ff3f670e9e/scripts/validate-sessions.py#L102

peter-gy:
> Great, let's do that. I have some code in the gallery-examples repo that implements a staleness check:
> 
> https://github.com/marimo-team/gallery-examples/blob/3f588ca4f77914b3137429819ba672ff3f670e9e/scripts/validate-sessions.py#L102

Thanks for the pointer @akshayka. I took the same core idea of comparing `code_hash` in order-insensitive manner. In a follow up PR we might want to take into account the PEP 723 inline script metadata as well (e.g. sandboxed dependency versions changed, marimo version changed, python min version changed, etc.). For this we'd need to add something like `metadata.script_metadata_hash` to snapshot JSON files and I see that outside of this PR's scope.

akshayka:
> > Great, let's do that. I have some code in the gallery-examples repo that implements a staleness check:
> > https://github.com/marimo-team/gallery-examples/blob/3f588ca4f77914b3137429819ba672ff3f670e9e/scripts/validate-sessions.py#L102
> 
> Thanks for the pointer @akshayka. I took the same core idea of comparing `code_hash` in order-insensitive manner. In a follow up PR we might want to take into account the PEP 723 inline script metadata as well (e.g. sandboxed dependency versions changed, marimo version changed, python min version changed, etc.). For this we'd need to add something like `metadata.script_metadata_hash` to snapshot JSON files and I see that outside of this PR's scope.

@peter-gy, the changes look great, thank you!

> In a follow up PR we might want to take into account the PEP 723 inline script metadata as well (e.g. sandboxed dependency versions changed, marimo version changed, python min version changed, etc.). For this we'd need to add something like `metadata.script_metadata_hash` to snapshot JSON files and I see that outside of this PR's scope.

I think this is worth doing, and I agree that it was outside scope of this PR

PR #8538, opened 2026-03-02
Fix Panel DynamicMap and interactive streams not updating

### LINKED ISSUE #8527: HoloViews/Panel: DynamicMap and interactive streams (Selection, Range, etc.) do not update
### Describe the bug

In marimo, interactive streams in `HoloViews` (using the `Bokeh` backend) and `Panel` objects do not trigger live updates. While the initial plots render correctly, any interaction that should trigger a Python-side callback (such as selecting points, changing ranges, or using linked streams) fails to update the output.

This behavior is not limited to `Selection1D`; it appears to be a general issue with how `marimo` handles bi-directional communication for libraries that rely on Bokeh’s document synchronization. The same lack of reactivity is observed when wrapping these components in `panel`.

### To Reproduce
1. Create a new marimo notebook.
2. Use the attached following code (which uses `Selection1D` as a representative example)
3. Run the notebook
4. Use the "Box Select" tool to select points on the left plot
5. **Observation:** The right plot (`DynamicMap`) remains static and does not reflect the selection

### **Expected Behavior**
Interactive streams should synchronize with the marimo kernel. When a user interacts with a plot (e.g., selecting points or panning a range), the associated `DynamicMap` or `Panel` callback should execute and update the UI live, matching the behavior seen in Jupyter or a standalone Bokeh/Panel server.

### **Actual Behavior**
The UI renders the initial state but becomes "disconnected" from the Python kernel for any subsequent interactions. No updates are triggered in the `DynamicMap`.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.2",
  "editable": false,
  "location": "C:/Users/user/AppData/Local/uv/cache/archive-v0/sRMV6_hPe_WUfr1f_19XM/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "10",
  "Python Version": "3.12.10",
  "Locale": "--",
  "Binaries": {
    "Browser": "145.0.7632.117",
    "Node": "--",
    "uv": "0.7.3 (3c413f74b 2025-05-07)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.17.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.41.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce
```python
# /// script
# dependencies = [
#     "holoviews==1.22.1",
#     "marimo",
#     "numpy==2.4.2",
# ]
# requires-python = ">=3.12"
# ///

import marimo

__generated_with = "0.20.2"


@app.cell
def _():
    # Taken directly from https://holoviews.org/reference/streams/bokeh/Selection1D_points.html#selection1d-points

    import numpy as np

    import holoviews as hv
    from holoviews import opts, streams

    hv.extension('bokeh')

    opts.defaults(opts.Points(tools=['box_select', 'lasso_select']))

    # Declare some points
    points = hv.Points(np.random.randn(1000,2 ))

    # Declare points as source of selection stream
    selection = streams.Selection1D(source=points)

    # Write function that uses the selection indices to slice points and compute stats
    def selected_info(index):
        selected = points.iloc[index]
        if index:
            label = 'Mean x, y: {:.3f}, {:.3f}'.format(*tuple(selected.array().mean(axis=0)))
        else:
            label = 'No selection'
        return selected.relabel(label).opts(color='red')

    # Combine points and DynamicMap
    points + hv.DynamicMap(selected_info, streams=[selection])
    return


if __name__ == "__main__":
    app.run()
```

manzt:
Fixes #8527

The Panel comm layer was broadcasting the wrong notification type and `comm._comm` was falsy, causing Panel's push() to silently drop backend-initiated updates. On the frontend, stale document refs broke DynamicMap after HoloViews replaced the document post-embedding. Echo prevention now uses Bokeh's built-in `event.sync` flag instead of a mutable ref guard.

### DIFF
--- a/frontend/src/plugins/impl/panel/PanelPlugin.tsx
+++ b/frontend/src/plugins/impl/panel/PanelPlugin.tsx
@@ -279,7 +279,7 @@ const PanelSlot = (props: Props) => {
         // Bokeh tags server-applied patch events with sync=false via
         // model.setv({...}, {sync: false}) inside apply_json_patch.
         // Only forward user-initiated events (sync=true, the default).
-        if (isSyncEvent(event) && event.sync !== false) {
+        if (isSyncEvent(event) && event.sync) {
           eventBufferRef.current?.add(event);
         }
       });


frontend/src/plugins/impl/panel/PanelPlugin.tsx
```diff
@@ -193,6 +235,11 @@ const PanelSlot = (props: Props) => {
         return;
       }
 
+      const doc = getDoc();
```
mscolnick: i last recall that calling a `useEvent` fn on the first render with throw an error. did you not hit that?

manzt: Yeah, `react-use-event-hook`'s `useEvent` throws if called before mount, which happens when Bokeh fires `on_change` synchronously during embedding (so i had to wrap in try/catch before). 

I switched to `@dnd-kit/utilities` version which no-ops before mount, making the protective try/catch unnecessary.

PR #8540, opened 2026-03-03
fix: handle matplotlib boxplot/violinplot dict outputs

### LINKED ISSUE #6838: Duplicated boxplots
### Describe the bug

The same boxplot is output 7 times. Other `matplotlib` plot types seem to work correctly.

Image

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.16.5",
  "editable": false,
  "location": ".../python3.13/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "24.6.0",
  "Processor": "arm",
  "Python Version": "3.13.7",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "141.0.7390.108",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.7.0",
    "packaging": "25.0",
    "psutil": "7.1.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.48.0",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.8.1",
    "python-lsp-ruff": "2.3.0",
    "ruff": "0.14.0"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```
import marimo as mo
import numpy as np

fig, ax = plt.subplots()
ax.boxplot([0])
```

ffmiruz:
ax.boxplot() and ax.violinplot() return dicts containing matplotlib Artist objects (as lists or direct values). The existing formatter only handled list/tuple of artists, causing the dict values to be formatted individually, resulting in duplicated plot outputs.

This extends the matplotlib special case in the structures formatter to also extract artists from dict values.

Fixes #6838

### DIFF
--- a/marimo/_output/formatters/structures.py
+++ b/marimo/_output/formatters/structures.py
@@ -4,8 +4,7 @@ from __future__ import annotations
 import json
 import sys
 from collections import defaultdict
-from collections.abc import Sequence
-from typing import Any, Union
+from typing import TYPE_CHECKING, Any, Union
 
 from marimo._messaging.mimetypes import KnownMimeType
 from marimo._output import formatting
@@ -15,6 +14,9 @@ from marimo._output.formatters.repr_formatters import maybe_get_repr_formatter
 from marimo._plugins.stateless.inspect import inspect
 from marimo._plugins.stateless.plain_text import plain_text
 from marimo._utils.flatten import CyclicStructureError, flatten
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
 
 
 def is_structures_formatter(


marimo/_output/formatters/structures.py
```diff
@@ -122,10 +122,25 @@ def _format_structure(
...
+                # Collect all artists from list/tuple/dict structures
+                artists: list[Any] = []
+                if isinstance(t, dict):
+                    for v in t.values():
+                        if isinstance(v, (list, tuple)):
+                            artists.extend(v)
+                        elif isinstance(v, matplotlib.artist.Artist):
+                            artists.append(v)
+                else:
+                    artists = list(t)
```
akshayka: Can this just be `artists = t` to avoid the copy?

marimo/_output/formatters/structures.py
```diff
@@ -122,10 +122,25 @@ def _format_structure(
...
+
+                if artists and all(
+                    isinstance(i, matplotlib.artist.Artist) for i in artists
+                ):
```
akshayka: Can you rewrite this so that it short-circuits on the first non-Artist encountered? 

For example, this logic does short-circuit: `if all(isinstance(i, matplotlib.artist.Artist) for i in t):`

marimo/_output/formatters/structures.py
akshayka: What happens if a structure has some artists, some not artists? It seems like this change silently drops the non-artists. Please add a test for this code path.

marimo/_output/formatters/structures.py
akshayka: This will lead to an unnecessary copy (iteration and memory allocation)

akshayka (changes requested):
Thanks for the PR; I have requested some changes.

marimo/_output/formatters/structures.py
ffmiruz: Got it. My intention was, together with casting tuple and dict into list so we have one clear branch where we do the check.
Also just in case, if you mean doing it literally like this `if all(isinstance(i, matplotlib.artist.Artist) for i in t):`, that won't work because it iterates on tuple key.  I failed to mention this in the PR.

marimo/_output/formatters/structures.py
ffmiruz: Mixed structure will fallthrough due failing line 141 check. This is the original behaviour before fix.
`hist()` follow this path. 
I will add more test.

marimo/_output/formatters/structures.py
akshayka: Thanks, I didn't mean doing it literally as `if all(isinstance(i, matplotlib.artist.Artist) for i in t)`, but rather that was just an example of an expression that would short-circuit

marimo/_output/formatters/structures.py
akshayka: Here is an example of a mixed structure failing

Image

tests/_output/formatters/test_structures.py
```diff
@@ -57,6 +57,221 @@ async def test_matplotlib_special_case(
...
+        await executing_kernel.run(
+            [
+                exec_req.get(
+                    """
+                    import matplotlib.pyplot as plt
+                    from marimo._output.formatting import get_formatter
+
+                    fig, ax = plt.subplots()
+                    boxplot_result = ax.boxplot([0])
+                    formatter = get_formatter(boxplot_result)
```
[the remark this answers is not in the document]
akshayka: The runtime closes figures so this is fine

akshayka (dismissed):
Thanks so much for the contribution! I pushed a small change for legibility, but your logic was correct

mscolnick (approved):

PR #8547, opened 2026-03-03
feat: consider PEP 723 metadata in session staleness logic

peter-gy:
As a follow up to #8533 this PR consolidates session snapshot handling and staleness checks so they can be reused across export paths and used to avoid unnecessary notebook re-execution.

It moves snapshot staleness/persistence logic into a shared module, adds `metadata.script_metadata_hash` to session snapshots, and includes PEP 723 metadata hashing in freshness decisions. It also centralizes automatic `__marimo__/session` persistence in `run_app_until_completion(...)`, while keeping `marimo export session` on a single-write path (`persist_session=False` there). Cache-hit logic is updated to include script metadata hash.

### DIFF
--- a/docs/guides/exporting.md
+++ b/docs/guides/exporting.md
@@ -157,7 +157,7 @@ Pass CLI args through to notebooks with `--`:
 marimo export session notebook.py -- --foo 123
 ```
 
-By default, marimo only rewrites session snapshots when they are stale (for example, when the notebook code has changed).
+By default, marimo only rewrites session snapshots when they are stale (for example, when notebook code changes or PEP 723 script metadata changes).
 Up-to-date snapshots are skipped.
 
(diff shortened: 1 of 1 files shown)


marimo/_utils/inline_script_metadata.py
```diff
@@ -247,3 +249,16 @@ def has_marimo_in_script_metadata(filepath: str) -> bool | None:
 
     dependencies = project.get("dependencies", [])
     return any(is_marimo_dependency(dep) for dep in dependencies)
+
+
+def script_metadata_hash_from_filename(name: str) -> str | None:
+    project = _get_pyproject_from_filename(name)
```
dmadisetti, line 255: Do we want to invalidate on all pyproject changes? Or just deps? Probably easier for all, just wondering if there

PR #8558, opened 2026-03-03
fix: dont rasterize images in the /pdf endpoint

mscolnick:
This PR reverts to using the active session to `updateCellOutputsWithScreenshots` so that we don't need to rasterize on the backend. This could show a different state and requires another run. Rasterization is still used in the CLI.

Also:
- removes feature flag for server-side-pdf-export since this caused bad conditionals
- enabled PDF export in wasm just to use browser print

**PDF Export Simplification:**
- Removed the import and usage of `PDFRasterizationOptions` and `collect_pdf_png_fallbacks` from the `export_as_pdf` endpoint, streamlining the PDF export process by eliminating rasterization logic.
- **Why** - we can screenshot the current view and populate outputs with PNGs

**Live Notebook Server Command Updates:**
- Added the `-y` flag to suppress prompts and the `--no-sandbox` flag to avoid unnecessary sandboxing when launching the notebook server in `_build_command`, making the process more suitable for automated or internal usage. 
- **Why** - this can cause the live server to hang when it has inline metadata

### DIFF
--- a/frontend/src/components/editor/actions/__tests__/pdf-export.test.ts
+++ b/frontend/src/components/editor/actions/__tests__/pdf-export.test.ts
@@ -15,10 +15,10 @@ describe("runServerSidePDFDownload", () => {
 
     expect(downloadPDF).toHaveBeenCalledWith({
       filename: "slides.py",
-      webpdf: true,
+      webpdf: false,
       preset: "document",
-      includeInputs: false,
-      rasterServer: "live",
+      includeInputs: true,
+      rasterServer: "static",
     });
   });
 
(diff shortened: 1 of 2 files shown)


dmadisetti (approved):

dmadisetti (dismissed):

frontend/src/components/editor/actions/pdf-export.ts
```diff
@@ -18,9 +18,9 @@ export async function runServerSidePDFDownload(opts: {
 
   await downloadPDF({
     filename,
-    webpdf: true,
+    webpdf: false,
     preset,
-    includeInputs: false,
-    rasterServer: "live",
+    includeInputs: true,
+    rasterServer: "static",
```
akshayka, line 24: I am reverting to what we had in 0.20.2 to produce Jupyter-style PDFs FYI

PR #8559, opened 2026-03-03
feat: add expanded option to accordion

### LINKED ISSUE #8549: Accordion layout elements visible by default
### Description

The [accordion](https://docs.marimo.io/api/layouts/accordion/) layout in Marimo is initialised with its elements collapsed by default. The user has to click on the individual elements to display them at runtime.

### Suggested solution

If it's within scope, it would be nice to have an accordion argument (defaulting to `False`) that initialises the layout and displays the expanded elements at startup.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

ryptotalent:
Adds `expanded` parameter to `mo.accordion()` that expands all items by default when set to `True`. Defaults to `False` for backward compatibility.

Closes #8549

### DIFF
--- a/frontend/src/components/editor/navigation/navigation.ts
+++ b/frontend/src/components/editor/navigation/navigation.ts
@@ -661,6 +661,7 @@ export function useCellEditorNavigationProps(
   const temporarilyShownCodeActions = useTemporarilyShownCodeActions();
   const keymapPreset = useAtomValue(keymapPresetAtom);
   const hotkeys = useAtomValue(hotkeysAtom);
+  const userConfig = useAtomValue(userConfigAtom);
 
   const vimCommandModeShortcut = useMemo(() => {
     const shortcut = hotkeys.getHotkey("command.vimEnterCommandMode");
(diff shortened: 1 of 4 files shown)


frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -720,7 +721,11 @@ export function useCellEditorNavigationProps(
         }
       } else {
         // For non-vim mode, regular Escape exits to command mode
-        if (evt.key === "Escape") {
+        // Only if the configuration option is enabled
+        if (
+          evt.key === "Escape" &&
+          userConfig.keymap.enter_command_mode_on_escape !== false
+        ) {
```
mscolnick, line 728: from another PR?

marimo/_plugins/stateless/accordion.py
```diff
@@ -11,7 +11,10 @@
 
 @mddoc
 def accordion(
-    items: dict[str, object], multiple: bool = False, lazy: bool = False
+    items: dict[str, object],
+    multiple: bool = False,
+    lazy: bool = False,
+    expanded: bool = False,
```
mscolnick, line 17: does this expand all? is there a case you may just want to expand first first one? or middle one?

cc @akshayka @dmadisetti regarding API

marimo/_plugins/stateless/accordion.py
```diff
@@ -22,6 +25,7 @@ def accordion(
         lazy: a boolean, whether to lazily load the accordion content.
             This is a convenience that wraps each accordion in a `mo.lazy`
             component.
+        expanded: if True, all accordion items are expanded by default
```
dmadisetti, line 28: It's not clear how `expanded` and `multiple` interact. It seems like `mutliple=False` isn't respected with this?

PR #8563, opened 2026-03-04
only display "Jump to page" when >100 pages

Light2Dark:
## 📝 Summary

100 is chosen because that's the clamping point where we introduce ellipsis.

image

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/pagination.tsx
+++ b/frontend/src/components/data-table/pagination.tsx
@@ -37,6 +37,8 @@ import {
 } from "../ui/select";
 import type { DataTableSelection, PageRange } from "./types";
 
+const MAX_PAGES_BEFORE_CLAMPING = 100;
+
 interface DataTablePaginationProps<TData> {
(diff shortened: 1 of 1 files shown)


frontend/src/components/data-table/pagination.tsx
```diff
@@ -325,35 +325,39 @@ export const PageSelector = ({
...
-              if (e.key === "Enter") {
-                handleJump();
-              }
-              e.stopPropagation();
-            }}
-            className="h-6 text-xs"
-            data-testid="page-jump-input"
-          />
-        </div>
+        {totalPages > 50 && (
```
mscolnick: why 50? should it be based on when we do the clamping (which might be 50?)

Light2Dark: We clamp at 100, but I felt 50 was a good number so you don't have to scroll too much. We could change later I think

I've updated to 100

PR #8567, opened 2026-03-04
Add section on opencode agent

koaning:
I noticed that we did not have a guide on how to set up opencode. Adding it in this PR.

### DIFF
--- a/docs/guides/editor_features/agents.md
+++ b/docs/guides/editor_features/agents.md
@@ -94,13 +94,13 @@ codex
 
 [OpenCode](https://opencode.ai/) is an open source AI coding agent built for the terminal, but also supports ACP.
 
-**Installation and login:**
+**Installation:**
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/guides/editor_features/agents.md
```diff
@@ -90,6 +90,61 @@ codex
...
+
+    ```bash
+    npx stdio-to-ws "npx opencode-ai acp" --port 3023
+    ```
+
+=== "Windows"
+
+    ```bash
+    npx stdio-to-ws "cmd /c npx opencode-ai acp" --port 3023
+    ```
```
akshayka, line 119: Are these still necessary? Our frontend doesn't automatically run this in the terminal for you?

koaning: CleanShot 2026-03-04 at 17 33 07@2x

I think so?

PR #8574, opened 2026-03-04
Add documentation page for real-time collaboration

manzt:
marimo has a working experimental RTC implementation but I could not find any documentation for it anywhere (and neither can our users). These changes add a standalone doc page covering how to enable it, what syncs, and current limitations. The page is intentionally not added to the mkdocs nav so it isn't discoverable in the sidebar, but it can be linked to directly when giving context to users asking about collaborative editing.


mscolnick:
RTC still does not work well (and its behind feature flag). i dont think we want to document this yet.

akshayka:
Closing until we get RTC working better

PR #8579, opened 2026-03-05
feat(image): add `normalize` parameter to `mo.image`

### LINKED ISSUE #8569: mo.image always normalize image intensity from min-max to 0-255, making inter-image comparison impossibile
## The problem
The current implementation of image normalization scaling forces a global normalization based on individual image's min and max pixel values here: [image.py#L71](https://github.com/marimo-team/marimo/blob/c0b58212cd4126fe78cac763bf84d32ed2d98011/marimo/_plugins/stateless/image.py#L71)
`src = (src - src.min()) / (src.max() - src.min()) * 255.0`

While this ensures high contrast for a single image, it makes it impossible to compare different images within a dataset. For example, a "dark" image and a "light" image will both be scaled to the same $0-255$ range, destroying the relative physical or optical meaning of the pixel intensities.

Also, if an image is compose of a single pixel value, this yields a division by zero warning.

As a workaround, now I'm adding a black and white border around images so that normalization is actually bypassed.

See this screenshot: 

Image

## Suggested Fix
I'm not sure if this was by design, but I would suggest to add the possibility to pass a `vmin` and `vmax` values for normalization, or a flag to disable it.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.4",
  "editable": false,
  "location": "/home/[redacted]/.cache/uv/archive-v0/qCRvGHMauZmGiIDOhHiyJ/lib/python3.14/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.17.0-14-generic",
  "Processor": "x86_64",
  "Python Version": "3.14.0",
  "Locale": "--",
  "Binaries": {
    "Browser": "145.0.7632.116",
    "Node": "v20.16.0",
    "uv": "0.9.8"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.17.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.41.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "marimo",
#     "numpy",
#     "pillow"
# ]
# ///
import marimo

__generated_with = "0.20.4"


@app.cell
def _():
    import marimo as mo
    import numpy as np

    dark_image = np.full((100, 100), 50, dtype=np.uint8)
    light_image = np.full((100, 100), 200, dtype=np.uint8)

    mo.hstack([
        mo.image(dark_image),
        mo.image(light_image)
    ], justify="space-around")
    return dark_image, light_image, mo, np


@app.cell
def _(dark_image, light_image, mo, np):
    def _add_borders(image):
        # 1. Add 1px black border (0)
        black_border = np.pad(image, 1, constant_values=0)

        # 2. Add 1px white border (255)
        return np.pad(black_border, 1, constant_values=255)

    mo.hstack([
        mo.image(_add_borders(dark_image)),
        mo.image(_add_borders(light_image))
    ], justify="space-around")
    return


if __name__ == "__main__":
    app.run()
```

weiguangli-io:
## Summary

Adds a `normalize` parameter to `mo.image()` to control whether array pixel intensities are rescaled from `[min, max]` to `[0, 255]`.

- **`normalize=True`** (default): current behavior preserved -- pixel intensities are rescaled to the full `[0, 255]` range
- **`normalize=False`**: pixel values are clipped to `[0, 255]` without rescaling, preserving relative intensity differences across images

This also fixes a division-by-zero warning when displaying constant images (all pixels have the same value).

## Motivation

As reported in #8569, the forced normalization makes it impossible to visually compare different images within a dataset -- a "dark" image and a "light" image both get scaled to the same `0-255` range, destroying the relative physical meaning of pixel intensities.

## Changes

- `marimo/_plugins/stateless/image.py`:
  - Added `normalize` parameter to `_normalize_image()` (internal helper)
  - Added `normalize` parameter to `image()` (public API), defaulting to `True` for backward compatibility
  - Added division-by-zero guard for constant images
  - When `normalize=False`, pixel values are clipped to `[0, 255]` instead of rescaled

- `tests/_plugins/stateless/test_image.py`:
  - Added `test_image_normalize_flag` -- verifies pixel values are preserved when `normalize=False`
  - Added `test_image_normalize_constant_image` -- verifies constant images don't trigger division-by-zero

## Test plan

- [x] All existing tests pass (13/13 in `test_image.py`, 15/15 in `test_image_compare.py`)
- [x] New tests pass (2 new tests)
- [x] `ruff check` passes with no errors

Closes #8569

### DIFF
--- a/marimo/_plugins/stateless/image.py
+++ b/marimo/_plugins/stateless/image.py
@@ -21,7 +21,7 @@ Tensor = Any
 ImageLike = Union[Image, Tensor]
 
 
-def _normalize_image(src: ImageLike) -> Image:
+def _normalize_image(src: ImageLike, normalize: bool = True) -> Image:
     """Normalize an image-like object to a standard format.
 
     This function handles a variety of input types, including lists, arrays,
@@ -39,6 +39,9 @@ def _normalize_image(src: ImageLike) -> Image:
     Args:
         src: An image-like object. This can be a list, array, tensor, or a
             file-like object.
+        normalize: If True, rescale pixel intensities from [min, max] to
+            [0, 255]. If False, pixel values are clipped to [0, 255] without
+            rescaling.
 
     Returns:
         A BytesIO object or other Image type.
@@ -68,7 +71,21 @@ def _normalize_image(src: ImageLike) -> Image:
             if hasattr(src, "toarray"):
                 src = src.toarray()
             src = numpy.array(src)
-        src = (src - src.min()) / (src.max() - src.min()) * 255.0
+        if normalize:
+            src_min = src.min()
+            src_max = src.max()
+            if src_min != src_max:
+                src = (src - src_min) / (src_max - src_min) * 255.0
+            else:
+                # Avoid division by zero for constant images
+                import numpy
+
+                src = numpy.full_like(src, fill_value=src_min, dtype=float)
+                src = numpy.clip(src, 0, 255)
+        else:
+            import numpy
+
+            src = numpy.clip(src, 0, 255).astype(float)
         img = _Image.fromarray(src.astype("uint8"))
         # io.BytesIO is one of the Image types.
         normalized_src: Image = io.BytesIO()
@@ -102,6 +119,7 @@ def image(
     rounded: bool = False,
     style: Optional[dict[str, Any]] = None,
     caption: Optional[str] = None,
+    normalize: bool = True,
 ) -> Html:
     """Render an image as HTML.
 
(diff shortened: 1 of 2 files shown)


weiguangli-io:

akshayka:
I have converted this to a draft because the API needs further discussion on the issue

PR #8592, opened 2026-03-05
Add server registry and SSE execute endpoint

manzt:
Running marimo instances now write a small JSON file to `~/.local/state/marimo/servers/` at startup so external tools (like shell scripts) can discover them without importing marimo or hitting an HTTP endpoint. Registration is opt-in: only servers started with `--no-token` and `--no-skew-protection` are registered, since those are the ones that have explicitly relaxed local access. The registry entry is kept minimal — just `host`, `port`, `base_url`, `pid`, `version` — with no secrets or session data.

### DIFF
--- a/marimo/_mcp/code_server/main.py
+++ b/marimo/_mcp/code_server/main.py
@@ -8,7 +8,6 @@ in a running marimo kernel via the scratchpad.
 
 from __future__ import annotations
 
-import asyncio
 import os
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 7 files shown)


marimo/_server/session_registry.py
```diff
@@ -0,0 +1,191 @@
...
+                    path.unlink(missing_ok=True)
+                    continue
+
+                entries.append(entry)
+            except (json.JSONDecodeError, TypeError, KeyError) as e:
+                LOGGER.warning(
+                    "Skipping invalid registry entry %s: %s", path, e
+                )
+                # Remove corrupted entries
+                path.unlink(missing_ok=True)
```
mscolnick: may want another `except` here

marimo/_server/session_registry.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def _sessions_dir() -> Path:
+    return marimo_state_dir() / SESSIONS_DIR_NAME
+
+
+def _entry_path(server_id: str) -> Path:
+    # Replace colons/slashes with underscores for safe filenames
+    safe_id = server_id.replace(":", "_").replace("/", "_")
+    return _sessions_dir() / f"{safe_id}.json"
```
mscolnick: maybe already safe, but the host could also have `[]`

PR #8599, opened 2026-03-06
fix: `mo.mpl.interactive` renders blank in `marimo run` mode (#8452)

### LINKED ISSUE #8452: Matplotlib figures complaining about being started outside of main thread when run as app
### Describe the bug

`/tmp/marimo_17894/__marimo__cell_Hbol_.py:1: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.`
Getting this error for `marimo>=0.19.0` when creating Matplotlib figures for use with `mo.mpl.interactive` with `marimo run`. `marimo edit` works fine and displays the figure, and so does `marimo==0.18.4` with `marimo run`.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.2",
  "editable": false,
  "location": "/home/ec2-user/miniconda3/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.1.158-180.294.amzn2023.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.13.5",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.19",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.12.0",
    "packaging": "24.2",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.3.2",
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "openai": "2.21.0",
    "pandas": "2.3.1",
    "polars": "1.32.0",
    "pyarrow": "21.0.0",
    "ruff": "0.12.4",
    "sqlglot": "27.5.1"
  },
  "Experimental Flags": {
    "performant_table_charts": false,
    "rtc_v2": false,
    "chat_modes": false
  }
}
```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.20.2"

with app.setup:
    import marimo as mo


@app.cell
def _():
    fig = plt.figure()
    ax = fig.add_subplot()
    ax.plot([1, 2], [3, 4])

    mo.mpl.interactive(fig)
    return


if __name__ == "__main__":
    app.run()
```

mscolnick:
Closes #8452

## Description

Since v0.19.0 `marimo run` executes kernels in threads instead of
subprocesses. In run mode the session is auto-instantiated from the
WebSocket handler, and the browser WebSocket API does not support
custom headers — so the `x-runtime-url` header that `mpl.interactive`
relies on is absent.

When `_get_remote_url()` returned `""`, `_template()` still prepended
the proxy-style `/mpl/{fig_id}` prefix to every URL. But the embedded
matplotlib server only exposes top-level routes (`/ws`, `/_static/…`,
`/mpl.js`), so all static assets and the WebSocket connection 404'd,
producing a blank rectangle.

### Changes

- **`marimo/_plugins/stateless/mpl/_mpl.py`**
  - `_template()`: when there is no remote URL, generate direct URLs
    to the embedded server *without* the `/mpl/{fig_id}` proxy prefix.
  - Suppress the harmless "Starting a Matplotlib GUI outside of the
    main thread" `UserWarning` around WebAgg figure-manager creation
    (the rendering is Agg-based, no actual GUI toolkit is involved).

### DIFF
--- a/marimo/_plugins/stateless/mpl/_mpl.py
+++ b/marimo/_plugins/stateless/mpl/_mpl.py
@@ -14,6 +14,7 @@ import mimetypes
 import os
 import threading
 import time
+import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, Union
 
@@ -211,10 +212,21 @@ def _convert_scheme_to_ws(url: str) -> str:
 
 
 def _template(fig_id: str, port: int) -> str:
-    base_url = _get_remote_url() or f"http://localhost:{port}/"
-    base_url_and_path = f"{base_url}/mpl/{fig_id}"
-    base_url_and_path_ws = f"{base_url}/mpl/{port}"
-    ws_base_url = _convert_scheme_to_ws(base_url_and_path_ws)
+    remote_url = _get_remote_url()
+    if remote_url:
+        # Route through the marimo server proxy, which expects
+        # /mpl/{fig_id}/... for static assets and /mpl/{port}/ws for
+        # the WebSocket connection.
+        base_url_and_path = f"{remote_url}/mpl/{fig_id}"
+        ws_base = f"{remote_url}/mpl/{port}"
+    else:
+        # Direct access to the embedded matplotlib server.  The server
+        # only has top-level routes (/_static, /ws, /mpl.js, …), so we
+        # must NOT add the /mpl/{fig_id} prefix that the proxy expects.
+        base_url_and_path = f"http://localhost:{port}"
+        ws_base = f"http://localhost:{port}"
+
+    ws_base_url = _convert_scheme_to_ws(ws_base)
 
     return html_content % {
         "ws_uri": f"{ws_base_url}/ws?figure={fig_id}",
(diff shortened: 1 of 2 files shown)


marimo/_plugins/stateless/mpl/_mpl.py
```diff
@@ -211,10 +212,21 @@ def _convert_scheme_to_ws(url: str) -> str:
...
+        # /mpl/{fig_id}/... for static assets and /mpl/{port}/ws for
+        # the WebSocket connection.
+        base_url_and_path = f"{remote_url}/mpl/{fig_id}"
+        ws_base = f"{remote_url}/mpl/{port}"
+    else:
+        # Direct access to the embedded matplotlib server.  The server
+        # only has top-level routes (/_static, /ws, /mpl.js, …), so we
+        # must NOT add the /mpl/{fig_id} prefix that the proxy expects.
+        base_url_and_path = f"http://localhost:{port}"
+        ws_base = f"http://localhost:{port}"
```
dmadisetti, line 227: This value is rendered to the user right? I don't think this will work outside of running locally.

mscolnick: oh yes you are right. let me see if I can get the request url another way then

mscolnick:
pulling the plug on the websocket approach: https://github.com/marimo-team/marimo/pull/8612

PR #8611, opened 2026-03-07
fix: isolation when running multiple notebooks in an app server

akshayka:
**Summary.** This PR introduces process-level isolation when serving multiple apps from the same server (`marimo run directory/`, `create_asgi_app()`); clients of any given app are still run in threads in the app's process for efficiency. This fixes a critical bug in which different apps shared the same Python globals, leading to undefined behavior such as collisions in `sys.modules. It does make multi-app servers consume slightly more RAM.

Process isolation also applies to multi-app servers started with `--sandbox`. (This PR also fixes a bug in which `--sandbox` run servers silently dropped stdout/stderr.)

**Dependency on pyzmq.** The proposed implementation also adds a dependency on `pyzmq` for multi-app servers, to pave the way for  allowing sandboxed multi-app servers . It would be possible to design a different solution that used multiprocessing instead of `pyzmq`, at the cost of not supporting package sandboxes. It is perhaps worth discussing whether we are okay with making pyzmq a required dependency of marimo. 

**Feature flag.** App isolation is feature-flagged, currently opt-in.

**Context.** When marimo was originally designed, `marimo run` only ever served a single notebook. A single process could safely serve multiple clients of the same notebook since they all share the same code.

When multiple app serving was introduced, we continued serving all clients from a single process, even though the clients were potentially running different programs. When two notebooks both `import utils` but expect different implementations from different directories, whichever app loads first wins, and the second app silently gets the wrong module. Similar problems may exist for other Python globals.

**This PR.** This PR fixes the problem by running each app in its own OS process. Multiple clients of the *same* app still share a process (as kernel threads), which allows for fast and cheap sessions. The isolation boundary is per-app, not per-client.

```
Before (shared process — sys.modules collisions):

  ┌──────────────── Main Process ─────────────────┐
  │                                               │
  │   Kernel(app1, client A)    ← all kernels     │
  │   Kernel(app1, client B)      share one       │
  │   Kernel(app2, client C)      sys.modules     │
  │   Kernel(app2, client D)                      │
  └───────────────────────────────────────────────┘

After (per-app process isolation):

  ┌──────────────── Main Process ─────────────────┐
  │           (HTTP, WebSocket, routing)           │
  └──────────────────┬──────────────┬──────────────┘
                     │ ZMQ          │ ZMQ
       ┌─────────────▼──┐   ┌──────▼──────────────┐
       │  App Process    │   │  App Process        │
       │  (app1.py)      │   │  (app2.py)          │
       │                 │   │                     │
       │  Kernel: cl. A  │   │  Kernel: cl. C      │
       │  Kernel: cl. B  │   │  Kernel: cl. D      │
       └─────────────────┘   └─────────────────────┘
        isolated sys.modules  isolated sys.modules
```

**IPC.** Each app process communicates with the main process over 4 shared ZeroMQ sockets (not per-client). Kernel commands and stream output are multiplexed over these shared channels using session ID tagging:

```
IPC channels (4 shared ZMQ sockets per app process):

  Main Process                             App Subprocess
  ────────────                             ──────────────
  mgmt     [PUSH] ─────────────────────▶ [PULL]  mgmt loop
  response [PULL] ◀───────────────────── [PUSH]  (create/stop kernel)
  cmd      [PUSH] ──[sid, channel, msg]─▶ [PULL]  dispatcher ──▶ kernel queues
  stream   [PULL] ◀──[sid, msg]───────── [PUSH]  collector  ◀── kernel output
```

```
Per session (main-process side):

  AppProcessQueueManager
    control_queue    ──┐
    ui_element_queue ──┤──▶ cmd socket (tagged with session_id)
    completion_queue ──┤
    input_queue      ──┘
    stream_queue     ◀──── stream receiver thread (regular Queue)
```

**When this path is activated.**

- `create_asgi_app()`: always enables process isolation (the whole point is multi-app)
- `marimo run app1.py app2.py` / directory serving: auto-enables when multiple files detected
- `marimo run app.py` (single file): no change, uses existing thread-based kernels
- `marimo edit`: no change, uses existing process-based kernels

### DIFF
--- a/tests/_session/app_host/test_app_host.py
+++ b/tests/_session/app_host/test_app_host.py
@@ -446,9 +446,7 @@ class TestAppHostMultipleClients:
                     config_manager=Mock(
                         with_overrides=Mock(
                             return_value=Mock(
-                                get_config=Mock(
-                                    return_value=DEFAULT_CONFIG
-                                )
+                                get_config=Mock(return_value=DEFAULT_CONFIG)
                             )
                         )
                     ),


marimo/_server/asgi.py
```diff
@@ -462,6 +462,12 @@ async def root():
     else:
         auth_token = AuthToken(token)
 
+    from marimo._dependencies.dependencies import DependencyManager
+
+    DependencyManager.zmq.require(
+        "for running multiple notebooks with create_asgi_app()"
+    )
```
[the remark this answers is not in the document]
mscolnick: this is right. but if we are making `pyzmq` required, do we still ned this?

marimo/_session/app_host/host.py
```diff
@@ -0,0 +1,310 @@
...
+            user_config=user_config,
+            virtual_files_supported=virtual_files_supported,
+            redirect_console_to_browser=redirect_console_to_browser,
+            log_level=log_level,
+        )
+        self._conn.mgmt.send(encode_mgmt_command(cmd))
+
+        response_timeout_ms = 30_000
+        if self._conn.response.poll(timeout=response_timeout_ms):
+            data = self._conn.response.recv()
```
mscolnick: do wee need to check that this create command is only sent once? what happens if you open the same app at the same time while it starts up?

marimo/_server/asgi.py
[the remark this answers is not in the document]
akshayka: We had the discussion of only requiring pyzmq for python < 3.15

marimo/_session/queue.py
```diff
@@ -30,6 +33,32 @@ def put_nowait(self, item: T, /) -> None: ...
     def empty(self) -> bool: ...
 
 
+def route_control_request(
```
mscolnick, line 36: nice dedupe/dry

marimo/_session/app_host/host.py
```diff
@@ -0,0 +1,310 @@
...
+            try:
+                if self._conn is None:
+                    break
+
+                frames = self._conn.stream.recv_multipart()
+                session_id = frames[0].decode()
+                payload = pickle.loads(frames[1])
+
+                if isinstance(payload, KernelExited):
+                    self._session_ids.discard(session_id)
```
mscolnick: should we lock between the discard and the empty check? tiny but exists

marimo/_server/asgi.py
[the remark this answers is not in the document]
mscolnick: ah ok

akshayka:
There's a race on app startup I need to fix before merging

marimo/_session/managers/ipc.py
```diff
@@ -115,66 +114,17 @@ def close_queues(self) -> None:
         self._ipc.close_queues()
 
     def put_control_request(self, request: commands.CommandMessage) -> None:
-        # Completions are on their own queue
-        if isinstance(request, commands.CodeCompletionCommand):
-            self.completion_queue.put(request)
-            return
-
-        self.control_queue.put(request)
-        # Update UI elements are on both queues so they can be batched
```
mscolnick, line 124: looks like there was a bug - missing `ModelCommand` that is now fixed by the DRYness

marimo/_session/app_host/host.py
```diff
@@ -0,0 +1,381 @@
...
+        self._process = subprocess.Popen(
+            cmd,
+            stdin=subprocess.PIPE,
+        )
+
+        proc_stdin = self._process.stdin
+        if proc_stdin is None:
+            self.shutdown()
+            raise RuntimeError("Failed to open stdin for app host")
+        proc_stdin.write(args.encode_json())
```
mscolnick: possible to get errors? move into the try/except?

marimo/_session/app_host/host.py
```diff
@@ -0,0 +1,381 @@
...
+                # _closed.  A blocking recv_multipart() would prevent
+                # context.destroy() from completing.
+                if not conn.stream.poll(timeout=1000):
+                    # If the subprocess died, no more KernelExited
+                    # sentinels will arrive.  Fire _on_empty so the
+                    # pool can clean up this AppHost.
+                    if not self.is_alive():
+                        with self._session_lock:
+                            self._session_ids.clear()
+                        self._fire_on_empty()
```
mscolnick, line 127: worth clearing the `_stream_receivers` like in shutdown?

```
        with self._stream_lock:
            for q in self._stream_receivers.values():
                try:
                    q.put(None)
                except Exception:
                    pass
            self._stream_receivers.clear()
```

akshayka: `on_empty` triggers the shutdown logic

PR #8615, opened 2026-03-09
fix: forward IPC kernel subprocess output to console

koaning:
## Summary

- When using `marimo run --sandbox .` (multi-file sandbox mode), cell errors and print output were silently lost because the IPC kernel subprocess's stdout/stderr pipes were captured for the `KERNEL_READY` handshake but never read again
- Spawn daemon threads after startup to continuously forward the subprocess's stdout/stderr to the parent process, matching the behavior of single-file `marimo run`

## Test plan

- [x] Added `TestForwardSubprocessOutput::test_stdout_and_stderr_are_forwarded` — verifies both stdout and stderr from a subprocess are forwarded after the KERNEL_READY line is consumed
- [x] Added `TestForwardSubprocessOutput::test_handles_closed_dest_gracefully` — verifies no crash when destination stream is closed during forwarding
- [ ] Manual: run `marimo run --sandbox .` with a notebook containing `raise ValueError(...)` and confirm the traceback appears in the console

### DIFF
--- a/tests/_cli/test_cli.py
+++ b/tests/_cli/test_cli.py
@@ -1095,12 +1095,9 @@ def test_cli_run_sandbox_directory_shows_cell_errors() -> None:
                 break
             time.sleep(0.5)
 
-        stderr_text = b"".join(stderr_chunks).decode(
-            errors="replace"
-        )
+        stderr_text = b"".join(stderr_chunks).decode(errors="replace")
         assert "sandbox_error_test" in stderr_text, (
-            "Expected cell error in stderr, "
-            f"got:\n{stderr_text}"
+            f"Expected cell error in stderr, got:\n{stderr_text}"
         )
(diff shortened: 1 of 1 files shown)


marimo/_session/managers/ipc.py
```diff
@@ -376,6 +383,47 @@ def kernel_connection(self) -> TypedConnection[KernelMessage]:
...
+) -> None:
+    """Read lines from *pipe* and write them to *dest* until EOF."""
+    assert pipe is not None
+    try:
+        for line in pipe:
+            try:
+                dest.write(line.decode("utf-8", errors="replace"))
+                dest.flush()
+            except ValueError:
+                # dest was closed (e.g. during shutdown)
```
akshayka: You may need buffering for performance

PR #8619, opened 2026-03-09
fix: tag install location on ModuleNotFound to distinguish installation on server vs kernel

dmadisetti:
## 📝 Summary

Tag installation requests with either server-side or notebook level installation for `--sandbox` mode.
Tested in:

- [x]  - sandbox
- [ ]  - sandbox home
- [x]  - marimo-jupyter-extension

We might consider a more visual indication of installation location?

### DIFF
--- a/tests/_server/api/endpoints/test_export.py
+++ b/tests/_server/api/endpoints/test_export.py
@@ -451,8 +451,6 @@ def test_auto_export_ipynb_missing_nbformat_notifies_once(
     """Missing-nbformat alert fires at most once per session."""
     from unittest.mock import patch
 
-    from marimo._server.api.utils import notify_server_missing_packages
-
     session = get_session_manager(client).get_session(SESSION_ID)
     assert session
     session.app_file_manager.filename = temp_marimo_file


marimo/_runtime/packages/utils.py
```diff
@@ -44,6 +44,24 @@ def is_python_isolated() -> bool:
     )
 
 
+def can_server_auto_install() -> bool:
```
mscolnick: maybe just better to prompt the user? its not that big of a deal, and they can change the version (and maybe index-url in future)

marimo/_dependencies/dependencies.py
```diff
@@ -284,7 +285,11 @@ def which(pkg: str) -> bool:
         return shutil.which(pkg) is not None
 
     @staticmethod
-    def require_many(why: str, *dependencies: Dependency) -> None:
+    def require_many(
+        why: str,
+        *dependencies: Dependency,
+        source: Literal["kernel", "server"] = "kernel",
```
mscolnick: should we just make this required? this forces the caller to add this and wont lead to bugs

marimo/_dependencies/errors.py
```diff
@@ -1,14 +1,23 @@
...
 
     package_names: list[str]
+    source: Literal["kernel", "server"]
 
-    def __init__(self, package_names: list[str], msg: str) -> None:
+    def __init__(
+        self,
+        package_names: list[str],
+        msg: str,
+        source: Literal["kernel", "server"] = "kernel",
```
mscolnick: same here; should we just make this required? this forces the caller to add this and wont lead to bugs

marimo/_messaging/notification.py
```diff
@@ -336,12 +336,16 @@ class MissingPackageAlertNotification(
...
+        isolated: Whether auto-install is possible in this environment.
+        source: Which Python environment to install into. "kernel" (default)
+                installs in the kernel's venv; "server" installs in the
+                server's own Python env (e.g. for formatter tools like ruff).
     """
 
     name: ClassVar[str] = "missing-package-alert"
     packages: list[str]
     isolated: bool
+    source: Literal["kernel", "server"] = "kernel"
```
mscolnick, line 348: can we reduce the places where we have this default? ideally it all gets funneled through and don't need defaults

mscolnick: cc @manzt does this need to be optional to avoid breaking marimo-LSP

marimo/_server/api/endpoints/export.py
```diff
@@ -460,7 +461,24 @@ async def auto_export_as_ipynb(
...
-            LOGGER.error("Cannot snapshot to IPYNB: nbformat not installed")
+            from marimo._messaging.notification import (
+                MissingPackageAlertNotification,
+            )
+            from marimo._runtime.packages.utils import can_server_auto_install
+            from marimo._session.utils import send_message_to_consumer
+
+            LOGGER.warning("Cannot snapshot to IPYNB: nbformat not installed")
+            session_id = app_state.get_current_session_id()
+            if session_id is not None:
```
mscolnick: we should do this once per session. if it fails to install, it might keep popping up

marimo/_server/api/endpoints/export.py
```diff
@@ -460,7 +461,24 @@ async def auto_export_as_ipynb(
...
+                send_message_to_consumer(
+                    session=session,
+                    operation=MissingPackageAlertNotification(
+                        packages=["nbformat"],
+                        isolated=can_server_auto_install(),
```
mscolnick: i think if we are running `marimo edit nb.py --sandbox` this might fail, and we should be able to install.

marimo/_server/api/endpoints/editing.py
```diff
@@ -229,4 +244,29 @@ async def install_missing_packages(request: Request) -> BaseResponse:
...
+            for pkg, version in cmd.versions.items():
+                await pkg_manager.install(pkg, version=version or None)
+        return SuccessResponse()
+
+    # Default ("kernel"): dispatch to kernel via ZeroMQ control queue.
+    app_state.require_current_session().put_control_request(
+        cmd,
+        from_consumer_id=ConsumerId(app_state.require_current_session_id()),
+    )
+    return SuccessResponse()
```
mscolnick: this could still end with `    return await dispatch_control_request(request, InstallPackagesRequest)` right?

marimo/_server/api/endpoints/editing.py
```diff
@@ -140,14 +142,27 @@ async def format_cell(request: Request) -> FormatResponse:
...
-        if app_state.session_manager.sandbox_mode is SandboxMode.MULTI:
-            # Re-raise without name so error handler won't send install notification
-            raise ModuleNotFoundError(
-                "Server does not have a formatter. Please install ruff"
-            ) from None
-        raise
+        session_id = app_state.get_current_session_id()
+        session = app_state.get_current_session()
+        if session_id is not None and session is not None:
+            send_message_to_consumer(
```
mscolnick: this logic (although maybe only 20 lines) seems deduped. could be worth DRYing

marimo/_server/api/endpoints/editing.py
```diff
@@ -234,4 +247,24 @@ async def install_missing_packages(request: Request) -> BaseResponse:
...
+    )
+
+    body = await parse_request(request, cls=InstallPackagesRequest)
+    cmd = body.as_command()
+
+    if cmd.source == "server":
+        # Install into the server's own Python environment (sys.executable).
+        # Used when the server itself needs a package (e.g. nbformat for
+        # IPYNB auto-export when running with --sandbox).
+        pkg_manager = create_package_manager(cmd.manager, python_exe=None)
```
mscolnick: do we want to use `python_exe=sys.executable`? is it safe long-term to rely on the default?

marimo/_server/errors.py
```diff
@@ -88,17 +88,21 @@ async def handle_error(request: Request, response: Any) -> Any:
...
                     packages = response.package_names
+                    source = response.source
                 elif isinstance(response, ModuleNotFoundError):
                     if not response.name:
                         return JSONResponse(
                             {"detail": str(response)}, status_code=500
                         )
                     packages = [response.name]
+                    source = "kernel"
+                isolated = True if source == "server" else is_python_isolated()
```
mscolnick, line 101: why is it isolated if `source == "server"`?

marimo/_session/managers/ipc.py
```diff
@@ -261,6 +262,11 @@ def start_kernel(self) -> None:
                 err=True,
             )
 
+            # Override UV env vars so the kernel subprocess sees the sandbox
+            # venv as its environment, not the outer uv project.
+            env["VIRTUAL_ENV"] = str(Path(venv_python).parent.parent)
+            env.pop("UV_PROJECT_ENVIRONMENT", None)
+
```
mscolnick: are there tests for this? maybe we can have a function `construct_env` which takes in the previous env and compares the expected output

tests/_server/api/endpoints/test_editing.py
```diff
@@ -113,6 +113,24 @@ def test_install_missing_packages(client: TestClient) -> None:
     assert "success" in response.json()
 
 
+@with_session(SESSION_ID)
+def test_install_missing_packages_server_source(client: TestClient) -> None:
```
mscolnick, line 146: this does not actually test what it says?

marimo/_messaging/notification.py
manzt: I _think_ this should be ok because we just decode JSON to dict to forward to LSP:

https://github.com/marimo-team/marimo-lsp/blob/7ca67543511726ba6972667787a3b2a7a322018e/src/marimo_lsp/session_consumer.py#L30-L33

Issue #7518, opened 2025-12-16
Copy or filter by md/html cell value in table errors out


yan-hic:
### Describe the bug

I was using the example at https://github.com/marimo-team/marimo/issues/7506#issuecomment-3657258640 to test behavior when cell is not plain-text. 

- Filtering by cell raises below:

`Function call (name: search, args: {'query': '', 'filters': [{'column_id': 'k', 'operator': 'not_in', 'value': [{'_serialized_mime_bundle': {'mimetype': 'text/markdown', 'data': '<span class="markdown prose dark:prose-invert contents"><span class="paragraph"><a href="https://www.google.com/search?q=14" rel="noopener noreferrer" target="_blank">14</a></span></span>'}}]}], 'page_number': 0, 'page_size': 10}) failed with exception 'is_in' cannot check for Int64 values in List(Struct({'_serialized_mime_bundle': Struct({'mimetype': String, 'data': String})})) data Resolved plan until failure: ---> FAILED HERE RESOLVING 'sink' <--- FILTER [(col("k").is_in([[{{"text/markdown","<span class="markdown prose dark:prose-invert contents"><span class="paragraph"><a href="https://www.google.com/search?q=14" rel="noopener noreferrer" target="_blank">14</a></span></span>"}}]]).not()) | (col("k").is_null())].strict_cast(Boolean) FROM DF ["_marimo_row_id", "k"]; PROJECT */2 COLUMNS`

- Copying does not fail but copies this:

`
{"_serialized_mime_bundle":{"mimetype":"text/markdown","data":"<span class=\"markdown prose dark:prose-invert contents\"><span class=\"paragraph\"><a href=\"https://www.google.com/search?q=72\" rel=\"noopener noreferrer\" target=\"_blank\">72</a></span></span>"}}
`

### Will you submit a PR?

- [ ] Yes

### Environment

0.18.4

mscolnick:
This is definitely a bug. We do the formatting on the backend, so we will need to send the unformatted code as well as the formatted display, which will likely touch a few code paths.

Light2Dark:
> This is definitely a bug. We do the formatting on the backend, so we will need to send the unformatted code as well as the formatted display, which will likely touch a few code paths.

I'm thinking we can parse `serialized_mime_bundle` in the frontend to extract the "rendered" value.

yan-hic:
bumping - error is showing on filter-by-value (right click) and when copy/pasting cell value (not in copy but in rendered pasted text).

[...]

PR #8622, opened 2026-03-09
Pass raw data for tables during search and copy text/html to clipboard

Light2Dark:
## 📝 Summary

Closes https://github.com/marimo-team/marimo/issues/7518.

When there is format_mapping, we pass raw data alongside data. So, the frontend will use the rawData to filter datasets.

This also copies both `text/plain` & `text/html` to clipboard, therefore fixing the original issue to copy text/html values.

## 📋 Checklist

### DIFF
--- a/frontend/src/components/data-table/utils.ts
+++ b/frontend/src/components/data-table/utils.ts
@@ -41,7 +41,7 @@ export async function loadTableData<T = object>(
 export async function loadTableAndRawData<T>(
   tableData: TableData<T>,
   rawTableData?: TableData<T> | null,
-): Promise<[TableData<T>, TableData<T> | undefined]> {
+): Promise<[T[], T[] | undefined]> {
   if (rawTableData) {
     return Promise.all([loadTableData(tableData), loadTableData(rawTableData)]);
   }


frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -587,13 +597,18 @@ export const LoadingDataTableComponent = memo(
...
         const searchResults = await searchResultsPromise;
         tableData = searchResults.data;
+        rawTableData = searchResults.raw_data;
         totalRows = searchResults.total_rows;
         cellStyles = searchResults.cell_styles || {};
         cellHoverTexts = searchResults.cell_hover_texts || {};
       }
       tableData = await loadTableData(tableData);
+      const rawRows = rawTableData
+        ? await loadTableData(rawTableData)
```
mscolnick: could we `Promise.all()` this? i assume they are not filenames so maybe they dont hit a waterfall. 

might be easier to pull into a function for readability 
```ts
function load(): Promise<[Data, Data | undefined]> {
   if (rawTableData) {
        return Promise.all(loadTableData(), loadTableData())
   }

    ...
}
```

frontend/src/utils/copy.ts
```diff
@@ -2,21 +2,34 @@
...
     return;
   }
 
-  await navigator.clipboard.writeText(text).catch(async () => {
-    // Fallback to prompt
+  if (html && navigator.clipboard.write) {
+    try {
+      const item = new ClipboardItem({
+        "text/html": new Blob([html], { type: "text/html" }),
+        "text/plain": new Blob([text], { type: "text/plain" }),
```
mscolnick, line 23: nice!

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1369,7 +1377,9 @@ def clamp_rows_and_columns(manager: TableManager[Any]) -> str:
                 data = data.select_columns(columns_to_select)
 
             try:
-                return data.to_json_str(self._format_mapping)
+                formatted = data.to_json_str(self._format_mapping)
+                raw = data.to_json_str() if self._format_mapping else None
```
mscolnick, line 1381: small, small optimization we could consider later is only including the raw data for the columns that have `format_mapping`, so we dont serialize more than needed. but since its only 10-20 rows at a time, i think this is ok without the complexity

mscolnick: on second thought, is this needed for search though to always be enabled?

frontend/src/components/data-table/utils.ts
```diff
@@ -87,3 +89,88 @@ export function stringifyUnknownValue(opts: {
...
+    }
+  }
+  return cell.getValue();
+}
+
+/**
+ * Get the raw (unformatted) value for a row/column from the table,
+ * or undefined if raw data is not available.
+ */
+export function getRawRowValue<TData>(
```
mscolnick: similar name to `getRawCellValue`. could you add a small note when this should be used over the other?

PR #8626, opened 2026-03-10
fix: re-render mutable Html children in vstack/hstack on flush

### LINKED ISSUE #8618: Spinner update not working when placed manually inside layout
### Describe the bug

I want to control the location of a spinner that shows up when the user clicks a button. To do this, I wanted to use vstack or something similar. However, when I place a spinner inside a stacked UI element, then the updates are not pushed. The only way to get updates, is not accept that the output is manually appended. I have added a minimal example

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.4",
  "editable": false,
  "location": "c:/Users/elmoe/work/gitlab/borehole-diagnostics-tool/.venv/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.12.9",
  "Locale": "--",
  "Binaries": {
    "Browser": "145.0.7632.110",
    "Node": "v20.18.1",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.17.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.41.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.20.4"
app = marimo.App()


@app.cell
def _():
    import marimo as mo
    import asyncio

    return asyncio, mo


@app.cell
def _(mo):
    run_btn = mo.ui.run_button(label="Run spinner stacked")
    return (run_btn,)


@app.cell
def _(mo):
    run_btn_2 = mo.ui.run_button(label="Run spinner inside layout")
    return (run_btn_2,)


@app.cell
async def _(asyncio, mo, run_btn):
    async def spinner_stacked():

        ui = mo.vstack([
            run_btn,
            mo.md("## Spinner below")
        ])
        mo.output.replace(ui)
        if run_btn.value:
            with mo.status.spinner(title="Initial...") as spinner:

                for i in range(1, 4):
                    await asyncio.sleep(1)
                    spinner.update(f"Step {i} of 3...")
        mo.output.append(mo.md("Done"))

    await spinner_stacked()
    return


@app.cell
async def _(asyncio, mo, run_btn_2):


    async def run_with_spinner_in_layout():
        ui_0 = mo.vstack([
            run_btn_2,
            mo.md("## Spinner below")
        ])
        mo.output.replace(ui_0)
        if run_btn_2.value:
            with mo.status.spinner(title="Initial…") as spinner:

                ui = mo.vstack([
                    run_btn_2,
                    mo.md("## Spinner below"),
                    spinner,
                    mo.md("_If updates worked, you would see Step 1, Step 2, Step 3 above spinner with this line shown as well_"),
                ])
                mo.output.replace(ui)

                # These updates are NOT shown: the replaced output is static
                for i in range(1, 4):
                    await asyncio.sleep(1)
                    spinner.update(f"Step {i} of 3…")
            mo.output.replace(ui_0)
            mo.output.append(mo.md("Done"))


    await run_with_spinner_in_layout()
    return


if __name__ == "__main__":
    app.run()

### PR DESCRIPTION
**Pull request by VishakBaddur**
Fixes #8618

## Problem
`mo.status.spinner` updates were silently lost when the spinner was placed inside a layout container like `mo.vstack` or `mo.hstack`.

Root cause: `_flex()` eagerly evaluated `as_html(item).text` for each child at construction time, baking a frozen HTML string into the layout. When `spinner.update()` later mutated `spinner._text` and called `output.flush()`, the parent vstack re-sent its own frozen HTML, the spinner's new state was never visible.

## Fix
`_FlexContainerHtml` now stores live `Html` references instead of frozen strings, and overrides the `text` property to call `_build_text()` on every access. This means each `flush()` re-reads the current `.text` of every child, propagating updates through the layout hierarchy.

## Testing
Added `test_mutable_html_children_update_live` to `tests/_plugins/stateless/test_flex.py` — verifies that mutating a child's `_text` after embedding it in `vstack`/`hstack` is reflected in the parent's `.text`.

### DIFF
--- a/marimo/_plugins/stateless/flex.py
+++ b/marimo/_plugins/stateless/flex.py
@@ -27,7 +27,7 @@ class _FlexContainerHtml(Html):
 
     def __init__(
         self,
-        style: str,
+        style: Optional[str],
         live_children: list[Html],
         child_flexes: Optional[Sequence[Optional[float]]],
     ) -> None:
@@ -62,7 +62,7 @@ class _FlexContainerHtml(Html):
                 h.div(c.text, style=_item_style(i))
                 for i, c in enumerate(self._live_children)
             ]
-        return h.div(grid_items, style=self._style)
+        return h.div(grid_items, style=self._style or "")
 
     @property
     def text(self) -> str:  # type: ignore[override]

### CODE REVIEW

**Comment by VishakBaddur** [comment]

**Comment by VishakBaddur** [comment]
recheck

File: marimo/_plugins/stateless/flex.py [hunk 1]
```diff
@@ -60,32 +105,12 @@ def _flex(
...
-    def create_style_for_item(idx: int) -> Optional[str]:
-        if child_flexes is None:
-            return ""
-        child_flex = child_flexes[idx]
-        if child_flex is None:
-            return ""
-        style: dict[str, str | int | float | None] = {"flex": f"{child_flex}"}
-        # Only make the wrapper a flex container for nested stacks so their
-        # flex: 1 and justify work. Leaf content (e.g. mo.stat) fills the
-        # wrapper when it is a block.
```
Light2Dark, line 72: we should preserve the original comments

VishakBaddur: Good catch, restored the original comments. Thanks!

VishakBaddur:
recheck

PR #8630, opened 2026-03-10
add global previewed tables state

Light2Dark:
## 📝 Summary

Adds a global variable to track previewed lazy tables. Tabs are being used for lazy tables, which means the table completely unmounts. So, we form a stable Id based on object-id and random-id to preserve whether it has been previewed or not. Object-id and random-id are attributes of the component/

Other alternatives are to use Activity, but since table is unmounted, this doesn't work. We could maybe make a more global change to `tabs`. Or..not pass tabs from the backend, maybe just pass table alongside query plan

https://github.com/user-attachments/assets/130ee83a-f75a-4d89-a8b1-1ba0fa4aa610

## 📋 Checklist

### DIFF
--- a/frontend/src/plugins/impl/DataTablePlugin.tsx
+++ b/frontend/src/plugins/impl/DataTablePlugin.tsx
@@ -60,7 +60,7 @@ import { ContextAwarePanelItem } from "@/components/editor/chrome/panels/context
 import { Alert, AlertTitle } from "@/components/ui/alert";
 import { Button } from "@/components/ui/button";
 import { DelayMount } from "@/components/utils/delay-mount";
-import { type CellId, findCellId } from "@/core/cells/ids";
+import { type CellId, findCellId, UIElementId } from "@/core/cells/ids";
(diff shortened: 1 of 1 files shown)


mscolnick:
somewhat hacky, but pretty pragmatic. i think this would be ok to ship this workaround

frontend/src/plugins/impl/DataTablePlugin.tsx
```diff
@@ -377,21 +382,66 @@ export const DataTablePlugin = createPlugin<S>("marimo-table")
     );
   });
 
+/**
+ * Tracks which lazy tables have been previewed across remounts (e.g. tab switches).
+ * Keyed by objectId (stable across remounts) with randomId as value
+ * (changes on cell re-execution, so stale entries are naturally invalidated).
+ */
+const previewedTables = new Map<string, string>();
```
mscolnick: `objectId` can be typed to `UIElementId`

PR #8636, opened 2026-03-11
fix: propagate notebook `__doc__` to cell execution namespace (#8634)

### LINKED ISSUE #8634: `__doc__` does not act as expected
### Describe the bug

The following notebook does not return the `__doc__` object as expected:
~~~
"""This is a test"""

import marimo

__generated_with = "0.20.2"


@app.cell
def _():
    __doc__
    return


if __name__ == "__main__":
    app.run()
~~~

# Expected result

```'This is a test'```

# Actual result
```'Created for the marimo kernel'```

# Additional Information

1. If you add a new cell with ```from my_notebook import __doc__```, the result is then correct.
2. If you then delete the new import cell, the result is once again incorrect, but different:
~~~
"Built-in functions, types, exceptions, and other objects.\n\nThis module provides direct access to all 'built-in'\nidentifiers of Python; for example, builtins.len is\nthe full name for the built-in function len().\n\nThis module is not normally accessed explicitly by most\napplications, but can be useful in modules that provide\nobjects with the same name as a built-in value, but in\nwhich the built-in of that name is also needed."
~~~

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.2",
  "editable": false,
  "location": ".../.venv/lib/python3.12/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "23.5.0",
  "Processor": "i386",
  "Python Version": "3.12.9",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "145.0.7632.160",
    "Node": "v23.7.0",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.1",
    "narwhals": "2.15.0",
    "packaging": "26.0",
    "psutil": "7.2.1",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "pandas": "3.0.1"
  },
  "Experimental Flags": {
    "layouts": true
  }
}
```

</details>

### Code to reproduce

```
"""This is a test"""

import marimo

__generated_with = "0.20.2"


@app.cell
def _():
    __doc__
    return


if __name__ == "__main__":
    app.run()
```

### PR DESCRIPTION
**Pull request by mscolnick**
When a marimo notebook has a module-level docstring (e.g., `"""This is a
test"""`), cells referencing `__doc__` would return `"Created for the
marimo kernel."` instead of the notebook's actual docstring.

## Root cause

`create_main_module()` in `patches.py` always hardcodes `doc="Created
for the marimo kernel."` when creating the `__main__` module. The
notebook's header docstring (parsed and stored in `App._header`) was
never propagated to the execution namespace.

## Changes

- Added `extract_docstring_from_header()` utility in `patches.py` that
  uses `ast.get_docstring()` to extract the actual string value from the
  raw header text.
- Added optional `doc` parameter to `create_main_module()` and
  `patch_main_module()` to allow setting a custom `__doc__`.
- Added `docstring` field to `AppMetadata` so it can be threaded from
  notebook loading through to kernel creation.
- Updated all `AppMetadata` creation sites (session manager, export,
  pyodide bootstrap) to populate `docstring` from the app's header.
- Updated `AppScriptRunner` and `AppKernelRunner` to pass the docstring
  directly to `create_main_module()`.

Closes #8634

### DIFF
--- a/docs/examples/markdown/admonitions.md
+++ b/docs/examples/markdown/admonitions.md
@@ -1,5 +1,6 @@
 # Admonitions
 
+marimo supports admonition blocks using the [PyMdown Extensions Blocks](https://facelessuser.github.io/pymdown-extensions/extensions/blocks/) syntax. See the [Admonition plugin docs](https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/admonition/) for the full format and options.
 
 /// marimo-embed-file
     size: xlarge
--- a/frontend/e2e-tests/global-setup.ts
+++ b/frontend/e2e-tests/global-setup.ts
@@ -23,7 +23,7 @@ async function globalSetup(_config: FullConfig) {
       while (retries > 0) {
         try {
           await page.goto(url, {
-            waitUntil: "networkidle",
+            waitUntil: "load",
             timeout: 15_000,
           });
           console.log(`✅ ${app} ready`);
--- a/frontend/e2e-tests/global-teardown.ts
+++ b/frontend/e2e-tests/global-teardown.ts
@@ -10,10 +10,14 @@ async function globalTeardown() {
   console.log("🧹 Cleaning up test environment...");
 
   try {
-    // Kill any remaining marimo processes
+    // Kill marimo processes and their parent uv processes.
+    // On macOS, uv doesn't forward SIGTERM to children, so Playwright's
+    // webServer termination hangs waiting for the uv process to exit.
+    // Using SIGKILL (-9) ensures processes die immediately.
     try {
-      await execAsync("pkill -f 'marimo.*--headless' || true");
-      console.log("✅ Cleaned up marimo processes");
+      await execAsync("pkill -9 -f 'marimo.*--headless' || true");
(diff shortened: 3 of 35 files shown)

### CODE REVIEW

File: marimo/_runtime/patches.py [hunk 1]
```diff
@@ -169,15 +170,28 @@ def set_index_urls(urls):
         del glbls["sys"]
 
 
+def extract_docstring_from_header(header: str | None) -> str | None:
+    """Extract the Python docstring value from a notebook header string."""
+    if not header:
+        return None
+    try:
+        tree = ast.parse(header)
```
dmadisetti: `ast_parse` since ast can emit stdout

PR #8652, opened 2026-03-11
fix: Nesting of scriptmode runtimes

dmadisetti:
## 📝 Summary

Imported notebooks with sql cells like:

```python
@app.cell
def _():
    from mynotebook import app as myapp
    myapp.run()
```

would fail because sql used the parent context instead of the imported context. This PR leverages the context installation mechanism used by `app.embed` to isolate context globals for consistent lookup and use.


marimo/_runtime/app/script_runner.py
```diff
@@ -88,40 +88,47 @@ def _run_synchronous(
...
-                        outputs[cid] = output
-                    except MarimoRuntimeException as e:
-                        unwrapped_exception: BaseException | None = e.__cause__
-
-                        if isinstance(unwrapped_exception, MarimoStopError):
-                            self._cancel(cid)
-                        else:
-                            raise e
-                    finally:
+            with script_context.install():
```
dmadisetti, line 99: Poor git diff. Really just indented the block.

dmadisetti:
@akshayka if you have a second I'll take a sanity check on this- but I'm pretty confident this is the right approach

marimo/_runtime/context/script_context.py
```diff
@@ -53,6 +53,9 @@ def graph(self) -> DirectedGraph:
 
     @property
     def globals(self) -> dict[str, Any]:
+        if self._glbls is not None:
+            return self._glbls
+        # fallback: standalone path before _glbls is wired up
         with patch_main_module_context(
             create_main_module(
                 file=None, input_override=None, print_override=None
```
akshayka: Not sure I understand this. Is it still needed?

marimo/_runtime/context/script_context.py
```diff
@@ -141,18 +144,21 @@ def app(self) -> InternalApp:
 
 def initialize_script_context(
```
akshayka: Can you rename this function since it no longer initializes the context?

marimo/_runtime/context/script_context.py
```diff
@@ -141,18 +144,21 @@ def app(self) -> InternalApp:
...
     app: InternalApp, stream: Stream, filename: str | None
-) -> None:
-    """Initializes thread-local/session-specific context.
+) -> ScriptRuntimeContext:
+    """Creates a ScriptRuntimeContext for use with install().
```
akshayka: The docstring shouldn't be coupled to how the object should be used

```suggestion
    """Creates a ScriptRuntimeContext
```

akshayka (commented):
Should `app.run()` just raise when called from a kernel context, and recommend usage of `app.embed()` instead? What is the use case for calling `app.run()` from a running notebook? Does this close an associated open GitHub issue?

marimo/_runtime/app/script_runner.py
```diff
@@ -88,40 +88,47 @@ def _run_synchronous(
         ) as module:
             glbls = module.__dict__
             glbls.update(self._glbls)
+            script_context._glbls = glbls
```
akshayka: This does feel a bit hacky, shouldn't there be a proper API for this if it's needed?

akshayka (commented):
Following up from my previous review. I guess we can't throw if `app.run()` is called in a notebook session, since that would be a breaking change.

We don't have precedent for running a script context inside of a kernel context, and that nesting is confusing to me (even if it ends up being correct). Can you please update the PR description and documentation with more context on why this makes sense?

dmadisetti:
Moving over to draft in case it comes up again

PR #8666, opened 2026-03-12
enable storage inspector by default

Light2Dark:
Enables the experimental storage inspector for all users by default.

Users can still disable it via **Settings > Experimental > Storage Inspector**.


Light2Dark:
> nit: obstore should not be capitalized (you can see this in their docs)
> 
> Why let users disable remote storage? Are we not confident in its correctness or performance?

I think it's safe to release, but opting to be more defensive and get more feedback (since it's documented as experimental). We can remove the feat flag in subsequent release imo.

Thanks for the obstore correction

PR #8667, opened 2026-03-12
Replace synchronous exec endpoint with SSE streaming

manzt:
The `/api/kernel/execute` endpoint (formerly `/scratchpad/execute`) now streams results as server-sent events instead of returning a single JSON response. This fixes stdout/stderr being silently dropped and makes errors appear as clean plain text instead of browser-targeted HTML.

The old endpoint waited for execution to finish, then read the final state. This had a known race with the buffered console writer (which flushes every 10ms) that was patched with a 50ms sleep. Streaming eliminates the race by yielding console events as they arrive. A 50ms drain after the idle sentinel handles any trailing output from the buffer, matching the pattern already used by the MCP code path.

Two additional fixes: compile errors in `run_scratchpad` now broadcast an idle status so the listener receives the done sentinel instead of timing out after 30s. And a `plain_text_traceback` flag on `ExecuteScratchpadCommand` lets the endpoint opt into plain-text tracebacks via a context var in `write_traceback`, so API consumers see standard Python tracebacks rather than Pygments-highlighted HTML.

The SSE protocol is:
 

```
    event: stdout
    data: {"data": "hello\n"}

    event: stderr
    data: {"data": "Traceback ...\nValueError: boom\n"}

    event: done
    data: {"success": true, "output": {"mimetype": "...", "data": "..."}}
```

### DIFF
--- a/marimo/_server/scratchpad.py
+++ b/marimo/_server/scratchpad.py
@@ -5,7 +5,13 @@ from __future__ import annotations
 
 import asyncio
 import json
-from typing import TYPE_CHECKING, Any, NotRequired, TypedDict
+import sys
+from typing import TYPE_CHECKING, Any, TypedDict
+
+if sys.version_info < (3, 11):
+    from typing_extensions import NotRequired
+else:
+    from typing import NotRequired
 
 from marimo._ai._tools.types import CodeExecutionResult
 from marimo._messaging.cell_output import CellChannel


marimo/_runtime/runtime.py
```diff
@@ -2247,9 +2251,15 @@ async def handle_sync_graph(
         async def handle_execute_scratchpad(
             request: ExecuteScratchpadCommand,
         ) -> None:
-            with http_request_context(request.request):
-                await self.run_scratchpad(request.code)
-            broadcast_notification(CompletedRunNotification())
+            from marimo._messaging.tracebacks import _plain_text_traceback
+
+            token = _plain_text_traceback.set(request.plain_text_traceback)
```
mscolnick: if you look at `http_request_context` you can do a similar context manager

marimo/_runtime/commands.py
```diff
@@ -324,11 +324,14 @@ class ExecuteScratchpadCommand(Command):
...
         code: Python code to execute.
         request: HTTP request context if available.
+        plain_text_traceback: When True, tracebacks are plain text, not HTML.
     """
 
     code: str
     # incoming request, e.g. from Starlette or FastAPI
     request: Optional[HTTPRequest] = None
+    # When True, write_traceback emits plain text instead of HTML.
+    plain_text_traceback: bool = False
```
mscolnick: cc @dmadisetti @akshayka for API

manzt: What if we just inspected the incoming HTTP `Accept` header to decide what to do with the traceback?

PR #8670, opened 2026-03-12
Add internal code mode module for programmatic notebook control

manzt:
This introduces `marimo._code_mode`, an internal agent-only API that gives programmatic access to a running marimo notebook. The motivating use case is letting agents (e.g. from a scratchpad) insert, delete, replace, and reorder cells without going through the frontend UI. e.g.,

```python
import marimo._code_mode as cm

    # Install packages (queued, installed before cell ops)
    ctx.install_packages("pandas", "polars>=0.20")

    # Cell ops (appends at end by default)
    cid = ctx.create_cell("import pandas as pd")
    ctx.create_cell("df = pd.DataFrame()", after=cid)
    ctx.create_cell("setup()", before=cid, hide_code=True, disabled=True)
    
    ctx.update_cell("my_cell", code="x = 42")
    ctx.update_cell("other", hide_code=False, disabled=True)
    ctx.delete_cell("old_cell")
    ctx.move_cell("my_cell", after="other_cell")

    # Set UI element values (batched)
    ctx.set_ui_value(slider, 10)

# Dry-run compile check is on by default; disable with:
async with cm.get_context(check=False) as ctx:
    ...
```

### DIFF
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -397,7 +397,8 @@ class AsyncCodeModeContext:
     def _resolve_target(self, target: str) -> CellId_t:
         """Resolve a cell ID or name to a ``CellId_t``.
 
-        Checks the live graph first, then pending adds.
+        Checks the live graph first, then pending adds (by ID and by
+        name), then queued renames from ``edit_cell``.
         """
         # Try the live graph.
         try:
@@ -409,6 +410,16 @@ class AsyncCodeModeContext:
         cid = CellId_t(target)
         if cid in self._pending_adds:
             return cid
+
+        # Try pending adds (by name).
+        for pending_id, add_op in self._pending_adds.items():
+            if add_op.name == target:
+                return pending_id
+
+        # Try queued renames from edit_cell ops.
+        for op in self._ops:
+            if isinstance(op, _UpdateOp) and op.name == target:
+                return op.new_cell_id or op.cell_id
 
         raise KeyError(
             f"Cell {target!r} not found in notebook or pending adds"
--- a/tests/_code_mode/test_context.py
+++ b/tests/_code_mode/test_context.py
@@ -416,6 +416,47 @@ class TestSummary:
         assert "edited code of cell" in captured.out
 
 
+class TestResolveTarget:
+    async def test_create_after_pending_add_by_name(self, k: Kernel) -> None:
+        """Can reference a just-added cell by name in a subsequent add."""
+        ctx = AsyncCodeModeContext(k)
+
+        async with ctx as nb:
+            nb.create_cell("x = 1", name="first")
+            nb.create_cell("y = x + 1", after="first")
+
+        assert k.globals["x"] == 1
+        assert k.globals["y"] == 2
+
+        # "first" should come before the second cell in ordering.
+        ids_notifs = _ids_notifs(k)
+        cell_ids = ids_notifs[0].cell_ids
+        assert len(cell_ids) == 2
+
+    async def test_create_after_renamed_cell(self, k: Kernel) -> None:
+        """Can reference a cell by its new name after edit_cell renames it."""
+        await k.run(
+            [
+                ExecuteCellCommand(cell_id="0", code="a = 1"),
+                ExecuteCellCommand(cell_id="1", code="b = 2"),
+            ]
+        )
+        ctx = AsyncCodeModeContext(k)
+        _clear_messages(k)
+
+        async with ctx as nb:
+            nb.edit_cell("0", name="renamed")
+            nb.create_cell("c = a + b", after="renamed")
+
+        assert k.globals["c"] == 3
+
+        # New cell should be after "0" (renamed), before "1".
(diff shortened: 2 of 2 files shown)


marimo/_code_mode/_edits.py
```diff
@@ -0,0 +1,136 @@
...
+@dataclass(frozen=True)
+class NotebookCellData:
+    """Immutable cell descriptor (code and/or config).
+
+    Use ``.replace()`` to derive a modified copy, or ``.edit()`` to
+    produce a replace-edit for ``ctx.apply_edit()``.
+    """
+
+    code: str | None = None
+    config: CellConfig | None = None
```
mscolnick: does this need to be exported now?

marimo/_code_mode/_edits.py
```diff
@@ -0,0 +1,136 @@
...
+    draft: bool = False
+    cell_id: CellId_t = field(default_factory=lambda: CellId_t(str(uuid4())))
```
mscolnick: will claude get tripped up on `CellId_t`

marimo/_code_mode/_edits.py
manzt: This shouldn't go across API boundaries (yet). I'm experimenting with an `ApplyEditNotification`

marimo/_code_mode/_edits.py
manzt: I haven't experienced that. The skill tells claude to look at the imports for _edits.py and _context.py to see where symbols are imported from.

marimo/_code_mode/_context.py
```diff
@@ -0,0 +1,379 @@
...
+    def __init__(self, ctx: AsyncCodeModeContext) -> None:
+        self._ctx = ctx
+
+    def _cell_ids(self) -> list[CellId_t]:
+        return list(self._ctx.graph.cells.keys())
+
+    def __len__(self) -> int:
+        return len(self._ctx.graph.cells)
+
+    def __getitem__(self, index: int) -> NotebookCellData:
```
mscolnick: could make this more bullet-proof and check `index: int | str` and we lookup index, then cell-id, then cell-name

marimo/_code_mode/_edits.py
```diff
@@ -0,0 +1,136 @@
...
+    _index: int | None = field(default=None, repr=False, compare=False)
+
+    def replace(self, **kwargs: Any) -> NotebookCellData:  # noqa: ANN401
+        """Return a copy with the given fields replaced."""
+        return dataclasses.replace(self, **kwargs)
+
+    def edit(
```
dmadisetti: An aside, since I don't think we need to be fully harness engineering but I thought this blog post was interesting: https://blog.can.ac/2026/02/12/the-harness-problem/

marimo/_code_mode/_context.py
```diff
@@ -0,0 +1,392 @@
...
+    # Read-only attributes
+    # ------------------------------------------------------------------
+
+    @property
+    def graph(self) -> DirectedGraph:
+        """The notebook's dependency graph (cells, defs, refs, config)."""
+        return self._kernel.graph
+
+    @property
+    def globals(self) -> dict[str, Any]:
```
dmadisetti, line 377: Should this maybe be a copy or read only view with a comment to prevent claude mutating them?

marimo/_code_mode/_context.py
```diff
@@ -0,0 +1,392 @@
...
+        existing_id_set = set(self.graph.cells.keys())
+        # Snapshot existing code so we can detect true changes vs moves
+        existing_code = {
+            cid: self.graph.cells[cid].code for cid in existing_id_set
+        }
+        plan_ids = {e.cell_id for e in plan}
+
+        # Delete cells that are in the graph but not in the plan
+        for cid in existing_id_set - plan_ids:
+            self.graph.delete_cell(cid)
```
dmadisetti: Does apply edit require full overwrite of the notebook? I'm missing something here, we're not inserting these deleted cells back in are we?

mscolnick (dismissed):

marimo/_code_mode/_context.py
manzt: I've  adjusted this now... the models struggled with the edit API

marimo/_code_mode/_context.py
mscolnick: i think might be a good idea as well (if at least in a follow). could implement your own `frozendict`

marimo/_code_mode/_context.py
```diff
@@ -0,0 +1,626 @@
...
+        # Try pending adds (by cell ID).
+        cid = CellId_t(target)
+        if cid in self._pending_adds:
+            return cid
+
+        raise KeyError(
+            f"Cell {target!r} not found in notebook or pending adds"
+        )
+
+    def create_cell(
```
mscolnick, line 456: these functions might be better with more docs and examples (for AI)

manzt: added

PR #8682, opened 2026-03-13
fix(ui): slider editable input ignores debounce bug (#8677)

### LINKED ISSUE #8677: Slider debounce prevents value updates when using editable input
### Describe the bug

I'm not sure if this applies to other controls as well. As far as I can see, the idea of the `debounce` option is to avoid triggering expensive computations when a control is still being edited. For this reason, the value only updates once the slider has essentially been let go.

The debounce flag however seems to not take into account the `include_input` flag, which adds an editable input for a slider. When both `debounce` and `include_input` are set to `True`, the editable input is completely unusable. Neither pressing enter after inserting a number, nor using the up/down arrows triggers a cell update. Even worse, the actual value displayed by the cell is still changed.

So, for example, say that the value of the slider was 4. I edit the input to 8 and press enter. The slider now displays 8, but nothing else is run. Then you can pick up the slider, move it around, and drop it to 8. Again nothing will happen, this time because the "value" of the slider has technically not changed -- even though this time we *did* trigger a drag stop event. So now we need to drag the slider somewhere else, starting a new computation, and then re-slide it back to 8 to finally trigger the computation we wanted in the first place.

I think that return events on the editable input, as well as clicks on the up/down arrows should trigger an update, whether debounce is on or off.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.4",
  "editable": false,
  "OS": "Linux",
  "Processor": "x86_64",
  "Python Version": "3.13.1",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.11.0",
    "uv": "0.6.10"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.17.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.41.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "loro": "1.10.3",
    "pandas": "3.0.1"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```
slider = mo.ui.slider(0, 10, debounce=True, include_input=True)

# --- another cell
print(slider.value)
```

pranay0703:
When `debounce=True` and `include_input=True` are set on a slider, editing the numeric input does not trigger cell execution. This fixes the issue by forcing the NumberField's onChange to immediately setValue regardless of if the parent slider is configured with debounce, since numeric input edit events are explicitly requested by the user.

Fixes #8677

### DIFF
--- a/frontend/src/plugins/impl/__tests__/SliderPlugin.test.tsx
+++ b/frontend/src/plugins/impl/__tests__/SliderPlugin.test.tsx
@@ -3,11 +3,11 @@
 import { act, fireEvent, render } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { z } from "zod";
+import { SetupMocks } from "@/__mocks__/common";
 import { initialModeAtom } from "@/core/mode";
 import { store } from "@/core/state/jotai";
 import type { IPluginProps } from "../../types";
 import { SliderPlugin } from "../SliderPlugin";
-import { SetupMocks } from "@/__mocks__/common";
 
 SetupMocks.resizeObserver();
 


pranay0703:
I have read the contribution rules and verified that all requirements are met: frontend linting and typechecks passed, UI tests were added, and all tests run successfully.

pranay0703:
recheck

pranay0703:

pranay0703:
recheck

frontend/src/plugins/impl/__tests__/SliderPlugin.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { initialModeAtom } from "@/core/mode";
+import { store } from "@/core/state/jotai";
+import type { IPluginProps } from "../../types";
+import { SliderPlugin } from "../SliderPlugin";
+
+class ResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
```
Light2Dark: recently we have used `SetupMocks.resizeObserver()`. Can we do the same here.

Light2Dark (changes requested):

pranay0703:
test: use SetupMocks for ResizeObserver in SliderPlugin test

- Replaces the manual ResizeObserver mock with `SetupMocks.resizeObserver()` to align with repo conventions.
- Fixes minor typing issues in the createProps setup for the slider plugins test.
- Adds missing expect assertion to fix vitest/expect-expect eslint warning.

Issue #6960, opened 2025-10-28
Issues logging with loguru and marimo


wlach:
### Describe the bug

I'm seeing some strange interactions between marimo and [loguru](https://loguru.readthedocs.io/en/stable/index.html).

Two (possibly related, possibly not) problems:

1. When running in interactive mode (marimo edit ...) I don't see any logging unless stderr is flushed (e.g. via a print statement). e.g.:

<img width="962" height="897" alt="Image" src="https://github.com/user-attachments/assets/49a53607-0482-414c-9504-24dbd20686dd" />

2. When running an export (e.g. via `uv run marimo export html logging-example.py -o out.html`), somehow loguru's json serializer gets engaged (despite marimo having no interaction with loguru?) and it writes json to both the log and standard out for at least one log message:

<img width="962" height="897" alt="Image" src="https://github.com/user-attachments/assets/1d01750f-d0a5-4f75-b1c5-25dd9141c6d4" />

```
❯ uv run marimo export html logging-example.py -o out.html
This notebook has inlined package dependencies.
Run in a sandboxed venv containing this notebook's dependencies? [Y/n]: y
Running in a sandbox: /opt/homebrew/bin/uv run --isolated --no-project --compile-bytecode --with-requirements /var/folders/g_/7hlq07nj1xqgc4j9550pqrf40000gn/T/tmp829x4rjf.txt --python >=3.11 marimo export html logging-example.py -o out.html
Warning: The file 'out.html' already exists. Overwrite? [y/N]: y
Test writing to stderr.
{'record': {'elapsed': 'PT0.003586S', 'exception': None, 'extra': {}, 'file': {'name': '__marimo__cell_vblA_.py', 'path': '/var/folders/g_/7hlq07nj1xqgc4j9550pqrf40000gn/T/marimo_38236/__marimo__cell_vblA_.py'}, 'function': '<module>', 'level': {'icon': 'ℹ️', 'name': 'INFO', 'no': 20}, 'line': 2, 'message': 'Test log message 1', 'module': '__marimo__cell_vblA_', 'name': '__main__', 'process': {'id': 38236, 'name': 'Process-1'}, 'thread': {'id': 8540954944, 'name': 'MainThread'}, 'time': {}}}2025-10-28 03:56:05.046 | INFO     | __main__:<module>:1 - Test log message 2
Test writing to stderr.
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.17.2",
[...]

PR #8697, opened 2026-03-16
Fix loguru messages not being printed

Light2Dark:
## 📝 Summary

Fixes https://github.com/marimo-team/marimo/issues/6960

image

Loguru's Message class is a str subclass with __slots__ = ("record",) ([source](https://github.com/Delgan/loguru/blob/master/loguru/_handler.py#L27-L28)).

When this Message object flows through marimo's console output pipeline and reaches msgspec serialization, the custom enc_hook sees the `__slots__` attribute and serializes it as {"record": {...}} instead of the string value. 

This causes the frontend to receive a dict where it expects a string, so the console output either doesn't render or renders as a raw record dict (during export).

The fix coerces str subclasses to plain str at the Stdout.write() / Stderr.write() boundary, stripping any extra slots/attributes before the data enters the serialization pipeline.

Afaik, the record dict is internal loguru metadata that's not meant to be in the output

## 📋 Checklist

### DIFF
--- a/marimo/_messaging/types.py
+++ b/marimo/_messaging/types.py
@@ -38,14 +38,21 @@ class NoopStream(Stream):
 def _ensure_plain_str(s: str) -> str:
     """Coerce str subclasses to plain ``str``.
 
-    Some libraries (e.g. loguru) emit str subclasses whose ``__dict__``
-    carries extra metadata (loguru's ``Message.record``).  When such an
+    Some libraries (e.g. loguru) emit str subclasses whose ``__slots__``
+    carry extra metadata (loguru's ``Message.record``).  When such an
     object reaches msgspec serialization the encoder may serialize the
-    ``__dict__`` instead of the string value, corrupting console output.
+    slots/attributes instead of the string value, corrupting console output.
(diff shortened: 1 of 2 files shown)


marimo/_messaging/types.py
```diff
@@ -35,6 +35,28 @@ def write(self, data: KernelMessage) -> None:
...
+
+    Raises TypeError for non-str inputs to preserve io.TextIOBase.write()
+    semantics.
+    """
+    if type(s) is not str:
+        if not isinstance(s, str):  # pyright: ignore[reportUnnecessaryIsInstance]
+            raise TypeError(  # pyright: ignore[reportUnreachable]
+                f"write() argument must be a str, not {type(s).__name__}"
+            )
+        return str(s)
```
dmadisetti, line 56: I wonder if

```suggestion
      if type(s) is not str:
          if issubclass(type(s), str):
              return str(s)
          raise TypeError(f"write() argument must be a str, not {type(s).__name__}")
```

Would fix pyright complaints? But looks fine. Note we had a similar issue with weave wrapping: https://github.com/marimo-team/marimo/pull/8078/changes

If it comes up again, we might want to elevate this to a util

PR #8698, opened 2026-03-16
feat: support dragging range slider track to move entire range

### LINKED ISSUE #5240: drag range_slider to keep constant range
### Description

With a range_slider with x_min and x_max defined, I would like to drag the middle of the range_slider and move that range such that x_max - x_min stays the same

### Suggested solution

Provide a parameter to range_slider which allows this

### Will you submit a PR?

- [x] Yes

### Alternative

Not that I know of. Would have to do something custom

### Additional context

_No response_

Sushit-prog:
Closes #5240

## What changed

Dragging the filled track between the two handles now moves 
the entire range together, preserving the selected width. 
Individual handles still work independently. No new parameters 
— this is default behavior for all range sliders.

Frontend-only change in `range-slider.tsx`.

## How it works

On pointerdown, the drag start position and value are captured 
once. On pointermove, the delta is computed from that fixed 
anchor and applied to both handles simultaneously. Pointer 
capture ensures smooth dragging even when the cursor leaves 
the element.

## Testing
- ✅ Track drag preserves range width
- ✅ Individual handle drag works independently
- ✅ Step snapping during track drag
- ✅ Vertical orientation
- ✅ Discrete steps
- ✅ Boundary clamping at min/max

### DIFF
--- a/frontend/src/plugins/impl/RangeSliderPlugin.tsx
+++ b/frontend/src/plugins/impl/RangeSliderPlugin.tsx
@@ -80,7 +80,6 @@ const RangeSliderComponent = ({
   showValue,
   fullWidth,
   disabled,
-
   valueMap,
 }: RangeSliderProps): JSX.Element => {
   const id = useId();
@@ -123,6 +122,7 @@ const RangeSliderComponent = ({
           step={step}
           orientation={orientation}
           disabled={disabled}
+          // Triggered on all value changes
           onValueChange={(nextValue: number[]) => {
             setInternalValue(nextValue);
             if (!debounce) {
--- a/marimo/_plugins/ui/_impl/input.py
+++ b/marimo/_plugins/ui/_impl/input.py
@@ -189,7 +189,6 @@ class slider(UIElement[Numeric, Numeric]):
         debounce (bool): Whether to debounce the slider to only send the value
             on mouse-up or drag-end. Defaults to False.
         disabled (bool, optional): Whether the slider is disabled. Defaults to False.
-
         orientation (Literal["horizontal", "vertical"]): The orientation of the
             slider, either "horizontal" or "vertical". Defaults to "horizontal".
         show_value (bool): Whether to display the current value of the slider.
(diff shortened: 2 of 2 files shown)


Sushit-prog:

Light2Dark:
I'm confused, I thought we didn't want to add a new parameter

Sushit-prog:
> I'm confused, I thought we didn't want to add a new parameter

Hi @Light2Dark! Thanks for the review. I understand the confusion, based on the discussion, I can update this to make dragging the middle range bar (between the two handles) always move both handles together as default behaviour, without needing the fixed_range parameter at all. Would that be the preferred approach? Happy to update the PR!

Sushit-prog:
Updated the PR based on feedback, removed the fixed_range parameter entirely from both Python and frontend. Middle-drag on the range bar now works as default behaviour for all range sliders, implemented as a frontend-only change in range-slider.tsx. No new Python parameters added. Dragging the blue bar between the two thumbs moves both handles together while keeping the range width constant.

Sushit-prog:
The only failing backend test is test_ibis_table.py::TestTemporalColSummaries::test_time_column which appears to be a pre-existing issue with sqlglot/ibis unrelated to this PR. All other 6623 tests pass. The frontend changes are complete middle-drag on the range bar now works as default behavior with no new parameters added.

dmadisetti:
Thanks for iterating!  A few of the copilot issues look valid. Additionally I think you can undo some of the python changes. The implementation also seems a little broken right now:


https://github.com/user-attachments/assets/35fbda93-0eef-46ac-9b87-1439cf9221b9


Ping back when you have a screencast of it working! Thanks again!

Sushit-prog:
Hi @dmadisetti, fixed all the issues. Here's a summary of what was addressed: 
- Fixed the broken drag behaviour — track drag now moves both handles 
  smoothly while preserving the range width consistently
- Released pointer capture on pointerup
- Fixed stale `props.value` in `onValueCommit` using `currentDragValue` ref
- Captured track bounding rect once on pointerdown to prevent drift 
  during re-renders
- Added `stopPropagation` on pointermove to prevent Radix interference
- Converted `mergedRef` to `useCallback`
- Handled `props.steps` array for correct step snapping
- Reverted Python docstring `Notes:` section as suggested

Also updated the PR title to reflect the actual implementation.


https://github.com/user-attachments/assets/d327131e-6ded-473f-a129-c540137cab90

dmadisetti:
Super smooth! I was still able to drag when disabled though

Sushit-prog:
@dmadisetti Fixed track drag is now prevented when the 
slider is disabled.

Sushit-prog:
Done! Reverted all changes in input.py and RangeSliderPlugin.tsx 
back to main. Only range-slider.tsx is changed now.

PR #8705, opened 2026-03-16
Use explicit mulitprocessing `getcontext("spawn")`

mscolnick:
Fixes deadlock in `test_cli_export_html` and `test_cli_export_html_no_code` that causes 2-minute timeouts on Linux CI.

### Problem

PR #8625 switched these tests from subprocess to in-process `CliRunner`. The HTML export path creates a session in `SessionMode.EDIT`, which spawns a kernel via `multiprocessing.Process`. On Linux, the default start method is `fork`, but `fork()` in a multi-threaded process (pytest + CliRunner) deadlocks the child — the kernel never sends `CompletedRunNotification` and `await instantiated_event.wait()` hangs forever.

On macOS `spawn` is already the default, so this only manifests in CI.

### Fix

Use `get_context("spawn").Process(...)` instead of bare `Process(...)` in `kernel.py`, consistent with how queues are already created in `queue.py` via `get_context("spawn")`.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -33,7 +33,7 @@ jobs:
   test:
     needs: changes
     if: ${{ needs.changes.outputs.playwright == 'true' }}
-    timeout-minutes: 13 # 2024-01-18 avg: 5.0m max: 7.0m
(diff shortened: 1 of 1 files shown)


akshayka:
The server process is multi-threaded, so the bug likely could also manifest in real usage, not just CI.

Issue #8701, opened 2026-03-16
mo.ui.table column filter passes string values to decimal columns, causing narwhals InvalidOperationError on filter/search


yan-hic:
### Describe the bug

When using mo.ui.table(...) on a dataframe containing decimal columns, filtering a numeric column from the table UI can fail because the filter value is treated as a string instead of being coerced to the column’s numeric type.

In my case, filtering the Tax column for 0.00 produced:

```python
narwhals.exceptions.InvalidOperationError:
'is_in' cannot check for List(String) values in Decimal(18, 2) data
```
This appears to come from marimo’s table search/filter path.

Workaround: none really as the regular filter option (from column header) has only the condition `between` - can't be used to exclude a given value but include all other positive and negative numbers.

Btw, error in table was not forwarded to cell error, so notebook (and AI when referring @Error) doesn't detect any error.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
Replace this line with the output of marimo env. Leave the backticks in place.
```

</details>

Light2Dark:
Thanks for raising, we can fix this.

yan-hic:
Great. Obviously the same happens on negating the `is_in`.

Should I log a feature request for more conditions in addition to `between`, or do you have that in the works already?

Light2Dark:
Don't think we have any new filters in the works, please request them! :)

PR #8706, opened 2026-03-16
Convert decimals to float64 for filtering in tables

Light2Dark:
## 📝 Summary

Fixes https://github.com/marimo-team/marimo/issues/8701.

image

When the UI sends column filter values (e.g., the user selects "0.00" from the column dropdown), they arrive as strings, JSON doesn't have a decimal type. We have some logic to coerce to certain types (eg. date, datetimes, object) but we did not handle Decimals. So the fix involves:

1. Coercing filter values to Float64 for comparison. Float64 is chosen because we can't convert to Decimal and know the precision needed, thus causing errors. There is a small precision loss (documented in the test).

2. Fix Object dtype sampling in handle_filter_rows to work with LazyFrames (previously drop_nulls() inside select() silently failed)

## 📋 Checklist


yan-hic:
Thx guys for the quick turnaround

PR #8713, opened 2026-03-16
Replace pickle with msgspec for IPC serialization

manzt:
Pickle serialization across the ZeroMQ IPC boundary is fragile because any change to the command types (renaming, adding/removing fields) can silently break deserialization between the host and kernel processes. `msgspec.msgpack` gives us a proper (binary) wire format with well-defined schema evolution rules tied to the existing `Command` struct definitions.

Each Pull channel now constructs a `msgspec.msgpack.Decoder[T]` from its known message type at creation time. The receiver thread uses this decoder to deserialize incoming messages back into their correct types, leveraging the discriminated union tags already on `Command` subclasses.

### DIFF
--- a/marimo/_ipc/connection.py
+++ b/marimo/_ipc/connection.py
@@ -8,20 +8,22 @@ import queue
 import sys
 import typing
 
+import msgspec
+
 from marimo import _loggers
 from marimo._ipc.queue_proxy import PushQueue, start_receiver_thread
 from marimo._ipc.types import ConnectionInfo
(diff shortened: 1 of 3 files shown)


mscolnick (approved):

mscolnick (approved):

manzt:
I ran a stress test here: https://gist.github.com/manzt/8ba26e3aae0ecba016c7632549f73ea6

mscolnick (approved):

PR #8729, opened 2026-03-17
Make transitionCell pure by injecting parseOutline

manzt:
The marimo-lsp extension imports `transitionCell` from the frontend via `@marimo-team/frontend/unstable_internal`. Previously `cell.ts` directly imported `parseOutline` from `dom/outline.ts`, which transitively pulled in `sanitize.ts`, DOMPurify, jotai atoms, and storage utilities that all assume a browser environment. This caused `ReferenceError: document is not defined` and `window is not defined` in the marimo-lsp Node test suite.

Rather than sprinkling `typeof window` guards throughout the codebase, `parseOutline` is now an optional callback on `transitionCell`. The frontend passes it in at the single call site in `cells.ts`, while marimo-lsp callers get a pure function with no browser dependencies.

\

### DIFF
--- a/frontend/src/core/cells/cell.ts
+++ b/frontend/src/core/cells/cell.ts
@@ -3,20 +3,15 @@ import { logNever } from "@/utils/assertNever";
 import { invariant } from "@/utils/invariant";
 import { type Seconds, Time } from "@/utils/time";
 import { extractAllTracebackInfo, type TracebackInfo } from "@/utils/traceback";
+import { parseOutline } from "../dom/outline";
 import type { CellMessage, OutputMessage } from "../kernel/messages";
 import type { RuntimeState } from "../network/types";
 import { collapseConsoleOutputs } from "./collapseConsoleOutputs";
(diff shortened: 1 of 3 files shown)


Light2Dark (commented):
I think there's a similar issue if you import `cells/session.ts`. It looks like `getOutline` solves this with 

```ts
function getOutline(html: string): Outline | null {
  // Some JS runtimes (e.g. Node.js for the VSCode extension) do not
  // expose DOMParser globally. In those environments parsing isn't
  // possible (without a polyfill), so return just return `null`.
  if (typeof DOMParser === "undefined") {
    return null;
  }
```

can we do the same in DOMPurify (in `core/sanitize.ts`)
```ts
if (typeof document !== "undefined") {
  DOMPurify.addHook("beforeSanitizeAttributes", (node) => { ... });
  DOMPurify.addHook("afterSanitizeAttributes", (node) => { ... });
}
```

manzt:
Sorry for the delay here. Your suggestion is much cleaner :)

PR #8748, opened 2026-03-18
Re-derive language adapter when cell code changes externally

manzt:
When `updateEditorCodeFromPython` receives new code (e.g., from a `code_mode` `edit_cell` call), it blindly uses the current language adapter to transform the incoming code. If a markdown cell is edited to contain plain Python (or vice versa), the adapter doesn't switch and the UI renders the cell with the wrong language mode.

This adds a check that compares the current adapter against what `languageAdapterFromCode` detects for the incoming Python code. When they differ, `switchLanguage` is called with `keepCodeAsIs: true` to update the adapter, metadata, and editor extensions before transforming the code.

### DIFF
--- a/frontend/src/core/codemirror/language/utils.ts
+++ b/frontend/src/core/codemirror/language/utils.ts
@@ -3,7 +3,11 @@
 import type { EditorState } from "@codemirror/state";
 import type { EditorView } from "@codemirror/view";
 import { replaceEditorContent } from "../replace-editor-content";
-import { languageAdapterState } from "./extension";
+import {
+  languageAdapterFromCode,
+  languageAdapterState,
+  switchLanguage,
+} from "./extension";
 import { languageMetadataField } from "./metadata";
 
 /**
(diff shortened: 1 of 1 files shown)


frontend/src/core/codemirror/language/utils.ts
mscolnick, line 38: this is used in a fair amount of places. there is a particular case when Codemirror doesn't like when you try to update the extensions/editor while its mounting. im not sure if that would be hit, but worth testing this a bit

mscolnick (approved):

frontend/src/core/codemirror/language/utils.ts
manzt: Hmm, do you think something like first transform with the current adapter then wait an animation frame / timeout to re-transform?


```ts
if (correctAdapter.type !== currentAdapter.type) {
  requestAnimationFrame(() => {
    switchLanguage(edit, { /* ... */ });
    // re-transform
    const newAdapter = editor.state.field(languageAdapterState);
    const [code] = newAdapter.transformIn(pythonCode);
    replaceEditorContent(editor, code);
  })
  // transform with current
  const [code] = currentAdapter.transformIn(pythonCode);
  replaceEditorContent(editor, code);
  return code;
} 
```

mscolnick: yea i'm not sure when to call this logic. where you seeing bugs with what i was mentioning? if not, i think without `requestAnimationFrame` is fine.

PR #8776, opened 2026-03-19
fix: use python -m pip instead of pip --python for PipPackageManager

KJyang-0114:
## Summary

On macOS with pip-installed Python (e.g. python.org installer), only 'pip3' is
available in PATH, not 'pip'. This caused `PipPackageManager.is_manager_installed()`
to return False because `shutil.which('pip')` could not find it, making marimo
fail to detect any optional dependencies.

## Fix

1. **`is_manager_installed()`**: Add fallback check via `python -m pip --version`
   when `pip` is not in PATH.

2. **All pip commands**: Use `python -m pip` instead of `pip --python python_exe`.
   The `-m pip` invocation is the [officially supported way](https://pip.pypa.io/en/stable/user_guide/)
   and works reliably across all platforms.

## Files Changed

- `marimo/_runtime/packages/pypi_package_manager.py`
  - Added `DependencyManager` import
  - Override `is_manager_installed()` with `python -m pip` fallback
  - `install_command()`, `uninstall()`, `list_packages()` now use `python -m pip`
- `tests/_runtime/packages/test_pypi_package_manager.py`
  - Updated test assertions to match new command format

## Testing

- All 70 tests in `test_pypi_package_manager.py` pass
- `ruff check` passes

## CLA

I have read the CLA D

### DIFF
--- a/marimo/_runtime/packages/pypi_package_manager.py
+++ b/marimo/_runtime/packages/pypi_package_manager.py
@@ -126,12 +126,13 @@ class PipPackageManager(PypiPackageManager):
         """Check if pip is available.
 
         On some platforms (e.g. macOS with pip-installed Python from python.org),
-        only `pip3` is available in PATH, not `pip`. We fall back to checking
-        via `python -m pip` to handle these cases.
+        only `pip3` is available in PATH, not `pip`. We first try the method we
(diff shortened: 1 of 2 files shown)


KJyang-0114:
Closing as duplicate of #8779 which contains the same pip fix with additional context. The pip fix (python -m pip) and tests are identical.

Issue #8687, opened 2026-03-14
LSP diagnostics (Ruff) persist after fixing code


daizutabi:
### Describe the bug

LSP diagnostics (such as Ruff's "Undefined name") are not cleared and remain visible in the cell editor even after the code is fixed (e.g., changing `a` to `a = 1`).

This has been observed on Windows (Native, WSL2, devcontainer). I am not sure if this issue persists on other platforms.

| 1. Enter `a` (Error appears) | 2. Change to `a = 1` (Error remains) |
| :--- | :--- |
| <img width="200" alt="Image" src="https://github.com/user-attachments/assets/9b9363de-bc8b-446d-a454-b75532701244" /> | <img width="200" alt="Image" src="https://github.com/user-attachments/assets/c84f9122-2a1b-49a9-8487-a3b88c19851c" /> |

I experimented with the code in `frontend/src/core/codemirror/lsp/notebook-lsp.ts` and found that the `publishDiagnostics` notification for clearing errors was missing the `version` field.

After I manually added the `version` field to the "clear" loop in the `patchProcessNotification` method, the diagnostics started clearing correctly.

However, I am not certain if this is the correct way to fix the issue.

(Update: This fix also resolved persistent diagnostics in basedpyright.)

```typescript
// frontend/src/core/codemirror/lsp/notebook-lsp.ts

// ... inside patchProcessNotification's clear loop
for (const cellDocumentUri of cellsToClear) {
  previousProcessNotification({
    method: "textDocument/publishDiagnostics",
    params: {
      uri: cellDocumentUri,
      version: cellVersion, // <--- Adding this line fixed the issue
      diagnostics: [],
    },
  });
}
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.4",
  "editable": true,
  "location": "/workspaces/marimo/marimo",
  "OS": "Linux",
  "OS Version": "6.6.87.2-microsoft-standard-WSL2",
  "Processor": "x86_64",
  "Python Version": "3.12.13",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.22.0",
    "uv": "0.6.17"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.17.0",
    "packaging": "26.0",
[...]

PR #8390, opened 2026-02-20
Fix Ty LSP startup and stale diagnostics in notebook editor

axsseldz:
## 📝 Summary

This PR fixes Ty LSP reliability when enabling it from Settings and fixes stale/incorrect diagnostics that could remain after code was corrected. It also adds a regression test to protect the Ty startup path.

Fixes [#8687](https://github.com/marimo-team/marimo/issues/8687).

## 🔍 Root Cause

- `save_user_config` startup logic was effectively tied to Copilot-style behavior, so Ty enablement did not always trigger LSP startup in time.
- WebSocket connection attempts to `/lsp/ty` could happen before the Ty process was fully ready to accept connections.
- Frontend config state updates could be overwritten by stale async responses.
- Diagnostics forwarding in notebook LSP could accept stale updates and not always clear old messages reliably.

Observed sandbox errors (`marimo edit --sandbox repro_ty.py`):

```text
[E 260220 00:07:57 middleware:496] Failed to connect to ws://localhost:3122/lsp/ty after 3 attempts. Final error: Multiple exceptions: [Errno 61] Connect call failed ('::1', 3122, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 3122)
[E 260220 00:07:57 middleware:575] WebSocket proxy error for ws://localhost:3122/lsp/ty: Multiple exceptions: [Errno 61] Connect call failed ('::1', 3122, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 3122)
[E 260220 00:07:57 middleware:580] LSP server appears to be down at ws://localhost:3122/lsp/ty. Check if the LSP server started successfully.
[E 260220 00:07:57 middleware:415] Error proxying websocket: Cannot call "send" once a close message has been sent.
```

## 🔍 Description of Changes

- Updated config-save LSP startup condition to use generic LSP-enabled detection (`any_lsp_server_running`) instead of Copilot-only logic.
- Improved backend LSP startup robustness by waiting for server readiness after process spawn.
- Improved proxy websocket behavior for LSP routes:
  - retry tuning,
  - startup attempt on LSP connection failures,
  - safe close handling to avoid noisy close-after-close errors.
- Updated frontend config merge behavior to apply only dirty values after save, avoiding stale overwrite issues.
- Updated notebook diagnostics handling:
  - drop stale diagnostics by version,
  - forward diagnostics with strictly increasing local versions,
  - ensure clear notifications carry version consistently.
- Added regression test:
[...]

PR #8780, opened 2026-03-19
fix(lsp): add version field when clearing diagnostics

KJyang-0114:
## Summary
LSP diagnostics (e.g. Ruff "Undefined name") are not cleared from the cell editor after fixing the code. The issue occurs because the `textDocument/publishDiagnostics` notification for clearing diagnostics is missing the `version` field.

According to the LSP spec, the client ignores `publishDiagnostics` notifications where the `version` field does not match the current document version. The clear-path code was not including this field, so the diagnostics persisted.

## Fix
Added `version: cellVersion` to the `publishDiagnostics` params in the cells-to-clear loop inside `patchProcessNotification`:

```diff
 for (const cellDocumentUri of cellsToClear) {
   previousProcessNotification({
     method: "textDocument/publishDiagnostics",
     params: {
       uri: cellDocumentUri,
+      version: cellVersion,
       diagnostics: [],
     },
   });
 }
```

This mirrors what the add/update path already does at line ~820.

## Testing
- Verified the fix locally on Windows + WSL2 (issue reporter's environment)
- Fix also resolved persistent diagnostics in basedpyright
- The fix is minimal (1 line) and only affects the clear path

## Risk
- **Low**: Single-line addition, only affects the diagnostics-clear path
- **Rollback**: Revert this single line

## Issue
Fixes marimo-team/marimo#8687

### DIFF
--- a/frontend/src/core/codemirror/lsp/notebook-lsp.ts
+++ b/frontend/src/core/codemirror/lsp/notebook-lsp.ts
@@ -832,6 +832,7 @@ export class NotebookLanguageServerClient implements ILanguageServerClient {
               method: "textDocument/publishDiagnostics",
               params: {
                 uri: cellDocumentUri,
+                version: cellVersion,
                 diagnostics: [],
               },
             });


KJyang-0114:

KJyang-0114:
recheck

mscolnick:
@KJyang-0114 thanks for the contribution. this might have already been fixed by https://github.com/marimo-team/marimo/pull/8390

Light2Dark:
closing this since likely fixed

PR #8784, opened 2026-03-20
fix: clear error state on disabled-transitively cells when ancestor recovers

### LINKED ISSUE #8072: Disabled cells' ancestor errors cannot be cleared
### Describe the bug

If a disabled cell's ancestor has an error, that disabled cell will also show as having an error. Fixing the original error doesn't remove the error state from the disabled cell.

Image

(cell-7 is just fine now.)

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.19.4",
  "editable": false,
  "location": "/opt/venv/lib/python3.14/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "5.15.146.1-microsoft-standard-WSL2",
  "Processor": "x86_64",
  "Python Version": "3.14.2",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10",
    "narwhals": "2.15.0",
    "packaging": "25.0",
    "psutil": "7.2.1",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.51.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "duckdb": "1.4.3",
    "nbformat": "5.10.4",
    "openai": "2.15.0",
    "pandas": "2.3.3",
    "polars": "1.37.1",
    "pyarrow": "22.0.0",
    "python-lsp-ruff": "2.3.0",
    "ruff": "0.14.13",
    "sqlglot": "28.6.0"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

VishakBaddur:
Fixes #8072

## Root Cause

When a disabled-transitively cell's ancestor had an error and then recovered, the disabled cell permanently showed the ancestor's error state.

`run_stale_cells()` in `runtime.py` only re-queues non-disabled cells:
```python
if cell_impl.stale and not self.graph.is_disabled(cid):
    cells_to_run.add(cid)
```

So disabled-transitively cells never got re-queued and never had a chance to reset their `run_result_status` from `"exception"` to `"disabled"`.

## Fix

- Added `is_any_ancestor_errored()` to `DirectedGraph`
- In `run_stale_cells()`, after building `cells_to_run`, reset `run_result_status` to `"disabled"` for any disabled-transitively cell whose ancestor no longer has an error

## Testing

Added `test_is_any_ancestor_errored` to `tests/_runtime/test_dataflow.py` verifying the new graph method correctly detects and clears ancestor error states.

### DIFF
--- a/marimo/_smoke_tests/disabled_ancestor_error.py
+++ b/marimo/_smoke_tests/disabled_ancestor_error.py
@@ -5,7 +5,7 @@
 
 import marimo
 
-__generated_with = "0.21.1"
+__generated_with = "0.23.1"
 app = marimo.App()
 
 
--- a/tests/_cli/test_sandbox.py
+++ b/tests/_cli/test_sandbox.py
@@ -133,7 +133,7 @@ def test_construct_uv_cmd_marimo_edit_empty_file() -> None:
         additional_deps=[],
     )
     assert "--refresh" in uv_cmd
-    assert uv_cmd[0].endswith("uv")
+    assert os.path.basename(uv_cmd[0]).startswith("uv")
     assert uv_cmd[1] == "run"
 
 
@@ -148,7 +148,7 @@ def test_construct_uv_cmd_marimo_edit_file_no_sandbox(
         additional_deps=[],
     )
     assert "--refresh" in uv_cmd
-    assert uv_cmd[0].endswith("uv")
+    assert os.path.basename(uv_cmd[0]).startswith("uv")
     assert uv_cmd[1] == "run"
 
 
@@ -164,7 +164,7 @@ def test_construct_uv_cmd_marimo_edit_sandboxed_file(
         additional_deps=[],
     )
     assert "--refresh" not in uv_cmd
-    assert uv_cmd[0].endswith("uv")
+    assert os.path.basename(uv_cmd[0]).startswith("uv")
     assert uv_cmd[1] == "run"
 
 


dmadisetti:
Hi @VishakBaddur are you still interested in contributing this? I'm not actually seeing this change reflected in the smoke test I just pushed.

image

If you want to update with a screenshot once you get this working that would be great.

VishakBaddur:
Hi @dmadisetti , thanks for the smoke test and for pushing this forward! I'll investigate why the error state isn't clearing in the UI. Looks like I need to also emit a frontend notification after resetting the run_result_status. Will update the PR shortly.

VishakBaddur:
Hi @dmadisetti , thanks for the smoke test, it helped pinpoint the exact gap. I've pushed a fix that addresses both the backend and frontend sides:

Root cause (two parts):
Backend: run_stale_cells() was resetting run_result_status but never emitting any frontend notifications, so the UI had no signal to update.
Frontend: transitionCell() in cell.ts did not reset the errored flag when receiving "disabled-transitively" status, leaving the red has-error border even after the output was cleared.

Fix:
Call cell_impl.set_runtime_state("disabled-transitively") to keep backend state consistent and broadcast status to the frontend
Call CellNotificationUtils.broadcast_empty_output(cell_id, status="disabled-transitively") to replace the stale error output in the UI
Reset nextCell.errored = false in the "disabled-transitively" case in transitionCell() — this is safe because in all other code paths where "disabled-transitively" is sent, errored is already false, and if an error message follows (as in mutate_graph), it will correctly re-set errored = true
Broadened is_any_ancestor_errored to include "marimo-error" in addition to "exception"

dmadisetti:
@VishakBaddur this still doesn't work as expected

image

Can you update with a screenshot when you're ready?

VishakBaddur:
Hi @dmadisetti , the fix is now working end-to-end. Here's what changed:
The previous approach only handled the backend state but missed two critical issues:

Wrong code path: The clear-stale logic was in run_stale_cells(), but the UI element toggle triggers set_ui_element_value() → _run_cells() directly. So the fix never fired during the actual bug reproduction.

Frontend never received a clear signal: Even when the backend reset run_result_status, no frontend notification was sent to clear the stale error output and errored/stopped flags.

The actual fix:

Backend: Snapshot disabled cells in error/cancelled state before running. After _run_cells() completes, broadcast empty output + correct status for any cell whose ancestor has now recovered. This lives in _run_cells() so it covers all code paths.
Frontend: Clear stopped flag in disabled-transitively case; clear stopped/errored in idle case when non-error output arrives.


Screenshot 2026-04-04 at 10 24 31 AM

Screenshot 2026-04-04 at 10 25 04 AM

marimo/_runtime/runtime.py
```diff
@@ -1402,12 +1402,39 @@ async def _run_cells(self, cell_ids: set[CellId_t]) -> None:
...
                 while cell_ids := await self._run_cells_internal(cell_ids):
                     LOGGER.debug("Running state updates ...")
                     if self.lazy() and cell_ids:
                         self.graph.set_stale(cell_ids, prune_imports=True)
                         break
                 LOGGER.debug("Finished run.")
+                # Clear stale error state from disabled cells whose ancestor
+                # recovered. Uses pre-run snapshot since run_result_status is
+                # updated during the run.
+                for _cid in _pre_run_errored_disabled:
```
dmadisetti: nit: these shouldn't be underscore prefixed (implies they are unused)

dmadisetti:
@VishakBaddur great job! Much cleaner implementation. We still get the red highlights:

image

But I don't think that's blocking. I think this is a great starting point if you'd like to get it in as is. Tiny code clean up comment, but let me know if you're happy with this

VishakBaddur:
Hi @dmadisetti , addressed the nit, removed the underscore prefixes. Happy to merge as-is if you are!

PR #8787, opened 2026-03-20
feat(plotly): add reactive bar selection support

axsseldz:
## 📝 Summary

Fix Plotly bar chart selection behavior so selections are accurate, and reactive for both click and box/lasso interactions.

## 🔍 Description of Changes

- Fixed duplicate bar payload entries when Plotly already sends explicit selected bar points and marimo also applies backend fallback extraction.
- Updated backend bar selection filtering to require overlap on both axes, so selecting empty space above bars no longer selects bars that merely share the same x-axis position.
- Added support for single-click selection on bar traces in the frontend Plotly plugin, so clicking a bar now returns the selected bar value reactively.
- Added Python regression tests covering:
  - empty-space box selections not selecting bars
  - explicit Plotly bar points not being duplicated
  - existing bar selection behavior remaining intact
- Added a frontend unit test verifying that clicking a bar updates Plotly selection state correctly.

## 📋 Checklist

CC: @mscolnick @nojaf


frontend/src/plugins/impl/plotly/PlotlyPlugin.tsx
```diff
@@ -221,7 +222,7 @@ export const PlotlyComponent = memo(
             selections:
               "selections" in evt ? (evt.selections as unknown[]) : [],
             points: extractPoints(evt.points),
-            indices: evt.points.map((point) => point.pointIndex),
+            indices: evt.points.map(getPointIndex).filter(isDefined),
```
nojaf: Bit of a nitpick but you are looping the collection twice here and could do this in one go.

A simple for loop and populate a new array can easily avoid this.

marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -1409,58 +1423,129 @@ def _extract_bars_fallback(
...
-            # Check if bar position is within selection range
             pos_in_range = False
 
             if _is_orderable_value(pos_val) and _is_orderable_value(pos_min):
-                # Orderable axis (numeric or datetime): direct comparison
-                # Parse datetime bounds (frontend sends ISO strings)
                 pos_min_p = _parse_datetime_bound(pos_min)
                 pos_max_p = _parse_datetime_bound(pos_max)
-                # TODO: Consider bar width in future implementations
-                # Currently treating bars as points at their position value
```
manzt, line 1455: I notice a bunch of comments have been removed in this PR. Are these TODO's fixed or is this intentional?

marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -1409,58 +1423,129 @@ def _extract_bars_fallback(
...
+    base = _get_indexed_value(getattr(trace, "base", None), point_idx)
+    if base is None:
+        base = 0
+
+    lower = min(base, point_value)
+    upper = max(base, point_value)
+    return not (selection_max < lower or selection_min > upper)
+
+
+def _build_bar_point(
```
manzt, line 1931: This function shares some similar logic with `_build_histogram_sample_point`. Could we try to consolidate the shared extraction logic for things like metadat?

marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -1409,58 +1423,129 @@ def _extract_bars_fallback(
...
+        return (int(curve_number), int(point_index))
+    return None
+
+
+def _bar_value_in_selection_range(
+    trace: Any,
+    point_idx: int,
+    point_value: Any,
+    selection_min: Any,
+    selection_max: Any,
```
manzt: Can we avoid the use of `Any` here? This logic assumes numbers but what if it is a categorical or string this will just blow up.

manzt: If the inputs can truly be `Any` then we should have some guard that checks they are actually numbers (falling back with a log warning "Cannot compare ..., treating as out of range")

frontend/src/plugins/impl/plotly/PlotlyPlugin.tsx
axsseldz: all good, thanks for the comment!, i replaced the map(...).filter(...) path with a single pass helper so we only walk the points once

marimo/_plugins/ui/_impl/plotly.py
axsseldz: Oh i see, i just pulled the shared per point metadata extraction into a small helper and reused it from both _build_bar_point and _build_histogram_sample_point

marimo/_plugins/ui/_impl/plotly.py
axsseldz: yep, updated this, i removed the any based comparison path, narrowed the inputs, and added a guard so non-numeric values return False instead of blowing up

marimo/_plugins/ui/_impl/plotly.py
axsseldz: oh yeah my bad, should be bc i iterated multiple times in this file, added them back 👍

Issue #8676, opened 2026-03-13
marimo.persistent_cache default save location


KusmiTeaBBBDetox:
### Describe the bug

# `marimo.persistent_cache` function is not consistent with documentation

```python
@mo.persistent_cache()
def get_cached_df():
    [...]
    return df
```

When I run the notebook like this:

```bash
marimo edit .\reports\folder\notebook.py
```

the cache is created in:

```
.\__marimo__\cache
```

instead of:

```
.\reports\folder\__marimo__\cache
```

This behavior is inconsistent with the documentation. Since in the example above the caching is not done in the directory of the notebook file

![Screenshot](https://github.com/user-attachments/assets/2904cc74-ae69-4fa1-ab23-5c0b5a3cf23d)

Documentation reference:  
https://docs.marimo.io/api/caching/#marimo.lru_cache--context-manager-for-lru-caching-the-return-value-of-a-block-of-code

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.4",
  "editable": false,
  "location": "C:/Users/---/anaconda3/envs/spotiflowEnv/lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 140 Stepping 1, GenuineIntel",
  "Python Version": "3.10.19",
  "Locale": "--",
  "Binaries": {
    "Browser": "144.0.7559.133",
    "Node": "--",
    "uv": "0.9.29 (1f1321d84 2026-02-03)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10",
    "narwhals": "2.15.0",
    "packaging": "25.0",
    "psutil": "7.2.1",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.20",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "loro": "1.10.3",
    "openai": "2.18.0",
    "pandas": "2.3.3",
    "polars": "1.37.1"
  },
  "Experimental Flags": {}
}
```

</details>

[...]

PR #8797, opened 2026-03-20
fix: follow pycache_prefix for __marimo__ location

### LINKED ISSUE #7173: Run on read-only filesystem (Place ./__marimo__ somewhere else)
### Description

When running on a read-only filesystem marimo will log errors as it's trying to create `./__marimo__/session`

```
[E 251114 08:57:31 serialize:392] Failed to create parent directories: [Errno 30] Read-only file system: '/app/notebooks/example/__marimo__'
[E 251114 08:57:31 background_task:55] Error in task: [Errno 30] Read-only file system: '/app/notebooks/example/__marimo__
```

Similar issues have been brought up before about cache, but that's possible to re-locate somewhere else with the experimental.cache config. It would be nice if the whole `__marimo__` could be moved somewhere else

### Suggested solution

Config option to specify `__marimo__` path

dmadisetti:
## 📝 Summary

fixes #7173 #8676

The `__marimo__` path was previously immutable. This PR introduces the convention of following `sys.pycache_prefix`.

Additionally, this PR fixes the cache location for notebooks in nested directories. Note, this will break cache, but that's fine because next release contains a different breaking PR #8793

There may be some opengraph dump location I'm missing @peter-gy

### DIFF
--- a/marimo/_metadata/opengraph.py
+++ b/marimo/_metadata/opengraph.py
@@ -14,7 +14,7 @@ from urllib.parse import urlparse
 import msgspec
 
 from marimo import _loggers
-from marimo._utils.paths import MARIMO_DIR_NAME, get_marimo_dir
+from marimo._utils.paths import MARIMO_DIR_NAME, notebook_output_dir
 from marimo._utils.scripts import read_pyproject_from_script
 
 LOGGER = _loggers.marimo_logger()
@@ -160,8 +160,11 @@ def derive_title_from_path(filepath: str) -> str:
 def default_opengraph_image(filepath: str) -> str:
     """Return the default relative image path for a given notebook."""
     stem = Path(filepath).stem
-    return (
-        f"{MARIMO_DIR_NAME}/assets/{stem}/{DEFAULT_OPENGRAPH_IMAGE_FILENAME}"
+    return str(
+        Path(MARIMO_DIR_NAME)
+        / "assets"
+        / stem
+        / DEFAULT_OPENGRAPH_IMAGE_FILENAME
     )
 
 
@@ -169,7 +172,7 @@ def default_opengraph_image_abs(filepath: str) -> Path:
     """Return the absolute image path for a given notebook."""
     stem = Path(filepath).stem
     return (
-        get_marimo_dir(filepath)
+        notebook_output_dir(filepath)
         / "assets"
         / stem
         / DEFAULT_OPENGRAPH_IMAGE_FILENAME
--- a/marimo/_save/stores/file.py
+++ b/marimo/_save/stores/file.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 from marimo._runtime.runtime import notebook_dir
 from marimo._save.stores.store import Store
-from marimo._utils.paths import get_marimo_dir
+from marimo._utils.paths import notebook_output_dir
 
 
 def _valid_path(path: Path) -> bool:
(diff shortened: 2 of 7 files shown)


marimo/_save/stores/file.py
```diff
@@ -14,14 +15,22 @@ def _valid_path(path: Path) -> bool:
...
+    @property
+    def save_path(self) -> Path:
+        if self._resolved_save_path is None:
+            self._resolved_save_path = self._default_save_path()
+        return self._resolved_save_path
+
     def _default_save_path(self) -> Path:
         if (root := notebook_dir()) is not None:
-            return Path(root / "__marimo__" / "cache")
+            return get_marimo_dir(root) / "cache"
```
mscolnick: we have something already called `get_marimo_dir` for the package. maybe we can call that one differently `get_marimo_whl_dir` or `get_marimo_installation_dir`?

marimo/_metadata/opengraph.py
```diff
@@ -160,12 +160,24 @@ def derive_title_from_path(filepath: str) -> str:
 def default_opengraph_image(filepath: str) -> str:
     """Return the default relative image path for a given notebook."""
     stem = Path(filepath).stem
-    return f"__marimo__/assets/{stem}/{DEFAULT_OPENGRAPH_IMAGE_FILENAME}"
+    return (
+        f"{MARIMO_DIR_NAME}/assets/{stem}/{DEFAULT_OPENGRAPH_IMAGE_FILENAME}"
```
mscolnick: should the paths be windows compat?

marimo/_utils/paths.py
```diff
@@ -51,6 +52,49 @@ def maybe_make_dirs(filepath: Path) -> None:
...
+def get_marimo_dir(notebook_path: Path | str | None) -> Path:
+    """Compute the __marimo__ output directory for a given notebook.
+
+    When ``sys.pycache_prefix`` is set and the notebook path is absolute, the
+    directory tree is mirrored under the prefix (similar to how Python mirrors
+    ``__pycache__`` directories).  Otherwise the ``__marimo__`` directory is
+    placed next to the notebook file.
+
+    Resolution order:
+        1. ``sys.pycache_prefix`` (if set and path is absolute): mirror tree
```
mscolnick, line 67: how often is `sys.pycache_prefix` set? will this break existing notebooks given `<notebook_parent>/__marimo__` is always the expectation.

dmadisetti: Unsure, it was unset in the 4-5 envs I just tried (including molab).
I can add some docs so this feels less magic

marimo/_save/stores/file.py
dmadisetti: Went with renaming this to marimo_output

marimo/_metadata/opengraph.py
```diff
@@ -160,12 +160,27 @@ def derive_title_from_path(filepath: str) -> str:
 def default_opengraph_image(filepath: str) -> str:
```
mscolnick, line 160: @peter-gy is this used in the file-system or a URL? should this be posix or detect the OS file separator?

peter-gy: It is a notebook-relative internal path that we store in OpenGraph metadata and later resolve on disk when serving the thumbnail. I'd keep this posix and use `default_opengraph_image_abs` introduced by this pr for actual fs operations.

peter-gy:
@dmadisetti I don't see any missing opengraph dump. in #8097 the main related parts I introduced were (1) default thumbnail write path in `thumbnail.py`, (2) default-image existence check in `opengraph.py` and (3) serving images under `__marimo__/` from `/og/thumbnail`. Your PR covers all three.

mscolnick:
We can do a breaking release next.

PR #8804, opened 2026-03-20
Discourage casual cell naming in code mode

manzt:
Perhaps more a personal preference thing... but agents were naming every cell they created, adding visual clutter. The docstring now frames `name` as a human-facing label for special cases.


marimo/_code_mode/_context.py
```diff
@@ -550,10 +545,6 @@ def create_cell(
             # Create and run
             cid = ctx.create_cell("x = 1")
```
manzt, line 551: The skill mentions the ability to name cells in the improving notebooks reference.

PR #8811, opened 2026-03-21
feat: use variable name as download filename in dataframe viewer

### LINKED ISSUE #8095: dataframe download from rich dataframe viewer automatically names download file by variable name

Sushit-prog:
## Summary

Fixes the download filename regression where files were being 
named `df.csv` instead of the actual variable name.

## Root Cause

The previous fix used `infer_variable_name()` which relies on 
identity checks (`is`) that fail when dataframes are 
wrapped/transformed, causing it to fall back to `df`.

## Fix

Use `ctx.ui_element_registry.bound_names(self._id)` to look up 
the actual variable name at runtime via a shared `get_bound_name()` 
utility function.

## Changes
- `utils/dataframe.py` — added `get_bound_name()` shared utility, 
  `_create_download_virtual_file()` helper and `filename` parameter 
  to `download_as()`
- `dataframes/dataframe.py` — uses `get_bound_name()` to get variable 
  name at download time
- `table.py` — same as above
- `test_dataframe.py` — added test verifying correct filename

Fixes #8095

### DIFF
--- a/marimo/_plugins/ui/_impl/dataframes/dataframe.py
+++ b/marimo/_plugins/ui/_impl/dataframes/dataframe.py
@@ -13,7 +13,6 @@ from typing import (
     cast,
 )
 
-from marimo import _loggers
 from marimo._messaging.mimetypes import KnownMimeType
 from marimo._output.hypertext import is_non_interactive
 from marimo._output.rich_help import mddoc
@@ -56,8 +55,6 @@ from marimo._utils.methods import getcallable
 from marimo._utils.narwhals_utils import is_narwhals_lazyframe, make_lazy
 from marimo._utils.parse_dataclass import parse_raw
 from marimo._utils.variable_name import infer_variable_name
-
-LOGGER = _loggers.marimo_logger()
 
 TOO_MANY_ROWS = 100_000
 
@@ -196,10 +193,6 @@ class dataframe(UIElement[dict[str, Any], DataFrameType]):
             label="",
             args={
                 "columns": self._get_column_types(),
-                # dataframe-name is used by the frontend for display purposes.
-                # The download filename is resolved at download time in _download_as()
-                # using get_bound_name(), which correctly handles wrapped/transformed
-                # dataframes where infer_variable_name() would fall back to 'df'.
                 "dataframe-name": dataframe_name,
                 "total": rows,
                 "page-size": page_size,
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -72,7 +72,6 @@ from marimo._utils.narwhals_utils import (
     can_narwhalify_lazyframe,
     unwrap_narwhals_dataframe,
 )
-from marimo._utils.variable_name import infer_variable_name
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
@@ -513,10 +512,6 @@ class table(
         self._max_columns: Optional[int] = None
         max_columns_arg: Union[int, str]
 
-        # Infer the variable name before add_selection_column() mutates data,
-        # so the identity check still matches the caller's original variable.
-        download_file_name = infer_variable_name(data, "download")
-
         has_stable_row_id = False
         if selection is not None:
             data, has_stable_row_id = add_selection_column(data)
@@ -749,7 +744,6 @@ class table(
                 "max-height": int(max_height)
                 if max_height is not None
                 else None,
-                "download-file-name": download_file_name,
             },
             on_change=on_change,
             functions=(
@@ -1128,10 +1122,13 @@ class table(
                 data_url = mo_data.arrow(table_manager.to_arrow_ipc()).url
                 return data_url, "arrow"
             except NotImplementedError:
-                LOGGER.debug("Arrow export not implemented, falling back.")
+                LOGGER.debug(
+                    "Arrow export not implemented, falling back to CSV."
+                )
             except Exception as e:
                 LOGGER.error("Unexpected error exporting Arrow: %s", e)
 
+        # Try CSV
         try:
             data_url = mo_data.csv(table_manager.to_csv({})).url
             return data_url, "csv"
@@ -1140,6 +1137,7 @@ class table(
         except Exception as e:
             LOGGER.error("Unexpected error exporting CSV: %s", e)
 
+        # Fallback to JSON
         try:
             data_url = mo_data.json(
                 table_manager.to_json({}, ensure_ascii=True)
--- a/marimo/_plugins/ui/_impl/utils/dataframe.py
+++ b/marimo/_plugins/ui/_impl/utils/dataframe.py
@@ -22,7 +22,7 @@ LOGGER = _loggers.marimo_logger()
 DEFAULT_CSV_ENCODING = "utf-8"
 
 
-def get_bound_name(element_id: str) -> str | None:
+def get_bound_name(element_id: UIElementId) -> str | None:
     """Get the bound variable name for a UI element.
 
     Looks up the element's bound names from the UI element registry
@@ -37,12 +37,9 @@ def get_bound_name(element_id: str) -> str | None:
     """
     try:
         ctx = get_context()
-        bound = sorted(
-            ctx.ui_element_registry.bound_names(UIElementId(element_id))
-        )
+        bound = sorted(ctx.ui_element_registry.bound_names(element_id))
         return bound[0] if bound else None
-    except Exception:
-        LOGGER.debug("Error getting bound names for download filename")
+    except ContextNotInitializedError:
         return None
 
 
@@ -106,7 +103,7 @@ def download_as(
         ValueError: If unrecognized format.
 
     Returns:
-        tuple: (url, user-facing filename without extension) for the downloaded file.
+        tuple: (url, user-facing filename with extension) for the downloaded file.
     """
     if drop_marimo_index:
         # Remove the selection column if exists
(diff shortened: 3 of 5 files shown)


Sushit-prog:
@Light2Dark @dmadisetti ready for review!

Here's a screenshot showing the fix working — downloaded file 
is now named after the variable (some_frame.csv) instead of df.csv.

Screenshot 2026-03-22 022235


Could you add the appropriate label so all CI checks can run?

Sushit-prog:
@Light2Dark  @dmadisetti  gentle ping, all checks are passing 
and the fix is working as shown in the screenshot. 
Happy to make any changes if needed!

marimo/_plugins/ui/_impl/utils/dataframe.py
```diff
@@ -58,7 +88,8 @@ def download_as(
     csv_encoding: str | None = None,
     csv_separator: str | None = None,
     json_ensure_ascii: bool = True,
-) -> str:
+    filename: str | None = None,
+) -> tuple[str, str]:
```
dmadisetti, line 84: Opposed to a tuple can we share with `DownloadAsResponse`? Maybe new name and location.

marimo/_plugins/ui/_impl/utils/dataframe.py
```diff
@@ -73,35 +104,58 @@ def download_as(
...
             else get_default_csv_encoding()
         )
-        return mo_data.csv(
+        vfile = mo_data.csv(
             manager.to_csv(encoding=encoding, separator=csv_separator)
-        ).url
+        )
+        return (vfile.url, vfile.filename)
     elif ext == "json":
-        # Use strict JSON to ensure compliance with JSON spec
```
Light2Dark, line 98: many of these types of comments are being removed, we should avoid that.

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -340,23 +344,34 @@ def _download_as(self, args: DownloadAsArgs) -> str:
...
+        bound_filename: str | None = None
+        try:
+            from marimo._runtime.context import get_context
+
+            ctx = get_context()
+            bound = sorted(ctx.ui_element_registry.bound_names(self._id))
+            if bound:
+                bound_filename = bound[0]
+        except Exception:
+            LOGGER.debug("Error getting bound names for download filename")
```
Light2Dark: should this be a shared util function?

marimo/_plugins/ui/_impl/utils/dataframe.py
```diff
@@ -73,35 +104,58 @@ def download_as(
...
+            data = manager.to_json(
+                encoding=None, ensure_ascii=json_ensure_ascii, strict_json=True
+            )
+        elif ext == "parquet":
+            data = manager.to_parquet()
+        else:
+            raise ValueError(
+                "format must be one of 'csv', 'json', or 'parquet'."
+            )
+        return _create_download_virtual_file(data, ext, filename)
```
Light2Dark: Sorry, just to check my understanding, why do we need this? And the custom function?

Sushit-prog:
@Light2Dark addressed all review comments:
1. Restored removed comments
2.Extracted bound_names lookup into shared get_bound_name() utility
3. Used in both dataframe.py and table.py

The _create_download_virtual_file function is needed because 
mo_data.csv/json/parquet() generate random filenames internally 
via VirtualFileLifecycleItem with no way to pass a custom name. 
Happy to simplify further if you have a preferred approach!

Sushit-prog:
@Light2Dark the Test BE / macos-latest failure (test_terminal_ws) 
appears to be a pre-existing flaky test unrelated to this PR — 
it fails due to a WebSocket file descriptor issue in terminal.py 
which is unrelated to the dataframe download changes. 

All dataframe and table tests pass locally. Happy to investigate 
further if needed!

Also, the Test BE / ubuntu-latest / Py 3.14 failure is 
the test_terminal_ws flaky test — unrelated to this PR's 
dataframe download changes. Is this a known issue on Python 3.14?

Light2Dark:
Yeah, I don't think those are related. Sorry about that, sometimes tests are flakey 😅

marimo/_plugins/ui/_impl/utils/dataframe.py
```diff
@@ -73,35 +138,59 @@ def download_as(
...
     Raises:
         ValueError: If unrecognized format.
         NotImplementedError: If the table format is not supported.
 
     Returns:
-        str: The URL to download the table data.
+        tuple: (url, user-facing filename) for the downloaded file.
     """
     if drop_marimo_index:
-        # Remove the selection column if exists
```
Light2Dark, line 85: don't remove this

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1116,13 +1122,27 @@ def _to_chart_data_url(
...
+
+        if is_simple_list:
+            try:
+                data_url = mo_data.json(
+                    table_manager.to_json({}, ensure_ascii=True)
+                ).url
+                return data_url, "json"
+            except Exception as e:
+                LOGGER.error("Failed to export table data as JSON: %s", e)
+                raise
```
Light2Dark: what is this for?

tests/_plugins/ui/_impl/dataframes/test_dataframe.py
```diff
@@ -345,8 +344,7 @@ def test_dataframe_download_encoding_utf8_sig() -> None:
             download_csv_encoding="utf-8-sig",
         )
 
-        # CSV should include BOM
```
Light2Dark, line 348: don't remove comments like these

tests/_plugins/ui/_impl/test_table.py
```diff
@@ -1359,8 +1359,7 @@ def test_download_as_ignores_cell_selection() -> None:
     table._search(SearchTableArgs(query="2", page_size=10, page_number=0))
     # Make a cell selection; download should still include the filtered view
     table._convert_value([{"rowId": "0", "columnName": "a"}])
-    # Use JSON format to avoid optional dependencies
```
Light2Dark, line 1362: here too

marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -189,6 +196,9 @@ def __init__(
             label="",
             args={
                 "columns": self._get_column_types(),
+                # dataframe-name is set at init time using infer_variable_name.
+                # For downloads, the actual filename comes from bound_names at
+                # download time via _download_as returning DownloadAsResponse.
```
Light2Dark: doesn't this mean we can use the dataframe-name for download?

Sushit-prog:
@Light2Dark @dmadisetti addressed all remaining review comments:

1. **Simplified `download_as()`** :removed the complex `_create_download_virtual_file()` 
helper and the duplicate `if filename is not None` branch entirely. Now always uses 
`mo_data.csv/json/parquet()` for the URL and simply returns the bound variable name 
as the filename. The frontend's `downloadByURL(url, filename)` already handles them 
as separate arguments so no custom virtual file needed.

2. **Answers Light2Dark's question** : "doesn't this mean we can use dataframe-name 
for download?" — `dataframe-name` is a static init-time arg that can't be updated 
after render, but we don't need to. `DownloadAsResponse` returns the correct filename 
directly from `get_bound_name()` at download time, and the frontend already uses it.

3. **Restored all deleted comments** : `# CSV should include BOM`, 
`# Use JSON format to avoid optional dependencies`, 
`# Remove the selection column if exists`

4. **Strengthened test** : now properly mocks `mo_data.csv` and explicitly asserts 
`result.filename == "my_dataframe"` and `result.url` comes from mo_data.

5. **`is_simple_list` block** : this is in `_to_chart_data_url` (chart rendering), 
not in `_download_as`. It was not added by this PR.

All backend tests pass locally. CI frontend build confirmed passing.

Sushit-prog:
@Light2Dark  @dmadisetti  addressed all remaining review comments:

1. **Simplified `download_as()`** : removed `_create_download_virtual_file()` 
and the duplicate `if filename is not None` branch. Now always uses 
`mo_data.csv/json/parquet()` for the URL and returns the bound variable name 
as filename directly. No custom virtual file needed since `downloadByURL(url, filename)` 
handles them as separate arguments.

2. **Removed `is_simple_list` block** from `_to_chart_data_url` — this was 
incorrectly added and was breaking existing tests.

3. **Restored all deleted comments** : `# CSV should include BOM`, 
`# Use JSON format to avoid optional dependencies`, 
`# Remove the selection column if exists`

4. **Clarified comment** in `dataframe.py` explaining why `dataframe-name` is 
init-time only and how `get_bound_name()` correctly resolves the filename at 
download time.

5. **Strengthened test**:  properly mocks `mo_data.csv` and explicitly asserts 
`result.filename == "my_dataframe"`.

All 15 dataframe tests and 21 table tests pass locally.

PR #8813, opened 2026-03-21
fix: render data-tooltip in portal to prevent clipping in overflow containers

### LINKED ISSUE #3237: tooltip clipped inside container

abhiyadav2345:
## Summary

Fixes #3237 — tooltips created with `data-tooltip` were clipped by parent containers using `overflow:hidden` (e.g. grid layout cells).

**Root cause:** The previous implementation used CSS `::before`/`::after` pseudo-elements with `position: absolute`, which are constrained by the nearest positioned ancestor and clipped by any ancestor with `overflow: hidden`.

**Fix:** Replace the CSS pseudo-element approach with a React `Tooltip` component transform in `RenderHTML.tsx`, following the existing `wrapDocHoverTargets` pattern already in the codebase. The Radix UI `Tooltip` renders its content via `TooltipPortal` (browser top layer), so it is never clipped by any ancestor container.

## Changes

- **`frontend/src/plugins/core/RenderHTML.tsx`** — Add `wrapTooltipTargets` transform that wraps any `[data-tooltip]` element in the existing `Tooltip` component (with `usePortal=true` by default)
- **`frontend/src/css/md-tooltip.css`** — Remove the `::before`/`::after` CSS rules (now handled by React); keep the `cursor: pointer` and `text-decoration: underline dotted` visual hint

## Before / After

**Before:** Tooltip clipped by container with `overflow: hidden`

**After:** Tooltip renders in portal (top layer), visible regardless of ancestor overflow settings

## Notes

As suggested in the issue by @mscolnick — using the portal/top-layer approach. Implemented via the existing Radix UI `Tooltip` (which already uses `TooltipPortal`) rather than raw Popover API, to stay consistent with the codebase patterns.

### DIFF
--- a/frontend/src/plugins/core/__test__/RenderHTML.test.ts
+++ b/frontend/src/plugins/core/__test__/RenderHTML.test.ts
@@ -223,8 +223,7 @@ describe("wrapTooltipTargets", () => {
   });
 
   test("data-tooltip on nested element wraps only that element", () => {
-    const html =
-      '<p>Outer <span data-tooltip="tip">inner</span> text</p>';
+    const html = '<p>Outer <span data-tooltip="tip">inner</span> text</p>';
     expect(parseHtml({ html })).toMatchInlineSnapshot(`
       <p>
         Outer 


abhiyadav2345:

abhiyadav2345:
Added 3 tests for `wrapTooltipTargets` in `RenderHTML.test.ts` (commit 8451dad):

- **basic wrap** — `data-tooltip` span gets wrapped in `<Tooltip content="...">...`
- **no wrap** — element without `data-tooltip` is left as-is
- **nested** — only the inner element with `data-tooltip` gets wrapped; outer element is unaffected

All 17 tests passing.

PR #8817, opened 2026-03-22
`NotebookDocument` model and backend integration

manzt:
Today, notebook structure is scattered across the front end and backend. These diverge silently (now especially with `marimo._code_mode`), and cause inconsistency bugs.

This PR introduces `NotebookDocument`, an event-sourced ordered list of cells that both the frontend and kernel will read from and write to:

```python
@dataclass(slots=True)
class NotebookCell:
    id: CellId_t
    code: str
    name: str = ""
    config: CellConfig = field(default_factory=CellConfig)
```

Seven event types mutate it:

| Event | Fields | Semantics |
| --- | --- | --- |
| `CellCreated` | `id, code, name, config, after` | New cell inserted after `after` (or appended) |
| `CellDeleted` | `id` | Cell removed |
| `CellMoved` | `id, after` | Single cell repositioned |
| `CellsReordered` | `cell_ids` | Full ordering replaced (bulk moves) |
| `CodeChanged` | `id, code` | Cell code updated |
| `NameChanged` | `id, name` | Cell renamed |
| `ConfigChanged` | `id, config` | Cell config changed |

The document is populated from the `CellManager` at session creation and "snapshotted" onto `ExecuteScratchpadCommand` so the kernel process has access.

### DIFF
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -162,10 +162,10 @@ class _CellsView:
 
     def _build_at(self, cell_id: CellId_t) -> NotebookCellData:
         cell_impl = self._ctx.graph.cells[cell_id]
-        meta = self._ctx._kernel.cell_metadata.get(cell_id)
+        doc_cell = self._ctx._document.get(cell_id)
         return NotebookCellData(
             code=cell_impl.code,
-            config=meta.config if meta else CellConfig(),
+            config=doc_cell.config if doc_cell else CellConfig(),
             name=self._cell_name(cell_id),
             id=cell_id,
         )
@@ -691,8 +691,8 @@ class AsyncCodeModeContext:
         config: CellConfig | None = None
         if hide_code is not None or disabled is not None or column is not None:
             # Start from existing config and override provided fields.
-            meta = self._kernel.cell_metadata.get(cell_id)
-            existing = meta.config if meta else CellConfig()
+            doc_cell = self._document.get(cell_id)
+            existing = doc_cell.config if doc_cell else CellConfig()
             config = CellConfig(
                 hide_code=hide_code
                 if hide_code is not None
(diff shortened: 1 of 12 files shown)


marimo/_code_mode/_context.py
```diff
@@ -89,11 +96,6 @@ class NotebookCellData:
 LOGGER = _loggers.marimo_logger()
 
 
-# Module-level store for cell names set via code_mode.
-# Persists across context manager invocations within the same kernel.
-_cell_names: dict[CellId_t, str] = {}
-
```
manzt, line 95: This was a hack.

marimo/_session/state/document.py
```diff
@@ -0,0 +1,295 @@
...
+
+    from marimo._ast.cell_manager import CellManager
+
+
+# ------------------------------------------------------------------
+# Document cell
+# ------------------------------------------------------------------
+
+
+class DocumentCell(msgspec.Struct):
```
mscolnick: what about `NotebookCell`?

marimo/_session/session.py
```diff
@@ -231,6 +232,9 @@ def __init__(
             ttl_seconds if ttl_seconds is not None else _DEFAULT_TTL_SECONDS
         )
         self.session_view = SessionView()
+        self.session_view.document = NotebookDocument.from_cell_manager(
+            app_file_manager.app.cell_manager
```
mscolnick, line 236: would love explore to possibility and effort to removing/consolidating `cell_manager`

marimo/_session/state/document.py
manzt: Good call. Changed

marimo/_session/session.py
manzt: I have this noted in a set of follow up PRs.

PR #8818, opened 2026-03-22
Frontend → session document event sync

manzt:
The document model from #8817 is populated at startup but doesn't yet reflect user actions in the browser. This PR closes that gap with a single endpoint and a frontend middleware.

A reducer middleware intercepts actions that modify cells (creates, deletes, moves, code edits, renames) and queues typed document events. Events are debounced (400ms) and flushed as a single batch to `POST /api/document/events`. The events use msgspec's tagged union for discriminated deserialization so one endpoint handles all event types:

```
POST /api/document/events
{"events": [
  {"type": "cells-reordered", "cell_ids": ["c", "a", "b"]},
  {"type": "code-changed", "id": "c", "code": "x = 42"}
]}
```

After this PR the session's `NotebookDocument` stays in sync with what the user sees, so when code_mode later snapshots it for the kernel, the ordering and code reflect the user's latest edits and drag-and-drop moves.

### DIFF
--- a/frontend/src/core/cells/document-events-middleware.ts
+++ b/frontend/src/core/cells/document-events-middleware.ts
@@ -10,12 +10,12 @@
  */
 
 import { debounce } from "lodash-es";
-import type { CellId } from "./ids";
-import type { NotebookState } from "./cells";
 import { kioskModeAtom } from "../mode";
 import { getRequestClient } from "../network/requests";
 import type { DocumentEventsRequest } from "../network/types";
 import { store } from "../state/jotai";
+import type { NotebookState } from "./cells";
+import type { CellId } from "./ids";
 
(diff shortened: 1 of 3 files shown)


frontend/src/core/cells/document-events-middleware.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * Middleware for the notebook reducer. Intercepts actions that modify
+ * document structure and queues document events.
+ */
+export function documentEventsMiddleware(
+  _prevState: NotebookState,
+  newState: NotebookState,
+  action: { type: string; payload: unknown },
```
manzt, line 48: Code smell. This should be strictly typed.

PR #8825, opened 2026-03-23
Fix asyncio tasks starved by blocking control loop

### LINKED ISSUE #8820: Asyncio tasks are very slow
### Describe the bug

I am trying to run `rendercanvas` and `jupyter_rfb` in the Marimo notebook. They render images which are streamed to the front-end, while user events (like mouse clicks and movement) are send back. While this works fine in Jupyter or VSCode notebooks, running interactive applications close to 100 FPS, in Marimo it is super slow.

In trying to debug I ran into another issue: [MO-5572](https://linear.app/marimo/issue/MO-5572/stdout-in-async-task-is-ignored). In the minimal proof of concept for that issue I incidentally  exposed the performance problem. I'll just repeat the code here:

```py
mport marimo

__generated_with = "0.20.4"


@app.cell
def _():
    import time
    import asyncio

    async def foo():
        et = time.time() + 1
        count = 0
        while time.time() < et:
            count += 1
            await asyncio.sleep(0.01)
        print("done", count)

    return asyncio, foo


@app.cell
def _(asyncio, foo):
    return (x,)


if __name__ == "__main__":
    app.run()
```

In the terminal:

```
done 4
```

So, while running `asyncio.sleep(0.01)` for 1 second takes ±90 iterations, when it is run as a separate task it gets only 4 iterations.

Any idea why this happens and/or how this can be fixed?

cc @manzt

manzt:
Fixes #8820

The kernel's control loop called `queue.get(timeout=0.1)` directly, blocking the asyncio event loop for up to 100ms per iteration. Any background tasks created by user code via `create_task` or `ensure_future` could only run in the brief gap between blocking calls, resulting in ~4 iterations/sec instead of ~90 for a task doing `await asyncio.sleep(0.01)`. This blocking also starved `loop.call_soon` callbacks and other event loop work scheduled outside cell execution.

The fix offloads the blocking `queue.get` to a thread via `run_in_executor`, keeping the event loop free to service background tasks while waiting for messages. The WASM/Pyodide path is unaffected since it already uses a native `asyncio.Queue`.

@almarklein with`main` of jupyter_rfb:

### DIFF
--- a/tests/_runtime/test_control_loop.py
+++ b/tests/_runtime/test_control_loop.py
@@ -16,41 +16,6 @@ import pytest
 
 
 @pytest.mark.asyncio
-async def test_blocking_queue_get_starves_background_tasks():
-    """Blocking queue.get() prevents background tasks from running.
-
-    This demonstrates the bug fixed by switching to run_in_executor:
-    a blocking get(timeout=0.1) holds the event loop for ~100ms per
-    iteration, so a background task sleeping for 10ms only wakes
-    a handful of times per second.
-    """
-    q: queue.Queue[None] = queue.Queue()
-    count = 0
-
-    async def background():
-        nonlocal count
-        end = time.time() + 0.5
-        while time.time() < end:
-            count += 1
-            await asyncio.sleep(0.01)
-
-    task = asyncio.create_task(background())
-
-    # Simulate the OLD control loop: blocking get()
-    end = time.time() + 0.5
-    while time.time() < end:
-        try:
-            q.get(timeout=0.1)
-        except queue.Empty:
-            await asyncio.sleep(0)
-
-    await task
-    # With blocking get(timeout=0.1), the task is starved:
-    # ~5 iterations per 0.5s instead of ~45.
-    assert count < 15
-
-
-@pytest.mark.asyncio
 async def test_executor_queue_get_does_not_starve_background_tasks():
     """run_in_executor keeps the event loop free for background tasks.
 


manzt:
```python
import numpy as np
import jupyter_rfb
import asyncio
import time

class PerfTest(jupyter_rfb.RemoteFrameBuffer):
    i = 0
    n = 0

    def get_frame(self):
        if self.i >= self.n:
            return None
        h, w = 100, 300
        array = np.zeros((h, w, 3), np.uint8)
        # Rainbow progress bar
        progress = (self.i + 1) / self.n
        bar_w = int(w * progress)
        for x in range(bar_w):
            hue = x / w
            # HSV-to-RGB (hue only, full saturation/value)
            r, g, b = [
                int(c * 255)
                for c in __import__("colorsys").hsv_to_rgb(hue, 1.0, 1.0)
            ]
            array[10 : h - 10, x, :] = [r, g, b]
        self.i += 1
        self.request_draw()
        return array

    def run(self, n=50):
        self.i = 0
        self.n = n
        self.reset_stats()
        self.request_draw()


w = PerfTest(css_height="120px")
w
```

```python
w.run(50)
```

almarklein:
Thanks for the quick response! I tried this branch locally, and can confirm it works fast again! 🚀

marimo/_runtime/runtime.py
```diff
@@ -3570,19 +3570,19 @@ def _enqueue_control_request(req: CommandMessage) -> None:
     async def control_loop(kernel: Kernel) -> None:
         from queue import Empty
 
+        loop = asyncio.get_running_loop()
+        TIMEOUT_S = 0.1
```
almarklein: The timeout can now be higher or even removed;  When it expires on the timeout, it immediately enters again.

dmadisetti (dismissed):
timeout could be removed, but I think this is fine

PR #8827, opened 2026-03-23
Prefer tomllib over tomlkit for reading TOML

manzt:
`tomlkit` is only needed for writing TOML (it preserves formatting), but `read_toml` was using it for reads too. For a single call this doesn't matter much, but we have a lot of tests that create a kernel fixture which calls:

```
get_store() → get_config() → read_pyproject_marimo_config()
```

`tomlkit` is pure Python and ~65x slower than `tomllib` (C extension, stdlib since 3.11). Across ~960 kernel fixtures that adds up fast.

```sh
hyperfine \
  --warmup 1 \
  --runs 5 \
  -n 'before (tomlkit)' \
    'git stash -q && uv run --group test pytest tests/_runtime/test_runtime.py -q --timeout=120 -x --no-header -p no:sugar 2>/dev/null; git stash pop -q' \
  -n 'after (tomllib)' \
    'uv run --group test pytest tests/_runtime/test_runtime.py -q --timeout=120 -x --no-header -p no:sugar 2>/dev/null'
# Benchmark 1: before (tomlkit)
#   Time (mean ± σ):     24.444 s ±  0.370 s    [User: 20.753 s, System: 3.308 s]
#   Range (min … max):   24.063 s … 24.969 s    5 runs
# 
# Benchmark 2: after (tomllib)
#   Time (mean ± σ):      5.397 s ±  0.027 s    [User: 4.263 s, System: 0.815 s]
#   Range (min … max):    5.362 s …  5.425 s    5 runs
# 
# Summary
#   after (tomllib) ran
#     4.53 ± 0.07 times faster than before (tomlkit)
```

### DIFF
--- a/tests/_config/test_config_packages.py
+++ b/tests/_config/test_config_packages.py
@@ -27,7 +27,7 @@ def mock_cwd(tmp_path: Path):
 def test_infer_package_manager_from_pyproject():
     # Test poetry detection
     with patch(
-        "marimo._config.packages.read_toml",
+        "marimo._config.packages.toml_reader.read",
         return_value={"tool": {"poetry": {}}},
     ):
         assert (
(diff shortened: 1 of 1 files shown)


marimo/_utils/toml.py
```diff
@@ -1,54 +1,53 @@
...
         import tomllib
 
         return tomllib.loads(s)
+    else:
+        import tomlkit
+
+        return tomlkit.loads(s).unwrap()
 
 
 def is_toml_error(e: Exception) -> bool:
```
mscolnick: we could rename to `is_toml_read_error` so it doesn't get used on `is_toml_write_error` (which would fail this check). there is no issue currently

manzt: Good callout. I made https://github.com/marimo-team/marimo/pull/8835 to try to consolidate this coupling.

PR #8828, opened 2026-03-23
Skip readiness polling in mocked LSP server tests

manzt:
Three LSP tests call `server.start()` with a mock subprocess that never listens on a port, so `_wait_until_ready` burns its full 5s timeout each time. These tests check lifecycle behavior, not readiness so we can patch it out.

15.9 s → 0.66 s.

```sh
hyperfine \
  --warmup 1 --runs 5 \
  -n 'before' 'git stash -q && uv run --group test pytest tests/_server/test_lsp.py -q --timeout=120 -x --no-header -p no:sugar 2>/dev/null; git stash pop -q' \
  -n 'after' 'uv run --group test pytest tests/_server/test_lsp.py -q --timeout=120 -x --no-header -p no:sugar 2>/dev/null'

# Benchmark 1: before
#   Time (mean ± σ):     15.876 s ±  0.052 s    [User: 0.703 s, System: 0.193 s]
#   Range (min … max):   15.818 s … 15.924 s    5 runs
#
# Benchmark 2: after
#   Time (mean ± σ):     661.7 ms ±   0.9 ms    [User: 559.0 ms, System: 92.9 ms]
#   Range (min … max):   660.6 ms … 663.1 ms    5 runs
#
# Summary
#   after ran
#    23.99 ± 0.09 times faster than before
```

### DIFF
--- a/tests/_server/test_lsp.py
+++ b/tests/_server/test_lsp.py
@@ -84,9 +84,7 @@ async def test_base_lsp_server_start_stop(
 
     # Skip readiness polling: the mock process doesn't listen on a TCP
     # port, so _wait_until_ready would poll for its full 5 s timeout.
-    with mock.patch.object(
-        server, "_wait_until_ready", return_value=True
-    ):
+    with mock.patch.object(server, "_wait_until_ready", return_value=True):
         alert = await server.start()
     assert alert is None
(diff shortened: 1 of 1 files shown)


mscolnick:
nice catch! would love to find a way to catch these sooner too

PR #8832, opened 2026-03-23
fix: keep markdown preview in sync on local edits

peter-gy:
## Summary

Fixes the below-illustrated markdown preview sync bug where the preview could miss the last typed character or stay stale after local edits.

https://github.com/user-attachments/assets/194628bc-2591-4fcf-bd04-fcc4e80a7aac

## Details

- Kept the local markdown autorun fix in the frontend editor path.
- Marked Loro's initial reconciliation dispatch as RTC sync, the same way later sync updates are marked.
- Markdown autorun now:
  - runs for local edit transactions (`input`, `delete`, `undo`, `redo`)
  - ignores formatting-only changes
  - ignores RTC sync transactions, including init sync
- Testing:
  - markdown autorun gating
  - RTC init-sync annotation

### DIFF
--- a/frontend/src/core/codemirror/cells/extensions.ts
+++ b/frontend/src/core/codemirror/cells/extensions.ts
@@ -373,8 +373,9 @@ function shouldAutorunMarkdownUpdate({
       return false;
     }
 
-    // Prefer explicit local edit transactions, but fall back to focused local
-    // doc changes for paths that don't annotate a user event.
+    // Prefer explicit local edit transactions, but keep a focused fallback for
+    // local rewrite paths like split-cell, which can update markdown content
+    // without annotating a user event.
     return (
       MARKDOWN_AUTORUN_USER_EVENTS.some((kind) => tr.isUserEvent(kind)) ||
       hasFocus


frontend/src/core/codemirror/cells/extensions.ts
```diff
@@ -330,6 +336,52 @@ function cellCodeEditing(hotkeys: HotkeyProvider): Extension[] {
...
+    // Ignore RTC sync changes to avoid duplicate runs from remote edits.
+    if (tr.annotation(loroSyncAnnotation) !== undefined) {
+      return false;
+    }
+
+    // Prefer explicit local edit transactions, but fall back to focused local
+    // doc changes for paths that don't annotate a user event.
+    return (
+      MARKDOWN_AUTORUN_USER_EVENTS.some((kind) => tr.isUserEvent(kind)) ||
+      hasFocus
```
mscolnick, line 381: What scenario requires the `hasFocus` fallback? Could you add a brief comment?

peter-gy: Just pushed some clarification in the comment. Essentially, when a focused markdown cell is rewritten by a local action like split-cell, that edit does not carry a `userEvent`, so `isUserEvent(...)` alone would skip markdown autorun. In that case, `hasFocus` is the fallback signal that this is still happening in the actively edited local cell, so we still autorun.

mscolnick: that makes total sense, thanks for clarifying

PR #8836, opened 2026-03-23
feat: display startup tips in CLI

peter-gy:
## Summary

Adds marimo tips to the CLI startup banner shown by server-starting commands.

## Details

For this initial impl I kept things minimal on purpose:

- tips are defined statically
- tip selection is random on each startup without any CLI state persisted
- tips are TTY-only so that we don1t pollute non-interactive caller logs
- `--quiet` suppresses tips
- malformed tip commands fail open instead of breaking startup logging
- if the filter removes every tip, we fall back to the full pool rather than showing nothing

For the relevance logic we derive a small startup context from the current invocation and use it to avoid obviously redundant tips. For example, if I run:

`marimo edit --headless marimo/_tutorials/intro.py --watch`

we should not suggest:

`marimo edit notebook.py --watch`

## Demo

Command-backed tips render as a short summary plus an indented action line. Link-backed tips render the same way.

### Tip with command

Screenshot 2026-03-23 at 16 46 23

### Tip with link

Screenshot 2026-03-23 at 16 47 42

### DIFF
--- a/marimo/_cli/tips.py
+++ b/marimo/_cli/tips.py
@@ -36,7 +36,7 @@ CLI_STARTUP_TIPS: Final[tuple[CliTip, ...]] = (
         command="marimo tutorial intro",
     ),
     CliTip(
-        text="Convert a Jupyter notebook",
+        text="Convert a Jupyter notebook to a marimo notebook",
         command="marimo convert notebook.ipynb -o notebook.py",
     ),
     CliTip(


marimo/_cli/tips.py
```diff
@@ -0,0 +1,196 @@
...
+    CliTip(
+        text="Run a notebook as a web app",
+        command="marimo run notebook.py",
+    ),
+    CliTip(
+        text="Watch for external edits and auto-reload",
+        command="marimo edit notebook.py --watch",
+    ),
+    CliTip(
+        text="Run a notebook in an isolated environment",
```
akshayka:
```suggestion
        text="Run a notebook in an isolated virtual environment",
```

marimo/_cli/tips.py
```diff
@@ -0,0 +1,196 @@
...
+    link: str | None = None
+
+
+CLI_STARTUP_TIPS: Final[tuple[CliTip, ...]] = (
+    CliTip(
+        text="Run a notebook as a web app",
+        command="marimo run notebook.py",
+    ),
+    CliTip(
+        text="Watch for external edits and auto-reload",
```
akshayka:
```suggestion
        text="Reload the editor when the notebook file is edited externally",
```

marimo/_cli/tips.py
```diff
@@ -0,0 +1,196 @@
...
+    CliTip(
+        text="Run a notebook in an isolated environment",
+        command="marimo edit --sandbox notebook.py",
+    ),
+    CliTip(
+        text="Open the intro tutorial",
+        command="marimo tutorial intro",
+    ),
+    CliTip(
+        text="Convert a Jupyter notebook",
```
akshayka:
```suggestion
        text="Convert a Jupyter notebook to a marimo notebook",
```

PR #8842, opened 2026-03-24
Add new `NotebookDocument` model

manzt:
This PR introduces the core data model, with wiring to session (#8843), frontend (#8844), and kernel (#8845).

`NotebookDocument` is an ordered list of `NotebookCell` entries that applies `Transaction`s atomically. A transaction is a tuple of typed ops (`CreateCell`, `DeleteCell`, `MoveCell`, `ReorderCells`, `SetCode`, `SetName`, `SetConfig`) with a `source` tag identifying the writer (e.g. `"frontend"`, `"kernel"`, `"file-watch"`). 

The document validates ops for conflicts (e.g. delete + update on the same cell) before applying. Version is `None` on creation and stamped by `apply()`.

```python
doc = NotebookDocument([NotebookCell(id="a", code="x = 1", ...)])
tx = Transaction(ops=(SetCode(cell_id="a", code="x = 2"),), source="kernel")
applied = doc.apply(tx)  # stamps version → Transaction(version=1)
```

PR #8843, opened 2026-03-24
[nd 2/5] Wire `NotebookDocument` into the session

manzt:
With #8842, this PR gives it a home. `session.document` is populated from the `CellManager` at startup and added to the `Session` protocol so all session code can reach it.

The key piece is the interception in `session.notify()`. Kernel notifications arrive as serialized bytes via the stream distributor, so we peek at the JSON tag to identify document transactions, deserialize them, apply to `session.document` (which stamps the version), and re-broadcast the versioned result. Server-side callers (like the file watcher) pass typed objects and hit the same `apply()` path.


marimo/_session/session.py
```diff
@@ -388,13 +419,45 @@ def notify(
         from_consumer_id: Optional[ConsumerId],
     ) -> None:
         """Write an operation to the session consumer and the session view."""
+        # Intercept document transactions: apply to session.document
+        # (stamps version), then re-serialize for broadcast.
+        # Kernel notifications arrive as bytes via the stream distributor;
+        # server-side callers (e.g. file-watch) pass typed objects.
+        if isinstance(operation, bytes) and (
+            operation.find(b'"notebook-document-transaction"') >= 0
```
mscolnick: is this fragile if this is inside some text corpus?

marimo/_session/session.py
```diff
@@ -388,13 +419,45 @@ def notify(
...
+        notif: NotebookDocumentTransactionNotification,
+    ) -> NotebookDocumentTransactionNotification:
+        """Apply to session.document, return with version stamped."""
+        try:
+            applied = self.document.apply(notif.transaction)
+            return NotebookDocumentTransactionNotification(
+                transaction=applied,
+            )
+        except (ValueError, KeyError):
+            LOGGER.warning("Failed to apply document transaction")
```
mscolnick, line 465: could we print the error here

marimo/_session/session.py
```diff
@@ -388,13 +419,45 @@ def notify(
...
+        ):
+            try:
+                operation = self._apply_document_transaction(
+                    msgspec.json.decode(
+                        operation,
+                        type=NotebookDocumentTransactionNotification,
+                    )
+                )
+            except Exception:
+                LOGGER.warning("Failed to decode document transaction")
```
mscolnick: could we print the error here too

marimo/_session/session.py
```diff
@@ -71,6 +76,29 @@
...
+    """
+    return NotebookDocument(
+        [
+            NotebookCell(
+                id=cd.cell_id,
+                code=cd.code,
+                name=cd.name,
+                config=cd.config,
+            )
+            for cd in cell_manager.cell_data()
```
dmadisetti, line 100: `cell_manager.build_document` per discussion?

PR #8844, opened 2026-03-24
[nd 3/5] Frontend transaction sync

manzt:
Prev #8842, #8843. This closes the frontend ↔ session loop in three steps:

**Frontend → session.** A reducer middleware intercepts cell actions (create, delete, move, rename, set-config) as the user performs them, batches them into transaction ops, debounces (400ms), and flushes to `POST /api/document/transaction`. A suppress guard prevents echo loops when applying server-originated ops.

**Session → frontend.** `applyTransactionOps` maps each op to the corresponding cell reducer action.

**Queued-cell staleness.** When the kernel queues a cell for execution, the frontend now snapshots `lastCodeRun`. Previously only the user-initiated "Run" button did this, so kernel-initiated runs (e.g. via code_mode) left cells visually stale even after successful execution.

### DIFF
--- a/marimo/_code_mode/__init__.py
+++ b/marimo/_code_mode/__init__.py
@@ -34,12 +34,12 @@ from __future__ import annotations
 
 from marimo._code_mode._context import (
     AsyncCodeModeContext,
-    NotebookCellData,
     get_context,
 )
+from marimo._notebook.document import NotebookCell
 
 __all__ = [
     "AsyncCodeModeContext",
-    "NotebookCellData",
+    "NotebookCell",
     "get_context",
 ]
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -24,9 +24,7 @@ Usage::
 from __future__ import annotations
 
 import sys
-from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, overload
-from uuid import uuid4
 
 from marimo import _loggers
 from marimo._ast.cell import CellConfig, CellImpl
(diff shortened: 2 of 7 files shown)


frontend/src/core/cells/apply-transaction.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import type { NotificationMessageData } from "../kernel/messages";
+import type { CellConfig } from "../network/types";
+import type { CellId } from "./ids";
+
+type Transaction =
+  NotificationMessageData<"notebook-document-transaction">["transaction"];
+type TransactionOp = Transaction["ops"][number];
+
+interface CellActions {
```
mscolnick: might be able to `Pick<CellActions, ...>` here.

manzt: I think i ran into some circular reference (because the type is inferred). We could probably have an explicit type somewhere that we reuse as a type-only import.

frontend/src/core/cells/document-transaction-middleware.ts
```diff
@@ -0,0 +1,164 @@
...
+type DocumentAction =
+  | { type: "createNewCell"; payload: { cellId: CellId } }
+  | { type: "deleteCell"; payload: { cellId: CellId } }
+  | { type: "moveCell"; payload: { cellId: CellId } }
+  | { type: "sendToTop"; payload: { cellId: CellId } }
+  | { type: "sendToBottom"; payload: { cellId: CellId } }
+  | { type: "dropCellOverCell"; payload: unknown }
+  | { type: "dropCellOverColumn"; payload: unknown }
+  | { type: "updateCellCode"; payload: { cellId: CellId; code: string } }
+  | { type: "updateCellName"; payload: { cellId: CellId; name: string } };
```
manzt: @mscolnick similar here... our middleware interface isn't strictly typed on the reducer. I hard-coded types to try to reduce the diff.

frontend/src/core/cells/apply-transaction.ts
```diff
@@ -0,0 +1,150 @@
...
+    switch (op.type) {
+      case "create-cell": {
+        // Determine the anchor cell and insertion direction.
+        // `after` means insert after that cell (before=false).
+        // `before` means insert before that cell (before=true).
+        // Neither means append at the end.
+        let cellId: CellId | "__end__" = "__end__";
+        let before = false;
+        if (op.after) {
+          cellId = op.after as CellId;
```
mscolnick: i think these `as` could be cleaned up now from your other PRs

frontend/src/core/cells/apply-transaction.ts
mscolnick: ah yes it is inferred. maybe we can make middleware additive

mscolnick: maybe creates other problems/smells, but something like this

image

PR #8848, opened 2026-03-24
Fix table cell styles/hover texts when sorting desc

### LINKED ISSUE #8847: `style_cell` styles lost when sorting table in descending order
### Describe the bug

When using `mo.ui.table` with a `style_cell` callback, cell styles are correctly applied on initial
render and survive ascending sort, but are lost when any column is sorted in descending order.

Reproduction:

```
import marimo as mo
import pandas as pd

df = pd.DataFrame({
    "Name": ["alpha", "beta", "gamma", "delta"],
    "Score": [1.0, 2.0, 3.0, 4.0],
})

def style(row, col, val):
    if col == "Score":
        try:
            v = int(val)
        except (ValueError, TypeError):
            return {}
        bg = {1: "rgba(39,174,96,0.2)", 2: "rgba(243,156,18,0.2)",
            3: "rgba(230,126,34,0.3)", 4: "rgba(231,76,60,0.3)"}.get(v, "")
        return {"background-color": bg} if bg else {}
    return {}

table = mo.ui.table(df, style_cell=style)
table
```
1. Observe that Score cells are color-coded on initial render.
2. Click the Score column header to sort ascending — colors persist.
3. Click again to sort descending — colors disappear.

In marimo/_plugins/ui/_impl/table.py, _style_cells() (line ~1263) reverses the row ID range when
descending=True:
```
if descending and total_rows != "too_many":
    row_ids = range(
        total_rows - 1 - skip, total_rows - 1 - skip - take, -1
    )
```
However, `self._searched_manager` already contains the data in sorted order (sorting was applied in
`_apply_filters_query_sort())`. The first page of results is always at positions `0..page_size-1` in
the sorted manager, regardless of sort direction. The reversed range generates style dict keys
(e.g., 199, 198, ..., 150) that don't match any row IDs on the displayed page (0, 1, ..., 49), so
the frontend discards them.

Suggested fix:
Remove the descending reversal in `_style_cells()`. The row IDs should always be `range(skip, skip +
take)` since the searched manager is already sorted:

```
# Before:
if response.all_rows or response.error:
    row_ids = range(skip, skip + take)
    if descending and total_rows != "too_many":
        row_ids = range(
            total_rows - 1 - skip, total_rows - 1 - skip - take, -1
        )

# After:
if response.all_rows or response.error:
    row_ids = range(skip, skip + take)
```

The same issue likely affects `_hover_cells()` which has identical logic.

Workaround:

Monkey-patch _style_cells to ignore the descending flag:
```
from marimo._plugins.ui._impl import table as _tbl
_orig = _tbl.table._style_cells

def _patched(self, skip, take, total_rows, descending=False):
    return _orig(self, skip, take, total_rows, descending=False)

_tbl.table._style_cells = _patched
```
Version: marimo 0.21.1

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

_No response_

gilangjavier:
Fixes #8847.

When the table has no active filters/search (GetRowIdsResponse.all_rows=True), _style_cells and _hover_cells were generating row id ranges that can diverge from the current sorted view, causing styles/hover texts to disappear (notably for descending sorts by a non-index column).

This change derives row ids from the current page of the searched manager when a stable row id is available, and falls back to positional ranges otherwise.

Adds a pandas regression test for descending sort by a non-index column.

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -1238,18 +1238,48 @@ class table(
         )
         return column_preview
 
+    def _resolve_row_ids(
+        self,
+        skip: int,
+        take: int,
+        total_rows: Union[int, Literal["too_many"]],
+    ) -> Union[list[Any], range]:
+        response = self._get_row_ids(EmptyArgs())
+
+        # Clamp take to the available row count for non-lazy tables
+        if total_rows != "too_many" and skip + take > total_rows:
+            take = total_rows - skip
+
+        if response.all_rows or response.error:
+            fallback_row_ids: range = range(skip, skip + take)
+
+            if response.all_rows and self._has_stable_row_id:
+                try:
+                    page = self._searched_manager.take(take, skip)
+                    # Keep provided row IDs as-is to support non-integer IDs.
+                    return page.data[INDEX_COLUMN_NAME].to_list()
+                except Exception as exc:
+                    LOGGER.warning(
+                        "Falling back to positional row IDs after failed "
+                        "stable-row-id lookup: %s",
+                        exc,
+                        exc_info=True,
+                    )
+
+            return fallback_row_ids
+
+        return response.row_ids[skip : skip + take]
+
     def _style_cells(
         self,
         skip: int,
         take: int,
         total_rows: Union[int, Literal["too_many"]],
-        descending: bool = False,
+        row_ids: Optional[Union[list[Any], range]] = None,
     ) -> Optional[CellStyles]:
         """Calculate the styling of the cells in the table."""
         if self._style_cell is None:
             return None
-
-        del descending
 
         def do_style_cell(row: str, col: str) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


Light2Dark (dismissed):

Light2Dark:
can address the copilot comments, I think some are valid.

gilangjavier:
I have read the CLA Document and I hereby sign the CLA\n\nrecheck\n\nI also pushed a follow-up commit to address the review notes: 7f71a0854\n\nChanges made:\n- keep stable row IDs as-is (no forced int conversion) in style/hover paths\n- add warning logs when stable row-id lookup fails and we fall back to positional IDs\n- reuse resolved row IDs between style and hover generation to avoid duplicate page fetches\n- remove unused descending plumbing in style/hover internals\n\nValidation run:\n- uv run --group test pytest tests/_plugins/ui/_impl/test_table.py -q

PR #8845, opened 2026-03-24
[nd 4/5] Emit transactions from `code_mode`

manzt:
Prev #8842, #8843, #8844. This PR closes the kernel → session loop.

**Transaction emission.** `_apply_ops` previously assembled three separate legacy notifications (`UpdateCellCodesNotification` grouped by stale status, plus `UpdateCellIdsNotification`). Now it builds typed `Op`s via `_plan_to_document_ops()` and broadcasts a **single** `NotebookDocumentTransactionNotification`.

**Document context.** The execution endpoint snapshots `session.document.cells` into `ExecuteScratchpadCommand.notebook_cells`. The kernel handler sets a `ContextVar` from it so `AsyncCodeModeContext` can read cell state from the document.`_CellsView` now delegates entirely to the document and returns `NotebookCell` directly, removing the `NotebookCellData` wrapper and the module-level `_cell_names` dict.

PR #8849, opened 2026-03-24
[nd 6/5] Back `NotebookDocument` cell text with Loro CRDT

manzt:
Prev #8842, #8843, #8844, #8845, #8846. Experiment with composing the document model with a CRDT for cell text.

The document event model from the previous PRs sequences structural changes via transactions, but cell text was still plain strings with Loro running as a parallel system. This explores a split-ownership model: `NotebookDocument` keeps structural metadata (ordering, names, configs), Loro owns cell text. One source of truth for each concern, no reconciliation.

```python
doc = NotebookDocument(create_doc())
doc.add_cell(cell_id, code="x = 1", name="__", config=CellConfig())
doc.loro_doc.commit()

doc.get_cell(cell_id).code  # materializes from LoroText.to_string()
```

Non-interactive writes (kernel, file-watch, code_mode) go through `SetCode` ops which write to Loro. Character-level edits flow through the Loro WebSocket directly. The session's `LoroDoc` is shared with the RTC doc manager — server-originated mutations auto-broadcast to connected clients. On the frontend, code changes flow exclusively through Loro; the transaction middleware no longer handles `set-code`.

### DIFF
--- a/marimo/_notebook/_loro.py
+++ b/marimo/_notebook/_loro.py
@@ -1,0 +1,19 @@
+# Copyright 2026 Marimo. All rights reserved.
+"""Typed wrappers for ``loro`` constructors.
+
+The ``loro`` stubs omit return types on ``__new__``, which triggers
+mypy ``no-untyped-call``.  These helpers provide correctly-typed
+construction so the rest of the codebase stays clean.
+"""
+
+from __future__ import annotations
+
+from loro import LoroDoc, LoroText
+
+
+def create_doc() -> LoroDoc:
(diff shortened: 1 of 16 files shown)


marimo/_notebook/_loro.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2026 Marimo. All rights reserved.
+"""Typed wrappers for ``loro`` APIs with incomplete stubs.
+
+The ``loro`` stubs omit return types on ``__new__`` and the
```
dmadisetti, line 4: We should raise upstream tbh

PR #8855, opened 2026-03-24
fix: route stdout from asyncio tasks to correct cell

### LINKED ISSUE #8816: Stdout in async task is ignored?
### Describe the bug

```py
import marimo

__generated_with = "0.20.4"


@app.cell
def _():
    import time
    import asyncio

    async def foo():
        et = time.time() + 1
        count = 0
        while time.time() < et:
            count += 1
            await asyncio.sleep(0.01)
        print("done", count)

    return asyncio, foo


@app.cell
async def _(foo):
    await foo()
    return


@app.cell
def _(asyncio, foo):
    return (x,)


@app.cell
def _(x):
    x
    return


if __name__ == "__main__":
    app.run()
```

Running this looks like this:

Image

BillionClaw:
## 📝 Summary

Closes #8816

## 🔍 Description of Changes

When a cell uses `asyncio.create_task()` to spawn a background task that calls `print()`, the output was being silently dropped instead of being routed to the cell's output in the frontend.

**Root cause:** `print_override()` (the patched `print` function) relied on `execution_context.cell_id` to route print output to the correct cell. However, when an async cell is executed, `asyncio.ensure_future()` schedules the coroutine to run asynchronously. By the time the task actually executes and calls `print()`, the `execution_context` for the creating cell may have been cleared, causing `print_override()` to fall back to the original `print()` which writes to a redirected file descriptor that the Watcher doesn't capture.

**Fix:** Introduced a `contextvars.ContextVar` (`_asyncio_task_cell_id`) that captures the `cell_id` when the async cell execution begins. This context variable is automatically propagated to any asyncio tasks created via `create_task()` within the cell via Python's context variable inheritance mechanism. When `print_override()` finds no `execution_context`, it falls back to this task-level `cell_id`.

### Changes:

1. **New file `marimo/_messaging/_async_task_context.py`**: Defines `_asyncio_task_cell_id`, a `contextvars.ContextVar[str | None]` shared between `print_override` and the cell runner.

2. **`marimo/_messaging/print_override.py`**: Modified to fall back to `_asyncio_task_cell_id.get()` when `execution_context` is `None`. This ensures print output from async tasks is still routed correctly.

3. **`marimo/_runtime/runner/cell_runner.py`**: When running a coroutine cell, wraps the execution in a context manager that sets `_asyncio_task_cell_id` to the cell's ID. Python's context variable mechanism propagates this to any `create_task()` tasks.

4. **`tests/_messaging/test_print_override.py`**: Added two new tests:
   - `test_print_override_asyncio_task_with_execution_context`: Verifies `execution_context` takes precedence when both are available
   - `test_print_override_asyncio_task_fallback_to_task_cell_id`: Verifies the task cell_id is used when `execution_context` is `None`

## 📋 Checklist

### DIFF
--- a/marimo/_messaging/_async_task_context.py
+++ b/marimo/_messaging/_async_task_context.py
@@ -7,6 +7,7 @@ print_override can route print() output to the correct cell.
 
 This module defines the context variable separately to avoid circular imports.
 """
+
 from __future__ import annotations
 
 import contextvars
@@ -15,6 +16,6 @@ import contextvars
 # When an async cell creates tasks, we capture the cell_id so that print_override
 # can route print() output to the correct cell even when the asyncio task runs
 # after the execution context has been cleared.
-_asyncio_task_cell_id: contextvars.ContextVar[str | None] = contextvars.ContextVar(
-    "_asyncio_task_cell_id", default=None
+_asyncio_task_cell_id: contextvars.ContextVar[str | None] = (
+    contextvars.ContextVar("_asyncio_task_cell_id", default=None)
 )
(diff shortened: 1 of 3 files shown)


dmadisetti:
Thanks! Can you sign the CLA and provide a screenshot of the correct functionality?

PR #8860, opened 2026-03-25
fix: recover polling file watcher from transient file absence

### LINKED ISSUE #8624: File watcher crash reporting missing file and never recovers
### Describe the bug

Marimo was started via `uv run marimo edit --watch`. A notebook was opened. The notebook was also edited concurrently via `vim`. At some point, the file watcher crashes with:
```
[W 260309 19:16:14 file_watcher:84] File at [filename.py] does not exist.
Task exception was never retrieved
future: <Task finished name='Task-21' coro=<PollingFileWatcher._poll() done, defined at [mydir]/.venv/lib/python3.13/site-packages/marimo/_utils/file_watcher.py:81> exception=FileNotFoundError('File at [filename.py] does not exist.')>
  File "[mydir]/.venv/lib/python3.13/site-packages/marimo/_utils/file_watcher.py", line 85, in _poll
    raise FileNotFoundError(f"File at {self.path} does not exist.")
FileNotFoundError: File at [filename.py] does not exist.
```

Apologies for redacting the file names. This has happened to me essentially every single time I'm using the file watcher feature. The file always exists as far as I can see (I don't think `vim` saving would temporarily delete it..); most of the time the changes are picked up correctly by Marimo, but after a while this error happens, and I guess that afterwards the watcher is dead and is not spawned anymore, so I'm forced to completely relaunch Marimo.

I haven't yet figured out if there's a pattern unfortunately. I'm not sure if the "reload" feature might be interfering, as I usually also have a bunch of `[E 260309 18:51:07 file_change_handler:222] Error loading file: invalid syntax. Perhaps you forgot a comma? ([filename.py], line 182)` errors, which prevent the reloading of a file into Marimo.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.20.4",
  "editable": false,
  "location": "[mydir]/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "Processor": "x86_64",
  "Python Version": "3.13.1",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.11.0",
    "uv": "0.6.10"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.17.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.41.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "loro": "1.10.3",
    "pandas": "3.0.1"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

Krishnachaitanyakc:
## Summary

Fixes #8624

The `PollingFileWatcher._poll()` method crashed with an unhandled `FileNotFoundError` when the watched file was temporarily missing. This commonly occurs with editors like **vim** that save via a delete-and-rename cycle, causing a brief window where the file does not exist. Once the exception propagated, the polling task died permanently and the file watcher never recovered -- forcing users to restart marimo.

**Changes:**

- Instead of raising `FileNotFoundError` immediately when the file is absent, the watcher now tolerates up to `MAX_MISSING_POLLS` (5) consecutive absences before gracefully stopping
- If the file reappears within that window, the missing counter resets and the watcher resumes normal operation, firing the change callback as expected
- Added `_missing_count` instance variable and `MAX_MISSING_POLLS` class constant to `PollingFileWatcher`

## Test plan

- [x] Added `test_polling_file_watcher_transient_missing` -- verifies the watcher survives a brief file deletion (simulating vim save) and correctly detects the file change when it reappears
- [x] Added `test_polling_file_watcher_permanently_missing` -- verifies the watcher gracefully stops after `MAX_MISSING_POLLS` consecutive checks when the file is truly gone
- [x] All existing file watcher tests continue to pass
- [x] `ruff check` and `ruff format` pass on changed files

### DIFF
--- a/tests/_utils/test_file_watcher.py
+++ b/tests/_utils/test_file_watcher.py
@@ -6,6 +6,7 @@ import os
 import shutil
 from pathlib import Path
 from tempfile import NamedTemporaryFile
+from unittest.mock import patch
 
 import pytest
 
@@ -151,34 +152,32 @@ async def test_file_watcher_manager() -> None:
         os.remove(tmp_path2)
 
 
-async def test_polling_file_watcher_transient_missing() -> None:
+@patch("marimo._utils.file_watcher.PollingFileWatcher.POLL_SECONDS", 0.05)
+async def test_polling_file_watcher_transient_missing(tmp_path: Path) -> None:
     """Test that the watcher survives a transient file deletion (e.g. vim save).
 
     Editors like vim save by writing to a temp file, deleting the original,
     and renaming the temp. The polling watcher should tolerate the brief
     absence and fire the callback once the file reappears with new content.
     """
-    with NamedTemporaryFile(delete=False) as tmp_file:
-        tmp_path = Path(tmp_file.name)
-        tmp_file.write(b"original")
+    tmp_file = tmp_path / "nb.py"
+    tmp_file.write_bytes(b"original")
 
     callback_calls: list[Path] = []
 
     async def test_callback(path: Path) -> None:
         callback_calls.append(path)
 
-    PollingFileWatcher.POLL_SECONDS = 0.05
-
     loop = asyncio.get_event_loop()
-    watcher = PollingFileWatcher(tmp_path, test_callback, loop)
+    watcher = PollingFileWatcher(tmp_file, test_callback, loop)
     watcher.start()
 
     await asyncio.sleep(0.1)
 
     # Simulate vim-style save: delete original, then recreate
-    os.remove(tmp_path)
+    os.remove(tmp_file)
     await asyncio.sleep(0.1)  # File is missing for a poll cycle
-    with open(tmp_path, "w") as f:  # noqa: ASYNC230
+    with open(tmp_file, "w") as f:  # noqa: ASYNC230
         f.write("updated content")
 
     # Wait for the watcher to detect the change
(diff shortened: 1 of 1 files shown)


tests/_utils/test_file_watcher.py
```diff
@@ -151,6 +151,85 @@ async def callback3(path: Path) -> None:
         os.remove(tmp_path2)
 
 
+async def test_polling_file_watcher_transient_missing() -> None:
```
mscolnick: for a slightly more concise setup, you can use `tmp_path`:

```
async def test_polling_file_watcher_transient_missing(tmp_path: Path) -> None:
    tmp_file = tmp_path / "nb.py"
    tmp_file.write(b"original")
```

this will also handle cleanup

tests/_utils/test_file_watcher.py
```diff
@@ -151,6 +151,85 @@ async def callback3(path: Path) -> None:
...
+    with NamedTemporaryFile(delete=False) as tmp_file:
+        tmp_path = Path(tmp_file.name)
+        tmp_file.write(b"original")
+
+    callback_calls: list[Path] = []
+
+    async def test_callback(path: Path) -> None:
+        callback_calls.append(path)
+
+    PollingFileWatcher.POLL_SECONDS = 0.05
```
mscolnick: also for `PollingFileWatcher.MAX_MISSING_POLLS`, you should be able to use a patch decorator, that will handle cleanup

```
@patch("marimo._utils.file_watcher.PollingFileWatcher.MAX_MISSING_POLLS", 0.05)
async def test_polling_file_watcher_transient_missing()
```

Krishnachaitanyakc:

Svalorzen:
Hi, I'm the original issue opener, thanks for trying to fix the problem! I have a question about this fix: as far as I understand, even now, if the file watcher crashes, marimo is still technically working and the python kernel is not deactivated. The notebook stays open, and in theory marimo would still be able to save the file from the version kept in memory. With the change in this PR, if the file is re-created/marimo saves it, will the file watcher automatically start working again?

If not, perhaps there could be some kind of button to force it to start again without necessarily rebooting marimo?

PR #8869, opened 2026-03-25
Avoid normalizing cloudpathlib providers

### LINKED ISSUE #8868: Not possible to use Path extended from cloudpathlib
### Describe the bug

In order to mitigate issue [MO-5191](https://linear.app/marimo/issue/MO-5191/file-browser-does-not-work-with-smb-path) I [added a new provider for cloudpathlib](<https://cloudpathlib.drivendata.org/stable/contributing/#adding-a-new-provider>) locally. However due to a call to [normalize_path](<https://github.com/marimo-team/marimo/blob/3cf9f4aff78997c274c5a731f712f752690212f7/marimo/_utils/paths.py#L54>) in [file_browser.py:175](<https://github.com/marimo-team/marimo/blob/3cf9f4aff78997c274c5a731f712f752690212f7/marimo/_plugins/ui/_impl/file_browser.py#L175>) the string is converted from "smb://server.com/share/..." to "smb:/server.com/share/..." (note the missing "/" after "smb:/").

```
    # Frontend can't handle relative paths, so normalize it and make it absolute
    # Use normalize_path to avoid symlink resolution
    self._initial_path = self._create_path(normalize_path(initial_path))
```

The reason that this works for the "build in" S3Path, GSPath and AzureBlobPath is a check in [marimo/\_utils/paths.py:76](<https://github.com/marimo-team/marimo/blob/3cf9f4aff78997c274c5a731f712f752690212f7/marimo/_utils/paths.py#L76>):

```
# Skip normalization for cloud paths (e.g., S3Path, GCSPath, AzurePath)
# os.path.normpath corrupts URI schemes like s3:// by reducing them to s3:/
if path.__class__.__module__.startswith("cloudpathlib"):
    return path
```

Above check looks a bit fragile to me, but I am not sure what the best solution is.

Currently my workaround is to "rename" my provider like this:

```
    self.__class__.__module__ = "cloudpathlib"
```

### Will you submit a PR?

- [X] Yes

### Environment

<details><summary>\`\`\`</summary>

{<br>  "marimo": "0.20.4",<br>  "editable": false,<br>  "location": "/home/xxxx/.venv/lib/python3.12/site-packages/marimo",<br>  "OS": "Linux",<br>  "OS Version": "6.17.0-19-generic",<br>  "Processor": "x86_64",<br>  "Python Version": "3.12.3",<br>  "Locale": "--",<br>  "Binaries": {<br>    "Browser": "146.0.7680.153",<br>    "Node": "--",<br>    "uv": "0.9.26"<br>  },<br>  "Dependencies": {<br>    "click": "8.3.1",<br>    "docutils": "0.22.4",<br>    "itsdangerous": "2.2.0",<br>    "jedi": "0.19.2",<br>    "markdown": "3.10.1",<br>    "narwhals": "2.15.0",<br>    "packaging": "26.0",<br>    "psutil": "7.2.1",<br>    "pygments": "2.19.2",<br>    "pymdown-extensions": "10.20",<br>    "pyyaml": "6.0.3",<br>    "starlette": "0.52.1",<br>    "tomlkit": "0.14.0",<br>    "typing-extensions": "4.15.0",<br>    "uvicorn": "0.40.0",<br>    "websockets": "16.0"<br>  },<br>  "Optional Dependencies": {<br>    "altair": "6.0.0",<br>    "loro": "1.10.3",<br>    "openai": "2.17.0",<br>    "pandas": "3.0.0",<br>    "polars": "1.39.3",<br>    "pyarrow": "23.0.1",<br>    "pytest": "8.4.2"<br>  },<br>  "Experimental Flags": {}<br>}

```
```

</details>

### Code to reproduce

*No response*

johannesc:
Cloudpathlib providers typically have a path like
"s3://my-bucket/folder/" which, when normalized is converted to "s3:/my-bucket/folder/" which will not work.

Instead of checking the module, which prevents external providers, isinstance() is now used to detect regular Path objects.

## 📝 Summary

Closes #8868

## 🔍 Description of Changes

For more details see #8868.

## 📋 Checklist

- [x] ruff check and ruff format pass on changed files

### DIFF
--- a/marimo/_utils/paths.py
+++ b/marimo/_utils/paths.py
@@ -117,7 +117,7 @@ def normalize_path(path: Path) -> Path:
     """
     # Skip normalization for cloud paths (e.g., S3Path, GCSPath, AzurePath)
     # os.path.normpath corrupts URI schemes like s3:// by reducing them to s3:/
-    if path.__class__.__module__.startswith("cloudpathlib"):
+    if "CloudPath" in [cls.__name__ for cls in type(path).__mro__]:
         return path
 
     # Make absolute if relative (relative to current working directory)


johannesc:

johannesc:
> The fix works correctly today because `cloudpathlib.CloudPath` does not inherit from `pathlib.Path`, so `not isinstance(path, Path)` correctly short-circuits for cloud paths. However, the guard is now broader than intended: any object that isn't a `pathlib.Path` subclass will be silently returned unchanged, which could mask bugs where an unexpected type is passed in. The original `__module__` check was fragile (string-based), but it was at least explicit about the intent.
> 
> Worth noting: the docstring comment on line 74 still says `# Skip normalization for cloud paths (e.g., S3Path, GCSPath, AzurePath)`, which is now misleading since the condition no longer has any cloudpathlib-specific knowledge. If cloudpathlib ever changes `CloudPath` to subclass `pathlib.Path` (unlikely but not impossible), this check would silently stop protecting cloud paths.
> 
> A middle ground worth considering: keep the `isinstance` check but add a narrower fallback, or at least update the comment to reflect that this now protects against _any_ non-`pathlib.Path` subclass, not just cloudpathlib types. A test covering a mock `CloudPath`-like object that doesn't inherit `pathlib.Path` would also help ensure this assumption doesn't regress silently.

I changed to using `if "CloudPath" in [cls.__name__ for cls in type(path).__mro__]` instead which is a bit tighter check. Also note that there is another pull-request (#8879) which perhaps is better (if so I can just close this one)?

mscolnick:
Closed by https://github.com/marimo-team/marimo/pull/8929

PR #8882, opened 2026-03-26
feat(table): add default_sort for mo.ui.table initial ordering

### LINKED ISSUE #3860: Add default_sort parameter to `mo.ui.table`

phusi319:
Closes #3860

This PR adds support for `default_sort` in `mo.ui.table` to control initial table ordering.

### Summary
- **Backend (`marimo/_plugins/ui/_impl/table.py`)**
  - Adds a `default_sort` argument to `ui.table`.
  - Validates that the requested sort column exists in table columns.
  - Applies initial search sorting with ascending order by default.
  - Exposes `default-sort` in component args for frontend initialization.
- **Frontend (`frontend/src/plugins/impl/DataTablePlugin.tsx`)**
  - Accepts `defaultSort` in plugin schema/data.
  - Initializes table sorting state from `defaultSort` as ascending (`desc: false`).
- **Tests (`tests/_plugins/ui/_impl/test_table.py`)**
  - Adds coverage that initial render uses `default_sort` ordering.
  - Adds validation test that missing `default_sort` column raises an error.

Behavior note: sorting is ascending by default when `default_sort` is provided.

### DIFF
--- a/frontend/src/plugins/impl/DataTablePlugin.tsx
+++ b/frontend/src/plugins/impl/DataTablePlugin.tsx
@@ -255,6 +255,7 @@ export const DataTablePlugin = createPlugin<S>("marimo-table")
         .nullable()
         .default(null),
       showDownload: z.boolean().default(false),
+      defaultSort: z.string().optional(),
       showFilters: z.boolean().default(false),
       showColumnSummaries: z
         .union([z.boolean(), z.enum(["stats", "chart"])])
@@ -500,8 +501,15 @@ export const LoadingDataTableComponent = memo(
 
     const search = props.search;
     const setValue = props.setValue;
+    const initialSorting = useMemo<SortingState>(
+      () =>
+        props.defaultSort
+          ? [{ id: props.defaultSort, desc: false }]
+          : Arrays.EMPTY,
+      [props.defaultSort],
+    );
     // Sorting/searching state
-    const [sorting, setSorting] = useState<SortingState>([]);
+    const [sorting, setSorting] = useState<SortingState>(initialSorting);
     const [paginationState, setPaginationState] =
       React.useState<PaginationState>({
         pageSize: props.pageSize,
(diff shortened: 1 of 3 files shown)


phusi319:

phusi319:
Thank you for the thorough review! I believe all three concerns have been addressed in the current code:

### 1.  Breaking change - `default_sort` parameter position
The parameter is already **keyword-only** (placed after `*,` at line 473-474):
`python
max_columns: MaxColumnsType = MAX_COLUMNS_NOT_PROVIDED,
*,
default_sort: Optional[str] = None,
`
This preserves backwards compatibility since callers cannot pass it positionally.

### 2.  Validation consistency with `_internal_lazy`
The validation runs **regardless** of `_internal_lazy` (lines 679-684):
`python
if default_sort is not None:
    existing_columns = set(self._manager.get_column_names())
    if default_sort not in existing_columns:
        raise ValueError(...)

if not _internal_lazy:  # <-- validation above is outside this block
    ...
`

### 3.  Frontend initial page fast-path
The `canShowInitialPage` logic (lines 570-580) handles the default sort case:
`javascript
const canShowInitialPage =
  searchQuery === "" &&
  paginationState.pageIndex === 0 &&
  filters.length === 0 &&
  (sorting.length === 0 ||
    (sorting.length === 1 &&
      Boolean(props.defaultSort) &&
      sorting[0]?.id === props.defaultSort &&
      sorting[0]?.desc === false)) &&  // Allow initial page when sort matches defaultSort
  !props.lazy &&
  !pageSizeChanged;
`
This allows the fast-path when sorting exactly matches the default sort, avoiding unnecessary latency.

Let me know if I've misunderstood any of the feedback!

Light2Dark:
Converting to a draft for now, thanks for the contribution! https://github.com/marimo-team/marimo/issues/3860
But let's get more feedback first.

PR #8884, opened 2026-03-26
Strengthen `createReducer` types to enforce action shapes

manzt:
Follow up to #8842 

The generic `ReducerAction<any>` type allowed dispatching actions with arbitrary `type` and `payload` fields, which meant typos and shape mismatches in action objects were silently accepted. This replaces it with a derived `ReducerActionOf<RH>` union that infers the exact `type`/`payload` pairs from the reducer handler map, giving end-to-end type safety from dispatch through middleware.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14900,14 +14900,14 @@ snapshots:
       chai: 5.3.3
       tinyrainbow: 2.0.0
 
-  '@vitest/mocker@3.2.4(msw@2.12.7(@types/node@20.19.30)(typescript@5.9.3))(rolldown-vite@7.3.1(@types/node@20.19.30)(esbuild@0.27.3)(jiti@2.6.1)(terser@5.44.1)(yaml@2.8.3))':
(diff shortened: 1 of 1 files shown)


frontend/src/core/cells/__tests__/runs.test.ts
```diff
@@ -148,15 +148,15 @@ describe("RunsState Reducer", () => {
           run_id: runId,
           cell_id: testCellId,
           timestamp: runStartTimestamp + 5000,
-          status: "success",
+          status: "running"
```
manzt: These were real type errors ("success" isn't in our types)

frontend/src/core/cells/__tests__/runs.test.ts
```diff
@@ -226,7 +226,8 @@ describe("RunsState Reducer", () => {
           status: "running",
           output: {
             channel: "stderr",
-            text: "Error occurred",
+            data: "Error occurred",
+            mimetype: "text/plain",
```
manzt, line 230: same

PR #8888, opened 2026-03-26
fix: skip row limit enforcement for EXPLAIN queries

### LINKED ISSUE #8328: Explain SQL-Statement previews rows with no chance to scroll
### Describe the bug

I want to analyse a rather large SQL-Statement. I add the explain-Keyword in front of the query. When displaying the result, I see the notification that marimo only shows a preview of the first 10 rows, which is the default value in my setting. However, there is no way to navigate through the following pages like there is when displaying tables.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.19.11",
  "editable": false,
  "location": "~/ivp/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.12.69+deb13-amd64",
  "Processor": "",
  "Python Version": "3.13.5",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "0.10.0"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.13.0",
    "packaging": "25.0",
    "psutil": "7.1.3",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "pandas": "2.3.3",
    "polars": "1.35.2",
    "pyarrow": "22.0.0",
    "pytest": "9.0.2",
    "python-lsp-ruff": "2.3.0",
    "ruff": "0.15.0"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

VishakBaddur:
Fixes #8328

## Root Cause

EXPLAIN queries were being truncated to the default result limit (e.g. 10 rows) before the EXPLAIN-specific display logic ran.

`enforce_own_limit` was `True` for EXPLAIN queries, so `df.limit()` / `df.head()` truncated the dataframe first. Then `extract_explain_content()` only saw the truncated data — and since `plain_text()` has no pagination, users had no way to scroll through the full output.

## Fix

One-line change: add `not is_explain_query(query)` to the `enforce_own_limit` condition:
```python
enforce_own_limit = not has_limit and default_result_limit is not None and not is_explain_query(query)
```

EXPLAIN queries now bypass the row limit entirely, so all rows are passed to `extract_explain_content()`.

## Testing

Added `test_explain_query_not_truncated_by_limit` to `tests/_sql/test_duckdb.py`: verifies that EXPLAIN output is not truncated when `MARIMO_SQL_DEFAULT_LIMIT` is set.

### DIFF
--- a/marimo/_sql/sql.py
+++ b/marimo/_sql/sql.py
@@ -106,7 +106,11 @@ def sql(
     except OSError:
         default_result_limit = None
 
-    enforce_own_limit = not has_limit and default_result_limit is not None and not is_explain_query(query)
+    enforce_own_limit = (
+        not has_limit
+        and default_result_limit is not None
+        and not is_explain_query(query)
+    )
 
     custom_total_count: Optional[Literal["too_many"]] = None
     if enforce_own_limit:
--- a/tests/_sql/test_duckdb.py
+++ b/tests/_sql/test_duckdb.py
@@ -295,13 +295,14 @@ def test_duckdb_engine_sql_output_formats(
 @pytest.mark.skipif(not HAS_DUCKDB, reason="requires duckdb")
 @pytest.mark.skipif(not HAS_POLARS, reason="requires polars")
 def test_explain_query_not_truncated_by_limit(
-    duckdb_connection: "duckdb.DuckDBPyConnection",
+    duckdb_connection: duckdb.DuckDBPyConnection,
 ) -> None:
     """EXPLAIN queries should not be truncated by the default result limit.
 
     Regression test for https://github.com/marimo-team/marimo/issues/8328
     """
     import os
+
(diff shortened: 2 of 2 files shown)


Light2Dark:
Thanks for the contribution. Since the issue cannot be reproduced, closing this for now until we understand the problem better.

PR #8900, opened 2026-03-27
fix: remove notifications with overlapping work as new document transactions

mscolnick:
This pull request removes the legacy "sync cell IDs" mechanism from both the frontend and backend, consolidating cell reordering and synchronization under the newer document transaction system. This simplifies the codebase, reduces redundancy, and ensures cell order changes are handled consistently.


marimo/_runtime/runtime.py
```diff
@@ -2152,10 +2151,6 @@ async def instantiate(self, request: CreateNotebookCommand) -> None:
             LOGGER.info("App is already instantiated, skipping instantiation.")
             return
 
-        broadcast_notification(
-            UpdateCellIdsNotification(cell_ids=list(request.cell_ids))
-        )
```
mscolnick, line 2157: if we need this back, we can `broadcast_notification(`NotebookDocumentTransactionNotification(with reorder cells change)`

marimo/_runtime/runtime.py
```diff
@@ -2153,7 +2154,12 @@ async def instantiate(self, request: CreateNotebookCommand) -> None:
             return
 
         broadcast_notification(
-            UpdateCellIdsNotification(cell_ids=list(request.cell_ids))
+            NotebookDocumentTransactionNotification(
+                transaction=Transaction(
+                    changes=(ReorderCells(cell_ids=tuple(request.cell_ids)),),
+                    source="kernel",
+                )
+            )
```
manzt, line 2162: do we want to apply here?

```py
applied = session.document.apply(Transaction(...))
broadcast_notification(NotebookDocumentTransactionNotification(transaction=applied))
```

marimo/_session/state/serialize.py
```diff
@@ -91,17 +91,7 @@ def serialize_session_view(
...
-        else:
-            # TODO: This is a problem for exporting to HTML, but not for caching the session view for replays. Something seems off.
-            LOGGER.debug(
-                "When serializing session view, the notebook-order of cells was "
-                "not known. This may cause issues when attempting to "
-                "reconstruct the notebook state from the serialized session "
-                "view."
-            )
-            cell_ids = view.cell_notifications.keys()
+        cell_ids = view.cell_notifications.keys()
```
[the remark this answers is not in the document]
manzt: yeah would be nice to get this from session.document. We could honestly have the document on a ContextVar if we wanted DI.

marimo/_runtime/runtime.py
mscolnick: i dont think so. i think that would always be a noop

marimo/_session/state/serialize.py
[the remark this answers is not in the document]
mscolnick: yea i need to revisit this. i think with the document, i can now wire `cell_ids` properly removing their optionality

PR #8904, opened 2026-03-27
refactor: enforce cell_ids parameter in session serialization and caching

mscolnick:
- Updated `serialize_session_snapshot` and `persist_session_view_to_cache` functions to require `cell_ids` as a mandatory parameter.
- Enhanced `CachingExtension` to include the `document` parameter in `SessionCacheManager`.
- Adjusted `SessionCacheWriter` to utilize `cell_ids` from the `NotebookDocument` for improved session management.
- Updated tests to reflect changes in the serialization process, ensuring compatibility with the new mandatory `cell_ids` parameter.

### DIFF
--- a/marimo/_server/export/_session_cache.py
+++ b/marimo/_server/export/_session_cache.py
@@ -59,7 +59,7 @@ def serialize_session_snapshot(
     view: SessionView,
     *,
     notebook_path: str | Path | None,
-    cell_ids: Iterable[CellId_t] | None = None,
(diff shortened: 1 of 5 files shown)


mscolnick:
> I supposed there's never a case where `cell_id != view.cell_id` ? Looks good with passing tests

Unlikely. It's possible the view has a “leak” in edit mode and doesn't clean up notifications in some cases but that's ok - source of truth is now the document.

PR #8905, opened 2026-03-27
feat: unify row viewer and column explorer

kirangadhave:
## 📝 Summary

Combines the row viewer and column explorer in a single tabbed "Tab Explorer Pane"

## 🔍 Description of Changes
- Single tabbed panel with **Rows** and **Columns** tabs, replacing two independent panels
  - Single toolbar button toggles the panel open/close; tab selection lives inside the panel
  - When only one tab is available (e.g. `mo.ui.dataframe` which has `showColumnExplorer=false`), renders without a tab bar
  - Tab type persists when toggling the panel on/off
- Minor UI improvements on side bar
  - row viewer search textbox is visually consistent with column explorer search
  - "select/deselect row" text is replaced with a checkbox
  - Better formatting for top bar (pin, focus, close btn)

https://github.com/user-attachments/assets/b4bf1e19-3e78-4bfd-9f98-e5de5a83c6b2

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,22 +1,21 @@
 ## 📝 Summary
 
 <!--
-Provide a concise summary of what this pull request is addressing.
-
 If this PR closes any issues, list them here by number (e.g., Closes #123).
+
+Detail the specific changes made in this pull request. Explain the problem addressed and how it was resolved. If applicable, provide before and after comparisons, screenshots, or any relevant details to help reviewers understand the changes easily.
 -->
 Closes #
 
-## 🔍 Description of Changes
+## 📋 Pre-Review Checklist
+<!-- These checks need to be completed before a PR is reviewed -->
 
-<!--
-Detail the specific changes made in this pull request. Explain the problem addressed and how it was resolved. If applicable, provide before and after comparisons, screenshots, or any relevant details to help reviewers understand the changes easily.
--->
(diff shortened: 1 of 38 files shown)


Light2Dark:
1. Could we move the tab selection to the area at the top? And probably make it much less prominent. It's a bit too big and has too much padding.

image

2. The powerful thing about column explorer is that you can click on the columns to view charts & summary stats. Hence why it was called column explorer. Could we communicate to the user that they can click on the column? 

3. Maybe reduce the size of the `select` text, and slightly reduce the size of the select checkbox. Might be hard, I wouldn't spend too much time on it.

4. Looks like some things are squished when the panel is too small. We could increase the min-width of the panel/even get rid of the `<<` and `>>` buttons I think

image

kirangadhave:
> 1. Could we move the tab selection to the area at the top? And probably make it much less prominent. It's a bit too big and has too much padding.

I like this. I have this implemented:
Screenshot 2026-03-30 at 10 49 46 AM

> 2. The powerful thing about column explorer is that you can click on the columns to view charts & summary stats. Hence why it was called column explorer. Could we communicate to the user that they can click on the column?

How about we expand the first column by default? and also add accordion like expand/collapse icons to the left side of each column header?

> 3. Maybe reduce the size of the `select` text, and slightly reduce the size of the select checkbox. Might be hard, I wouldn't spend too much time on it.

I actually increased the size, so this should be easy to undo

> 4. Looks like some things are squished when the panel is too small. We could increase the min-width of the panel/even get rid of the `<<` and `>>` buttons I think

i did increase the min-width, I can do it a few more points.

PR #8911, opened 2026-03-28
fix(ai): buffer streaming preamble text before code fence appears

### LINKED ISSUE #8880: AI cell generation emits conversational text as Python code
### Describe the bug

When using "Generate AI cell" or "Inline AI edit", some models (e.g. Haiku 4.5) emit conversational text before or instead of code fences. This causes two problems:

1. **Preamble text as a Python cell (Generate AI cell):** The preamble text (e.g. "I'll create a cell that...") gets created as a separate Python cell containing plain English, resulting in two staged cells instead of one.

2. **Mixed text and code in a single cell (Inline AI edit):** The AI's full response — preamble text, code, and trailing conversational text (e.g. clarifying questions) — all land in a single cell as-is, which fails to execute.

See the attached screenshots for examples of both cases.

**Hypothesized root cause (via Claude):** The streaming parser treats the buffer as raw Python code when no code fences are present yet, so early conversational chunks become a Python cell. For inline edits, the fence-stripping logic only activates when the very first streamed chunk starts with backticks — any conversational preamble causes the entire response to pass through unmodified.

**Steps to reproduce:**
1. Use "Generate AI cell" or "Inline AI edit" with a model prone to conversational preamble (e.g. Haiku 4.5)
2. Use any prompt (e.g. "fibonacci" or "create a function that reverses strings")
3. Observe either a preamble text cell above the code cell, or a single cell mixing prose and code

Image

Image

Image

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.21.1",
  "editable": false,
  "location": "/Users/kyrre.wahl.kongsgaard/projects/notebooks/.venv/lib/python3.12/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.3.0",
  "Processor": "arm",
  "Python Version": "3.12.10",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "146.0.7680.165",
    "Node": "v24.10.0",
    "uv": "0.9.7 (0adb44480 2025-10-30)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.18.1",
    "packaging": "25.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.42.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "anywidget": "0.9.21",
    "duckdb": "1.5.1",
    "ibis-framework": "12.0.0",
    "loro": "1.10.3",
    "mcp": "1.26.0",
    "nbformat": "5.10.4",
    "openai": "2.29.0",
    "pandas": "3.0.1",
    "polars": "1.39.3",
    "pyarrow": "23.0.1",
    "pytest": "9.0.2",
    "python-lsp-ruff": "2.3.0",
    "ruff": "0.15.7",
    "sqlglot": "30.0.3",
    "watchdog": "6.0.0"
  },
  "Experimental Flags": {
    "chat_modes": true,
    "external_agents": true,
    "mcp_docs": true,
    "storage_inspector": false,
    "chat_sidebar": true
  }
}
```

</details>

### Code to reproduce

_No response_

temrjan:
## Summary

- Buffer incoming AI stream chunks in `CellCreationStream` until a code fence (` ``` `) appears, preventing conversational preamble from becoming a separate Python cell
- Flush buffered content as a cell in `stop()` if the stream ends without any fence (backward compatibility for models that return code without fences)
- Add tests for: preamble + fence, code without fence, fence from first chunk

## Root cause

`CellCreationStream.stream()` called `codeToCells(buffer)` on every chunk. Before any fence arrived, `codeToCells` treated the entire buffer as Python code (line 166 of `completion-utils.ts`: `if (!code.includes("```")) return [{ language: "python", code }]`), creating a cell from preamble text like "I'll create a fibonacci function...".

## What this does NOT fix

The "Inline AI edit" flow (problem #2 in the issue) uses a different code path through the backend `without_wrapping_backticks` function. That is a separate issue.

## Test plan

- [x] All 25 staged-cells tests pass
- [x] TypeScript clean (no new errors)
- [x] Test: preamble text buffered, cell created only when fence arrives
- [x] Test: code without fence → cell created on stream end (backward compat)
- [x] Test: fence in first chunk → cell created immediately (no delay)

Fixes #8880

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/frontend/src/core/ai/__tests__/staged-cells.test.ts
+++ b/frontend/src/core/ai/__tests__/staged-cells.test.ts
@@ -419,7 +419,7 @@ describe("onStream", () => {
     expect(mockUpdateCellEditor).not.toHaveBeenCalled();
   });
 
-  it("should create cells when text-delta is received and a stream has been created", () => {
+  it("should buffer text without fences and create cell on stop", () => {
     const { result } = renderHook(() => useStagedCells(store));
     result.current.onStream({ type: "text-start", id: "test-id" });
 
@@ -433,6 +433,12 @@ describe("onStream", () => {
       delta: "some code",
     });
 
+    // Cell should NOT be created yet — waiting for a fence or stream end
+    expect(mockCreateNewCell).not.toHaveBeenCalled();
+
+    // When the stream ends, the buffered code is flushed as a cell
+    result.current.onStream({ type: "text-end", id: "test-id" });
+
     expect(mockCreateNewCell).toHaveBeenCalledWith({
       cellId: "__end__",
       code: "some code",
@@ -441,7 +447,50 @@ describe("onStream", () => {
     });
   });
 
-  it("should handle delta chunks", () => {
+  it("should not create cell from preamble when fence appears later", () => {
+    const { result } = renderHook(() => useStagedCells(store));
+    result.current.onStream({ type: "text-start", id: "test-id" });
+
+    const mockCellId = cellId("mock-cell-id");
+    vi.mocked(CellId.create).mockReturnValue(mockCellId);
+
+    // Preamble text without fence — should be buffered
+    result.current.onStream({
+      type: "text-delta",
+      id: "test-id",
+      delta: "I'll create a fibonacci function.\n\n",
+    });
+
+    expect(mockCreateNewCell).not.toHaveBeenCalled();
+
+    // Now fence arrives — cell should be created with only the code
+    result.current.onStream({
+      type: "text-delta",
+      id: "test-id",
+      delta: "```python\nsome code",
+    });
+
+    expect(mockCreateNewCell).toHaveBeenCalledWith({
+      cellId: "__end__",
+      code: "some code",
+      before: false,
+      newCellId: "mock-cell-id",
+    });
+
+    // More code arrives — cell should be updated
+    result.current.onStream({
+      type: "text-delta",
+      id: "test-id",
+      delta: "\nmore code\n```",
+    });
+
+    expect(vi.mocked(updateEditorCodeFromPython)).toHaveBeenCalledWith(
+      mockCellHandle.current.editorViewOrNull,
+      "some code\nmore code",
+    );
+  });
+
+  it("should handle delta chunks with fences", () => {
     const { result } = renderHook(() => useStagedCells(store));
     result.current.onStream({ type: "text-start", id: "test-id" });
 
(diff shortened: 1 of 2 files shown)


temrjan:

frontend/src/core/ai/__tests__/staged-cells.test.ts
```diff
@@ -441,37 +447,68 @@ describe("onStream", () => {
     });
   });
 
-  it("should handle delta chunks", () => {
```
manzt, line 444: This test covered the case where the first chunk is a partial fence, which got replaced. Would we add that scenario back?

manzt (commented):
Nice. I _think_ buffering until a fence appears makes sense, and the `stop()` fallback for fence-less responses is a good backward-compat touch.

frontend/src/core/ai/__tests__/staged-cells.test.ts
temrjan: Added! The new test (`should buffer partial fence and create cell when fence completes`) covers the split-fence scenario: first chunk has just ``````, second chunk completes the fence — no premature cell creation.

frontend/src/core/ai/staged-cells.ts
```diff
@@ -282,6 +291,15 @@ class CellCreationStream {
...
+    // (model returned code without fences), create a cell from the buffer.
+    if (this.buffer.trim() && this.createdCells.length === 0) {
+      const completionCells = codeToCells(this.buffer);
+      for (const cell of completionCells) {
+        const newCellId = this.onCreateCell(cell.code);
+        this.createdCells.push({ cellId: newCellId, cell });
+      }
+    }
     // Clear all state
     this.buffer = "";
```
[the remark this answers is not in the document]
temrjan: Good catch — applied in 2d784fc. `CellCreationStream` is single-use (new instance per `text-start`), so this is purely defensive, but it makes the comment accurate.

Light2Dark:
needs some manual testing as I couldn't reproduce this issue.

temrjan:
Hi @Light2Dark, thanks for the review!

Issue #8880 actually describes **two distinct bugs**:

1. **Generate AI cell** — conversational preamble becomes its own Python cell.
2. **Inline AI edit** — preamble, code, and trailing text all land mixed in one cell.

This PR addresses only Bug 1 (frontend-only change to `staged-cells.ts`). Bug 2 goes through a different code path (`without_wrapping_backticks` in `marimo/_server/ai/providers.py`) and would need a separate fix. Worth noting that Inline AI edit would still exhibit Bug 2's symptom — the two look similar but stem from different code paths.

### Repro for Bug 1

1. Configure **Anthropic Haiku 4.5** as the AI model (any model that emits conversational preamble works)
2. Use **Generate AI cell** (not Inline AI edit)
3. Prompt: `fibonacci`

**On `main`:** two staged cells — one with prose, one with code.
**On this branch:** one staged cell with only the code; the preamble is buffered until a fence appears, then discarded.

The screenshots in #8880 show the exact symptom on `main`.

### Validation

- 4 unit tests cover the buffering paths (preamble + fence, no-fence flushing, fence in first chunk, split fence). CI is green.
- @kyrre (original reporter) — would you be able to confirm this PR resolves Bug 1 in your setup?

Happy to record a side-by-side GIF (`main` vs. this branch) if it would help.

kirangadhave:
@cubic-dev-ai

PR #8915, opened 2026-03-29
fix: style_cell and hover_template lost when sorting in descending order

### LINKED ISSUE #8847: `style_cell` styles lost when sorting table in descending order

ManasVardhan:
## Summary

Fixes #8847. `style_cell` styles (and `hover_template` texts) were lost when sorting a table in descending order.

## Root Cause

`_style_cells()` and `_hover_cells()` generated row IDs assuming they correlated with the sort order:
- Ascending: `range(skip, skip + take)` 
- Descending: `range(total - 1 - skip, ..., -1)`

This only worked when sorting by the index column. When sorting by **any other column**, the actual `_marimo_row_id` values on the displayed page are arbitrary and do not follow sequential or reversed-sequential patterns. The frontend looks up styles by `_marimo_row_id`, so the mismatched keys caused all styles to be silently discarded.

## Fix

Introduces `_get_page_row_ids()` which reads the actual `_marimo_row_id` values from the sorted `_searched_manager` data for the current page. This ensures style/hover dict keys always match what the frontend uses for lookup, regardless of sort column or direction.

For tables without stable row IDs (list/dict data), positional indices are returned since both backend and frontend use positional indexing.

The `descending` parameter is removed from `_style_cells()` and `_hover_cells()` since it is no longer needed.

## Additional fixes
- Removes `xfail` from `test_cell_search_df_hover_texts_sorted` (hover with sorted data now works)
- Fixes pre-existing test bug where `sort` was passed as a bare `SortArgs` instead of `[SortArgs(...)]`

## Tests

- `test_cell_styles_descending_non_index_column`: 4-row table sorted by Score (not index) descending. Verifies all row IDs present with correct style values.
- `test_cell_styles_sorted_by_value_with_pagination`: 20-row table sorted by Value descending with page_size=5. Verifies page 0 contains the correct (non-sequential) row IDs.
- All existing style/hover tests continue to pass.

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -1263,14 +1263,17 @@ class table(
         if response.all_rows:
             if self._has_stable_row_id:
                 try:
-                    # Slice to the requested page via take(), then collect
-                    # to an eager frame. This avoids materializing the full
+                    # Slice to the requested page via take(), then read
+                    # the index column. This avoids materializing the full
                     # index column and handles lazy backends (polars
                     # LazyFrame, duckdb, ibis) that do not support direct
                     # slice indexing on the data attribute.
-                    page_data = self._searched_manager.take(
-                        take, skip
-                    ).as_frame()
+                    page_manager = self._searched_manager.take(take, skip)
+                    page_frame = getattr(page_manager, "as_frame", None)
+                    if page_frame is not None:
+                        page_data = page_frame()
+                    else:
+                        page_data = page_manager.data
                     return cast(
                         list[int],
                         page_data[INDEX_COLUMN_NAME].to_list(),


ManasVardhan:

mscolnick:
@ManasVardhan it looks like tests are failing. mind taking a look?

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1238,12 +1238,43 @@ def _preview_column(self, args: PreviewColumnArgs) -> ColumnPreview:
...
+            if self._has_stable_row_id:
+                try:
+                    all_ids = cast(
+                        list[int],
+                        self._searched_manager.data[
+                            INDEX_COLUMN_NAME
+                        ].to_list(),
+                    )
+                    return all_ids[skip : skip + take]
+                except Exception:
```
mscolnick, line 1281: worth `LOGGER.warning()` here?

mscolnick (dismissed):

mscolnick:
@ManasVardhan sorry for the back and forth. looks like lint it failing. lmk if you want to handle or want us to

marimo/_plugins/ui/_impl/table.py
ManasVardhan: Already there -- the `except` block logs `LOGGER.warning("Failed to read %s from searched manager; falling back to positional indices", INDEX_COLUMN_NAME)` right below. Might be collapsed in the diff view.

ManasVardhan:
hi @mscolnick , should be good now.

mscolnick:
looks like a bad test that is irrelevant. this looks great, thank you!

Issue #8324, opened 2026-02-15
Form Submit button, Ctrl+Enter, and Shift+Enter all have different behavior


maxTarlov:
### Describe the bug

If you have a text input inside of a form with `clear_on_submit=True`, neither Ctrl+Enter nor Shift+Enter trigger the form to clear. Furthermore,  Shift+Enter bypasses the form validation callback and form on_change callback. "Submit" runs the text input's on_change handler twice

<img width="687" height="817" alt="Image" src="https://github.com/user-attachments/assets/086c82c6-7492-48bf-91d8-9203c5473543" />

### Will you submit a PR?

- [ ] Yes

### Environment

Reproduced in a molab notebook Feb 15, 2026

<details>

```
{
  "marimo": "0.19.11",
  "editable": false,
  "location": "/usr/local/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "4.4.0",
  "Processor": "",
  "Python Version": "3.13.11",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.13.1",
    "uv": "0.10.2"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.16.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.20.1",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.40.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "anywidget": "0.9.21",
    "duckdb": "1.4.4",
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "openai": "2.20.0",
    "pandas": "3.0.0",
    "polars": "1.38.1",
    "pyarrow": "23.0.0",
    "pytest": "9.0.2",
    "python-lsp-ruff": "2.3.0",
    "python-lsp-server": "1.14.0",
    "ruff": "0.15.0",
    "sqlglot": "28.6.0"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```
# /// script
# requires-python = ">=3.13"
[...]

PR #8918, opened 2026-03-29
fix(form): route Ctrl+Enter through onSubmit and stop double-firing on_change

Mystery01092000:
## Summary

Fixes three related submit-inconsistency bugs in `mo.ui.form` reported in #8324.

### Bug 1 — `Ctrl+Enter` bypassed `onSubmit`

`FormPlugin.tsx` called `setValue(newValue)` directly in the `onKeyDown` handler, skipping the `<form>`'s `onSubmit`. This meant `shouldValidate` and `clearOnSubmit` were silently ignored when using the keyboard shortcut.

**Fix:** Replace `setValue(newValue)` with `formDiv.current?.requestSubmit()`, which routes the keyboard shortcut through the same `onSubmit` path as the button click.

### Bug 2 — `Shift+Enter` bypassed the form wrapper

`OnBlurredInput`'s `onKeyDown` in `input.tsx` fired `setValue` on any `Enter` keypress with no modifier check. `Shift+Enter` inside a wrapped text input would commit the input value directly, completely bypassing the form's buffering layer.

**Fix:** Add a `shiftKey` guard — only plain `Enter` (no `Shift`) commits the value.

### Bug 3 — Wrapped element's `on_change` fired twice per submission

`form._convert_value()` called `element._update(value)`, which always triggers the element's `_on_change` callback. This caused the wrapped element's `on_change` to fire unexpectedly during form submission (in addition to the form's own `on_change`).

**Fix:** Add `UIElement._update_value()` — identical to `_update()` but without the `_on_change` call. `form._convert_value()` now calls `_update_value()` so the wrapped element's `on_change` only fires when the element itself is interacted with, not when the form extracts its submitted value.

## Files changed

| File | Change |
|---|---|
| `frontend/src/plugins/impl/FormPlugin.tsx` | `requestSubmit()` instead of direct `setValue()` in `onKeyDown` |
| `frontend/src/components/ui/input.tsx` | `shiftKey` guard in `OnBlurredInput.onKeyDown` |
| `marimo/_plugins/ui/_core/ui_element.py` | New `_update_value()` method (silent update, no callback) |
| `marimo/_plugins/ui/_impl/input.py` | `form._convert_value` uses `_update_value` not `_update` |
| `frontend/src/plugins/impl/__tests__/FormPlugin.test.tsx` | New test file covering all three fixes |
| `tests/_plugins/ui/_impl/test_input.py` | Three new regression tests for the Python-side fix |

## Test plan

- [ ] `Ctrl+Enter` triggers validation when `shouldValidate=True`
- [ ] `Ctrl+Enter` clears inputs when `clearOnSubmit=True`
- [ ] `Shift+Enter` inside a text input does not submit the form
- [ ] Submit button calls `setValue` exactly once
- [ ] Wrapped element's `on_change` does not fire on form submission
- [ ] Form's `on_change` fires exactly once per submission
- [ ] `_update_value()` updates value without triggering `on_change`

---

> **Note:** Please let me know if any adjustments are needed to match the project's style or conventions. I'm happy to iterate.

### DIFF
--- a/frontend/src/components/ui/input.tsx
+++ b/frontend/src/components/ui/input.tsx
@@ -195,7 +195,10 @@ export const OnBlurredInput = React.forwardRef<
       onChange={(event) => setInternalValue(event.target.value)}
       onBlur={() => setValue(internalValue || "")}
       onKeyDown={(event) => {
-        if (event.key !== "Enter") {
+        // Only commit on plain Enter. Shift+Enter should not submit the
+        // enclosing form — let the event bubble naturally (e.g. newline in a
+        // textarea, or no-op in a single-line input).
+        if (event.key !== "Enter" || event.shiftKey) {
           return;
         }
         setValue(internalValue || "");
--- a/frontend/src/plugins/impl/FormPlugin.tsx
+++ b/frontend/src/plugins/impl/FormPlugin.tsx
@@ -149,11 +149,13 @@ export const FormWrapper = <T,>({
           "bg-(--amber-1) border-(--amber-7)": !synchronized && bordered,
         })}
         onKeyDown={(evt) => {
-          // Handle enter + ctrl/meta key
+          // Handle enter + ctrl/meta key — route through the form's onSubmit
+          // so that validation and clearOnSubmit are applied consistently,
+          // just as when the submit button is clicked.
           if (evt.key === "Enter" && (evt.ctrlKey || evt.metaKey)) {
             evt.preventDefault();
             evt.stopPropagation();
-            setValue(newValue);
+            formDiv.current?.requestSubmit();
           }
         }}
       >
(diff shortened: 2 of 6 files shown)


Mystery01092000:

Mystery01092000:
recheck

manzt (changes requested):
I pulled this down and tested locally but wasn't able to verify that Bug 3 (the double-fire) is fixed. Wrapping a `mo.ui.text(on_change=...)` in a form and submitting still fires the element's `on_change`.

Have you verified these change yourself? Could you share a screen recording or a minimal repro snippet showing the correct behavior?

mscolnick:
Going to move to draft while it needs changes.

PR #8931, opened 2026-03-30
improvement: inline audio and video in HTML export

### LINKED ISSUE #8801: option to configure exporting audio in HTML
### Description

There is currently no way to export audio in the HTML and it's very frustrating.

### Suggested solution

1) Add an option to export audio as well as video in the HTML 

2) document what method is used to do the conversion to give user more control over what will be exported (repr_html, anywidget, javascript tags ...) 
It may be outside the scope of this issue, but I'm still struggling to understand how it works.

- [ ] Yes

### Alternatives

_No response_

### Additional context

Context:
I'm currently working on an exploratory notebook on interpretability of audio classifiers and it's really important for me to be able to share very small audio snippets, one second, in the HTML for users to be able to follow along.

mscolnick:
Closes #8801

When exporting a marimo notebook to HTML, `<audio>` and `<video>` elements with virtual file URLs (`./@file/...`) were not embedded as data URIs — only `` tags were. This meant audio clips broke in exported HTML files.

Added `"audio"` and `"video"` to `VIRTUAL_FILE_ALLOWED_TAGS` so they are inlined just like images. Also added a 10MB size cap (`MAX_VIRTUAL_FILE_INLINE_BYTES`) — files exceeding this limit are skipped and should be served from the `public/` folder instead.

### DIFF
--- a/tests/_server/api/test_middleware.py
+++ b/tests/_server/api/test_middleware.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import io
 import json
+import socket
 import sys
 import time
 from multiprocessing import Process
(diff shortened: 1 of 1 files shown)


marimo/_convert/common/dom_traversal.py
```diff
@@ -304,6 +307,26 @@ def replace_virtual_files_with_data_uris(
...
+                        max_inline_bytes,
+                    )
+                    # Return a text placeholder so users see a clear
+                    # message instead of a broken ./@file/ link.
+                    msg = (
+                        f"File too large to inline "
+                        f"({parsed[0]} bytes, limit {max_inline_bytes})"
+                    )
+                    return build_data_url(
+                        "text/plain",
```
dmadisetti, line 327: text or warning/ error?

mscolnick: i dont want to change the layout structure. this only changes the URL

mscolnick:
> Can we have a test that check to see if we dump to public when the file size is exceeded?

We don't dump to public on behalf of the user. Not in scope of this PR

PR #8932, opened 2026-03-30
feat: persistent top toolbar in data table with search, chart-builder, and other actions 

kirangadhave:
## 📝 Summary

Separate actions and navigation on the table UI into top toolbar and bottom bar respectively.

## Changes
- Added a new top bar
  - Moved search to the top bar. Shows up as a magnifying glass and expands into textbox on click
  - Renamed `Download` to `Export` and replaced with an icon for consistency
  - Flattened the export menu into download and copy without sub-menus
    - Uniform format descriptions in the menu
- Bottom bar
  - row/col count on left
  - pagination in the middle
  - stats on the right. When no cells are selected shows a message indicating cells can be selected.
- Pagination
  - removed border from buttons for a cleaner look
  - page selector also accepts numeric input
- Dataframe displays
  - Raw dataframes now use `pagination=None` instead of `True`, matching `mo.ui.Table` defaults. The auto detection logic runs on `pagination=False` and hides pagination control if there is only 1 page.
- Renames
  -  Renamed download and other  related files to export
  - Renamed TableActions to TableBottomBar

https://github.com/user-attachments/assets/c0943c16-0491-44ae-93a6-bfa9d1cee7e2

### DIFF
--- a/frontend/src/components/data-table/TableBottomBar.tsx
+++ b/frontend/src/components/data-table/TableBottomBar.tsx
@@ -143,11 +143,12 @@ export const TableBottomBar = <TData,>({
   };
 
   return (
-    <div className="flex items-center shrink-0 pt-1">
-      <div className="flex-1 text-sm text-muted-foreground px-2 flex items-center gap-1">
-        {renderTotal()}
+    <div className="flex lg:grid lg:grid-cols-[1fr_auto_1fr] items-center shrink-0 pt-1">
+      <div className="flex flex-col text-sm text-muted-foreground px-2 shrink-0">
+        <div className="flex items-center gap-1">{renderTotal()}</div>
+        <CellSelectionStats table={table} className="lg:hidden" />
       </div>
-      <div className="flex items-center justify-center">
+      <div className="ml-auto lg:ml-0 lg:justify-self-center flex items-center shrink-0">
         {pagination && (
           <DataTablePagination
             table={table}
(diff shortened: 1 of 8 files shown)


kirangadhave:
> * The "_Select multiple cells ..._" text feels heavy handed. Could it be in a tooltip?

The main purpose of this is to indicate that cells can be selected

kirangadhave:
> * A possible solution could be to left align the page navigator (put it just to the right of the "rows, columns" text)

An issue with this is that we have a `Clear Selection` text that appears when we select some rows. This will cause the page navigator to be pushed out of the way. I will make the navigator more compact and add a different layout on smaller widths.

Light2Dark:
- I would change the Export icon to something like a Download.
- Small style change is the page size select box and the page number select box can be a little taller now. We made them smaller since there were many icons before I think.

I remember design suggesting this, maybe can take some inspiration? (not all). We should ask them for small screen designs too haha

image

kirangadhave:
@akshayka I have addressed some of your comments here:

> The horizontal space between the icons in the top bar is a little disorienting to me. I like to have all actions in one place.

The search box is now always open and all other actions are next to it. The actions also have labels

Screenshot 2026-03-31 at 12 33 09 AM

> The "Select multiple cells ..." text feels heavy handed. Could it be in a tooltip?

Switched the text to `No Selection` with a tooltip showing complete message
Screenshot 2026-03-31 at 12 34 31 AM

> The stats text overlaps with the page navigator in compact view, see image. 

Made the table responsive, on smaller widths the layout shifts.

Regular:
Screenshot 2026-03-31 at 12 35 39 AM

Smaller screens:
Screenshot 2026-03-31 at 12 37 50 AM

I have another ticket where I want to rethink the responsive behavior more widely, we can revisit the table responsiveness then.

PR #8969, opened 2026-04-01
fix altair not setting to full-width when width: container

Light2Dark:
## 📝 Summary

The existing data-container-width attribute approach only checks for width at the top level of the spec.
There are cases where that may be missing. 

- Use vega-embed's own fit-x class detection and add the width: 100%
- Fix nested/concat charts not setting to full-width

image

### unchanged
image

image

image

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/frontend/src/plugins/impl/vega/vega.css
+++ b/frontend/src/plugins/impl/vega/vega.css
@@ -5,7 +5,7 @@
 }
 
 .vega-embed[data-container-width="container"],
-.vega-embed:has(.chart-wrapper.fit-x) {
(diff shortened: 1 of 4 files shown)


frontend/src/plugins/impl/vega/vega.css
Light2Dark, line 134: we do this here too, these are default styles we ported over

Light2Dark:
I added in the smoke test if we manually did fit-x, and it was fine.

PR #8993, opened 2026-04-02
Decouple skew protection from agent-facing endpoints

manzt:
Skew protection exists to prevent frontend/server version mismatch, but it was also gating agent-only functionality: the `/api/kernel/execute` endpoint, `/api/sessions` listing, and local server discovery. Agents don't go through the frontend, so the skew token requirement doesn't apply to them.

The skew protection middleware now skips `/api/kernel/execute`. The `/api/sessions` endpoint drops its extra `enable_auth` and `skew_protection` guards since `@requires("edit")` already handles auth. Server registry discovery still requires `--no-token` because filesystem-based discovery has no auth layer, but no longer requires `--no-skew-protection`.

### DIFF
--- a/marimo/_server/api/middleware.py
+++ b/marimo/_server/api/middleware.py
@@ -137,7 +137,7 @@ class SkewProtectionMiddleware:
         ):
             return await self.app(scope, receive, send)
         # If /api/kernel/execute, skip (agent-only endpoint)
-        if request.url.path.endswith("/api/kernel/execute"):
+        if request.url.path.rstrip("/").endswith("/api/kernel/execute"):
             return await self.app(scope, receive, send)
         # If ws, skip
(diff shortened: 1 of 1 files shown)


marimo/_server/api/middleware.py
```diff
@@ -136,6 +136,9 @@ async def __call__(
             "application/x-www-form-urlencoded"
         ):
             return await self.app(scope, receive, send)
+        # If /api/kernel/execute, skip (agent-only endpoint)
+        if request.url.path.endswith("/api/kernel/execute"):
```
manzt:
```suggestion
        if request.url.path.rstrip("/").endswith("/api/kernel/execute"):
```

mscolnick (dismissed):

PR #8998, opened 2026-04-02
feat: indicate server side installations

dmadisetti:
image


dmadisetti:
@mscolnick  It only makes a difference in sandbox mode because an installation on say `nbformat` w/o sandbox will be importable by the kernel (same venv)

marimo/_schemas/generated/notifications.yaml
```diff
@@ -2,9 +2,9 @@
 components:
   schemas:
     AlertNotification:
-      description: "User-facing alert message.\n\nAttributes:\n    title: Alert title.\n\
```
manzt, line 5: Some odd churn here in the schemas. I think this can come from generating without the specific python version? I think the preferred way is to `make fe-codegen` (uses Python 3.12).

dmadisetti: Hmm, this was from `make fe-codegen`. Committing because it only looks like whitespace and thrashing won't be awful

PR #9001, opened 2026-04-02
chore: migrate eslint/biome ignore to oxlint ignore, turn off some rules

kirangadhave:
Bought down warnings from 12k -> ~280.
turned off some overly strict rules coming from `correctness` category in oxlint config

### DIFF
--- a/frontend/.oxlintrc.json
+++ b/frontend/.oxlintrc.json
@@ -26,7 +26,7 @@
     // === Marimo custom rules ===
     "marimo/add-event-listener-object": "error",
     "marimo/remove-event-listener-object": "error",
-    "marimo/prefer-object-params": "off",
+    "marimo/prefer-object-params": "warn",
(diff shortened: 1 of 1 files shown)


frontend/.oxlintrc.json
```diff
@@ -26,16 +26,38 @@
     // === Marimo custom rules ===
     "marimo/add-event-listener-object": "error",
     "marimo/remove-event-listener-object": "error",
-    "marimo/prefer-object-params": "warn",
+    "marimo/prefer-object-params": "off",
```
mscolnick: this is useful as a warning still. it can show up in the IDE and people can choose to fix or ignore it. im sure it gives a lot of existing issue, but at least stops the bleeding (while we can slowing heal the wounds)

kirangadhave: hmm there are a bunch of these in our code base, so i turned it off.  Definitely see the value in it though.

PR #9010, opened 2026-04-03
feat(plotly): add reactive box plot selection support

axsseldz:
### ## 📝 Summary

Makes `go.Box` (and `px.box`) plots fully selection-reactive in `mo.ui.plotly`. Users can now drag a box/lasso, click a box body, or click individual jittered points to get row-level data in Python.

## 🔍 Description of Changes

- **Frontend:** added `type === "box"` to `shouldHandleClickSelection` so box click events are forwarded to Python.
- **Python:** added `_append_box_points_to_selection` which handles three cases: range/lasso with individual points already sent by Plotly, range/lasso without points (extract from figure data), and click events (expand `pointNumbers` into individual rows). Supports vertical/horizontal orientation and categorical axes.
- **Tests:** 12 Python tests + 1 frontend test covering all selection modes.
- **Example:** `examples/third_party/plotly/box_chart.py` — three chart variants (single-trace, grouped, horizontal) each with a live stats summary and `mo.ui.table`.

## Selection
box1

## Click Selection
box2

## 📋 Pre-Review Checklist

- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it.
- [x] Video or media evidence is provided for any visual changes (optional). 

## ✅ Merge Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [x] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Tests have been added for the changes made.

@mscolnick @nojaf

PR #9011, opened 2026-04-03
feat(plotly): add reactive violin plot selection support

axsseldz:
## 📝 Summary

Makes `go.Violin` / `px.violin` plots selection-reactive in `mo.ui.plotly`.

## 🔍 Description of Changes

- **Frontend:** added `type === "violin"` to `shouldHandleClickSelection`.
- **Python:** added `_append_violin_points_to_selection`. Handles vertical/horizontal orientations and categorical axes.
- **Tests:** 11 Python tests + 1 frontend test.
- **Example:** `examples/third_party/plotly/violin_chart.py` — single-trace, grouped (`px.violin`), and horizontal variants with live stats and `mo.ui.table`.

## Box Selection
violin1

## Click Selection
violin2

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/frontend/src/plugins/impl/plotly/__tests__/PlotlyPlugin.test.tsx
+++ b/frontend/src/plugins/impl/plotly/__tests__/PlotlyPlugin.test.tsx
@@ -152,8 +152,13 @@ describe("PlotlyPlugin", () => {
 
     expect(setValue).toHaveBeenCalledTimes(1);
     const updater = setValue.mock.calls[0][0] as (value: unknown) => unknown;
-    const result = updater({}) as { points: unknown[]; indices: unknown[] };
-    expect(result.points).toHaveLength(1);
-    expect(result.indices).toHaveLength(1);
+    expect(updater({})).toEqual({
+      selections: [],
+      points: [
+        { x: "Group A", y: 3, curveNumber: 0, pointNumber: 0, pointIndex: 0 },
+      ],
+      indices: [0],
+      range: undefined,
+    });
   });
 });
--- a/marimo/_plugins/ui/_impl/plotly.py
+++ b/marimo/_plugins/ui/_impl/plotly.py
@@ -2460,10 +2460,12 @@ def _extract_violin_points_from_range(
 ) -> list[dict[str, Any]]:
     """Extract violin plot underlying data points that fall within a selection range.
 
-    For each violin trace, the category position (x for vertical, y for
-    horizontal) is compared against the selection range.  All sample rows
-    belonging to a selected violin group are returned regardless of their
-    individual value.
+    For each violin trace, both the category axis (x for vertical, y for
+    horizontal) and the value axis are compared against the selection range.
+    Only sample rows whose category group overlaps the selection *and* whose
+    individual value falls within the value-axis bounds are returned.  When the
+    value-axis range is absent from ``range_data`` (e.g. a selection spanning
+    the full y-axis) all rows in matching categories are included.
     """
     if DependencyManager.numpy.has():
         return _extract_violin_points_numpy(figure, range_data)
@@ -2473,7 +2475,14 @@ def _extract_violin_points_from_range(
 def _extract_violin_points_numpy(
     figure: go.Figure, range_data: dict[str, Any]
 ) -> list[dict[str, Any]]:
-    """Extract violin plot data points from a selection range using numpy."""
+    """Extract violin plot data points from a selection range using numpy.
+
+    Filters by both the category axis (which violin group falls inside the
+    selection box) and the value axis (which individual sample values fall
+    inside the selection box).  When ``points`` is disabled in the Plotly
+    figure, Plotly does not send individual point coordinates, so we derive
+    inclusion from the underlying trace arrays.
+    """
     import numpy as np
 
     x_range = range_data.get("x")
@@ -2488,10 +2497,12 @@ def _extract_violin_points_numpy(
         orientation = getattr(trace, "orientation", "v") or "v"
         if orientation == "h":
             cat_range = y_range
+            val_range = x_range
             cat_data = getattr(trace, "y", None)
             val_data = getattr(trace, "x", None)
         else:
             cat_range = x_range
+            val_range = y_range
             cat_data = getattr(trace, "x", None)
             val_data = getattr(trace, "y", None)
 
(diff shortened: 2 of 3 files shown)


Light2Dark:
Thanks @axsseldz , will you be refactoring the logic later on? I feel some of the PRs share similar logic
I.e, the logic for box and violin plots (https://github.com/marimo-team/marimo/pull/9010) looks similar.

Would you prefer if we merge one, and then you can work on derivatives (like this) incrementally?

axsseldz:
> Thanks @axsseldz , will you be refactoring the logic later on? I feel some of the PRs share similar logic I.e, the logic for box and violin plots (#9010) looks similar.
> 
> Would you prefer if we merge one, and then you can work on derivatives (like this) incrementally?

sounds good 👍 merging one first works, then i’ll iterate on the derivatives

marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -1955,6 +1969,423 @@ def _build_bar_point(
...
+                    [_safe_category_get(seen_order, v, -1) for v in cat_arr],
+                    dtype=np.float64,
+                )
+                cat_mask = (cat_max > positions - 0.5) & (
+                    cat_min < positions + 0.5
+                )
+        else:
+            cat_mask = np.ones(n, dtype=bool)
+
+        selected_indices = np.where(cat_mask)[0]
```
Light2Dark, line 2577: **No value-range filtering in fallback extraction.** Both `_extract_violin_points_numpy` and `_extract_violin_points_fallback` only filter by category position — the value axis range is ignored entirely. If a user draws a tight box over just the top half of a violin (with `points` disabled), they'll get back **all** rows in that category, not just those within the y-range.

This is arguably acceptable when `points` is disabled since there are no visible individual points to select, but it could surprise users. At minimum, the docstring should document this as intentional behavior. Ideally, when both axes are present in `range_data`, rows should also be filtered by the value dimension.

marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -1955,6 +1969,423 @@ def _build_bar_point(
...
+        if cat_len == len(getattr(trace, val_key, [])):
+            cat_value = _get_indexed_value(cat_source, point_idx)
+        else:
+            indexed = _get_indexed_value(cat_source, 0)
+            cat_value = indexed if indexed is not None else cat_source
+
+    point: dict[str, Any] = {
+        val_key: val_value,
+        "pointIndex": point_idx,
+        "curveNumber": trace_idx,
```
Light2Dark, line 2754: **Missing `pointNumber` for consistency.** The existing `_build_bar_point` includes both `pointIndex` and `pointNumber`:

```python
point: dict[str, Any] = {
    "x": x_value,
    "y": y_value,
    "curveNumber": trace_idx,
    "pointIndex": int(point_idx),
    "pointNumber": int(point_idx),  # <-- present in bar
}
```

But `_build_violin_sample_point` only includes `pointIndex`. Since `_get_selection_point_id` falls back to `pointNumber` when `pointIndex` is absent, this inconsistency could cause dedup failures in mixed-trace figures (e.g., violin + bar). Consider adding `"pointNumber": point_idx` here.

marimo/_plugins/ui/_impl/plotly.py
```diff
@@ -1955,6 +1969,423 @@ def _build_bar_point(
...
+            # Plotly already sent the individual selected data points because
+            # points is enabled.  Use them as-is; do NOT re-extract from the
+            # range (which can fail on categorical axes and would discard richer
+            # hovertemplate fields like custom ids).
+            clean_points = existing_violin + existing_non_violin
+            # Preserve incoming indices from the frontend payload rather than
+            # recomputing from pointIndex only — Plotly may send pointNumber
+            # instead. Map each point object to its original index by identity.
+            incoming_index_map = {
+                id(p): idx for idx, p in zip(all_indices, all_points) if p
```
Light2Dark, line 2333: **Nit: `id(p)` identity mapping is fragile.** This works only because the point dicts come directly from the frontend payload and haven't been copied. If any middleware ever clones or reconstructs these dicts (e.g., serialization round-trip, defensive copy), this mapping silently breaks and falls through to the `_get_selection_point_id` fallback.

A positional index mapping would be more robust:
```python
incoming_index_map = {
    i: idx for i, (idx, p) in enumerate(zip(all_indices, all_points)) if p
}
```
Then look up by position in `clean_points` using the original list index. Not blocking, but worth noting.

frontend/src/plugins/impl/plotly/__tests__/PlotlyPlugin.test.tsx
```diff
@@ -111,4 +111,49 @@ describe("PlotlyPlugin", () => {
...
+            curveNumber: 0,
+          },
+        ],
+      });
+    });
+
+    expect(setValue).toHaveBeenCalledTimes(1);
+    const updater = setValue.mock.calls[0][0] as (value: unknown) => unknown;
+    const result = updater({}) as { points: unknown[]; indices: unknown[] };
+    expect(result.points).toHaveLength(1);
```
Light2Dark: **Test assertion is less strict than the bar test above it.** The bar test on line 99 uses `expect(updater({})).toEqual({...})` to assert the exact payload shape. This violin test only checks `toHaveLength`, which would miss regressions in the actual point content. Consider matching the bar test's pattern:

```typescript
expect(result).toEqual({
  selections: [],
  points: [
    { x: "Group A", y: 3, curveNumber: 0, pointNumber: 0, pointIndex: 0 },
  ],
  indices: [0],
  range: undefined,
});
```

tests/_plugins/ui/_impl/test_plotly.py
```diff
@@ -3012,3 +3014,232 @@ def test_bar_chart_preserves_indices_when_empty_points_exist() -> None:
 
     assert result == [{"x": "B", "y": 20, "curveNumber": 0, "pointIndex": 1}]
     assert plot.indices == [1]
+
+
+# ============================================================================
+# Violin Plot Tests
```
Light2Dark, line 3340: **Some test gaps to consider:**

- **Mixed traces**: No test for a figure containing violin + scatter (or violin + bar). Since `_convert_value` processes all trace type handlers sequentially and they mutate `selection_data` in place, interaction between handlers matters.
- **Lasso selection**: The code checks `isinstance(lasso_value, dict)` but no test exercises this path for violin traces.
- **Empty `pointNumbers`**: What happens if `pointNumbers` is `[]`? The click path would skip expansion but also not append the original point — worth verifying.

Light2Dark (commented):
**Review from Shahmir's Cursor-guided agent.** 

If these are inaccurate, please ignore. Otherwise, I think the code looks fine. Testing something new - Shahmir

PR #9024, opened 2026-04-03
tests: speed up slow snippet tests

mscolnick:
### `tests/conftest.py`
New session-scoped `default_snippets` fixture that calls `read_snippets()` once
per session instead of independently in each snippet-related test.

### `tests/_snippets/test_snippets.py`
Uses `default_snippets` fixture instead of calling `read_snippets()` directly.

### `tests/_pyodide/test_pyodide_session.py`
Mocks `read_snippets` with the `default_snippets` fixture in the bridge test.

### DIFF
--- a/marimo/_save/loaders/lazy.py
+++ b/marimo/_save/loaders/lazy.py
@@ -33,8 +33,6 @@ from marimo._save.stubs.lazy_stub import (
 
 LOGGER = _loggers.marimo_logger()
 
-_MISSING = object()
-
 
 def to_item(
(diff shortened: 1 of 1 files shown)


marimo/_save/loaders/lazy.py
```diff
@@ -33,6 +33,8 @@
 
 LOGGER = _loggers.marimo_logger()
 
+_MISSING = object()
```
mscolnick: will remove this file change for another PR

PR #9029, opened 2026-04-03
fix: exit with non-zero code when marimo-pair skill is missing

mscolnick:
The prompt command printed an error when the skill was not found but
continued executing and exited successfully. Now it raises SystemExit(1)
so callers can detect the failure. Also fixes the test assertions to
match the actual error message ("could not be found").

### DIFF
--- a/marimo/_cli/pair/commands.py
+++ b/marimo/_cli/pair/commands.py
@@ -151,7 +151,6 @@ def prompt(
                 "https://github.com/marimo-team/marimo-pair",
                 err=True,
             )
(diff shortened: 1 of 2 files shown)


marimo/_cli/pair/commands.py
```diff
@@ -151,6 +151,7 @@ def prompt(
                 "https://github.com/marimo-team/marimo-pair",
                 err=True,
             )
+            raise SystemExit(1)
```
mscolnick: removing this

PR #9033, opened 2026-04-03
fix: gracefully handle delete of unknown cell and read document on reconnect

mscolnick:
Two fixes:
1. Frontend: catch errors when deleting a cell ID not present in any
   column (e.g. kernel-only cells never sent to the UI) so the
   remaining transaction changes still apply instead of crashing.
2. Backend: `_extract_cell_data` now reads from `session.document`
   instead of the cell_manager, so document mutations from
   `notebook-document-transaction` are reflected when clients reconnect.

### DIFF
--- a/marimo/_server/api/endpoints/files.py
+++ b/marimo/_server/api/endpoints/files.py
@@ -169,6 +169,14 @@ async def save(
     session = app_state.require_current_session()
     contents = session.app_file_manager.save(body)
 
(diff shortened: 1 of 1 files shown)


mscolnick:
> Nice, easy fix. Are there other places we should be using _extract_cell_data with document instead?

That was pretty consolidated, but we have plans to remove `cell_manager` which should simplify

PR #9037, opened 2026-04-03
fix: show which cell defines a variable in multiply-defined error

mscolnick:
When code mode validation detects a multiply-defined name, the error
message now identifies the existing cell(s) that already define it,
making it actionable instead of requiring the user to search manually.

### DIFF
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -928,10 +928,14 @@ class AsyncCodeModeContext:
             )
             if new_multiply_defined:
                 # Show which existing cell(s) already define each name.
-                batch_ids = {op.cell_id for op in ops}
+                # Only exclude cells that were actually (re)registered
+                # during this dry run.  Moves and config-only updates
+                # do not register code, so their pre-existing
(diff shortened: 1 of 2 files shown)


marimo/_code_mode/_context.py
```diff
@@ -927,10 +927,24 @@ def _dry_run_compile(self, ops: list[_Op]) -> None:
...
+                for name in sorted(new_multiply_defined):
+                    existing = graph.get_defining_cells(name) - batch_ids
+                    if existing:
+                        labels = ", ".join(
+                            self._cell_label(cid) for cid in sorted(existing)
+                        )
+                        details.append(
+                            f"  - {name!r} is already defined in cell {labels}"
+                        )
+                    else:
```
dmadisetti, line 946: I don't think this branch should be possible?

dmadisetti: Oh, maybe a re-def in the ops. Worth adding a test for I think

mscolnick: thanks `test_multiply_defined_two_new_cells_in_batch`

PR #9057, opened 2026-04-06
Surface graph errors on code_mode NotebookCell

manzt:
When a cell has a dataflow error (multiply-defined variable, cycle), the `status` field already reported `"marimo-error"` but the `error` field was `None` because these errors prevent execution.

`NotebookCell` now exposes an `errors` property that combines both graph errors from the kernel's error registry and runtime exceptions into a uniform `CellError` type:

```py
for cell in ctx.cells:
	for err in cell.errors:
		err.kind       # "graph" or "runtime"
		err.msg        # human-readable description
		err.exception  # the Exception for runtime errors, else None
```

### DIFF
--- a/tests/_code_mode/test_cells_view.py
+++ b/tests/_code_mode/test_cells_view.py
@@ -426,7 +426,9 @@ class TestEnrichedCellStatus:
         )
         enriched = EnrichedCell(cell, impl)
         assert enriched.status == "exception"
-        assert enriched.error is err
+        assert len(enriched.errors) == 1
+        assert enriched.errors[0].kind == "runtime"
+        assert enriched.errors[0].exception is err
 
     def test_code_edited_becomes_stale(self) -> None:
         cell = _cell("a", "x = 2")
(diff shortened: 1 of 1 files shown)


marimo/_code_mode/_context.py
```diff
@@ -95,6 +96,40 @@
     "running",  # currently executing
 ]
 
+CellErrorKind = Literal["graph", "runtime"]
+
+
+class CellError:
```
mscolnick, line 107: any reason this could not be a dataclass with `frozen` and `slots`?

manzt: nope! that's better. i'll implement

marimo/_code_mode/_context.py
```diff
@@ -258,13 +295,29 @@ def status(self) -> CellStatusType | None:
         return rr
 
     @property
-    def error(self) -> Exception | None:
-        """The exception from the last run, if any.
+    def errors(self) -> list[CellError]:
```
dmadisetti, line 298: `error` -> `errors` is causing a CI break

PR #9067, opened 2026-04-06
fix: reset Plotly axis settings when switching between different chart types

### LINKED ISSUE #5898: Plotly plots are changing when switching between plots in a cell
### Describe the bug

I created multiple different plotly plots (using go.Figure()), and I'm trying to show them in a single cell, switching between them using marimo ui elements.

When switching between the plots, they suddenly change and distort in an unexpected way.

I provided reproduction code, in which you need to:

1. Observe the plot (radio value 2).
2. Switch the radio ui value to 3.
3. View the plot (radio value 3).
4. Comment the last line in the last cell (`mo.ui.plotly(_fig)`) and then uncomment it.
5. View the plot again (radio value 3) - it'll be different then before (and it'll be correct now).
6. Switch the radio ui value to 2.
7. Observe the plot (radio value 2) - it'll be different then before (and it'll be **incorrect** now).

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.16",
  "editable": false,
  "OS": "Linux",
  "OS Version": "6.1.0-37-cloud-amd64",
  "Processor": "",
  "Python Version": "3.10.6",
  "Binaries": {
    "Browser": "--",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.8.2",
    "narwhals": "2.0.1",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.19.2",
    "pyyaml": "6.0.2",
    "starlette": "0.47.2",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "duckdb": "1.3.0",
    "nbformat": "5.10.4",
    "pandas": "2.1.4",
    "polars": "1.30.0",
    "pyarrow": "15.0.2",
    "pytest": "8.4.1",
    "python-lsp-server": "1.13.0",
    "ruff": "0.12.4",
    "sqlglot": "26.25.3"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

```python
import marimo

__generated_with = "0.14.16"


@app.cell
def _():
    import marimo as mo

    from plotly import graph_objects as go
    import numpy as np
    import pandas as pd
    return go, mo, np, pd


@app.cell
def _(mo):
    r = mo.ui.radio(['1', '2', '3'], value='2')
    r
    return (r,)


@app.cell
def _(go, mo, np, pd, r):
    """
    Creating 2 figures of different types (scatter and histogram, although I think it happens with other types as well)
    """

    _fig1 = go.Figure()

    _n = 100

    _x = np.arange(0, _n, 0.1)
    _y = np.sin(_x)

    _fig1.add_trace(
        go.Scatter(
            x=_x,
            y=_y
        )
    )

    _fig2 = go.Figure()

    _fig2.add_trace(
        go.Histogram(
            x=np.random.normal(0, 1, _n * 10),
            nbinsx=100
        )
    )

    _fig3 = go.Figure()

    df = pd.DataFrame(_y).describe().T.drop(columns='count')

    _fig3.add_trace(
        go.Bar(
            x=df.columns,
            y=df.values[0]
        )
    )

    # Shown figure is controlled with a radio ui, although even hard coded changes lead to the same issue
    _fig = _fig1 if r.value == '1' else _fig2 if r.value == '2' else _fig3

    mo.ui.plotly(_fig)
    return


if __name__ == "__main__":
    app.run()

```

mscolnick:
Closes #5898

When switching between different Plotly trace types (e.g., scatter → histogram → bar), axis settings from the previous chart bled into the new chart, causing distortion. Now axis settings are only preserved when trace types are compatible; otherwise they are reset to let Plotly auto-compute correct ranges.

### DIFF
--- a/frontend/src/plugins/impl/plotly/usePlotlyLayout.ts
+++ b/frontend/src/plugins/impl/plotly/usePlotlyLayout.ts
@@ -41,7 +41,7 @@ export function createInitialLayout(figure: Figure): Partial<Plotly.Layout> {
 
 /**
  * Returns true if two figures have compatible trace types.
- * When traces are incompatible (different types or count), axis settings
+ * When traces are incompatible (different types, count, or order), axis settings
  * from the old figure should not be preserved as they would distort the
  * new chart. See https://github.com/marimo-team/marimo/issues/5898
  */


frontend/src/plugins/impl/plotly/usePlotlyLayout.ts
```diff
@@ -121,6 +149,9 @@ export function usePlotlyLayout({
     return structuredClone(originalFigure);
   });
 
+  // Track the previous figure to detect trace type changes
+  const prevFigureRef = useRef(figure);
```
manzt, line 153: nit: at first glance this ref looks redundant bc i _think_ `figure` state still holds the old value when this effect runs. but you'd probably need to disable to linting rule.

PR #9069, opened 2026-04-06
fix: Tab in AI chat triggers autocomplete instead of indenting (#5718)

### LINKED ISSUE #5718: Tab in AI chat indents opposed to autocomplete
### Describe the bug

Hitting tab (at least the first time) indents chat opposed to auto complete.

I noticed this on @koaning's video, and have run into it myself. Not a biggie, but still annoying. See https://youtu.be/NIBprn5cEZA?t=347

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

HEAD

</details>

### Code to reproduce

Type `@` and tab in a new chat

mscolnick:
The Tab keymap entry in the chat input's resource extension was missing `preventDefault: true`, so the browser/CodeMirror default Tab behavior (indentation) fired even when the autocomplete handler ran.

Closes #5718

### DIFF
--- a/frontend/src/core/codemirror/ai/resources.ts
+++ b/frontend/src/core/codemirror/ai/resources.ts
@@ -72,6 +72,7 @@ export function resourceExtension(opts: {
       keymap.of([
         {
           key: "Tab",
+          preventDefault: true,
           run: (view: EditorView) => {
             return acceptCompletion(view);
           },


frontend/src/core/codemirror/ai/resources.ts
```diff
@@ -72,6 +72,7 @@ export function resourceExtension(opts: {
       keymap.of([
         {
           key: "Tab",
+          preventDefault: true,
           run: (view: EditorView) => {
             return acceptCompletion(view);
```
mscolnick, line 77: this would prevent tabbing in here, which i _think_ is actually ok. Cursor does the same

PR #9072, opened 2026-04-06
fix: table hangs when sorting/page-size change triggers cell re-render with style_cell/hover_template

### LINKED ISSUE #9004: mo.ui.table becomes unusable after sorting and changing row count on ~100k rows
### Describe the bug

I was happily playing with column sorting, then changed the limit of rows to bring into the table (dataframe).
The cell with `mo.ui.table` becomes then unusable in the browser.

Findings:
- initial render succeeds.
- Changing the row count before sorting succeeds.
- If I then sort by Tax, ascending or descending, and change the row count again while that sort is active, the table hangs.
- if I wait long enough (as browser will ask: wait or exit), the table area shows skeleton/blinking cells and scrollbar thumb is very small but table never becomes usable

Sorting is one confirmed trigger sequence but there may be others. I suspect full-text search as I faced the exact same behavior, but can't deterministically repro (not in below nb at least).

### Will you submit a PR?

- [ ] Yes

### Environment

mo 0.22
win/linux

### Code to reproduce

See below a repro + the app in molab: https://molab.marimo.io/notebooks/nb_aK2iAc2ukeR6T3GrLB4S9z/app
### steps:

1. sort by Tax (asc or desc)
2. change the size of the df e.g.  180k+ rows
--> browser hangs and tab mem usage goes balistic

```
import marimo

__generated_with = "0.22.0"
app = marimo.App(width="wide")


@app.cell
def _():
    import marimo as mo
    import numpy as np
    import polars as pl

    return mo, np, pl


@app.cell
def _(mo):
    rows = mo.ui.slider(
        start=100,
        stop=200_000,
        step=100,
        value=100_000,
        include_input=True,
        show_value=True,
        label="Rows",
    )
    show_summaries = mo.ui.switch(
        value=False,
        label="Column summaries",
    )

    controls = mo.hstack([rows, show_summaries], justify="start", gap=1.0)
    controls
    return rows, show_summaries


@app.cell
def _(mo, np, pl, rows, show_summaries):
    n_rows = int(rows.value or 100_000)
    rng = np.random.default_rng(0)

    df = pl.DataFrame(
        {
            "row_id": np.arange(n_rows),
            "Tax": np.round(rng.uniform(0, 280, size=n_rows), 2),
            "TaxBase": np.round(rng.uniform(0, 45_000, size=n_rows), 2),
            "TaxCode": rng.choice(["U1", "U2", "R1", "R2", "Z0"], size=n_rows),
            "State": rng.choice(
                ["CA", "NY", "TX", "GA", "FL", "WA", "IL", "NJ"],
                size=n_rows,
            ),
            "Taxable": rng.choice([True, False], size=n_rows, p=[0.85, 0.15]),
            "col_01": rng.integers(0, 1_000, size=n_rows),
            "col_02": rng.integers(0, 2_000, size=n_rows),
            "col_03": rng.integers(0, 3_000, size=n_rows),
            "col_04": rng.integers(0, 4_000, size=n_rows),
            "col_05": rng.integers(0, 5_000, size=n_rows),
            "col_06": rng.integers(0, 6_000, size=n_rows),
            "col_07": rng.integers(0, 7_000, size=n_rows),
            "col_08": rng.integers(0, 8_000, size=n_rows),
            "col_09": rng.integers(0, 9_000, size=n_rows),
            "col_10": rng.integers(0, 10_000, size=n_rows),
            "col_11": rng.integers(0, 11_000, size=n_rows),
            "col_12": rng.integers(0, 12_000, size=n_rows),
        }
    )

    def style_cell(row_id: str, column_name: str, value: object) -> dict[str, str]:
        if column_name != "Tax":
            return {}
        try:
            if float(value) >= 250:
                return {"backgroundColor": "light-dark(#fee2e2, #7f1d1d)"}
        except (TypeError, ValueError):
            return {}
        return {}

    def hover_template(row_id: str, column_name: str, value: object) -> str:
        return f"row={row_id} | {column_name}={value}"

    # Control note: I have not seen the same issue with mo.ui.dataframe here.
    mo.ui.table(
        df,
        page_size=20,
        show_column_summaries=show_summaries.value, # in case it makes a difference
        show_data_types=False,
        freeze_columns_left=["row_id"],
        style_cell=style_cell,
        hover_template=hover_template, # to repro what I use e.g. description of the underlying dimension code/value
    )
    return


if __name__ == "__main__":
    app.run()
```

kirangadhave:
Fixes #9004 

## Causes:
There were a bunch of different bugs which came together to cause the browser to hang. The general flow is:

1. Cell re-execution causes the `useAsyncData` to fire twice
  a. I have not addressed this behavior in the PR, since I am not sure if it's correct behavior or a bug
2. The second call to `useAsyncData` fires transition from `loading` to `pending` which wipes the previous data and sets `isPending = true`, which causes the `LoadingTable` to render with `totalRows` (which could be `100_000` or whatever the slider is set to), this adds a lot of animated DOM elements (100k rows, 8 cols = 800k animated cells added to dom)
3. full dataframe scans in _style_cells/_hover_cells `O(page_size * columns * total_rows)`. We are iterating on rows in a page, a column and running the query: `df.filter(nw.col("_marimo_row_id") == int(row))` for each

## Fixes:
1. I clamped the LoadingTable skeleton rows to pageSize, so the table renders only limited rows when `isPending = true`
1. Preserve data on `loading` to `loading` transitions (happens due to `useAsyncData` firing twice). The data wipe should only happen on `pending`
1. compute a lookup dict once instead of scanning the full dataframe per cell `O(page_size * columns)`. Using `self._searched_manager.take()` to get the slice equal to page_size and only run lookups on that.

https://github.com/user-attachments/assets/4c15005a-4d2f-49d4-87b1-1e1931a58eee

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -1315,19 +1315,25 @@ class table(
         )
 
         page_manager = self._searched_manager.take(take, skip)
+
+        if self._has_stable_row_id:
+            cell_row_ids = [str(rid) for rid in row_ids]
+        else:
+            cell_row_ids = [str(i) for i in range(take)]
+
         all_cells = [
-            TableCoordinate(row_id=str(i), column_name=col)
-            for i in range(take)
+            TableCoordinate(row_id=rid, column_name=col)
+            for rid in cell_row_ids
             for col in columns
         ]
         selected = page_manager.select_cells(all_cells)
 
-        # Group selected cells by page position, keyed by original row ID
-        pos_to_row_id = {str(i): str(rid) for i, rid in enumerate(row_ids)}
+        cell_to_orig: dict[str, str] = {
+            cid: str(rid) for cid, rid in zip(cell_row_ids, row_ids)
+        }
         lookup: dict[str, dict[str, Any]] = {}
         for cell in selected:
-            cell_row = str(cell.row)
-            row_str = pos_to_row_id.get(cell_row, cell_row)
+            row_str = cell_to_orig.get(str(cell.row), str(cell.row))
             if row_str not in lookup:
                 lookup[row_str] = {}
             lookup[row_str][cell.column] = cell.value


marimo/_plugins/ui/_impl/table.py
```diff
@@ -1294,39 +1294,73 @@ def _get_page_row_ids(
...
+            for i, rid in enumerate(page_row_ids):
+                row_dict: dict[str, Any] = {}
+                for col in columns:
+                    row_dict[col] = page_data[col][i]
+                lookup[str(rid)] = row_dict
+        else:
+            for i, rid in enumerate(row_ids):
+                row_dict = {}
+                for col in columns:
+                    row_dict[col] = page_data[col][i]
```
mscolnick: worth checking we have good test coverage here. im not sure if all dataframe libraries or lazy/eager libraries support this

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1294,39 +1294,73 @@ def _get_page_row_ids(
...
             skip, take, response
         )
 
-        return {
-            str(row): {col: do_style_cell(str(row), col) for col in columns}
-            for row in row_ids
-        }
+        page_manager = self._searched_manager.take(take, skip)
+        page_frame = getattr(page_manager, "as_frame", None)
+        page_data = (
```
mscolnick: we might lose types at this point. would be good to be explicit in this type to get it back

marimo/_plugins/ui/_impl/table.py
kirangadhave: good catch. I'll go back to using the previous `select_cells` approach, but only on the page_size length subset.

marimo/_plugins/ui/_impl/table.py
```diff
@@ -1294,39 +1294,73 @@ def _get_page_row_ids(
...
+            page_frame() if page_frame is not None else page_manager.data
+        )
```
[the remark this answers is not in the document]
kirangadhave: added a check

marimo/_plugins/ui/_impl/table.py
kirangadhave: should be fixed with above change

PR #9077, opened 2026-04-07
improvement: add cross-origin authentication token handling for WebSocket URLs

mscolnick:
- Implemented logic to append the `access_token` as a query parameter for cross-origin WebSocket connections when an `authToken` is provided.
- Updated tests to verify the correct behavior of `access_token` inclusion based on origin and token presence.
- Enhanced URL handling to strip `/health` from redirected URLs while preserving query parameters.

### DIFF
--- a/frontend/src/core/runtime/runtime.ts
+++ b/frontend/src/core/runtime/runtime.ts
@@ -152,9 +152,15 @@ export class RuntimeManager {
    */
   getLSPURL(lsp: "pylsp" | "basedpyright" | "copilot" | "ty" | "pyrefly"): URL {
     if (lsp === "copilot") {
-      // For copilot, don't include any query parameters
+      // For copilot, strip all query parameters except the auth token.
+      // Copilot doesn't understand arbitrary query params, but we still
+      // need access_token for cross-origin authentication.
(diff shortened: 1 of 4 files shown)


frontend/src/core/runtime/runtime.ts
```diff
@@ -88,6 +88,15 @@ export class RuntimeManager {
       searchParams,
       /* restrictToKnownQueryParams =*/ false,
     );
+
+    // For cross-origin runtimes, pass the auth token as a query parameter.
+    // WebSocket connections cannot send custom headers (no Authorization
+    // header), and cross-origin cookies are blocked by browsers, so the
+    // access_token query param is the only way to authenticate.
+    if (!this.isSameOrigin && this.config.authToken) {
+      url.searchParams.set(KnownQueryParams.accessToken, this.config.authToken);
```
manzt, line 97: The existing copilot LSP path wipes `url.search`, which would strip this token cross-origin. Just to flag (pre-existing, Claude helped me find)

https://github.com/marimo-team/marimo/blob/push-vswxuvkkvyyu/frontend/src/core/runtime/runtime.ts#L141-L152

PR #9088, opened 2026-04-07
fix: replace bare except with except Exception in parse_dataclass

harshadkhetpal:
## Summary

Replace bare `except:` with `except Exception:` in `marimo/_utils/parse_dataclass.py`.

Bare `except:` catches all exceptions including `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which should generally not be caught. In this case the handler immediately re-raises, so narrowing to `Exception` preserves the intent while following Python best practices (PEP 8).

**Change:**
```python
# Before
except:
    raise

# After
except Exception:
    raise
```

## Testing
No behavior change — the except clause still catches and re-raises all standard exceptions.

### DIFF
--- a/marimo/_utils/parse_dataclass.py
+++ b/marimo/_utils/parse_dataclass.py
@@ -131,7 +131,7 @@ class _DataclassParser:
                 # catch expected exceptions when conversion fails
                 except (TypeError, ValueError):
                     continue
-                except:
+                except BaseException:
                     raise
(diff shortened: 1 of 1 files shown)


Light2Dark:
Since it doesn't change anything, we probably won't merge this. You may consider remove some of the ruff ignore-rules and that may pick this up.

PR #9120, opened 2026-04-10
fix: crash on non-dict YAML frontmatter during markdown conversion

lawrence3699:
Converting a markdown file with non-dict YAML frontmatter (e.g. a YAML list) crashes with an unhandled `AssertionError`:

```
AssertionError: Expected dict, got <class 'list'>
```

**Reproduction:**

```bash
cat > test.md <<'EOF'
---
- one
- two
---

hello
EOF

marimo convert test.md
```

**Cause:** `yaml.load()` in `_utils/yaml.py` asserts `isinstance(response, dict)`. `extract_frontmatter()` in `to_ir.py` only catches `yaml.YAMLError`, so the `AssertionError` escapes.

**Fix:** Catch `(yaml.YAMLError, AssertionError)` — matching the sibling code in `from_ir.py:74` which already handles this correctly.

**Checks run:**
- `pytest tests/_convert/markdown/test_markdown_conversion.py` — 12/12 passed
- `ruff check` and `ruff format --check` on changed files — clean

### DIFF
--- a/marimo/_convert/markdown/to_ir.py
+++ b/marimo/_convert/markdown/to_ir.py
@@ -533,7 +533,7 @@ def extract_frontmatter(text: str) -> tuple[dict[str, str], str]:
         try:
             return yaml.load(yaml_content), body
         # If there's an error in parsing YAML, ignore the meta and proceed.
-        except yaml.YAMLError:
+        except (yaml.YAMLError, AssertionError):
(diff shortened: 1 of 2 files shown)


dmadisetti:
Thanks @lawrence3699, looks good could you please sign the CLA ?

PR #9126, opened 2026-04-10
fix: wrap plugin slot in TooltipProvider to fix tooltip regression

mscolnick:
Each plugin mounts its own React root inside a shadow DOM, so there's no
ambient TooltipProvider from the outer app. When user HTML with a Radix
Tooltip was passed as a label/children to a plugin, it threw
"Tooltip must be used within TooltipProvider".

Adds a single TooltipProvider in PluginSlot and removes the now-redundant
per-plugin wrappers.


Light2Dark (approved):
nice! There's another bug where button tooltips are misplaced, but I think a separate issue.

PR #9131, opened 2026-04-10
fix: sanitize custom.css

mscolnick:


dmadisetti (approved):

dmadisetti:
This is sufficent, still requires file access and opening an untrusted notebook

PR #9139, opened 2026-04-10
fix: solve failing windows test with absolute path resolution

dmadisetti:
## 📝 Summary

Fixes CI failure in windows

### DIFF
--- a/marimo/_server/files/path_validator.py
+++ b/marimo/_server/files/path_validator.py
@@ -195,10 +195,14 @@ class PathValidator:
                     )
                 )
 
-                filepath_base = directory if directory.is_absolute() else Path.cwd()
+                filepath_base = (
+                    directory if directory.is_absolute() else Path.cwd()
+                )
                 # Original-form absolute directory (no long-path conversion).
                 # filepath_base / directory is a no-op when directory is absolute.
-                directory_original = Path(os.path.normpath(filepath_base / directory))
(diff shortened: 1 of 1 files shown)


marimo/_server/files/path_validator.py
```diff
@@ -196,18 +195,42 @@ def validate_inside_directory(
...
                 else:
                     filepath_base = Path.cwd()
+                    directory_original = Path(
+                        os.path.normpath(str(filepath_base / directory))
+                    )
 
-                filepath_normalized = (
-                    self._normalize_path_without_resolving_symlinks(
-                        filepath, filepath_base
+                if not filepath.is_absolute():
```
mchav: nit:

```python
filepath_base = directory if directory.is_absolute() else Path.cwd()

abs_filepath = Path(os.path.normpath(filepath_base / filepath))

try:
    # Both paths are in the same (possibly short) form, so relative_to is reliable.
    # Reuse directory_normalized as the prefix to avoid unneeded conversions.
    relative = abs_filepath.relative_to(directory_original)
    filepath_normalized = directory_normalized / relative
except ValueError:
    # filepath is not inside directory; normalize it independently
    filepath_normalized = self._normalize_path_without_resolving_symlinks(
        filepath, filepath_base
    )
```

mchav (dismissed):

mchav (approved):

Issue #9127, opened 2026-04-10
Invalid session id in mounted Marimo apps


liquidcarbon:
### Describe the bug

I'm running Marimo mounted to a fastapi app and I've had this problem that bugs me and my users for many months.

Usually run the app on one uvicorn worker via gunicorn.  I've experimented with granian, too.

Sometimes for no obvious reason, other times when I try to gracefully reload the app, e.g. with `kill -HUP`, if someone is actively interacting with the app, we can get into this state where some marimo pages spin forever after loading partially, other FastAPI endpoints stop working altogether, and we cannot recover from this without killing the process.

I can't figure out how to reproduce this on purpose.

```
... (more startette messages up top)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 193, in __call__
    response = await self.dispatch_func(request, call_next)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/marimo/_server/api/middleware.py", line 205, in dispatch
    return await call_next(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 168, in call_next
    raise app_exc from app_exc.__cause__ or app_exc.__context__
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 144, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/marimo/_server/api/auth.py", line 266, in __call__
    await super().__call__(scope, receive, send)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/authentication.py", line 48, in __call__
    await self.app(scope, receive, send)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/marimo/_server/api/auth.py", line 255, in wrapped_app
    await app(scope, receive, send)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/cors.py", line 96, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/data/code/tdw2/.venv/lib/python3.12/site-packages/starlette/middleware/cors.py", line 154, in simple_response
[...]

PR #9181, opened 2026-04-14
fix: use shared memory for virtual files when running with app isolation 

akshayka:
This change updates isolated apps to use shared memory for virtual
file storage.

Isolated apps run in different processes from the main server. This
means they need to place virtual file data in shared memory, but
historically run-mode has run in-process and used in-memory
storage.

The change looks large because we're lifting the choice of storage
out of the kernel context, letting the caller decide what storage
backend to use. This required touching many files. In practice the change is small.

This fixes the issue identified in the following comment: https://github.com/marimo-team/marimo/issues/9127#issuecomment-4225391036

### DIFF
--- a/marimo/_ipc/types.py
+++ b/marimo/_ipc/types.py
@@ -35,7 +35,6 @@ class KernelArgs(msgspec.Struct):
     # Whether to use run-mode config (autorun) vs edit-mode config (lazy)
     is_run_mode: bool = False
     # Runtime behavior flags
-    virtual_files_supported: bool = True
     redirect_console_to_browser: bool = True
 
     def encode_json(self) -> bytes:
--- a/marimo/_ipc/types.py
+++ b/marimo/_ipc/types.py
@@ -35,6 +35,8 @@ class KernelArgs(msgspec.Struct):
     # Whether to use run-mode config (autorun) vs edit-mode config (lazy)
     is_run_mode: bool = False
     # Runtime behavior flags
+    # virtual_files_supported is ignored, but kept for backward compatibility
+    virtual_files_supported: bool = True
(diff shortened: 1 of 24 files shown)


marimo/_runtime/context/kernel_context.py
```diff
@@ -154,15 +155,17 @@ def create_kernel_context(
...
-    storage = (
+    # Storage is chosen explicitly by the caller. None means virtual files
+    # are not supported; we still construct an (inert) InMemoryStorage so
+    # the registry has a backend, but ctx.virtual_files_supported is False.
+    storage: VirtualFileStorage = (
         SharedMemoryStorage()
-        if mode == SessionMode.EDIT
+        if virtual_file_storage == "shared_memory"
         else InMemoryStorage()
     )
```
akshayka, line 170: This is the main change in this PR.

marimo/_ipc/types.py
```diff
@@ -35,7 +35,6 @@ class KernelArgs(msgspec.Struct):
     # Whether to use run-mode config (autorun) vs edit-mode config (lazy)
     is_run_mode: bool = False
     # Runtime behavior flags
-    virtual_files_supported: bool = True
```
akshayka, line 38: @manzt I can keep this argument and just ignore it to prevent a breaking change, lmk. It looks like it was ignored before

dmadisetti (dismissed):

marimo/_ipc/types.py
manzt: We extend the KernelManager impl in the VS Code extension, so we probably just need to bump the minimum marimo version in the next release.

https://github.com/marimo-team/marimo/actions/runs/24527579413

PR #9190, opened 2026-04-14
fix: only toast cell logs in app mode

mscolnick:
Previously we were toasting on all internal errors. When only this should be happening in run mode. Also this skips it for the scratch pad (which i saw when used by code_mode)

### DIFF
--- a/frontend/src/core/cells/logs.ts
+++ b/frontend/src/core/cells/logs.ts
@@ -7,7 +7,7 @@ import { Strings } from "@/utils/strings";
 import type { CellMessage, OutputMessage } from "../kernel/messages";
 import { isErrorMime } from "../mime";
 import type { CellId } from "./ids";
-import { getInitialAppMode } from "../mode";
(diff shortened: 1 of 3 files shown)


frontend/src/core/cells/logs.ts
```diff
@@ -83,7 +84,14 @@ export function getCellLogsForMessage(cell: CellMessage): CellLog[] {
         error.type === "internal",
     );
 
-    if (exceptionErrors.length > 0 && !didAlreadyToastError) {
+    // Only show the toast in app mode: edit mode already surfaces errors in
+    // the cell UI, so toasting there would be noisy and duplicative. Read the
+    // atom directly so an unset initial mode (e.g. in tests/islands) simply
+    // returns undefined instead of throwing and masking real errors.
+    const isAppMode = store.get(initialModeAtom) === "read";
```
dmadisetti, line 91: Fair, this was a little annoying

dmadisetti (approved):

PR #9191, opened 2026-04-14
feat: auto-save in code-mode and marimo-pair

mscolnick:
Agent mutations via `code_mode` / `marimo-pair` emit kernel document transactions but never hit disk. `NotificationListenerExtension` now intercepts kernel-sourced transactions and triggers a best-effort save; run mode and unnamed notebooks are skipped, and failures surface as an `AlertNotification` toast instead of raising.

  - New `SerialTaskRunner` FIFO-dispatches the blocking save off the event loop so a slow older save can't clobber a newer one, and routes `on_error` back to the loop thread.
  - `AppFileManager` gains `save_from_cells(cells)` plus a reentrant `_save_lock` serializing every public writer.
  - `generate_filecontents` normalizes empty cell names to `DEFAULT_CELL_NAME` (empty names emitted invalid `def ():`).

**Note:** the frontend still thinks the UI is not saved (because it was saved in the backend which is a new concept). I think this is ok for now, but something we can track in a follow.

### DIFF
--- a/marimo/_session/notebook/file_manager.py
+++ b/marimo/_session/notebook/file_manager.py
@@ -76,11 +76,9 @@ class AppFileManager:
         # Track the last saved content to avoid reloading our own writes
         self._last_saved_content: str | None = None
 
-        # Serializes concurrent writers (e.g. frontend Ctrl+S racing
-        # with code_mode auto-save, save_app_config during rename, …).
-        # Reentrant so public entry points can hold the lock around the
-        # full "mutate app + _save_file" sequence, and _save_file can
-        # defensively re-acquire it for direct callers.
+        # Serializes concurrent writers. Reentrant so public entry points
+        # can wrap the full "mutate app + _save_file" sequence while
+        # ``_save_file`` re-acquires for any direct caller.
         self._save_lock = threading.RLock()
 
     @property
@@ -187,10 +185,7 @@ class AppFileManager:
     ) -> str:
         """Save notebook to storage using appropriate format handler.
 
-        All file writes and mutations of ``_last_saved_content`` / ``_filename``
-        go through this method. The ``_save_lock`` serializes concurrent
-        writers — e.g. frontend Ctrl+S racing with code_mode auto-save, or a
-        ``save_app_config`` call during a ``rename``.
+        All file writes go through this method under ``_save_lock``.
 
         Args:
             path: Target file path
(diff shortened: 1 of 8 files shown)


marimo/_utils/serial_task_runner.py
```diff
@@ -0,0 +1,150 @@
...
+        on_error: Callable[[Exception], None] | None = None,
+    ) -> None:
+        """Run ``work()`` on the serial worker thread.
+
+        Offloads to the executor when called from the event loop;
+        otherwise runs inline on the caller thread (e.g. the
+        ``QueueDistributor`` worker thread). ``on_error`` is invoked
+        with any exception raised by ``work`` — posted back to the event
+        loop when off-loop, inline otherwise. A failing ``on_error`` is
+        logged and swallowed.
```
manzt, line 67: cool

tests/_server/test_file_manager.py
```diff
@@ -228,6 +228,148 @@ def test_save_cannot_rename(app_file_manager: AppFileManager) -> None:
...
+        t2.join(timeout=10)
+
+        assert not t1.is_alive(), (
+            "frontend save thread did not terminate within 10s "
+            "(likely deadlock in AppFileManager write lock)"
+        )
+        assert not t2.is_alive(), (
+            "autosave thread did not terminate within 10s "
+            "(likely deadlock in AppFileManager write lock)"
+        )
```
manzt, line 359: wierd copilot didn't see this?

marimo/_session/notebook/file_manager.py
```diff
@@ -398,37 +414,70 @@ def save(self, request: SaveNotebookRequest) -> str:
...
+                # Remove the layout from the config
+                self.app.update_config({"layout_file": None})
+
+            return self._save_file(
+                filename_path,
+                notebook=self.app.to_ir(),
+                persist=request.persist,
+            )
+
+    def save_from_cells(self, cells: Sequence[NotebookCell]) -> str:
```
manzt, line 452: I'm guessing we can improve our "save" paths overall in the future to derive from our NotebookDocument, for a followup.

manzt (approved):

tests/_server/test_file_manager.py
mscolnick: those are old comments

marimo/_session/notebook/file_manager.py
mscolnick: yep, definitely

Issue #9182, opened 2026-04-14
NotImplementedError: when using asyncio.create_subprocess_exec on Windows


ketoprakmaster:
## Summary:
Users on Windows cannot spawn subprocesses using asyncio (e.g., via nodriver, playwright, SeleniumBase, or asyncio.create_subprocess_exec) within a marimo notebook. This results in a NotImplementedError.
## Most Likely Technical Root Cause:
On Windows, asyncio provides two event loop implementations:
1. SelectorEventLoop: The legacy loop. It does not support subprocesses.
2. ProactorEventLoop: Uses Windows I/O Completion Ports (IOCP) and does support subprocesses.
Marimo currently initializes using SelectorEventLoop on Windows. Because the loop is already active when the kernel starts, calling asyncio.set_event_loop_policy() within a cell is ineffective. 
## Impact:
Any library that relies on asyncio subprocesses is broken on Windows. Examples include:
- Browser automation tools (e.g., nodriver, playwright).
- CLI tools wrapped in asyncio.create_subprocess_exec.
## Current Workarounds:
- Remote Connection: Running the subprocess externally and connecting via WebSocket.
- Thread Offloading: Spawning a separate OS thread to run a dedicated ProactorEventLoop for the subprocess logic.
---
### With the help of Gemini

---
### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.0",
  "editable": false,
  "location": "C:/Users/ACER/AppData/Roaming/uv/tools/marimo/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "AMD64 Family 23 Model 104 Stepping 1, AuthenticAMD",
  "Python Version": "3.14.0",
  "Locale": "--",
  "Binaries": {
    "Browser": "146.0.7680.178",
    "Node": "v25.9.0",
    "uv": "0.11.6 (65950801c 2026-04-09 x86_64-pc-windows-msvc)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.18.1",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.21",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
[...]

PR #9194, opened 2026-04-14
fix: use the proactor event loop policy on Windows in edit-mode runtimes

akshayka:
This change updates the kernel to use the Proactor event loop in Windows in edit mode, partially fixing #9182.

Before this change, the kernel process inherited the event loop policy of the server process (which is the selector policy, because we use the `add_reader` API). 

This is only a partial fix because it doesn't address when the kernel runs in-process (`marimo run notebook.py`). We could maybe change the event loop policy even in the in-process situation, but we would have to double check that doing so wouldn't break the server as a side-effect.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -3519,18 +3519,22 @@ def launch_kernel(
     # - Run mode (not edit) uses autorun config regardless of IPC
     is_subprocess = is_edit_mode or is_ipc
 
+    loop_factory: Callable[[], asyncio.AbstractEventLoop] | None = None
     if is_subprocess:
         restore_signals()
 
-        # The kernel subprocess does not use add_reader() (the server side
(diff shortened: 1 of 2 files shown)


mscolnick (commented):
could we add a test that this gets set in a test? or prevent someone from accidentally removing this code path.

mscolnick (dismissed):

mscolnick:
@akshayka tests are currently failing:
```
FAILED tests/_runtime/test_runtime.py::TestLaunchKernelEventLoop::test_windows_pre_314_installs_proactor_event_loop_policy - NameError: name 'windows_events' is not defined
FAILED tests/_runtime/test_runtime.py::TestLaunchKernelEventLoop::test_run_mode_on_windows_does_not_touch_event_loop_policy - NameError: name 'windows_events' is not defined
```

PR #9196, opened 2026-04-14
fix: allow for data-uris in mpl-interactive and marimo-panel after a cell has run

mscolnick:
This PR updates the logic for trusting virtual file URLs in marimo plugins, allowing data uri after some cell has been run that use inline base64 data URLs (WASM and VS Code runtimes)


manzt (approved):
Nice fix. For the `extension` -> `extensionUrl` change in the plugin, would you be down to just upgrade the front end to accept either in the zod schema? Otherwise, we can just bump minimum-kernel-version in marimo-lsp.

PR #9212, opened 2026-04-15
feat: support columns in marimo-pair

mscolnick:
Incoming transactions with set-config column updates only touched
cell.config.column metadata, leaving the MultiColumn tree unchanged
so boundary-anchor transactions never actually reorganized cells
across columns on the replica. Sort set-config changes last and
dispatch a new rebuildCellColumns action that regroups cells via
MultiColumn.fromIdsAndColumns using the reorder-cells flat order.

My smoke test:
```
  * Started with 2 cells "import marimo as mo" and `print(123)`
  1. Add a dataframe in the second column
  2. Add a chart above that as well
  3. Move the chart back to the first column
  4. Move the 123 cell to a 3rd column
  5. Add markdown headers above each column
``

### DIFF
--- a/frontend/src/core/cells/__tests__/apply-transaction.test.ts
+++ b/frontend/src/core/cells/__tests__/apply-transaction.test.ts
@@ -106,6 +106,14 @@ function pretty(s: NotebookState): string {
   return `\n${lines.join("\n")}\n`;
 }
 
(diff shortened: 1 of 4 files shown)


manzt (approved):
very nice

PR #9218, opened 2026-04-16
feat: visually distinguish null, empty, whitespace, NaN, and Infinity in table cells

### LINKED ISSUE #9155: In the table viewer, visually distinguish nulls, empty strings, and strings with only whitespace
### Description

When using a table viewer in marimo, I want to be able to tell the difference between a null value, an empty string, and a string with only whitespace

Consider
```sql
FROM UNNEST(['', null, 'null', ' ', '   '])
```

In marimo, it looks like

Image

whereby none of  `'', null, ' ', '   '` can be differentiated

### Suggested solution

The [DuckDB UI](https://duckdb.org/2025/03/12/duckdb-ui) (`$ duckdb -ui`) is the only place I have seen this fully solved. A solution like that would be great.

Image

Importantly it also differentiates empty strings, and strings with only whitespace

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

kirangadhave:
Closes #9155

## Summary
- Null values render as `None` in an italic muted pill
- Empty strings render as blank in cells, `<empty>` in filter dropdown
- Whitespace-only strings show visible markers: `␣` (space), `→` (tab), `↵` (newline)
- NaN renders as `NaN`, Infinity as `inf`/`-inf` — all in italic muted pills
- String "NaN"/"Infinity" in numeric columns detected correctly via column type context
- Filter-by-values dropdown uses the same sentinel rendering

Screenshot 2026-04-16 at 11 37 50 AM

Screenshot 2026-04-16 at 11 38 05 AM

Screenshot 2026-04-16 at 11 38 25 AM

### DIFF
--- a/frontend/src/components/data-table/column-header.tsx
+++ b/frontend/src/components/data-table/column-header.tsx
@@ -66,7 +66,6 @@ import {
   renderSortFilterIcon,
   renderSorts,
 } from "./header-items";
-import { isNumericType } from "./types";
 import { detectSentinel, stringifyUnknownValue } from "./utils";
 
 const TOP_K_ROWS = 30;
@@ -649,12 +648,10 @@ const PopoverFilterByValues = <TData, TValue>({
             {filteredData.map(([value, count], rowIndex) => {
               const isSelected = chosenValues.has(value);
               const valueString = stringifyUnknownValue({ value });
-              const isNumericColumn = isNumericType(
+              const sentinel = detectSentinel(
+                value,
                 column.columnDef.meta?.dataType,
               );
-              const sentinel = detectSentinel(value, {
-                isNumericColumn,
-              });
 
               return (
                 <CommandItem
--- a/frontend/src/components/data-table/column-header.tsx
+++ b/frontend/src/components/data-table/column-header.tsx
@@ -53,6 +53,8 @@ import {
   SelectValue,
 } from "../ui/select";
 import { type ColumnFilterForType, Filter } from "./filters";
+import { SentinelCell } from "./sentinel-cell";
+import { detectSentinel } from "./utils";
 import {
   ClearFilterMenuItem,
   FilterButtons,
@@ -65,6 +67,7 @@ import {
   renderSortFilterIcon,
   renderSorts,
 } from "./header-items";
+import { isNumericType } from "./types";
 import { stringifyUnknownValue } from "./utils";
 
 const TOP_K_ROWS = 30;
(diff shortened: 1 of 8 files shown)


frontend/src/components/data-table/column-header.tsx
```diff
@@ -53,6 +53,8 @@ import {
   SelectValue,
 } from "../ui/select";
 import { type ColumnFilterForType, Filter } from "./filters";
+import { SentinelCell } from "./sentinel-cell";
+import { detectSentinel } from "./utils";
```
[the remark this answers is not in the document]
kirangadhave: fixed

Light2Dark (commented):
I think you made a nice notebook to show the different values, I think it can be added under `marimo/_smoke_tests`

Also, I expect this to make a release highlight, could you share an image of the new changes.

Light2Dark:
@kirangadhave you may also want to look at `marimo/_smoke_tests/tables/nans.py`, there are some other NaN-like values (pd.NA, pd.NaT). Not saying we need to handle them, but just bringing to your attention. There's also np.inf, np.nan. Not sure if they have the same values 🥲

And `marimo/_smoke_tests/tables/complex_types.py` there's some Decimal nans

frontend/src/components/data-table/sentinel-cell.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return [...str].map((ch) => WHITESPACE_MARKERS[ch] ?? ch).join("");
+}
+
+function describeWhitespace(str: string): string {
+  const CHAR_NAMES: Record<string, string> = {
+    " ": "space",
+    "\t": "tab",
+    "\n": "newline",
+    "\r": "newline",
+  };
```
Light2Dark: maybe move this outside of the function

frontend/src/components/data-table/sentinel-cell.tsx
```diff
@@ -0,0 +1,91 @@
...
+    tooltip: () => "None",
+    ariaLabel: () => "None",
+  },
+  "empty-string": {
+    label: () => "<empty>",
+    tooltip: () => "<empty>",
+    ariaLabel: () => "empty string",
+  },
+  whitespace: {
+    label: (value) => renderWhitespaceMarkers(value as string),
```
Light2Dark: should the assertions be replaced with String casting? so we can avoid asserting them
```suggestion
    label: (value) => renderWhitespaceMarkers(String(value)),
```

Light2Dark (commented):
This is a nice implementation! I think works for alot of whitespace types. And the attention to aria-label stuff is great, I will remember it too.

Image

side note, i dont think we need to fix this (obscure whitespace chars) but dang haha. I can add this to a smoke test somewhere
image

frontend/src/components/data-table/sentinel-cell.tsx
kirangadhave: haha, I defined this outside the fn. I think claude decided to move it inside for some reason

kirangadhave:
> And the attention to aria-label stuff is great, I will remember it too.

I have it my global CLAUDE.md, to remind me about accessibility when reviewing my work

Light2Dark (approved):
Mind adding a picture too? (We can use for release notes)

PR #9224, opened 2026-04-16
Restore plain text tracebacks and fix exit codes for code mode

manzt:
Fixes a regression from #8376 brought back HTML-wrapping for errors in code mode. Additionally, errors in child cells now also cause the done event to report `success: false`, so `execute-code.sh` callers get exit code 1. Finally, simplifies `_print_summary` to just report `(error)` in the for cells since it appears in the traceback already.

### DIFF
--- a/marimo/_server/scratchpad.py
+++ b/marimo/_server/scratchpad.py
@@ -118,7 +118,9 @@ class ScratchCellListener(EventAwareExtension):
                 and msg.output.data
             ):
                 err = msg.output.data[0]
-                exc_type = getattr(err, "exception_type", None) or type(err).__name__
+                exc_type = (
+                    getattr(err, "exception_type", None) or type(err).__name__
+                )
                 short_id = str(msg.cell_id)[:8]
                 self.child_error_summaries.append(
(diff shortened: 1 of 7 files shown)


marimo/_server/scratchpad.py
```diff
@@ -105,10 +106,25 @@ def on_notification_sent(
...
+                msg.output is not None
+                and msg.output.channel == CellChannel.MARIMO_ERROR
+                and isinstance(msg.output.data, list)
+                and msg.output.data
+            ):
+                err = msg.output.data[0]
+                exc_type = (
+                    getattr(err, "exception_type", None) or type(err).__name__
+                )
+                short_id = str(msg.cell_id)[:8]
```
mscolnick, line 124: i think this is always short? maybe could lead to a bug otherwise?

mscolnick (approved):

marimo/_server/scratchpad.py
manzt: yeah i changed to just the msg.cell_id, so the model gets the full context.

PR #9229, opened 2026-04-16
Fix: mo.ui.refresh typing and docs

### LINKED ISSUE #9213: marimo.ui.refresh value attribute has type str, not int
### Describe the bug

The doc string for refresh says:
`    Attributes:
        value (int): The time in seconds since the refresh has been activated.`
However, refresh.value is actually of type 'str' taking the form "\<refresh period\> (\<count\>)". You can see that in current refresh demo notebook:
https://docs.marimo.io/api/inputs/refresh/#refresh

Implementing the documented behavior where the value be the time in seconds (int, float, or whatever) would be useful from a programmatic standpoint, and so that feels like the right solution. I haven't traced down the internal magic to make that happen, yet.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

_No response_

akshayka:
The value of ui.refresh is typically a string, but it was documented to be an int.

The value was weird in that it started as an int of 0, then changed to '<interval> (number of refreshes)`.

Rather than keep this initialization behavior, this change updates ui.refresh's initial value to be an empty string. This is a breaking change, but at least they are both false-y.

This change also updates typing and docs to say the value is a str.

Fixes #9213

### DIFF
--- a/marimo/_plugins/ui/_impl/refresh.py
+++ b/marimo/_plugins/ui/_impl/refresh.py
@@ -11,7 +11,7 @@ if TYPE_CHECKING:
 
 
 @mddoc
-class refresh(UIElement[int, int]):
+class refresh(UIElement[str, str]):
     """A refresh button that will auto-refresh its descendants for a given interval.
 
     Each option value can either be a number (int or float) in seconds or a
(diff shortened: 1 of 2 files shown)


kirangadhave (dismissed):
🚀

PR #9232, opened 2026-04-16
feat: code-mode .screenshot() api

mscolnick:
Add `ctx.screenshot()` to code-mode. Adds the ability for agents to capture PNG screenshots of cell outputs during code-mode sessions.

  - Launches a headless Chromium browser (via Playwright) connected to the existing running server in ?kiosk=true mode
  - Browser is lazily initialized on first call and reused across multiple screenshot() calls
  - Cleaned up automatically when the async with context manager exits
  - Auth token is scoped — injected only at the /execute endpoint into request.meta["screenshot_auth_token"], then passed as access_token query param to Playwright

  API

```
      ctx.create_cell("import altair as alt; chart = alt.Chart(...)")
      ctx.run_cell(...)

  # Screenshot by index, name, cell ID, or NotebookCell
  img = await ctx.screenshot(0)
  img = await ctx.screenshot("my_cell")
  img = await ctx.screenshot(ctx.cells[-1])

  # Options
  url = await ctx.screenshot("my_cell", as_data_url=True)
  await ctx.screenshot("my_cell", save_to="out.png")
  await ctx.screenshot("my_cell", timeout_ms=60_000)
```

Also adds `ctx.find_cell_defining_object(obj)` to resolve a live Python object back to the cell that defines it. While often the object is in another cell, this can help with some of the cases.

### DIFF
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -82,11 +82,13 @@ from marimo._utils.formatter import DefaultFormatter
 
 if TYPE_CHECKING:
     from collections.abc import Iterator, Sequence
+    from os import PathLike
     from types import TracebackType
 
     from typing_extensions import Self
 
     from marimo._ast.cell_manager import CellManager
+    from marimo._code_mode.screenshot import _ScreenshotSession
     from marimo._runtime.dataflow import DirectedGraph
     from marimo._runtime.runtime import Kernel
 
@@ -577,6 +579,7 @@ class AsyncCodeModeContext:
         self._ui_updates: list[tuple[UIElementId, Any]] = []
         self._cells_to_run: set[CellId_t] = set()
         self._entered = False
+        self._screenshot_session: _ScreenshotSession | None = None
 
     def _require_entered(self) -> None:
         if not self._entered:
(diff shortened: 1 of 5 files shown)


marimo/_server/api/endpoints/execution.py
```diff
@@ -301,10 +301,23 @@ async def sse_generator() -> AsyncGenerator[str, None]:
...
+                    # code-mode screenshot support.  We use the
+                    # server's own host/port (from config) rather
+                    # than the request's Host header to prevent
+                    # header-spoofing attacks.
+                    http_req.meta["screenshot_auth_token"] = str(
+                        app_state.session_manager.auth_token
+                    )
+                    base_url = app_state.base_url.rstrip("/")
+                    http_req.meta["screenshot_server_url"] = (
+                        f"http://{app_state.host}:{app_state.port}{base_url}"
```
manzt: I'm guessing this is fine (preferred) to hard code scheme here?

marimo/_code_mode/_context.py
```diff
@@ -1116,6 +1121,204 @@ def run_cell(self, target: str) -> None:
...
+        Example::
+
+            cell_id = ctx.find_cell_defining_object(chart)
+            if cell_id is not None:
+                image = await ctx.screenshot(cell_id)
+        """
+        try:
+            globals_map = self.globals
+        except Exception:
+            return None
```
manzt: When/how does this raise an exception?

marimo/_code_mode/_context.py
```diff
@@ -1116,6 +1121,204 @@ def run_cell(self, target: str) -> None:
...
+
+        names = [name for name, value in globals_map.items() if value is obj]
+        if not names:
+            return None
+
+        try:
+            graph_cells = self.graph.cells
+        except Exception:
+            return None
```
manzt: Similar here. Can we drop these try/except blocks, i'm not sure how/when they would fail.

manzt (dismissed):
Nice. Added some additional tests.

manzt (approved):

PR #9233, opened 2026-04-16
Add `ctx.packages` namespace to code mode

manzt:
Code mode's single `install_packages()` method dispatched to `InstallPackagesCommand`, which scans the kernel for all other missing modules and re-runs affected cells on success. That's the right default for auto-install on import error, but surprising when an agent is explicitly asking to add one package.

These changes introduces a `ctx.packages` namespace with `add`, `remove`, and `list`, installed into `marimo/_code_mode/_packages.py`.

`add` goes through a custom install loop that keeps the streaming frontend notifications (queued → installing → installed/failed with logs) but drops the missing-module scan and cell re-runs (installing exactly what was asked, nothing more). `remove` and `list` are thin wrappers over `PackageManager.uninstall` / `list_packages` and update the script metadata for sandboxed notebooks the same way the HTTP endpoints do.

Mutations are queued and flushed on context exit before cell ops, so newly installed packages are importable from cells added in the same batch. `list` returns the current environment state and raises if `add` or `remove` have been queued. The alternative (stale data or silently overlaying pending ops) is worse than a hard error.

        ctx.packages.list()
        ctx.packages.add("pandas", "numpy>=1.26")
        ctx.packages.remove("old-pkg")

### DIFF
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -37,6 +37,11 @@ from marimo._ast.cell import (
 from marimo._ast.cell_id import CellIdGenerator
 from marimo._ast.compiler import compile_cell
 from marimo._ast.names import SETUP_CELL_NAME
+from marimo._code_mode._packages import (
+    Packages,
+    _AddPackage,
+    _RemovePackage,
+)
 from marimo._code_mode._plan import (
     _AddOp,
     _build_plan,
@@ -71,7 +76,6 @@ from marimo._runtime.commands import (
     CommandMessage,
     DeleteCellCommand,
     ExecuteCellCommand,
-    InstallPackagesCommand,
     UpdateUIElementCommand,
 )
 from marimo._runtime.context import get_context as _get_runtime_context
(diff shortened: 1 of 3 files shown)


marimo/_code_mode/_context.py
```diff
@@ -589,14 +593,25 @@ def _require_entered(self) -> None:
                 "Without 'async with', operations are silently lost."
             )
 
+    def __getattr__(self, name: str) -> Any:
+        # Legacy alias: `ctx.install_packages(...)` was the pre-namespace
+        # API. Kept as a hidden shim for in-flight skills / examples;
+        # does not appear in dir() or IDE completion. Prefer
+        # `ctx.packages.add(...)` in new code.
+        if name == "install_packages":
```
manzt, line 605: `ctx.install_packages` is referenced in the skill, so we should probably keep around for a bit: https://github.com/marimo-team/marimo-pair/pull/36

Could use as a hook/nudge to prompt the user to upgrade the skill.

marimo/_code_mode/_packages.py
```diff
@@ -0,0 +1,269 @@
...
+    result: list[str] = []
+    for pkg in packages:
+        if isinstance(pkg, (list, tuple)):
+            result.extend(pkg)
+        else:
+            result.append(pkg)
+    return result
+
+
+class Packages:
```
mscolnick, line 57: probably worth adding the help class here

mscolnick (approved):

PR #9248, opened 2026-04-17
fix: kill kernel's process group on shutdown

akshayka:
On shutdown, kill the kernel's entire process group so that child processes (started by the kernel or by the user) are also killed.

Also add a poller to the kernel that makes the kernel kill itself if it detects the server died (this logic was mostly copied from ipykernel).


akshayka:
I am closing this in favor of #9257. This is too complicated.

PR #9257, opened 2026-04-17
fix: kill kernel's process group on shutdown

akshayka:
This change updates the server to kill the entire kernel process group in Unix-like systems, not just the kernel process. It also adds clean-up logic in the event that the parent process dies.

**User-spawned processes.** Before this change, processes started by user-code in the kernel (such as `sleep 100`) weren't shut down on kernel shutdown, leading to leaks. With this change, the entire process group is killed. There are a few unit tests that exercise this, as well as a smoke test.

**Zombie processes.** This change adds background tasks that reap zombie kernel processes while the server is running. This will only make a difference for long-lived servers, as on server shutdown zombies should be reaped by the operating system.

**Orphaned kernels.** Additionally, on Unix-like systems, this change adds a poller to the kernel process that checks whether the server that started it is still alive. If the server dies, the poller causes the kernel process to terminate. For example, if you start a marimo edit session, then manually kill just the server, the kernel and its associated processes (which usually have `multiprocessing` in their command, you the reviewer may have seen such processes hanging around) will now also be cleaned up:

```
[W 260418 07:40:14 parent_poller:31] Parent server appears to have exited, shutting down.
```

**Single-file sandbox sessions.** A parent poller is started to clean up uv and marimo if the CLI process is terminated.

These cleanup mechanisms are applied to edit-mode sessions, IPC (sandbox) sessions, and AppHost sessions.

These changes are modeled after similar logic in `ipykernel`.

### DIFF
--- a/marimo/_runtime/parent_poller.py
+++ b/marimo/_runtime/parent_poller.py
@@ -1,0 +1,163 @@
+# Copyright 2026 Marimo. All rights reserved.
+"""Poll another process and exist when that process exits.
+
+Ported from ipykernel.parentpoller (BSD-licensed) to avoid orphaned kernel
+processes when the marimo server is killed ungracefully (e.g. SIGKILL, OOM).
+Without this, a crashed server can leave the kernel and its children running
+forever.
+"""
+
+from __future__ import annotations
+
+import os
+import signal
+import sys
+import time
+from dataclasses import dataclass
+from threading import Event, Thread
+from typing import TYPE_CHECKING, Final
+
+from marimo import _loggers
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+LOGGER = _loggers.marimo_logger()
+
+_PARENT_POLL_INTERVAL_SECONDS: Final[float] = 1.0
+_PARENT_SHUTDOWN_WAIT_SECONDS: Final[float] = 1.0
+
+
+def kill_own_process_group() -> None:
+    """Force-kill this process and all peers in its process group."""
+    try:
+        os.killpg(os.getpgrp(), signal.SIGKILL)
+    except ProcessLookupError:
+        pass
+    except OSError:
+        LOGGER.debug(
+            "Failed to kill kernel process group; exiting current process.",
+            exc_info=True,
+        )
+    os._exit(1)
+
+
+@dataclass(frozen=True)
+class ParentPollerHandle:
+    """State shared between a parent poller and its owner."""
+
+    exit_detected: Event
+    cleanup_complete: Event
+
+    def mark_cleanup_complete(self) -> None:
+        """For the caller to signal to the poller it can exit."""
+        self.cleanup_complete.set()
+
+    def finalize_if_parent_died(self) -> None:
+        if not self.exit_detected.is_set():
+            return
+
+        self.mark_cleanup_complete()
+        kill_own_process_group()
+
+
+class ParentPollerUnix(Thread):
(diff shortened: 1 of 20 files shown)


akshayka:
Multiple tests in CI are failing that are not related to this change.

marimo/_runtime/parent_poller.py
```diff
@@ -0,0 +1,78 @@
...
+import sys
+import time
+from threading import Thread
+
+from marimo import _loggers
+
+LOGGER = _loggers.marimo_logger()
+
+
+class ParentPollerUnix(Thread):
```
akshayka, line 23: I can add the Windows-equivalent poller in a follow-up

marimo/_utils/subprocess.py
```diff
@@ -271,3 +276,105 @@ def safe_popen(
...
+    if _process_finished(process):
+        return
+
+    LOGGER.warning(
+        "Process %s did not respond to SIGTERM. Force killing.", pid
+    )
+    if pgid is not None:
+        os.killpg(pgid, signal.SIGKILL)
+    elif pid is not None:
+        os.kill(pid, signal.SIGKILL)
```
dmadisetti: `except ProcessLookupError:`

marimo/_utils/subprocess.py
```diff
@@ -271,3 +276,105 @@ def safe_popen(
...
+    if pid is None:
+        return
+
+    if is_windows():
+        # TODO(akshayka): Investigate whether we need to kill an entire
+        # process group on Windows, and if so how ...
+        process.terminate()
+        return
+
+    pgid = os.getpgid(pid)
```
dmadisetti, line 362:
```suggestion
    try:
        pgid = os.getpgid(pid)
    except ProcessLookupError:
         # The process is already dead. Since kernel processes call setsid()
         # to become their own process group leader (PID == PGID), we can
         # use the PID as the PGID to clean up any remaining children in
         # the group.
         pgid = pid
```

marimo/_utils/subprocess.py
```diff
@@ -271,3 +276,105 @@ def safe_popen(
...
+    target_pgid: int | None
+    if pgid == os.getpgrp():
+        # This should never happen ... the kernel process makes sure to
+        # call setsid and become the group leader
+        LOGGER.warning("The target's pgid matches the server's (%d)", pgid)
+        target_pgid = None
+        process.terminate()
+    else:
+        target_pgid = pgid
+        os.killpg(pgid, signal.SIGTERM)
```
dmadisetti, line 372:
```suggestion
        os.killpg(pgid, signal.SIGTERM)
        try:
            os.killpg(pgid, signal.SIGTERM)
        except ProcessLookupError:
            # The entire process group is already dead
            return
```

dmadisetti (commented):
Maybe a little aggressive in testing, but I manually killed processes and got a few stack traces with the process lookup. I think they might be worth guarding against? Aside from that- seemed to clean up everything.

marimo/_utils/subprocess.py
akshayka: I don't think this makes sense? The process doesn't exist

marimo/_utils/subprocess.py
akshayka: I changed the callsites instead to ignore the exception, I prefer letting this function raise

dmadisetti (approved):

PR #9260, opened 2026-04-18
fix(file_browser): single-click selects, double-click navigates

kirangadhave:
## 📝 Summary

Stacked on PR https://github.com/marimo-team/marimo/pull/9259

Splits `file_browser` row click dispatch so selection and navigation are separate gestures:
- Single click: toggles selection on selectable rows
- Double click: navigates into a directory

Reason:
Clicking the row and navigating into the folder is confusing behavior. Switched clicking to select instead

https://github.com/user-attachments/assets/9ffd802f-9b09-44c3-9bff-8e1bc39c2f50

### DIFF
--- a/frontend/src/plugins/impl/FileBrowserPlugin.tsx
+++ b/frontend/src/plugins/impl/FileBrowserPlugin.tsx
@@ -340,7 +340,7 @@ export const FileBrowser = ({
   fileRows.push(
     <TableRow
       className="hover:bg-accent select-none"
-      key={"Parent directory"}
+      key="Parent directory"
       onClick={() => setNewPath(PARENT_DIRECTORY)}
     >
       <TableCell className="w-[50px] pl-4">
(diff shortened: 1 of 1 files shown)


mscolnick (approved):

frontend/src/plugins/impl/FileBrowserPlugin.tsx
```diff
@@ -376,22 +374,27 @@ export const FileBrowser = ({
...
+          "cursor-default": !isSelectable && !file.is_directory,
         })}
-        onClick={() =>
-          handleClick({
-            path: filePath,
-            name: file.name,
-            isDirectory: file.is_directory,
-          })
-        }
+        onClick={() => {
```
mscolnick, line 379: if this is not selectable, should we set the path?

alternatively, maybe an icon to jump in? some people dont find double click discoverable

PR #9262, opened 2026-04-18
Suppress `marimo-ui-value-update` echo for user-initiated changes

manzt:
The kernel broadcasts a `marimo-ui-value-update` after every UI element value change, including user-initiated ones where the frontend already has the value. The echo is invisible over WebSocket but lands stale on high-latency transports, where `broadcastMessage` applies it and visibly snaps widgets backward mid-drag (the slider lag in marimo-team/marimo-lsp#515).

These changes gate the broadcast behind a `notify_frontend` kwarg that defaults off to scope the feature for code-mode only.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -782,7 +782,7 @@ class App:
         self,
         request: UpdateUIElementCommand,
         *,
-        notify_frontend: bool = False,
+        notify_frontend: bool,
     ) -> bool:
         app_kernel_runner = self._get_kernel_runner()
         return await app_kernel_runner.set_ui_element_value(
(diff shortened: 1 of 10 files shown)


marimo/_ast/app.py
```diff
@@ -779,10 +779,15 @@ def _run_cell_sync(
         return output, _Namespace(defs, owner=self)
 
     async def _set_ui_element_value(
-        self, request: UpdateUIElementCommand
+        self,
+        request: UpdateUIElementCommand,
+        *,
+        notify_frontend: bool = False,
```
mscolnick: Is it worth removing the default so it's always explicit? seems like you covered all the cases anyways

manzt: Yeah that's great call.

manzt:
i tried the copilot agent from my phone to make the naming fix, no CLA lol

mscolnick (approved):

PR #9282, opened 2026-04-20
ci(test_be): use uv-managed Python instead of setup-python

akshayka:
The hostedtoolcache interpreter from actions/setup-python has ABI quirks that caused lazy submodule imports (numpy.random, numpy.testing, matplotlib.backends.backend_svg/backend_webagg_core) to raise ModuleNotFoundError on CI despite loading fine locally. It also caused test_acts_as_test to fail with "No module named pytest" because sys.executable pointed outside the project venv.

Switch to `uv python install` so the matrix interpreter comes from uv's managed builds (python-build-standalone), and add an explicit `uv sync --group <group>` before each pytest run so the project venv is the source of truth for subprocesses.

3.14 is now in uv's managed Python catalog, so we no longer have to use the hostedtoolcache interpreter.

This PR also fixes a bug in the pytest-changed plugin which was preventing the `test-all` label from doing its job.

### DIFF
--- a/.github/workflows/test_be.yaml
+++ b/.github/workflows/test_be.yaml
@@ -149,8 +149,14 @@ jobs:
         with:
           fetch-depth: 0 # Fetch all history for git diff
 
+      # Use uv's managed Python rather than the hostedtoolcache build from
+      # actions/setup-python: the hostedtoolcache interpreter has ABI quirks
+      # that caused numpy/matplotlib lazy submodule imports to fail on CI.
       - name: 🐍 Setup uv
         uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7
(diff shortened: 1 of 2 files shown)


dmadisetti (commented):
Fewer deps seems like a good thing. I wonder if this will break minimal? I think sync will attempt to use the lock file

akshayka:
> Fewer deps seems like a good thing. I wonder if this will break minimal? I think sync will attempt to use the lock file

I removed the sync. I also just pushed a fix to the pytest changed plugin so it respects the test-all label. I will let the tests run and see what happens.

dmadisetti (dismissed):
Ok, confirmed minimal still works: `numpy==2.2.6` vs `numpy==2.4.4` in the resolved deps.

PR #9284, opened 2026-04-20
tests: save and restore main for islands tests

akshayka:
Unit tests that exercise multiprocessing typically need to save and restore main. This change applies our save-and-restore-main utility to islands tests to fix CI (`test_build` and `test_render` are failing without this change.)

In a follow-up change I will attempt applying the utility globally.


mscolnick (approved):

mscolnick:
there is a bug in a pandas test. ill fix in another PR

PR #9296, opened 2026-04-21
Fix type-check `Channel.Pull` with NewType and union msg_type

manzt:
PR #8713 was merged despite mypy errors in marimo/_ipc/connection.py (that's on me). `Channel.Pull` declared `msg_type: type[T]`, which rejects both the `KernelMessage` NewType and the `CommandMessage` / `BatchableCommand` union aliases that the actual callers pass.

Overload `Channel.Pull` the same way msgspec's `Decoder` and pass `KernelMessage` (rather than `bytes`) for the stream channel so the return type matches `Channel[KernelMessage]`.


manzt:
I can make a follow up for redshift, altair locale, ibis import-untyped type errors.

mscolnick (approved):

PR #9305, opened 2026-04-21
fix: do not mutate the test environment during tests

akshayka:
This PR fixes a test regression caused by tests that mutated the test virtual environment.

A few code mode tests were calling `uv add` and `uv remove`. This synced the environment to marimo's dependencies and deleted all the test dependencies. This led to many errors that looked different (multiprocessing.spawn failing to start, various packages failing to import) but that were all symptoms of the same issue .

In this change, we mock the package manager. We also add a test fixture that detects venv environment corruption and fails fast to prevent this from happening again.

Before this change, 45+ tests were failing. With the following four tests are failing on Ubuntu, macOS, and Windows. These tests are unrelated to the root cause addressed in this PR.

```
FAILED tests/_plugins/ui/_impl/test_table.py::test_download_as[df0] - AssertionError
FAILED tests/_plugins/ui/_impl/test_table.py::test_get_data_url_no_deps - AssertionError: assert False
 +  where False = <built-in method startswith of str object at 0x7f8e887c9390>('data:application/json;base64,')
 +    where <built-in method startswith of str object at 0x7f8e887c9390> = 'data:text/csv;base64,dmFsdWUKMQoyCjMK'.startswith
 +      where 'data:text/csv;base64,dmFsdWUKMQoyCjMK' = GetDataUrlResponse(data_url='data:text/csv;base64,dmFsdWUKMQoyCjMK', format='csv').data_url
FAILED tests/_server/api/endpoints/test_assets.py::test_serve_static_allowed_files - RuntimeError: File at path /home/runner/work/marimo/marimo/marimo/_static/manifest.json does not exist.
FAILED tests/_runtime/test_complete.py::test_get_completion_options_bails_out_when_timeout_elapsed - AssertionError: assert 'function' == ''
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -126,7 +126,16 @@ def _assert_venv_not_wiped(
     Checking the canary before and after each test pinpoints the exact
     offender: see `tests/_code_mode/test_context.py::TestPackages` for
     the required mocking pattern.
+
+    Under pytest-xdist, other workers run concurrently; the post-check
+    can fire on a worker whose current test did not cause the mutation.
+    We soften the post-check message in that case and recommend
+    rerunning with `-p no:xdist` to identify the culprit.
     """
+    import os
+
+    under_xdist = "PYTEST_XDIST_WORKER" in os.environ
+
     if _venv_canary_path is None:
         # Could not establish a canary (e.g. hypothesis not installed on
         # disk); nothing to protect.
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -86,6 +86,74 @@ def pytest_collection_modifyitems(
...
+
+    spec = importlib.util.find_spec("hypothesis")
+    if spec is None or spec.origin is None:
+        return None
+    path = Path(spec.origin)
+    return path if path.exists() else None
+
+
+@pytest.fixture(autouse=True)
+def _assert_venv_not_wiped(
```
mscolnick, line 111: is this run on all tests?

akshayka: Yep, but should be cheap: 

Image

PR #9247, opened 2026-04-17
perf(complete): bail out of type inference when completion budget expires

mscolnick:
Completing against heavy libraries (`pandas`, `numpy`, `torch`, `aiohttp`, ...) on a cold jedi cache was paying the full per-item inference cost for every completion, even when the list is large enough that we already skip docstrings. On molab this made the first `pd.` tab take ~11s; on a local Mac with a cleared cache, ~5.5s.

- Add `compute_type=True` kwarg to `_get_completion_option`; when False, skip the `completion.type` access (and the docstring/signature path, which re-triggers the same inference).
- Collapse the split oversize/normal branches of `_get_completion_options` into one loop that tracks a time budget. Once the budget expires, remaining completions are returned with name only.

Per-completion options cost: 29.6 ms (OLD) -> 19.0 ms (NEW), ~36% across the suite. Worst-case savings for heavy libs bigger:
  - `aiohttp` (169 attrs): ~9.6s projected -> 2.8s measured (~71%)
  - `pandas` cold (141 attrs): ~11.3s pre-patch -> ~2.0s capped (~82%)

```
  Projected savings, same-lib cold

  ┌────────────────────────────────────────────┬────────┬───────────┬─────────────────┬───────────────┐
  │                    lib                     │ #comps │    OLD    │       NEW       │     saved     │
  ├────────────────────────────────────────────┼────────┼───────────┼─────────────────┼───────────────┤
  │ pandas (molab, measured pre-patch)         │    141 │ 11,300 ms │ ~2,000 ms (cap) │ ~9.3 s (82 %) │
  ├────────────────────────────────────────────┼────────┼───────────┼─────────────────┼───────────────┤
  │ aiohttp (projected OLD = 169 × 57 ms/comp) │    169 │ ~9,600 ms │        2,839 ms │ ~6.8 s (71 %) │
  ├────────────────────────────────────────────┼────────┼───────────┼─────────────────┼───────────────┤
  │ rich (measured OLD, projected NEW cap)     │     98 │  5,547 ms │ ~2,000 ms (cap) │ ~3.5 s (64 %) │
  └────────────────────────────────────────────┴────────┴───────────┴─────────────────┴───────────────┘

  Local cold cache, same scenarios before vs after

  ┌──────────────────────────────────────┬──────────┬──────────┬──────────────────────────────────┐
  │               scenario               │  before  │  after   │                Δ                 │
  ├──────────────────────────────────────┼──────────┼──────────┼──────────────────────────────────┤
[...]

PR #9306, opened 2026-04-21
fix: unblock five test_be failures left after #9305

akshayka:
Unblocks the test_be CI after #9305. Each of these five failures has an independent root cause — grouping them since the fixes are small and share the same goal.

- **`test_download_as[df0]`** — the Parquet short-circuit in `_download_as` required `polars OR (pandas AND pyarrow)`, rejecting pyarrow-alone environments even though pyarrow can write Parquet natively. Relax to `polars OR pyarrow`. The `_parquet_without_libs_reports_missing_packages` and `_parquet_with_pandas_only_prompts_pyarrow` regression tests still pass.

- **`test_get_data_url_no_deps`** — #9246 rewrote `DefaultTableManager.to_csv_str` to use the stdlib `csv` module, so `_to_chart_data_url` now returns CSV instead of falling through to JSON. Update the test expectation.

- **`test_get_data_url_values`** — same #9246 change: stdlib CSV emits column name `"value"` for single-column data, where the old pandas fallback produced integer column `0`. Update the first `expected_df` to `{"value": [1, 2, 3]}`. (Landed as a separate commit after I found it on a broader sweep with the `test-optional` group.)

- **`test_serve_static_allowed_files`** — `serve_static` returned `FileResponse` without checking existence, so Starlette raised `RuntimeError` at request time when the file was missing. CI copies `index.html` and `favicon.ico` into `_static/` but not `manifest.json`, which is how this surfaced. Add an `is_file()` guard that 404s. The test passed locally only because `_static/` doesn't exist on dev machines and `PathValidator` rejected upstream.

- **`test_get_completion_options_bails_out_when_timeout_elapsed`** — #9247 switched the production code to `time.monotonic` but the test still mocked `time.time`, so the budget never expired and all completions got types set. Mock `time.monotonic`.

## Test plan

- [x] Ran all four originally-failing tests; they pass.
- [x] Ran the affected modules (`test_table.py`, `test_assets.py`, `test_complete.py`) with the `test` group — no new failures; pre-existing `test_index*` failures need a built `_static/` which CI provides.
- [x] Ran the same modules with `test-optional` (pandas + polars + pyarrow + altair) — uncovered and fixed `test_get_data_url_values`.
- [x] Confirmed the Parquet "missing packages" prompts still fire correctly under pandas-only and no-lib environments.
- [ ] Let CI verify Linux/Windows.

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -905,7 +905,7 @@ class table(
             has_pandas = DependencyManager.pandas.has()
             has_pyarrow = DependencyManager.pyarrow.has()
 
-            if not (has_polars or (has_pandas and has_pyarrow)):
+            if not (has_polars or has_pyarrow):
                 # if pandas is installed and pyarrow is not, prompt to install pyarrow
                 if has_pandas:
                     return DownloadAsResponse(
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -900,29 +900,12 @@ class table(
                 of ``TableCell`` from cell-selection modes).
         """
         # Short-circuit Parquet when no parquet-capable lib is importable.
-        # For `DefaultTableManager` (plain-Python data), `to_parquet` routes
-        # through `_as_table_manager`, which requires pandas or polars — so the
-        # strict guard applies. Native pandas/polars/pyarrow frames export via
-        # narwhals, where pyarrow alone is sufficient for pyarrow frames and
-        # polars alone for polars frames, so the guard is relaxed accordingly.
         if args.format == "parquet":
-            from marimo._plugins.ui._impl.tables.default_table import (
-                DefaultTableManager,
-            )
-
             has_polars = DependencyManager.polars.has()
             has_pandas = DependencyManager.pandas.has()
             has_pyarrow = DependencyManager.pyarrow.has()
 
-            is_default_manager = isinstance(
-                self._searched_manager, DefaultTableManager
-            )
-            if is_default_manager:
-                can_export = has_polars or (has_pandas and has_pyarrow)
-            else:
-                can_export = has_polars or has_pyarrow
-
-            if not can_export:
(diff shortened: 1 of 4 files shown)


marimo/_plugins/ui/_impl/table.py
```diff
@@ -900,12 +900,29 @@ def _download_as(self, args: DownloadAsArgs) -> DownloadAsResponse:
...
-            if not (has_polars or (has_pandas and has_pyarrow)):
+            is_default_manager = isinstance(
+                self._searched_manager, DefaultTableManager
+            )
+            if is_default_manager:
+                can_export = has_polars or (has_pandas and has_pyarrow)
+            else:
+                can_export = has_polars or has_pyarrow
+
+            if not can_export:
```
akshayka: @kirangadhave would appreciate your review on this on particular

kirangadhave: we don't have a `pyarrow` table manager so this code path won't work

tests/_plugins/ui/_impl/test_table.py
```diff
@@ -2227,10 +2227,14 @@ def test_lazy_dataframe_with_non_lazy_dataframe(df: Any):
...
-    assert data == [{"value": 1}, {"value": 2}, {"value": 3}]
-    assert response.format == "json"
+    # DefaultTableManager.to_csv_str uses the stdlib csv module and works
+    # without pandas/polars/pyarrow, so _to_chart_data_url returns CSV before
+    # falling through to JSON.
+    assert response.data_url.startswith("data:text/csv;base64,")
+    assert from_data_uri(response.data_url)[1].decode("utf-8") == (
+        "value\n1\n2\n3\n"
+    )
+    assert response.format == "csv"
```
kirangadhave, line 2247: good catch

marimo/_plugins/ui/_impl/table.py
akshayka: reverted

kirangadhave (approved):
🚀

PR #9309, opened 2026-04-21
fix: consistent mangling for private imports

### LINKED ISSUE #9151: import of internal modules is not working any more >0.21.1
### Describe the bug

With version: marimo 23.1 there is an error when import internal modules (startswith "_").
Works with marimo 21.

I'm importing a local module like that:
```python
from cama import _camboardscan, _camera, _scanner
```
I get a name error: 
``NameError: name _cell_ffec47fb-6ddf-4e37-a31b-0d21dc8967db_camera is not defined```

Workaround would be using aliases but this feels a bit clumsy. As I'm using marimo to explore things and later integrate it maybe into my code I would have to rename everything.

Background: I use on start of development mostly internal modules and providing later on an public interface which makes it easier to mess around with my code without breaking any "public" imports for other users.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
NameError: name '_cell_ffec47fb-6ddf-4e37-a31b-0d21dc8967db_camera' is not defined
```

</details>

### Code to reproduce

_No response_

dmadisetti:
## 📝 Summary

A scoping change introduced in #8762 changed `_` prefixed imports by turning then into privates.
While this is consistent with the remainder of marimo's private pattern, it is technically a breaking change as of 0.22.0, and the edge case was not correctly handled. The change lead to a bug where imports referenced in scope were not correctly addressed. This PR handles the mangling behavior for these import cases.

An alternative PR, for retaining the behavior, is here: https://github.com/marimo-team/marimo/pull/9308 but this requires ref tracking a little differently such that the `_` pattern is properly placed into globals.

closes #9151

### DIFF
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -280,9 +280,6 @@ class ScopedVisitor(ast.NodeVisitor):
             #   import [a.b.c] - we define a
             #   from foo import [a] - we define a
             #   from foo import [*] - we don't define anything
-            #
-            # Note:
-            # Don't mangle - user has no control over package name
             basename = node.name.split(".")[0]
             if basename == "*":
                 # Use the ImportFrom node's line number for consistency
@@ -298,7 +295,13 @@ class ScopedVisitor(ast.NodeVisitor):
                     f"{line} SyntaxError: Importing symbols with `import *` "
                     "is not allowed in marimo."
                 )
-            return basename
+            # Previously we did not mangle. So this is technically breaking
+            # from 0.22.0 (#8762) but now consistent with private naming
+            # conventions.
+            mangled = self._if_local_then_mangle(basename)
+            if mangled != basename:
+                node.asname = mangled
+            return mangled
         else:
             node.asname = self._if_local_then_mangle(node.asname)
             return node.asname
--- a/marimo/_ast/visitor.py
+++ b/marimo/_ast/visitor.py
@@ -298,22 +298,34 @@ class ScopedVisitor(ast.NodeVisitor):
             # Mangle _-prefixed imports consistently with other
             # private naming conventions. We set asname (not name)
             # because name is the actual module path to import —
-            # changing it would alter what gets imported. asname
+            # changing it would alter what gets imported. `asname`
             # only controls the local binding:
             #   from my_module import _private
             # becomes:
             #   from my_module import _private as _mangled
             mangled = self._if_local_then_mangle(basename)
             if mangled != basename:
+                # An alternative would be rewriting
+                #   import _pkg.submodule
+                #   _pkg.submodule.attr
+                # as
+                #   import _pkg.submodule as _mangled
+                #   _mangled.attr
+                # but this would increase the complexity of
+                # visitor a fair amount.
+                # TODO(dmadisetti): Consider revising.
                 if "." in node.name:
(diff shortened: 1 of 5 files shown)


mscolnick (commented):
can we also test renaming the private import?

marimo/_ast/visitor.py
```diff
@@ -298,7 +295,13 @@ def _get_alias_name(
                     f"{line} SyntaxError: Importing symbols with `import *` "
                     "is not allowed in marimo."
                 )
-            return basename
+            # Previously we did not mangle. So this is technically breaking
+            # from 0.22.0 (#8762) but now consistent with private naming
+            # conventions.
+            mangled = self._if_local_then_mangle(basename)
```
mscolnick, line 306: why did this require actual code changes? I thought it was just to add a verification test?

dmadisetti: The current implementation is broken, The test I put in caught the edge case where a scoped import no longer works.

mscolnick: got it, maybe worth thinking through some other edge cases then (lazy imports etc)

marimo/_ast/visitor.py
```diff
@@ -298,7 +295,13 @@ def _get_alias_name(
...
+            if mangled != basename:
+                node.asname = mangled
```
akshayka, line 329: I don't understand this branch. Isn't `node.asname` `None`? Why are we assigning the mangled basename to it? Should it be `name.name = mangled`?

akshayka (commented):
Can you make the PR description more specific?

> and the edge case was not correctly handled.

Which edge case?

> The change lead to a bug where imports referenced in scope were not correctly addressed. This PR handles the mangling behavior for these import cases.

What do you mean by not correctly addressed? How does this PR fix it?

marimo/_ast/visitor.py
dmadisetti: We can't rename a module but:

`import _private as _mangled_private` is fine

I do think I'm missing the

`import _private.submodule`

case though. Will revise and add a comment.

marimo/_ast/visitor.py
```diff
@@ -298,7 +295,13 @@ def _get_alias_name(
                     f"{line} SyntaxError: Importing symbols with `import *` "
                     "is not allowed in marimo."
                 )
-            return basename
+            # Previously we did not mangle. So this is technically breaking
+            # from 0.22.0 (#8762) but now consistent with private naming
+            # conventions.
+            mangled = self._if_local_then_mangle(basename)
+            if mangled != basename:
```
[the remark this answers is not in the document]
dmadisetti: Thanks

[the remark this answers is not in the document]
dmadisetti: but not the right fix

akshayka:
> Can you make the PR description more specific?
> 
> > and the edge case was not correctly handled.
> 
> Which edge case?
> 
> > The change lead to a bug where imports referenced in scope were not correctly addressed. This PR handles the mangling behavior for these import cases.
> 
> What do you mean by not correctly addressed? How does this PR fix it?

@dylan can you address these comments? This PR is subtle enough that I would like the intended behavior to be very clear.

dmadisetti:
Closing out for simpler method

PR #9312, opened 2026-04-21
ci: don't cancel in-flight runs on main

akshayka:
This change updates our test workflows to run to completion on main, even if new commits land. 

Previously, a second commit landing on main before the first one's tests finished would cancel the earlier run, leaving no completed signal for that SHA. That hurts bisectability and can mask regressions.

Workflows on PRs are still cancelled when new commits land.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,8 +8,8 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,8 +8,8 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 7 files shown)


.github/workflows/test_cli.yaml
```diff
@@ -6,8 +6,8 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
mscolnick, line 10: do you need both of these changes? isn't `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` enough?

mscolnick (dismissed):

.github/workflows/test_cli.yaml
akshayka: GitHub actions only allows one pending run per group, automatically canceling any other pending runs.  [https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#using-concurren[…]ferent-scenarios](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#using-concurrency-in-different-scenarios)

.github/workflows/docs.yml
```diff
@@ -8,8 +8,8 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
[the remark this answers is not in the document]
akshayka: reverted docs changes

PR #9313, opened 2026-04-21
fix(runtime): restore sys.modules['__main__'] after run-mode kernel

### LINKED ISSUE #9293: patch_main_module leaks mutated sys.modules['__main__'] to the host
### Describe the bug

`marimo._runtime.patches.patch_main_module` unconditionally mutates `sys.modules['__main__']` via `patch_sys_module` and provides no restore path. `patch_main_module_context` exists as a context-manager sibling that saves and restores around a scope, but `marimo._runtime.runtime.launch_kernel` calls the non-context version at kernel startup, so the replacement persists for the lifetime of the process that hosts `launch_kernel`.

In subprocess kernel modes (EDIT with `is_edit_mode`, IPC with `is_ipc`, both flagged as `is_subprocess` in `launch_kernel`) the mutation is contained in a subprocess that exits after the run, so callers do not observe it. In RUN mode the kernel runs in a thread of the host process and shares `sys.modules`, so the host's `__main__` is replaced and never restored. Any code in the host that later inspects `sys.modules['__main__']` sees marimo's synthetic module.

The bug is already visible inside this repository. Three test-layer workarounds exist today for the same root cause:

- `tests/_server/test_asgi.py::TestASGIAppBuilder.setUp / tearDown`: *"Kernel threads running in RUN mode may modify `sys.modules['__main__']` via `patch_main_module`. Save it so `tearDown` can restore it and prevent contamination of later tests."*
- `tests/_server/test_session_manager.py::_preserve_main_module` (autouse fixture): *"Earlier tests (e.g. in test_asgi.py) may start kernel threads in RUN mode that call `patch_main_module()`, permanently replacing `__main__` with a module whose `__file__` points to a now-deleted temp file. On Windows the multiprocessing 'spawn' start method reads `__main__.__file__` to bootstrap the child process, so a stale path causes `FileNotFoundError` and the parent hangs on `listener.accept()`."*
- `tests/conftest.py::_save_and_restore_main` (autouse fixture added in #9286, which is the global follow-up to the `save_and_restore_main` decorator from #9284): *"Restore `sys.modules['__main__']` after each test. marimo's kernel swaps out the main module via `patch_main_module`; without this fixture, that swap leaks into subsequent tests."* Companion fixture `_ensure_main_has_file` in the same file patches `__main__.__file__` with a fake path when it's missing, which is a secondary symptom of the same leak. `tests/conftest.py::MockedKernel.teardown` also does its own manual save/restore (`sys.modules["__main__"] = self._main`).

The escalation pattern (one file → a test-utility decorator → a global autouse fixture) suggests the workaround surface keeps growing. A runtime-level fix in the kernel's lifecycle makes these test-side workarounds unnecessary and prevents a category of Windows-specific hangs and Unix-specific pickling errors that are currently only mitigated in test code.

The concrete downstream failure outside the test suite, on POSIX: after a RUN-mode session runs, `multiprocessing.get_context('spawn').Process(...).start()` in the host fails to pickle targets defined in the real `__main__`. `ForkingPickler` looks them up on the now-synthetic `sys.modules['__main__']` and raises `PicklingError: Can't pickle <function _noop at ...>: attribute lookup _noop on __main__ failed`. On Windows, the failure mode documented in `_preserve_main_module` above (`FileNotFoundError` from `__main__.__file__` pointing at a deleted temp path, parent hang on `listener.accept()`) is a separate symptom of the same root cause.

Reproduced on a clean `uv pip install marimo==0.23.1` in `python:3.12.3-slim`, no marimo source modifications. Output of the repro script (provided under "Code to reproduce" below):

```
before: main id=0x7ffffee53290
before: main has _noop = True
after: main id=0x7ffffeb2c770
after: main has _noop = False
spawn attempt:
... PicklingError: Can't pickle <function _noop at 0x...>: attribute lookup _noop on __main__ failed
```

`id(main)` differs before vs after, so `sys.modules['__main__']` is a different object. `_noop`, defined in the real main, is absent from the synthetic one. Any subsequent spawn call that references it by name fails.

#### Proposed fix

Add refcounted `save_main_module()` / `restore_main_module()` helpers to `marimo/_runtime/patches.py`, paired with the existing `patch_main_module`. Call `save_main_module()` in `KernelManagerImpl.start_kernel` for the RUN-mode (thread) path before thread creation, and `restore_main_module()` in `KernelManagerImpl.close_kernel` after the kernel thread exits.

Refcounted so that concurrent RUN-mode sessions in a single host share one captured original and restore only on the last release. This matches the existing TODO note inside `patch_main_module` about races between concurrent kernel threads; the refcount preserves LIFO or interleaved exit ordering across sessions. For subprocess kernels the helpers are not called (subprocess exits anyway), so EDIT and IPC behavior is unchanged.

While touching `close_kernel`, also add a bounded `self.kernel_task.join(timeout=5.0)` after `StopKernelCommand` in the RUN-mode branch. The existing comment in that branch acknowledges the tradeoff: *"We don't join the kernel thread because we don't want the server to block on it finishing."* That is correct for an interactive server, but batch callers observe close as incomplete since the kernel thread may still be finalizing when `close_kernel` returns, racing any host code that reads process state immediately after. A 5-second upper bound is long enough for a cooperative shutdown following `StopKernelCommand` and short enough that stuck kernels do not block server-mode session close meaningfully. If the thread is still alive at the deadline, cleanup is skipped rather than restoring `__main__` under a live thread; a warning is logged.

Total fix is three files, approximately 170 lines including comments and new unit tests:

- `marimo/_runtime/patches.py`: add `save_main_module` / `restore_main_module` with a module-level refcount + `threading.Lock`.
- `marimo/_session/managers/kernel.py::KernelManagerImpl`: call the helpers in `start_kernel` and `close_kernel`, add the bounded join, log a warning on timeout.
- `tests/_runtime/test_patches.py`: three unit tests for the helpers.

#### Alternative implementations I considered but did not prototype

- **Try/finally inside `launch_kernel`**: save the caller's `__main__` at function entry, restore in a `finally` around the function body. Simpler (no new patches.py API), but requires indenting the ~290-line body of `launch_kernel` into a `try:` block. LIFO ordering only; does not handle interleaved multi-session exits.
- **Wrap the kernel's main loop with the existing `patch_main_module_context`**: requires separating `patches.create_main_module` from `patches.patch_main_module` at the call site, and wrapping the main loop in a `with` block. Same indenting cost as the try/finally and the same LIFO limitation.

I chose the refcounted API because the existing TODO in `patch_main_module` already flags overlapping RUN-mode sessions as a concern, and the refcount is the minimum invariant that addresses it. Happy to take a different direction if preferred.

#### Scope

In scope:
- `marimo/_runtime/patches.py`: new `save_main_module` / `restore_main_module`, refcounted with `threading.Lock`.
- `marimo/_session/managers/kernel.py::KernelManagerImpl`: call the helpers in `start_kernel` (thread branch) and `close_kernel` (thread branch), add bounded join.
- `tests/_runtime/test_patches.py`: unit tests for the helpers.

Out of scope:
- `marimo/_pyodide/pyodide_session.py::patch_main_module` call site. Pyodide runs in a WASM environment without a persistent host to pollute. Left untouched.
- `marimo/_runtime/runtime.py`. The runtime keeps calling `patch_main_module` (non-context); the save/restore pair around it is managed by `KernelManagerImpl`.
- The process branch of `close_kernel`. Subprocess is terminated; host state restore is not meaningful.
- Removing the existing test-layer workarounds (`test_asgi.py` setUp/tearDown, `test_session_manager.py::_preserve_main_module`, the `conftest.py::_save_and_restore_main` autouse fixture from #9286). Those can be retired in follow-up PRs once this runtime fix lands. Not part of this bug report.

#### Testing

Local prototype on Linux in `python:3.12.3-slim`, current `origin/main`, `uv sync --group dev --group test`:

- `tests/_runtime/test_patches.py`: 14 passed (11 pre-existing + 3 new for the helpers), 1 skipped (polars not installed).
- `tests/_messaging/test_thread_local_proxy.py`: 21 passed.
- `tests/_session/managers/`: 23 passed.
- `tests/_server/export/test_exporter.py::test_run_until_completion_with_stop`, `::test_run_until_completion_with_console_output`: 2 passed.

Not run: the full `tests/` suite, `tests/_islands/`, `tests/_server/test_asgi.py`, `tests/_server/test_session_manager.py`, frontend-dependent HTML export tests.

End-to-end verification of the fix under RUN mode requires a way to reach RUN mode from a public entry point. Today `run_app_until_completion` hardcodes `mode=SessionMode.EDIT`; to exercise the restore path I invoked `SessionImpl.create(mode=SessionMode.RUN, ...)` from a thread in a local branch. With the fix applied, `sys.modules['__main__']` identity is preserved across the session lifetime; without it, the identity changes and subsequent `multiprocessing.get_context('spawn').Process(...).start()` raises `PicklingError` as shown above. I can include this integration test with the PR if helpful.

### Will you submit a PR?

- [x] Yes

### Environment

<details>

```
{
  "marimo": "0.23.1",
  "editable": false,
  "location": "/tmp/v/lib/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.12.76-linuxkit",
  "Processor": "",
  "Python Version": "3.12.3",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "0.11.7 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.3.2",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.20.0",
    "packaging": "26.1",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.44.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["marimo==0.23.1"]
# ///

import sys
import traceback

from marimo._runtime.patches import patch_main_module


def _noop() -> None:
    return None


def _report(label: str) -> None:
    main = sys.modules.get("__main__")
    print(f"{label}: main id={id(main):#x}")
    print(f"{label}: main has _noop = {hasattr(main, '_noop')}")


def main() -> None:
    _report("before")
    patch_main_module(file=None, input_override=None, print_override=None)
    _report("after")

    import multiprocessing as mp

    ctx = mp.get_context("spawn")
    try:
        p = ctx.Process(target=_noop)
        p.start()
        p.join()
        print(f"spawn attempt: OK (exitcode={p.exitcode})")
    except Exception:
        print("spawn attempt:")
        traceback.print_exc()


if __name__ == "__main__":
    main()
```

jacobcbeaudin:
## What

Closes #9293.

Run-mode (thread) kernels share `sys.modules` with the host process. `patches.patch_main_module` replaces `sys.modules['__main__']` with a synthetic module and has no restore path, so after a `SessionMode.RUN` session ends, the host is left with the synthetic main. Any later code that reads `sys.modules['__main__']` (notably `multiprocessing.spawn.get_preparation_data` during subprocess bootstrap, and `ForkingPickler` during target pickling) gets the wrong module. Three existing test-layer workarounds in this repo (`test_asgi.py` setUp/tearDown, `test_session_manager.py::_preserve_main_module`, `conftest.py::_save_and_restore_main` from #9286) currently paper over the leak.

## How

- **`marimo/_runtime/patches.py`**: adds `save_main_module` / `restore_main_module`, refcounted with a `threading.Lock`. First save captures the host's original `__main__`; last release restores it. Unbalanced save leaks the refcount for the process lifetime; unbalanced restore is a no-op. Debug-level logs on capture and restore so forensic traces can locate the save/restore points on a polluted host.
- **`marimo/_session/managers/kernel.py`**:
  - `KernelManagerImpl.__init__` initializes a per-instance flag and `threading.Lock` that together make the wrappers idempotent against double `close_kernel` and safe under concurrent callers.
  - Private methods `_save_host_main_module`, `_restore_host_main_module`, `_stop_and_join_run_mode_kernel` encapsulate the run-mode bookkeeping. `start_kernel` / `close_kernel` become short call sites that delegate to these methods.
  - `close_kernel` now bounds the kernel thread join at 5 seconds; on timeout it logs a warning and skips the `__main__` restore rather than touching process state under a live cell.
  - `start_kernel` exception path balances the save if thread setup fails.
- **`tests/_runtime/test_patches.py`**: 6 unit tests covering the refcount primitive (basic cycle, overlapping savers, over-release no-op, concurrent thread safety, absent `__main__` edge case, end-to-end host spawn after save/patch/restore).
- **`tests/_runtime/_patches_spawn_target.py`**: tiny helper module hosting a top-level `noop_target` function used by the end-to-end spawn test.
- **`tests/_session/managers/test_kernel.py`**: 7 unit tests covering the application-layer wrappers (flag transition, idempotent save, save-without-restore leak contract, idempotent restore, concurrent-restore safety, restore-without-save no-op, full cycle integration).

## Scope

- **EDIT-mode** and **IPC-mode** kernel paths are unchanged (their mutation is contained in a subprocess that exits).
- The **pyodide** `patch_main_module` call site is unchanged (WASM environment with no persistent host to pollute).
- The **process branch** of `close_kernel` is unchanged (subprocess is terminated; host state restore is not meaningful).

Existing test-layer workarounds (`test_asgi.py`, `test_session_manager.py::_preserve_main_module`, `conftest.py::_save_and_restore_main` from #9286) are left in place; they can be retired in a follow-up once this lands.

## Testing

Run on Linux (`python:3.12.3-slim`, `uv sync --group dev --group test`):

- `tests/_runtime/test_patches.py`: all pass (11 pre-existing + 6 new; 1 skipped for polars).
- `tests/_session/managers/test_kernel.py`: 7 new tests pass.
- `tests/_session/managers/`: 23 pre-existing tests still pass.
- `tests/_messaging/test_thread_local_proxy.py`: 21 tests pass.
- `tests/_server/export/test_exporter.py::test_run_until_completion_with_stop`, `::test_run_until_completion_with_console_output`: pass.

Not run: full `tests/` suite, `tests/_islands/`, `tests/_server/test_asgi.py`, `tests/_server/test_session_manager.py`, frontend-dependent HTML export tests (need `marimo/_static/index.html` which is not produced by plain `uv sync`).

## Notes for reviewers

- Refcount pattern mirrors `marimo/_messaging/thread_local_streams.py` for consistency; I kept the module-level globals style rather than introducing a class wrapper.
- Alternative implementations considered and rejected:
  - Try/finally wrapping `launch_kernel`'s body (~290-line indent change; LIFO only).
  - `patch_main_module_context` around the kernel main loop (same indent cost; LIFO only).

  The refcount approach handles overlapping server sessions correctly, which the existing TODO inside `patch_main_module` already flags as a concern.
- Happy to adjust naming, split tests, drop defensive ones, or reshape structure on review.

### DIFF
--- a/marimo/_runtime/patches.py
+++ b/marimo/_runtime/patches.py
@@ -15,9 +15,9 @@ from marimo import _loggers
 from marimo._ast.parse import ast_parse
 from marimo._dependencies.dependencies import DependencyManager
 from marimo._runtime import marimo_browser, marimo_pdb
+from marimo._utils.platform import is_pyodide
 
 LOGGER = _loggers.marimo_logger()
-from marimo._utils.platform import is_pyodide
 
 Unpatch = Callable[[], None]
 
--- a/marimo/_runtime/patches.py
+++ b/marimo/_runtime/patches.py
@@ -6,13 +6,17 @@ import contextlib
 import functools
 import sys
 import textwrap
+import threading
 import types
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
+from marimo import _loggers
 from marimo._ast.parse import ast_parse
 from marimo._dependencies.dependencies import DependencyManager
 from marimo._runtime import marimo_browser, marimo_pdb
+
+LOGGER = _loggers.marimo_logger()
 from marimo._utils.platform import is_pyodide
 
 Unpatch = Callable[[], None]
@@ -58,6 +62,58 @@ def patch_webbrowser() -> None:
 
 def patch_sys_module(module: types.ModuleType) -> None:
     sys.modules[module.__name__] = module
+
+
+# Refcount-based save/restore for the process's ``sys.modules['__main__']``.
+# Run-mode (thread) kernels share sys.modules with the host; pairing every
+# ``patch_main_module`` with save/restore keeps the host's main module from
+# leaking the kernel's synthetic module after the session ends.
+_main_save_lock = threading.Lock()
+_main_save_count = 0
+_main_original: types.ModuleType | None = None
+
+
+def save_main_module() -> None:
+    """Pair with ``restore_main_module`` to scope a ``patch_main_module`` call.
+
+    The first call in a process captures the current ``sys.modules['__main__']``;
+    subsequent overlapping calls share that capture and the last paired
+    ``restore_main_module`` puts the original back.
+
+    Callers must balance save/restore. An unbalanced save increments the
+    refcount permanently for the process lifetime and prevents the original
+    ``__main__`` from ever being restored; an unbalanced restore is a no-op.
+    """
+    global _main_save_count, _main_original
+    with _main_save_lock:
+        if _main_save_count == 0:
+            _main_original = sys.modules.get("__main__")
+            LOGGER.debug(
+                "save_main_module captured original __main__=%r",
+                _main_original,
+            )
+        _main_save_count += 1
+
+
+def restore_main_module() -> None:
+    """Release a reference acquired by ``save_main_module``.
+
+    Restores the captured original ``sys.modules['__main__']`` when the last
+    outstanding reference is released. Calling with no outstanding reference
+    is a no-op.
+    """
+    global _main_save_count, _main_original
+    with _main_save_lock:
+        if _main_save_count == 0:
+            return
+        _main_save_count -= 1
+        if _main_save_count == 0 and _main_original is not None:
+            sys.modules["__main__"] = _main_original
+            LOGGER.debug(
+                "restore_main_module restored __main__ to %r",
+                _main_original,
+            )
+            _main_original = None
 
 
 def patch_pyodide_networking() -> None:
--- a/marimo/_session/managers/kernel.py
+++ b/marimo/_session/managers/kernel.py
@@ -32,6 +32,11 @@ if TYPE_CHECKING:
 
 LOGGER = _loggers.marimo_logger()
 
+# Seconds to wait for the RUN-mode kernel thread to exit after we send
+# StopKernelCommand. Long enough that a cooperative kernel winds down
+# cleanly, short enough that a stuck kernel does not block a server.
+_THREAD_KERNEL_JOIN_TIMEOUT_S = 5.0
+
 
 class KernelManagerImpl(KernelManager):
     """Kernel manager using multiprocessing Process or threading Thread.
@@ -62,6 +67,13 @@ class KernelManagerImpl(KernelManager):
         # Only used in edit mode
         self._read_conn: TypedConnection[KernelMessage] | None = None
         self._virtual_file_storage = virtual_file_storage
+
+        # Tracks an outstanding save_main_module awaiting its paired
+        # restore. Guarded by _main_save_lock so concurrent close_kernel
+        # or start_kernel calls on the same instance cannot double-release
+        # the __main__ refcount.
+        self._main_save_lock: threading.Lock = threading.Lock()
+        self._main_save_outstanding: bool = False
 
     def start_kernel(self) -> None:
         # We use a process in edit mode so that we can interrupt the app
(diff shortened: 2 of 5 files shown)


akshayka (commented):
Hey @jacobcbeaudin, the CI issues are resolved, so we (marimo team) no longer need this fix urgently. 

Can you help me understand your motivation for attempting to fix this? Are you hitting this bug in real usage of marimo, or in tests?

If you are hitting it in real usage I will consider fixing it myself. This is not an easy area of marimo to contribute to.

jacobcbeaudin:
@akshayka thanks for the review, and understood on the complexity of this area. Motivation for the fix is on the issue: https://github.com/marimo-team/marimo/issues/9293#issuecomment-4291908334.

Given #9304 is in flight and your point about this being a delicate area, closing this PR. If useful after #9304 lands, happy to contribute the two orthogonal pieces (end-to-end spawn-after-cycle test and the bounded `close_kernel` join) as separate small follow-ups.

PR #9317, opened 2026-04-22
ci: try enabling xdist on windows

akshayka:
Recent changes to CI appear to have fixed Windows under xdist. Perhaps the global save-and-restore-main fixture.


mscolnick (approved):
nice! this is great

PR #9328, opened 2026-04-22
ci: don't cancel in-flight docs and release runs on main

akshayka:
Match the behavior from #9312: on pushes to main, use a per-run concurrency group so a newer push never cancels an older one. Cancelled workflows show as a red X on the commit in the GitHub UI, which makes the UI not useful. For docs we also want to guarantee the build actually completes.

To keep this safe without cancellation, each workflow re-checks that its SHA is still the tip of main before shipping artifacts and skips the final step if it's been superseded, so a stale older run can't clobber a newer deploy/publish. This keeps the workflow status green.

- docs.yml: per-run group, gate the Vercel deploy on tip-of-main
- release.yml: per-run group on main only (tags keep cancel-in-progress), gate TestPyPI publishes before and after build_dev
- release-dev.yml: recheck tip-of-main before the npm publishes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,8 +13,11 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # On main: unique group per run so nothing is cancelled (avoids a red X on the commit
+  # when a newer push supersedes this one). On tags: per-ref group with cancel-in-progress,
+  # preserving the original production-release behavior.
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -72,19 +75,67 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      is_latest: ${{ steps.check.outputs.is_latest }}
+    steps:
+      - id: check
+        env:
+          GH_SHA: ${{ github.sha }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          LATEST_SHA=$(git ls-remote "https://github.com/${GH_REPO}.git" refs/heads/main | awk '{print $1}')
```
dmadisetti, line 127: V low chance of RC, but good enough

dmadisetti (approved):

PR #9330, opened 2026-04-22
fix: harden trust-bearing window globals and gate script loading

mscolnick:
Follow-up to #9318. Extends the frozen first-party marker pattern to
`__MARIMO_STATIC__` (backs the virtual-file allowlist) and
`__MARIMO_MOUNT_CONFIG__`, and adds a shared `hasTrustedNotebookContext()`
predicate so `RenderHTML.replaceSrcScripts` refuses notebook-authored
`<script src>` in untrusted edit mode before any user interaction.

### DIFF
--- a/frontend/src/core/static/static-state.ts
+++ b/frontend/src/core/static/static-state.ts
@@ -5,20 +5,44 @@ import type { MarimoStaticState, StaticVirtualFiles } from "./types";
 
 declare global {
   interface Window {
-    __MARIMO_STATIC__?: MarimoStaticState;
+    __MARIMO_STATIC__?: Readonly<MarimoStaticState>;
   }
 }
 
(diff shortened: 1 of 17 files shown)


frontend/src/core/static/static-state.ts
```diff
@@ -5,20 +5,44 @@ import type { MarimoStaticState, StaticVirtualFiles } from "./types";
...
+    return false;
+  }
+  if (
+    candidate.modelNotifications !== undefined &&
+    !Array.isArray(candidate.modelNotifications)
+  ) {
+    return false;
+  }
+  return true;
+}
```
peter-gy, line 43: should we also validate that `files` is a string-to-string map, instead of only checking if it's an object?

peter-gy (approved):

PR #9331, opened 2026-04-22
fix: `app.run()` memory leak

### LINKED ISSUE #9198: Memory is not released when calling notebook as script
### Describe the bug

I'm working on a fairly long notebook A, which at the moment takes nearly all of my RAM. To keep going, I wanted to create a separate notebook B, call (not embed) notebook A and extract some variables I need, and keep going from there.

For some reason however, it seems that with importing and calling `app.run()`, the memory used to run A is not released once done -- I did try calling `gc.collect()` but nothing changed, and I even made a separate python module external to B trying to call A from there and just return my used variables. Still, at the end of the function, the memory allocated persists.

My current solution will be to add a new step to A to serialize the variables I might need, and load them from B, but it would be much nicer if calling A directly would allow me to just take what I need from it and then free the rest of the memory. Perhaps there's already a way to do so, but unfortunately I have not found it.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.0",
  "editable": false,
  "OS": "Linux",
  "OS Version": "6.8.0-107-generic",
  "Processor": "x86_64",
  "Python Version": "3.13.1",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.11.0",
    "uv": "0.6.10"
  },
  "Dependencies": {
    "click": "8.3.2",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.19.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.44.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "loro": "1.10.3",
    "pandas": "3.0.2"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

dmadisetti:
## 📝 Summary

Closes #9198

Since module scopes are persistent, the continual use of `app.run()` progressively fills up memory even after user accessible variables are not longer present. This PR introduces reference tracking on the `_Namespace` mapping class which clears the module dictionary on cleanup.

cc @VishakBaddur: I just saw your comments on the initial issue. Good job on determining the root cause, but I think context manager is unneeded. The reference tracking here should silently cleanup without explicit management from the user.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -6,6 +6,7 @@ import inspect
 import os
 import sys
 import threading
+import weakref
 from collections.abc import (
     Callable,
     Iterable,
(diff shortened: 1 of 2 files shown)


akshayka:
@dmadisetti this seems over-engineered? If there is a leak, it might be because if this reference cycle:  https://github.com/marimo-team/marimo/blob/main/marimo/_ast/app.py#L621-L626

The `_owner` attribute appears unused ...

dmadisetti:
@akshayka `_Namespace` is cleaned up, the module namespace outlives everything. My initial fix is maybe the easiest solution: https://github.com/marimo-team/marimo/pull/9331/changes/c065beddb7d2e6d2c13ee809f704d729832274e9

and we can go with that. We just have the scoping information, so we can leverage it (for relatively free).

akshayka:
@dmadisetti I looked at your commit and I don't understand. `glbls` is just a Python dict defined in `run` and not bound to any instance variables. Is there an underlying side-effect that needs to be eliminated? I don't see any "persistent" module.

EDIT: marimo pair brought me up to speed, this is more involved than I thought. https://molab.marimo.io/notebooks/nb_ULHALbwMsQzjB24qU1bCaC

akshayka:
Hmm. Not sure how we can fix #9198 while retaining Python semantics ... I think your first commit might have unintended side-effects, clearing globals when objects (such as values pulled out of the returned `defs`) still reference them seems unsafe. While this PR also looks brittle. It feels like a proper fix would have to change how we execute cells, to prevent the entire notebook's globals from being saved on its returned definitions.

dmadisetti:
> I think your first commit might have unintended side-effects

Absolutely agree. The follow-up fix cleans up everything it can (if `defs` gets dropped, everything not tracked is dropped). Then as tracked objects are cleaned up the dependency chain is cleaned up too.

>  It feels like a proper fix would have to change how we execute cells, to prevent the entire notebook's globals from being saved on its returned definitions.

It seems to me, we always want the full notebook to run (it may have side effects etc). Retaining all variables also make sense since we don't know what the user will want. But curious if you have other ideas around this?

akshayka:
> The follow-up fix cleans up everything it can (if defs gets dropped, everything not tracked is dropped). Then as tracked objects are cleaned up the dependency chain is cleaned up too.

My concern is that it feels like there would be many edge cases that could get confusing, for example if user code mutated things in ways that was not trackable by static analysis / required refs. So I'm not sure if this can be done 100% correctly.

I was thinking we might just recommend to users that if they want reusability while controlling memory consumption, they opt for reusable functions instead of `app.run()`. `app.run()` is basically the same thing as `import my_regular_python_module` in that it executes the full notebook like you say. In other words, we may be able to close the original issue by making a docs change.

dmadisetti:
> My concern is that it feels like there would be many edge cases that could get confusing, for example if user code mutated things in ways that was not trackable by static analysis / required refs. So I'm not sure if this can be done 100% correctly.

Hmm the only edgecase I can think of is `globals()["value"]`, which breaks marimo's reactivity model anyway. FFI isn't a concern here because we're just removing the ref count to the namespace dict, even then only when we're trying to clean up the object anyway. Let me know if I'm missing something. I agree on the docs change regardless

PR #9344, opened 2026-04-23
fix: fix file deletion race in file watcher

akshayka:
This fixes a race condition in the file watcher in which the file was deleted after an existence check.


akshayka:
Failure is unrelated

mscolnick (approved):

PR #9349, opened 2026-04-23
feat: editable filter pills

kirangadhave:
## 📝 Summary

- Click a column-filter pill in a data-table to open an inline
  editor: pick column, operator, and value; Apply / Close / Remove.
- Editable types: number / text / boolean / select. Date/datetime/time
  remain read-only — separate PR.
- Also: new `FilterByValuesPicker` (controlled top-K dropdown, used in
  the select value slot); `Combobox` forwards `id` onto its trigger so
  `<label htmlFor>` works.

## Scope

- `filter-pill-editor.tsx` (new): popover with Combobox / Select /
  value slot. Snapshot-based rehydration restores the pre-edit values
  when the user navigates back to the same column + operator.
- `filter-by-values-picker.tsx` (new): fully-controlled, emits on
  every toggle. Intended to replace the column-header popover in a
  follow-up.
- `filter-pills.tsx`: three-segment pill (col | op | val),
  focusable button trigger, `calculateTopKRows` threaded through to
  the select value slot.
- `combobox.tsx`: new optional `id` prop forwarded to the trigger.

https://github.com/user-attachments/assets/5722c3d9-5023-4c40-838e-ccc0cb9912eb

---
## Summary by cubic
Make filter pills editable via an inline popover editor. Users can change column, operator, and value directly from the pill; date/datetime/time pills remain read‑only.

- **New Features**
  - Inline editor on pill click with Apply, Close, and Remove actions.
  - Supported types: number, text, boolean, select; others render read‑only.
  - Three-segment pills show column | operator | value for quick scanning.
  - New `FilterByValuesPicker`: controlled top‑K values dropdown for select filters, powered by `calculateTopKRows`.
  - `DataTable` accepts `calculateTopKRows` and passes it to pills; `DataTablePlugin` wires `calculate_top_k_rows`.
  - `Combobox` forwards `id` to its trigger to enable `<label htmlFor>` association.

- **Bug Fixes**
  - `FilterByValuesPicker`: fixed async deps; smarter search (smartMatch + includes); Select All affects only visible items with indeterminate state; correct truncation notice; stable keys.
  - `FilterPillEditor`: `is_null`/`is_not_null` preserve column type across round‑trips; `Checkbox` now supports an indeterminate state (minus icon) used by Select All.

<sup>Written for commit 7f2a1f241b3e3ea7bd301af16807d99508683698. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/components/data-table/filter-by-values-picker.tsx
+++ b/frontend/src/components/data-table/filter-by-values-picker.tsx
@@ -1,0 +1,225 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+"use no memo";
+
+import type { Column } from "@tanstack/react-table";
+import { ChevronDownIcon } from "lucide-react";
+import { useMemo, useState } from "react";
+import { useAsyncData } from "@/hooks/useAsyncData";
+import { ErrorBanner } from "@/plugins/impl/common/error-banner";
+import type { CalculateTopKRows } from "@/plugins/impl/DataTablePlugin";
+import { cn } from "@/utils/cn";
+import { Logger } from "@/utils/Logger";
+import { Spinner } from "../icons/spinner";
+import { Button } from "../ui/button";
+import { Checkbox } from "../ui/checkbox";
+import {
+  Command,
+  CommandEmpty,
+  CommandInput,
+  CommandItem,
+  CommandList,
+} from "../ui/command";
+import { Popover, PopoverContent, PopoverTrigger } from "../ui/popover";
+import { SentinelCell } from "./sentinel-cell";
+import { detectSentinel, stringifyUnknownValue } from "./utils";
+
+const TOP_K_ROWS = 30;
+
+interface Props<TData, TValue> {
+  column: Column<TData, TValue>;
+  calculateTopKRows?: CalculateTopKRows;
+  chosenValues: unknown[];
+  onChange: (values: unknown[]) => void;
+}
+
+export const FilterByValuesPicker = <TData, TValue>({
+  column,
+  calculateTopKRows,
+  chosenValues,
+  onChange,
+}: Props<TData, TValue>) => {
+  const [open, setOpen] = useState(false);
+
+  const chosenValuesSet = useMemo(
+    () => new Set(chosenValues),
+    [chosenValues],
+  );
+
+  const selectedValuesStr = useMemo(() => {
+    if (chosenValuesSet.size === 0) {
+      return "Select values…";
+    }
+    const items = [...chosenValuesSet].map((v) =>
+      stringifyUnknownValue({ value: v }),
+    );
+    return `[${items.join(", ")}]`;
+  }, [chosenValuesSet]);
+
+  return (
+    <Popover open={open} onOpenChange={setOpen}>
+      <PopoverTrigger asChild={true}>
+        <Button
+          variant="outline"
+          size="xs"
+          className="h-6 mb-1 w-full justify-between font-normal"
+        >
+          <span
+            className={cn(
+              "truncate",
+              chosenValuesSet.size === 0 && "text-muted-foreground",
+            )}
+          >
+            {selectedValuesStr}
+          </span>
+          <ChevronDownIcon className="h-4 w-4 opacity-50 shrink-0" />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent className="w-80 p-0">
+        <PickerBody
+          column={column}
+          calculateTopKRows={calculateTopKRows}
+          chosenValues={chosenValuesSet}
+          onChange={onChange}
+        />
+      </PopoverContent>
+    </Popover>
+  );
+};
+
+interface PickerBodyProps<TData, TValue> {
+  column: Column<TData, TValue>;
+  calculateTopKRows?: CalculateTopKRows;
+  chosenValues: Set<unknown>;
+  onChange: (values: unknown[]) => void;
+}
+
+const PickerBody = <TData, TValue>({
+  column,
+  calculateTopKRows,
+  chosenValues,
+  onChange,
+}: PickerBodyProps<TData, TValue>) => {
+  const [query, setQuery] = useState<string>("");
+
+  const { data, isPending, error } = useAsyncData(async () => {
+    if (!calculateTopKRows) {
+      return null;
+    }
+    const res = await calculateTopKRows({ column: column.id, k: TOP_K_ROWS });
+    return res.data;
+  }, []);
+
+  const filteredData = useMemo(() => {
+    if (!data) {
+      return [];
+    }
+    try {
+      return data.filter(([value, _count]) =>
+        value === undefined
+          ? false
+          : String(value).toLowerCase().includes(query.toLowerCase()),
+      );
+    } catch (error_) {
+      Logger.error("Error filtering data", error_);
+      return [];
+    }
+  }, [data, query]);
+
+  const handleToggle = (value: unknown) => {
+    const next = new Set(chosenValues);
+    if (next.has(value)) {
+      next.delete(value);
+    } else {
+      next.add(value);
+    }
+    onChange([...next]);
+  };
+
+  const allChecked = chosenValues.size === filteredData.length;
+
+  const handleToggleAll = () => {
+    if (!data) {
+      return;
+    }
+    onChange(allChecked ? [] : filteredData.map(([value]) => value));
+  };
+
+  if (isPending) {
+    return <Spinner size="medium" className="mx-auto mt-12 mb-10" />;
+  }
+
+  if (error) {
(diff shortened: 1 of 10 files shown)


frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -0,0 +1,225 @@
...
+}: PickerBodyProps<TData, TValue>) => {
+  const [query, setQuery] = useState<string>("");
+
+  const { data, isPending, error } = useAsyncData(async () => {
+    if (!calculateTopKRows) {
+      return null;
+    }
+    const res = await calculateTopKRows({ column: column.id, k: TOP_K_ROWS });
+    return res.data;
+  }, []);
```
mscolnick: `column.id` in deps

mscolnick: also maybe `calculateTopKRows` since that might change if the dataframe changes

frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  const filteredData = useMemo(() => {
+    if (!data) {
+      return [];
+    }
+    try {
+      return data.filter(([value, _count]) =>
+        value === undefined
+          ? false
+          : String(value).toLowerCase().includes(query.toLowerCase()),
```
mscolnick: we have `smartMatch` logic somewhere if you want to use that

frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -0,0 +1,225 @@
...
+  }, [data, query]);
+
+  const handleToggle = (value: unknown) => {
+    const next = new Set(chosenValues);
+    if (next.has(value)) {
+      next.delete(value);
+    } else {
+      next.add(value);
+    }
+    onChange([...next]);
```
mscolnick: we have `arrayToggle()`. or could add this to `sets.ts`

frontend/src/components/data-table/filter-by-values-picker.tsx
kirangadhave: column and calculateTopKRows wouldn't change when the Picker is open

kirangadhave: Actually you're right, it's _correct_ to have them in dep array, so i'll add it.

frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -0,0 +1,218 @@
...
+  };
+
+  const allChecked = chosenValues.size === filteredData.length;
+
+  const handleToggleAll = () => {
+    if (!data) {
+      return;
+    }
+    onChange(allChecked ? [] : filteredData.map(([value]) => value));
```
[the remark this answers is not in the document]
kirangadhave: oh this is a good catch

frontend/src/components/data-table/filter-pill-editor.tsx
```diff
@@ -0,0 +1,464 @@
...
+  type,
+  operator,
+  draft,
+}: {
+  type: EditableFilterType;
+  operator: UiOperator;
+  draft: DraftValue;
+}): ColumnFilterValue | undefined {
+  if (operator === "is_null" || operator === "is_not_null") {
+    return Filter.text({ operator: operator as OperatorType });
```
mscolnick: does this need to handle numbers? if `type === "number"`, then `Filter.number({ operator: operator as OperatorType });`

frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -0,0 +1,225 @@
...
+    const next = new Set(chosenValues);
+    if (next.has(value)) {
+      next.delete(value);
+    } else {
+      next.add(value);
+    }
+    onChange([...next]);
+  };
+
+  const allChecked = chosenValues.size === filteredData.length;
```
mscolnick: maybe rename to `allVisibleChecked`? also would be true when `filteredData.length === 0`

mscolnick: (i guess this is an existing bug?)

frontend/src/components/data-table/filter-pill-editor.tsx
kirangadhave: oh yeah, this will mess up the operators if i keep returning `text` always

frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -0,0 +1,238 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+"use no memo";
```
Light2Dark, line 2: is it possible for these files to not have `use no memo` ? It might avoid the need to memoize everything manually.

frontend/src/components/data-table/filter-pill-editor.tsx
```diff
@@ -0,0 +1,471 @@
...
+    return "text";
+  }
+  if (value.type === "boolean") {
+    return "boolean";
+  }
+  if (value.type === "select") {
+    return "select";
+  }
+  // date/datetime/time fall back to text; callers should guard. supported in future
+  return "text";
```
Light2Dark, line 373: One improvement for future maintenance is to use `logNever(value.type)` and exhaustively check each case. So the typechecker will warn when a new `type` has been added but we did not add the handling here.

Same for getUiOperator

frontend/src/components/data-table/filter-pill-editor.tsx
```diff
@@ -0,0 +1,471 @@
...
+    return Filter.text({
+      text: draft.text,
+      operator: "contains" as OperatorType,
+    });
+  }
+  if (type === "boolean") {
+    if (operator === "is_true") {
+      return Filter.boolean({
+        value: true,
+        operator: "is_true" as OperatorType,
```
Light2Dark: can we remove the `as OperatorType` assertions?

Light2Dark (commented):
Looks great! 

1. I think there's too much vertical space between the filter pills and the search bar, imo can reduce it.
 
2. Might be an existing bug, if I have a `filter by values` and try to add another filter, it shows up empty here.

Image

3. If I apply an empty min/max, it ends up as NaN. I think it should just be no min/no max. Not sure

Image

4. Not sure what this filter button means

Image

5. Nit, I think the X icon in the filter pill should just be bolder when you hover on it, doesn't need to have a red circle background. Because the pill is already blue when hovering on it.
image

kirangadhave:
Interesting, the vertical space is for sure a bug. I'm the video you can see the space is not quite that big.

Light2Dark:
> Interesting, the vertical space is for sure a bug. I'm the video you can see the space is not quite that big.

It looks kinda large in your video too
image

kirangadhave:
> It looks kinda large in your video too

I see what you mean. I was looking at something else

frontend/src/components/data-table/filter-by-values-picker.tsx
kirangadhave: tanstack table can break without "use no memo"

frontend/src/components/data-table/filter-pill-editor.tsx
kirangadhave: I will add this when i add support for full set of operators in my follow up PR

kirangadhave:
> If I apply an empty min/max, it ends up as NaN. I think it should just be no min/no max. Not sure

Hmm I'm not able to replicate this
Screenshot 2026-04-24 at 10 29 35 AM

kirangadhave:
> Not sure what this filter button means

this is an existing feature. I think it's only supposed to indicate that the column is filtered on currently. I will improve it's handling and visuals when i work on the actual filter popover PR.

kirangadhave:
> Might be an existing bug, if I have a filter by values and try to add another filter, it shows up empty here.

Existing bug. This will go away when i add a filter popup. The Filter by Values option will just become the `in` operator in filter.

Light2Dark:
> > If I apply an empty min/max, it ends up as NaN. I think it should just be no min/no max. Not sure
> 
> Hmm I'm not able to replicate this 

I could only reproduce it when min has a value, and then I remove the min value and click apply, then it shows up as NaN. Im okay to fix this in follow-ups though, I can have a look too

kirangadhave:
> I could only reproduce it when min has a value, and then I remove the min value and click apply, then it shows up as NaN. Im okay to fix this in follow-ups though, I can have a look too

this will go away most probably. When i expand operator support I will add dedicated >, <, etc. operators. so between will always take two args

Light2Dark (approved):

PR #9350, opened 2026-04-23
Correlate scratchpad completion with `run_id`

### LINKED ISSUE #9255: marimo pair: Claude Code doesn't see cell errors automatically

manzt:
Closes #9302 
Fixes #9255, and flips the 4 xfail integration tests from #9342 flip to passing.

The ``ScratchCellListener`` used to fire its done sentinel on the scratch cell's ``idle`` status (+ 50ms grace for flushing stderr/stdout). Anything broadcast after that grace  was silently dropped from the SSE stream.

These changes add an optional `run_id` to ``ExecuteScratchpadCommand`` and ``CompletedRunNotification``. The api endpoint and MCP mint a UUID and pass it to both the command and the ``ScratchCellListener``; the listener now fires its sentinel only on the matching ``CompletedRunNotification``. 

Note: The reason we couldn't just observe `CompletedRunNotification` directly (#9302) is because the ``CompletedRunNotificationn`` from ``session.instantiate`` trips up the listener early (separate run).

---
## Summary by cubic
Correlate each scratchpad run with a `run_id` and wait for the matching `CompletedRunNotification` before finishing. This prevents early success in `/api/kernel/execute`, streams compile-time errors, and simplifies the `done` SSE.

- **Bug Fixes**
  - Add optional `run_id` to `ExecuteScratchpadCommand` and `CompletedRunNotification`; `/api/kernel/execute` and the MCP code server mint a UUID and `ScratchCellListener` waits for the matching completion, ignoring others.
  - Always emit `CompletedRunNotification` in a `finally` so listeners don’t hang if `run_scratchpad` raises.
  - Stream compile-time errors to `stderr` and keep console output, so `SyntaxError` diagnostics are visible before `done`.
  - Add integration tests for `ctx.create_cell` validation and the `skip_validation=True` path to cover early `RuntimeError` reporting and graph-state failures.

- **Migration**
  - `done` SSE now returns `{ success, output }` only; the `error` field was removed. On failure `output` is `{ mimetype: "text/plain", data: "" }`.
  - OpenAPI and generated TS types updated: `CompletedRunNotification.run_id?`, `ExecuteScratchpadCommand.runId?`.

<sup>Written for commit a01e1522215b3a46890e5697d0b695a89ace7290. Summary will update on new commits.</sup>

### DIFF
--- a/tests/_server/test_scratchpad_integration.py
+++ b/tests/_server/test_scratchpad_integration.py
@@ -22,10 +22,10 @@ fresh session via a per-test ``session`` fixture that calls
 snapshotted as SSE lines with volatile paths and version-specific
 traceback noise scrubbed.
 
-Each test's snapshot encodes the **correct** expected output.
-Scenarios currently broken (see issue #9255) are marked
-``xfail(strict=True)`` so a fix turns them into ``XPASS`` failures
-and the fix PR must remove the markers.
+Each test's snapshot encodes the expected SSE output for that
+scenario — failure cases land in ``done`` with
+``{success: false, output: {mimetype: "text/plain", data: ""}}``;
+error detail arrives earlier in the stream as ``stderr`` SSE events.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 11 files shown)


tests/_server/test_scratchpad_integration.py
```diff
@@ -572,10 +616,68 @@ def test_ctx_run_cell_cascade_error(session: _Session) -> None:
...
+    assert lines == snapshot(
+        [
+            "event: stdout",
+            'data: {"data": "created cell \'<cid>\'\\n"}',
+            "",
+            "event: done",
+            'data: {"success": false, "output": {"mimetype": "text/plain", "data": ""}}',
+            "",
+        ]
+    )
```
manzt, line 711: @mscolnick some drive-by, additional code mode tests.

mscolnick (dismissed):

mscolnick (approved):

PR #9365, opened 2026-04-24
fix: decode tuple/frozenset payloads with non-finite floats

mscolnick:
Python's `json.dumps` emits bare `NaN`/`Infinity`/`-Infinity` for non-finite floats inside embedded tuple/frozenset payloads (e.g. `text/plain+tuple:[NaN]`). The outer JSON still parses because those tokens live inside a JSON string, but `formatTuplePayload` and `formatFrozensetPayload` used strict `JSON.parse` and threw.

Swap both to `jsonParseWithSpecialChar` (same parser we already use at the outer level) and render non-finite elements as `float('nan')` / `float('inf')` / `-float('inf')`, matching the scalar-key form.

### DIFF
--- a/frontend/src/components/editor/output/JsonOutput.tsx
+++ b/frontend/src/components/editor/output/JsonOutput.tsx
@@ -18,6 +18,7 @@ import { memo, useState } from "react";
 import type { OutputMessage } from "@/core/kernel/messages";
 import { cn } from "@/utils/cn";
 import { copyToClipboard } from "@/utils/copy";
+import { jsonParseWithSpecialChar } from "@/utils/json/json-parser";
 import { isUrl } from "@/utils/urls";
 import { useTheme } from "../../../theme/useTheme";
 import { logNever } from "../../../utils/assertNever";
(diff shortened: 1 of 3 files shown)


frontend/src/components/editor/output/JsonOutput.tsx
```diff
@@ -386,14 +387,35 @@ function renderLeaf(leaf: string, render: LeafRenderer): React.ReactNode {
...
+// so render them as the same `float(...)` literals we use for scalar
+// float keys (see `decodeKeyForCopy`).
+function formatCollectionItems(items: unknown[]): string {
+  return items
+    .map((x) => {
+      if (typeof x === "number" && !Number.isFinite(x)) {
+        if (Number.isNaN(x)) {
+          return "float('nan')";
+        }
+        return x > 0 ? "float('inf')" : "-float('inf')";
```
Light2Dark, line 402: do you know if this affects `pd.NaT`, `np.nan`

Light2Dark (approved):

PR #9376, opened 2026-04-24
Scope filter-by-values top-K to exclude the filter being edited

kirangadhave:
## Summary
- Re-opening a "filter by values" pill re-computed the picker's top-K list against the already-filtered rows, so values excluded by the current filter vanished from the picker. Narrowing `role` to `[admin, user]` hid `guest` entirely — the user could no longer broaden the filter back out without clearing it.
- `_calculate_top_k_rows` now strips conditions targeting the requested column before computing top-K. Filters on other columns and the active search query still narrow the result, so cross-column filter context is preserved.
- Fast path: when no active filter targets the requested column, the cached `_searched_manager` is reused — zero extra cost for the common case.

### Before
https://github.com/user-attachments/assets/ff33a5f0-d8ae-4332-9870-987f2d99446d

### After
https://github.com/user-attachments/assets/f34d590e-805e-49d3-b26c-d94825b03c4a

### DIFF
--- a/marimo/_plugins/ui/_impl/table.py
+++ b/marimo/_plugins/ui/_impl/table.py
@@ -227,6 +227,45 @@ _DATATYPE_TO_CATEGORY: dict[str, str] = {
 }
 
 
+def _filters_reference_column(
+    group: FilterGroup, target_column: ColumnName
+) -> bool:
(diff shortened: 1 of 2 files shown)


kirangadhave:
@cubic

Light2Dark (approved):

PR #9377, opened 2026-04-24
POC: add --execute flag to marimo export html-wasm

dmadisetti:
## 📝 Summary

Run the notebook during `marimo export wasm-html --execute` and embed cell outputs as a session snapshot so users see pre-computed results before Pyodide loads. Requires a small FE change to preserve snapshots on kernel, but left it scoped to wasm.

Importantly, `--execute` leverages the pyodide packages lock and python version expected in WASM. This has the added advantage we can check for incompatible packages on export and warn accordingly.

---

I ran into the issue of incompatible server vs wasm packages when looking at sharing cache artifacts and thought that session previews were probably an ideal place to start for introducing `--execute`.

cc @peter-gy


marimo/_cli/export/pyodide_constraints.py
```diff
@@ -0,0 +1,163 @@
...
+LOGGER = _loggers.marimo_logger()
+
+# Pyodide version matching frontend/package.json — update together.
+PYODIDE_VERSION = "0.27.7"
+
+# Derived from the lockfile's info.python field (Pyodide 0.27.7 → 3.12.7).
+PYODIDE_PYTHON_VERSION = "3.12"
+
+_LOCKFILE_URL = (
+    "https://cdn.jsdelivr.net/pyodide/v{version}/full/pyodide-lock.json"
```
mscolnick, line 21: we should our lockfile: `https://wasm.marimo.app/pyodide-lock.json?v={marimo_version}&pyodide=v{pyodide_version}`

dmadisetti:
Following up with a more scoped PR

PR #9381, opened 2026-04-24
feat: prompt user to build fe at first run

ouatu-ro:
When `marimo edit` is run from a repo checkout without built frontend assets, the current error is `index.html not found and no asset_url configured`, which is confusing for contributors. This changes that path to return a more actionable message telling the user to run `make fe` and restart marimo.


ouatu-ro:

dmadisetti (approved):
Hey @ouatu-ro thanks for this. Looks like a nice QOL for new devs

PR #9386, opened 2026-04-25
fix: restore cell selection highlight for lazy/image cells in mo.ui.table

### LINKED ISSUE #9189: `mo.ui.table` visual selection does not work when cells are `lazy`
### Describe the bug

When I populate a table with `mo.lazy` cells, the selection logic works (`table.value` contains the correct cell) but the visual background indicator does not trigger when I select.

Without lazy:

Image

With lazy:
Image

Note that the visual selection still works when I click on the white gap between the image and the border of the cell.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
N/A: molab
```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "marimo>=0.23.1",
# ]
# ///

import marimo

__generated_with = "0.23.0"
app = marimo.App(
    width="medium",
    css_file="/usr/local/_marimo/custom.css",
    auto_download=["html"],
)


@app.cell
def _():
    import marimo as mo

    return (mo,)


@app.cell
def _(mo):
    table = mo.ui.table(
        [
            {
                "bird_1": (
                    mo.image(
                        "https://www.shutterstock.com/image-photo/hummingbirds-small-birds-that-important-600w-2460672183.jpg"
                    )
                ),
                "bird_2": (
                    mo.image(
                        "https://www.shutterstock.com/image-photo/blue-jay-bird-on-green-600w-2430565815.jpg"
                    )
                ),
            },
        ],
        selection="single-cell",
    )
    table
    return (table,)


@app.cell
def _(table):
    table.value
    return


if __name__ == "__main__":
    app.run()

```

VishakBaddur:
Closes #9189

## 📝 Summary

When `mo.ui.table` contains `mo.lazy(mo.image(...))` or `mo.image()` cells with `selection="single-cell"`, the visual highlight does not appear when clicking the image, even though `table.value` updates correctly.

**Root cause:** The existing highlight uses a `::before` pseudo-element with `z-index: -10` on the inner content div. `` elements(both inside shadow DOM (`mo.lazy`) and in light DOM (`mo.image`)), create browser compositing layers that always paint above `z-index: -10` absolutely-positioned elements. This is a fundamental browser rendering constraint, not a CSS specificity issue.

Verified in devtools:
- `td.style.backgroundColor` → invisible behind image ✗
- `td.style.boxShadow = 'inset ...'` → invisible behind image ✗
- `td.style.outline = '2px solid blue'` → visible above image ✓

This also explains why clicking the white gap/padding around the image DID show the highlight — the padding area has no compositing layer.

**Fix:** Apply the selection outline directly on the `<td>` element, which renders in the browser's decoration layer above all content including image compositing layers. One file changed, 4 lines added.

## Screenshots
Screenshot 2026-04-25 at 6 03 25 PM
Screenshot 2026-04-25 at 6 08 54 PM

Screenshot 2026-04-25 at 6 09 40 PM

## Pre-Review Checklist

- [x] All 462 frontend tests pass
- [x] `mo.lazy(mo.image(...))` - blue outline visible on selection
- [x] `mo.image(...)` without lazy - blue outline visible on selection  
- [x] Text cells - blue outline visible on selection
- [x] Dark mode verified (`--blue-7` resolves correctly)

### DIFF
--- a/frontend/src/components/data-table/renderers.tsx
+++ b/frontend/src/components/data-table/renderers.tsx
@@ -164,13 +164,17 @@ export const DataTableBody = <TData,>({
       );
 
       const title = cell.getHoverTitle?.() ?? undefined;
+      const isCellSelected = cell.getIsSelected?.() || false;
       return (
         <TableCell
           tabIndex={0}
           {...getCellDomProps(cell.id)}
           key={cell.id}
           className={cn(
-            "whitespace-pre truncate max-w-[300px] outline-hidden border-r border-r-border/75",
+            "whitespace-pre truncate max-w-[300px] border-r border-r-border/75",
+            isCellSelected
+              ? "outline outline-2 outline-(--blue-7) -outline-offset-2"
+              : "outline-hidden",
             cell.column.getColumnWrapping &&
               cell.column.getColumnWrapping?.() === "wrap" &&
               COLUMN_WRAPPING_STYLES,


Light2Dark (commented):
I believe the current selection highlight is from another component overlaying the cell. And it was for performance reasons. I'm not sure if this is performant.

VishakBaddur:
Thanks for the review @Light2Dark! You're right that the `::before` pseudo-element was the original approach.

The issue is that `img` elements (both inside `mo.lazy`'s shadow DOM and in plain `mo.image()` cells) can be promoted to their own compositing layers. When promoted, they paint above the `::before` highlight when implemented with a negative z-index, making the highlight invisible for image cells.

This also matches what I observed in devtools: background/box-shadow were hidden, while `outline` (which renders in the element's decoration layer) remained visible.

On performance: `cell.getIsSelected?.()` reads from already-computed TanStack state, it's the same call `columns.tsx` already makes per cell per render (line 269), and the O(n²) complexity is already acknowledged in the codebase (`feature.ts` line 69). For large tables, virtualization (`MIN_ROWS_TO_VIRTUALIZE = 100`) limits rendered cells anyway.

Happy to explore alternatives if there's a preferred pattern here, e.g., a CSS-only approach or an overlay strategy used elsewhere in the codebase.

Light2Dark:
I have a fix for this, it's due to the `use-cell-range-selection` hook. It treated all marimo elements as interactive, when it shouldn't.

VishakBaddur:
Thanks for digging into this @Light2Dark , the isInteractiveTarget() fix makes sense. I had approached it from the visual side (outline rendering above compositing layers), but fixing the click registration at the source is cleaner. Happy to close #9386 if this covers it.

PR #9401, opened 2026-04-27
Remove suggested solution from PR template

akshayka:
The problem statement is more interesting than the solution


kirangadhave (approved):
🚀

PR #9404, opened 2026-04-27
wasm: check platform instead of sys.modules for pyodide check

mscolnick:
Switch to use `sys.platform = "emscripten"` than potentially corrupted `sys.modules`


dmadisetti (approved):
Nice cleanup

PR #9405, opened 2026-04-27
refactor: consolidate `CellManager` and `NotebookDocument`

kirangadhave:
## Summary

`CellManager` now wraps a `NotebookDocument` instead of holding its own
dict of `CellData`, and `Session.document` is a property that reads
through `cell_manager.document`. Removes a class of drift bugs where the
two were synced once at session startup and then diverged across save,
file-watch reload, and code-mode autosave.

## What changed

- **`CellManager` wraps the document.** Removed `_cell_data` dict is gone; `get_cell_data` returns a fresh synthesized view each call.
- **`Session.document` is a `@property`** returning `app.cell_manager.document`. No more manual rebinding when the manager or app is swapped. Save-endpoint workaround removed.
- **Transitional helpers** preserve identity of `_document` and `_compiled_cells` for callers that still rebuild the cell list as a unit: `CellManager._replace_state_from` and `NotebookDocument._replace_cells`. `InternalApp.with_data` uses the former. A follow-up will route both through `Transaction.apply`.
- **File-watch reload bridge.** `AppFileManager.reload()` still swaps in a fresh app, so the coordinator now snapshots `prev_document` before the swap, builds the diff `Transaction`, and passes it to the strategy. A follow-up will rework `reload()` to apply the diff in place.

## Tests
Invariants for the document property, `get_cell_data` view freshness, `_replace_state_from` identity preservation, `_compiled_cells` lockstep rekeying (with an **xfail** for the latent inner `Cell._cell.cell_id` mismatch — separate follow-up), `_replace_cells` version bump, and `InternalApp.with_data` identity + compiled-cell carryover.

### DIFF
--- a/marimo/_ast/app.py
+++ b/marimo/_ast/app.py
@@ -294,7 +294,7 @@ class App:
             strict=False,
         ):
             cell = None
-            cell_data = self._cell_manager._cell_data.get(cell_id)
+            cell_data = self._cell_manager.get_cell_data(cell_id)
             new_cell_id = app._cell_manager.create_cell_id()
             # If the cell exists, the cell data should be set (ie not None).
             if cell_data is not None:
@@ -1014,23 +1014,31 @@ class InternalApp:
         names: Iterable[str],
         configs: Iterable[CellConfig],
     ) -> InternalApp:
-        new_cell_manager = CellManager()
+        """Rewrite the cell list from textual fields, in place.
+
+        Mutates ``self._app._cell_manager`` rather than replacing it,
+        so any caller holding ``app.cell_manager`` or
+        ``app.cell_manager.document`` (notably ``Session.document``)
+        continues to see live state without rebinding.
+
+        Cells that survive the rewrite keep their compiled ``Cell``;
+        callers from save flows pass the frontend's snapshot, which
+        renames/reorders/reconfigures cells but doesn't recompile.
+        """
+        cm = self._app._cell_manager
+        prev_compiled = dict(cm._compiled_cells)
+        rebuilt = CellManager(prefix=cm.prefix)
         for cell_id, code, name, config in zip(
             cell_ids, codes, names, configs, strict=False
         ):
-            cell = None
-            # If the cell exists, the cell data should be set.
-            cell_data = self._app._cell_manager._cell_data.get(cell_id)
-            if cell_data is not None:
-                cell = cell_data.cell
-            new_cell_manager.register_cell(
+            rebuilt.register_cell(
                 cell_id=cell_id,
                 code=code,
                 name=name,
                 config=config,
-                cell=cell,
+                cell=prev_compiled.get(cell_id),
             )
-        self._app._cell_manager = new_cell_manager
+        cm._replace_state_from(rebuilt)
         return self
 
     async def run_cell_async(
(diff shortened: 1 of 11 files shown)


marimo/_ast/app.py
```diff
@@ -1014,23 +1014,31 @@ def with_data(
         names: Iterable[str],
         configs: Iterable[CellConfig],
     ) -> InternalApp:
-        new_cell_manager = CellManager()
```
mscolnick, line 1017: i guess this was a bug missing the `prefix`?

marimo/_ast/cell_manager.py
```diff
@@ -208,13 +217,22 @@ def register_cell(
...
-            cell=cell,
-        )
+        resolved_config = config or CellConfig()
+        existing = self._document.get(cell_id)
+        if existing is not None:
+            existing.code = code
+            existing.name = name
+            existing.config = resolved_config
+        else:
+            self._document._cells.append(
```
mscolnick, line 229: i wonder if we should go through the `.apply(Transaction()` API. it does increment the version which is not ideal

marimo/_ast/app.py
kirangadhave: `with_data` is only called from file save. The prefix on top level notebooks is `""`, so strictly we don't need to do this. But this is more consistent.

marimo/_ast/cell_manager.py
kirangadhave: I have some follow up issues filed, one of that is to do all mutations using Transactions, we can revisit it then. Maybe we can add a silent-apply or something which doesn't bump version?

kirangadhave:
@cubic-dev-ai

mscolnick (dismissed):
looks great! letting copilot do a pass as well

marimo/_session/session.py
```diff
@@ -278,6 +251,19 @@ def __init__(
...
+
+        Derived from ``self.app_file_manager.app.cell_manager.document``
+        rather than stored, so any code path that swaps the underlying
+        ``CellManager`` or ``app`` (save round-trip, file-watch reload,
+        export reload) is automatically picked up — no rebinding needed
+        at the call sites. Read-only by design: the document's identity
+        belongs to the cell manager.
+        """
+        return self.app_file_manager.app.cell_manager.document
+
```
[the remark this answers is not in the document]
kirangadhave: added in stacked PR

marimo/_ast/cell_manager.py
```diff
@@ -208,13 +217,22 @@ def register_cell(
         if cell_id is None:
             cell_id = self.create_cell_id()
```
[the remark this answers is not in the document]
kirangadhave: added

marimo/_session/file_change_handler.py
```diff
@@ -87,58 +153,22 @@ def handle_reload(
...
+        }
+
+        if transaction.changes:
+            # AppFileManager.reload already brought session.document to the
+            # post-reload state, so we don't apply(); we only broadcast.
+            # Bump the document version manually so the stamped transaction
+            # is consistent with what apply() would produce.
+            new_doc = session.document
+            new_doc._version += 1
+            stamped = structs_replace(transaction, version=new_doc._version)
```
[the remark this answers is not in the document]
kirangadhave: fixed in the follow up PR

mscolnick (approved):

PR #9410, opened 2026-04-28
fix(frontend): make disconnect indicator clickable to reconnect

### LINKED ISSUE #9385: Disconnect indicator unclickable; reconnect requires page reload after WS retries exhausted
### Describe the bug

When using Marimo over an SSH-tunneled kernel, when the kernel gets disconnected (especially for an extended period of time like overnight if SSH is disconnected), the only way to reconnect is to refresh the page. The broken link indicator gives the impression of being clickable but is a no-op. Refreshing resets your page position which is very disruptive and slow.

**Root cause (two layers)**

  1. The frontend hook's reconnect gate (`shouldTryReconnecting`) only resets on a successful `onOpen`, so once `partysocket`'s 10-retry budget is exhausted, our code never calls `ws.reconnect()` again.
  2. `partysocket@1.1.10` itself has an upstream bug: `_connectLock` leaks in the `_connect()` max-retries early-return, so even if the gate is reset, `ws.reconnect()` silently no-ops. Fixed upstream in [partykit#322 → partysocket
  1.1.13](https://github.com/partykit/partykit/pull/322).

  The combo means the broken-link icon's "should be clickable" UX is unrecoverable without a page reload.

I have a PR ready for this that I can submit soon which makes this UX much more intuitive and robust. It revamps the kernel reconnection behavior and makes the icon truly clickable to attempt a graceful reconnection.

### Will you submit a PR?

- [x] Yes

### Environment

<details>

```
{
  "marimo": "0.23.2",
  "editable": false,
  "location": "[REDACTED]/.venv/lib/python3.11/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.8.0-1047-aws",
  "Processor": "x86_64",
  "Python Version": "3.11.14",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v20.19.5",
    "uv": "0.10.0"
  },
  "Dependencies": {
    "click": "8.3.2",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.19.0",
    "packaging": "26.1",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.44.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "pandas": "3.0.2"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

  1. `marimo edit notebook.py` accessed via SSH port-forward (`ssh -L 2718:localhost:2718 host`). Or simulate locally: open notebook, DevTools → Network → Offline, wait ~3 minutes, back online.
  2. Broken-link icon appears. Hover → tooltip "App disconnected". Click → no effect.
  3. Even after the network recovers, the WS doesn't reconnect; only `Cmd-R` works.

bestvibes:
## 📝 Summary

Closes #9385 

**This pull request was primarily authored by a coding agent.**

### Problem

When marimo is accessed over an SSH-forwarded tunnel (or any flaky link), the kernel WebSocket drops on tunnel reset or laptop sleep. The browser shows the broken-link indicator in the top-left, and the only recovery is a full page reload — disruptive because it resets scroll position, cell focus, and any unsaved client state, despite the kernel surviving the WS drop in edit mode and the server already supporting session resume on a fresh connection (`marimo/_server/api/endpoints/ws/ws_session_connector.py`).

The frontend was forcing the reload because of two compounding issues:

1. **One-shot reconnect gate**: `useMarimoKernelConnection.shouldTryReconnecting` only resets on a successful `onOpen`. After partysocket's 10-retry budget is exhausted, the gate stays closed and `ws.reconnect()` is never called again from our code.
2. **Upstream partysocket bug**: in `partysocket@1.1.10`, `_connectLock` leaks in the `_connect()` max-retries early-return — even if the gate were reset, `ws.reconnect()` would silently no-op because the lock is permanently held. Fixed upstream in [partykit#322](https://github.com/cloudflare/partykit/pull/322), released as partysocket 1.1.13.
3. **Stuck spinner**: the `onClose` default branch always sets state to `CONNECTING`, even after partysocket has exhausted its retry budget — the spinner outlives the actual retry loop with no escape.

The broken-link icon also gives the visual impression of being clickable but is a no-op `<div>` today.

### What changed

This is two commits — a dependency bump (precondition) followed by the frontend fix.

#### `chore(frontend): bump partysocket 1.1.10 → 1.1.13`

Patch-version bump pulling in the upstream `_connectLock` fix. No API-surface changes — verified that `retryCount`, `reconnect()`, and `maxRetries` signatures are identical between 1.1.10 and 1.1.13. The lockfile diff is exactly 6 lines, all about partysocket.

#### `fix(frontend): make disconnect indicator clickable to reconnect`

- **Disconnect indicator becomes a real `<button>`** (`status.tsx`). Native button gives Enter/Space activation and accessibility; previously it was a `<div>` with a tooltip. When the connection is in a recoverable CLOSED state, clicking calls a new `reconnect()` exposed by `useMarimoKernelConnection`. When the connection is `CLOSED` for any other reason, the button is `disabled`.
- **`reconnect()` probes the runtime first** via the existing `runtimeManager.isHealthy()` (`/health`). If the server is genuinely unreachable (e.g. the SSH tunnel is dead), we transition straight back to `CLOSED + KERNEL_DISCONNECTED` in ~1s instead of burning partysocket's full retry budget on the silent click. The probe also resets the one-shot reconnect gate so subsequent automatic retries work.
- **Surface `CLOSED` when partysocket has given up.** In the `onClose` default branch, when `ws.retryCount >= MAX_RETRIES`, transition to `CLOSED + KERNEL_DISCONNECTED` instead of `CONNECTING`. This makes the disconnect icon reappear and become clickable again, so the user can manually retry without reloading. Without this, the spinner stayed up forever even after partysocket had stopped trying.
- **Refactor**: the inline switch on `event.reason` in `onClose` is extracted into a pure `classifyCloseEvent` helper (`close-handler.ts`) so the regression behavior can be unit-tested. The hook is now thin glue: classify → setConnection → optionally close transport or call `tryReconnecting`.
- **Tests**: `close-handler.test.ts` (15 cases) covers each terminal reason plus the retry-budget-exhaustion path. `status.test.tsx` (3 cases) covers the click → callback wiring and the disabled state.

### Heads-up for review

- **`runtimeManager.isHealthy()` has a side effect** (`setDOMBaseUri()` on success). It was previously only triggered during initial startup; now it can fire on each manual reconnect. Almost certainly a no-op in practice (URL hasn't changed) but flagging in case you want it gated.
- **`IConnectionTransport.retryCount` was added** to the transport interface so the close handler can detect partysocket give-up. This leaks a partysocket-specific concept into a transport-agnostic interface; an alternative would be to have the transport emit a synthetic close event with a "gave-up" reason. Open to that direction if preferred — current approach was chosen for minimal diff.
- **No automatic-behavior changes** for users on a healthy connection. The only paths that exercise the new code are user-initiated clicks and the previously-stuck post-give-up case.
- **No new user-facing strings.** `"kernel not found"` / tooltip text reuse existing copy; only new string is the tooltip suffix `"— click to reconnect"`.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist


bestvibes:

frontend/src/core/websocket/close-handler.ts
```diff
@@ -0,0 +1,100 @@
...
+ * - `retry`: transient close; the consumer should reconnect.
+ */
+export type CloseDecision =
+  | { kind: "terminal"; status: ConnectionStatus; closeTransport: boolean }
+  | { kind: "gave-up"; status: ConnectionStatus }
+  | { kind: "retry"; status: ConnectionStatus };
+
+export interface CloseEventLike {
+  code?: number;
+  reason?: string;
```
kirangadhave: It might be worth creating a literal type

```ts
type CloseReason = 
	"MARIMO_ALREADY_CONNECTED"
	| "MARIMO_WRONG_KERNEL_ID"
	// and so on ...
```

Just gives us some additional type safety. I can do this in a follow up PR.
No action needed here.

frontend/src/core/websocket/close-handler.ts
```diff
@@ -0,0 +1,100 @@
...
+      if (context.retryCount >= context.maxRetries) {
+        return {
+          kind: "gave-up",
+          status: {
+            state: WebSocketState.CLOSED,
+            code: WebSocketClosedReason.KERNEL_DISCONNECTED,
+            reason: "kernel not found",
+          },
+        };
+      }
```
kirangadhave: We should use `logNever(e.reason)` in the default branch and move the retry/gave-up logic after the switch. That way if a new MARIMO_* reason gets added server-side, we'll at least get a console warning instead of silently falling into retry logic

frontend/src/core/websocket/useMarimoKernelConnection.tsx
```diff
@@ -340,6 +344,26 @@ export function useMarimoKernelConnection(opts: {
...
 
+  // Manual reconnect. Probes /health first to fail fast when the runtime
+  // is unreachable, instead of waiting on partysocket's retry budget.
+  const reconnect = async () => {
+    if (ws.readyState === WebSocket.OPEN) {
+      return;
+    }
+    shouldTryReconnecting.current = true;
+    setConnection({ state: WebSocketState.CONNECTING });
+    const healthy = await runtimeManager.isHealthy();
```
kirangadhave, line 443: `isHealthy` should not have side effects. Thanks for pointing it out. I will fix it in a follow up.

frontend/src/core/websocket/close-handler.ts
kirangadhave, line 1: we don't need this to be in a separate file. let's add this as an function in `useMarimoKernelConnection.tsx`. It's not going to be used anywhere else for now.

frontend/src/core/websocket/useMarimoKernelConnection.tsx
```diff
@@ -340,6 +344,26 @@ export function useMarimoKernelConnection(opts: {
     }
   };
 
+  // Manual reconnect. Probes /health first to fail fast when the runtime
+  // is unreachable, instead of waiting on partysocket's retry budget.
+  const reconnect = async () => {
+    if (ws.readyState === WebSocket.OPEN) {
```
kirangadhave: should probably also early return on `CONNECTING`. If user double clicks on the indicator, it will fire off a duplicate request.

frontend/src/core/websocket/useMarimoKernelConnection.tsx
```diff
@@ -340,6 +344,26 @@ export function useMarimoKernelConnection(opts: {
...
+      return;
+    }
+    shouldTryReconnecting.current = true;
```
kirangadhave, line 441: we should change `shouldTryReconnecting.current` to `false` if the health probe fails

kirangadhave (changes requested):
@bestvibes great job. Just a few comments to address, and then should be good to merge. Thanks a lot!

bestvibes:
Thanks for the review @kirangadhave, let me know your thoughts on the latest changes!

kirangadhave:
@cubic-dev-ai

frontend/src/components/editor/header/status.tsx
```diff
@@ -37,11 +40,24 @@ export const StatusOverlay: React.FC<{
...
+      onReconnect ? "App disconnected — click to reconnect" : "App disconnected"
+    }
+  >
+    <button
+      type="button"
+      className={cn(topLeftStatus, "bg-transparent border-0 p-0")}
+      aria-label={onReconnect ? "Reconnect to app" : "App disconnected"}
+      data-testid="disconnected-indicator"
+      onClick={onReconnect}
+      disabled={!onReconnect}
```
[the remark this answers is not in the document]
kirangadhave: oh this is a good callout. disabling a button will also consume all pointer events, so tooltip won't be visible. You can do something like this:
https://github.com/marimo-team/marimo/blob/e355c4ca5b539d0fc719f827e7e0cd02d44ff578/frontend/src/components/data-table/export-actions.tsx#L257

you wrap the button in a `span` and set/unset it's tab index

kirangadhave (changes requested):
almost there. Can you please address the copilot and cubic.dev PR comments? I think we should be good to merge after that. I missed adding those reviews in last iteration.

Also it would be great if you add a video of the button in action.

bestvibes:
Comments addressed! Here's a short video, it's hard to repro the scenario succinctly since it needs to wait for the retry budget to get exhausted but I think I caught it here.

https://github.com/user-attachments/assets/9530e982-95a3-43f4-9641-508840f2e315

kirangadhave (approved):
🚀 thanks a lot for addressing the changes. Great job on your first PR to marimo!

PR #9413, opened 2026-04-28
improvement: patch polars I/O in wasm

mscolnick:
Polars I/O fails in pyodide because the Rust core can't reach the network and its Python fallbacks reference modules (fsspec/aiohttp) that aren't installed, surfacing as NameError and friends. URL-based `read_csv / read_parquet / scan_*` etc. are unusable today.

Add a small WASM patch framework (`WasmPatchSet`) that wraps a target callable: try original, on caught exception run a fallback; if the fallback also raises, re-raise the original with the fallback chained so users see the real error. Returns a single unpatch handle and is a no-op outside pyodide.

Use it to wrap polars read_csv/scan_csv, read_parquet/scan_parquet, read_ipc/scan_ipc, read_ndjson/scan_ndjson, and read_json. The fallback resolves the source to bytes (URLs via sync XMLHttpRequest, file-likes/paths/bytes locally), decodes via pyarrow, and returns polars via pl.from_arrow. scan_* materialize eagerly then .lazy() — true streaming wouldn't work in WASM anyway.


marimo/_runtime/_wasm/_polars.py
```diff
@@ -0,0 +1,267 @@
+# Copyright 2026 Marimo. All rights reserved.
+"""WASM-only fallback patches for polars.
```
dmadisetti, line 2: I wonder if we can make a `_runtime/_wasm/*`

marimo/_runtime/patches.py
```diff
@@ -458,72 +454,3 @@ def enhanced_init(
         SignatureParamName.infer = wrap_infer(original_dynamic_infer)
     if not getattr(SignatureParamName.__init__, "patched", False):
         SignatureParamName.__init__ = enhanced_init
-
-
-def patch_polars_write_json() -> Unpatch:
```
dmadisetti, line 463: nice refactor

marimo/_runtime/_wasm/_polars.py
mscolnick: yea i can move files

PR #9415, opened 2026-04-28
fix: update snapshots and types for altair v6.1.0 / vega-lite v6.4.1

mscolnick:
CI was failing on main after the altair upgrade landed:
- snapshots referenced vega-lite v6.1.0 schema and old view-hash names
- alt.layer now returns LayerChart | FacetChart, breaking add_common_config
- vlc.get_*_locale signatures became Literal-typed

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -65,8 +65,6 @@
         "!/.*@oxc.*/",
         // react-hook-form is pinned to 7.54.2
         "!/.*react-hook-form.*/",
-        // vega-lite is pinned to 6.3.1
-        "!/.*vega-lite.*/",
         // pyodide needs to be manually updated
         "!/.*pyodide.*/"
       ],
--- a/.github/workflows/test_fe.yaml
+++ b/.github/workflows/test_fe.yaml
@@ -181,10 +181,6 @@ jobs:
             echo "Error: react-hook-form version in package.json must be exactly 7.54.2. As it breaks mo.ui.dataframe"
             exit 1
           fi
-          if ! grep -q '"vega-lite": "6.3.1"' package.json; then
-            echo "Error: vega-lite version in package.json must be exactly 6.3.1. Newer versions break numeric tooltips"
-            exit 1
-          fi
           if ! grep -q '"html-to-image": "1.11.13"' package.json; then
             echo "Error: html-to-image version in package.json must be exactly 1.11.13. Newer versions break matplotlib toolbars"
             exit 1
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -155,7 +155,7 @@
     "typescript-memoize": "^1.1.1",
     "use-acp": "0.2.6",
     "use-resize-observer": "^9.1.0",
-    "vega-lite": "6.4.1",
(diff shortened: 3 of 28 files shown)


mscolnick:
Verified the numeric tooltip bug is no longer there.

<details><summary>Details</summary>
<p>

```
import marimo

__generated_with = "0.23.3"
app = marimo.App()


@app.cell
def _():
    import marimo as mo

    mo.md(
        """
        # Verify vega-lite tooltip fix

        This notebook reproduces the numeric-tooltip bug fixed by
        [vega-lite#9707](https://github.com/vega/vega-lite/pull/9707) (released in
        vega-lite **6.4.2**).

        The bug: when hovering near a point on a chart that uses
        `selection_point(nearest=True)`, the tooltip would show `undefined`
        instead of the actual numeric value.

        **Hover over the points below.** The tooltip should display real
        numbers — not `undefined`.
        """
    )
    return (mo,)


@app.cell
def _():
    import altair as alt
    import pandas as pd

    return alt, pd


@app.cell
def _(alt, pd):
    df = pd.DataFrame(
        {
            "x": list(range(20)),
            "y": [v * v for v in range(20)],
        }
    )

    nearest = alt.selection_point(
        fields=["x"], nearest=True, on="mouseover", empty=False
    )

    line = alt.Chart(df).mark_line().encode(x="x:Q", y="y:Q")
    points = (
        alt.Chart(df)
        .mark_point(size=80, filled=True)
        .encode(
            x="x:Q",
            y="y:Q",
            opacity=alt.condition(nearest, alt.value(1), alt.value(0)),
            tooltip=["x:Q", "y:Q"],
        )
        .add_params(nearest)
    )

    chart = (line + points).properties(
        title="Hover near points — tooltip should show numbers, not `undefined`"
    )
    chart
    return


@app.cell
def _(mo):
    mo.md("""
    ## Datasource panel chart

    The same pattern is used by the datasource preview chart for date
    columns (see `DateChartBuilder` in `marimo/_data/charts.py`). If the
    upgrade is good, hovering anywhere along the area below should show
    the count and date — never `undefined`.
    """)
    return


@app.cell
def _(alt, pd):
    import datetime as _dt

    dates = pd.DataFrame(
        {
            "date": [
                _dt.date(2024, 1, 1) + _dt.timedelta(days=i)
                for i in range(60)
            ],
            "count": [i % 7 + 1 for i in range(60)],
        }
    )

    nearest_d = alt.selection_point(
        fields=["date"], nearest=True, on="mouseover", empty=False
    )

    base = alt.Chart(dates)
    area = base.mark_area(opacity=0.4).encode(x="date:T", y="count:Q")
    pts = (
        base.mark_point(size=80, filled=True)
        .encode(
            x="date:T",
            y="count:Q",
            opacity=alt.condition(nearest_d, alt.value(1), alt.value(0)),
            tooltip=[
                alt.Tooltip("date:T"),
                alt.Tooltip("count:Q", format=",.0f"),
            ],
        )
        .add_params(nearest_d)
    )
    (area + pts).properties(width="container", height=200)
    return


if __name__ == "__main__":
    app.run()

```

</p>
</details>

PR #9426, opened 2026-04-30
fix: force Content-Disposition on table export downloads

mscolnick:
Pass `?download=1` from the data-table export menu so the virtual-file
endpoint returns `Content-Disposition: attachment`. Without this, iframed
deployments where `<a download>` is ignored (sandboxed iframes, restrictive
Permissions-Policy) navigated to the JSON instead of saving it.


mscolnick:
noticed this issue in `molab`

kirangadhave (approved):
🚀

PR #9435, opened 2026-05-01
improvement: fall back to disk storage when shared memory cannot allocate

mscolnick:
`SharedMemoryStorage.store()` had no error handling around `shared_memory.SharedMemory(create=True, size=...)`, so an OSError (e.g. /dev/shm full) would crash the kernel and drop the virtual file. Add a `FallbackStorage` composite that tries each backend in order and falls back on `OSError`, plus a cross-process-safe `DiskStorage` backend; wire EDIT mode and the manager's ad-hoc cross-process reader to use both tiers.


marimo/_runtime/virtual_file/storage.py
```diff
@@ -248,6 +274,239 @@ def has(self, key: str) -> bool:
...
+    Cross-process safe: any process can read a key by computing the same path.
+    Used as a fallback tier when ``SharedMemoryStorage`` cannot allocate
+    (e.g., ``/dev/shm`` is full).
+    """
+
+    def __init__(self, base_dir: str | Path | None = None) -> None:
+        self._base_dir = (
+            Path(base_dir)
+            if base_dir is not None
+            else Path(tempfile.gettempdir()) / "marimo-vfs"
```
dmadisetti, line 289: I think it might be verifying that this gets mapped to tmpfs on molab, otherwise this will be very slow when actually used. Currently maps to `/tmp` (expected) but our mounts should this to be on disk

dmadisetti (dismissed):
Left a comment but not addressable here pinged @mchav

PR #9440, opened 2026-05-03
fix: resolve @file virtual-file URLs to absolute in HTML renderer

### LINKED ISSUE #9432: mo.image() not rendered on molab.marimo.io in edit mode
### Describe the bug

`mo.image()` not rendered on `molab.marimo.io` in edit mode, however it does work on the app mode.

See examples:

This does not work:
 https://molab.marimo.io/notebooks/nb_FgWdGrSJSmpU63tcVtwB9r

This does work:
https://molab.marimo.io/notebooks/nb_FgWdGrSJSmpU63tcVtwB9r/app

Note: I think this is similar, but not exactly the same as this issue:  https://github.com/marimo-team/marimo/issues/5731

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
n/a see https://molab.marimo.io/notebooks/nb_FgWdGrSJSmpU63tcVtwB9r
```

</details>

### Code to reproduce

_No response_

VishakBaddur:
## Problem

`mo.image()` (and `mo.audio()`, `mo.pdf()`) fails to render in edit mode on molab but works in app mode. Closes #9432.

**Root cause:** The backend intentionally generates relative virtual file URLs like `./@file/SIZE-filename` (it can't know the mount path at creation time). When the page URL has no trailing slash — e.g., molab edit mode at `/notebooks/nb_xxx` — the browser resolves `./@file/...` to `/notebooks/@file/...`, dropping the notebook ID and causing a 404.

App mode works because the URL ends with `/app`, providing an extra path segment that makes relative resolution correct.

Verified empirically:

```js
new URL('./@file/123.png', 'https://molab.marimo.io/notebooks/nb_xxx')
// → https://molab.marimo.io/notebooks/@file/123.png  ❌ (notebook ID dropped)

new URL('./@file/123.png', 'https://molab.marimo.io/notebooks/nb_xxx/app')
// → https://molab.marimo.io/notebooks/nb_xxx/@file/123.png  ✓
```

## Fix

Add `resolveVirtualFileUrl()` helper in `RenderHTML.tsx` that resolves `@file/` URLs against the runtime base URL with a guaranteed trailing slash, making the URL absolute and unambiguous regardless of the page URL's trailing slash.

Apply it in:
- `replaceVirtualFileSrc` — handles ``, `<audio>`, `<video>`, `<source>` (covers `mo.image()`, `mo.audio()`)
- `replaceValidIframes` — handles `<iframe>` (covers `mo.pdf()`)

Data URLs and external `https://` URLs are untouched.

## Testing

- 4 new regression tests: `./@file/` rewriting, `@file/` without `./`, non-`@file` URLs (unchanged), `data:` URLs (unchanged)
- 305/305 test files pass, 4722/4722 tests pass, zero regressions

### DIFF
--- a/frontend/src/components/data-table/renderers.tsx
+++ b/frontend/src/components/data-table/renderers.tsx
@@ -164,13 +164,17 @@ export const DataTableBody = <TData,>({
       );
 
       const title = cell.getHoverTitle?.() ?? undefined;
+      const isCellSelected = cell.getIsSelected?.() || false;
       return (
         <TableCell
           tabIndex={0}
           {...getCellDomProps(cell.id)}
           key={cell.id}
           className={cn(
-            "whitespace-pre truncate max-w-[300px] outline-hidden border-r border-r-border/75",
+            "whitespace-pre truncate max-w-[300px] border-r border-r-border/75",
+            isCellSelected
+              ? "outline outline-2 outline-(--blue-7) -outline-offset-2"
+              : "outline-hidden",
             cell.column.getColumnWrapping &&
               cell.column.getColumnWrapping?.() === "wrap" &&
               COLUMN_WRAPPING_STYLES,
--- a/frontend/src/components/data-table/renderers.tsx
+++ b/frontend/src/components/data-table/renderers.tsx
@@ -164,17 +164,13 @@ export const DataTableBody = <TData,>({
       );
 
       const title = cell.getHoverTitle?.() ?? undefined;
-      const isCellSelected = cell.getIsSelected?.() || false;
       return (
         <TableCell
           tabIndex={0}
           {...getCellDomProps(cell.id)}
           key={cell.id}
           className={cn(
-            "whitespace-pre truncate max-w-[300px] border-r border-r-border/75",
-            isCellSelected
-              ? "outline outline-2 outline-(--blue-7) -outline-offset-2"
-              : "outline-hidden",
+            "whitespace-pre truncate max-w-[300px] outline-hidden border-r border-r-border/75",
             cell.column.getColumnWrapping &&
               cell.column.getColumnWrapping?.() === "wrap" &&
               COLUMN_WRAPPING_STYLES,
--- a/frontend/src/plugins/core/RenderHTML.tsx
+++ b/frontend/src/plugins/core/RenderHTML.tsx
@@ -38,11 +38,18 @@ interface Options {
 // using the runtime base, ensuring a trailing slash so the notebook-ID path
 // segment is never dropped during relative resolution.
 function resolveVirtualFileUrl(src: string): string {
-  const base = getRuntimeManager().httpURL;
-  if (!base.pathname.endsWith("/")) {
-    base.pathname += "/";
+  // We intentionally do not reuse asRemoteURL() here: that function does not
+  // guarantee a trailing slash, so relative resolution of "./@file/..." would
+  // drop the last path segment on deployments like molab where the page URL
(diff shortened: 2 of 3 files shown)


dmadisetti:
Thanks! LGTM, but I think we need to make sure this is general to base URL. I'll look at it in a bit more detail but pinged @mchav who should be able to determine a bit more molab context

frontend/src/components/data-table/renderers.tsx
```diff
@@ -164,13 +164,17 @@ export const DataTableBody = <TData,>({
       );
 
       const title = cell.getHoverTitle?.() ?? undefined;
+      const isCellSelected = cell.getIsSelected?.() || false;
```
mchav: How is this related to the current change?

frontend/src/plugins/core/RenderHTML.tsx
```diff
@@ -32,6 +33,39 @@ interface Options {
   alwaysSanitizeHtml?: boolean;
   additionalReplacements?: ReplacementFn[];
 }
+// Resolve a virtual file URL (./@file/... or @file/...) to an absolute URL
+// using the runtime base, ensuring a trailing slash so the notebook-ID path
+// segment is never dropped during relative resolution.
+function resolveVirtualFileUrl(src: string): string {
```
mchav, line 40: This seems similar to https://github.com/marimo-team/marimo/blob/a8dd4b6716d116ab0ba6b8eeaf32c7d50b53eace/frontend/src/core/runtime/config.ts#L54

Can we reuse that instead of creating a new function?

frontend/src/plugins/core/__test__/renderHTML-sanitization.test.tsx
```diff
@@ -129,3 +129,55 @@ describe("renderHTML sanitization integration", () => {
     );
   });
 });
+
+describe("replaceVirtualFileSrc - virtual file URL rewriting", () => {
```
mchav, line 141: These tests don't actually repro the bug. We shoudl mock the runtime manager and check against complete strings rather than substrings.

mchav (changes requested):

frontend/src/components/data-table/renderers.tsx
VishakBaddur: You're right, that's unrelated, removed it.

frontend/src/plugins/core/RenderHTML.tsx
VishakBaddur: I've restructured it to mirror asRemoteURL's pattern (blob: handling + same base resolution). The reason I kept it as a separate function rather than calling asRemoteURL directly is that asRemoteURL doesn't guarantee a trailing slash, and without that, new URL("./@file/...", base) drops the last path segment when the base URL has no trailing slash (the exact bug we're fixing). Added a comment in the code explaining this.

frontend/src/plugins/core/__test__/renderHTML-sanitization.test.tsx
VishakBaddur: Updated the tests to mock getRuntimeManager with a base URL that has no trailing slash (http://localhost:2718/notebooks/nb_xxx), which reproduces the actual molab edit-mode scenario. Assertions now check complete URL strings instead of substrings.

mscolnick:
Going to close this so it does not get accidentally review more. the issue is on molab and it is already fixed in the latest molab that we are currently doing a slow rollout of.

PR #9453, opened 2026-05-04
Autosave only during code-mode

kirangadhave:
## Summary

- Tighten `Transaction.source` to `Literal["frontend", "kernel", "code-mode", "file-watch"]` so the closed set is type-checked.
- Rename code-mode's transaction emission from `source="kernel"` to `source="code-mode"`, and gate `_maybe_autosave` on it.

## Why
`run_app_then_export_as_ipynb` (and any other path using `run_app_until_completion` in EDIT mode) was rewriting source `.py` files as a side effect.

The fix flips the autosave default from "all kernel-sourced transactions persist" to "only code-mode transactions persist."

### DIFF
--- a/marimo/_code_mode/_context.py
+++ b/marimo/_code_mode/_context.py
@@ -1587,7 +1587,7 @@ class AsyncCodeModeContext:
             internal_ops=ops,
         )
         if doc_ops:
-            tx = Transaction(changes=tuple(doc_ops), source="kernel")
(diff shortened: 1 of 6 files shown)


kirangadhave:
@cubic-dev-ai

mscolnick (approved):

mscolnick (dismissed):

kirangadhave:
removing bash-focus label which was auto-added

mscolnick (approved):

PR #9458, opened 2026-05-05
chore: regenerate openapi types for cell_manager TransactionSource

kirangadhave:
Codegen follow-up to #9457 — regenerates `packages/openapi/api.yaml` and `src/api.ts` to include the `cell_manager` literal in `TransactionSource`.

### DIFF
--- a/packages/openapi/api.yaml
+++ b/packages/openapi/api.yaml
@@ -5054,6 +5054,7 @@ components:
           type: array
         source:
           enum:
+          - cell_manager
           - code-mode
           - file-watch
           - frontend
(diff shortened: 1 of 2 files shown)


kirangadhave:
doesn't need bug-bash. internal change.

packages/openapi/api.yaml
```diff
@@ -5054,6 +5054,7 @@ components:
           type: array
         source:
           enum:
+          - cell_manager
```
dmadisetti, line 5057: Maybe in a follow up `cell-manager`? Weird it's serializing different

dmadisetti (approved):

PR #9459, opened 2026-05-05
chore: rename cell_manager TransactionSource to cell-manager

kirangadhave:
Use kebab-case for the `cell-manager` transaction source, matching the other sources (`frontend`, `kernel`, `code-mode`, `file-watch`).

Follow-up to #9457 and #9458.

## Test plan
- [x] `make check` passes


dmadisetti (approved):
l-g-t-m

PR #9462, opened 2026-05-06
refactor: `SetConfig` is now full-replacement instead of partial-update

kirangadhave:
## Summary

- `SetConfig` was a partial update where `None` meant "leave unchanged", which collided with the legitimate `column=None` value (cell outside any column layout). Any diff producer that emitted `SetConfig(column=None, ...)` to reset a cell's column was silently dropped on apply.
- Made `SetConfig` a full replacement: `column`/`disabled`/`hide_code` are required and overwrite the cell's config directly. Producers that already emitted full triples (file-watch reload, code-mode plan) need no changes; the `app.setup` setup-cell producer now reads existing config first. Frontend producers (`columnChanges`, `updateCellConfig`) and the apply consumer mirror the full-replacement semantic.


kirangadhave:
@cubic-dev-ai

Light2Dark (approved):
makes sense!

PR #9465, opened 2026-05-06
feat: add new W&B models to llm-info data

mscolnick:
This update introduces several new models to the llm-info data, including DeepSeek V3.1, Google Gemma 4 31B, IBM Granite 4.1 8B, and others.


Light2Dark (approved):
Do we need to update molab to use these?

PR #9472, opened 2026-05-07
fix: remount <img> on src change in mo.Html to avoid stale paint

mscolnick:
When mo.Html re-renders with new image URLs, React reconciliation reuses
the existing  DOM nodes and only updates the src attribute. With
flaky CDNs/CORS proxies the new request can stall or fail and the
browser keeps the previous image painted, so the user sees stale (or
blank) images even though the rendered HTML source is correct.

Key  elements by src+index in RenderHTML so React unmounts and
remounts them when the src changes, mirroring the existing iframe
remount workaround.

### DIFF
--- a/frontend/src/plugins/core/RenderHTML.tsx
+++ b/frontend/src/plugins/core/RenderHTML.tsx
@@ -6,6 +6,7 @@ import parse, {
   type HTMLReactParserOptions,
 } from "html-react-parser";
 import React, {
+  cloneElement,
   isValidElement,
   type JSX,
   type ReactNode,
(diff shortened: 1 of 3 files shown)


frontend/src/plugins/core/RenderHTML.tsx
```diff
@@ -169,6 +170,35 @@ const addCopyButtonToCodehilite: TransformFn = (
...
+    return undefined;
+  }
+  const src = domNode.attribs?.src;
+  if (!src) {
+    return undefined;
+  }
+  // data: URIs can be very large; hash them so keys stay bounded.
+  const keyPart = src.startsWith("data:") ? `data:${djb2(src)}` : src;
+  return `${keyPart}-${index}`;
+};
```
kirangadhave: For `data:` URIs we shouldn't have a stale-repaint issue. They're decoded from local bytes and rendered synchronously, so there's no in-flight request to stall or fail.

we can skip `data:` URIs entirely:

```suggestion
// Build a React key for an <img> derived from its src so the element
// remounts when src changes, instead of reusing the existing DOM node.
// Reusing an <img> across src changes can leave the previous image painted
// (e.g. when the new request is slow/blocked, served stale by a CDN, or
// fails CORS), so the user sees the old image even though the HTML source
// is up to date.
const imageKeyFor = (domNode: DOMNode, index: number): string | undefined => {
  if (!(domNode instanceof Element) || domNode.name !== "img") {
    return undefined;
  }
  const src = domNode.attribs?.src;
  if (!src || src.startsWith("data:")) {
    return undefined;
  }
  return `${src}-${index}`;
};
```

That lets us drop `djb2` entirely and no need to hash huge strings

kirangadhave (commented):
Suggestion to skip keys for data uris

kirangadhave (approved):
🚀

PR #9473, opened 2026-05-07
fix: support HTTP Range requests on virtual files (#9460)

### LINKED ISSUE #9460: mo.audio doesn't work in Safari
### Describe the bug

mo.audio appears to work in Chrome and Firefox but in Safari it looks like this:

Image

Tested in Safari Version 26.2 (21623.1.14.11.9) on masOC 26.2 (25C56)

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.5",
  "editable": false,
  "location": "/Users/yairchu/dev/soundradix/Drum3D/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.2.0",
  "Processor": "arm",
  "Python Version": "3.13.1",
  "Locale": "C/C",
  "Binaries": {
    "Browser": "140.0.7339.134",
    "Node": "v23.11.0",
    "uv": "0.7.12 (dc3fd4647 2025-06-06)"
  },
  "Dependencies": {
    "click": "8.3.3",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.20.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.46.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.1.0",
    "loro": "1.10.3",
    "pandas": "3.0.2"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
_sr = 44100
mo.audio(
    0.01 * np.sin(math.tau * np.cumsum(np.linspace(660, 110, 100000)) / _sr),
    _sr,
    normalize=False,
)
```

mscolnick:
Safari's `<audio>/<video>` elements require a 206 Partial Content response
to their initial range probe. Without it, mo.audio renders a disabled
player. Chrome and Firefox are more forgiving, so the bug was Safari-only.

Parse Range headers in the `/@file/` endpoint and return 206 with
`Content-Range/Content-Length`, advertise `Accept-Ranges:` bytes on every
response, and 416 for invalid ranges. Full responses still omit
`Content-Length` to preserve the h11 fix from #8928. Adds an optional start
offset to the chunked virtual-file readers so partial reads don't
allocate the full buffer.

Closes #9460

### DIFF
--- a/marimo/_runtime/virtual_file/storage.py
+++ b/marimo/_runtime/virtual_file/storage.py
@@ -38,11 +38,18 @@ class VirtualFileStorage(Protocol):
         key: str,
         byte_length: int,
         chunk_size: int = DEFAULT_CHUNK_SIZE,
+        start: int = 0,
     ) -> Iterator[bytes]:
         """Read buffer data by key in chunks.
 
         Yields chunks of bytes, avoiding allocating the full buffer at once.
         Useful for streaming large files over HTTP.
+
+        Args:
+            key: storage key
+            byte_length: total number of bytes to yield (after applying ``start``)
+            chunk_size: chunk size in bytes
+            start: offset in bytes to begin reading from (default 0)
 
         Raises:
             KeyError: If key not found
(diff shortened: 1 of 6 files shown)


kirangadhave (approved):
🚀 tested in Safari

PR #9478, opened 2026-05-07
feat: add Mermaid theme customization options

williambdean:
This PR adds Mermaid theme customization to `mo.mermaid(...)` so users can style diagrams without manually rewriting Mermaid strings.

## Summary

- Add `theme` and `theme_variables` to `mo.mermaid(...)`
- Wire both through the Mermaid stateless plugin and frontend renderer
- Preserve existing default behavior when no theme is provided
- Re-render Mermaid output when theme/theme variables change
- Add backend/frontend tests and sanitizer coverage
- Update docs with Mermaid reference links and behavior notes

## Behavior

- `theme_variables` + `theme=None` -> marimo automatically uses `theme="base"`
- `theme_variables` + explicit non-`base` theme -> raises `ValueError`

## Examples

Theme only:

```python
mo.mermaid(
    diagram,
    theme="neutral",
)
```

Theme variables (auto-uses `theme="base"` when theme is omitted):

```python
mo.mermaid(
    diagram,
    theme_variables={
        "primaryColor": "#E8EEF5",
        "primaryTextColor": "#1F2937",
        "primaryBorderColor": "#64748B",
        "lineColor": "#475569",
    },
)
```

## Validation

- `uv run --group test pytest tests/_plugins/stateless/test_mermaid.py`
- `pnpm test src/plugins/layout/__test__/MermaidPlugin.test.ts src/plugins/core/__test__/sanitize.test.ts`

### DIFF
--- a/marimo/_plugins/stateless/mermaid.py
+++ b/marimo/_plugins/stateless/mermaid.py
@@ -1,36 +1,16 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-from typing import Literal
-
 from marimo._output.hypertext import Html
 from marimo._output.rich_help import mddoc
 from marimo._plugins.core.web_component import build_stateless_plugin
 from marimo._utils.dicts import remove_none_values
 
-MermaidTheme = Literal[
-    "base",
-    "dark",
-    "default",
-    "forest",
-    "neutral",
-    "null",
-]
-
-_MERMAID_THEMES: set[MermaidTheme] = {
-    "base",
-    "dark",
-    "default",
-    "forest",
-    "neutral",
-    "null",
-}
-
 
 @mddoc
 def mermaid(
     diagram: str,
-    theme: MermaidTheme | None = None,
+    theme: str | None = None,
     theme_variables: dict[str, str] | None = None,
 ) -> Html:
     """Render a diagram with Mermaid.
@@ -41,8 +21,9 @@ def mermaid(
 
     Args:
         diagram: a string containing a Mermaid diagram
-        theme: optional Mermaid theme (`base`, `dark`, `default`,
-            `forest`, `neutral`, or `null`). If not provided, marimo picks
+        theme: optional Mermaid theme. See
+            https://mermaid.js.org/config/theming.html#available-themes
+            for available themes. If not provided, marimo picks
             a default based on app theme, unless `theme_variables` are
             provided.
         theme_variables: optional Mermaid `themeVariables` overrides.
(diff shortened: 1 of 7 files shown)


marimo/_plugins/stateless/mermaid.py
```diff
@@ -30,11 +64,42 @@ def mermaid(diagram: str) -> Html:
...
+                "primaryTextColor": "#1F2937",
+                "primaryBorderColor": "#64748B",
+                "lineColor": "#475569",
+                "tertiaryColor": "#F8FAFC",
+            },
+        )
         ```
     """
+    if theme is not None and theme not in _MERMAID_THEMES:
+        raise ValueError(
```
dmadisetti: Can we validate on mermaid itself? This forces us to always keep mermaid defs up to date

marimo/_plugins/stateless/mermaid.py
```diff
@@ -30,11 +64,42 @@ def mermaid(diagram: str) -> Html:
...
+            f"theme must be one of {sorted(_MERMAID_THEMES)}, got {theme!r}"
+        )
+
+    if theme_variables is not None:
+        if theme is None:
+            theme = "base"
+        elif theme != "base":
+            raise ValueError(
```
dmadisetti, line 66: Why?

dmadisetti (commented):
Nice! I think the default green can be a little wearing. Left some comments, but I think the api kwargs make sense!

marimo/_plugins/stateless/mermaid.py
williambdean: the mermaid docs say colors cannot be overridden if it is not "base" theme. If non-base is passed as well as the color overrides, there is nothing stated to the user (Maybe in the console log, but I would have to check). 
This provides faster enforcement.

dmadisetti: Thanks for the clarification! Linking out to the mermaid docs on failure might be a better message

williambdean: Added the mermaid docs link in the raise explicitly.

marimo/_plugins/stateless/mermaid.py
williambdean: I've removed any checks here and hard coded themes. Added link to the docs in the python docstring. These is no raise that happens now. Mermaid will just silently do nothing.

williambdean:
Thanks for the feedback @dmadisetti . Everything you raised is addressed.

dmadisetti (approved):

PR #9480, opened 2026-05-08
feat: support HTTP DuckDB queries in WASM notebooks

peter-gy:
Motivated by marimo-team/quarto-marimo#74, marimo-team/jupyter-book-marimo#1, and #9413.

DuckDB remote file queries fail in Pyodide because DuckDB-WASM [can't use httpfs](https://duckdb.org/2024/10/02/pyodide#limitations). Therefore, URL-based SQL like `FROM
  'https://...'` and `read_csv/read_parquet/read_json('https://...')` are unusable in WASM notebooks today.

This PR adds a DuckDB WASM fallback layer for `mo.sql`, SQL cells, raw `duckdb.sql/query/execute/query_df`, connection SQL methods, and direct `duckdb.read_csv/read_parquet/read_json` calls.

It translates queries such as

```sql
SELECT * FROM read_csv('https://example.com/cars.csv')
SELECT * FROM 'https://example.com/cars.csv'
-- or duckdb.read_csv('https://example.com/cars.csv') via Python API
```

into

```sql
SELECT * FROM __marimo_wasm_duckdb_remote_0
```

where `__marimo_wasm_duckdb_remote_0` is bound to a fetched pandas DataFrame, which DuckDB can query through Python replacement scans.

Underneath, the fallback layer:

- uses sqlglot to analyze SQL and extract supported static remote file references from the AST
- fetches remote files through Python/urllib via marimo's shared WASM fetch util
- decodes fetched DuckDB file bytes into pandas DataFrames
- hands those DataFrames back to DuckDB under generated table names

Unsupported or dynamic cases are left to DuckDB's normal path. The patch is a no-op outside Pyodide, and in Pyodide the DuckDB SQL fallback requires `sqlglot` for AST analysis.

Tested with unit coverage for the rewrite/fetch/read paths and manually in the WASM playground against hosted CSV, parquet, JSON, and GeoJSON datasets.

## WASM Playground Demo

Demonstrates that the Pyodide-build of marimo supports querying remote files with DuckDB across cases like:

- CSV via `mo.sql`: direct URL scan
- Parquet via `duckdb.sql`: `read_parquet(...)`
- Direct `duckdb.read_csv` Python API with patched options (custom delimiter)
- JSON / GeoJSON path
- Connection API via `duckdb.connect`

https://github.com/user-attachments/assets/a2b310ed-2677-4f67-81da-3d8ea2145254

### DIFF
--- a/frontend/src/core/islands/worker/worker.tsx
+++ b/frontend/src/core/islands/worker/worker.tsx
@@ -87,7 +87,7 @@ const requestHandler = createRPCRequestHandler({
     await pyodideReadyPromise; // Make sure loading is done
 
     if (shouldLoadDuckDBPackages(code)) {
-      // DuckDB SQL and remote readers need these packages loaded up front.
+      // Add pandas and duckdb to the code for mo.sql and for remote duckdb sources
       code = `import pandas\n${code}`;
       code = `import duckdb\n${code}`;
       code = `import sqlglot\n${code}`;
--- a/frontend/src/core/islands/worker/worker.tsx
+++ b/frontend/src/core/islands/worker/worker.tsx
@@ -9,6 +9,7 @@ import {
 } from "rpc-anywhere";
 import type { NotificationPayload } from "@/core/kernel/messages";
 import type { ParentSchema } from "@/core/wasm/rpc";
+import { shouldLoadDuckDBPackages } from "@/core/wasm/utils";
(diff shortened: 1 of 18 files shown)


mscolnick (approved):
this is REALLY amazing! great cleanup and isolation. super excited to use this

PR #9483, opened 2026-05-08
refactor: replace MarimoFileKey alias with FileKey ADT

mscolnick:
Introduces a tagged union (NewFileKey | PathFileKey) for notebook keys,
replacing the `MarimoFileKey = str` alias and `__new__` magic-string
sentinel. Boundary code parses raw query-param strings via
`parse_file_key` once, internal code pattern-matches by type, and
`serialize_file_key` round-trips back to the wire format where needed
(session initialization_id, service-worker injection, log messages).

Also tightens the new-file sentinel from a `startswith` prefix check to
an exact match, so non-sentinel keys consistently flow through normal
path validation instead of returning the blank notebook.

### DIFF
--- a/marimo/_server/api/endpoints/assets.py
+++ b/marimo/_server/api/endpoints/assets.py
@@ -37,6 +37,11 @@ from marimo._server.templates.templates import (
     inject_script,
     notebook_page_template,
 )
+from marimo._server.workspace import (
+    FileKey,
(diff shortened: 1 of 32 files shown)


kirangadhave:
@cubic-dev-ai

dmadisetti (approved):

PR #9484, opened 2026-05-08
feat: enable chain shadowing for deterministic variable redefinition

### LINKED ISSUE #7815: Multiple definitions automatically act like merged cells
### Description

https://docs.marimo.io/guides/understanding_errors/multiple_definitions/?h=variable#merge-cells describes a way to deal with operations where you need to have mutable state -- making a bigger cell for them. You can still output incrementally through the cell.

The reason why this works is because it decouples two uses of cells in marimo:
1. Cells are the unit of execution, e.g. execution is rerun or marked stale at the cell level; dependencies are tracked at the cell level
2. Cells are the (default) unit of output, because the last expression in each cell is displayed

This seems like it could be done automatically, without requiring user hacks like `mo.output.append`.

### Suggested solution

I think there are multiple directions this could go, and would be very open to ideas.

But as a starting place, here's one proposal, that feels pretty minimal:
- Multiple definition errors are only raised if the cells are non-consecutive
- If consecutive cells define the same variable, they are all run as a unit and any change in any of them reruns them all/marks them all stale
- These units function like connected-components, so if cells 1, 2, and 3 all modify `a` and cells 3, 4, and 5 all modify `b`, cells 1-5 all become one unit
- The "units" are visible to the user as a bracket or something connecting the cells

- [x] Yes

### Alternatives

Here are some more ambitious alternatives:
- Multiple definitions are allowed in non-consecutive cells, so long as multiple such units aren't interleaved
- Cells with multiple definitions are still run independently, taking advantage of the persistent cache to "undo" the effects of a later mutating cell

### Additional context

_No response_

esauflores:

## Summary
Allow variable redefinition across cells when cells form a deterministic linear chain in the dependency graph. This enables AI-generated pipelines and sequential data transformations without requiring underscore prefix syntax.

## Changes
- Add implicit shadowing edges in edges.py for chain determinism
- Rewrite get_multiply_defined() in graph.py with topo-sort validation
- Simplify validate_graph.py to delegate to graph.get_multiply_defined()
- Add comprehensive tests for chain shadowing scenarios (9 new tests)
- Update all affected tests across 12 files

## Motivation
Closes #7815. Many users (especially those using AI/LLM to generate notebooks) encounter MultipleDefinitionError when writing sequential data pipelines like sequential data transformations.

With chain shadowing, these cells form a deterministic execution chain and are allowed.

## Testing
- All 571 affected tests pass
- Lint passes on changed files
- Pre-existing test failures are unrelated to this change

## Notes
- I am unable to sign the CLA at this time. Guidance appreciated.


esauflores:
Closing to re-submit properly. Apologies for the premature PR.

PR #9502, opened 2026-05-11
feat(wasm): render notebook snapshot while Pyodide initializes

mscolnick:
Previously the WASM loader blocked the entire app behind a spinner until
Pyodide finished downloading. With the new snapshot embedded by
`marimo export html-wasm --execute`, render cells (code + cached outputs)
on first paint and let Pyodide download in the background.

- Gate spinner on whether cells exist, not whether Pyodide is pending
- Toast Pyodide init failures instead of throwing when a snapshot is shown
- Add a footer "Pyodide" indicator that mirrors the "Kernel" one


dmadisetti (approved):
+1 Nice

PR #9505, opened 2026-05-11
fix(terminal): send initial size when WebSocket opens

mscolnick:
The mount-time fitAddon.fit() can fire before the terminal WS reaches
OPEN state (common in iframe + tunneled deployments), in which case the
resize message is silently dropped and the PTY stays at its default 0x0
winsize. bash then falls back to COLUMNS=80, producing horizontal-scroll
input lines and incorrect wrapping.

Send the current dimensions from the open handler so the PTY is always
sized before the user starts typing.


mscolnick:
Addressed the redundant `fitAddon.fit()`.

Skipping the test for now, very hard to mock this stuff out.

PR #9510, opened 2026-05-11
feat(table): allow hosts to size-gate downloads

kirangadhave:
This is step 1 toward fixing [marimo-team/marimo-lsp#542](https://github.com/marimo-team/marimo-lsp/issues/542); the companion change will live in marimo-lsp, which will set the atom to a concrete VS Code–appropriate limit at renderer init.

## Summary

Adds `size_bytes` to `DownloadAsResponse` and a new `downloadSizeLimitAtom` (jotai) that gates `mo.ui.table` exports and clipboard copies when the serialized payload exceeds a host-supplied limit. The default policy is `null` (gate dormant); hosts opt in by setting the atom at init time.

This is a bandaid until streaming exports lands — the goal is to give VS Code's NotebookRenderer iframe (and any other sandboxed host) a way to refuse downloads it can't fulfill, rather than failing silently.

- Backend: `download_as` now returns `(url, filename, size_bytes)`; the size is computed once from the already-serialized bytes per format (csv/json/parquet). Threaded through both `ui.table._download_as` and `ui.dataframe._download_as`.
- Frontend: `ExportMenu` reads `downloadSizeLimitAtom`; when set and \`size_bytes > limitBytes\`, a danger toast fires and the download/copy is skipped. Fail-open when \`size_bytes\` is absent (older backend) or the atom is unset.
- Schema: `DownloadAsSchema` zod output mirrors the new optional field.

This is blocking https://github.com/marimo-team/marimo-lsp/pull/554

## Test plan

- [x] \`make py-check\` clean
- [x] \`make fe-check\` clean
- [x] Backend tests for \`download_as\` size correctness (csv/json/parquet) and \`_download_as\` populating \`size_bytes\`
- [x] Manual: in browser, with atom unset, big-table export still works (no regression)
- [x] Manual: with the atom temporarily defaulted to \`{ limitBytes: 1, ... }\`, any non-empty table export and clipboard copy triggers the toast and skips the action

### DIFF
--- a/frontend/src/components/data-table/export-actions.tsx
+++ b/frontend/src/components/data-table/export-actions.tsx
@@ -113,11 +113,10 @@ export const ExportMenu: React.FC<ExportActionProps> = (props) => {
       data-testid="export-button"
       size="xs"
       variant="text"
-      aria-disabled={disabled}
+      disabled={disabled}
       className={cn(
         "print:hidden text-xs gap-1",
         open ? "text-primary" : "text-muted-foreground",
-        disabled && "opacity-50 cursor-not-allowed",
       )}
     >
       <DownloadIcon className="w-3.5 h-3.5" />
@@ -249,16 +248,16 @@ export const ExportMenu: React.FC<ExportActionProps> = (props) => {
   };
 
   return (
-    <DropdownMenu
-      modal={false}
-      open={open}
-      onOpenChange={(next) => setOpen(next && !disabled)}
-    >
+    <DropdownMenu modal={false} open={open} onOpenChange={setOpen}>
       <Tooltip
         content={disabled ? policy?.unavailableMessage : "Export"}
         open={open ? false : undefined}
       >
-        <DropdownMenuTrigger asChild={true}>{button}</DropdownMenuTrigger>
+        <DropdownMenuTrigger asChild={true} disabled={disabled}>
+          <span tabIndex={disabled ? 0 : -1} className="inline-flex">
+            {button}
+          </span>
+        </DropdownMenuTrigger>
       </Tooltip>
       <DropdownMenuContent side="bottom" className="print:hidden">
         <DropdownMenuLabel className="text-xs text-muted-foreground">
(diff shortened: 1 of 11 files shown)


kirangadhave:
@cubic-dev-ai

frontend/src/components/data-table/export-actions.tsx
```diff
@@ -149,6 +161,18 @@ export const ExportMenu: React.FC<ExportActionProps> = (props) => {
     if (!result) {
       return;
     }
+    if (
+      policy &&
+      result.size_bytes != null &&
+      result.size_bytes > policy.limitBytes
```
mscolnick: so the result could be sent over the wire, but not downloaded? wouldnt this require getting the full result anyways?

kirangadhave: yes we pay the serialization cost on the backend, but we only get the url and size here (result object includes `url`, `size_bytes` and `filename`). The actual download happens down on line 168 in `downloadByUrl`

The crash happens when sending the large serialized string over json not when serializing in the backend.

kirangadhave: Updates: my initial hypothesis was that we are hitting the V8 limit on a single string size (512MB) and that's what causing the crash. I was wrong. The V8 actually handles strings upto 1G fine. 
Usually the `result` object has a data url, but in vscode extension we disable virtual file and use the `result.url` is actually a data uri which encodes all data.

Sending it over the wire is fine for VSCode. 

When we try to download the data uri (we simulate a click on `a` tag in iframe), the Electron main process SIGTRAPs when handling the huge URI. I verified this by creating a standalone electron app and trying to download a huge data URI in there.

The above gate fixes that issue by never triggering the `downloadByUrl`.

kirangadhave:
@cubic-dev-ai

frontend/src/components/data-table/export-actions.tsx
```diff
@@ -94,12 +101,19 @@ const labelForCopyFormat = (format: CopyFormat): string =>
 export const ExportMenu: React.FC<ExportActionProps> = (props) => {
   const { locale } = useLocale();
   const [open, setOpen] = React.useState(false);
+  const policy = useAtomValue(downloadSizeLimitAtom);
```
manzt, line 104: nice

manzt (approved):
looks great. nice solution

PR #9521, opened 2026-05-12
feat(files): use multipart/form-data for /api/files/create

mscolnick:
Switches the HTTP upload path from base64-in-JSON to multipart/form-data,
eliminating the 33% size overhead and pairing naturally with File/Blob on
the frontend. The WASM/Pyodide bridge keeps its JSON+base64 wire format
since the JS<->Py RPC boundary cannot carry multipart; both transports
share a single client-side FileCreateInput interface.

Also adds:
- path-traversal hardening in OSFileSystem.create_file_or_directory
- parse_multipart_request helper in marimo/_server/api/utils.py

Two followups:
- streaming so the full file doesn't need to be loaded into memory in the server
- batching uploads on the frontend


frontend/src/core/network/requests-network.ts
```diff
@@ -298,10 +311,15 @@ export function createNetworkRequests(): EditRequests & RunRequests {
     },
     sendCreateFileOrFolder: async (request) => {
       await waitForConnectionOpen();
+      const formData = new FormData();
+      formData.append("path", request.path);
+      formData.append("type", request.type);
+      formData.append("name", request.name);
+      if (request.file) {
```
kirangadhave, line 318: do we want to make the request if there is no file?

kirangadhave (approved):
🚀

frontend/src/core/network/requests-network.ts
mscolnick: yea, i could check (will do it in a followup)

PR #9527, opened 2026-05-12
feat(files): stream uploads to disk instead of buffering

mscolnick:
Previously /api/files/create read the entire UploadFile into memory
before writing to disk, peaking at ~100 MB for a 100 MB upload. Now we
drain the upload in 1 MiB chunks straight to a `.part` temp file and
atomically rename on success, so peak memory stays bounded regardless
of file size and a failed upload never leaves a half-written file at
the final path.

- `OSFileSystem.stream_create_file` does the chunked write with atomic
  rename and cleanup on failure
- `parse_multipart_request` now hands callers un-read `UploadFile`
  handles (instead of pre-read bytes), so streaming is possible without
  giving up the small-payload `.read()` path
- Directories and the default-template notebook still go through the
  in-memory `create_file_or_directory` path; only real file content
  streams

### DIFF
--- a/marimo/_server/api/endpoints/file_explorer.py
+++ b/marimo/_server/api/endpoints/file_explorer.py
@@ -4,11 +4,15 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from starlette.authentication import requires
+from starlette.exceptions import HTTPException
 
 from marimo import _loggers
 from marimo._server.api.deps import AppState
 from marimo._server.api.utils import parse_multipart_request, parse_request
-from marimo._server.files.os_file_system import OSFileSystem
+from marimo._server.files.os_file_system import (
+    OSFileSystem,
+    UploadTooLargeError,
+)
 from marimo._server.models.files import (
     FileCopyRequest,
     FileCopyResponse,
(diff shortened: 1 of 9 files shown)


marimo/_server/files/os_file_system.py
```diff
@@ -133,33 +153,32 @@ def open_file(self, path: str, encoding: str | None = None) -> str | bytes:
...
         if name.strip() == "":
             raise ValueError("Cannot create file or directory with empty name")
-        # Names that traverse out of `path` or escape via separators are
-        # rejected. Validation belongs here (not in the endpoint) so every
-        # caller of OSFileSystem — HTTP, WASM bridge, scripts — is covered.
-        if (
-            "/" in name
-            or "\\" in name
-            or "\x00" in name
-            or name in (".", "..")
```
kirangadhave, line 156: new code does not handle `. or ..`.

maybe also add a test for this, so regression is caught.

marimo/_server/api/endpoints/file_explorer.py
```diff
@@ -120,15 +120,20 @@ async def create_file_or_directory(
...
+            if upload is not None and parsed.body.type in ("file", "notebook"):
+                info = await file_system.stream_create_file(
+                    parsed.body.path, parsed.body.name, upload
+                )
+            else:
+                info = file_system.create_file_or_directory(
+                    parsed.body.path, parsed.body.type, parsed.body.name, None
+                )
         return FileCreateResponse(success=True, info=info)
     except Exception as e:
```
kirangadhave, line 145: `stream_create_file` throws `ValueError` for large files. We catch and return this as 200 response with error message. We should return `413`

marimo/_server/files/os_file_system.py
```diff
@@ -192,6 +211,62 @@ def create_file_or_directory(
...
+
+        Avoids loading the full payload into memory (the HTTP multipart
+        path can otherwise buffer 100 MB at once). Writes to a ``.part``
+        temp file and atomically renames on success so a failed upload
+        doesn't leave a half-written file at the final path.
+        """
+        self._validate_create_name(name)
+
+        full_path = Path(path) / name
+        full_path = _generate_unique_path(full_path)
```
kirangadhave: We can have a toctou race here. We are getting the file path, then do a bunch of work and the actually write on line 257 with replace. if some other concurrent upload happens meanwhile and picks up the same name we might have overwrites.

kirangadhave (changes requested):

marimo/_server/files/os_file_system.py
mscolnick: handled in `if name in DISALLOWED_NAMES:`. it was dup

marimo/_server/api/endpoints/file_explorer.py
mscolnick: yea, something i can look into a followup. but would diverge from the wasm path

kirangadhave (approved):
🚀

PR #9528, opened 2026-05-12
perf(frontend): parallelize file uploads with bounded concurrency

mscolnick:
Add Semaphore + mapWithConcurrency utility (cap N concurrent promises)
and apply it with N=5 to: file-tree uploads (previously sequential),
file-tree move/refreshAll, filesToBase64, and vega data resolution
(single shared semaphore across the recursive spec traversal).

### DIFF
--- a/frontend/src/utils/semaphore.ts
+++ b/frontend/src/utils/semaphore.ts
@@ -5,6 +5,7 @@
  * tasks. Waiters are released in FIFO order.
  */
 export class Semaphore {
+  private readonly maxPermits: number;
   private permits: number;
   private waiters: Array<() => void> = [];
 
@@ -14,6 +15,7 @@ export class Semaphore {
         `Semaphore permits must be a positive integer, got ${permits}`,
       );
     }
+    this.maxPermits = permits;
     this.permits = permits;
   }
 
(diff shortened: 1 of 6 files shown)


frontend/src/utils/semaphore.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+  }
+
+  release(): void {
+    const next = this.waiters.shift();
+    if (next) {
+      next();
+      return;
+    }
+    this.permits++;
```
kirangadhave, line 53: if we keep calling `release()`, the permits keep increasing beyond the initial value.

frontend/src/utils/semaphore.ts
```diff
@@ -0,0 +1,78 @@
...
+  get available(): number {
+    return this.permits;
+  }
+
+  /** Number of waiters queued for a permit. */
+  get pending(): number {
+    return this.waiters.length;
+  }
+
+  acquire(): Promise<void> {
```
kirangadhave, line 32: maybe add an `AbortSignal`? something to support cancelling?

It can be future work

frontend/src/utils/semaphore.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (this.permits > 0) {
+      this.permits--;
+      return Promise.resolve();
+    }
+    return new Promise<void>((resolve) => {
+      this.waiters.push(resolve);
+    });
+  }
+
+  release(): void {
```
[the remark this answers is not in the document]
mchav: Or if we must keep them public acquire should return an idempotent release token specifically for this. That way double release is a no-op.

kirangadhave (approved):
🚀

PR #9532, opened 2026-05-12
fix(mpl): contain comm callback errors in mpl_interactive

kirangadhave:
## Summary

A raising callback inside `mpl_interactive._handle_comm_msg` (e.g. matplotlib's `AttributeError: 'Axes' object has no attribute '_pan_start'` from `end_pan` without a prior `start_pan`) was propagating up through `MarimoComm.handle_msg` → `WIDGET_COMM_MANAGER.receive_comm_message` → the kernel's request handler, terminating `asyncio.run` and killing the kernel subprocess. The uvicorn parent kept serving so the UI just reported "stopped responding".

This wraps the `handle_json` and `_handle_download` calls in `try/except Exception` and logs via `LOGGER.exception` so a buggy callback is contained while the traceback still surfaces for diagnosis. `BaseException` is intentionally not caught.

Fixes the kernel-crash half of the matplotlib pan/zoom issue (see linked Linear issue). The underlying matplotlib `_pan_start` bug is tracked separately.

## Test plan

- [x] Manual repro via the linked Linear repro (pan → re-run cell → zoom → pan): exception logged, kernel stays alive, figure remains responsive.
- [x] Direct programmatic injection: `_handle_comm_msg` with a raising `handle_json` returns cleanly and a follow-up message is processed.

### DIFF
--- a/marimo/_plugins/ui/_impl/from_mpl_interactive.py
+++ b/marimo/_plugins/ui/_impl/from_mpl_interactive.py
@@ -274,23 +274,37 @@ class mpl_interactive(UIElement[ModelIdRef, dict[str, Any]]):
             self._handle_download(fmt)
             return
         else:
-            # Forward to figure manager (mouse events, toolbar actions, etc.)
-            self._figure_manager.handle_json(event)  # type: ignore[no-untyped-call]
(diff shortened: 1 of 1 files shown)


Light2Dark (approved):
should we add tests?

PR #9552, opened 2026-05-14
chore(asyncio): hardening pass — utilities, deprecated API cleanup, lifespan fix

mscolnick:
- Add marimo/_utils/asyncio_utils.py with supervised_task, fire_and_forget,
  and cancel_and_wait helpers (strong-ref task tracking, exception logging,
  shutdown dance).
- Fix Lifespans._manager swallowing CancelledError, which was breaking
  shutdown propagation through the ASGI lifespan chain.
- Replace the type-confused run_async() helper in session_manager with
  fire_and_forget at its three callers.
- Replace 11 deprecated asyncio.get_event_loop() sites with
  get_running_loop() (or asyncio.to_thread in tool_manager).
- Migrate api/lifespans.py and execution.py hand-rolled set+callback /
  cancel-and-await patterns to the new helpers.
- ConnectionDistributor now pins the loop at start() and reuses it on stop().

### DIFF
--- a/marimo/_server/api/lifespans.py
+++ b/marimo/_server/api/lifespans.py
@@ -117,9 +117,7 @@ async def mcp(app: Starlette) -> AsyncIterator[None]:
             LOGGER.warning(f"Failed to connect MCP servers: {e}")
             return None
 
-    # ``background_connect_mcp_servers`` swallows its own errors, so the
-    # supervisor's logging callback would never fire — but suppress it
-    # explicitly to make the contract obvious (this task is awaited).
+    # Awaited below — opt out of supervisor logging to avoid duplicate logs.
     task = supervised_task(
         background_connect_mcp_servers(),
         name="mcp.connect",
@@ -129,8 +127,6 @@ async def mcp(app: Starlette) -> AsyncIterator[None]:
 
     yield
 
-    # No-op if the task already completed normally; otherwise cancel and
-    # await. After this returns, ``task.done()`` is True.
     await cancel_and_wait(task)
     if task.cancelled():
         return
--- a/marimo/_server/api/lifespans.py
+++ b/marimo/_server/api/lifespans.py
@@ -27,6 +27,7 @@ from marimo._server.utils import initialize_mimetypes
 from marimo._server.uvicorn_utils import close_uvicorn
 from marimo._server.workspace import NEW_FILE
 from marimo._session.model import SessionMode
+from marimo._utils.asyncio_utils import cancel_and_wait, supervised_task
 from marimo._utils.subprocess import cancel_pending_reaps
 
 if TYPE_CHECKING:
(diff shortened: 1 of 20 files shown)


marimo/_server/api/lifespans.py
```diff
@@ -119,22 +117,30 @@ async def background_connect_mcp_servers() -> MCPClient | None:
             LOGGER.warning(f"Failed to connect MCP servers: {e}")
             return None
 
-    task = asyncio.create_task(background_connect_mcp_servers())
-    background_tasks.add(task)  # Keep a reference to prevent GC
-    task.add_done_callback(background_tasks.discard)  # Clean up when done
+    task = supervised_task(
```
kirangadhave, line 121: `supervised_task` will log any exception from `background_connect_mcp_servers()` via its done-callback, and the caller below also has `except Exception: LOGGER.error(...)` — connect failures will be logged twice. Either drop the supervisor's logging here (`on_exception=lambda _: None`) or use plain `asyncio.create_task` since this caller handles its own errors.

marimo/_server/api/lifespans.py
```diff
@@ -119,22 +117,30 @@ async def background_connect_mcp_servers() -> MCPClient | None:
...
+        name="mcp.connect",
+        registry=background_tasks,
+    )
 
     yield
 
-    # Shutdown
-    task.cancel()
+    if not task.done():
+        await cancel_and_wait(task)
```
kirangadhave: This branching can collapse: `cancel_and_wait` already handles the done case, so call it unconditionally and then just check `task.cancelled() or not task.done()` to bail out before reading the result. Less conditional, same semantics.

marimo/_server/api/lifespans.py
```diff
@@ -119,22 +117,30 @@ async def background_connect_mcp_servers() -> MCPClient | None:
...
+
+    # ``task.result()`` re-raises CancelledError (which is a BaseException
+    # on 3.11+ and not caught by ``except Exception``). Short-circuit if
+    # the loop cancelled the task before we got here.
+    if task.cancelled():
+        return
+
     try:
-        mcp_client = await task
+        mcp_client = task.result()
```
kirangadhave: Consider splitting into two `try/except` blocks. Exceptions from `task.result()` are connect-time failures ("Error connecting to MCP servers"), while exceptions from `disconnect_from_all_servers()` are disconnect-time failures. The merged catch produces a misleading "Error during MCP cleanup" log for connect-time errors.

marimo/_utils/asyncio_utils.py
```diff
@@ -0,0 +1,158 @@
...
+T = TypeVar("T")
+
+# Module-level registry for fire_and_forget. The event loop holds only
+# weak references to tasks; without a strong reference here the task can
+# be garbage-collected mid-flight. The done-callback discards on
+# completion, so this never grows unbounded under normal use.
+_BACKGROUND_TASKS: set[asyncio.Task[Any]] = set()
+
+
+def _log_task_exception(task: asyncio.Task[Any]) -> None:
```
kirangadhave, line 33: The "awaited tasks that handle their own errors should not use `supervised_task`" rule is buried in this inline comment — please surface it in the public `supervised_task` docstring. Note that `background_connect_mcp_servers()` at `lifespans.py:121` currently violates this (awaited + handled, but still supervised → duplicate logs).

tests/_utils/test_asyncio_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+async def test_cancel_and_wait_swallows_cancelled_error() -> None:
+    async def sleeper() -> None:
+        await asyncio.sleep(10)
+
+    task = asyncio.create_task(sleeper())
+    await asyncio.sleep(0)
+    await cancel_and_wait(task)
+    assert task.cancelled() or task.done()
```
kirangadhave: `task.cancelled()` implies `task.done()`, so the `or` here is a tautology. Probably just:

```python
assert task.cancelled()
```

marimo/_server/api/lifespans.py
mscolnick: done

PR #9554, opened 2026-05-15
refactor:  Add `kernel_session()` as context manager, DRY up tests

mscolnick:
- Add `kernel_session()` context manager to `marimo/_runtime/kernel_lifecycle.py`
  and use it from `launch_kernel` so create+teardown are paired by the type
  system.
- Add `tests/_runtime/_helpers/` with canonical `MockStream`/`MockStdout`/
  `MockStderr`/`MockStdin`, `default_app_metadata`/`default_user_config`
  factories, and `mocked_kernel_session()` returning a `TestKernel` bundle.
- Rewire `tests/conftest.py` fixtures (`k`, `strict_kernel`, `lazy_kernel`,
  `run_mode_kernel`, `mocked_kernel`, etc.) to delegate to
  `mocked_kernel_session()`. `MockedKernel` is preserved as a thin back-compat
  wrapper.


tests/_runtime/_helpers/session.py
```diff
@@ -0,0 +1,119 @@
...
+import contextlib
+import dataclasses
+import sys
+from typing import TYPE_CHECKING
+
+from marimo._config.config import DEFAULT_CONFIG
+from marimo._messaging.print_override import print_override
+from marimo._runtime.kernel_lifecycle import KernelArgs, kernel_session
+from marimo._runtime.marimo_pdb import MarimoPdb
+from marimo._runtime.virtual_file import VirtualFileStorageType
```
kirangadhave, line 20: can be typing only import

kirangadhave (approved):
🚀 only nits

tests/_runtime/_helpers/session.py
kirangadhave: Only applies to `VirtualFileStorageType`

mscolnick: probably. maybe we can enforce with linting?

PR #9561, opened 2026-05-15
Fix: expand dict transformation

### LINKED ISSUE #4583: New dataframe transform: Polars Native Expand Dict Transformation
### Description

I'm proposing to extend `Expand Dict` dataframe transformation that implements Polars' column unnest functionality. Currently, marimo has an `Expand Dict` feature, but it fails when a column contains null values, throwing the error: `'NoneType' object cannot be converted to 'PyDict'`.

### Suggested solution

The native Polars method df.unnest("column_name") handles null values gracefully without failing. This new transformation would leverage the Polars unnest method directly, making it more robust when working with dictionaries that contain null values. 

It would allow users to flatten nested dictionaries into separate columns without having to pre-process the data to handle null values. 

I am willing to submit a PR to implement this functionality if desired.

### Alternative

_No response_

### Additional context

Error message:
Image

Shamik-07:
## 📝 Summary

Using narwahls to convert all backend to polars and then using the `unnest` function of polars for expanding the dict and then convert it back to the original backend. 
Closes #4583 

https://github.com/user-attachments/assets/8213e093-a3c9-4de5-88a2-badce54d19c8

## 📋 Pre-Review Checklist

## ✅ Merge Checklist


tests/_plugins/ui/_impl/dataframes/test_handlers.py
```diff
@@ -1733,22 +1733,32 @@ def test_explode_columns(df: DataFrameType) -> None:
...
-    )
     @pytest.mark.parametrize(
         ("df", "expected"),
         list(
             zip(
                 create_test_dataframes(
-                    {"A": [{"foo": 1, "bar": "hello"}], "B": [1]}
+                    {"A": [{"foo": 1, "bar": "hello"}], "B": [1]},
                 ),
                 create_test_dataframes(
```
mscolnick, line 1767: polars should already be created in this `create_test_dataframes`. so we dont need to add the dataframe below

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -492,7 +492,12 @@ def handle_explode_columns(
     def handle_expand_dict(
         df: DataFrame, transform: ExpandDictTransform
     ) -> DataFrame:
-        return df.explode(transform.column_id)
+        # Convert to Polars for native unnest, which handles null values
+        # correctly (fixes: 'NoneType' cannot be converted to 'PyDict' issue #4583)
+        collected_df, undo = collect_and_preserve_type(df)
+        polars_df = collected_df.to_polars()
```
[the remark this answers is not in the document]
mscolnick: this is correct. we cannot use polars directly and can only use narwhals. 

we can check if the narwhals backend/implementation in polars

mscolnick (dismissed):
need to take an optional dep on polars

tests/_plugins/ui/_impl/dataframes/test_handlers.py
Shamik-07: Fixed

Shamik-07:
> need to take an optional dep on polars

Done. 
Added `allow_none_equals_nan` for `assert_frame_equal_with_nans` as `None!=NaN`, which was causing the use of `assert_frame_equal_with_nans` to fail for `test_expand_dict`.

Shamik-07:
There are some pandas CI errors that i am looking into.

Shamik-07:
> There are some pandas CI errors that i am looking into.

This is happening because of data conversion mismatch between pandas and narwahls with mixed data types
```
"Could not convert '3' with type str: tried to convert to double"
```
for 
```
test_print_code_result_matches_actual_transform_pandas(
    transform=ExpandDictTransform(
        type=TransformType.EXPAND_DICT,
        column_id='strings',
    ),
)
```
So my only option is to fallback to pandas backend processing separately for the `unnest`.
This should fix it.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -492,7 +492,24 @@ def handle_explode_columns(
...
+        collected_df, undo = collect_and_preserve_type(df)
+        native_df = collected_df.to_native()
+
+        # Keep pandas handling fully pandas-native so mixed/object columns in
+        # unrelated fields do not trigger Arrow coercion errors.
+        if nw.dependencies.is_pandas_dataframe(native_df):
+            import pandas as pd
+
+            result_df = native_df.copy()
+            expanded = pd.json_normalize(
```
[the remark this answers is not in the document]
Shamik-07: Interesting...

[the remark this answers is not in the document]
Shamik-07: @mscolnick need more feedback on this.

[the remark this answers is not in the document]
mscolnick: i actually dont have enough context on the issue. is `df.explode(transform.column_id)` not sufficient enough? that seems pretty "native" to me

Shamik-07:
@mscolnick 
Apart from the cubic ai's review comment, we started this because `expand_dict` was having trouble with `NaN` rows.
We have certainly fixed that, but the issue is regarding deeply nested dicts being expanded by pandas if we are using `pd.json_normalize` vs. polars `unnest` will only unnest a single level by default. This will yield discrepancies between both the backend outputs.

**Current Behaviour**
e.g. df with nested columns
image

polars unnest
image


pandas unnest
image


**Proposal**
Let's unnest only a single level in pandas thus matching polars unnest. 
This will always give us the same backend agnostic output 
image


I have added the fix to this PR, please let me know your thoughts.

Shamik-07:
https://github.com/user-attachments/assets/913bbf61-2f80-40ad-b098-6e1d8ebb0bcd

Shamik-07:
@mscolnick Just a gentle nudge.

kirangadhave:
@cubic-dev-ai

kirangadhave:
@Shamik-07 can you please update the video in the PR to a higher res version? I'm having difficulty reading text in it

Shamik-07:
> @Shamik-07 can you please update the video in the PR to a higher res version? I'm having difficulty reading text in it

Unfortunately, due to GH upload limitations, i can't upload a high res video. 
PFA the notebook instead. 
[issue_4583_expand_dict_null_values.py](https://github.com/user-attachments/files/28482361/issue_4583_expand_dict_null_values.py)

kirangadhave (changes requested):
please address copilot review. We can do a manual round of reviews after that.

Shamik-07:
> please address copilot review. We can do a manual round of reviews after that.

Thanks, fixed.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +498,35 @@ def handle_explode_columns(
...
+                result_df.pop(transform.column_id).map(
+                    lambda value: {} if value is None else value
+                ),  # type: ignore[arg-type]
+                max_level=0,
+            )
+            expanded.index = result_df.index
+            return undo(nw.from_native(result_df.join(expanded)))
+
+        DependencyManager.polars.require(
```
kirangadhave: non pandas dataframes are forced to go through `polars` conversion here. For `duckdb` or `ibis`, that means forcing `polars` installation. We should not do that.

Narwhals has `struct.field`, so we can do:

```python
schema = df.collect_schema()
fields = [f.name for f in schema[col].fields]
df.with_columns(
    [nw.col(col).struct.field(f).alias(f) for f in fields]
).drop(col)
```

This approach also stays lazy. Pandas approach with `json_serialize` is correct.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +498,35 @@ def handle_explode_columns(
...
+
+            result_df = native_df.copy()
+            # max_level=0 was used so that pandas doesn't recursively unnest dicts
+            # causing mismatch between pandas vs. polars df
+            # using the map function to replace the None values
+            # Replace top-level null rows so pandas 2.x can normalise them, needed for
+            # older versions of pandas running on py310 otherwise CI will fail
+            expanded = pd.json_normalize(
+                result_df.pop(transform.column_id).map(
+                    lambda value: {} if value is None else value
```
kirangadhave: Also check for `NaN` here

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +498,35 @@ def handle_explode_columns(
...
+        collected_df, undo = collect_and_preserve_type(df)
+        native_df = collected_df.to_native()
+
+        # Keep pandas handling fully pandas-native so mixed/object columns in
+        # unrelated fields do not trigger Arrow coercion errors.
+        if nw.dependencies.is_pandas_dataframe(native_df):
+            import pandas as pd
+
+            result_df = native_df.copy()
+            # max_level=0 was used so that pandas doesn't recursively unnest dicts
```
kirangadhave: the comment is narrating the code, simplify to explain the why instead.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +498,35 @@ def handle_explode_columns(
...
+            # Replace top-level null rows so pandas 2.x can normalise them, needed for
+            # older versions of pandas running on py310 otherwise CI will fail
+            expanded = pd.json_normalize(
+                result_df.pop(transform.column_id).map(
+                    lambda value: {} if value is None else value
+                ),  # type: ignore[arg-type]
+                max_level=0,
+            )
+            expanded.index = result_df.index
+            return undo(nw.from_native(result_df.join(expanded)))
```
kirangadhave, line 534: Duplicate column names after unnest will throw error here. Handle it gracefully.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +498,35 @@ def handle_explode_columns(
...
+
+        DependencyManager.polars.require(
+            why="to expand dict/struct columns for non-pandas backends"
+        )
+        polars_df = collected_df.to_polars()
+        unnested = polars_df.unnest(transform.column_id)
```
kirangadhave: Duplicate column names after unnest will throw error here. Handle it gracefully.

tests/_plugins/ui/_impl/dataframes/test_handlers.py
```diff
@@ -47,8 +47,8 @@
 
 pytest.importorskip("ibis")
 pd = pytest.importorskip("pandas")
-pytest.importorskip("polars")
 pytest.importorskip("pyarrow")
+pytest.importorskip("polars")
```
kirangadhave, line 51: import order change is unnecessary

kirangadhave (changes requested):
Requesting a major change which should remove hard dependency on `polars` for other backends which support `structs`.

Also please address other nits.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
Shamik-07: Are you sure that all backends, which support `struct` schema would necessarily have `struct.field`?

Shamik-07:
@kirangadhave 

Thanks for the review comments. 
I have addressed your review comments.
I still have a question, are you sure that all backends, which support struct schema would necessarily have struct.field?

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
kirangadhave: `struct.field` is part of the narwhals' public api, so it should translate okay. We support `pandas`, `polars`, `ibis`, `pyarrow` and `duckdb` backends. Except `pandas` all others have `struct.field` support.

It might be worth adding tests though. Thoughts?

Shamik-07: Yes, you are correct about narwahl's api. 

[create_test_dataframes](https://github.com/Shamik-07/marimo/blob/fix/expand_dict_transformation/tests/_plugins/ui/_impl/dataframes/test_handlers.py#L58) is used for generating the dataframes for the expand dict tests. 

To support `duckdb`, should i create a new test for expand dict or it doesn't matter as the [tests](https://github.com/Shamik-07/marimo/blob/fix/expand_dict_transformation/tests/_plugins/ui/_impl/dataframes/test_handlers.py#L48-L52) are only supported for `pandas, polars, ibis, pyarrow`?

tests/_plugins/ui/_impl/dataframes/test_handlers.py
```diff
@@ -1760,11 +1781,82 @@ def test_expand_dict(df: DataFrameType, expected: DataFrameType) -> None:
...
+        nw_expected = collect_df(expected)
+        result_cols = sorted(nw_result.columns)
+        expected_cols = sorted(nw_expected.columns)
+        assert_frame_equal_with_nans(
+            nw_expected.select(expected_cols),
+            nw_result.select(result_cols),
+            allow_none_equals_nan=True,
+        )
+
+    @pytest.mark.xfail(
```
kirangadhave: `xfail` is used to mark tests where failure is expected. Raising error is the expected __pass__ case here, so don't mark as xfail.

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +497,81 @@ def handle_explode_columns(
...
+                    return {}
+                if isinstance(value, float) and math.isnan(value):
+                    return {}
+                return value
+
+            # Keep expansion shallow and replace top-level null/nan entries so
+            # pandas and other backends agree on expand-dict behaviour.
+            expanded = pd.json_normalize(
+                result_df.pop(transform.column_id).map(normalise_empty_dict),
+                # type: ignore[arg-type]
```
kirangadhave: type ignore comment on new line doesn't do anything

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +497,81 @@ def handle_explode_columns(
...
+                raise nw.exceptions.InvalidOperationError(
+                    "Cannot expand dict because it would duplicate existing "
+                    f"columns: {duplicate_columns}"
+                )
+            expanded.index = result_df.index
+            return undo(nw.from_native(result_df.join(expanded)))
+
+        schema = collected_df.collect_schema()
+        dtype = schema.get(transform.column_id)
+        fields = getattr(dtype, "fields", None)
```
kirangadhave: use `isinstance(dtype, nw.Struct)` here?

marimo/_plugins/ui/_impl/dataframes/transforms/handlers.py
```diff
@@ -497,7 +497,81 @@ def handle_explode_columns(
...
+                        .struct.field(field_name)
+                        .alias(field_name)
+                        for field_name in field_names
+                    ]
+                )
+                .drop(transform.column_id)
+                .select(expanded_columns)
+            )
+
+        raise nw.exceptions.InvalidOperationError(
```
kirangadhave, line 569: Let's add a test for this

kirangadhave (changes requested):
Thanks again for the changes. We will be good to merge after this set of changes. Especially the test related ones are must do.

Shamik-07:
> Thanks again for the changes. We will be good to merge after this set of changes. Especially the test related ones are must do.

Thank you again for the detailed review.

kirangadhave (approved):
:rocket:

kirangadhave:
Thanks @Shamik-07 for your contribution and addressing multiple review rounds

kirangadhave (approved):

PR #9586, opened 2026-05-18
refactor: support multiple markdown flavors in `marimo export md`

peter-gy:
## 📝 Summary

Refactors the internals of `marimo export md` so markdown export can target multiple markdown flavors through a shared renderer interface. Supports https://github.com/marimo-team/quarto-marimo/ and https://github.com/marimo-team/jupyter-book-marimo.

Prior to this, markdown export only needed to distinguish between marimo's default markdown format (pymdown) and Quarto, so an `is_qmd` branch in the exporter was simple enough to handle markdown output differences. As we add MyST /
Jupyter Book support, that does not scale well as each target has its own way to represent executable cells, page-level metadata, callouts, options, and tabsets.

This PR introduces explicit markdown flavors for:

- pymdown (already existed): marimo's existing markdown authoring/export surface
- qmd (already existed): Quarto markdown
- mystmd (new addition, hence the `enhancement` label on this PR): MyST markdown, including the `{marimo} / {marimo-config}` surface used by marimo-team/jupyter-book-marimo#1

The source authoring surface is still assumed to be marimo markdown / pymdown. The flavor layer maps from that source representation into each target format.

## Details

This refactor sets up marimo core to support markdown flavor-specific rendering of callouts, tabs, etc. It allows us to first build a small intermediate export document instead of directly concatenating strings and if-else branching on detected markdown flavor.

image

**As part of a follow-up**, flavor implementations can then decide how to render:

- top-level metadata, e.g. YAML frontmatter versus `{marimo-config}` for mystmd
- executable code cell syntax
- cell options
- PyMdown admonitions / callouts
- tabsets
- unknown directive fallbacks

For Quarto, PyMdown admonitions are mapped to Quarto callouts and tab groups are mapped to panel tabsets.
The QMD exporter no longer injects `filters: marimo-team/marimo`, since the Quarto marimo engine extension now auto-detects marimo code blocks.

For MyST, marimo cells are emitted as `{marimo}` directives and page-level execution config is emitted as `{marimo-config}`, because executable plugins do not currently get a clean frontmatter access path. The spelling remains `unparsable` throughout; no support for `unparseable` spelling to stay coherent and to ensure we do not express the same thing in multiple ways in the codebase.

### PyMdown

https://facelessuser.github.io/pymdown-extensions/

```md
/// Tip | Data sources panel

    Click the database "barrel" icon in the left toolbar
    to see all dataframes and in-memory tables that your notebook has access to.
///
```

Screenshot 2026-05-18 at 21 11 49

### QMD

https://quarto.org/docs/authoring/markdown-basics.html

```qmd
::: {.callout-tip title="Data sources panel"}

Click the database "barrel" icon in the left toolbar to see all dataframes and in-memory tables that your notebook has access to.

:::
```

Screenshot 2026-05-18 at 21 04 50

### MyStmd

https://mystmd.org/

```md
:::{tip} Data sources panel

Click the database "barrel" icon in the left toolbar to see all dataframes and in-memory
tables that your notebook has access to.

:::
```

Screenshot 2026-05-18 at 21 07 24


marimo/_cli/export/commands.py
```diff
@@ -363,6 +365,12 @@ def export_callback(file_path: MarimoPath) -> ExportResult:
     type=bool,
     help=_sandbox_message,
 )
+@click.option(
```
dmadisetti, line 408: can we just detect on the file extension?

marimo/_convert/markdown/flavor/qmd.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+        guard = "```"
+        while guard in code:
+            guard += "`"
+
+        head = self._code_fence_head(guard, language, attribute_str)
+        parts = [head, code, guard, ""]
+        return "\n".join(parts)
+
+    def render_directive(self, block: DirectiveBlock) -> str:
```
dmadisetti: I'm not sure we should render these differently?

marimo/_convert/markdown/to_ir.py
```diff
@@ -48,6 +53,10 @@
 
 MARIMO_MD = "marimo-md"
 MARIMO_CODE = "marimo-code"
+_MYST_MARIMO_HEADER_RE = re.compile(
```
dmadisetti: Is there a reference parse or format we can document for reference?

marimo/_convert/markdown/to_ir.py
```diff
@@ -318,6 +336,9 @@ def __init__(
         self.preprocessors.register(
             FrontMatterPreprocessor(self), "frontmatter", 100
         )
+        self.preprocessors.register(
```
dmadisetti: Can we conditionally add this?

marimo/_tutorials/markdown_format.md
```diff
@@ -135,14 +134,10 @@ print("This code cell has a syntax error"
 and on notebook save, will annotate the cell for you:
 
 ````md
```
dmadisetti, line 137: Hmm, this is a content change. Any issues keeping this? We have intentional errors in other tutorials

dmadisetti (commented):
I had a couple reservations initially, but I think it's a good abstraction.

I think my major concern is handling admonitions differently. Ideally we don't want to change user code- so if anything I think they should be less of a parse concern and more of a render one.

marimo/_convert/markdown/flavor/qmd.py
```diff
@@ -0,0 +1,253 @@
...
+                transformed.extend(split_pymdown_blocks(block.text))
+            else:
+                transformed.append(block)
+        return group_pymdown_tabs(transformed)
+
+    def prepare_metadata(
+        self, metadata: dict[str, str | list[str]]
+    ) -> dict[str, str | list[str]]:
+        qmd_metadata = metadata.copy()
+        qmd_metadata.setdefault("filters", [_MARIMO_QMD_FILTER])
```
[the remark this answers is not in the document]
dmadisetti: The engine addition no longer needs a `filter` to work as well.

marimo/_cli/export/commands.py
peter-gy: we can match on `qmd`, but keeping the `--flavor` flag is needed for mystmd as it also `md` just as the default pymdown format.

marimo/_tutorials/markdown_format.md
peter-gy: I removed `unparseable="true"` so that we do not need to maintain both `unparseable` and `unparsable` as supported options. In all other parts of the codebase we have `unparsable` and it feels like having the alias `unparseable` in addition just creates noise.

marimo/_convert/markdown/flavor/qmd.py
peter-gy: mapping from pymdown syntax to other targets here.

> Ideally we don't want to change user code- so if anything I think they should be less of a parse concern and more of a render one.

I understand this concern but one of the main objectives of this PR is to have a place where we can define these different logics so that users who just want to use a static renderer like quarto or jupyter book do not need to be familiar with these custom syntaxes (only with that of marimo/pymdown) but can just run an export with the right flavor and let quarto-marimo / jupyter-book-marimo take care of the rest.

marimo/_tutorials/markdown_format.md
dmadisetti: That's not what this block was. I think that was just a typo

peter-gy:
@dmadisetti as per our conversation I reduced the scope of this PR to carry only the flavor-based refactoring and the introduction of `mystmd`,  without baking in callout/tabset etc. rendering-specific logic.

dmadisetti (approved):
LGTM! thanks for following up with the callouts etc

dmadisetti:
Type check and test failure clearly unrelated

PR #9588, opened 2026-05-18
fix(kernel): callback bug fixes in cache clear, dataset connections, and model updates

mscolnick:
- cache: ClearCacheCommand now dispatches Loader.clear() unconditionally
  instead of gating on BasePersistenceLoader, so MemoryLoader (and any
  other subclass) is also purged. Loader.clear is a safe no-op by default.

- datasets: preview_datasource_connection used get_engine_catalog, which
  rejected raw QueryEngine connections (e.g. sqlite3.Connection) and left
  the frontend with no DataSourceConnectionsNotification. Switch to
  get_sql_connection so engine_to_data_source_connection can emit a
  connection with an empty databases list.

- kernel_request_handlers: document the empty-state contract in
  _handle_receive_model_message — ``if ui_element_id and state`` only
  filters empty diffs, since receive_comm_message returns (id, state)
  only for ModelUpdateMessage with a non-empty id, observe callbacks
  already fire inside comm.handle_msg, and resulting state setters take
  the elif branch. Add a regression test pinning both cases.

### DIFF
--- a/marimo/_runtime/callbacks/cache.py
+++ b/marimo/_runtime/callbacks/cache.py
@@ -27,14 +27,13 @@ class CacheCallbacks:
     async def clear_cache(self, request: ClearCacheCommand) -> None:
         del request
         from marimo._save.cache import CacheContext
-        from marimo._save.loaders import BasePersistenceLoader
 
         ctx = get_context()
         saved = 0
         for obj in ctx.globals.values():
             if isinstance(obj, CacheContext):
-                if isinstance(obj.loader, BasePersistenceLoader):
-                    obj.loader.clear()
+                # Loader.clear is a no-op by default; safe on any subclass.
+                obj.loader.clear()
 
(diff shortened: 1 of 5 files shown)


marimo/_runtime/callbacks/cache.py
```diff
@@ -27,14 +27,13 @@ def register(self, router: RequestRouter) -> None:
...
         from marimo._save.cache import CacheContext
-        from marimo._save.loaders import BasePersistenceLoader
 
         ctx = get_context()
         saved = 0
         for obj in ctx.globals.values():
             if isinstance(obj, CacheContext):
-                if isinstance(obj.loader, BasePersistenceLoader):
-                    obj.loader.clear()
+                # Loader.clear is a no-op by default; safe on any subclass.
```
dmadisetti, line 35: I think we can remove the comment, this may have only been in here fore typing

dmadisetti (approved):

PR #9589, opened 2026-05-18
refactor(runtime): narrow callback deps, drop get_context in cache

mscolnick:
Introduce a GlobalsView protocol for CacheCallbacks and ExternalStorageCallbacks
so they no longer take the full Kernel; CacheCallbacks now reads globals via
the injected scope instead of get_context(). Move the env snapshot from
Kernel into SecretsCallbacks. Promote _if_autorun_then_run_cells to a public
maybe_autorun_cells on Kernel.


dmadisetti (approved):
Nice!

PR #9597, opened 2026-05-19
feat(table): expand column filter operators and pill-editor UX

kirangadhave:
> **Reviewer note**: please review **commit by commit**.

## Summary

Expands the table's column-filter capabilities and overhauls the filter-pill editor.

- **Filter models**: typed text/number filter models, shared operators module, native between for numbers, value-drop for unary text ops.
- **Column header**: full operator pickers for text and number columns. `FilterByValuesList` gains creatable mode and fixes the chosen-outside-top-K case.
- **Pill editor**: full per-dtype operator support, operator-aware label/value formatting, plus UX polish (Apply gated by form submit + tooltip, draft preserved across compatible operator changes, keyboard focus flow, values picker as popover, right-sized value slots).
- **Regex filter**: slash-bracketed input for the new `regex` text operator.
- **Bug fix**: column-header text filter no longer eats special characters via key-propagation.
- **Creatable values picker:** the `in` / `not_in` value picker now supports typing values that aren't in the top-K — they appear as a "+ Add" item and, once selected, render at the bottom of the picker with no count so they stay visible and removable.

### Column Header Filtering

https://github.com/user-attachments/assets/435052c5-3365-4b81-a8f0-1b6fcddbca46

### Values picker

https://github.com/user-attachments/assets/4f400f6e-7b91-48b1-9030-4036fc80c0bd

## Test plan

- [x] Filter a text column with every operator (contains, starts/ends with, equals, regex, is_empty, is_not_empty, ...)
- [x] Filter a number column with every operator (between, eq, lt, gt, ...)
- [x] Open the pill editor, switch operators: draft value preserved where compatible, cleared otherwise
- [x] Tab through column, operator, value, Apply
- [x] Apply button disabled with tooltip until value is valid; form submit also applies
- [x] FilterByValuesList: type to create a new value, confirm values outside top-K still display
- [x] Regex input accepts `/pattern/flags`
- [x] Special chars (`/`, `[`, `]`, etc.) typed into the column-header text filter reach the input

### DIFF
--- a/frontend/src/components/data-table/filter-by-values-picker.tsx
+++ b/frontend/src/components/data-table/filter-by-values-picker.tsx
@@ -60,6 +60,7 @@ export const FilterByValuesPicker = <TData, TValue>({
     <Popover open={open} onOpenChange={setOpen}>
       <PopoverTrigger asChild={true}>
         <Button
+          type="button"
           variant="outline"
           size="xs"
           className="h-6 mb-1 w-full justify-between font-normal"
--- a/frontend/src/components/data-table/filter-by-values-picker.tsx
+++ b/frontend/src/components/data-table/filter-by-values-picker.tsx
@@ -32,6 +32,7 @@ interface Props<TData, TValue> {
   calculateTopKRows?: CalculateTopKRows;
   chosenValues: unknown[];
   onChange: (values: unknown[]) => void;
+  creatable?: boolean;
 }
 
 export const FilterByValuesPicker = <TData, TValue>({
@@ -39,6 +40,7 @@ export const FilterByValuesPicker = <TData, TValue>({
   calculateTopKRows,
   chosenValues,
   onChange,
+  creatable = false,
 }: Props<TData, TValue>) => {
   const [open, setOpen] = useState(false);
 
@@ -79,6 +81,7 @@ export const FilterByValuesPicker = <TData, TValue>({
           calculateTopKRows={calculateTopKRows}
           chosenValues={chosenValuesSet}
           onChange={onChange}
+          creatable={creatable}
         />
       </PopoverContent>
     </Popover>
--- a/frontend/src/components/data-table/filter-by-values-picker.tsx
+++ b/frontend/src/components/data-table/filter-by-values-picker.tsx
@@ -229,7 +229,8 @@ export const FilterByValuesList = <TData, TValue>({
             : `Search among the top ${data.length} values`
         }
         autoFocus={true}
-        onValueChange={(value) => setQuery(value.trim())}
+        value={query}
+        onValueChange={setQuery}
         onKeyDown={(e) => {
           if (e.key === "Enter" && canCreate) {
             e.preventDefault();
--- a/frontend/src/components/data-table/filter-by-values-picker.tsx
+++ b/frontend/src/components/data-table/filter-by-values-picker.tsx
@@ -90,6 +90,7 @@ interface FilterByValuesListProps<TData, TValue> {
   calculateTopKRows?: CalculateTopKRows;
   chosenValues: Set<unknown>;
   onChange: (values: unknown[]) => void;
+  creatable?: boolean;
 }
 
 /**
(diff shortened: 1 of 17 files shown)


kirangadhave:
@cubic-dev-ai

frontend/src/components/data-table/filter-by-values-picker.tsx
```diff
@@ -133,13 +138,48 @@ export const FilterByValuesList = <TData, TValue>({
...
+    creatable &&
+    trimmedQuery !== "" &&
+    !mergedData.some(([v]) => String(v) === trimmedQuery);
+
+  const commitCreate = () => {
+    if (!canCreate) {
+      return;
+    }
+    onChange([...chosenValues, trimmedQuery]);
+    setQuery("");
```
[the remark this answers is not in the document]
kirangadhave: addressed

frontend/src/components/data-table/regex-input.tsx
```diff
@@ -0,0 +1,61 @@
...
+        className,
+      )}
+    >
+      <Slash />
+      <Input
+        ref={ref}
+        id={id}
+        type="text"
+        value={value}
+        onChange={(e) => onChange(e.target.value)}
```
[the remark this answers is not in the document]
kirangadhave: The `/` characters are decorative to show it's a regex input. Only the actual pattern string is sent to backend

frontend/src/components/data-table/filter-pill-editor.tsx
```diff
@@ -302,53 +334,107 @@ interface ValueSlotProps<TData, TValue> {
...
+      <div className="w-48">
+        <FilterByValuesPicker
+          column={column}
+          calculateTopKRows={calculateTopKRows}
+          chosenValues={v.values ?? []}
+          onChange={(next) =>
+            onChange({ kind: "multi-text", values: next.map(String) })
+          }
+          creatable={true}
+        />
```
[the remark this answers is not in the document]
kirangadhave: makes sense

Light2Dark:
This looks good, nice!

frontend/src/components/data-table/filter-pills.tsx
```diff
@@ -209,19 +210,62 @@ function formatValue(
...
+      case "ends_with":
+        return {
+          operator: OPERATOR_LABELS[value.operator].toLowerCase(),
+          value: `"${value.text}"`,
+        };
+    }
   }
   if (value.type === "date") {
-    return formatMinMax(value.min?.toISOString(), value.max?.toISOString());
+    return formatMinMaxLegacy(
```
Light2Dark, line 253: why is this called legacy?

frontend/src/components/data-table/filter-pill-editor.tsx
```diff
@@ -191,9 +194,29 @@ export const FilterPillEditor = <TData,>({
...
+  useEffect(() => {
+    const firstInput = valueSlotRef.current?.querySelector<HTMLElement>(
+      'input, [role="combobox"], button',
+    );
+    if (firstInput) {
+      firstInput.focus();
+    } else {
+      operatorTriggerRef.current?.focus();
+    }
+  }, [draftType, draftOperator]);
```
Light2Dark, line 210: there's also a FocusScope from "react-aria" component, which sounds like what this is doing. But non-blocking

Light2Dark (approved):

frontend/src/components/data-table/filter-pills.tsx
kirangadhave: Previously we overloaded the between operator to support all comparison operators. The UI now enforces separate operators for everything, but we still support previous behaivor (for e.g. backend can send `min: null && max: 7` which becomes `val <= 7`).

The formatter handles the legacy behavior.

frontend/src/components/data-table/filter-pill-editor.tsx
kirangadhave: FocusScope doesn't actually replicate this completely. it focuses on first focusable element on mount (we want to focus on first input OR operator). Also the useEffect runs on operator change, and adjusts focus accordingly, which FocusScope won't do.

PR #9600, opened 2026-05-19
feat(ui): support disabled on dropdown and multiselect

### LINKED ISSUE #9579: mo.ui.dropdown should support being disabled
### Description

Most controls (text area, button, slider, etc) support being disabled with `disabled=True`, but `mo.ui.dropdown` does not.

- [x] Yes

### Alternatives

_No response_

### Additional context

_No response_

kirangadhave:
## Summary

- Adds a `disabled` kwarg to `mo.ui.dropdown` and `mo.ui.multiselect`, matching the pattern already used by other inputs (`text`, `slider`, `checkbox`, `radio`, `button`, etc.).
- Native `<select>` path uses the native `disabled` attribute.
- Searchable dropdown and multiselect (Combobox) block popover toggling and chip removal via `aria-disabled` and pointer-events styling; the trigger remains a `<div>` to avoid changing focus/keyboard semantics for existing combobox usages.

Closes #9579

https://github.com/user-attachments/assets/4ee53d9b-539c-40a9-8fbd-a98c322bcc46

## Test plan

- [x] `make check`
- [x] Python: new `test_dropdown_disabled`, `test_multiselect_disabled`
- [x] Frontend: new tests for disabled native select and disabled searchable combobox; existing tests updated
- [x] Manual: smoke notebook with enabled + disabled variants of dropdown, searchable dropdown, and multiselect

### DIFF
--- a/frontend/src/components/ui/combobox.tsx
+++ b/frontend/src/components/ui/combobox.tsx
@@ -173,7 +173,7 @@ export const Combobox = <TValue,>({
       <Popover
         open={open}
         onOpenChange={(v) => {
-          if (disabled) {
+          if (disabled && v) {
             return;
           }
           setOpen(v);
--- a/frontend/src/components/ui/combobox.tsx
+++ b/frontend/src/components/ui/combobox.tsx
@@ -39,6 +39,7 @@ interface ComboboxCommonProps<TValue> {
   className?: string;
   id?: string;
   keepPopoverOpenOnSelect?: boolean;
+  disabled?: boolean;
 }
 
 type ComboboxFilterProps =
@@ -95,6 +96,7 @@ export const Combobox = <TValue,>({
   chipsClassName,
   keepPopoverOpenOnSelect,
   id,
+  disabled = false,
   ...rest
 }: ComboboxProps<TValue>) => {
   const [open = false, setOpen] = useControllableState({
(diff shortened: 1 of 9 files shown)


mscolnick (dismissed):

frontend/src/components/ui/combobox.tsx
```diff
@@ -168,15 +170,25 @@ export const Combobox = <TValue,>({
...
+      <Popover
+        open={open}
+        onOpenChange={(v) => {
+          if (disabled) {
+            return;
+          }
+          setOpen(v);
+        }}
+      >
         <PopoverTrigger asChild={true}>
```
Light2Dark, line 182: should this trigger have disabled prop?

frontend/src/components/ui/combobox.tsx
Light2Dark, line 199: should we add aria-disabled on the popoverContent?

frontend/src/components/ui/combobox.tsx
kirangadhave: `PopoverTrigger` has a `disabled` prop which just gets forwarded the the child because of `asChild={true}`. The child here is a `div` which can't do anything with the `disabled` prop.

We could switch to using `button` instead of `div`, but that would need testing all previous usages.

frontend/src/components/ui/combobox.tsx
kirangadhave: i don't think it's needed. `onOpenChange` is gated, so PopoverContent is never actually mounted when disabled is true. We have it on trigger which actually gets read by assistive tech.

frontend/src/components/ui/combobox.tsx
```diff
@@ -168,15 +170,25 @@ export const Combobox = <TValue,>({
...
   return (
     <div className={cn("relative")} {...rest}>
-      <Popover open={open} onOpenChange={setOpen}>
+      <Popover
+        open={open}
+        onOpenChange={(v) => {
+          if (disabled) {
+            return;
+          }
+          setOpen(v);
```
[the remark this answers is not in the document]
kirangadhave: makes sense

frontend/src/components/ui/combobox.tsx
Light2Dark: oh nice, missed that. Then, its okay for me

Light2Dark (approved):

PR #9606, opened 2026-05-19
fix(files): show .git and .venv in file browser

kirangadhave:
## Summary

- list_files unconditionally stripped `.git` and `.venv`, so the file tree's "Show hidden files" toggle could never reveal them. Split the ignore list: listing only filters `.`, `..`, `.DS_Store`; recursive search still skips `.git`, `.venv`, `node_modules`, `__pycache__`, and `site-packages` to avoid traversing pack files and site-packages.
- File-explorer toggle button now swaps to an open-eye icon in the accent color when hidden files are visible.

https://github.com/user-attachments/assets/6026a7d4-1584-4989-9ed6-66e8b7b6b40a

## Test plan

- [x] `uv run --group test pytest tests/_server/files/test_os_file_system.py`
- [x] `make py-check`
- [x] `make fe-check`
- [x] Manual: dotfiles hidden by default, visible after toggle; file search does not return entries from `.git`/`.venv`.

### DIFF
--- a/marimo/_server/files/os_file_system.py
+++ b/marimo/_server/files/os_file_system.py
@@ -21,10 +21,14 @@ from marimo._utils.files import natural_sort
 
 LOGGER = _loggers.marimo_logger()
 
-IGNORE_LIST = [
+LIST_IGNORE_LIST = [
     ".",
     "..",
     ".DS_Store",
+]
+
+SEARCH_IGNORE_LIST = [
+    *LIST_IGNORE_LIST,
     "__pycache__",
     "node_modules",
     ".git",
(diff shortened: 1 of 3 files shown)


marimo/_server/files/os_file_system.py
```diff
@@ -21,10 +21,14 @@
 
 LOGGER = _loggers.marimo_logger()
 
-IGNORE_LIST = [
+LIST_IGNORE_LIST = [
```
Light2Dark, line 24: maybe needs a better name

kirangadhave: any suggestions? i was going for clarity and consistency

Light2Dark (approved):

marimo/_server/files/os_file_system.py
Light2Dark: maybe base_ignore_list? i think its fine though nevermind

PR #9379, opened 2026-04-24
fix: def declaration within function

Shamik-07:
## 📝 Summary

Previously, F12/command usage would jump effectively using first match semantics and could jump to an outer/global definition when the cursor was on a local symbol use. Now, the jump follows lexical scope first,  and fixes the issue where symbols inside def blocks did not resolve correctly.

Tested with the code snippets to reproduce this issue and added the same to the tests. 
Closes #1430 

## 📋 Pre-Review Checklist

- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it.
- [ ] Video or media evidence is provided for any visual changes (optional). 

## ✅ Merge Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [x] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Tests have been added for the changes made.

I have read the CLA Document and I hereby sign the CLA

PR #9607, opened 2026-05-19
Tighten import handling and fix SetComprehension typo

manzt:
Follow up to #9379

`ImportStatement` was collecting every `VariableName` direct child as a binding, including the module path in `from m import ...` and the imported name when shadowed by `as`.

Also:

- Removed the dead `ImportFromStatement` branch — the Lezer Python grammar only emits `ImportStatement`, so it never matched.
- Fixed pre-existing `SetComprehension` typo in reactive-references/analyzer.ts (grammar node is `SetComprehensionExpression`); set comprehensions now correctly create a scope.

### DIFF
--- a/frontend/src/core/codemirror/go-to-definition/__tests__/commands.test.ts
+++ b/frontend/src/core/codemirror/go-to-definition/__tests__/commands.test.ts
@@ -253,6 +253,73 @@ a = 10`;
     `);
   });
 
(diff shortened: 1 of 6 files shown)


mscolnick:
@cubic-dev-ai

mscolnick (approved):

PR #9610, opened 2026-05-19
refactor(frontend): clean up runtime health probe and transport close handling

kirangadhave:
Closes MO-6153

## Summary

Two small follow-ups to the disconnect-indicator fix (#9410):

1. **Runtime health probe** is split into a pure `probeHealth()` (just fetches /health) and a side-effectful `reconcileFromHealth()` (rewrites `config.url` on redirect, sets `<base href>`). The mid-session footer poll uses the pure probe so a redirected /health no longer silently mutates `<base href>` after the page is loaded. Bootstrap (`init`) and manual reconnect use reconcile (the latter so a runtime that has moved is followed).

2. **Transport close handling** no longer leaks partysocket's `retryCount` through the public transport interface. A new `WsTransport` wraps `ReconnectingWebSocket` and rewrites the close-event reason to `MARIMO_TRANSPORT_EXHAUSTED` when partysocket gives up. `classifyCloseEvent` becomes a pure function of the reason; the exhaustion case lives alongside the other `MARIMO_*` reasons. Adds a `CloseReason` literal union so `logNever` in the default branch is type-checked.

Follow-up tracked separately: centralize backend/frontend close-reason strings.

### DIFF
--- a/frontend/src/core/websocket/transports/partysocket.ts
+++ b/frontend/src/core/websocket/transports/partysocket.ts
@@ -1,0 +1,90 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+
+import ReconnectingWebSocket from "partysocket/ws";
+import type {
+  ConnectionEvent,
+  ConnectionTransportCallback,
+  IConnectionTransport,
+} from "./transport";
+
+// Per-`reconnect()` retry budget. After exhaustion, partysocket stops silently;
+// the wrapper rewrites the close-event reason to surface the give-up.
+export const MAX_RETRIES = 10;
+
+export const TRANSPORT_EXHAUSTED_REASON = "MARIMO_TRANSPORT_EXHAUSTED";
+
+export class PartysocketTransport implements IConnectionTransport {
+  private inner: ReconnectingWebSocket;
+  private closeWrappers = new WeakMap<
+    ConnectionTransportCallback<"close">,
+    ConnectionTransportCallback<"close">
+  >();
+
+  constructor(urlProvider: () => string) {
+    this.inner = new ReconnectingWebSocket(urlProvider, undefined, {
+      maxRetries: MAX_RETRIES,
+      debug: false,
+      startClosed: true,
+      // long timeout — the server can become slow when many notebooks are open.
+      connectionTimeout: 10_000,
+    });
+  }
+
+  get readyState(): WebSocket["readyState"] {
+    return this.inner.readyState as WebSocket["readyState"];
+  }
+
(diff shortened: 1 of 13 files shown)


frontend/src/core/websocket/useMarimoKernelConnection.tsx
```diff
@@ -80,16 +77,24 @@ import {
 
 const SUPPORTS_LAZY_KERNELS = true;
 
+export type CloseReason =
```
mscolnick, line 84: maybe a comment these are derived from the backend (or better if we can make it a shared openapi enum)

frontend/src/core/websocket/transports/partysocket.ts
```diff
@@ -0,0 +1,90 @@
...
+  IConnectionTransport,
+} from "./transport";
+
+// Per-`reconnect()` retry budget. After exhaustion, partysocket stops silently;
+// the wrapper rewrites the close-event reason to surface the give-up.
+export const MAX_RETRIES = 10;
+
+export const TRANSPORT_EXHAUSTED_REASON = "MARIMO_TRANSPORT_EXHAUSTED";
+
+export class PartysocketTransport implements IConnectionTransport {
```
mscolnick: what about `WsTranport`?

kirangadhave:
@cubic-dev-ai

frontend/src/core/websocket/transports/ws.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  addEventListener<T extends ConnectionEvent>(
+    event: T,
+    callback: ConnectionTransportCallback<T>,
+  ): void {
+    if (event === "close") {
+      const userCb = callback as ConnectionTransportCallback<"close">;
+      const wrapper: ConnectionTransportCallback<"close"> = (e) => {
+        if (this.inner.retryCount >= MAX_RETRIES) {
+          userCb(
```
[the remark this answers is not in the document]
kirangadhave: adding tests in next commit

frontend/src/core/websocket/useMarimoKernelConnection.tsx
kirangadhave: added comment + issue to track the shared openapi enum (it's a bigger refactor)

frontend/src/core/websocket/transports/partysocket.ts
kirangadhave: So rename Partysocket to WsTransport? makes sense

mscolnick (approved):

PR #9621, opened 2026-05-20
support tool approvals for pydantic-ai chatbot

Light2Dark:
## 📝 Summary

Supports more AI sdk parts, refactors logic to be more maintainable.

- ChatMessage no longer loses unmodeled SDK fields. Approvals (and e.g. `callProviderMetadata`, `providerExecuted`, `preliminary`) live on AI SDK parts that marimo's typed dataclasses don't model. The message now snapshots the raw wire payload per-part in _raw_parts, so those fields survive every round-trip
- `pydantic_ai._build_ui_messages` uses the raw payload. It now calls `message.raw_or_dumped_parts()` so the approval/tool state the frontend just sent us makes it into the agent run unmodified. The old `asdict` + `_remove_none_values` path was lossy.
- `sanitize_part` strips keys the AI SDK's { ...part, state, ... } spread can leak from prior tool states (e.g. a stale output clinging to an approval-requested part).
- hasPendingToolCalls (frontend) rewritten. The old predicate treated "every tool ready & no trailing text" silently looped whenever an assistant message ended in a non-text part (file, source-url, data-*, reasoning) after a completed tool call. The fix uses some native AI sdk logic and some testing.

image

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/marimo/_ai/llm/_impl.py
+++ b/marimo/_ai/llm/_impl.py
@@ -838,11 +838,17 @@ class pydantic_ai(ChatModel):
             messages=ui_messages,
         )
 
-        adapter = VercelAIAdapter(
-            agent=self.agent,
-            run_input=run_input,
-            sdk_version=AI_SDK_VERSION,
-        )
+        try:
+            adapter = VercelAIAdapter(
+                agent=self.agent,
+                run_input=run_input,
+                sdk_version=AI_SDK_VERSION,
+            )
+        except TypeError:
+            adapter = VercelAIAdapter(
+                agent=self.agent,
+                run_input=run_input,
+            )
         event_stream = adapter.run_stream(model_settings=model_settings)
         async for event in event_stream:
             if serialized := self._serialize_vercel_ai_chunk(event):
--- a/marimo/_ai/llm/_impl.py
+++ b/marimo/_ai/llm/_impl.py
@@ -1,16 +1,14 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-import dataclasses
 import json
 import os
 import re
 from typing import TYPE_CHECKING, Any, cast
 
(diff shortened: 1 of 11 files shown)


marimo/_ai/llm/_impl.py
```diff
@@ -747,18 +745,14 @@ def _build_ui_messages(
...
                     list[UIMessagePart],
-                    [
-                        self._remove_none_values(
-                            dataclasses.asdict(part)
-                            if dataclasses.is_dataclass(part)
-                            else part
-                        )
-                        for part in message.parts
-                    ],
+                    [sanitize_part(p) for p in message.raw_or_dumped_parts()],
```
kirangadhave, line 755: we don't want to remove `None` values anymore?

kirangadhave, line 755: Codex review highlighted this:

>`sanitize_part()` builds its allowlist from pydantic field aliases only, so typed marimo dataclass dumps like `ToolInvocationPart(...)` lose snake_case keys such as `tool_call_id` before pydantic-ai can validate them. For example, an `output-available` tool part is filtered down without either `toolCallId` or `tool_call_id`, causing `UIMessage(...)` validation to fail for chat histories containing typed marimo tool parts. This path should either preserve field names in the allowlist or convert dataclass dumps to the frontend alias form before sanitizing.

I'm not sure about implementations, so leave this to your discretion.

Light2Dark: With the new sanitize_part function, we only filter fields that pydantic-ai accepts, so None will get removed naturally. Thanks though! I did not write the description well.

Light2Dark: I don't think users can hit this issue, but will monitor and try out

PR #9629, opened 2026-05-20
perf(autoreload): skip stdlib/site-packages on per-cell check

### LINKED ISSUE #9628: Module autoreload leads to a significant performance drop
### Describe the bug

Switching the "on module change" option from "off" to either "lazy" or "autorun" causes every single cell to take about 16ms-80ms to run depending on the machine (instead of the usual 1ms-2ms). As a one time penalty this is insignificant, but when a single UI interaction causes a dozen cells to rerun it can lead to over a a second of delay, which is rather significant.

I am assuming the cause is marimo checking for module changes on every single cell run. If that's the case one possible fix would be to have it only check once per execution run of cells, and another possible fix would be to have it check via a watchdog in the background so that it's independent of cell execution.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.5",
  "editable": false,
  "location": "/var/home/umnikos/.local/lib/python3.14/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.17.7-ba29.fc43.x86_64",
  "Processor": "",
  "Python Version": "3.14.3",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v25.9.0",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.1.7",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.21.0",
    "packaging": "25.0",
    "psutil": "7.0.0",
    "pygments": "2.20.0",
    "pyyaml": "6.0.2",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.46.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "polars": "1.40.1",
    "python-lsp-ruff": "2.3.1",
    "ruff": "0.15.12"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

It happens even on empty notebooks so here's my `marimo.toml` file instead:

<details>

```
[completion]
signature_hint_on_typing = false
activate_on_typing = true
copilot = false

[display]
cell_output = "below"
code_editor_font_size = 14
dataframes = "rich"
default_width = "medium"
default_table_page_size = 10
reference_highlighting = true
theme = "light"
default_table_max_columns = 50

[server]
follow_symlink = false
browser = "default"

[language_servers.pylsp]
enable_flake8 = false
enable_pyflakes = false
enable_mypy = true
enable_pydocstyle = false
enable_pylint = false
enable_ruff = true
enabled = false

[diagnostics]
sql_linter = true

[formatting]
line_length = 79

[ai.custom_providers]

[ai.models]
displayed_models = []
custom_models = []

[mcp]
presets = ["marimo"]

[mcp.mcpServers]

[save]
format_on_save = false
autosave = "after_delay"
autosave_delay = 1000

[keymap]
preset = "default"

[keymap.overrides]

[runtime]
std_stream_max_bytes = 1000000
auto_reload = "lazy"
output_max_bytes = 8000000
auto_instantiate = false
watcher_on_save = "lazy"
show_tracebacks = false
reactive_tests = true
on_cell_change = "autorun"
default_sql_output = "auto"
default_csv_encoding = "utf-8"

[package_management]
manager = "pip"

[snippets]
include_default_snippets = true
custom_paths = []

```

</details>

mscolnick:

Fixes #9628.

With `auto_reload` set to `lazy` or `autorun`, every cell run was calling `ModuleReloader.check(sys.modules, reload=True)`, which iterates *all* of `sys.modules` and does `os.stat` on each entry. With ~1000 modules in scope (typical), that adds 16–80ms per cell — compounded across the dozen cells re-running on a UI interaction it becomes a >1s lag.

This change adds an opt-in `skip_non_user_modules=True` flag on `ModuleReloader.check`. When set, stdlib and site-packages module names are recorded in a persistent skip set (classified by `sysconfig` prefixes) and short-circuited on subsequent calls.

`AutoreloadManager.cell_scope` (the hot per-cell path) opts in. The background `ModuleWatcher` keeps the default behavior and continues to scan every module on its 1s loop, so edits inside an installed package are still detected — just at watcher latency rather than cell-entry latency. Editable installs (`pip install -e .`, `uv add --editable`) have `__file__` outside site-packages, so they are correctly classified as user code and reload with no latency change.

### Benchmark

Driving `ModuleReloader.check()` directly, 200 iterations post-warmup. Issue-shaped workload: ~2.5k modules (heavy stdlib + numpy/pandas/etc.) + 5 user files in a tmp dir.

| path | median | p95 |
|------|-------:|----:|
| before | 4.88 ms | 6.15 ms |
| after  | 0.91 ms | 1.01 ms |

**~4 ms saved per cell run, 5.4× median speedup.**

Scale curve (median µs, varying user-module count):

| user mods | sys.modules | before | after | speedup |
|----------:|------------:|-------:|------:|--------:|
| 0   | 2514 |  5037 |  873 | 5.8× |
| 5   | 2519 |  5245 |  802 | 6.5× |
| 25  | 2539 |  6082 | 1693 | 3.6× |
| 100 | 2614 |  8342 | 4421 | 1.9× |
| 500 | 3014 | 12489 | 8398 | 1.5× |

The win narrows as user-code grows, by design: the optimization only filters out non-user-code.

### DIFF
--- a/marimo/_runtime/reload/autoreload.py
+++ b/marimo/_runtime/reload/autoreload.py
@@ -75,16 +75,28 @@ def safe_hasattr(obj: M, attr: str) -> bool:
 
 
 def _non_user_module_roots() -> tuple[str, ...]:
-    """Filesystem prefixes that hold stdlib + site-packages modules."""
+    """Filesystem prefixes that hold stdlib + site-packages modules.
+
+    Each entry is normalized and terminated with a separator so that a raw
+    prefix check on a normalized path matches whole directory boundaries
+    (e.g. `/usr/lib/python3.13/` does not match `/usr/lib/python3.13-mine/`).
+    """
     roots: set[str] = set()
     for key in ("stdlib", "platstdlib", "purelib", "platlib"):
         p = sysconfig.get_path(key)
         if p:
             roots.add(p)
-    # Covers stdlib for non-venv interpreters, where sysconfig's stdlib path
-    # may not match the runtime location.
-    roots.add(sys.base_prefix)
-    return tuple(roots)
+    # Fallback for builds where sysconfig's stdlib path is missing or
+    # differs from the runtime location of the stdlib.
+    roots.add(os.path.dirname(os.__file__))
+
+    normalized: set[str] = set()
+    for r in roots:
+        n = os.path.normcase(os.path.realpath(r))
+        if not n.endswith(os.sep):
+            n += os.sep
+        normalized.add(n)
+    return tuple(normalized)
 
 
 def modules_imported_by_cell(
@@ -174,9 +186,11 @@ class ModuleReloader:
         # for thread-safety
         self.lock = threading.Lock()
         self._module_dependency_finder = ModuleDependencyFinder()
-        # Names known to live in stdlib/site-packages; populated lazily on the
-        # hot per-cell path. Entries are never evicted: a module whose
-        # __file__ moves between roots at runtime would not be re-evaluated.
+        # Names known to live in stdlib/site-packages. Populated lazily by
+        # callers that pass `skip_non_user_modules=True` (the hot per-cell
+        # path), and then consulted by every `check()` call. Entries are
+        # never evicted: a module whose `__file__` moves between roots at
+        # runtime would not be re-evaluated.
         self._skip: set[str] = set()
         self._non_user_roots = _non_user_module_roots()
 
@@ -186,13 +200,14 @@ class ModuleReloader:
     def _is_user_module(self, module: types.ModuleType) -> bool:
         """True for modules whose source lives outside stdlib/site-packages.
 
-        Editable installs (e.g. ``pip install -e .``) point ``__file__`` at the
+        Editable installs (e.g. `pip install -e .`) point `__file__` at the
         source tree, so they are correctly classified as user code.
         """
         f = safe_getattr(module, "__file__", None)
         if not f:
             return False
-        return not f.startswith(self._non_user_roots)
+        path = os.path.normcase(os.path.realpath(f))
+        return not path.startswith(self._non_user_roots)
 
     def filename_and_mtime(
         self, module: types.ModuleType
@@ -246,10 +261,12 @@ class ModuleReloader:
 
         Also patches existing objects with hot-reloaded ones.
 
-        When ``skip_non_user_modules`` is True, stdlib and site-packages
-        modules are skipped via a persistent cache — intended for the per-cell
-        hot path. The background ``ModuleWatcher`` keeps the default so that
-        edits to installed packages remain detectable at watcher latency.
+        When `skip_non_user_modules` is True, modules whose `__file__` is
+        under stdlib/site-packages are added to a persistent skip set and
+        won't be stat-ed on this or future calls. Intended for the per-cell
+        hot path. Once populated, the skip set short-circuits every caller
+        — including the background `ModuleWatcher` — so edits inside
+        installed packages are not hot-reloaded.
 
         Returns a set of modules that were found to have been modified.
         """
@@ -261,17 +278,16 @@ class ModuleReloader:
         # and also iterates over it
         with self.lock:
             modified_modules: set[types.ModuleType] = set()
-            skip = self._skip if skip_non_user_modules else None
             # materialize the module keys, since we'll be reloading while
             # iterating
             for modname in list(modules.keys()):
-                if skip is not None and modname in skip:
+                if modname in self._skip:
                     continue
                 m = modules.get(modname, None)
                 if m is None:
                     continue
-                if skip is not None and not self._is_user_module(m):
-                    skip.add(modname)
+                if skip_non_user_modules and not self._is_user_module(m):
+                    self._skip.add(modname)
                     continue
 
                 module_mtime = self.filename_and_mtime(m)
--- a/marimo/_runtime/reload/autoreload.py
+++ b/marimo/_runtime/reload/autoreload.py
@@ -297,7 +297,12 @@ class ModuleReloader:
                     if self._skip[modname] == current_file:
                         is_non_user = True
                     else:
+                        # Rebound to a different file — drop all cached
+                        # state for this name so the new module starts
+                        # from a clean mtime baseline.
                         del self._skip[modname]
+                        self.modules_mtimes.pop(modname, None)
+                        self.stale_modules.discard(modname)
                         is_non_user = False
                 else:
                     is_non_user = False
--- a/marimo/_runtime/reload/autoreload.py
+++ b/marimo/_runtime/reload/autoreload.py
@@ -188,12 +188,13 @@ class ModuleReloader:
         # for thread-safety
         self.lock = threading.Lock()
         self._module_dependency_finder = ModuleDependencyFinder()
-        # Names known to live in stdlib/site-packages. Populated by every
-        # `check()` call (memoizing the `_is_user_module` classification);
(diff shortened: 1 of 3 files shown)


marimo/_runtime/reload/autoreload.py
```diff
@@ -206,12 +236,21 @@ def cell_uses_stale_modules(self, cell: CellImpl) -> bool:
...
+        modules: dict[str, types.ModuleType],
+        reload: bool,
+        *,
+        skip_non_user_modules: bool = False,
     ) -> set[types.ModuleType]:
         """Check timestamps of modules, optionally reload them.
 
         Also patches existing objects with hot-reloaded ones.
 
+        When ``skip_non_user_modules`` is True, stdlib and site-packages
```
akshayka:
```suggestion
        When `skip_non_user_modules` is True, stdlib and site-packages
```

marimo/_runtime/reload/autoreload.py
```diff
@@ -206,12 +236,21 @@ def cell_uses_stale_modules(self, cell: CellImpl) -> bool:
...
+        modules are skipped via a persistent cache — intended for the per-cell
+        hot path. The background ``ModuleWatcher`` keeps the default so that
```
akshayka:
```suggestion
        hot path. The background `ModuleWatcher` keeps the default so that
```

akshayka: What does "keeps the default" mean?

marimo/_runtime/reload/autoreload.py
```diff
@@ -73,6 +74,19 @@ def safe_hasattr(obj: M, attr: str) -> bool:
         return False
 
 
+def _non_user_module_roots() -> tuple[str, ...]:
+    """Filesystem prefixes that hold stdlib + site-packages modules."""
+    roots: set[str] = set()
+    for key in ("stdlib", "platstdlib", "purelib", "platlib"):
+        p = sysconfig.get_path(key)
```
akshayka, line 88: TIL

marimo/_runtime/reload/autoreload.py
```diff
@@ -160,10 +174,26 @@ def __init__(self) -> None:
...
+        self._skip: set[str] = set()
+        self._non_user_roots = _non_user_module_roots()
 
         # Timestamp existing modules
         self.check(modules=sys.modules, reload=False)
 
+    def _is_user_module(self, module: types.ModuleType) -> bool:
+        """True for modules whose source lives outside stdlib/site-packages.
+
+        Editable installs (e.g. ``pip install -e .``) point ``__file__`` at the
```
akshayka:
```suggestion
        Editable installs (e.g. `pip install -e .`) point `__file__` at the
```

The convention in our codebase is to use single backticks in docstrings. Double backticks are an artifact of RST, which we don't use. We should update our agents.md

marimo/_runtime/reload/autoreload.py
```diff
@@ -222,12 +261,18 @@ def check(
         # and also iterates over it
         with self.lock:
             modified_modules: set[types.ModuleType] = set()
+            skip = self._skip if skip_non_user_modules else None
```
akshayka: This can be deleted

marimo/_runtime/reload/autoreload.py
```diff
@@ -222,12 +261,18 @@ def check(
...
             # materialize the module keys, since we'll be reloading while
             # iterating
             for modname in list(modules.keys()):
+                if skip is not None and modname in skip:
```
akshayka:
```suggestion
                if modname in self._skip:
```

marimo/_runtime/reload/autoreload.py
```diff
@@ -222,12 +261,18 @@ def check(
...
+                    continue
                 m = modules.get(modname, None)
                 if m is None:
                     continue
+                if skip is not None and not self._is_user_module(m):
```
akshayka:
```suggestion
                if skip_non_user_modules and self._is_user_module(m):
```

marimo/_runtime/reload/autoreload.py
```diff
@@ -222,12 +261,18 @@ def check(
...
+                    skip.add(modname)
```
akshayka:
```suggestion
                    self._skip.add(modname)
```

akshayka (commented):
Overall LGTM, code style comments

marimo/_runtime/reload/autoreload.py
mscolnick: updated to be more descriptive

marimo/_runtime/reload/autoreload.py
```diff
@@ -73,6 +74,31 @@ def safe_hasattr(obj: M, attr: str) -> bool:
         return False
 
 
+def _non_user_module_roots() -> tuple[str, ...]:
```
dmadisetti, line 79: functools.cache?

marimo/_runtime/reload/autoreload.py
```diff
@@ -160,10 +186,29 @@ def __init__(self) -> None:
         # for thread-safety
         self.lock = threading.Lock()
         self._module_dependency_finder = ModuleDependencyFinder()
+        # Names known to live in stdlib/site-packages. Populated lazily by
+        # callers that pass `skip_non_user_modules=True` (the hot per-cell
+        # path), and then consulted by every `check()` call. Entries are
+        # never evicted: a module whose `__file__` moves between roots at
+        # runtime would not be re-evaluated.
+        self._skip: set[str] = set()
+        self._non_user_roots = _non_user_module_roots()
```
dmadisetti: if functools, we don't have to set

marimo/_runtime/reload/autoreload.py
```diff
@@ -160,10 +186,29 @@ def __init__(self) -> None:
...
         self.check(modules=sys.modules, reload=False)
 
+    def _is_user_module(self, module: types.ModuleType) -> bool:
+        """True for modules whose source lives outside stdlib/site-packages.
+
+        Editable installs (e.g. `pip install -e .`) point `__file__` at the
+        source tree, so they are correctly classified as user code.
+        """
+        f = safe_getattr(module, "__file__", None)
+        if not f:
```
dmadisetti, line 209: false positive on c libraries? Unsure, but I think so. Maybe that's fine

akshayka (dismissed):

marimo/_runtime/reload/autoreload.py
```diff
@@ -225,9 +281,14 @@ def check(
             # materialize the module keys, since we'll be reloading while
             # iterating
             for modname in list(modules.keys()):
+                if modname in self._skip:
+                    continue
                 m = modules.get(modname, None)
                 if m is None:
                     continue
+                if skip_non_user_modules and not self._is_user_module(m):
```
dmadisetti: could always add, that way when the skip path comes along it can jump early. Gating should be just on the continue

dmadisetti (commented):
I think the continue logic should be tied to `skip_non_user_modules`

mscolnick:
thanks @dmadisetti, i had that but removed from the comments. will add back `skip_non_user_modules` to just the hot path

marimo/_runtime/reload/autoreload.py
mscolnick: nice, done

PR #9630, opened 2026-05-20
feat(file_browser): restore selection_mode='all' and accept list form

### LINKED ISSUE #9402: file_browser() support for selecting either/both files and directories from one instance
### Description

My use case is a browser for parquet, json and deltalake files.

"file" works fire for the first two, but deltalake are actually directories, so I can't have a single ergonomic UI unless I can select delta lake directories in the same `file_browser` as parquet and json files...

### Suggested solution

An "both" option to file_browser's `selection_mode` parameter.

- [ ] Yes

### Alternatives

There's aren't any nice options, I just bailed and forced the user to put full path in a text box as the hack around 😢 

### Additional context

_No response_

kirangadhave:
## Summary

- Restore the `selection_mode="all"` string on `mo.ui.file_browser` (regressed in #4130).
- Also accept a list/tuple form, e.g. `selection_mode=["file", "directory"]`.
- Internally normalized to a `frozenset` of selectable kinds; the wire format stays a single string (`"file" | "directory" | "all"`), so the frontend (which already handles `"all"`) is unchanged.
- Default remains `"file"`; no behavior change for existing callers.

Closes #9402

### DIFF
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -14,12 +14,55 @@ from typing import (
 from marimo import _loggers
 from marimo._output.rich_help import mddoc
 from marimo._plugins.ui._core.ui_element import UIElement
-from marimo._plugins.validators import validate_one_of
 from marimo._runtime.functions import Function
 from marimo._utils.files import natural_sort
 from marimo._utils.paths import is_cloudpath, normalize_path
 
 LOGGER = _loggers.marimo_logger()
+
+
+_VALID_KINDS: Final[frozenset[str]] = frozenset({"file", "directory"})
+
+
+def _normalize_selection_mode(
+    value: object,
+) -> frozenset[str]:
+    """Normalize ``selection_mode`` to a frozenset of selectable kinds.
+
+    Accepted inputs:
+        - ``"file"``      -> ``{"file"}``
+        - ``"directory"`` -> ``{"directory"}``
+        - ``"all"``       -> ``{"file", "directory"}``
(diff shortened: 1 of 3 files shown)


kirangadhave:
@cubic-dev-ai

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -114,8 +167,10 @@ class file_browser(
         filetypes (Sequence[str], optional): The file types to display in each
             directory; for example, filetypes=[".txt", ".csv"]. If None, all
             files are displayed. Defaults to None.
-        selection_mode (Literal["file", "directory"], optional): Either "file" or "directory". Defaults to
-            "file".
+        selection_mode (str | Sequence[str], optional): Which kinds of entries
+            the user can select. Accepts one of "file" (default), "directory",
+            "all", or a list/tuple containing "file" and/or "directory".
+            "all" is equivalent to ["file", "directory"]. Defaults to "file".
```
Light2Dark, line 173: doesn't this mean ["file", "directory"] is the same as "all". What's the reason to have both list & str support?

kirangadhave: yeah it's same as `all`. If we add something like `symlinks` or some other type in future, we can pick and choose with this pattern. `all` will always be all supported types. just a bit of over-engineering for future proofing.

Light2Dark: got it, im usually not keen on over-eng, but leaving it to you.

Light2Dark (approved):

PR #9635, opened 2026-05-21
automation script to pull models.yml

Light2Dark:
## 📝 Summary

image

- Models.yml is used to populate our models list. We manually updated this. This PR introduces a script to automate, although not perfect.
- Add automation script that needs to be run manually to pull latest models data. By default pulls latest 10 models from each provider. We can pass params to control this.
- Enriched the models data with more info (cost, capabilities)
- Ive changed the data structure of models.yml to be more readable and collapsible by provider. Else, we could have duplicates under the previous model. Models are grouped under provider and sorted by release date.
- Adds a skill to write descriptions.

Pulling this data from https://models.dev. It has an open-source API. Considered openrouter as well, but it has a slightly different structure. Anyway, easy to changeover if needed.

Some models don't exist on the API, maybe we can cross-check. I've removed them manually for now (eg. `gpt-5.5-codex-spark`)

We could put this into a github actions workflow in the future.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist


packages/llm-info/src/sync-models.ts
```diff
@@ -0,0 +1,304 @@
...
+  }
+  return total;
+}
+
+export async function syncModels(options: SyncOptions): Promise<SyncResult> {
+  const {
+    modelsYamlPath,
+    write = true,
+    mode = "append",
+    maxPerProvider,
```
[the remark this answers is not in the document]
Light2Dark: This was addressed in `2f7425a` — the dedup landed in `merge.ts` rather than `sync-models.ts`, so the diff for this file looks unchanged but the underlying behaviour is fixed.

`mergeModels` now accumulates candidates per *marimo* provider into a single `Map<modelId, AiModel>`, then calls `sortAndTrim` once per marimo provider. See `merge.ts` ~L174-L212 (`candidatesByProvider`).

Regression test in `packages/llm-info/src/__tests__/sync-models.test.ts`:

```
it("enforces `maxPerProvider` after deduping across mapped providers", ...)
```

passes `google` + `google-vertex` with overlapping ids and `maxPerProvider: 2`, and asserts the merged bucket has exactly 2 entries. So `-n 5` cannot produce 10 anymore.

packages/llm-info/src/sources/models-dev.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export interface ModelsDevProvider {
+  id: string;
+  name?: string;
+  models: Record<string, ModelsDevModel>;
+}
+
+export type ModelsDevApi = Record<string, ModelsDevProvider>;
+
+const DEFAULT_URL = "https://models.dev/api.json";
```
kirangadhave, line 43: most providers have their own urls which might have most up to date info. It might be worth it just maintain those urls instead.

kirangadhave: Also how much work would it be to fetch this live? i.e instead of reading from models.json, make this query

Light2Dark: Unfortunately, it doesn't have the full information we need like cost and capability. Different providers give a different structure. So, this is more unified.

Light2Dark: forgot to attach a link: https://developers.openai.com/api/reference/typescript/resources/models/methods/list
https://platform.claude.com/docs/en/api/typescript/models/list

kirangadhave (approved):
🚀 This is good. Should we have a job that run this periodically and opens a PR? Say weekly or monthly?

Light2Dark:
> 🚀 This is good. Should we have a job that run this periodically and opens a PR? Say weekly or monthly?

Yeah, I can follow-up on that. Thanks!

PR #9650, opened 2026-05-21
perf(chrome): preload panel chunks on hover/focus

mscolnick:
Wrap all sidebar/dev-panel lazy imports with reactLazyWithPreload and
prefetch the corresponding chunk when the user hovers or focuses the
icon/tab. On mount, idle-preload the persisted selected panel(s) if
their parent was last left open. Adds a theme-matched skeleton for the
terminal panel so first open isn't a blank flash.

### DIFF
--- a/frontend/src/components/editor/chrome/wrapper/lazy-panels.ts
+++ b/frontend/src/components/editor/chrome/wrapper/lazy-panels.ts
@@ -1,0 +1,91 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+
+import { reactLazyWithPreload } from "@/utils/lazy";
+import { Logger } from "@/utils/Logger";
+import type { PanelType } from "../types";
+
+// Preloading is best-effort: a chunk-load failure here should not surface as
+// an unhandled rejection (the panel will retry the import when actually
+// rendered, where Suspense/ErrorBoundary handle the failure).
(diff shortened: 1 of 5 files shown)


frontend/src/components/editor/chrome/wrapper/lazy-panels.ts
```diff
@@ -0,0 +1,91 @@
...
+);
+export const LazyCachePanel = reactLazyWithPreload(
+  () => import("../panels/cache-panel"),
+);
+
+// Preloader registry: hovering an icon/tab calls into this map to warm the
+// corresponding chunk. Two panel types (chat and agents) share the "ai" slot,
+// so we preload both.
+export const PANEL_PRELOADERS: Record<PanelType, () => void> = {
+  files: safePreload(LazyFileExplorerPanel),
```
kirangadhave, line 73: why not wrap the panel object with `reactLazyWithPreload`:
```ts
export const LazyCachePanel = safePreload(
	reactLazyWithPreload(() => import("../panels/cache-panel"))
)
```

kirangadhave (approved):
🚀 nice!

frontend/src/components/editor/chrome/wrapper/lazy-panels.ts
mscolnick: that could work too. i guess it would matter if it was used in more places

PR #9651, opened 2026-05-21
chore(openapi): regenerate api.yaml after docstring + dep drift

mscolnick:
## Summary

`packages/openapi/api.yaml` drifted out of sync with the generator, breaking `tests/_cli/test_cli_development.py::test_openapi_up_to_date` on every PR since yesterday afternoon.

Two changes that touched docstring rendering both landed without regenerating the schema file:

1. **`pydantic-ai-slim` v1.96.0** changed docstrings on types referenced by marimo's schema (e.g. `pydantic_ai.ui.vercel_ai.request_types.UIMessage`). The CLI test installs deps freely via `uv pip install --group=test-optional --group=dev marimo*.whl`, so it picked up the new version as soon as it was released.
2. **#9645** ("enforce single backticks in Python docstrings") rewrote docstrings across the codebase but didn't regenerate `api.yaml`.

This PR regenerates via:
```sh
marimo development openapi > packages/openapi/api.yaml
pnpm -w run codegen
```

The diff is **description-only** — no schema additions, removals, or renames. Just docstring formatting (RST → Markdown backticks, multi-line dedenting, etc).

## Followups (separate)

`test_openapi_up_to_date` is fragile because the CLI test resolves deps freely — any minor docstring change in a transitive dep silently breaks it until someone regenerates. Worth either pinning deps for that test or stripping `description` fields from the comparison.

## Test plan
- [x] `uv run --group test-optional pytest tests/_cli/test_cli_development.py::test_openapi_up_to_date -v` passes locally


mscolnick:
Closing — investigation was wrong. PR #9645 already regenerated `api.yaml` as part of the docstring style change; main has been correct since that merged at 17:12 UTC. The failures on PRs based on earlier main commits will resolve on rebase.

PR #9664, opened 2026-05-22
fix(cache): mo.cache raises KeyError: '__scratch__' in scratchpad

everettroeth:
## Why

`@mo.cache`, `@mo.lru_cache`, and `with mo.persistent_cache(...)` raise `KeyError: '__scratch__'` when used from the `/api/kernel/execute` scratchpad endpoint:

```
File "marimo/_save/save.py", line 232, in _set_context
    self.scoped_refs |= graph.cells[cell_id].defs & set(glbls.keys())
KeyError: '__scratch__'
```

`ctx.graph` returns the kernel's main graph (`marimo/_runtime/context/kernel_context.py:46`), but the scratchpad cell is registered only in the Runner's local graph (`marimo/_runtime/runtime.py:1706-1707`), so the decorator's cell lookup misses. The same root cause raises `CacheException` from `_cache_context.trace` (`save.py:711`) and `KeyError` from `mo.defs()` / `mo.refs()` (`runtime.py:195,222`).

## What

In `Kernel.run_scratchpad`, also register `SCRATCH_CELL_ID` in `self.graph` for the duration of the run, unregister in a `try/finally`. The scratch cell's `defs` and `refs` are cleared at `runtime.py:1702-1703`, so this adds no edges, triggers no reactive recomputation, and fires no callbacks. State updates are flushed after cleanup so reactive cells never observe `__scratch__`.

Three regression tests added under `TestExecution` in `tests/_runtime/test_runtime.py` alongside the existing `test_run_scratch*` tests.

### DIFF
--- a/marimo/_runtime/runtime.py
+++ b/marimo/_runtime/runtime.py
@@ -1706,31 +1706,51 @@ class Kernel:
         graph = dataflow.DirectedGraph()
         graph.register_cell(SCRATCH_CELL_ID, cell)
 
-        # Copy hooks and add scratchpad-specific hooks
-        scratchpad_hooks = self._hooks.copy()
-        scratchpad_hooks.add_on_finish(
-            self.packages_callbacks.missing_packages_hook
-        )
+        # Also register __scratch__ in the kernel's main graph so any
+        # runtime-context consumer that reads ctx.graph.cells[cell_id]
+        # during scratchpad execution resolves it correctly (e.g.,
+        # @mo.cache, @mo.lru_cache, mo.persistent_cache). The scratch
+        # cell's defs and refs are cleared above, so this adds no edges
+        # and has no reactive side effects. We unregister in a `finally`
+        # so a scratchpad crash cannot leave the main graph polluted.
+        did_register_in_main_graph = False
+        if SCRATCH_CELL_ID not in self.graph.cells:
+            self.graph.register_cell(SCRATCH_CELL_ID, cell)
+            did_register_in_main_graph = True
+        try:
+            # Copy hooks and add scratchpad-specific hooks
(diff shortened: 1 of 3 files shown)


marimo/_runtime/runtime.py
```diff
@@ -1706,31 +1706,51 @@ async def run_scratchpad(self, code: str) -> None:
...
-            self.packages_callbacks.missing_packages_hook
-        )
+        # Also register __scratch__ in the kernel's main graph so any
+        # runtime-context consumer that reads ctx.graph.cells[cell_id]
+        # during scratchpad execution resolves it correctly (e.g.,
+        # @mo.cache, @mo.lru_cache, mo.persistent_cache). The scratch
+        # cell's defs and refs are cleared above, so this adds no edges
+        # and has no reactive side effects. We unregister in a `finally`
+        # so a scratchpad crash cannot leave the main graph polluted.
+        did_register_in_main_graph = False
```
dmadisetti: I think it would be better to put this directly in cache opposed to runtime?

There's also an executor refactor that could be a better place to put this if we did want to keep it in runtime.

But thanks for opening this!

dmadisetti:
Thanks @everettroeth ! This also just came on our radar. Let me know what you think about my comment of handling this in cache itself. Patching __scratch__ into the graph may do weird things

marimo/_runtime/runtime.py
everettroeth: Thanks for the speedy review @dmadisetti! Sounds good yep, cleaner in cache. Updated.

everettroeth:

dmadisetti:
Nice! Thank you

dmadisetti (approved):

PR #9667, opened 2026-05-22
feat(file_browser): add filter param for regex and callable filtering

### LINKED ISSUE #8399: file browser should more flexible filtering
### Description

Currently, file filtering is limited to 'filetypes'. 

### Suggested solution

Have a 'regex' pattern or callback filtering function.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

kratos0718:
Closes #8399

## 📝 Summary

Adds a `filter` keyword argument to `mo.ui.file_browser()` so users can filter files using regex patterns or custom functions — not just file extensions.

**Three ways to use it:**
- Regex string: `filter=r"^train_"`
- Compiled pattern: `filter=re.compile(r"\.csv$", re.IGNORECASE)`
- Callable: `filter=lambda p: p.stat().st_size > 1_000_000`

Directories are always shown for navigation. Works together with `filetypes` and `ignore_empty_dirs`.

##  Merge Checklist

- [x] I have read the contributor guidelines
- [x] Documentation has been updated (docstring updated)
- [x] Tests have been added (8 new tests, 79 total pass)
- [x] AI generated code reviewed line-by-line by the PR author

### DIFF
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -316,6 +316,28 @@ class file_browser(
         path = self._path_cls(path_str, **kwargs)
         return path
 
+    def _passes_filter(self, file: Path) -> bool:
+        """Return True if `file` passes the configured `filter`.
+
+        Centralizes filter evaluation so `_list_directory` and
+        `_has_files_recursive` stay in sync. A callable filter that raises is
+        treated as "does not match" rather than crashing the whole directory
+        listing, so one bad file can't take down the browser.
+        """
+        if self._filter is None:
+            return True
+        if isinstance(self._filter, re.Pattern):
+            return self._filter.search(file.name) is not None
+        try:
+            return self._filter(file)
+        except Exception:
+            LOGGER.warning(
+                "file_browser filter callable raised on %s; treating as no match",
+                file,
+                exc_info=True,
+            )
+            return False
+
     def _has_files_recursive(
         self, directory: Path, max_depth: int = 100
     ) -> bool:
@@ -353,13 +375,8 @@ class file_browser(
                     ):
                         continue
                     # Apply regex or callable filter
-                    if self._filter is not None:
-                        if isinstance(self._filter, re.Pattern):
-                            if not self._filter.search(item.name):
-                                continue
-                        else:
-                            if not self._filter(item):
-                                continue
+                    if not self._passes_filter(item):
+                        continue
                     return True
                 elif item.is_dir() and not item.is_symlink():
                     # Skip directory symlinks to avoid infinite loops
@@ -417,13 +434,8 @@ class file_browser(
                     continue
 
             # Apply regex or callable filter to files
-            if self._filter is not None and not is_directory:
-                if isinstance(self._filter, re.Pattern):
-                    if not self._filter.search(file.name):
-                        continue
-                else:
-                    if not self._filter(file):
-                        continue
+            if not is_directory and not self._passes_filter(file):
+                continue
 
             # Skip empty directories if ignore_empty_dirs is enabled
             if self._ignore_empty_dirs and is_directory:
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -261,9 +261,7 @@ class file_browser(
             )
         elif isinstance(filter, str):
             self._filter = re.compile(filter)
-        elif isinstance(filter, re.Pattern):
-            self._filter = filter
-        elif callable(filter):
+        elif isinstance(filter, re.Pattern) or callable(filter):
             self._filter = filter
         else:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


kratos0718:

kratos0718:

kratos0718:
Hi, just following up on this PR. 
All checks are passing and CLA is signed. 
Happy to make any requested changes!

kratos0718:
Hi! Just checking in — all automated checks are passing and CLA is signed. Happy to make any requested changes or incorporate feedback. Thanks for your time!

kirangadhave:
@cubic-dev-ai

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -330,6 +352,14 @@ def _has_files_recursive(
...
                     ):
                         continue
+                    # Apply regex or callable filter
+                    if self._filter is not None:
+                        if isinstance(self._filter, re.Pattern):
+                            if not self._filter.search(item.name):
+                                continue
+                        else:
+                            if not self._filter(item):
+                                continue
```
kirangadhave: this logic is duplicated on `L421-L426`, extract to a helper. Easier to read and makes sure both don't diverge in future.

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -386,6 +416,15 @@ def _list_directory(
                 if extension.lower() not in self._filetypes:
                     continue
 
+            # Apply regex or callable filter to files
+            if self._filter is not None and not is_directory:
+                if isinstance(self._filter, re.Pattern):
+                    if not self._filter.search(file.name):
+                        continue
+                else:
+                    if not self._filter(file):
```
kirangadhave: the `callable` version of filter is not wrapped in `try...expect`. Handle it gracefully here.

kirangadhave: Also add a test case for _callable_ filter which raise exceptions

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -330,6 +352,14 @@ def _has_files_recursive(
                         and item.suffix.lower() not in self._filetypes
                     ):
                         continue
+                    # Apply regex or callable filter
+                    if self._filter is not None:
+                        if isinstance(self._filter, re.Pattern):
+                            if not self._filter.search(item.name):
```
kirangadhave: since we are use `search` instead of `match`, let's add a note in docstring indicating that `search` will match within the filename, use `^/$` to anchor

kratos0718:
Thanks for the review! Addressed both findings:

- **P1 (exception isolation):** filter evaluation now runs through a single `_passes_filter` helper that wraps a callable filter in `try/except`. If the callable raises on a file, that file is treated as a non-match and logged, instead of crashing the whole directory listing.
- **P2 (duplicated logic):** the regex/callable branching previously duplicated in `_list_directory` and `_has_files_recursive` now lives only in `_passes_filter`, so the two paths can't drift.

Added a regression test (`test_filter_callable_exception_is_isolated`) covering the exception-isolation case.

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -330,6 +352,14 @@ def _has_files_recursive(
...
+                                continue
+                        else:
+                            if not self._filter(item):
```
kirangadhave: maybe we should wrap this in it's own `try...except`?
for e.g. if we get OSError here, we still want to try filtering other files. For e.g. if directory has 100 files and the first we one we attempt to filter is a broken symlink or something, current setup will just skip the remaining 99 files.

we should catch the OSError exceptions, and return `False`. Error means the file doesn't pass our filter. Other errors should propagate up.

kirangadhave (changes requested):
Hi @kratos0718, the PR looks great. Please address the comments and we should be good to merge.

Overall note:
Please add logging with `LOGGER.debug(...)` to except blocks, e.g.
```python
try:
	return bool(self._filter(file))
except OSError as e:
	LOGGER.debug(f"file_browser filter could not evaluate {file}: {e}")
```

we should add logging in other places as well, but for now please add it to the new code you're adding in this PR.

kratos0718:
Thanks for the detailed review @kirangadhave! Addressed everything:

- **Duplicated filter logic** — extracted into a single `_passes_filter` helper, now used by both `_list_directory` and `_has_files_recursive`.
- **Exception handling** — the callable filter is wrapped in `try/except OSError`. An `OSError` (e.g. a broken symlink) is treated as a non-match so the remaining files still list; any other exception propagates up as you suggested.
- **Logging** — skipped files are logged with `LOGGER.debug(f"file_browser filter could not evaluate {file}: {e}")`.
- **Docstring** — noted that the regex filter uses `search` (matches within the filename) and that `^`/`$` should be used to anchor.
- **Tests** — added `test_filter_callable_oserror_is_isolated` (OSError isolated) and `test_filter_callable_non_oserror_propagates` (other errors propagate).

Ready for another look — thanks!

kirangadhave (approved):
🚀 Thanks for addressing the comments. Ready to merge.

PR #9673, opened 2026-05-22
Fix dropped error hints and improve error output UI

akshayka:
This change improves how statically parsed errors and exceptions are rendered in marimo's frontend, making them more compact, more legible, and conveying more information.

**Motivation.** Reactive notebooks can emit many errors. Our errors were taking up a lot of space (padding, redundant information), reducing information density. They were also a bit aggressively red in their styling, especially static errors such as multiple definition errors, which can make users feel alarmed or stressed.

**Changes.** Python's "Did you mean: ..." suggestions for NameError, AttributeError, and similar exceptions were never reaching the frontend; this change fixes that.

The rest of the change tidies how errors are presented, making them more compact and legible. The console-area traceback loses its collapsible accordion, which only repeated the exception message already shown above it. The output area loses its "See console area for traceback", unless outputs are configured to be shown above the cell.

Static errors (multiple definitions, cycles, setup-cell references, star imports) no longer include the aggressive red "This cell wasn't run because it has errors text". Instead, a subtle cell not run title is included, with a tooltip provided more context.

## Examples

**Multiple definition error.**

image

**With tooltip.**

image

**Exceptions.**

image


dmadisetti (commented):
For posterity this is what it currently looks like:

Image

Image

I thought we maybe mentioned automatically collapsing the stacktrace and just putting the error? I think a collapsable stack track still may make sense (see below for an extreme example)

Image

Aside from that, I think this is nice!

akshayka:
I don't think we should collapse stacktraces, they are valuable information

akshayka:
Thanks @dylan. I've made the traceback collapsible, thank you for catching that regression.

image

dmadisetti (approved):
Nice

PR #9675, opened 2026-05-23
Add padding between cell number and minimap dependency lines

akshayka:
In the dependency-graph minimap, the upstream whisker on a selected cell extended to within ~2px of the cell-number label's container, making the line visually collide with the number when a cell was selected.

This change shifts the dependency-dot SVG and the variable list 6px to the right so the whisker has breathing room against the label.

image


Light2Dark (approved):
Image

I don't really notice any issue with the original styles but deferring to you if there is

akshayka:
Thanks, for some configurations (when there is an ancestor and descendant) the line overlaps the number (Trevor reported the issue)

PR #9691, opened 2026-05-26
perf(table): lazy download-size RPC + first-page extrapolation

### LINKED ISSUE #9688: Very long loading time to display a polars Dataframe.
### Describe the bug

Displaying a polars dataframe inside a Marimo notebook takes so long compared to before, this seems to happen in the marimo versions 0.23.8 and 0.23.7, as in 0.23.6 it does work well. I'm using polars 1.41.0.

Image

Image

11s vs 400ms

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.7",
  "editable": false,
  "location": "C:/Users/EliottCrancee/Documents/Code/smol_dispatch/smol-dispatch/.venv/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Python Version": "3.13.9",
  "Locale": "--",
  "Binaries": {
    "Browser": "148.0.7778.179",
    "Node": "v25.0.0",
    "uv": "0.11.7 (9d177269e 2026-04-15 x86_64-pc-windows-msvc)"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.21.2",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.1.0",
    "tomlkit": "0.15.0",
    "uvicorn": "0.48.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "mcp": "1.27.1",
    "openai": "2.38.0",
    "pandas": "2.3.3",
    "polars": "1.41.0",
    "pyarrow": "24.0.0",
    "pytest": "9.0.3",
    "ruff": "0.15.14"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

inside a marimo cell :
```python
import numpy as np
import polars as pl

n_rows = 1_000_000
n_cols = 10

rng = np.random.default_rng(seed=42)

df = pl.DataFrame(
    {
        f"col_{i}": rng.random(n_rows)  # float random in [0, 1)
        for i in range(n_cols)
    }
)

df
```

kirangadhave:
## Summary

Stop blocking the polars/pandas DataFrame render on a full-table JSON serialization. The size measurement is needed only when a host sets `downloadSizeLimitAtom`, and even then it is now cheap.

Two layered fixes:
- New `get_size_bytes` RPC on `mo.ui.table` and `mo.ui.dataframe`. Standalone marimo never calls it; hosts call it only when their policy atom is set (vscode sets it currently).
- The RPC serializes a 100-row head sample and scales by `total_rows / sample_size`. Returns in single-digit ms even on million-row frames. Lives as `TableManager.estimate_size_bytes` so both UI widgets share one implementation.

For the regression's 1M-row repro, `table(df)._mime_()` drops from ~6.3 s to ~ms-scale, matching polars' own `_repr_html_()`. The new RPC also resolves in ~ms because it only sanitizes 100 rows.

Fixes https://github.com/marimo-team/marimo/issues/9688


marimo/_plugins/ui/_impl/dataframes/dataframe.py
```diff
@@ -199,7 +197,6 @@ def __init__(
                 "columns": self._get_column_types(),
                 "dataframe-name": dataframe_name,
                 "total": rows,
-                "size-bytes": self._get_json_size_bytes(self._manager),
```
dmadisetti, line 202: Was this this blocking bit?

marimo/_plugins/ui/_impl/tables/table_manager.py
```diff
@@ -24,6 +24,10 @@
 
 T = TypeVar("T")
 
+# Rows used when estimating the JSON-serialized size of the rendered data.
+# Bigger samples are more precise but cost more on the kernel control loop.
+SIZE_ESTIMATE_SAMPLE_ROWS = 100
```
dmadisetti, line 30: Could still be large depending on the entry. Logic looks reasonable, but wonder if it would still be better to just use narwhals: https://narwhals-dev.github.io/narwhals/api-reference/dataframe/?h=bytes#narwhals.dataframe.DataFrame.estimated_size

kirangadhave: The `estimated_size` gives size on heap. We want the size of the json string which is encoded in data-uri. That's the part that crashes vscode. 

For e.g. `null` values in the estimated size calculation (it uses `pyarrow`) add almost nothing in final size calculation. when stringifed to `None` they start adding up.

marimo/_plugins/ui/_impl/dataframes/dataframe.py
kirangadhave: yup. This used to json stringify the entire table

marimo/_plugins/ui/_impl/tables/table_manager.py
dmadisetti: Seems like a good proxy then

dmadisetti (approved):

marimo/_plugins/ui/_impl/tables/table_manager.py
kirangadhave: yeah I hope this goes away in future as we implement virtual file system in `marimo-lsp` and stream the files. I have a ticket to track the work, but it's on low prio for now

PR #9692, opened 2026-05-26
feat(ai): add model dropdown for Ollama and custom OpenAI-compatible providers

### LINKED ISSUE #5975: listing available models from openai-compatible model providers when configuring AI
### Description

When configuring the AI features in Marimo to use a custom model provider, I'm not always sure of what models are available on that provider. It may not be possible to run `ollama list` against the model provider.

### Suggested solution

I'd like Marimo to poll the available models using an appropriate API, and use the result to pre-fill the model selector. It appears that the `openai.models.list()` function does exactly that.

- [ ] Yes

### Alternatives

Alternative is a simple script to manually poll the model list. In the example below, I'm using Ollama via OpenWebUI to provide a security layer, but it would apply just as well to any other provider that uses an openai-compatible API. `apikey` has already been loaded from a secret storage mechanism.

```python
# Using Ollama library
import ollama

llm_url = 'https://llm.example.com/ollama'

oc = ollama.Client(host=llm_url, headers={'Authorization': f'Bearer {apikey}'})
for m in oc.list().models:
    print(m.model)
```

does the same thing as this

```python
# Using openai library
import openai

oai_url = f'{llm_url}/vi'

oac = openai.Client(api_key=apikey, base_url=oai_url)
for oam in oac.models.list():
    print(oam.id)
```

These give me exactly the same result: the list of models available which I can then use to configure Marimo, eg. `codellama:13b-python`, `codellama:13b-instruct`, `deepseek-r1`, `deepseek-coder-v2`, or any of a hundred others...

### Additional context

_No response_

jens-koesling:
## 📝 Summary
When adding a custom AI model in the AI Models settings, users previously had to manually type the model name. This PR adds automatic model discovery for Ollama and custom OpenAI-compatible providers.

## Changes
- New `useProviderModels` hook that fetches available models via the Marimo backend
- New backend endpoint `GET /api/ai/models` that proxies requests to OpenAI-compatible `/v1/models` endpoints
- Model text input is replaced with a dropdown when models are available
- Model selection resets when switching providers (only if selected from dropdown, not manually typed)
- CORS issues avoided by proxying through Marimo backend

## Testing
Tested with Ollama and custom OpenAI-compatible providers.

Closes #5975

## 📋 Pre-Review Checklist

Bildschirmfoto 2026-05-26 um 21 55 02
 

## ✅ Merge Checklist

> **Note:** Tests for the backend endpoint and frontend hook will be added in a follow-up PR.

### DIFF
--- a/frontend/src/components/app-config/use-provider-models.ts
+++ b/frontend/src/components/app-config/use-provider-models.ts
@@ -11,7 +11,9 @@ export interface ProviderModelsResult {
  * Fetches available models from an OpenAI-compatible endpoint
  * via the Marimo backend proxy to avoid CORS issues.
  */
-export function useProviderModels(baseUrl: string | null | undefined): ProviderModelsResult {
+export function useProviderModels(
+  baseUrl: string | null | undefined,
+): ProviderModelsResult {
   const [models, setModels] = useState<string[]>([]);
   const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
@@ -39,7 +41,7 @@ export function useProviderModels(baseUrl: string | null | undefined): ProviderM
           throw new Error(`HTTP ${response.status}`);
         }
 
-        const data = await response.json() as { models: string[] };
+        const data = (await response.json()) as { models: string[] };
         setModels(data.models ?? []);
       } catch (err) {
         if (err instanceof Error && err.name === "AbortError") {
--- a/frontend/src/components/app-config/use-provider-models.ts
+++ b/frontend/src/components/app-config/use-provider-models.ts
@@ -1,0 +1,61 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+import { useState, useEffect } from "react";
+
+export interface ProviderModelsResult {
+  models: string[];
+  isLoading: boolean;
+  error: string | null;
+}
+
+/**
+ * Fetches available models from an OpenAI-compatible endpoint
+ * via the Marimo backend proxy to avoid CORS issues.
+ */
+export function useProviderModels(baseUrl: string | null | undefined): ProviderModelsResult {
+  const [models, setModels] = useState<string[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    if (!baseUrl) {
+      setModels([]);
+      setError(null);
+      return;
+    }
+
+    const controller = new AbortController();
+
+    const fetchModels = async () => {
+      setIsLoading(true);
+      setError(null);
+
+      try {
+        const params = new URLSearchParams({ base_url: baseUrl });
+        const response = await fetch(`/api/ai/models?${params}`, {
+          signal: controller.signal,
+        });
+
+        if (!response.ok) {
+          throw new Error(`HTTP ${response.status}`);
+        }
+
+        const data = await response.json() as { models: string[] };
+        setModels(data.models ?? []);
+      } catch (err) {
+        if (err instanceof Error && err.name === "AbortError") {
+          return;
+        }
+        setError("Could not fetch models from endpoint");
+        setModels([]);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    fetchModels();
+
+    return () => controller.abort();
+  }, [baseUrl]);
+
+  return { models, isLoading, error };
+}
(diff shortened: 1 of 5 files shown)


jens-koesling:

marimo/_server/api/endpoints/ai.py
```diff
@@ -571,3 +571,61 @@ async def mcp_refresh(
...
+            data = response.json()
+            model_ids = sorted(
+                [m["id"] for m in data.get("data", []) if m.get("id")]
+            )
+            return StructResponse({"models": model_ids})
+    except Exception as e:
+        LOGGER.warning(
+            "Could not fetch models from %s: %s", models_url, str(e)
+        )
+        return StructResponse({"models": []})
```
dmadisetti: Do we just return nothing if we timeout?

dmadisetti (commented):
Thanks! I know ollama is a common option, but should be directly compatible with OpenAI api style inference. CC @Light2Dark for whether we want to support this. Hold on before hacking on this any more, but you'll also need to regen api schema (`make codegen`)

marimo/_server/api/endpoints/ai.py
jens-koesling: Good point on the timeout handling. I chose to return an empty list on error/timeout so the UI gracefully falls back to manual text input. with this way the settings dialog still works even if the local provider is temporarily unavailable. Happy to change this to return a proper HTTP error code if preferred.

Light2Dark:
Thanks for this @jens-koesling, I think another option was to call openai.clients list method.
Any thoughts on why you chose calling the API over the sdk?

```
import openai

oai_url = f'{llm_url}/v1'

oac = openai.Client(api_key=apikey, base_url=oai_url)
for oam in oac.models.list():
    print(oam.id)
```

jens-koesling:
> Thanks for this @jens-koesling, I think another option was to call openai.clients list method. Any thoughts on why you chose calling the API over the sdk?
> 
> ```
> import openai
> 
> oai_url = f'{llm_url}/v1'
> 
> oac = openai.Client(api_key=apikey, base_url=oai_url)
> for oam in oac.models.list():
>     print(oam.id)
> ```

Thanks for reviewing! Regarding the httpx vs SDK approach:  
I chose the direct HTTP approach because I was considering to support other local model providers beyond Ollama (specifically LM Studio and similar OpenAI-compatible servers). 
furthermore it is core dependency for Marimo while openai is optional. 

My question: does using the OpenAI SDK require `openai` to be installed as a dependency on the client side? Or is the `import openai` in your example importing it via the Python environment that's already available in Marimo? 

however the main advantage is that this approach works with any OpenAI-compatible endpoint (Ollama, LM Studio, vLLM, etc.) without requiring additional dependencies. 
I am only running local models and don't have `openai` installed, so I was not even considering I could use this ;)

if you prefer the SDK approach I can change request method, no problem :)

marimo/_server/api/endpoints/ai.py
```diff
@@ -3,6 +3,8 @@
 
 from typing import TYPE_CHECKING, Literal
 
+import httpx
+from pydantic import BaseModel
```
Light2Dark, line 7: Users may not have pydantic. We should not require this as an import, only as TYPE_CHECKING import if needed.

marimo/_server/api/endpoints/ai.py
```diff
@@ -571,3 +577,60 @@ async def mcp_refresh(
...
+
+    # Normalize URL
+    if base_url.endswith("/v1"):
+        models_url = f"{base_url}/models"
+    else:
+        models_url = f"{base_url}/v1/models"
+
+    try:
+        async with httpx.AsyncClient(timeout=5.0) as client:
+            response = await client.get(models_url)
```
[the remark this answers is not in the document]
Light2Dark: I think this is fine. It's user configured anyway.

frontend/src/components/app-config/use-provider-models.ts
```diff
@@ -0,0 +1,63 @@
...
+      setModels([]);
+      setError(null);
+      return;
+    }
+
+    const controller = new AbortController();
+
+    const fetchModels = async () => {
+      setIsLoading(true);
+      setError(null);
```
Light2Dark, line 32: consider using useAsyncData, it's a similar pattern

Light2Dark (commented):
Sorry for the many comments from copilot & me. I can also have a look at implementing the changes if you'd like.

- Id also add a comment on the frontend detailing why we do this in brief.

Issue #4165, opened 2025-03-19
Ability to hide columns in table viewer


delennc:
### Description

I want to be able to hide/show columns in the table viewer, so that I don't have to keep editing the code cell to specify what columns to show

### Suggested solution

similar to Excel functionality, where you can hide columns and then expand them again

mscolnick:
We can look into adding this into the core library. In the meantime, you can build this yourself in marimo:

```python
# cell 1
data = pl.read_csv(my_data.csv)

# cell 2
include_columns = mo.ui.multiselect(options=data.columns)
exclude_columns = mo.ui.multiselect(options=data.columns)
mo.hstack([include_columns, exclude_columns])

# cell 3
_data = data

if include_columns.value:
   _data = _data.select(include_columns.value)
if exclude_columns.value:
   _data = _data.drop(exclude_columns.value)

_data
```

aschreiber1:
This would be very useful if this features was included, its very hard when the model and the view layer have to show the same data (one of the reasons they are usually separate)

guiguid:
Hi, this is also very useful when using **data.value** returned by selected row.

The Ui doesn't need to show all column to the user, but the selected row need all column to process **data.value**.

<img width="1041" height="246" alt="Image" src="https://github.com/user-attachments/assets/ccda00b5-f474-4649-ae9c-f1d9b107eb20" />

```python
if data.value :
    # do what you want with selected row but with all column's values ...
```

mscolnick:
@guiguid in your request. how do you want to hide these columns? in code or via the GUI?

guiguid:
Hi Myles,
I think an option in mo.ui.table.
Example : 
```python
mo.ui.table(data=office_characters, pagination=True, column=[listofcolumntoshow])
```
by default : if "column" is omit : all column are shown.

Thx.

kirangadhave:
Fixed in https://github.com/marimo-team/marimo/pull/9687 and https://github.com/marimo-team/marimo/pull/9696

Issue #9220, opened 2026-04-16
[FR] Table: hidden_columns, column_widths, and full CSS pass-through in style_cell


yan-hic:
Three related gaps in `mo.ui.table` that currently require shadow-root CSS injection via anywidget to work around:

1. `hidden_columns: list[str]` — hide columns from display while keeping them in the DataFrame and `.value` selection. The data stays intact; only the visual rendering drops the column. Similar to `freeze_columns_left` / `wrapped_columns` in shape.
2. `column_widths: dict[str, int]` — pin specific columns to a fixed pixel width. Currently the only lever is auto-sizing.
3. `style_cell` full CSS pass-through — the frontend currently renders only `backgroundColor`, `color`, and `fontWeight` from the dict returned by `style_cell`. Properties like `backgroundImage`, `borderLeft`, `boxShadow` are present in `data-cell-styles` but never applied to the DOM. Rendering all standard CSS properties (or at least `background-image` and `border-left`) would make `style_cell` usable for accent bars and other cell-level visuals without JS injection.

yan-hic:
@kirangadhave glad you take this on !
Actually drag-to-resize columns would be very useful (xl, gsheet), especially if widths persist across refreshes e.g. saved in marimo.toml or notebook-level config. I can create a separate FR if better. Thx !

kirangadhave:
@yan-hic Actually it would be great if you can create a separate FR for `column_widths` so we can collect some feedback from the community. 

I have `hidden_columns` and `style_cell` on my todo.

kirangadhave:
@yan-hic Can you share code for `style_cell` issue? I am not able to reproduce it. I tried this snippet with `backgroundImage`, `borderLeft`, and `boxShadow` and they seem to correctly render

```python
df = pd.DataFrame(
    {
        "name": ["alpha", "beta", "gamma", "delta"],
        "score": [10, 42, 75, 99],
    }
)


def style(row_id, col_name, value):
    if col_name != "score":
        return {}
    if value >= 75:
        return {
            "backgroundColor": "#ecfdf5",
            "backgroundImage": "linear-gradient(90deg, #6ee7b7 0%, #ecfdf5 100%)",
            "color": "#065f46",
            "fontWeight": "600",
            "borderLeft": "4px solid #10b981",
            "boxShadow": "inset 0 0 0 1px #a7f3d0",
        }
    if value < 25:
        return {
            "backgroundColor": "#fef2f2",
[...]

PR #9696, opened 2026-05-27
feat(table): Column Explorer visibility controls + smart-search

### LINKED ISSUE #9419: Search table columns by name and display only matching column
### Description
In marimo UI tables, I would like the search bar to support searching by column name, not just table cell values.

When working with wide tables containing many columns, it is difficult to quickly locate a specific column. The current table search behavior appears to filter rows based on values, but it does not let me search for a column name and narrow the view to that column.

### Requested behavior
Allow users to type a column name into the table search bar and show only the matching column or columns.

For example, if a table has columns:

`user_id`, `created_at`, `status`, `revenue`

Searching for:

`revenue`

would display only the revenue column.

### Use case
This would make wide tables much easier to inspect, especially when exploring dataframes with many columns.

kirangadhave:
## Summary
Second of two PRs implementing `mo.ui.table` column visibility (RFC). Builds on #9687 (PR 1: Python kwargs, hooks, hidden-count display, recovery banner).

- Substring filter replaced with `smartMatch`: queries split on whitespace and each word must prefix a word in the column name.
- Click hides/shows the column. Eye icon when visible, eye-off when hidden.
- "Unhide all" link + hidden count in the panel header. Shown only when at least one user column is hidden.

https://github.com/user-attachments/assets/6f08b1cf-c5ba-474b-8a49-6cb86c9d5625

Related to #9220
Closes #9419

### DIFF
--- a/frontend/src/components/data-table/TableBottomBar.tsx
+++ b/frontend/src/components/data-table/TableBottomBar.tsx
@@ -13,7 +13,7 @@ import {
 } from "../editor/chrome/panels/context-aware-panel/context-aware-panel";
 import { Button } from "../ui/button";
 import { toast } from "../ui/use-toast";
-import { getUserColumnVisibilityCounts } from "./hooks/use-column-visibility";
+import { getColumnCountForDisplay } from "./hooks/use-column-visibility";
 import { DataTablePagination, prettifyRowColumnCount } from "./pagination";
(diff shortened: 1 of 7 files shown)


akshayka:
Also closes #4165 :)

akshayka (approved):
LGTM on UI/UX, based on watching the video

PR #9703, opened 2026-05-27
feat(ai): per-provider max_tokens defaults with optional override

kirangadhave:
## Summary

Drop the blanket `DEFAULT_MAX_TOKENS = 4096` so reasoning-heavy models (Kimi K2.6, DeepSeek, gpt-oss, etc.) stop truncating or hard-erroring at 4096.

- Non-Anthropic providers omit `max_tokens` from settings entirely when unset, letting pydantic-ai fall through to each provider's own default.
- Anthropic provider call keeps a floor of `32768` because its API requires the field.
- Adds an optional `ai.max_tokens` config field plus a Settings > AI checkbox that, when on, applies one explicit limit to every provider (including Anthropic).

image

Kimi K2.6 working with ~27k chars (roughly 6-7k tokens)
image

After UI override to 100 `max_tokens`, reasoning errors out at 550 chars
image

## Behavior
| Override      | Anthropic          | Other providers                                |
|---------------|--------------------|------------------------------------------------|
| Off (default) | server sends 32768 | `max_tokens` omitted; provider default applies |
| On, value `N` | `N`                | `N`                                            |

Closes MO-6314

### DIFF
--- a/frontend/src/components/app-config/ai-config.tsx
+++ b/frontend/src/components/app-config/ai-config.tsx
@@ -1286,11 +1286,27 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
           const isOn = field.value != null;
           return (
             <div className="flex flex-col gap-y-1">
-              <FormItem className={formItemClasses}>
-                <FormLabel className="font-normal">
-                  Override max output tokens
-                </FormLabel>
-                <FormControl>
+              <div className="flex items-center gap-x-2">
+                <FormItem className={formItemClasses}>
+                  <FormLabel className="font-normal">
+                    Max output tokens
+                  </FormLabel>
+                  <FormControl>
+                    <Input
+                      data-testid="ai-max-tokens-input"
+                      type="number"
+                      min={1}
+                      disabled={!isOn}
+                      className="w-28 h-6"
+                      value={field.value ?? 32768}
+                      onChange={(e) => {
+                        const n = Number.parseInt(e.target.value, 10);
+                        field.onChange(Number.isFinite(n) && n > 0 ? n : null);
+                      }}
+                    />
+                  </FormControl>
+                </FormItem>
+                <FormItem className={formItemClasses}>
                   <Checkbox
                     data-testid="ai-max-tokens-checkbox"
                     checked={isOn}
@@ -1313,34 +1329,15 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
                       onSubmit(form.getValues());
                     }}
                   />
-                </FormControl>
-              </FormItem>
-              <FormDescription>
-                We recommend leaving this off, as each provider applies its own
-                limit (except Anthropic, where we set a default). Use it to
-                control costs or raise limits for heavy-reasoning models.
-              </FormDescription>
-              <div className="flex flex-col gap-y-1 pl-6">
-                <FormItem className={formItemClasses}>
-                  <FormLabel className="font-normal">
-                    Max output tokens
-                  </FormLabel>
-                  <FormControl>
-                    <Input
-                      data-testid="ai-max-tokens-input"
-                      type="number"
-                      min={1}
-                      disabled={!isOn}
-                      className="m-0 inline-flex h-7 w-28"
-                      value={field.value ?? 32768}
-                      onChange={(e) => {
-                        const n = Number.parseInt(e.target.value, 10);
-                        field.onChange(Number.isFinite(n) && n > 0 ? n : null);
-                      }}
-                    />
-                  </FormControl>
+                  <FormLabel className="font-normal">Override</FormLabel>
                 </FormItem>
               </div>
+
+              <FormDescription>
+                Each provider sets its own max output tokens (Anthropic uses a
+                recommended default). Adjust to control costs or enable more
+                output.
+              </FormDescription>
             </div>
           );
         }}
--- a/frontend/src/components/app-config/ai-config.tsx
+++ b/frontend/src/components/app-config/ai-config.tsx
@@ -1279,6 +1279,82 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
         )}
       />
 
+      <FormField
+        control={form.control}
+        name="ai.max_tokens"
+        render={({ field }) => {
+          const isOn = field.value != null;
+          return (
+            <div className="flex flex-col gap-y-1">
+              <FormItem className={formItemClasses}>
+                <FormLabel className="font-normal">
+                  Override max output tokens
+                </FormLabel>
+                <FormControl>
+                  <Checkbox
+                    data-testid="ai-max-tokens-checkbox"
+                    checked={isOn}
+                    onCheckedChange={(checked) => {
+                      // null signals delete to the server; cast because
+                      // UserConfig (OpenAPI-derived) types max_tokens as
+                      // `number | undefined`, but zod accepts `null`.
+                      const next = (checked === true ? 32768 : null) as
+                        | number
+                        | undefined;
+                      // shouldDirty: true forces RHF to keep this in
+                      // dirtyFields even when `next` happens to equal the
+                      // form's defaultValue (e.g. untick → tick when disk
+                      // started with 32768). Otherwise getDirtyValues
+                      // would skip it and the save body would be empty.
+                      form.setValue("ai.max_tokens", next, {
+                        shouldDirty: true,
+                        shouldTouch: true,
+                      });
+                      onSubmit(form.getValues());
+                    }}
+                  />
+                </FormControl>
+              </FormItem>
+              <FormDescription>
+                Recommended: leave off. Each provider applies its own output
+                limit, which is usually the right value for the model in use.
+                Anthropic is the exception: its API requires an explicit limit,
+                so marimo sends 32,768 by default.
+              </FormDescription>
+              <div className="flex flex-col gap-y-1 pl-6">
+                <FormItem className={formItemClasses}>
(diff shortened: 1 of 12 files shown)


kirangadhave:
@cubic-dev-ai

frontend/src/components/app-config/ai-config.tsx
```diff
@@ -1279,6 +1279,82 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
...
+                      // dirtyFields even when `next` happens to equal the
+                      // form's defaultValue (e.g. untick → tick when disk
+                      // started with 32768). Otherwise getDirtyValues
+                      // would skip it and the save body would be empty.
+                      form.setValue("ai.max_tokens", next, {
+                        shouldDirty: true,
+                        shouldTouch: true,
+                      });
+                      onSubmit(form.getValues());
+                    }}
```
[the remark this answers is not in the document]
kirangadhave: That is by design. Once enabled, folks will immediately edit the input box if required

frontend/src/components/editor/ai/completion-utils.ts
```diff
@@ -40,13 +41,16 @@ export function getAICompletionBody({
...
+
   return {
     includeOtherCode: getCodes(""),
     context: {
       plainText: contextString,
       schema: [],
       variables: [],
     },
+    ...(maxTokens !== undefined && { maxTokens }),
   };
```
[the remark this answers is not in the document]
kirangadhave: it's expected. backend has `None-as-delete`

frontend/src/components/editor/ai/completion-utils.ts
```diff
@@ -84,6 +90,7 @@ export async function getAICompletionBodyWithAttachments({
         schema: [],
         variables: [],
       },
+      ...(maxTokens !== undefined && { maxTokens }),
     },
```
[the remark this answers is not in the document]
kirangadhave: expected

frontend/src/components/editor/ai/completion-utils.ts
```diff
@@ -40,13 +41,16 @@ export function getAICompletionBody({
...
+  const maxTokens = store.get(resolvedMarimoConfigAtom).ai?.max_tokens;
+
   return {
     includeOtherCode: getCodes(""),
     context: {
       plainText: contextString,
       schema: [],
       variables: [],
     },
+    ...(maxTokens !== undefined && { maxTokens }),
```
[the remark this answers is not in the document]
kirangadhave: `null` is okay

frontend/src/components/app-config/ai-config.tsx
```diff
@@ -1279,6 +1279,82 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
...
+                </FormLabel>
+                <FormControl>
+                  <Checkbox
+                    data-testid="ai-max-tokens-checkbox"
+                    checked={isOn}
+                    onCheckedChange={(checked) => {
+                      // null signals delete to the server; cast because
+                      // UserConfig (OpenAPI-derived) types max_tokens as
+                      // `number | undefined`, but zod accepts `null`.
+                      const next = (checked === true ? 32768 : null) as
```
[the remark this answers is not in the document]
kirangadhave: that's fine

frontend/src/components/app-config/ai-config.tsx
```diff
@@ -1279,6 +1279,82 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
...
+                      onChange={(e) => {
+                        const n = Number.parseInt(e.target.value, 10);
+                        field.onChange(Number.isFinite(n) && n > 0 ? n : null);
+                      }}
+                    />
+                  </FormControl>
+                </FormItem>
+                {isOn && (
+                  <FormDescription>
+                    Override applies to every provider, including Anthropic. Use
```
dmadisetti:
```suggestion
                     Use
```

frontend/src/components/app-config/ai-config.tsx
```diff
@@ -1279,6 +1279,82 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
...
+                      onSubmit(form.getValues());
+                    }}
+                  />
+                </FormControl>
+              </FormItem>
+              <FormDescription>
+                Recommended: leave off. Each provider applies its own output
+                limit, which is usually the right value for the model in use.
+                Anthropic is the exception: its API requires an explicit limit,
+                so marimo sends 32,768 by default.
```
dmadisetti:
```suggestion
                Models which require an explicit limit (Anthropic) will still send a  default of 32,768 token.
```

marimo/_config/manager.py
```diff
@@ -400,6 +400,11 @@ def save_config(
         # Merge the current config with the new config
         current_config = self._load_config()
         merged = merge_config(current_config, config)
+        # None-as-delete: any key whose merged value is None (typically because
+        # the incoming config explicitly sent null) is removed from disk. Lets
+        # the UI clear optional scalars (e.g. ai.max_tokens) without a separate
+        # delete primitive.
+        _drop_none_values(cast(dict[str, Any], merged))
```
dmadisetti, line 407: On nice. Thought we already did this?

dmadisetti (commented):
Just some comments on language. Think it's a little awkward right now.

Looks like a few clean up things snuck into there, but nice!

marimo/_config/manager.py
kirangadhave: we didn't earlier, but now we have it if required.

frontend/src/components/app-config/ai-config.tsx
```diff
@@ -1279,6 +1279,79 @@ export const AiAssistConfig: React.FC<AiConfigProps> = ({
...
+                      }}
+                    />
+                  </FormControl>
+                </FormItem>
+                {isOn && (
+                  <FormDescription>
+                     Use
+                    this to cap response length for cost control, or to raise
+                    the limit for reasoning-heavy models that need longer
+                    outputs.
```
kirangadhave:
```suggestion
                    Use this to cap response length for cost control, or to raise
                    the limit for reasoning-heavy models that need longer
                    outputs.
```

marimo/_config/manager.py
Light2Dark, line 407: We should note this for bug bash. We had issues in the past where config changes on the frontend didn't save, this whole function is a bit fragile.

marimo/_server/ai/constants.py
```diff
@@ -1,5 +1,5 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-DEFAULT_MAX_TOKENS = 4096
+ANTHROPIC_DEFAULT_MAX_TOKENS = 32768
```
Light2Dark, line 4: I'd add a link to where it mentions this is a limit.

Light2Dark (commented):
Is there a issue this is attached too? I'm wondering if we should avoid adding UI config, just keep it in the toml.

If not, imo the UI is abit verbose. I would just say "Max tokens" and the description about overriding. You should also use the <IsOverriding> component to check if it has been overrided by the backend.

marimo/_server/ai/constants.py
kirangadhave: the above number is not a limit. It's just a good enough default. The only "limit" is that we have to provide some max_token. We can bump this up or down as required

marimo/_config/manager.py
kirangadhave: I'll add it to notes for next release

kirangadhave:
>Is there a issue this is attached too? I'm wondering if we should avoid adding UI config, just keep it in the toml.

I feel there is value in UI config. Just so one can override it if they are using local models like Ollama with much lower limits. (i assume we can just add a local provider with url?)

> If not, imo the UI is abit verbose. I would just say "Max tokens" and the description about overriding. You should also use the component to check if it has been overrided by the backend.
How would backend overide it? Only source of truth is `toml`. Only thing backend has is a default value for anthropic provider

Do you suggest a specific copy for the text?

marimo/_config/manager.py
kirangadhave: do you think we need more tests around this?

kirangadhave:
> Is there a issue this is attached too? 

Only a linear issue. ID mentioned in PR desc

Light2Dark:
> Do you suggest a specific copy for the text?

Max tokens: [ disabled if override is unticked ]  [ ] Override
We recommend leaving this off as each provider applies their own limit, except for Anthropic where we set a default. Use this to control costs or raise limits for heavy-reasoning models.

maybe something like this?

> (i assume we can just add a local provider with url?)

yeap, thats possible

marimo/_config/manager.py
Light2Dark: Maybe..

The previous issues was if you add a custom model, then refresh, it wouldn't be saved.
Same if you ticked certain models in the AI models dropdown, and then refreshed, it would disappear. 
Same with deleting a custom model.

I would test ^ during bug bash

kirangadhave:
Fixed copy:

Screenshot 2026-05-28 at 11 18 20 PM

Reads much better now. Thanks @Light2Dark

Light2Dark:
> Fixed copy

i added a PR here https://github.com/marimo-team/marimo/pull/9720 for the UI, feel free to use it or leave it.

I had an issue where my config wasn't being saved on refresh. So it keeps showing this
image

Light2Dark (approved):
thanks for fixing!

PR #9705, opened 2026-05-27
fix: allow async cancellation

### LINKED ISSUE #2684: Cannot interrupt async cell
### Describe the bug

The following code:
```py
import asyncio

await asyncio.sleep(10)
```
cannot be interrupted: the cell runs until complete and then shows a `CancelledError`.

### Environment

```json
{
  "marimo": "0.9.11",
  "OS": "Linux",
  "OS Version": "6.8.0-47-generic",
  "Processor": "x86_64",
  "Python Version": "3.12.3",
  "Binaries": {
    "Browser": "130.0.6723.58",
    "Node": "v22.9.0"
  },
  "Dependencies": {
    "click": "8.1.3",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.1",
    "markdown": "3.7",
    "narwhals": "1.10.0",
    "packaging": "24.1",
    "psutil": "6.1.0",
    "pygments": "2.18.0",
    "pymdown-extensions": "10.11.2",
    "pyyaml": "6.0.2",
    "ruff": "0.6.9",
    "starlette": "0.41.0",
    "tomlkit": "0.13.2",
    "uvicorn": "0.32.0",
    "websockets": "12.0"
  },
  "Optional Dependencies": {
    "duckdb": "1.1.2"
  }
}
```

### Code to reproduce

```py
import asyncio

await asyncio.sleep(10)
```

dmadisetti:
## 📝 Summary

Closes #2684

Promotes scheduler to long lived singleton on kernel context for look-up of running cells. This in turn allows task cancellations on async cells, which were previously un-cancelable. Example of cancelation:

image

### DIFF
--- a/marimo/_runtime/runner/scheduler.py
+++ b/marimo/_runtime/runner/scheduler.py
@@ -1,21 +1,23 @@
 # Copyright 2026 Marimo. All rights reserved.
 """Scheduler: per-run cell queue, cancellation, and async-task tracking.
 
-Singular-scheduler invariant
-----------------------------
-At most one `Runner.run_all()` is on the stack per `KernelRuntimeContext`
-at any time. The kernel serializes runs — control requests queue and
-state-update cascades only re-enter after `run_all()` returns. Embedded
-apps push a child `KernelRuntimeContext` (not a nested scheduler on the
-same context). Under this invariant, `KernelRuntimeContext._active_scheduler`
-can be a singular field.
+Active-scheduler stack
+----------------------
+`KernelRuntimeContext._active_scheduler` points at the scheduler whose
+run is currently executing, so the SIGINT handler can route a cancel to
+it. Runs nest as a stack on a single context: a re-entrant `run_all`
+(e.g. code_mode `ctx.run_cell` driving a run from inside a running cell)
+suspends the outer run at its `await`, runs to completion, then restores
+the outer scheduler. `__aenter__` saves the previously-active scheduler
+and `__aexit__` restores it, so the field always names the innermost
+(currently executing) run. The kernel still serializes top-level runs —
+control requests queue and state-update cascades only re-enter after
+`run_all()` returns.
 
 A future non-blocking `AsyncScheduler.submit()` (returning before
-dispatch completes) will break this invariant by allowing concurrent
-schedulers on one context; that PR will need to promote
-`_active_scheduler` to a plural `OrderedDict[int, Scheduler]`.
-`__aenter__` below fails loudly if the invariant is ever silently
-broken.
+dispatch completes) would allow genuinely *concurrent* schedulers on one
+context; that PR will need to promote `_active_scheduler` from a saved
+stack to a plural `OrderedDict[int, Scheduler]`.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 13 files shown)


marimo/_runtime/runner/scheduler.py
```diff
@@ -87,3 +147,87 @@ def cancelled_cells(self) -> CancelledCells:
...
+        tracked: `ensure_future` and `_register_task` run as two
+        adjacent synchronous statements (the narrowest gap in pure
+        Python), then `_interrupted` is re-checked. A SIGINT delivered
+        between them flips `_interrupted` via `cancel_all`; the
+        re-check cancels the freshly-registered task before the loop
+        ever resumes it.
+        """
+        if self._interrupted:
+            coro.close()
+            raise asyncio.CancelledError
```
kirangadhave:
```suggestion
            raise asyncio.CancelledError()
```

marimo/_runtime/runner/scheduler.py
```diff
@@ -87,3 +147,87 @@ def cancelled_cells(self) -> CancelledCells:
...
+        task = asyncio.ensure_future(coro)
+        self._register_task(cell_id, task)
```
kirangadhave, line 171: what happens if SIGINT comes after scheduling the task and before registering it? If there is no other active task and current one is not yet registered, we will just throw, but the task will still complete in bg?

Maybe we can have a flag which we set to detect that a task is being registered and unset it after register_task

dmadisetti: Good catch — this is a real race, but I think it's acceptable rather than worth the machinery to close.

The window is the synchronous gap between `ensure_future` and `_register_task`: no `await` in between, so it's sub-millisecond and the task body hasn't actually started. If a SIGINT lands *there* and no other task is active, `has_active_tasks()` is `False`, so the handler takes the `raise MarimoInterrupt` branch instead of `cancel_all(); return`. The raise unwinds before we register or reach the `_interrupted` re-check, so that one task is never cancelled and runs to completion once the loop resumes.

Blast radius is just that single cell: `cancel_all()` still ran, so `_interrupted` is set and the rest of the run halts — only the in-flight cell's cancellation is missed. And since interrupt is a human-driven action, the recovery is simply hitting interrupt again, which now sees the registered task and cancels it.

Your flag idea is the right fix if we want to close it (a "registering" flag OR'd into `has_active_tasks()` so the handler cancels-and-returns instead of raising during the window — that keeps the existing re-check doing the actual cancel). I'd rather not grow the signal-path machinery for a sub-ms, self-correcting race right now, so leaving as-is and documenting. Happy to take the flag as a follow-up if you feel strongly.

dmadisetti: So not a no-op like I thought

Image

I benched the RC and it's `[RC] register gap 1208ns`

Not sure it's worth the added complexity. If this does crop up a bit I'd actually want to know

dmadisetti: Note the kernel is resilient to this, just a noisy message

kirangadhave: Thanks again for taking the time, this is super helpful to build my understanding. 

The reasoning makes sense. probably will only be an issue if the async task is super long and we hit the exact sub-ms window. Not worth added complexity.

kirangadhave (approved):
:rocket:

PR #9715, opened 2026-05-28
test(e2e): deflake toggle-cell-language spec

kirangadhave:
## Problem

The `toggle-cell-language` e2e test flakes in CI. After converting a cell to Markdown it asserts the code editor is hidden, but converting only hides the code on blur (code stays visible while the cell is focused, by design). The test never blurs the cell, so it races the dropdown-close blur against the 5s assertion timeout. This is a long-standing race; recent CI environment shifts started tipping it red. (This is Claude's hypothesis)

## Fix (test-only)

- Force a blur after "Convert to Markdown" (Escape, then click outside the cell) before asserting the editor is hidden, so hide-on-blur takes effect deterministically instead of racing.
- Add the `maybeRestartKernel` `beforeEach` guard the sibling specs already use, for session isolation (`title.py` is shared with `mode.spec.ts`).

No behavior change.

## Verification

Local, `--retries=0 --repeat-each=5 --workers=1`, chromium:

- With the blur: 5/5 pass.
- Without the blur: 5/5 fail.

### DIFF
--- a/frontend/e2e-tests/toggle-cell-language.spec.ts
+++ b/frontend/e2e-tests/toggle-cell-language.spec.ts
@@ -1,7 +1,7 @@
 /* Copyright 2026 Marimo. All rights reserved. */
 import { expect, test } from "@playwright/test";
 import { getAppUrl, resetFile } from "../playwright.config";
-import { openCellActions } from "./helper";
+import { maybeRestartKernel, openCellActions } from "./helper";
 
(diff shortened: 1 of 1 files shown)


dmadisetti (approved):
🎉  weird it started failing. I don't think it was flaky- reliably broken- but the functionality still works as far as I can see

PR #9730, opened 2026-05-29
chore: align local ruff with CI's dependency window

kirangadhave:
## Problem

`make py-check` runs `uv run ruff check --fix`, which **re-resolves** ruff rather than using a frozen lock (`uv.lock` is gitignored). Locally there was no upper bound, so it floated to whatever ruff `uv` resolved at setup time (observed 0.15.13 / 0.15.14 / 0.15.15 across environments). CI's lint job (`test_check` in `test_be.yaml`) runs the *same* command but with `UV_EXCLUDE_NEWER: "7 days"`, so CI resolves the latest ruff aged ≥7 days. The newer local ruff ships detections/fixes (`TC002`/`TC003` import moves, unused-import removals) that CI's windowed version lacks, so local checks rewrote files CI considered clean.

## Fix

Make local resolution reproduce CI's window at setup time:

- `pyproject.toml`: add `[tool.uv] exclude-newer = "7 days"`, mirroring CI's `UV_EXCLUDE_NEWER`. Local `uv run`/`uv lock` now resolve the same ruff (and other deps) CI would resolve when you set up. This is point-in-time: a fresh setup matches CI; a stale checkout matches once re-locked.
- `.pre-commit-config.yaml`: pin `ruff-pre-commit` to the version that window currently resolves (`v0.15.14`), since pre-commit.ci installs from the tag and isn't governed by the window. Renovate (`minimumReleaseAge: 7 days`) keeps the tag tracking.
- Commit the autofixes the windowed ruff produces: unused-import removals and type-only imports relocated into `TYPE_CHECKING`.

No version is hard-pinned; the floor stays `ruff>=0.15.9` and the window controls the upper bound.

## Verification

- `uv run ruff --version` -> 0.15.14 (== pre-commit rev == CI's windowed resolution)
- `uv run ruff check . && uv run ruff format --check .` -> clean
- `uvx ruff@0.15.14 check .` -> All checks passed
- `uvx pre-commit run ruff-check/ruff-format --all-files` -> Passed
- `make py-check` -> lint, format, typecheck (713 files), lock all pass
- All modified modules import cleanly; `tests/_runtime/test_manage_script_metadata.py` -> 14 passed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,7 +39,7 @@ repos:
         types: [python]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.10
+    rev: v0.15.15
     hooks:
       # Run the linter
       - id: ruff-check
--- a/marimo/_ai/_tools/tools/cells.py
+++ b/marimo/_ai/_tools/tools/cells.py
@@ -24,7 +24,6 @@ from marimo._messaging.notification import (
 from marimo._types.ids import CellId_t, SessionId
 
 if TYPE_CHECKING:
-    from marimo._ast.models import CellData
     from marimo._session import Session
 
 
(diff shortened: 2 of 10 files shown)


kirangadhave:
didn't realize the ruff version was released recently.

Issue #9743, opened 2026-06-01
Go to definition does not look outside the current cell


stephenmac7:
### Describe the bug

Using `gd` in vim mode or right clicking and selecting "Go to definition" does not actually go to definition if it is in another cell.

Command-click does go to the correct cell.

I have tried enabling LSP in the settings UI as well, to no avail.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.8",
  "editable": false,
  "location": "/home/smcintosh/.cache/uv/archive-v0/gWcAR7b-sNwNUvJC/lib/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.17.0-1015-azure",
  "Processor": "x86_64",
  "Python Version": "3.12.3",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v24.16.0",
    "uv": "0.11.17 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.4.1",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.22.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pymdown-extensions": "10.21.3",
    "pyyaml": "6.0.3",
    "starlette": "1.2.1",
    "tomlkit": "0.15.0",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.48.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "mcp": "1.27.2",
    "pandas": "3.0.3",
    "pyarrow": "24.0.0",
    "python-lsp-ruff": "2.3.1",
    "python-lsp-server": "1.14.0",
    "ruff": "0.15.15"
  },
  "Experimental Flags": {}
}
```

</details>

akshayka:
Thanks for reporting this issue. This makes sense for us to do. We have some wiring in place like you mentioned, if you ctrl/cmd+click a variable, we actually do go to the definition. We should be able to leverage that functionality to fix `gd` and right click "go to definition". We are open to contributions that fix this issue.

omkar-334:
hi @akshayka i'd like to work on this

AnshManwani:
[...]

PR #9747, opened 2026-06-01
[codex] Fix reactive go-to-definition lookup

nkgotcode:

## Summary
- prevent current-cell lookups from falling back to the first matching local variable when resolving reactive variables
- keep local scope resolution for same-cell definitions and then fall through to notebook reactive variable lookup
- add coverage for jumping from a reactive variable use in one cell to its defining cell

## Root cause
`goToDefinition` checked same-cell definitions first, but `goToVariableDefinition` always fell back to the first same-cell name match. For a reactive variable used in another cell, that fallback treated the usage token as a local match and stopped before the notebook variable lookup could jump to the defining cell.

## Validation
- `pnpm --filter @marimo-team/frontend test src/core/codemirror/go-to-definition/__tests__/utils.test.ts --run`
- Result: 1 test file passed, 3 tests passed.

### DIFF
--- a/frontend/src/core/codemirror/go-to-definition/commands.ts
+++ b/frontend/src/core/codemirror/go-to-definition/commands.ts
@@ -413,12 +413,16 @@ export function goToVariableDefinition(
   view: EditorView,
   variableName: string,
   usagePosition?: number,
+  fallbackToFirstMatch = true,
 ): boolean {
   const { state } = view;
   const from =
     (usagePosition !== undefined
       ? findScopedDefinitionPosition(state, variableName, usagePosition)
-      : null) ?? findFirstMatchingVariable(state, variableName);
+      : null) ??
+    (fallbackToFirstMatch
+      ? findFirstMatchingVariable(state, variableName)
+      : null);
 
   if (from === null) {
     return false;
--- a/frontend/src/core/codemirror/go-to-definition/commands.ts
+++ b/frontend/src/core/codemirror/go-to-definition/commands.ts
@@ -408,6 +408,8 @@ function findScopedDefinitionPosition(
  * @param view The editor view which contains the variable name.
  * @param variableName The name of the variable to select, if found in the editor.
  * @param usagePosition The position of the variable usage, if available.
+ * @param fallbackToFirstMatch Whether to fall back to the first matching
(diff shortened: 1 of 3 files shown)


nkgotcode:

nkgotcode:
recheck

akshayka:
@kirangadhave this might be right, can you confirm?

kirangadhave:
@cubic-dev-ai

frontend/src/core/codemirror/go-to-definition/commands.ts
```diff
@@ -413,12 +413,16 @@ export function goToVariableDefinition(
   view: EditorView,
   variableName: string,
   usagePosition?: number,
+  fallbackToFirstMatch = true,
```
kirangadhave, line 419: jsdoc doesn't reflect the new param, default value and what it does

frontend/src/core/codemirror/go-to-definition/commands.ts
```diff
@@ -413,12 +413,16 @@ export function goToVariableDefinition(
...
 ): boolean {
   const { state } = view;
   const from =
     (usagePosition !== undefined
       ? findScopedDefinitionPosition(state, variableName, usagePosition)
-      : null) ?? findFirstMatchingVariable(state, variableName);
+      : null) ??
+    (fallbackToFirstMatch
+      ? findFirstMatchingVariable(state, variableName)
+      : null);
```
kirangadhave: ternaries with null coalescing operator is difficult to parse visually. let's simplify:

```ts
let from: number | null = null;
if (usagePosition !== undefined) {
  from = findScopedDefinitionPosition(state, variableName, usagePosition);
}
if (from === null && fallbackToFirstMatch) {
  from = findFirstMatchingVariable(state, variableName);
}
```

kirangadhave (changes requested):
two small followups

nkgotcode:
Addressed the remaining JSDoc feedback in `39b195b74`: the comment now describes scoped-definition lookup, the `fallbackToFirstMatch` behavior, and its default.

kirangadhave:
Addresses: https://github.com/marimo-team/marimo/issues/9743

kirangadhave (approved):
:rocket: Thanks @nkgotcode

PR #9750, opened 2026-06-02
Unify CSV and TSV export through the backend

kirangadhave:
## Summary

Unifies TSV export with CSV by routing both TSV download and copy through the backend, and refactors the shared `download_as` helper into a format registry. Previously TSV was computed on frontend.

## Changes

**Backend**
- Refactor `download_as` from a stringly-typed `if/elif` chain into a format registry (`_EXPORT_FORMATS`), where each format owns its file extension and writer callable. Per-format config moves into nested value objects (`DownloadOptions` with `DelimitedOptions` / `JsonOptions`), removing the flat per-format kwargs.
- Add `tsv` as a registry entry that shares the delimited writer with CSV, differing only by a forced tab separator.
- Widen `DownloadAsArgs.format` to include `tsv`.

**Frontend**
- Add `tsv` to the download RPC schema (zod enum and type union).
- Add TSV to the download menu and route TSV copy through the backend via `fetchText`.
- Delete the client-side `jsonToTSV` / `formatValueForTSV` and their tests.

## Notes
- `mo.ui.dataframe` shares `table.py`'s `DownloadAsArgs` and the same export menu, so it gains TSV automatically. `download_csv_separator` continues to configure CSV only; TSV always uses a tab.
- The registry is designed so future export formats (e.g. geodata) slot in as a single entry plus a writer.
- Markdown copy intentionally stays client-side: there is no backend markdown serializer and, unlike TSV, it has no locale issue.

Closes MO-5917

### DIFF
--- a/marimo/_plugins/ui/_impl/utils/dataframe.py
+++ b/marimo/_plugins/ui/_impl/utils/dataframe.py
@@ -1,7 +1,8 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-from typing import Any, TypeVar, Union
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, TypeVar, Union
 
 from narwhals.typing import IntoDataFrame, IntoLazyFrame
 
@@ -16,6 +17,11 @@ from marimo._runtime.context.types import (
     get_context,
 )
 from marimo._types.ids import UIElementId
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from marimo._runtime.virtual_file import VirtualFile
 
 LOGGER = _loggers.marimo_logger()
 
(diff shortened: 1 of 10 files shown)


kirangadhave:
@cubic-dev-ai

marimo/_plugins/ui/_impl/utils/dataframe.py
```diff
@@ -77,61 +81,127 @@ def get_default_csv_encoding() -> str:
...
-        # Use strict JSON to ensure compliance with JSON spec
-        payload = manager.to_json(
-            encoding=None, ensure_ascii=json_ensure_ascii, strict_json=True
-        )
-        vfile = mo_data.json(payload)
-    elif ext == "parquet":
-        payload = manager.to_parquet()
-        vfile = mo_data.parquet(payload)
-    else:
-        raise ValueError("format must be one of 'csv', 'json', or 'parquet'.")
```
Light2Dark, line 134: Any thoughts on keeping all the logic within this function? I think we can avoid the indirection & thus simplify the code if so.
```python
    if ext not in _DOWNLOAD_FORMATS:
        allowed = ", ".join(map(repr, _DOWNLOAD_FORMATS))
        raise ValueError(f"format must be one of {allowed}.")

    if ext in ("csv", "tsv"):
        separator = "\t" if ext == "tsv" else options.delimited.separator
        data = manager.to_csv(
            encoding=options.delimited.encoding or get_default_csv_encoding(),
            separator=separator,
        )
    elif ext == "json":
        # Use strict JSON to ensure compliance with JSON spec
        data = manager.to_json(
            encoding=None,
            ensure_ascii=options.json.ensure_ascii,
            strict_json=True,
        )
    elif ext == "parquet":
        data = manager.to_parquet()
    else:
        assert_never(ext)

    vfile = mo_data.any_data(data, ext=ext)
    base_name = filename if filename is not None else "download"
    return (vfile.url, f"{base_name}.{ext}")
```

marimo/_plugins/ui/_impl/utils/dataframe.py
```diff
@@ -77,61 +81,127 @@ def get_default_csv_encoding() -> str:
...
+
+_EXPORT_FORMATS: dict[str, _ExportFormat] = {
+    "csv": _ExportFormat("csv", _serialize_delimited(None)),
+    "tsv": _ExportFormat("tsv", _serialize_delimited("\t")),
+    "json": _ExportFormat("json", _serialize_json),
+    "parquet": _ExportFormat("parquet", _serialize_parquet),
+}
+
+
 def download_as(
```
Light2Dark, line 166: might be a big refactor so not necessary, but would be nice for avoiding potential misuse by adding `*`
```suggestion
def download_as(
    manager: TableManager[Any],
    ext: str,
	*,
    drop_marimo_index: bool = False,
    options: DownloadOptions | None = None,
    filename: str | None = None,
) -> tuple[str, str]:
```

marimo/_plugins/ui/_impl/utils/dataframe.py
kirangadhave: I wanted to abstract is away so adding new formats doesn't require adding conditionals. 

Adding to FORMATS should suffice. That way we abstract away file format handling from the download logic. I feel it's easier to read that way without having to parse conditionals mentally

Light2Dark: Personally, I think it's harder to read because it's abstracted too much. But I'm okay

Light2Dark (approved):

PR #9751, opened 2026-06-02
fix: arg/kwarg collision for local numpy vars in caching

dmadisetti:
## 📝 Summary

We do not properly detect local (defined in same cell) kwarg/arg defs, causing invalid collisions.

Came across this when doing some benchmarking.


kirangadhave (approved):
🚀

PR #9578, opened 2026-05-18
feat(sharing): enforce sharing config as server-side security

nojaf:
## 📝 Summary

Closes #6318

This PR makes the existing `sharing.wasm` and `sharing.html` configuration flags control what the UI surfaces, so users cannot accidentally share notebook source code through molab or WebAssembly links. Previously `sharing.wasm` only hid "Create WebAssembly link" — the "Create molab notebook" button was left fully ungated, and exported HTML files showed the "Open in molab" button regardless of config.

### What changed

**`marimo edit` (live server)**

- `Share → Create molab notebook` was not gated at all. It now respects `sharing.wasm = false` and is hidden alongside the "Create WebAssembly link" item.

**Exported HTML files (static and WASM)**

The sharing config is now baked into the `user_config` JSON embedded in exported HTML. The `StaticBanner` component reads `sharing.wasm` from this embedded config and conditionally hides the "Open in molab" button. This means a hosted WASM notebook or a shared static HTML file will not offer an easy one-click path to send source code to an external service when the flag is set.

**Important limitation:** hiding the molab button is a convenience measure, not a security control. A WASM HTML file always embeds the full source code (Pyodide requires it to run the notebook). Anyone who receives the file can read the source directly from the HTML, and can trivially reconstruct a molab URL themselves — the lz-string compression used in those URLs is a well-known open-source scheme, reversible with any standard lz-string library. Blocking `molab.marimo.io` at the network level prevents the link from being *opened* in molab but does not prevent the source from being extracted from the file and shared by other means.

### Configuration

**Per-project or per-user** — committed `pyproject.toml` or user config:

```toml
# pyproject.toml  — project-level, applies to all contributors
[tool.marimo.sharing]
wasm = false
html = false
```

```toml
# ~/.config/marimo/marimo.toml  — user-level, applies to all notebooks
[sharing]
wasm = false
html = false
```

#### The risk

Several marimo features encode the full notebook source code and make it trivially accessible outside the organisation:

- **"Create WebAssembly link"** and **"Create molab notebook"** compress the source with `lz-string` and embed it in a URL hash fragment (`https://marimo.app/#code/...`). The encoding is not encryption — it is a well-known open-source compression scheme that any developer can decode in seconds with freely available tools.
- **Hosted WASM notebooks** contain the source code and, without this PR, include an "Open in molab" button that sends the code to an external service with one click.

[...]

PR #9756, opened 2026-06-02
feat(sharing): add MARIMO_RESTRICT_SHARING env var machine-wide

nojaf:
## 📝 Summary

This is a follow up to #9578. That PR made the `sharing.wasm`, `sharing.html`, and `sharing.molab` config flags control what sharing affordances the UI surfaces, and added the `molab` flag. During #9578 the maintainers (@mscolnick via DM) asked that the `MARIMO_RESTRICT_SHARING` env var be split out into its own PR, so this is that separate change.

### What changed

- **`marimo/_config/settings.py`**: add `RESTRICT_SHARING` to `GlobalSettings`, reading the `MARIMO_RESTRICT_SHARING` env var.
- **`marimo/_config/manager.py`**: when the env var is set, `EnvConfigManager.get_config()` injects `sharing = {"wasm": False, "html": False, "molab": False}`. `EnvConfigManager` is the last (highest priority) partial in `get_default_config_manager`, so this override takes precedence over any per-project or per-user TOML config.
- **`tests/_config/test_manager.py`**: tests that the flag injects the all-false sharing config, that nothing is injected when the flag is off, and that the env override wins over a user config that explicitly enables sharing.

### Why we want this

The `sharing.*` config flags require every project or user to opt in. In our company we run marimo in managed devpod/container environments, and there is no reliable way to ensure every user has the flag set in their own config, since users have admin access to their own environment and can edit config files. `MARIMO_RESTRICT_SHARING=1` lets infra admins set the restriction once at the container/pod spec level, outside the user's control, and have every notebook session inherit it without per-project configuration.

### Scope and honest framing

This is a UI-hiding/policy control, not a server-side security boundary. #9578 deliberately removed the server-side 403 gates (`exporting != sharing`), so the sharing config only drives what the UI surfaces (the editor Share dropdown and the molab button baked into exported HTML). This env var hides those affordances machine-wide but does not block a determined user: exported HTML still embeds source and endpoints still serve code. We pair it with network egress filtering (blocking `marimo.app`, `molab.marimo.io`, `static.marimo.app`) for defence in depth. Note also that `MARIMO_RESTRICT_SHARING=0` as a command prefix overrides the env var for that process, so it should be set in the devpod/container spec rather than inside the container.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

//cc @Light2Dark @akshayka, @dmadisetti

### DIFF
--- a/marimo/_config/manager.py
+++ b/marimo/_config/manager.py
@@ -36,6 +36,7 @@ from marimo._config.secrets import (
     mask_secrets_partial,
     remove_secret_placeholders,
 )
+from marimo._config.settings import GLOBAL_SETTINGS
 from marimo._config.utils import (
     get_or_create_user_config_path,
 )
@@ -331,6 +332,13 @@ class EnvConfigManager(PartialMarimoConfigReader):
             ["runtime", "auto_instantiate"],
             project_config,
         )
+        if GLOBAL_SETTINGS.RESTRICT_SHARING:
+            # Highest-priority override: hide every external sharing affordance.
+            project_config["sharing"] = {
+                "wasm": False,
+                "html": False,
+                "molab": False,
+            }
         if hide_secrets:
             return mask_secrets_partial(project_config)
         return project_config
--- a/marimo/_config/manager.py
+++ b/marimo/_config/manager.py
@@ -66,6 +66,8 @@ def get_default_config_manager(
         ProjectConfigManager(current_path),
         ScriptConfigManager(current_path),
         EnvConfigManager(),
+        # Always merged last; see SecurityConfigManager.
+        SecurityConfigManager(),
     )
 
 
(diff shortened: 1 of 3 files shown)


dmadisetti:
Thanks! I think that a new `Restricted`/`Exfil`/`Security` ConfigManager would be a better fit than hardcoding the override in `MarimoConfigManager`. Adding it to `get_default_config_manager` should be sufficient

nojaf:
Thanks @dmadisetti, that makes sense. Happy to pull it out into a dedicated reader (`RestrictedConfigManager`) wired into `get_default_config_manager` instead of special-casing it in `MarimoConfigManager`.

One nuance I want to confirm before I push: the reason it currently lives in `get_config_overrides()` (the post-merge clamp) rather than in a single reader is that `with_overrides()` appends its partial *after* the existing ones, so a reader placed in `get_default_config_manager` could be re-enabled by a later `with_overrides({"sharing": {...: True}})`. There's a test pinning that ("a later override cannot re-enable sharing").

Two ways to keep that guarantee with the cleaner reader design:

1. Add `RestrictedConfigManager` as a normal partial *and* have `with_overrides()` keep it last (re-append it after the new override), so the "cannot be re-enabled" property and test still hold.
2. Treat it as a plain partial exactly as you described, and accept that a later `with_overrides()` can re-enable sharing, i.e. drop that guarantee/test, on the grounds that this is a UI-hiding policy control and not a server-side boundary anyway.

Do you have a preference? My lean is (1) since it keeps the machine-wide enforcement intent, but (2) is simpler and matches your comment most literally. Either way I'll move the logic into the new reader.

dmadisetti:
> Add RestrictedConfigManager as a normal partial and have with_overrides() keep it last (re-append it after the new override), so the "cannot be re-enabled" property and test still hold.

I think this ideal. We can maybe make it a "SecurityConfig" or something to indicate it should always go last + create the surface area for other required enforcements in the future.

dmadisetti (approved):
Test looks unrelated. Nice @nojaf !

We should release later today

PR #9762, opened 2026-06-02
fix: serialize kernel-defined functions by value for spawn multiprocessing

### LINKED ISSUE #9717: Symbols not visible to child processes, breaking multiprocessing
### Describe the bug

Attempting to execute a notebook-defined function in another process with multiprocessing fails, because Python's multiprocessing bootstrap can't find the function:

```
Process SpawnProcess-1:1:
  File "/Users/aagrawal/.local/share/uv/python/cpython-3.14-macos-aarch64-none/lib/python3.14/multiprocessing/process.py", line 320, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/Users/aagrawal/.local/share/uv/python/cpython-3.14-macos-aarch64-none/lib/python3.14/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aagrawal/.local/share/uv/python/cpython-3.14-macos-aarch64-none/lib/python3.14/concurrent/futures/process.py", line 242, in _process_worker
    call_item = call_queue.get(block=True)
  File "/Users/aagrawal/.local/share/uv/python/cpython-3.14-macos-aarch64-none/lib/python3.14/multiprocessing/queues.py", line 120, in get
    return _ForkingPickler.loads(res)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^
AttributeError: module '__mp_main__' has no attribute 'compute_square'
```

```python
import multiprocessing
import time
from concurrent.futures import ProcessPoolExecutor
import numpy as np

# Embarrassingly parallel workload: calculate square of numbers
def compute_square(n):
    # Simulate some work with a small delay
    time.sleep(0.001)
    return n * n

def run_parallel_workload(num_workers, data_size=100):
    """Run parallel workload with specified number of workers"""
    data = list(range(data_size))
    
    start_time = time.time()
    with ProcessPoolExecutor(max_workers=num_workers) as executor:
        results = list(executor.map(compute_square, data))
    end_time = time.time()
    
    return end_time - start_time

# Test with varying numbers of cores
max_cores = multiprocessing.cpu_count()
core_counts = list(range(1, max_cores + 1))[:4]
times = []

for num_cores in core_counts:
    print("Number of cores")
    elapsed_time = run_parallel_workload(num_cores)
    times.append(elapsed_time)
    print(f"Cores: {num_cores}, Time: {elapsed_time:.2f}s")

# Plot time vs number of cores
plt.figure(figsize=(10, 6))
plt.scatter(core_counts, times, color='blue', s=50)
plt.plot(core_counts, times, color='lightblue', linestyle='-', alpha=0.7)
plt.xlabel('Number of Cores')
plt.ylabel('Execution Time (seconds)')
plt.title('Execution Time vs Number of Cores for Parallel Workload')
plt.grid(True, alpha=0.3)
plt.xticks(core_counts)
plt.gca()
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

_No response_

desty:

## Problem

Running a notebook-defined function in a subprocess via `multiprocessing` / `ProcessPoolExecutor` fails with the `spawn` start method — the default on macOS and Windows, and on Linux from Python 3.14:

```
AttributeError: Can't get attribute 'compute_square' on <module '__mp_main__'>
```

marimo executes cells by `exec`-ing them in a synthetic module registered as `sys.modules["__main__"]`, so cell-defined functions get `__module__ == "__main__"`. The standard library pickles them *by reference* (`__main__.<name>`), but a spawned child is a fresh interpreter whose `__main__`/`__mp_main__` never contains them. Re-executing the notebook file in the child doesn't help either, since cell functions live inside `@app.cell def _(): ...` closures, not at module top level.

This is the behavior already captured by the `xfail` on `test_run_in_processpool_executor`.

## Fix

Register a `cloudpickle`-backed `reducer_override` on `multiprocessing.reduction.ForkingPickler` for functions/classes whose `__module__` is `"__main__"`, so they serialize *by value*. cloudpickle keeps genuinely importable objects by-reference, so nothing else changes.

Two non-obvious points:
- `ForkingPickler.register(...)` does **not** work here — pickle dispatches functions/classes through its built-in `save_global`, which bypasses the reducer dispatch table. `reducer_override` is consulted first, so it takes effect.
- The override is monkeypatched onto the existing class rather than swapping the class, so modules that bound `ForkingPickler` at import time (e.g. `multiprocessing.queues`) pick it up.

## cloudpickle dependency — question for maintainers

I added `cloudpickle` as an **optional** dependency (registered in `DependencyManager`, added to the `test-optional` group). When it's absent, `patch_multiprocessing_reduction()` is a no-op and behavior is unchanged. cloudpickle is already present transitively in `uv.lock`.

If you'd prefer spawn-based multiprocessing to work out of the box for everyone, cloudpickle would need to be promoted to a **core dependency** — glad to do that if you're OK with it. I left it optional to respect the lean dependency set; let me know which you prefer.

## Tests

- `test_run_in_processpool_executor`: the `xfail` now applies only when cloudpickle is **not** installed; with it installed (the `test-optional` environment), the test runs and passes on macOS / Windows / 3.14.
- New `TestMultiprocessingReduction.test_kernel_main_function_pickled_by_value`: verifies a `__main__`-defined function round-trips through `ForkingPickler` even when the receiving `__main__` lacks it (simulating a spawned child).

Verified locally on macOS (Python 3.13, spawn): the previously-failing test passes, `test_patches.py` is green, and `ruff` / `mypy` are clean.

Fixes #9717

### DIFF
--- a/marimo/_dependencies/dependencies.py
+++ b/marimo/_dependencies/dependencies.py
@@ -216,6 +216,7 @@ class DependencyManager:
     anywidget = Dependency("anywidget")
     traitlets = Dependency("traitlets")
     watchdog = Dependency("watchdog")
+    cloudpickle = Dependency("cloudpickle")
     ipython = Dependency("IPython")
     ipywidgets = Dependency("ipywidgets")
     nbformat = Dependency("nbformat")
--- a/marimo/_runtime/patches.py
+++ b/marimo/_runtime/patches.py
@@ -213,6 +213,62 @@ def create_main_module(
     return _module
 
 
+_PATCHED_MP_REDUCTION = False
+
+
+def patch_multiprocessing_reduction() -> None:
+    """Make objects defined in the kernel's `__main__` module picklable by value.
+
+    marimo runs cells by `exec`-ing them in a synthetic module that is
+    registered as `sys.modules["__main__"]`. Functions and classes defined this
+    way pickle fine within the kernel process, but the standard library pickles
+    them *by reference* (as `__main__.<name>`). When such an object is sent to a
+    subprocess started with the `spawn` method — the default on macOS and
+    Windows, and on Linux from Python 3.14 — the child interpreter has its own
+    empty `__main__`/`__mp_main__` and cannot resolve the name, raising
+    `AttributeError: Can't get attribute '<name>' on <module '__mp_main__'>`.
+
+    Registering a `cloudpickle`-backed reducer makes these objects serialize by
+    value so they survive the trip. cloudpickle keeps genuinely importable
+    objects as by-reference, so this does not change behavior for anything that
+    already pickles correctly across processes.
+
+    No-op if `cloudpickle` is not installed.
+    """
+    global _PATCHED_MP_REDUCTION
+    if _PATCHED_MP_REDUCTION:
+        return
+
+    if not DependencyManager.cloudpickle.has():
+        return
+
+    from multiprocessing.reduction import ForkingPickler
+
+    import cloudpickle  # type: ignore[import-untyped]
+
+    def _reducer_override(self: Any, obj: Any) -> Any:  # noqa: ARG001
+        # Functions and classes defined in a cell live in the synthetic
+        # __main__ module; route them through cloudpickle so spawned
+        # subprocesses can reconstruct them by value. Everything else falls
+        # through to the default machinery by returning NotImplemented.
+        #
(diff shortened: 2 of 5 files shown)


desty:

PR #9766, opened 2026-06-02
fix(runtime): don't shadow builtin print unless mo.Thread is used (#9765)

### LINKED ISSUE #9765: Marimo replaces builtin print, breaking Numba @njit compilation
### Describe the bug

Numba fails to compile functions using print() inside a marimo notebook.

Minimal example:
```
from numba import njit

@njit
def test():
    print(123)

test()
```

This raises:
```
Untyped global name 'print': Cannot determine Numba type of <class 'function'>
```

However, replacing print with builtins.print works:
```
import builtins
from numba import njit

@njit
def test():
    builtins.print(123)

test()
```

Inspection shows that marimo replaces the builtin print with a Python function wrapper:
```

import builtins

print(print)
print(builtins.print)
print(print is builtins.print)
```

Output:

```
<function print at 0x7fd99e334fe0>
<built-in function print>
False
```

This appears to break libraries such as Numba that special-case the builtin print.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.8",
  "editable": false,
  "location": "/home/ar/Work/back/.venv/lib/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "7.0.10-201.fc44.x86_64",
  "Processor": "",
  "Python Version": "3.12.12",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.22.2",
    "uv": "0.11.2 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.4.1",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.21.2",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.2.1",
    "tomlkit": "0.15.0",
    "uvicorn": "0.48.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "mcp": "1.27.2",
    "pandas": "3.0.3",
    "polars": "1.41.2",
    "python-lsp-ruff": "2.3.1",
    "ruff": "0.15.15"
  },
  "Experimental Flags": {
    "external_agents": true
  }
}
```

</details>

### Code to reproduce

_No response_

mscolnick:
marimo unconditionally replaced the builtin `print` in every cell's
global namespace with `print_override` (used to route mo.Thread output
to the right cell). This broke libraries that special-case the genuine
builtin, e.g. numba's `@njit`, which requires `print is builtins.print`
and otherwise fails with "Untyped global name 'print': Cannot determine
Numba type of <class 'function'>".

Only patch `print` when the feature that needs it is actually used:
stop installing the override in create_main_module, and instead have
mo.Thread install it lazily into the kernel globals on construction.
Thread-free notebooks now keep the real builtin print, so numba and
similar libraries work again.

Removes the now-unused print_override plumbing through KernelArgs and
create_main_module/patch_main_module.

Fixes #9765

### DIFF
--- a/marimo/_pyodide/pyodide_session.py
+++ b/marimo/_pyodide/pyodide_session.py
@@ -478,7 +478,6 @@ def _launch_pyodide_kernel(
             control_queue=control_queue,
             set_ui_element_queue=set_ui_element_queue,
             virtual_file_storage=None,
-            print_override_fn=None,
         )
     )
 
--- a/marimo/_runtime/app/kernel_runner.py
+++ b/marimo/_runtime/app/kernel_runner.py
@@ -129,7 +129,6 @@ class AppKernelRunner:
             module=create_main_module(
                 filename,
                 input_override=None,
-                print_override=None,
                 doc=extract_docstring_from_header(app._app._header),
             ),
             user_config=DEFAULT_CONFIG,
--- a/marimo/_runtime/app/script_runner.py
+++ b/marimo/_runtime/app/script_runner.py
@@ -85,7 +85,6 @@ class AppScriptRunner:
             create_main_module(
                 file=self.filename,
                 input_override=None,
-                print_override=None,
                 doc=self._docstring,
             )
         ) as module:
@@ -120,7 +119,6 @@ class AppScriptRunner:
             create_main_module(
                 file=self.filename,
                 input_override=None,
-                print_override=None,
                 doc=self._docstring,
             )
         ) as module:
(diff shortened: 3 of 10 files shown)


akshayka (approved):
TIL. Thank you

PR #9777, opened 2026-06-03
Fix async task output stream capture

### LINKED ISSUE #8816: Stdout in async task is ignored?

nkgotcode:
## Summary

Fixes #8816.

This routes stdout and stderr through task-local stream targets when async tasks are active, so output emitted from scheduled async work is captured by the owning execution context instead of escaping to the process-level streams.

## Tests

- `../marimo/.venv/bin/python -m pytest tests/_runtime/test_async_task_stdout_repro.py tests/_messaging/test_thread_local_proxy.py tests/_runtime/test_redirect_streams.py tests/_messaging/test_streams.py -q`
  - `43 passed, 1 warning in 2.26s`
- `../marimo/.venv/bin/python -m pytest tests/_runtime/test_runtime.py::TestExecution::test_runtime_name_error_reference_caught -q`
  - `2 passed in 0.76s`

### DIFF
--- a/marimo/_messaging/streams.py
+++ b/marimo/_messaging/streams.py
@@ -7,6 +7,7 @@ import os
 import sys
 import threading
 from collections import deque
+from contextvars import ContextVar
 from typing import (
     TYPE_CHECKING,
     Protocol,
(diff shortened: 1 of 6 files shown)


mscolnick:
@cubic-dev-ai

PR #9780, opened 2026-06-03
feat(table): styled tooltip for mo.ui.table hover_template

kirangadhave:
## Summary
Replaces the native HTML `title` tooltip used by `mo.ui.table`'s `hover_template` with marimo's existing radix styled tooltip, for both modes: the row-level string template (`"{{first}} {{last}}"`) and the per-cell callable. Hover text is now consistent with the rest of the table UI and can support richer content in the future.

## Approach
A single **controlled** radix tooltip per table, not one per cell. A small hook tracks `{ rect, content }` of the hovered cell (piggybacking the existing cell `onMouseOver`), and one positioned, `pointer-events-none` trigger is anchored to that rect. Cost is O(1) in tooltip components regardless of cell count, which avoids the per-cell wrapping perf concern.

## Changes
- New `data-table/hover-tooltip/`: `content.ts` (template resolution, cell-level wins over row-level), `use-table-hover-tooltip.ts` (hover state, 400ms delay, drag/scroll suppression, focus/blur parity), `hover-tooltip.tsx` (the single controlled tooltip).
- `renderers.tsx`: removed the two native `title=` props and the inline `applyHoverTemplate`; wired the hook and render `<HoverTooltip>` once.

Screenshot 2026-06-03 at 5 01 19 PM

Closes MO-5746

### DIFF
--- a/frontend/src/components/data-table/__tests__/data-table.test.tsx
+++ b/frontend/src/components/data-table/__tests__/data-table.test.tsx
@@ -142,6 +142,39 @@ describe("DataTable", () => {
     expect(screen.getAllByText("per-cell tip").length).toBeGreaterThan(0);
   });
 
+  it("links the focused cell to the tooltip content for assistive tech", () => {
+    const testData: TestData[] = [{ id: 1, name: "Test 1" }];
+    const columns: ColumnDef<TestData>[] = [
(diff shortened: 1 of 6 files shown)


kirangadhave:
@cubic-dev-ai

dmadisetti (approved):

PR #9781, opened 2026-06-03
fix(frontend): coalesce outgoing document transactions to avoid intra-batch conflicts

kirangadhave:
## Problem

Debounced cell edits are buffered into a single document transaction. When an edit and a delete of the same cell land in one debounce window (e.g. type in a cell, Escape, then Shift-Backspace), the flushed batch contains both a `set-code` and a `delete-cell` for that cell. The server applies a transaction atomically and rejects it as an intra-transaction conflict, surfacing the "Failed to sync document transaction" toast and dropping the change.

Error popups:
image

## Fix

Add a pure `coalesceChanges` reconciliation pass in `document-changes.ts`, run in `flushChanges` before the batch is sent, so the transaction reflects net effect rather than edit history:

- For any cell deleted in the batch, drop its property/move changes (`set-code`, `set-name`, `set-config`, `move-cell`). The `create-cell` and `delete-cell` are kept so anchors referencing the cell stay resolvable server-side; create+delete is a net no-op.
- Collapse repeated `set-code`/`set-name`/`set-config` for a surviving cell to the last occurrence.

Both reductions preserve order. The pass is pure and unit-tested independently of the flush wiring.

## Tests

- Unit tests for `coalesceChanges` covering each reduction and anchor-safety.
- A `flush path coalescing` test that drives the real debounced flush and asserts the same-window edit+delete batch sends only the `delete-cell`, while cross-window edits send two clean transactions.

Closes MO-6335

### DIFF
--- a/frontend/src/core/cells/__tests__/document-changes.test.ts
+++ b/frontend/src/core/cells/__tests__/document-changes.test.ts
@@ -16,14 +16,32 @@
 import { python } from "@codemirror/lang-python";
 import { EditorState } from "@codemirror/state";
 import { EditorView } from "@codemirror/view";
-import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from "vitest";
 import { cellId } from "@/__tests__/branded";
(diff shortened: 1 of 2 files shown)


kirangadhave:
@cubic-dev-ai

mscolnick (approved):
nice find, i kept seeing this.

PR #9782, opened 2026-06-04
fix(messaging): stop reusable hint flicker on cell updates

kirangadhave:
## Problem

The reusable-definition hint (`?` / `reusable` badge on a top-level `def`/`class` cell) flickered whenever the cell received any update, and vanished permanently under lazy execution.

Root cause: `serialization` was the one `CellNotification` field that did not follow the partial-update convention. The frontend assigned it unconditionally, and the backend sent `serialization=None` on every queued/running/idle message. So each status update wiped the hint; under eager execution the post-execution broadcast re-set it (the flicker), and under lazy execution the cell never re-ran so the hint never came back. The reconnect/replay merge dropped it for the same reason.

## Fix

Adopt a three-state contract for the hint, with `msgspec.UNSET` as the "unchanged" sentinel so `None` is free to mean "clear":

- `msgspec.UNSET` (omitted on the wire) -> leave the hint unchanged
- `None` (explicit `null`) -> clear the hint (cell is no longer a top-level def)
- string -> set the hint

The frontend mirrors this as `undefined` (unchanged) vs `null` (clear) vs value. The post-execution hook now explicitly clears the hint when a cell stops being a top-level definition, and the session-view merge inherits the prior hint on `UNSET`, fixing the reconnect path too.

To clear the hint when a cell stops being a top-level def, we need to know if it was showing one before, otherwise every normal cell run would send a redundant clear. We track this on the `DirectedGraph` by cell id rather than `CellImpl`, since `CellImpl` is replaced when a cell's code changes, which is the exact transition we need to catch.

## Tests

- `cell.test.ts`: `transitionCell` keeps the hint when `serialization` is omitted, clears it on `null`, sets it on a string.
- `test_session_view.py`: the reconnect merge keeps the old hint on `UNSET` and applies a `None` clear.
- `test_runtime.py`: editing a top-level def into a plain assignment sends one clear; normal cell runs send none.

Follow-up tracked in MO-6367 to migrate remaining partial-update fields to the same convention.

Closes MO-6336


dmadisetti (approved):
🚀

PR #9794, opened 2026-06-05
Right/Left cell outputs

koaning:
This is my version 0 of what could be support for having the output of a cell render on the right.

CleanShot 2026-06-05 at 17 12 29@2x

I made this PR with the intent to easily share this with people, but there's a few things we should double check before considering merging this.

1. Having this for the right side is cool, but if you're used to a language where you read from right to left instead of left to right, it could also make sense to render this on the left.
2. Holding cells based on H1/H2 headers in markdown is currently broken.
3. Doing this makes the cells a bunch smaller to work from if there is output. But cells without output, do we want to have those be normal width?

---

## Follow-up updates

Since the prototype, the layout has been hardened (some of these fixes were made with the help of a coding agent):

- **Left output support** — output can now render on the `left` as well as the `right`. Both options are labelled `(experimental)` in settings.
- **Reader / kiosk view** — left/right output layouts now render correctly in reader and kiosk modes.
- **Side-by-side sizing** — removed the editor height clamp so the editor column matches the output, and stretched the code editor gutter height.
- **Hover-bridge fixes (addresses #4)** — the floating action rails (run/shoulder buttons) and the maximize/expand controls now stay reachable when moving the pointer toward them:
  - In side-by-side layouts the hover gutters are hosted on the full-width `.cell-columns` row instead of the single editor column, so they no longer cover the output or swallow pointer events near the resizer.
  - In stacked (above/below) layouts the output region gets its own right-edge hover bridge, so the maximize/expand controls remain reachable when hovering alongside the output (not just at the editor's height).

### Still open
- H1/H2 header-based cell stacking in markdown (concern #2).
- Whether output-less cells should keep full width (concern #3).

### DIFF
--- a/frontend/src/components/editor/actions/useConfigActions.tsx
+++ b/frontend/src/components/editor/actions/useConfigActions.tsx
@@ -165,6 +165,30 @@ export function useConfigActions() {
         });
       },
     },
+    {
+      label: "Config > Set cell output area: left",
+      hidden: config.display.cell_output === "left",
+      handle: () => {
+        handleUserConfig({
+          display: {
+            ...config.display,
+            cell_output: "left",
+          },
+        });
+      },
+    },
+    {
+      label: "Config > Set cell output area: right",
+      hidden: config.display.cell_output === "right",
+      handle: () => {
+        handleUserConfig({
+          display: {
+            ...config.display,
+            cell_output: "right",
+          },
+        });
+      },
+    },
   ];
 
   return actions.filter((a) => !a.hidden);
--- a/frontend/src/components/editor/actions/useConfigActions.tsx
+++ b/frontend/src/components/editor/actions/useConfigActions.tsx
@@ -1,7 +1,11 @@
 /* Copyright 2026 Marimo. All rights reserved. */
 
 import { useAppConfig, useResolvedMarimoConfig } from "@/core/config/config";
-import type { AppConfig, UserConfig } from "@/core/config/config-schema";
+import {
+  type AppConfig,
+  CELL_OUTPUT_POSITIONS,
+  type UserConfig,
+} from "@/core/config/config-schema";
 import { getAppWidths } from "@/core/config/widths";
 import { useRequestClient } from "@/core/network/requests";
 import type { ActionButton } from "./types";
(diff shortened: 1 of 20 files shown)


Light2Dark:
I think we should allow "left" | "right" ?
Looks good though!

koaning:
@Light2Dark I found an odd bug. Notice how my mouse cursor doesn't turn into a click pointer when it is too close to the center seperator? It's consistent across the scatterwidget and the dataframe one. 

https://github.com/user-attachments/assets/4f9fd21f-48cc-4c20-b904-7b0ff4a4cb1b

frontend/src/components/editor/actions/useConfigActions.tsx
```diff
@@ -165,6 +165,30 @@ export function useConfigActions() {
         });
       },
     },
+    {
+      label: "Config > Set cell output area: left",
```
mscolnick: this could likely be DRYed but with a `.map`

frontend/src/components/editor/notebook-cell.tsx
```diff
@@ -455,7 +481,17 @@ const EditableCellComponent = ({
...
+
+  // Side-by-side doesn't make sense for markdown cells: the output is just the
+  // rendered source, so a split view is redundant. Fall back to the stacked
+  // "below" layout (editor on top, rendered preview underneath).
+  const cellOutput =
+    isMarkdown &&
+    (configuredCellOutput === "left" || configuredCellOutput === "right")
+      ? "below"
+      : configuredCellOutput;
+  const isSideBySide = cellOutput === "left" || cellOutput === "right";
```
mscolnick: duplicate code? from vertical-cell layout

koaning:
Spotted an issue I think. 

Normally, a headers have a property that let you collapse. 

CleanShot 2026-07-02 at 22 38 17@2x

CleanShot 2026-07-02 at 22 38 21@2x

But in this branch this behaviour seems gone. Also for normal above/below cells. 

CleanShot 2026-07-02 at 22 38 30@2x

@Light2Dark I think the collapse icon move to the right. Was this intentional? 

CleanShot 2026-07-02 at 22 46 53@2x

koaning:
Closing this work after having a longer discussion with team-members. It's a neat idea, but it just has *so much* surface area that it is not worth it right now.

PR #9797, opened 2026-06-05
refactor: replace bulk-mutation helpers with diff Transactions

kirangadhave:
Routes the `with_data` save round-trip through `NotebookDocument.apply(Transaction)` and removes the last two non-transactional mutation helpers: `CellManager._replace_state_from` and `NotebookDocument._replace_cells`.

After this change, every structural mutation of a notebook document flows through one path: diff the old and new cell state into a `Transaction`, then `apply` it. There is no longer a bulk-replace side door that bypasses the document state machine.

## Why

The two bulk helpers reassigned the cell list wholesale and unconditionally bumped the document version, even for no-op saves. That diverged from the transactional path, which only bumps when something actually changes and keeps per-cell versions stable across format-on-save round trips (so an agent's prior read stays valid). Collapsing onto `apply(Transaction)` makes the version semantics consistent and gives a single, auditable mutation path.

## How

- Extract the cell-diff logic out of `file_manager.py` into a shared `marimo/_ast/cell_diff.py` (`build_transaction`), so `with_data` (in `_ast`) can reuse it without an `_ast` -> `_session` import.
- `with_data` now rebuilds a `CellManager` from the textual fields, diffs it against the live one, and applies the resulting transaction. State the document doesn't track (`_compiled_cells`, `unparsable`, `seen_ids`) is carried over at the call site.
- `reload()` (file-watch) calls the same shared `build_transaction`.
- Delete `_replace_state_from` and `_replace_cells` and migrate their tests onto the transactional path.

Closes MO-6039


marimo/_ast/app.py
```diff
@@ -1052,7 +1053,19 @@ def with_data(
                 config=config,
                 cell=prev_compiled.get(cell_id),
             )
-        cm._replace_state_from(rebuilt)
+
+        transaction, _ = build_transaction(
```
dmadisetti: should this maybe be on the cell manager itself?, tied to it with prev

marimo/_ast/app.py
```diff
@@ -1052,7 +1053,19 @@ def with_data(
...
+        )
+        cm.document.apply(transaction)
+
+        # _compiled_cells, unparsable, and seen_ids aren't in the document;
+        # copy them too. Mutate _compiled_cells in place so its holders keep
+        # the dict; union seen_ids so a deleted id isn't reused.
+        cm._compiled_cells.clear()
+        cm._compiled_cells.update(rebuilt._compiled_cells)
+        cm.unparsable = rebuilt.unparsable
+        cm._cell_id_generator.seen_ids |= rebuilt._cell_id_generator.seen_ids
```
dmadisetti: maybe wrapped into `build_and_apply_transaction`? Seems like this logic shouldn't be decoupled

marimo/_ast/app.py
kirangadhave: so something like `cm.build_transaction...`?

marimo/_ast/app.py
kirangadhave: yeah that makes sense. the logic is also duplicated in one other place.

dmadisetti (approved):
Thanks for the cleanup!

PR #9708, opened 2026-05-28
add pagination & remote search flow for remote storage

Light2Dark:
## 📝 Summary

Fixes https://github.com/marimo-team/marimo/issues/9662. This PR introduces a "Load more" button, that when clicked fetches more entries under the remote storage panel.

It also surfaces this button when a prefix search produces no entries, eg. folder/filename. A bug with Google Drive's file id has been fixed too.

https://github.com/user-attachments/assets/ab3ed8f6-4583-45e3-b307-a4c1dfe36079

## 📋 Pre-Review Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it.
- [x] Video or media evidence is provided for any visual changes (optional). 

## ✅ Merge Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Tests have been added for the changes made.

Issue #9662, opened 2026-05-22
Remote storage limits fo view


ivan-balashov:
### Describe the bug

I have a private aws s3 bucket with lots of elements. when I addit to marimo - it only shows the first N elements (I assume due to head operation) and I cannot select/explore other elements; (I see the folder started from A to N, and cannot see elements starts from W).

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.6",
  "editable": false,
  "location": "/home/jovyan/new_project/.venv/lib/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.12.83-113.160.amzn2023.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.12.8",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "0.7.8"
  },
  "Dependencies": {
    "click": "8.4.0",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.21.2",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pymdown-extensions": "10.21.3",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.15.0",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.47.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "polars": "1.40.1"
  },
  "Experimental Flags": {
    "external_agents": true
  }
}
```

</details>

### Code to reproduce

Add s3 bucket with lots of elements or prefix with lots of elements(I could not share  my bucket)

Light2Dark:
Thanks for reporting. This is because we did not implement proper pagination flow yet, so we only pull N items. But this is something I'd like to add soon. Involves frontend changes to `load more` and backend changes to store the pagination token.

Just to clarify, I believe you're talking about the remote storage feature. And it only lists N number of items. But if you did expand a folder within the list, you can see more items right? If not, then that's a bug.

ivan-balashov:
[...]

PR #9834, opened 2026-06-09
feat(storage): paginate remote storage listings with a "Load more" button

Light2Dark:
## 📝 Summary

Split out from #9708. Part of the work for https://github.com/marimo-team/marimo/issues/9662.

Adds pagination to remote storage listings:

- Threads a `page_token` through the storage backends, the `StorageListEntriesCommand`, and the `StorageEntriesNotification`. `list_entries` now returns a `StorageListResult` (entries + `next_page_token` + `may_have_more`).
- Surfaces a "Load more" button in the storage inspector that fetches and appends the next page, plus a "More files may exist" hint when a provider truncates a listing it cannot page through.

https://github.com/user-attachments/assets/0ec9e6e3-8228-48c2-b21a-2769b22a3dba

## 📋 Pre-Review Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it.
- [x] Video or media evidence is provided for any visual changes (optional). 

## ✅ Merge Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] Documentation has been updated where applicable, including docstrings for API changes.
- [x] Tests have been added for the changes made.

PR #9835, opened 2026-06-09
feat(storage): search the backend when no loaded entries match

Light2Dark:
## 📝 Summary

Split out from #9708. Part of the work for https://github.com/marimo-team/marimo/issues/9662.

When a prefix query produces no matches among the already-loaded entries, this pages through the backend to surface remote results:

- Entries are now matched by full path (not just basename), so partial path queries like `folder/x` resolve `folder/xsomething`. Because object stores evaluate prefixes on a path-segment basis, the parent directory is listed and filtered client-side.
- When there are no loaded matches, the panel offers a "Search more entries" action (the button or pressing Enter) that pages through the backend up to 5 pages per trigger, stopping early on a match or when the listing is exhausted. If it reaches the cap without a match, it shows "Continue searching" to fetch the next batch. Status rows reflect the search state (searching / found / no matches / capped / error).

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/frontend/src/components/storage/storage-inspector.tsx
+++ b/frontend/src/components/storage/storage-inspector.tsx
@@ -1040,7 +1040,10 @@ export const StorageInspector: React.FC = () => {
             value={searchValue}
             onValueChange={setSearchValue}
             onKeyDown={(event) => {
-              if (event.key === "Enter" && hasSearch) {
+              if (
+                event.key === "Enter" &&
+                namespaces.some(canContinueRemoteSearch)
+              ) {
                 event.preventDefault();
                 continueRemoteSearches();
               }
--- a/frontend/src/components/storage/storage-inspector.tsx
+++ b/frontend/src/components/storage/storage-inspector.tsx
@@ -43,10 +43,12 @@ import {
   useStorage,
   useStorageActions,
   useStorageEntries,
+  useStoragePageFetcher,
 } from "@/core/storage/state";
 import type {
   StorageEntry,
   StorageNamespace,
+  StoragePageMetadata,
   StoragePathKey,
 } from "@/core/storage/types";
 import { storagePathKey } from "@/core/storage/types";
@@ -94,6 +96,10 @@ function displayName(path: string): string {
   return parts[parts.length - 1] || trimmed;
 }
 
+function directoryPrefix(path: string): string {
+  return path.endsWith("/") ? path : `${path}/`;
+}
+
 /**
  * Stable, unique identity for an entry row. Prefer the
  * backend's stable id when present and fall back to the list index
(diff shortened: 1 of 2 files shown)


frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -680,6 +1039,12 @@ export const StorageInspector: React.FC = () => {
             className="h-6 m-1"
             value={searchValue}
             onValueChange={setSearchValue}
+            onKeyDown={(event) => {
+              if (event.key === "Enter" && hasSearch) {
+                event.preventDefault();
+                continueRemoteSearches();
+              }
+            }}
```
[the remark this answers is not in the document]
Light2Dark: Good catch — fixed in 113a4f007. The Enter handler now gates on `namespaces.some(canContinueRemoteSearch)` instead of `hasSearch`, so it only calls `preventDefault()` when a backend search will actually run; slashless queries fall through to CmdK's default selection.

frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -602,7 +788,29 @@ const StorageNamespaceSection: React.FC<{
               />
             );
           })}
-          {hasMore && (
+          {showRemoteResults &&
+            filteredRemoteEntries.map((entry) => {
+              const rowKey = storageEntryKey(
+                entry,
+                remoteEntries.indexOf(entry),
+              );
```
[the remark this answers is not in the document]
Light2Dark: Keeping `remoteEntries.indexOf(entry)` here intentionally. Using the `filteredRemoteEntries` map index reintroduces the key-instability we fixed earlier (#9834): for id-less entries the fallback key is `path::index`, and the *filtered* index shifts as the query narrows, remounting rows and dropping their `isExpanded` state. Keying off the position in the unfiltered `remoteEntries` keeps keys stable and mirrors the two main row blocks (`children.indexOf` / `entries.indexOf`). The O(n²) is negligible for realistic listing sizes; if it ever matters we can precompute a position map once across all three blocks.

frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -523,6 +629,103 @@ const StorageNamespaceSection: React.FC<{
     searchValue,
     entriesByPath,
   });
+  const searchPrefix = remoteSearchPrefix(searchValue);
+  const searchKey = storagePathKey(namespaceName, searchPrefix);
+  const remoteEntries =
+    searchPrefix === "" ? [] : (entriesByPath.get(searchKey) ?? []);
+  // The fetched page is the whole parent directory; we still need to filter
+  // it by the full search query before showing entries to the user.
+  const filteredRemoteEntries =
```
kirangadhave, line 638: why not just call `filterEntries`? filtering on empty array should return empty anyways.

frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -617,11 +825,162 @@ const StorageNamespaceSection: React.FC<{
...
+      const prefixMatches = filterEntries(prefixEntries, {
+        namespace: namespaceName,
+        searchValue: currentQuery,
+        entriesByPath,
+      });
+      if (prefixMatches.length > 0) {
+        return false;
+      }
+
+      const canFetchPrefix =
```
kirangadhave, line 902: We have a `function canSearchMoreRemoteEntries` which does the same check right?

frontend/src/components/storage/storage-inspector.tsx
```diff
@@ -523,6 +629,103 @@ const StorageNamespaceSection: React.FC<{
...
+            Continue searching
+          </Button>
+          <span className="text-[10px]">(or press Enter)</span>
+        </span>
+      );
+    }
+    if (remoteSearch.status === "exhausted" && !hasLoadedMatches) {
+      return "No matches";
+    }
+    if (!hasSearch && !isPending && entries.length === 0 && !error) {
```
kirangadhave, line 705: can combine with above as one predicate?

kirangadhave (approved):
🚀 Looks good, just some nits

frontend/src/components/storage/storage-inspector.tsx
Light2Dark: thanks! will address these in follow-up

PR #9839, opened 2026-06-09
feat: threading and multiprocessing adapters in WASM

peter-gy:
<details>

<summary>Original POC PR description</summary>

Drafting this as a POC on stubbing threading and multiprocessing in WASM notebooks.
This first slice adds a minimal threading patch for Pyodide:

- bootstrap WASM runtime patches before marimo imports runtime modules that capture
`threading.local`
- patch the small `threading` surface needed for the POC: `Thread`, `Event`, `local`, current thread, ident, enumerate/count
- give started WASM threads a synthetic identity
- make `threading.local` use that identity
- repair marimo runtime context storage if it was imported before the patch
- let `mo.Thread` hand async targets back to the patched runner

The main thing to review is the lifecycle: install early, install once, keep one interpreter-wide state object, and make marimo runtime context follow the synthetic thread identity.

## Review order

Probably easiest to read in this order:

1. `marimo/__init__.py` and `marimo/_pyodide/bootstrap.py`
2. `marimo/_runtime/_wasm/__init__.py`
3. `marimo/_runtime/_wasm/_concurrency/_install.py`
4. `marimo/_runtime/_wasm/_concurrency/_state.py`
5. `marimo/_runtime/_wasm/_concurrency/_threading.py`
6. `marimo/_runtime/_wasm/_concurrency/_wait.py`
7. `marimo/_runtime/threads.py`
8. `tests/_runtime/test_wasm_threading_poc.py`

## WASM Demo

https://github.com/user-attachments/assets/ea5e671a-8dc6-4830-adf0-79c9604e4e50

## Not in this PR / Follow-up Work

Leaving these out for now, but the intended approaches are:

- `ThreadPoolExecutor`: build on the same synthetic thread identity model, with
submitted callables running through asyncio tasks and returning normal-looking
`Future`s.
- `concurrent.futures.wait` / `as_completed`: adapt future completion to the Pyodide
wait bridge so timeout and completion behavior stay close to stdlib call sites.
- `multiprocessing.Process`: keep the process API callable in Pyodide, but execute the
target in the current interpreter and report lifecycle state through `start`, `join`,
`is_alive`, and `exitcode`.
- `multiprocessing.Queue`: use an in-memory queue implementation that works with the
process-shaped APIs above.
- `multiprocessing.Pool`: map pool work onto the same current-interpreter execution
model, mostly for common notebook/library code that imports `Pool`.
- `ProcessPoolExecutor`: layer futures over the process-shaped runner, so code using
the API can run without crashing in Pyodide.
- stream proxy repair: if stdout/stderr proxies captured a pre-patch `threading.local`,
swap them to the WASM local and sync state on restore.
- browser matrix / Pyodide acceptance tests: run the examples in a real WASM notebook
path, including `mo.Thread`, stdlib threading, futures, and multiprocessing-shaped
code.
- docs and lint-rule wording: explain which APIs marimo patches in WASM and which APIs
still have Pyodide-specific semantics.

</details>

Belongs to PR family of #9413 and #9480.

This PR makes the common Python concurrency APIs callable in WASM notebooks by installing browser-backed adapters for threading, futures, and process-shaped multiprocessing. The adapters run inside the current Pyodide interpreter on the browser event loop while preserving the API shape and lifecycle semantics that notebook and library code commonly expects.

## Summary

Adds WASM adapters for:

- `threading.Thread`, `threading.Event`, `threading.local`, current thread identity, and thread enumeration
- `concurrent.futures.ThreadPoolExecutor`, `Future.result`, callbacks, `wait`, and `as_completed`
- process-shaped `multiprocessing.Process`, `Queue`, `SimpleQueue`, and `Pool`
- `concurrent.futures.ProcessPoolExecutor`
- `mo.Thread` output routing from WASM workers back to the spawning cell

The branch also updates WASM lint rules and docs so supported process-shaped APIs are no longer flagged as unavailable, while we still call out unsupported native process primitives.

## Runtime model

As per the POC (e1c9570), all work stays in one Pyodide interpreter. Started threads, executor workers, and process-shaped targets are scheduled onto the browser-backed asyncio loop with synthetic thread and process identities.

Blocking-looking waits use the [Pyodide JSPI bridge](https://blog.pyodide.org/posts/jspi/) when needed, so calls like `thread.join()`, `Event.wait()`, `Future.result()`, `wait()`, `as_completed()`, `Queue.get()`, and pool result waits can yield to browser runtime work instead of failing immediately.

APIs that require OS processes, shared memory, pipes, managers, fork/forkserver contexts, or native synchronization remain unsupported.

### Support Levels

WASM notebooks run Python inside one browser-hosted Pyodide interpreter. Some concurrency APIs can preserve their Python API shape there, but their execution model differs from server-backed CPython. This PR introduces four support levels to make those differences explicit:

- `api-compatible`: the tested API shape and result behavior match the local Python contract for that operation.
- `serialized`: the API shape is available, but submitted work runs one task at a time in the current Pyodide interpreter.
- `cooperative-only`: waits, cancellation, and termination progress when Python yields back to Pyodide's event loop. Already-running Python code is not preempted.
- `blocked`: marimo rejects the API because the browser cannot provide the native process, synchronization, or shared-memory primitive it requires.

## Changes

- Installs WASM concurrency patches during Pyodide bootstrap before marimo runtime modules capture thread-local state.
- Keeps one interpreter-wide WASM runtime state for synthetic thread identity, current process ownership, live work tracking, and shutdown.
- Adds a serialized WASM executor for `ThreadPoolExecutor` and process-pool-shaped APIs.
- Adds process-shaped lifecycle support for `multiprocessing.Process`: `start`, `join`, `is_alive`, `exitcode`, `current_process`, `parent_process`, and `active_children`.
- Adds in-memory `Queue` and `SimpleQueue` adapters that preserve the standard put/get shape inside the browser interpreter.
- Adds `multiprocessing.Pool` support for `apply`, `map`, `starmap`, `imap`, async results, callbacks, close/join, and termination semantics that make sense in Pyodide.
- Adds `ProcessPoolExecutor` over the same process-shaped executor layer.
- Repairs pre-imported marimo runtime context storage and stdout/stderr stream proxies so `mo.Thread` output and progress updates route back to the correct cell.
- Extends WASM lint tests and docs to distinguish supported adapters from APIs that still cannot work in Pyodide.

## Testing

- Added runtime coverage for WASM threading, futures, multiprocessing process, queue, pool, and process pool adapters
- Added Pyodide browser matrix coverage for the public concurrency surfaces
- Added stream and message-type coverage for WASM thread output routing
- Added lint coverage for supported and unsupported WASM multiprocessing APIs

## Demo

Outputs from [this test notebook](https://github.com/user-attachments/files/28926946/notebook.py):

Screenshot 2026-06-14 at 13 51 12

Screenshot 2026-06-14 at 13 51 29

Screen Recording 2026-06-14 at 13 52 32

Screenshot 2026-06-14 at 13 53 57

Screenshot 2026-06-14 at 13 54 09

Screenshot 2026-06-14 at 13 54 21

Screenshot 2026-06-14 at 13 54 32

Screenshot 2026-06-14 at 13 54 45

Screenshot 2026-06-14 at 13 54 58

Screenshot 2026-06-14 at 13 55 12

## Follow-ups

Having all the above primitives introduced in this PR, we could add support for:

- `threading.Timer`, `cooperative-only`: can be implemented as a delayed `AsyncioThread` scheduled on the Pyodide event loop. It does not need native threads, only delayed cooperative execution plus `cancel()` before the callback runs.

- `multiprocessing.Event`, `cooperative-only`: we can map it to the existing `AsyncEvent` shape already used for `threading.Event` and queue waits. This would cover common library code that just needs `set`, `clear`, `is_set`, and `wait`.

- `multiprocessing.Lock` / `RLock`, `cooperative-only`: we could add same-interpreter locks backed by asyncio state and JSPI waits.

- `multiprocessing.Semaphore` / `BoundedSemaphore`, `cooperative-only`: we can build on the same cooperative wait primitive as queues.  Acquisition would only progress when Python yields to Pyodide.

- `multiprocessing.Condition`, `cooperative-only`: we could layer over the cooperative lock plus waiter list. This is mechanically straightforward once Lock/RLock exist and would have browser-event-loop scheduling semantics.

- `multiprocessing.JoinableQueue`, `serialized`: we could extend the in-memory queue with unfinished-task accounting, `task_done()`, and `join()`. Since `Queue` is already same-interpreter, this is probably the cleanest multiprocessing follow-up.

- `multiprocessing.Pipe`, `serialized`: can be implemented as paired in-memory endpoints with `send`, `recv`, `poll`, and `close`.

- `multiprocessing.pool.ThreadPool`, `serialized`: could be just aliased to the existing serialized executor/Pool machinery.

- `Process.sentinel`, `cooperative-only`: we can expose a synthetic wait token for adapted processes to support simple completion checks.

- Mixed pending `concurrent.futures.wait` / `as_completed`, `cooperative-only`: we can wrap observable foreign futures with callbacks and keep the current error for futures that cannot be observed without blocking the Pyodide event-loop lane.

### DIFF
--- a/marimo/_runtime/_wasm/_concurrency/_threading.py
+++ b/marimo/_runtime/_wasm/_concurrency/_threading.py
@@ -21,6 +21,7 @@ from marimo._runtime._wasm._concurrency._state import (
     create_task_in_empty_wasm_context,
     current_ident,
     current_identity,
+    current_process_owner,
     current_thread,
     current_thread_var,
     get_event_loop,
@@ -254,6 +255,8 @@ class AsyncioThread(ThreadIdentity, metaclass=AsyncioThreadMeta):
         self._done_future: asyncio.Future[None] | None = None
         self._task: asyncio.Task[None] | None = None
         self._exception: BaseException | None = None
+        self._suppress_excepthook_for: tuple[type[BaseException], ...] = ()
+        self._wasm_process_owner: Any | None = None
 
     @property
     def daemon(self) -> bool:
@@ -272,6 +275,7 @@ class AsyncioThread(ThreadIdentity, metaclass=AsyncioThreadMeta):
             self._ident = new_ident()
             self._native_id = self._ident
             self._started = True
+            self._wasm_process_owner = current_process_owner()
             loop = get_event_loop()
             self._done_future = loop.create_future()
             live_threads.add(self)
@@ -299,10 +303,12 @@ class AsyncioThread(ThreadIdentity, metaclass=AsyncioThreadMeta):
                 await result
         except asyncio.CancelledError as exc:
             self._exception = exc
-            self._call_excepthook(exc)
+            if not isinstance(exc, self._suppress_excepthook_for):
+                self._call_excepthook(exc)
         except BaseException as exc:
             self._exception = exc
-            self._call_excepthook(exc)
+            if not isinstance(exc, self._suppress_excepthook_for):
+                self._call_excepthook(exc)
         finally:
             current_thread_var.reset(token)
             self._finish()
--- a/marimo/_runtime/_wasm/_concurrency/_threading.py
+++ b/marimo/_runtime/_wasm/_concurrency/_threading.py
@@ -141,12 +141,22 @@ class AsyncLocal:
         return None
 
     def __setattr__(self, name: str, value: Any) -> None:
+        if name == "__dict__":
+            raise AttributeError(
+                f"{type(self).__name__!r} object attribute "
+                f"{name!r} is read-only"
+            )
         if name == "_storage" or self._data_descriptor(name) is not None:
             object.__setattr__(self, name, value)
             return
         self._namespace()[name] = value
 
     def __delattr__(self, name: str) -> None:
+        if name == "__dict__":
+            raise AttributeError(
+                f"{type(self).__name__!r} object attribute "
+                f"{name!r} is read-only"
+            )
         if self._data_descriptor(name) is not None:
             object.__delattr__(self, name)
             return
--- a/marimo/_runtime/_wasm/_concurrency/_threading.py
+++ b/marimo/_runtime/_wasm/_concurrency/_threading.py
@@ -1,0 +1,359 @@
+# Copyright 2026 Marimo. All rights reserved.
+"""Map a minimal `threading` surface to Pyodide asyncio tasks.
+
+The patch keeps `threading.Thread`, `threading.Event`, and `threading.local`
+callable in WASM notebooks. It does not create OS threads. Each started thread
+receives a synthetic identity so `threading.local` and marimo runtime context
+storage stay isolated.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import inspect
+import threading as _threading
+import traceback
+import weakref
+from typing import TYPE_CHECKING, Any, cast
+
+from marimo._runtime._wasm._concurrency._state import (
+    ThreadIdentity,
+    create_task_in_empty_wasm_context,
+    current_ident,
+    current_identity,
+    current_thread,
+    current_thread_var,
+    get_event_loop,
+    live_threads,
+    new_ident,
+    new_thread_name,
+    patch_state,
+    run_until_complete_in_empty_wasm_context,
+)
+from marimo._runtime._wasm._concurrency._wait import (
+    cooperative_wait,
+    wait_for_future,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable
+
+
+_ASYNC_LOCALS: weakref.WeakSet[AsyncLocal] = weakref.WeakSet()
+
+
+def clear_thread_local_state(ident: int) -> None:
+    """Remove per-thread storage when a synthetic thread finishes."""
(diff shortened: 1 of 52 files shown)


marimo/_runtime/_wasm/_concurrency/_threading.py
```diff
@@ -0,0 +1,359 @@
...
+        if self._finished or self._done_future is None:
+            return
+        if timeout is not None and timeout <= 0:
+            return
+        cooperative_wait(wait_for_future(self._done_future, timeout))
+
+    def is_alive(self) -> bool:
+        return self._started and not self._finished
+
+    def getName(self) -> str:
```
mscolnick, line 365: why the camelCase?

peter-gy: They mirror deprecated CPython `Thread` aliases. We can drop if we want to support only the modern api surface.

mscolnick (approved):
this is impressive and awesome. quite a large change, but i dont have any comments. its pretty isolated enough that if we need to revert or extract it to its own package, it could do so easily. 

the test coverage seems great and this is already a huge step above the existing behavior. we can test this out with some real cases

Issue #5102, opened 2025-05-27
Enhance mo.ui.multiselect


Yasin197:
### Describe the bug

using the mo.ui.multiselect widget it is hard to keep track of all the elements which was selected especially if the dropdown is long.

1. Typically, you should have a list just above the dropdown which contains all the elements which was selected (maybe slightly different highlighting too) 
 
2. when you hover over the element it should also show you which elements which was selected.

with these enhancements it should improve the UX greatly! 

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
Replace this line with the output of marimo env. Leave the backticks in place.
```

</details>

Yasin197:
unfortunately i won't be able to as i am not good with the front-end codebase. I raised this feature enhancement because this is the behavior we have in our in house strategic applications in our UIs which is a "must have" for greater adoption

Yasin197:
@mscolnick yes - i have something like this already as a work around, do you think this feature would make it into a release in the future?

mscolnick:
@Yasin197 - i think it could. most definitely with a community contribution.

yan-hic:
@mscolnick I agree multiselect could support "combo-style" ie. like multi feature of https://react-select.com/home, where user can see and deselect easily the selected entries instead of a generic "# selected". As per OP, if the list is long, it's cumbersome to scroll and remember which entries are selected.
Another cell can show the values, but doesn't help with deselecting.

Could be either an enhancement to multiple or a new UI component. Am not a ts/js developer but as workaround, can react-select be wrapped in an anywidget ?

PR #9849, opened 2026-06-11
refactor(select): shared select-core for multiselect, dropdown, and filter picker

### LINKED ISSUE #9388: mo.ui.multiselect: add bulk "Select/Deselect matching" selector on top "Select/Deselect all"
### Description

When a `multiselect` has hundreds of options, there's no way to bulk-select search results. You search, then click items one by one. Hitting Enter selects *all* options, not just the *matching* ones.

### Suggested solution

Add a `match_actions` parameter, defaulting to `False` so existing behavior is unchanged.

```python
mo.ui.multiselect(options=big_list, match_actions=True)
```

With `match_actions=True`, two actions appear in the dropdown while the user is typing a search:

* **Select matching** — adds all currently filtered options to the selection
* **Deselect matching** — removes them

When the search has text, matching items are displayed below Select matching / Deselect matching / Select all / Deselect all (\*). When the search is empty, only already-selected items are shown under Select all / Deselect all.

(\*) We could agree to hide Select all / Deselect all when search has text to simplify the UI.

Image

I have a working implementation (Python + React + tests) and I'm happy to open a PR if you're interested. Quick summary of the changes:

* **Python** (`input.py`): new `match_actions: bool = False` parameter on `multiselect.__init__`, passed as `match-actions` to the frontend
* **React** (`MultiselectPlugin.tsx`):
  * Pure functions `selectMatching()` / `deselectMatching()` / `getVisibleOptions()` extracted for testability
  * `ComboboxItem` entries shown conditionally when `matchActions && searchQuery`
  * Idle dropdown filtered to selected items when `matchActions` is on
* **Tests**: Python (`test_input.py`) and TypeScript (`MultiSelectPlugin.test.ts`) — all passing

- [X] Yes

### Alternatives

* **anywidget**: built a custom JS widget replicating the multiselect — it works but can't match the native look and feel. I think this belongs in the main `mo.ui.multiselect` widget anyway.

### Additional context

Related: [MO-2929](https://linear.app/marimo/issue/MO-2929/enhance-mouimultiselect) — similar scope but more focused on one specific interaction pattern.

kirangadhave:
## What

Extracts a shared `components/ui/select-core/` module and migrates the multiselect, dropdown (SearchableSelect), and data-table top-K filter picker onto it. The three widgets previously reimplemented their own search, selection, pinning, and bulk-action logic; they now share a headless hook and a renderer.

### New module: `ui/select-core`

- `useSelectList` — headless hook composed from four file-private hooks (search, pinning, toggle, bulk). Owns search query, filtered options, toggle semantics, freeze-on-open pinning, and search-scoped bulk select/deselect with cap-aware row config. Pinning and bulk actions are opt-in.
- `SelectList` — renderer that wires the hook's outputs into Command / checkbox markup, with `data-slot` hooks and render slots for customization, an optional compact-chip trigger, and a virtualized path above a row threshold.
- `utils` — generic selection helpers (`getVisibleOptions`, `getBulkActions`, `selectMatching` / `deselectMatching`, smart match).
- `BulkAction` is a discriminated union (`select-all` / `deselect-all` / `select-matching` / `deselect-matching`).

### New features
- `mo.ui.multiselect`: selected items now pin to the top of the dropdown, and the menu gains bulk **Select/Deselect all** (idle) and **Select/Deselect N matching** (while searching) actions scoped to the current search. The trigger shows selections as compact chips instead of an "N selected" count.
- `mo.ui.dropdown`: re-picking the current value now clears it (replacing the old `--` row).

### Migrations

- **Multiselect** becomes a thin `SelectList` wrapper. Pinning, freeze-on-open, the compact-chip trigger, search-scoped bulk rows, and `max_selections` enforcement are inherited from the core.
- **Dropdown** renders `SearchableSelect` through the core, dropping the `NONE_KEY` / `--` sentinel row in favor of the core's `allowSelectNone` clear-to-null gate (re-picking the current value clears it).
- **Filter picker** (data-table top-K) drives its search query and filtered rows through `useSelectList`; its own Command/checkbox/count markup, toggle-all-visible, and creatable path are unchanged.
- `multiselectFilterFn` and the old `data-table/value-chips` location are removed; `value-chips` now lives under `ui/`.

## Why

Three widgets had drifted into three slightly different implementations of the same select behavior. Consolidating onto one headless hook + renderer removes the duplication, makes selection semantics consistent across the widgets, and gives a single tested place for pinning and bulk actions.

https://github.com/user-attachments/assets/f5c1b334-c080-469a-a673-283636213ece

## Testing

- New unit tests for `useSelectList` and `select-core/utils`.
- Updated dropdown and multiselect plugin tests for the migrated paths.
- `make fe-check` passes.

Closes https://github.com/marimo-team/marimo/issues/9388, https://github.com/marimo-team/marimo/issues/5102


mscolnick (approved):

frontend/src/components/ui/select-core/types.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface Option<V = string> {
+  /** Selection identity; what the adapter serializes. */
+  value: V;
+  /** Human-readable string used for display, filtering, and chip text. */
+  label: string;
+  /** Whether the option can be selected. */
+  disabled?: boolean;
+  /** Arbitrary per-row payload (e.g. `{ count }`) read by render slots. */
+  data?: unknown;
+}
```
Light2Dark, line 13: could we do a pass to reduce comments like these

kirangadhave: sorry i missed this. I'll do a follow up

PR #9859, opened 2026-06-11
Fix async task output stream capture

### LINKED ISSUE #8816: Stdout in async task is ignored?

nkgotcode:
## Summary

Fixes #8816.

This routes stdout and stderr through task-local stream targets when async tasks are active, so output emitted from scheduled async work is captured by the owning execution context instead of escaping to the process-level streams.

## Tests

- `uv run --group test pytest tests/_runtime/test_async_task_stdout_repro.py tests/_runtime/test_threads.py tests/_runtime/test_redirect_streams.py -q`
  - `15 passed in 1.45s`

### DIFF
--- a/marimo/_messaging/streams.py
+++ b/marimo/_messaging/streams.py
@@ -208,28 +208,6 @@ class ThreadSafeStream(Stream):
                 self.console_msg_cv.notify()
 
 
-def _forward_os_stream(
-    standard_stream: Stdout | Stderr, fd: int, should_exit: threading.Event
-) -> None:
-    """Watch a file descriptor and forward it to a stream object."""
-
-    # This coarse try/except block silences exceptions; a raised exception
-    # at this point could cause bad errors, such as an infinite stream of data
-    # to be written to the fd/routed through the stream.
-    #
-    # TODO(akshayka): Make this loop bomb-proof, so that exceptions raised are
-    # exceptions we actually want to pay attention to; then store the exception
-    # and print it to the terminal later (outside an execution context).
-    try:
-        while not should_exit.is_set():
-            data = os.read(fd, 1024)
-            if not data:
-                break
-            standard_stream.write(data.decode())
-    except Exception:
-        ...
-
-
 class _NoOpWatcher:
     """A dummy watcher that does nothing, used when fd redirection is off."""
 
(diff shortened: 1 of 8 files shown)


marimo/_messaging/streams.py
```diff
@@ -107,6 +108,11 @@ def __init__(
         cell_id: CellId_t | None = None,
     ):
         self.pipe = pipe
+        self._cell_id = cell_id
```
dmadisetti: `cell_id` is already set?

dmadisetti: I suppose you could remove the other instead, but do we even need the fallback? When is the context var not going to be set?

marimo/_messaging/streams.py
```diff
@@ -145,6 +151,24 @@ def write(self, data: KernelMessage) -> None:
...
+        self._get_cell_id_context().set(cell_id)
+
+    def _get_cell_id_context(self) -> ContextVar[CellId_t | None]:
+        if not hasattr(self, "_cell_id_context"):
+            self._cell_id_context = ContextVar(
+                "marimo_stream_cell_id",
+                default=None,
+            )
+        return self._cell_id_context
```
dmadisetti: This is done in `__init__`, I think we can remove this

pyproject.toml
```diff
@@ -158,6 +158,8 @@ test = [
     "pytest-picked>=0.5.1",
     "pytest-sugar~=1.1.1",
     "pytest-xdist~=3.5",
+    # For formatter-backed tests
+    "ruff>=0.15.16",
```
dmadisetti: A bit of scope drift, I'd remove changes in this file

marimo/_messaging/thread_local_streams.py
dmadisetti, line 118: Can we make this a context manager instead? mutating `sys` with no guaranteed undo is less than ideal

dmadisetti (changes requested):
@nkgotcode looks reasonable at first pass thanks! But I had a number of stylistic comments and questions

Could you also fix your merge res issues? Appreciate it!

dmadisetti:
Actually refer to the original post, will provide an update- this may already be fixed

PR #9867, opened 2026-06-11
fix(table): recover from transient "function not found" in app mode

### LINKED ISSUE #9763: Sporadic "could not find function given request" in mo.ui.table
### Describe the bug

During the interaction with marimo tables in app mode, there are sporadic issues with any of the following:
- switching pages
- filtering
- sorting

There would be a message like below in a red frame:

```
Could not find function given request: InvokeFunctionCommand(id=48b59cd5-5626-4264-8e76-c99cf6fee2fd, fn=SFPL-0.calculate_top_k_rows)
```

The table cannot be used until the page is refreshed.  Usually it works fine after refresh.

This has been happening for at least a year.  Sometimes the error comes on the very first attempt to use UI, sometimes not.  Overall, I'd say about 5-10% of the time.  There is no rhyme or reason to it.  Sometimes it's just 20 rows and I can't get to the second page.  I cannot say with confidence if it does or does not happen in edit mode.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.6",
  "editable": false,
  "location": ".venv/lib/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "5.4.17-2136.336.5.1.el8uek.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.12.13",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.22.3",
    "uv": "0.1.24"
  },
  "Dependencies": {
    "click": "8.3.3",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.21.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.15.0",
    "uvicorn": "0.46.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.1.0",
    "anywidget": "0.11.0",
    "duckdb": "1.5.2",
    "loro": "1.10.3",
    "openai": "2.36.0",
    "pandas": "2.3.3",
    "pyarrow": "24.0.0",
    "pytest": "9.0.3",
    "python-lsp-ruff": "2.3.1",
    "ruff": "0.15.13",
    "sqlglot": "30.8.0",
    "watchdog": "6.0.0"
  },
  "Experimental Flags": {
    "multi_column": true
  }
}```

</details>


### Code to reproduce

_No response_

### PR DESCRIPTION
**Pull request by kirangadhave**
## Problem
In app/run mode, table interactions (page, sort, filter) sporadically fail with a terminal red frame: `Could not find function given request: ...calculate_top_k_rows`. The table stays unusable until a full page refresh. Could not reproduce the issue locally

This PR does not attempt the root-cause fix. It makes the symptom recoverable and adds the server-side signal needed to identify the trigger.

## What changed
- A "function not found" response means the function never executed, so re-issuing the request is side-effect-safe regardless of whether the function is idempotent. At the single funnel where plugins invoke kernel functions, a `found === false` result is now retried with bounded exponential backoff (150/300/600ms), re-reading the object-id on each attempt so a corrected id from a re-render is picked up. After retries are exhausted it surfaces a soft `FunctionNotFoundError` rather than the red frame. An execution error (`found === true`, the function ran and raised) is never retried and keeps its current behavior.

- The kernel already computed whether the function was found but discarded it before notifying the frontend, which left the client to infer the case by string-matching a status title. `found: bool` is now carried on `FunctionCallResultNotification`.

- The not-found branch now logs a structured warning (pid, notebook, requested namespace and function, whether the namespace was registered, and the set of registered namespaces). No argument values or user data are logged.

## Follow-up
The root-cause fix (startup-send gating, re-run id drift, multi-worker session affinity) is intentionally deferred until the new log identifies which trigger is happening in the field.

Closes #9763
Closes MO-6350

### DIFF
--- a/marimo/_runtime/kernel_request_handlers.py
+++ b/marimo/_runtime/kernel_request_handlers.py
@@ -167,13 +167,14 @@ class KernelRequestHandlers:
     async def _handle_function_call(
         self, request: InvokeFunctionCommand
     ) -> None:
-        status, ret, _ = await self._kernel.function_call_request(request)
+        status, ret, found = await self._kernel.function_call_request(request)
         LOGGER.debug("Function returned with status %s", status)
         broadcast_notification(
             FunctionCallResultNotification(
                 function_call_id=request.function_call_id,
                 return_value=ret,
                 status=status,
+                found=found,
             ),
         )
 
--- a/frontend/src/plugins/core/__test__/registerReactComponent.test.ts
+++ b/frontend/src/plugins/core/__test__/registerReactComponent.test.ts
@@ -283,7 +283,7 @@ describe("plugin function: not-found recovery", () => {
     expect(requestMock).toHaveBeenCalledTimes(4);
   });
 
-  test("does not retry an execution error (found:true)", async () => {
+  test("does not retry once the function is found (found:true)", async () => {
     const run = await mountWithFunction();
     requestMock.mockResolvedValue({
       function_call_id:
--- a/frontend/src/plugins/core/__test__/registerReactComponent.test.ts
+++ b/frontend/src/plugins/core/__test__/registerReactComponent.test.ts
@@ -3,10 +3,19 @@
 import ReactDOM from "react-dom/client";
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 import { z } from "zod";
+import { FUNCTIONS_REGISTRY } from "@/core/functions/FunctionRegistry";
+import type { FunctionCallResultMessage } from "@/core/kernel/messages";
+import { FunctionNotFoundError } from "@/utils/errors";
 import {
   isCustomMarimoElement,
   registerReactComponent,
 } from "../registerReactComponent";
+
+vi.mock("@/core/functions/FunctionRegistry", () => ({
+  FUNCTIONS_REGISTRY: { request: vi.fn() },
+}));
+
+const requestMock = vi.mocked(FUNCTIONS_REGISTRY.request);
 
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: marimo/_runtime/kernel_request_handlers.py [hunk 1]
```diff
@@ -167,13 +167,14 @@ async def _handle_receive_model_message(
     async def _handle_function_call(
         self, request: InvokeFunctionCommand
     ) -> None:
-        status, ret, _ = await self._kernel.function_call_request(request)
+        status, ret, found = await self._kernel.function_call_request(request)
```
dmadisetti, line 170: Oh looks like we even had this behavior wired up

dmadisetti (dismissed):

marimo/_runtime/kernel_request_handlers.py
kirangadhave: yeah we used it in `runtime.py` to search when we search for a function via current kernel or it's children's registry.

dmadisetti (approved):

liquidcarbon:
@kirangadhave thanks for your work on this.
Looks like this did not go into 0.23.10?
When it's released, will the logging you implemented go into regular logs, or only to debug?

PR #9876, opened 2026-06-13
fix(data): skip chart for nested/unknown column types

kirangadhave:
## Problem

Hovering a Polars `Struct` or `List` column in the datasets panel surfaced a confusing error and logged a spurious warning. These dtypes resolve to a `FieldType` of `"unknown"`, but chart generation still ran the full pipeline and failed deep in Altair's dataframe-interchange serializer with `ValueError: Unexpected DtypeKind: Struct(...)`. The caller caught the exception, logged a `WARNING` (which reads like a real failure), and shipped the raw serializer error to the frontend as `ColumnPreview.error`, as if charting had genuinely broken.

## Fix

Short-circuit `_get_altair_chart` for `"unknown"` field types, returning no chart and no error. The dtype is known to be unchartable up front, so there is nothing to attempt and nothing to report. The check lives right after the field-type lookup that the function already performs, so it adds no extra work. The stats path is on a separate branch and is unaffected, so aggregation stats still compute and display for these columns.

Before:
Screenshot 2026-06-12 at 6 57 31 PM

After:
Screenshot 2026-06-12 at 6 57 51 PM

Closes MO-6157

### DIFF
--- a/marimo/_data/preview_column.py
+++ b/marimo/_data/preview_column.py
@@ -235,6 +235,11 @@ def _get_altair_chart(
 
     (column_type, _external_type) = table.get_field_type(column_name)
 
+    # Nested/unknown dtypes (e.g. Polars Struct/List) can't be serialized
+    # through the dataframe interchange path, so skip charting silently
+    if column_type == "unknown":
+        return None, None, None, None
+
     if stats.total == 0:
         return None, None, "Table is empty", None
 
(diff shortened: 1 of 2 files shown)


mscolnick (approved):

marimo/_data/preview_column.py
```diff
@@ -235,6 +235,11 @@ def _get_altair_chart(
 
     (column_type, _external_type) = table.get_field_type(column_name)
 
+    # Nested/unknown dtypes (e.g. Polars Struct/List) can't be serialized
+    # through the dataframe interchange path, so skip charting silently
+    if column_type == "unknown":
+        return None, None, None, None
```
Light2Dark, line 241: Should we return a "No chart available" message. There may be users confused why there isn't a chart here.

Light2Dark (approved):

PR #9877, opened 2026-06-13
fix: preserve multiline string indentation in cell loader

### LINKED ISSUE #9851: Cell loader dedents lines inside multiline string literals — runtime value differs from Python semantics
## Bug

When a cell defines a multiline (triple-quoted) string whose content lines are indented, marimo's file loader dedents the cell body by the function indent **including lines inside the string literal**. The runtime value under `marimo run` / `marimo edit` then differs from what Python itself would produce when importing the same file.

## Minimal repro (marimo 0.19.11, python 3.14)

`nb.py`:

```python
import marimo

__generated_with = "0.19.11"
app = marimo.App()


@app.cell
def _():
    TEXT = """line0
  two_spaces
    four_spaces
      six_spaces"""
    return (TEXT,)
```

```python
import importlib.util
from marimo._ast.app import InternalApp

spec = importlib.util.spec_from_file_location("nb", "nb.py")
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)
cell = next(c for c in InternalApp(mod.app).cell_manager._cell_data.values()
            if "TEXT" in (c.code or ""))
ns = {}; exec(cell.code, ns)
print(repr(ns["TEXT"]))
# marimo:  'line0\n  two_spaces\nfour_spaces\n  six_spaces'
# python:  'line0\n  two_spaces\n    four_spaces\n      six_spaces'
```

Lines with a 4-space prefix lose exactly those 4 spaces (`    four_spaces` → `four_spaces`, `      six_spaces` → `  six_spaces`); lines with fewer keep theirs. `python nb.py` / importing the module gives the original string.

## Real-world impact

We embed Lean 4 proof source in notebook cells and send it to a verification API ([lean-marimo-quantum](https://github.com/Englund-Garage/lean-marimo-quantum)). Lean is indentation-sensitive: tactic lines indented ≥4 spaces arrived at the kernel at column 0 and failed with `expected '{' or indented tactic sequence` — but only when served via `marimo run`; the same file imported by Python verified fine. Confusing to debug because the divergence is invisible in the source. Anything indentation-sensitive (YAML, Python source strings, Makefiles, Lean) hits this.

Workaround we adopted: keep multiline literals in a plain imported module rather than in cells.

## Expected

Cell extraction should preserve string-literal content byte-for-byte (dedent decisions based on AST token positions rather than raw line prefixes), or `marimo check`/docs should at least warn about indented multiline strings in cells.

Implementation found by Claude (Anthropic) in collaboration with @dirkenglund.

ChidiebereNjoku:
**Description:**

Closes #9851

## Problem

When a cell defines a multiline (triple-quoted) string with indented content,
marimo's cell loader strips leading whitespace from lines inside the string
literal. This mutates user data — the runtime value of the string differs from
what Python itself produces when importing the same file.

**Repro:**
```python
@app.cell
def _():
    TEXT = """line0
  two_spaces
    four_spaces
      six_spaces"""
    print(TEXT)
    return (TEXT,)
```

**Before (broken):**
```
line0
  two_spaces
four_spaces      ← wrong, 2 spaces stripped from string content
  six_spaces
```

**After (fixed):**
```
line0
  two_spaces
    four_spaces
      six_spaces
```

## Root Cause

`textwrap.dedent` and the `fixed_dedent` helper in `parse.py` are indentation-blind — they strip leading whitespace from every line uniformly, including lines that are content inside multiline string literals. Python's semantics require that indentation inside string literals is part of the string's value and must never be touched.

Additionally, a stray `import tokenize` (module) on line 43 of `compiler.py` shadowed the `from tokenize import TokenError, tokenize` (function) on line 15, causing a `TypeError: 'module' object is not callable` in `ends_with_semicolon`.

## Solution

Introduce `marimo/_ast/dedent.py` with a token-aware `smart_dedent` function. It uses Python's `tokenize` module to identify lines that fall inside multiline string literals and marks them as protected. Only unprotected (code) lines participate in the base indentation calculation and stripping.

## Changes

- **`marimo/_ast/dedent.py`** (new): `smart_dedent` and `_get_protected_lines` — token-aware dedent that preserves string literal whitespace
- **`marimo/_ast/compiler.py`**: replace `textwrap.dedent` with `smart_dedent` in `cell_factory`; remove stray `import tokenize` that caused `TypeError`
- **`marimo/_ast/parse.py`**: replace `textwrap.dedent` in `fixed_dedent` with `smart_dedent`; preserve existing `refill` logic for AI-generated inconsistent indentation but skip `refill` on protected (string-content) lines
- **`tests/_ast/test_parse.py`**: update `test_fixed_dedent` snapshot to reflect corrected behaviour

## Pre-Review Checklist

- [ ] Documentation has been updated where applicable.
```

<!-- This is an auto-generated description by cubic. -->
<a href="https://cubic.dev/pr/marimo-team/marimo/pull/9877?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg"><img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"></picture></a>
<!-- End of auto-generated description by cubic. -->

### CODE REVIEW

**Comment by ChidiebereNjoku** [comment]

**Comment by mscolnick** [comment]
@cubic-ai

File: marimo/_ast/dedent.py [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+    """Return a bool list marking lines inside multiline string literals."""
+    lines = code.splitlines(keepends=True)
+    n = len(lines)
+    protected = [False] * n
+    try:
+        tokens = list(tokenize.generate_tokens(io.StringIO(code).readline))
+    except TokenError:
+        return protected
+    for tok_type, tok_string, tok_start, tok_end, _ in tokens:
+        if tok_type == tokenize.STRING and "\n" in tok_string:
```
mscolnick: this might not worth for f-strings

mscolnick:
can we add some tests for this specific function? i dont think this works on f-strings currently (hard to tell from just reading the existing tests)

marimo/_ast/parse.py
```diff
@@ -56,22 +56,31 @@ def ast_parse(
 
 def fixed_dedent(text: str) -> str:
```
dmadisetti: Can you move this to dedent while we're here?

dmadisetti (commented):
Thanks!

My largest concern is that this isn't exactly what the user sees in front end, but I understand how it's maybe useful for script mode comparison. We do rewrite the AST, but I wonder if one way conversions would be better and _only_ if the line falls under the standard indent.

Happy to be wrong, but it does seem like our current doc strings will get formatted weird and round tripping is not supported. What happens to:

```python
"""Cell to compute sum
e.g. a=1; b=2 \imples c=3
foo
"""
a + b
```

Under this scheme it should become

```python
def _():
    """Cell to compute sum
e.g. a=1; b=2 \imples c=3
foo
"""
    a + b
```

That seems a little counter intuitive and not currently in code. Conversely

 ```python
def _():
    """Cell to compute sum
    e.g. a=1; b=2 \imples c=3
    foo
    """
    a + b
```

becomes

 ```python
"""Cell to compute sum
    e.g. a=1; b=2 \imples c=3
    foo
    """
a + b
```

So maybe we only trigger in the following case:

```python
def _():
    """Cell to compute sum
  e.g. a=1; b=2 \imples c=3 <--- under indented!
    foo
    """
    a + b
```

```python
"""Cell to compute sum
  e.g. a=1; b=2 \imples c=3 <--- under indented!
    foo
    """
a + b
```


---

Can we also throw in a test for top level functions? e.g.

```python
@app.function
def foo():
    """I wonder what
    Happens to this ?
    """
    return ...
```

```python
@app.function
def bar():
    """Also wonder what
 Happens to that ?
    """
    return ...
```

ChidiebereNjoku:
@mscolnick Thanks for flagging this. f-strings work correctly because Python's tokenize module classifies f"""...""" as a STRING token just like regular triple-quoted strings, the token type is the same regardless of the f, r, or b prefix. I'll add explicit tests for f-strings and top-level function docstrings to make this clear.

ChidiebereNjoku:
@dmadisetti Good catches man. I'll move fixed_dedent into dedent.py to consolidate all dedent logic in one place. On the docstring concern, you're right that a docstring whose content lines start at column 0 (less indented than the function body) is a valid edge case. smart_dedent handles it correctly: those lines are marked protected and emitted byte-for-byte, so they won't be touched. I'll add a test covering this case.

ChidiebereNjoku:
@mscolnick  You were right. Python 3.12+ tokenizes f-strings as FSTRING_START/FSTRING_MIDDLE/FSTRING_END instead of a single STRING token, so they weren't being protected by the original code. Fixed _get_protected_lines to handle both cases. Added explicit tests for f-strings (test_fstring_interior_preserved, test_fstring_preserved) and top-level function docstrings as requested. All 14 new tests pass.

ChidiebereNjoku:
@dmadisetti Done, fixed_dedent is now in dedent.py alongside smart_dedent and _get_protected_lines. Also added test_docstring_content_at_column_zero which confirms that docstring content lines less indented than the function body are correctly preserved byte-for-byte.

dmadisetti:
Woops. I just broke this PR trying to merge res. Sorry this fell off my plate, for reference hit review to to let people know you're ready

image

ChidiebereNjoku:
> Woops. I just broke this PR trying to merge res. Sorry this fell off my plate, for reference hit review to to let people know you're ready
> 
> image

No worries. All conflicts resolved and CI is green again. Ready for re-review whenever you get a chance. @dmadisetti

ChidiebereNjoku:
All Copilot suggestions addressed.
blank line check, \r/\r\n newline detection via regex, docstring escaping, and the missing _NEWLINE_RE constant. All 104 tests passing locally. Ready for re-review. @mscolnick @dmadisetti

Thanks.

tests/_ast/test_parse.py
```diff
@@ -557,7 +560,7 @@ def foo():
     # Multiline string preserved
     assert fixed_dedent('    x = """hello\n    world"""') == snapshot('''\
 x = """hello
-world"""\
```
dmadisetti, line 560: @ChidiebereNjoku I don't think these tests should change

dmadisetti (changes requested):
Hmm not exactly, https://github.com/marimo-team/marimo/actions/runs/28353231401/job/84115621989?pr=9877 should still pass

You can test this with `marimo check examples/`

I know this has been in queue for a bit. I can take a pass if you'd like. We should just check to see if the string is under-indented in `smart_indent` before applying the "smart" indent

ChidiebereNjoku:
Thanks for the review. On the test changes, the snapshots changed because smart_dedent now correctly preserves whitespace inside multiline string literals (which is the whole point of this PR). For example, world""" previously had its leading spaces stripped (the bug), and now     world""" is preserved correctly. Should I revert these test changes to keep the old (broken) behaviour in fixed_dedent, or are you happy to accept the corrected snapshots? Happy to follow your preference here.

@dmadisetti

ChidiebereNjoku:
Fixed both issues, turns out your instinct about checking whether the string is under-indented before applying the smart indent was exactly right.


A string block now only shifts along with the surrounding code if its own minimum indentation is at least as much as what's being stripped. So blocks that are "in step" with the code — like mo.md(r"""...) written to visually match the function body — dedent properly now. Blocks with deliberately shallower or varied indentation (the original bug case) stay fully untouched.


That fixed the CI failure too, the 282 markdown-indentation warnings are gone, and I've included the ~200 auto-fixed example files since they're a direct result of this. Also updated the test snapshots to reflect the corrected output, since it's different from both the old buggy version and the earlier full-protection approach.


105 tests passing, original repro still holds up. Should be good for another look whenever you're free.

dmadisetti:
😬  there's a lot of failing tests here @ChidiebereNjoku I think I'll take this over. thanks for getting us this far though!

dmadisetti (approved):
Sorry this has been stalled @ChidiebereNjoku but I think it's fine now. For agent driven development I find "Test Driven Development" goes a long way.

ChidiebereNjoku:
Thank you @dmadisetti  for taking this over the finish line and for the extra polish, appreciate you sticking with this through all the back and forth. Good to have it in.

PR #9889, opened 2026-06-15
fix(formatter): wrap cells in function scope before ruff formatting

### LINKED ISSUE #9848: Extra lines around nested functions conflict with Ruff formatting
### Describe the bug

This code with a nested function is formatted by Marimo as:

```python
import marimo

__generated_with = "0.23.9"
app = marimo.App()

with app.setup:
    import marimo as mo


@app.cell
def _():
    x = 3


    def _foo():
        return x + 1


    print(_foo())
    return


if __name__ == "__main__":
    app.run()
```

However, Ruff will format the code as:

```python
import marimo

__generated_with = "0.23.9"
app = marimo.App()

with app.setup:
    import marimo as mo


@app.cell
def _():
    x = 3

    def _foo():
        return x + 1

    print(_foo())
    return


if __name__ == "__main__":
    app.run()
```

Notice there is only one empty line surrounding `_foo()` when formatted with Ruff.

I suspect this is because Marimo is treating the cell contents the same as the "file" scope, while running Ruff directly on the file will treat cells as functions.

This is probably related to marimo-team/marimo#6844.

### Will you submit a PR?

- [ ] Yes

### Environment

<details><summary>\`\`\`{</summary>

"marimo": "0.23.9",<br>  "editable": false,<br>  "location": "C:/Users/andrew/Code/Project/code/.venv/Lib/site-packages/marimo",<br>  "OS": "Windows",<br>  "OS Version": "11",<br>  "Processor": "Intel64 Family 6 Model 198 Stepping 2, GenuineIntel",<br>  "Python Version": "3.13.9",<br>  "Locale": "--",<br>  "Binaries": {<br>    "Browser": "149.0.7827.54",<br>    "Node": "v22.22.3",<br>    "uv": "0.9.28 (0e1351e40 2026-01-29)"<br>  },<br>  "Dependencies": {<br>    "click": "8.2.1",<br>    "docutils": "0.22",<br>    "itsdangerous": "2.2.0",<br>    "jedi": "0.19.2",<br>    "markdown": "3.8.2",<br>    "narwhals": "2.21.0",<br>    "packaging": "25.0",<br>    "psutil": "7.0.0",<br>    "pygments": "2.19.2",<br>    "pymdown-extensions": "10.21.2",<br>    "pyyaml": "6.0.2",<br>    "starlette": "0.47.3",<br>    "tomlkit": "0.13.3",<br>    "typing-extensions": "4.15.0",<br>    "uvicorn": "0.35.0",<br>    "websockets": "15.0.1"<br>  },<br>  "Optional Dependencies": {<br>    "altair": "6.1.0",<br>    "anywidget": "0.9.21",<br>    "duckdb": "1.3.2",<br>    "loro": "1.10.3",<br>    "mcp": "1.18.0",<br>    "nbformat": "5.10.4",<br>    "openai": "2.30.0",<br>    "pandas": "2.3.2",<br>    "polars": "1.32.3",<br>    "pyarrow": "21.0.0",<br>    "pytest": "8.4.1",<br>    "python-lsp-ruff": "2.2.2",<br>    "python-lsp-server": "1.13.1",<br>    "ruff": "0.12.11",<br>    "sqlglot": "30.7.0",<br>    "vegafusion": "2.0.3",<br>    "watchdog": "6.0.0"<br>  },<br>  "Experimental Flags": {}<br>}

```
```

</details>

### Code to reproduce

*No response*

VishakBaddur:
## 📝 Summary

Closes #9848

`RuffFormatter.format()` was passing raw cell code directly to `ruff format` via stdin. Ruff treats stdin as file scope and applies E302 (2 blank lines around top-level function definitions). But when the user runs `ruff` on the saved `.py` file, cells are already wrapped in `def _():` functions, so ruff applies E301 (1 blank line around nested defs). This creates a conflict - marimo's in-editor formatting and ruff on the saved file disagree on blank lines around nested functions.

**Fix:** Before sending to ruff, wrap each non-empty cell in `def _():\n` + indented body. After formatting, drop the first line, dedent, and strip. This makes marimo match what ruff produces on the saved file. Whitespace-only cells are left unwrapped (empty function body is invalid Python). If wrapping causes ruff to reject a cell, it falls back to formatting unwrapped.

## 📋 Pre-Review Checklist
- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on Discord, or the community discussions. (Issue: #9848)

## ✅ Merge Checklist
- [x] Documentation has been updated where applicable, including docstrings for API changes. (No public API changes; internal formatter only.)

### DIFF
--- a/marimo/_ast/parse.py
+++ b/marimo/_ast/parse.py
@@ -74,6 +74,24 @@ def fixed_dedent(text: str) -> str:
     return dedent("\n".join(map(refill, lines)))
 
 
+def unwrap_cell_body(formatted: str) -> str:
+    """Extract the body of a wrapped `def _():` cell after ruff formatting.
+
+    Used by RuffFormatter to unwrap cells that were temporarily wrapped in a
+    dummy function so ruff applies function-scope blank-line rules (E301)
+    instead of file-scope rules (E302).
+    """
+    tree = ast_parse(formatted)
+    fn = tree.body[0]
+    assert isinstance(fn, ast.FunctionDef)
+    assert fn.end_lineno is not None
+    extractor = Extractor(formatted)
+    raw = extractor.extract_from_offsets(
+        fn.body[0].lineno - 1, 0, fn.end_lineno - 1, fn.end_col_offset
+    )
+    return fixed_dedent(raw).strip()
+
+
 def extract_lineno(node: Node) -> int:
     if not isinstance(
         node, (ast.AsyncFunctionDef, ast.FunctionDef, ast.ClassDef)
--- a/marimo/_ast/parse.py
+++ b/marimo/_ast/parse.py
@@ -83,8 +83,12 @@ def unwrap_cell_body(formatted: str) -> str:
     """
     tree = ast_parse(formatted)
     fn = tree.body[0]
-    assert isinstance(fn, ast.FunctionDef)
-    assert fn.end_lineno is not None
+    if not isinstance(fn, ast.FunctionDef):
+        raise ValueError(
+            f"Expected a FunctionDef node, got {type(fn).__name__}"
+        )
+    if fn.end_lineno is None:
+        raise ValueError("FunctionDef node has no end_lineno")
     extractor = Extractor(formatted)
     raw = extractor.extract_from_offsets(
         fn.body[0].lineno - 1, 0, fn.end_lineno - 1, fn.end_col_offset
--- a/marimo/_utils/formatter.py
+++ b/marimo/_utils/formatter.py
@@ -1,14 +1,13 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-import ast
 import asyncio
 import subprocess
 import sys
 import textwrap
 
 from marimo import _loggers
-from marimo._ast.parse import Extractor, fixed_dedent
+from marimo._ast.parse import unwrap_cell_body
 from marimo._dependencies.dependencies import DependencyManager
 from marimo._types.ids import CellId_t
 
@@ -171,17 +170,7 @@ class RuffFormatter(Formatter):
             if not code.strip():
                 result[key] = wrapped_result.get(key, code)
             elif key in wrapped_result:
-                formatted = wrapped_result[key]
-                tree = ast.parse(formatted)
-                fn = tree.body[0]
-                extractor = Extractor(formatted)
-                raw = extractor.extract_from_offsets(
-                    fn.body[0].lineno - 1,
-                    0,
-                    fn.end_lineno - 1,
-                    fn.end_col_offset,
-                )
-                result[key] = fixed_dedent(raw).strip()
+                result[key] = unwrap_cell_body(wrapped_result[key])
 
         return result
 
--- a/marimo/_utils/formatter.py
+++ b/marimo/_utils/formatter.py
@@ -146,12 +146,23 @@ class RuffFormatter(Formatter):
     ) -> CellCodes:
         # Wrap each cell in a dummy function so ruff applies function-scope
         # blank-line rules (1 blank line around nested defs) instead of
-        # file-scope rules (2 blank lines). Cells with only whitespace are
-        # left unwrapped because an empty function body is invalid Python.
+        # file-scope rules (2 blank lines). Cells with no executable
+        # statements (whitespace-only or comment-only) are left unwrapped
+        # since an empty or comment-only function body is invalid Python.
+        skip_wrap = {
+            key
+            for key, code in codes.items()
+            if all(
+                not line.strip() or line.strip().startswith("#")
+                for line in code.splitlines()
+            )
+        }
         wrapped: CellCodes = {}
         for key, code in codes.items():
-            if code.strip():
-                wrapped[key] = "def _():\n" + textwrap.indent(code, "    ")
+            if key not in skip_wrap:
+                wrapped[key] = "\n".join(
+                    ["def _():", textwrap.indent(code, "    ")]
+                )
             else:
                 wrapped[key] = code
 
@@ -167,7 +178,7 @@ class RuffFormatter(Formatter):
         # parse mechanism in marimo/_ast/parse.py (Extractor + fixed_dedent).
         result: CellCodes = {}
         for key, code in codes.items():
-            if not code.strip():
+            if key in skip_wrap:
                 result[key] = wrapped_result.get(key, code)
             elif key in wrapped_result:
                 result[key] = unwrap_cell_body(wrapped_result[key])
--- a/marimo/_utils/formatter.py
+++ b/marimo/_utils/formatter.py
@@ -1,12 +1,14 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
+import ast
 import asyncio
 import subprocess
 import sys
 import textwrap
 
 from marimo import _loggers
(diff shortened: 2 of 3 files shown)


marimo/_utils/formatter.py
```diff
@@ -142,14 +143,50 @@ class RuffFormatter(Formatter):
...
             "format",
             "--line-length",
             str(self.line_length),
             stdin_filename=stdin_filename,
         )
 
+        # Unwrap: drop the def _(): line, dedent, strip.
+        result: CellCodes = {}
+        fallback_keys = []
+        for key, code in codes.items():
```
dmadisetti, line 180: I wonder if we can use our own parse mechanism here? This feels brittle but is maybe fine

marimo/_utils/formatter.py
```diff
@@ -142,14 +143,50 @@ class RuffFormatter(Formatter):
...
+            if not code.strip():
+                result[key] = wrapped_result.get(key, code)
+            elif key in wrapped_result:
+                lines = wrapped_result[key].split("\n")
+                result[key] = textwrap.dedent("\n".join(lines[1:])).strip()
+            else:
+                # Wrapping caused ruff to reject this cell; retry unwrapped.
+                fallback_keys.append(key)
+
+        if fallback_keys:
```
dmadisetti: Any known code paths trigger this or is this just defensive?

dmadisetti (commented):
Thanks again @VishakBaddur !

Test looks great. I wonder if we can trigger that "failed" branch though
Think you can give a shot at using `marimo/ast/_parse.py`?

marimo/_utils/formatter.py
VishakBaddur: Switched to Extractor + fixed_dedent from marimo/_ast/parse.py!

marimo/_utils/formatter.py
VishakBaddur: Couldn't find a real code path that triggers it(any valid cell body is valid inside def _():),  so removed it.

VishakBaddur:
Switched the unwrap to use marimo/_ast/parse.py and removed the fallback since I couldn't find a case that actually hits it. Let me know if you think there's an edge case I'm missing!

marimo/_utils/formatter.py
```diff
@@ -142,14 +145,46 @@ class RuffFormatter(Formatter):
...
 
+        # Unwrap: use the AST to extract the function body, mirroring the
+        # parse mechanism in marimo/_ast/parse.py (Extractor + fixed_dedent).
+        result: CellCodes = {}
+        for key, code in codes.items():
+            if not code.strip():
+                result[key] = wrapped_result.get(key, code)
+            elif key in wrapped_result:
+                formatted = wrapped_result[key]
+                tree = ast.parse(formatted)
```
dmadisetti: Woops, `ast_parse` since ast.parse issues logging side effects, also I think we can lift this into a helper in `marimo._ast.parse` (I thought it already existed, but maybe not)

dmadisetti (commented):
Another comment, sorry. Also realizing- can you add a top level function test? e.g.

```python
@app.function
def foo():
    """Something here"""

File: marimo/_utils/formatter.py (hunk 3 above)
Comment (VishakBaddur) [reply marimo/_utils/formatter.py]: Done, moved into unwrap_cell_body in marimo/_ast/parse.py using ast_parse. Added the @app.function test too.

File: marimo/_utils/formatter.py [hunk 4]
```diff
@@ -142,14 +144,36 @@ class RuffFormatter(Formatter):
...
     ) -> CellCodes:
-        return await ruff(
-            codes,
+        # Wrap each cell in a dummy function so ruff applies function-scope
+        # blank-line rules (1 blank line around nested defs) instead of
+        # file-scope rules (2 blank lines). Cells with only whitespace are
+        # left unwrapped because an empty function body is invalid Python.
+        wrapped: CellCodes = {}
+        for key, code in codes.items():
+            if code.strip():
```
dmadisetti: This may fail for pure comments too just realizing- can we add a test?

```python
def _():
    # just a comment
```

is also a syntax error

marimo/_ast/parse.py
```diff
@@ -74,6 +74,24 @@ def refill(line: str) -> str:
...
+    """Extract the body of a wrapped `def _():` cell after ruff formatting.
+
+    Used by RuffFormatter to unwrap cells that were temporarily wrapped in a
+    dummy function so ruff applies function-scope blank-line rules (E301)
+    instead of file-scope rules (E302).
+    """
+    tree = ast_parse(formatted)
+    fn = tree.body[0]
+    assert isinstance(fn, ast.FunctionDef)
+    assert fn.end_lineno is not None
```
dmadisetti: We should throw a value error instead of assert here I think.
Assertions are a little obtuse, and although this path should never be hit (type safety sanity check)

marimo/_utils/formatter.py
```diff
@@ -142,14 +144,36 @@ class RuffFormatter(Formatter):
...
+                wrapped[key] = "def _():\n" + textwrap.indent(code, "    ")
```
dmadisetti: While we're making changes, fstring?

dmadisetti (commented):
Sorry found a bug- but clean!

marimo/_utils/formatter.py
dmadisetti: nvm, the \n will make that annoying

```suggestion
                wrapped[key] = "\n".join("def _():", textwrap.indent(code, "    "))
```

or reuse the helper, which will take care of the comment case (because there will be a return)

marimo/_utils/formatter.py
```diff
@@ -142,14 +144,36 @@ class RuffFormatter(Formatter):
...
             "--line-length",
             str(self.line_length),
             stdin_filename=stdin_filename,
         )
 
+        # Unwrap: use the AST to extract the function body, mirroring the
+        # parse mechanism in marimo/_ast/parse.py (Extractor + fixed_dedent).
+        result: CellCodes = {}
+        for key, code in codes.items():
+            if not code.strip():
```
dmadisetti: If we just use the helper, then we won't need the conditional (all code contents should be handled)

marimo/_utils/formatter.py
VishakBaddur: fixed by precomputing skip_wrap, a set of keys whose code has no executable statements (all lines are whitespace or comments). Both the wrap and unwrap loops key off this set, so comment-only cells are never wrapped. I've also added a test covering this case.

marimo/_ast/parse.py
VishakBaddur: replaced both with raise ValueError

marimo/_utils/formatter.py
VishakBaddur: Used "\n".join(...) as suggested. Regarding the alternative to reuse the helper, if we wrapped comment-only cells and appended a pass to make the function body syntactically valid, the pass would end up preserved in the unwrapped output. Handling them via skip_wrap keeps the final output completely clean.

marimo/_utils/formatter.py
VishakBaddur: The skip_wrap set removes the need for the inline condition in both loops. wrap loop uses key not in skip_wrap, unwrap loop uses key in skip_wrap. No duplication, no condition repeated inline

dmadisetti (approved):
Nice!

PR #9895, opened 2026-06-15
feat(runtime): cell-level cached execution lifecycle

dmadisetti:
## Summary

This PR introduces the `CachedLifecycle` which can be turned on via `[marimo.tools.runtime.cache_cells]`.

When the Lifecycle is active, all executed code is attempted to be cache. Stub hydration occurs when a cell must be executed. When stub hydration fails (e.g. an Unhashable stub), this code change allows for cell rescheduling, forcing a rerun.

 - Followups include intelligent caching (i.e. do no cache cases where the caching process would become slow)
 - Surfacing caching information to the user

### DIFF
--- a/marimo/_runtime/executor/lifecycles/cached.py
+++ b/marimo/_runtime/executor/lifecycles/cached.py
@@ -16,7 +16,7 @@ There are 4 potential outcomes from attempted restoration:
 2. Stale cache
    In this case, a "UnhashableStub" requirement or cache mechanism failure
    indicates that the cell must rerun live, and may have ancestors that require
-   a re-run as well. In this case, the cell raises a MarimoCancelCellError with
+   a re-run as well. In this case, the cell raises a MarimoRescheduleError with
    the set of cells to requeue.
 3. Cache miss
    Cache was not found, ancestors do not require re-run, so the cell can run as
@@ -89,7 +89,7 @@ import time
 from typing import TYPE_CHECKING, cast
 
 from marimo import _loggers
-from marimo._runtime.exceptions import MarimoCancelCellError
+from marimo._runtime.exceptions import MarimoRescheduleError
 from marimo._runtime.executor.lifecycles import Skip
 from marimo._runtime.runner.result import RunResult
 from marimo._save.cache import Cache, CacheException
@@ -187,7 +187,7 @@ class CachedLifecycle:
             self._attempts.pop(cell_id, None)
             # Fall through to miss-path execution.
 
-        # Raises MarimoCancelCellError if any ref requires rehydration.
+        # Raises MarimoRescheduleError if any ref requires rehydration.
         self._preflight_refs(cell, glbls)
 
         self._exec_starts[cell_id] = time.time()
@@ -257,7 +257,7 @@ class CachedLifecycle:
         Walks `cell.refs` and checks each name in `glbls` for an
         `UnhashableStub` instance. If any are found, invalidates each producer's
         recorded manifest, drops this cell's attempt so teardown will no-op, and
-        raises `MarimoCancelCellError` with `cells_to_rerun` populated.
+        raises `MarimoRescheduleError` with `cells_to_rerun` populated.
         """
         cell_id = cell.cell_id
         stub_vars: list[str] = []
@@ -297,12 +297,12 @@ class CachedLifecycle:
         self._exec_starts.pop(cell_id, None)
 
         LOGGER.info(
-            "Rescheudling for %s: stub refs %s; Cell Ids %s",
+            "Rescheduling for %s: stub refs %s; Cell Ids %s",
             cell_id,
             stub_vars,
             fresh,
         )
-        raise MarimoCancelCellError(cells_to_rerun={cell_id} | fresh)
+        raise MarimoRescheduleError(cells_to_rerun={cell_id} | fresh)
 
     def _invalidate(self, cell_id: CellId_t) -> None:
         """Invalidate `cell_id`'s manifest so it re-runs live.
--- a/marimo/_runtime/executor/lifecycles/cached.py
+++ b/marimo/_runtime/executor/lifecycles/cached.py
@@ -73,6 +73,14 @@ class CachedLifecycle:
         # Per-cell manifest path, recorded on hit/save. Consumed when
         # this cell's pre-flight invalidates an upstream producer.
         self._manifest_keys: dict[CellId_t, str] = {}
+        # Producers already invalidated for re-execution during THIS run. A
+        # `CachedLifecycle` lives for one `Runner` (one `run_all`), so this
+        # set resets every reactive round — a later round self-heals once an
+        # upstream value recovers. Within a round, a producer that still
+        # yields a stub after its rerun is not requeued again, so a stub that
+        # can't be recomputed falls through to a tripwire raise instead of
+        # looping forever.
+        self._invalidated: set[CellId_t] = set()
 
     def setup(self, cell: CellImpl, glbls: MutableGlobals) -> Skip | None:
         cell_id = cell.cell_id
@@ -132,6 +140,14 @@ class CachedLifecycle:
         run_result: RunResult,
     ) -> None:
         cell_id = cell.cell_id
+        # Release the requeue guard only once the rerun actually replaced the
+        # stubbed value. A run can finish without raising yet still propagate
+        # a stub (e.g. `g = f` where `f` is a stub), which must stay guarded
+        # or the consumer requeues this producer forever.
+        if run_result.exception is None and not self._defines_stub(
+            cell, glbls
+        ):
+            self._invalidated.discard(cell_id)
         attempt = self._attempts.pop(cell_id, None)
         exec_start = self._exec_starts.pop(cell_id, None)
 
@@ -163,6 +179,11 @@ class CachedLifecycle:
             LOGGER.warning("Cache save failed for %s: %s", cell_id, e)
 
     @staticmethod
+    def _defines_stub(cell: CellImpl, glbls: MutableGlobals) -> bool:
+        """True if any name the cell defines is still an UnhashableStub."""
+        return any(_is_unhashable_stub(glbls.get(name)) for name in cell.defs)
+
+    @staticmethod
     def _restored_ui_defs(attempt: Cache, glbls: MutableGlobals) -> bool:
         """True if any def restored from the cache is a live UIElement."""
         from marimo._plugins.ui._core.ui_element import UIElement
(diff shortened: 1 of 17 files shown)


marimo/_runtime/runner/cell_runner.py
```diff
@@ -807,4 +831,17 @@ async def _dispatch_runnable(
...
+                        "Soft-cancel for %s; requeuing %s",
+                        cell_id,
+                        e.cells_to_rerun,
+                    )
+                    # Soft-cancel control signal from a lifecycle.
+                    # Move the cell back to queued state, and reschedule for
+                    # rerun after the relevant cells have been run.
+                    for rerun_id in e.cells_to_rerun:
+                        self.graph.cells[rerun_id].set_runtime_state("queued")
+                    self._scheduler.requeue_for_rerun(e.cells_to_rerun)
```
akshayka: I don't understand why `MarimoCancelCellError` signals that cells should be rerun.

Is this exception named appropriately? Additionally, even though `MarimoCancelCellError` was introduced in a previous PR, please add more documentation. It is not clear to me what a "soft-cancel" is, please be more precise.

marimo/_runtime/runner/cell_runner.py
```diff
@@ -430,25 +461,18 @@ async def run(self, cell_id: CellId_t) -> RunResult:
...
-                https://github.com/marimo-team/marimo/issues
-
-                Any additional context of what caused this error, such
-                as sample code to reproduce, will help us debug.
-                """
-            )
+        try:
+            run_result = self._finalize_run_result(raw_result, cell_id)
+        except BaseException:
+            LOGGER.exception(UNKNOWN_ERROR)
```
akshayka: Can `UNKNOWN_ERROR` be logged twice now? The control flow of two almost back-to-back try/except's here is somewhat confusing to me

marimo/_runtime/runner/cell_runner.py
```diff
@@ -157,6 +171,23 @@ def __init__(
...
             lifecycles.append(StrictLifecycle(self.graph))
+        if user_config is not None and user_config.get("runtime", {}).get(
+            "cache_cells", False
+        ):
+            from marimo._runtime.executor.lifecycles.cached import (
+                CachedLifecycle,
+            )
+
+            lifecycles.append(
+                CachedLifecycle(
```
akshayka, line 169: Nice, very clean

marimo/_runtime/runner/hook_context.py
```diff
@@ -42,6 +42,26 @@ def add(self, raising_cell: CellId_t, descendants: set[CellId_t]) -> None:
...
+    def discard(self, cell_id: CellId_t) -> None:
+        """Un-cancel a cell
+
+        That is remove it from the cancelled set and any descendant sets."""
+        self._by_raising_cell.pop(cell_id, None)
+        for raising_cell in list(self._by_raising_cell):
+            descendants = self._by_raising_cell[raising_cell]
+            descendants.discard(cell_id)
+            # A raiser with no cancelled descendants left is no longer a
+            # cancellation record; drop it so iteration and __bool__ stay honest.
```
akshayka:
```suggestion
            # A raiser with no cancelled descendants left is no longer a
            # cancellation record.
```

akshayka: Is this because `self._by_raising_cell[raising_cell]` also contains `raising_cell`? Otherwise a cell with no descendants could still be cancelled?

marimo/_runtime/runner/scheduler.py
```diff
@@ -117,6 +118,23 @@ def requeue(self, cell_ids: Iterable[CellId_t]) -> None:
         self._cells_to_run.clear()
         self._cells_to_run.extend(cell_ids)
 
+    def requeue_for_rerun(self, cells: set[CellId_t]) -> None:
+        """Soft-cancel by putting `cells` back at the head of the queue.
```
akshayka: Here, too, I don't understand what "Soft-cancel" means. Can you use more precise language?

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+
+On setup: hash the cell + consult the store.
+
+  - HIT:  restore defs into globals, return `Skip` with the cached
+          return value so the Evaluator short-circuits the body.
+  - MISS: pre-flight `cell.refs`: if any ref in `glbls` is an
+          `UnhashableStub` invalidate the producer's manifest and raise
+          `MarimoCancelCellError(cells_to_rerun=producers | self)`.
+          `Runner.run_all` catches the signal and hands it to
+          `Scheduler.requeue_for_rerun`.
```
akshayka: These comments are too specific. They should not reference implementation details of other modules. Please describe the semantics of what a hit and miss are, not the implementation details. For example, referencing `Runner.run_all` catching a signal and handing it to `Scheduler` is inappropriate.

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+
+LOGGER = _loggers.marimo_logger()
+
+
+def _is_unhashable_stub(value: object) -> bool:
+    return getattr(type(value), "__marimo_unhashable__", False) is True
+
+
+class CachedLifecycle:
+    """Skip cell exec on cache hit; backfill cell results on miss success."""
```
akshayka: What does it mean to "backfill cell results"? What is a "miss success"?

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+        self._loader = cast(
+            BasePersistenceLoader,
+            resolve_loader(PERSISTENT_LOADERS[loader])(name="cell_cache"),
+        )
+        # Per-cell state — populated in setup, consumed in teardown.
+        self._attempts: dict[CellId_t, Cache] = {}
+        self._exec_starts: dict[CellId_t, float] = {}
+        # Per-cell manifest path, recorded on hit/save. Consumed when
+        # this cell's pre-flight invalidates an upstream producer.
+        self._manifest_keys: dict[CellId_t, str] = {}
```
akshayka: What is a manifest path? What is stored in here?

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+
+            # UI construction is cheap and inherently
+            # session state.
+            LOGGER.debug(
+                "Cache hit for %s defines UI elements; running "
+                "live to register them with this session",
+                cell_id,
+            )
+            self._attempts.pop(cell_id, None)
+            # Fall through to miss-path execution.
```
akshayka: Does this mean that cells that define UI elements are always run, and never short-circuited? If so that could be a surprising "gotcha" to users, who may mix UI element construction and expensive code in the same cell.

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+                producers.update(self._graph.get_defining_cells(var_name))
+            except KeyError:
+                pass
+        producers.discard(cell_id)
+
+        fresh = {p for p in producers if p not in self._invalidated}
+        if not fresh:
+            LOGGER.warning(
+                "Pre-flight for %s: stub refs %s persist after producer "
+                "rerun; not requeuing (tripwire raises on access)",
```
akshayka: What does this mean? "Tripwire raises on access"?

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+        LOGGER.info(
+            "Pre-flight requeue for %s: stub refs %s; producers %s",
+            cell_id,
+            stub_vars,
+            fresh,
+        )
+        raise MarimoCancelCellError(cells_to_rerun={cell_id} | fresh)
+
+    def _invalidate(self, cell_id: CellId_t) -> None:
+        """Invalidate `cell_id`'s manifest so it re-runs live.
```
akshayka, line 307:
```suggestion
        """Invalidate `cell_id`'s manifest so it re-runs.
```

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+        return None
+
+    def teardown(
+        self,
+        cell: CellImpl,
+        glbls: MutableGlobals,
+        run_result: RunResult,
+    ) -> None:
+        cell_id = cell.cell_id
+        # Release the requeue guard.
```
akshayka: This comment says what the code does. Instead it should provide context: what does it mean if the conditions in the if statement are met, and why does that mean the cell should be invalidated?

marimo/_save/loaders/lazy.py
```diff
@@ -48,6 +47,10 @@
 LOGGER = _loggers.marimo_logger()
 
 
+# Runtime cache: type → loader string, populated by maybe_update_lazy_stub().
```
akshayka:
```suggestion
# Runtime cache: type -> loader string, populated by maybe_update_lazy_stub().
```

akshayka: This comment should reference `maybe_update_lazy_stub`. That's a non-local implementation detail. Additionally "type to loader string" does not tell me anything more than what's already in the type annotation. The comment should be about semantics of what this cache is and does and what is stored.

marimo/_runtime/executor/lifecycles/cached.py
akshayka: If this is the case, at the very least we should document it in the codebase with a TODO to document it in our public documentation.

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,252 @@
...
+          return value so the Evaluator short-circuits the body.
+  - MISS: pre-flight `cell.refs`: if any ref in `glbls` is an
+          `UnhashableStub` invalidate the producer's manifest and raise
+          `MarimoCancelCellError(cells_to_rerun=producers | self)`.
+          `Runner.run_all` catches the signal and hands it to
+          `Scheduler.requeue_for_rerun`.
+
+          Otherwise, fall through and let the body run.
+
+On teardown: backfill on successful miss; drop the attempt on a raised
```
akshayka: This is not clear. What is a successful miss? What is a backfill?

marimo/_runtime/executor/lifecycles/cached.py
dmadisetti: Moved over some of my docs. The diagram might be a but much, but generated with mermaid-ascii

marimo/_runtime/runner/cell_runner.py
dmadisetti: Renamed this "Reschedule", which I think removes the ambiguity of "Soft-Cancel"

marimo/_runtime/executor/lifecycles/cached.py
```diff
@@ -0,0 +1,321 @@
...
+            except (Exception, CacheException) as e:
+                LOGGER.warning("Cache restore failed for %s: %s", cell_id, e)
+                self._attempts.pop(cell_id, None)
+                # Fall through to miss-path execution.
+
+        if restored:
+            # Defer loading if restored and not a UI element.
+            # TODO(dmadisetti): Attempt to restore UI elements as well.
+            # Currently the UIElement class has UIDs that are session-specific.
+            # For now, UI construction is cheap and inherently session state.
```
akshayka: This can be expensive if the cell defining the UI element also has expensive code, right? If so, can you update the comment? "UI construction is cheap" is misleading.

akshayka (commented):
This is much better, thank you. I liked the renamed exception. A couple of comments

marimo/_runtime/runner/hook_context.py
dmadisetti: Should be applied now?

marimo/_runtime/executor/lifecycles/cached.py
dmadisetti: Yup! Removed

akshayka (approved):

PR #9898, opened 2026-06-15
feat(save): LazyStore dual-mode WASM backend

dmadisetti:
## Summary

This PR creates a WASM compatible LazyStore and introduces the `DualMode` store concept, where a targeted store can have different behaviors on different platforms.

The LazyLoader wasm `dual` loads values from an expected cache value using pyodide's native `fetch` mechanism, and falls back to caching in memory (since no disk writes in WASM)

### DIFF
--- a/marimo/_save/loaders/lazy.py
+++ b/marimo/_save/loaders/lazy.py
@@ -58,7 +58,7 @@ def maybe_update_lazy_stub(value: Any) -> str:
     """Return the loader strategy string for *value*, caching the result.
 
     Walks the MRO of `type(value)` against `LAZY_STUB_LOOKUP` (a
-    fq-class-name → loader-string registry).  Falls back to `"pickle"`
+    fq-class-name -> loader-string registry).  Falls back to `"pickle"`
     when no match is found.
     """
     value_type = type(value)
@@ -182,12 +182,7 @@ _FALLBACK_CACHE_STATE: CacheState | None = None
 
 
 def _cache_state() -> CacheState:
-    """Cache state for the current session, scoped to the root context.
-
-    Embedded apps run in child contexts (`create_kernel_context(parent=...)`);
-    walk to the root so the loader registry, poison set, and WASM store are
-    shared across the session rather than per child.
-    """
+    """Cache state for the current session, scoped to the root context."""
     ctx = safe_get_context()
     if ctx is not None:
         while ctx.parent is not None:
@@ -213,10 +208,7 @@ class LazyStore(Store):
     """Native store for `LazyLoader`.
 
     Delegates to an inner store (default `FileStore` at `__marimo__/cache/`)
-    and tracks which keys were written or read so `--execute` export knows
-    exactly which blobs to bundle. Does no environment checks; the WASM
-    variant (`WasmLazyStore`) adds HTTP-backed reads.
-    """
+    and tracks which keys were written or read."""
 
     def __init__(self, inner: Store | None = None) -> None:
         self._inner = inner if inner is not None else FileStore()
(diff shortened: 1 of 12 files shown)


marimo/_save/stores/store.py
```diff
@@ -23,4 +27,28 @@ def clear(self, key: str) -> bool:
         return False
 
 
+class WasmExportableStore(Store):
```
mscolnick: maybe consider just adding to `Store`? seems like useful/common functions, but idk if its always used

mscolnick: `get_batch` could have a default impl. `export_manifest` could be maybe be `export_keys()` or just `list_keys()`. it could default to empty if causes churn for other stores

marimo/_save/loaders/lazy.py
```diff
@@ -165,41 +179,233 @@ def from_item(item: Item) -> Any:
...
+        _WASM_DICT_STORE = DictStore()
+    return _WASM_DICT_STORE
+
+
+# Keys whose deserialization failed — never re-fetched over HTTP. Survives
+# across LazyStore instances.
+_POISONED_KEYS: set[str] = set()
+
+
+class LazyStore(WasmExportableStore):
```
mscolnick: is it possible to re-use `TieredStore` for some of these and just mix and match instead of integrate more deeply?

marimo/_save/loaders/lazy.py
```diff
@@ -165,41 +179,233 @@ def from_item(item: Item) -> Any:
...
+        self, keys: Iterable[str]
+    ) -> Iterator[tuple[str, bytes | None]]:
+        for k in keys:
+            yield k, self.get(k)
+
+    def export_manifest(self) -> list[str]:
+        return sorted(self._written_keys | self._touched_keys)
+
+
+class WasmLazyStore(LazyStore):
```
mscolnick, line 246: could this also re-use `TieredStore`?

marimo/_save/stores/store.py
dmadisetti: Yeah, a bit too much abstraction here

marimo/_save/loaders/lazy.py
dmadisetti: TieredStore will stack on these

marimo/_save/loaders/lazy.py
```diff
@@ -165,41 +175,247 @@ def from_item(item: Item) -> Any:
...
+        for key, data in results:
+            if data is not None:
+                self._inner.put(key, data)
+                self._touched_keys.add(key)
+            yield key, data
+
+
+# Registry of active loaders by name, so a recreated loader reuses the same
+# store (and so `flush_all`/export can reach every session loader).
+_ACTIVE_LAZY_LOADERS: dict[str, LazyLoader] = {}
```
mscolnick: does this need to be keyed by nested notebooks to avoid conflicts?

marimo/_save/loaders/lazy.py
```diff
@@ -165,41 +175,247 @@ def from_item(item: Item) -> Any:
...
+    if _WASM_DICT_STORE is None:
+        from marimo._save.stores.dict_store import DictStore
+
+        _WASM_DICT_STORE = DictStore()
+    return _WASM_DICT_STORE
+
+
+# Keys whose deserialization failed — never re-fetched over HTTP. Survives
+# across LazyStore instances.
+_POISONED_KEYS: set[str] = set()
```
mscolnick: does this need to be keyed by nested notebooks to avoid conflicts?

mscolnick (dismissed):

marimo/_save/loaders/lazy.py
```diff
@@ -399,37 +656,172 @@ def save_cache(self, cache: Cache) -> bool:
...
+                raise FileNotFoundError("Incomplete cache: missing blobs")
+            unpickled[key] = self._deserialize_blob(
+                key, data, ref_type_hints, return_ref, return_type_hint
+            )
+        return unpickled
+
+    def _dispatch_write(self, write_fn: Callable[[], None]) -> None:
+        write_fn()  # synchronous — no threads in Pyodide
+
+    def _on_restore_failure(
```
mscolnick, line 770: missing a test

marimo/_save/loaders/lazy.py
dmadisetti: We shouldn't have cache collisions

dmadisetti:
@mscolnick moved globals to existing shared object and added test

mscolnick (dismissed):

PR #9918, opened 2026-06-18
Fix/container gc retention

### LINKED ISSUE #9880: Multiple instances of potential garbage collection of live widgets
If I instantiate a widget like

```python
get_state, set_state = mo.state(False)
wgt = mo.ui.checkbox(on_change=set_state)
wgt
```

```python
get_state()
```
then all is well: the second cell updates as I click.  If, on the other hand, if I instantiate it like

```python
get_state, set_state = mo.state(False)
wgt = mo.ui.checkbox(on_change=set_state).style()
wgt
```

```python
get_state()
```
then `set_state` is never called and the second cell is frozen as `False`.

The issue is that `style` returns
```python
    return Html(h.div(children=as_dom_node(item).text, style=style_str))
```

This treats `item` as nothing more than the sum of its html string `as_dom_node(item).text`, wraps that html string into the `h.div`, and returns the composite `Html` object, dropping `item` on the floor.  In typical marimo usage, this isn't an issue because `item` is kept alive as a global or some child of some global.  But if the user creates `item` in such a way that no strong reference to it exists, for example because it's part of a dynamic tree of styled widgets, `item` is vulnerable to garbage collection and the Javascript representation of the widget loses connection with its Python half.

This is a bug in a variety of places; simply patching `style` is insufficient.  I haven't been through them all, but according to Claude Opus, 
* `mo.style`/`.style()`
* `mo.callout`/`.callout()`
* `mo.accordion`
* `mo.carousel`
* `mo.ui.tabs`
* `mo.sidebar`
* `mo.stat(slot=...)`

are all prone to dropping references and allowing garbage collection.

There is already a solution in the code, though, for `vstack` and `hstack`: the flexboxes inherit from `flex._FlexContainerHtml` which explicitly keeps alive its children by storing references to them, and `hypertext._BlockWrapped` does something similar.

I propose to add a class called `ContainerHtml` which explicitly keeps alive references to its children.  It's essentially just `_BlockWrapped` with a little flexibility built in:

```python
class ContainerHtml(Html):
    def __init__(self, children: Sequence[Html]) -> None:
        self._children: list[Html] = list(children)
        super().__init__(self._build_text())

    def _build_text(self) -> str:
        raise NotImplementedError

    @property
    def text(self) -> str:  # type: ignore[override]
        """Re-render children live on every access."""
        return self._build_text()
```

Functions like `style` and `accordion` would return a class like `StyledHtml` and `AccordionHtml`, derived from `ContainerHtml`, with their own implementations of `__init__` and `_build_text`.  Existing, similar classes like `_BlockWrapped` would be renamed/reparented for consistency.  Uses of `as_dom_node`/`as_html` to create containers would now all involve these container classes, so anyone who goes to make something new by reading old code would be more likely to adopt the new pattern instead of dropping the reference.

Potentially some refactor of `as_dom_node`/`as_html` to make them impossible to use without one of these classes would be warranted, but that would be more complicated and take a lot more thought.

I have a partial PR where I've applied the fix to `style` and `accordion` and it works great -- code similar to both snippets above functions correctly.  But this represents a refactor of central widget objects, including the stacks, and so as noted in CONTRIBUTING.md, I'm making a proposal and asking for feedback before I proceed further on this.

### Will you submit a PR?

- [x] Yes

### Environment

<details>

```
{
  "marimo": "0.23.9",
  "editable": true,
  "location": "/home/krohne/marimo-dev/marimo",
  "OS": "Linux",
  "OS Version": "6.14.0-37-generic",
  "Processor": "x86_64",
  "Python Version": "3.13.12",
  "Locale": "--",
  "Binaries": {
    "Browser": "148.0.7778.215",
    "Node": "--",
    "uv": "0.11.4 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.22.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.2.1",
    "tomlkit": "0.15.0",
    "uvicorn": "0.49.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.1.0",
    "anywidget": "0.9.21",
    "duckdb": "1.5.3",
    "loro": "1.10.3",
    "mcp": "1.27.2",
    "nbformat": "5.10.4",
    "openai": "2.41.0",
    "pandas": "3.0.3",
    "polars": "1.41.2",
    "pyarrow": "24.0.0",
    "pytest": "9.0.3",
    "ruff": "0.15.16",
    "sqlglot": "30.9.0"
  },
  "Experimental Flags": {
    "multi_column": true,
    "cache_panel": true,
    "isolate_apps": true
  }
}
```

</details>

### Code to reproduce

_No response_

edwardwkrohne:
## 📝 Summary
Closes #9880

Widgets created as

```python
get_state, set_state = mo.state(False)
wgt = mo.ui.checkbox(on_change=set_state).style()
wgt
```

would never fire their set_state methods.  The issue is that `style` returns 

```python
    return Html(h.div(children=as_dom_node(item).text, style=style_str))
```

which drops the actual python object `item`, preventing the corresponding frontend ui object from actually interacting with the backend.  

Now, `style` is a class that derives from a new class `ContainerHtml`, which explicitly contains strong references to a list of children to prevent them from being garbage collected.  The offending code survives as 

```python
    def _build_text(self) -> str:
        return h.div(children=self._children[0].text, style=self._style_str)
```

which uses the explicitly stored child `self._children[0]` instead of a local.  Similar treatment has been given to `accordion`, `callout`, `carousel`, `tabs`, and `sidebar`, all of which could potentially drop their children by converting to text and losing the original.

`tabs` was slightly different in that it was already a `UIElement`, which would have been incompatible with the implementation of the new `ContainerHtml` class, so it was modified to explicitly store its own children.

`sidebar` was reparented to `ContainerHtml` and stores its children that way.

`_BlockWrapped` was also reparented to `ContainerHtml`.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/9918?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/marimo/_output/hypertext.py
+++ b/marimo/_output/hypertext.py
@@ -14,7 +14,7 @@ from marimo._output.utils import flatten_string
 from marimo._utils.methods import getcallable
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import Iterator, Sequence
 
     from marimo._plugins.ui._core.ui_element import UIElement
     from marimo._plugins.ui._impl.batch import batch as batch_plugin
@@ -289,7 +289,39 @@ class Html(MIME):
         return self.text
 
 
-class _BlockWrapped(Html):
+class ContainerHtml(Html):
+    """Base class for Html wrappers that contain other renderables.
+
+    A `ContainerHtml` holds *strong* references to its children and re-renders
+    from their live `.text` on every access. This matters for two reasons:
+
+    - UI elements are only weakly referenced by the UI element registry, so a
+      container that merely froze `child.text` at construction time would let
+      its children be garbage collected -- breaking interactivity for elements
+      returned from helper functions. Holding strong references keeps the
+      children alive, mirroring `mo.hstack`/`mo.vstack`.
+    - Mutable children (e.g. `mo.status.spinner`) re-render live rather than
+      being frozen at construction time.
+
+    Subclasses store any extra rendering state on `self` *before* calling
+    `super().__init__()` and implement `_build_text` in terms of
+    `self._children`.
+    """
+
+    def __init__(self, children: Sequence[Html]) -> None:
+        self._children: list[Html] = list(children)
+        super().__init__(self._build_text())
+
+    def _build_text(self) -> str:
+        raise NotImplementedError
+
+    @property
+    def text(self) -> str:  # type: ignore[override]
+        """Re-render children live on every access."""
+        return self._build_text()
+
+
+class _BlockWrapped(ContainerHtml):
     """Wraps another Html in a plain block `<div>`.
 
     Used by :meth:`Html.center`, :meth:`Html.left`, and :meth:`Html.right`
@@ -303,17 +335,12 @@ class _BlockWrapped(Html):
     """
 
     def __init__(self, inner: Html) -> None:
-        self._inner = inner
-        super().__init__(self._build_text())
+        super().__init__([inner])
 
     def _build_text(self) -> str:
         from marimo._output.builder import h
 
-        return h.div(self._inner.text)
-
-    @property
-    def text(self) -> str:
-        return self._build_text()
+        return h.div(self._children[0].text)
 
 
 MARIMO_NO_JS_KEY = "MARIMO_NO_JS"
--- a/marimo/_plugins/stateless/accordion.py
+++ b/marimo/_plugins/stateless/accordion.py
@@ -9,8 +9,37 @@ from marimo._plugins.core.web_component import build_stateless_plugin
 from marimo._plugins.stateless.lazy import lazy as lazy_ui
 
 
-class _AccordionHtml(ContainerHtml):
-    """Html produced by `mo.accordion()`; keeps live references to its children."""
+@mddoc
+class accordion(ContainerHtml):
+    """Accordion of one or more items.
+
+    Args:
+        items: a dictionary of item names to item content; strings are
+            interpreted as markdown
+        multiple: whether to allow multiple items to be open simultaneously
+        lazy: a boolean, whether to lazily load the accordion content.
+            This is a convenience that wraps each accordion in a `mo.lazy`
+            component.
+
+    Returns:
+        An `Html` object.
+
+    Example:
+        ```python3
+        mo.accordion(
+            {"Tip": "Use accordions to let users reveal and hide content."}
+        )
+        ```
+
+        Accordion content can be lazily loaded:
+
+        ```python3
+        mo.accordion({"View total": expensive_item}, lazy=True)
+        ```
+
+        where `expensive_item` is the item to render, or a callable that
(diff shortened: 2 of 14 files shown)


edwardwkrohne:

kirangadhave:
~~@edwardwkrohne Please read and sign the CLA https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md#your-first-pr~~

nvm you already did

marimo/_output/hypertext.py
```diff
@@ -289,7 +289,39 @@ def _repr_html_(self) -> str:
         return self.text
 
 
-class _BlockWrapped(Html):
+class ContainerHtml(Html):
```
kirangadhave, line 292: This is identical to https://github.com/marimo-team/marimo/blob/c34002e3afaa7b49312142f8ec5828763eecd738/marimo/_plugins/stateless/flex.py#L16-L70

I will consolidate both classes in a follow up. @edwardwkrohne you can ignore this comment.

marimo/_plugins/stateless/style.py
```diff
@@ -33,16 +31,26 @@ def style(
...
+        combined_style = style or {}
+
+        # Add kwargs to combined_style, converting snake_case to kebab-case
+        for key, value in kwargs.items():
+            kebab_key = key.replace("_", "-")
+            combined_style[kebab_key] = value
+
+        self._style_str = ";".join(
+            [f"{key}:{value}" for key, value in combined_style.items()]
+        )
```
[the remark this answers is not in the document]
kirangadhave: Please address this

marimo/_plugins/stateless/accordion.py
kirangadhave: `accordion` is not a class not a function, so the `Returns:` part of doc string doesn't make sense here. Please remove this. 

Also verify on all functions -> class conversions in the PR.
`callout`, `style`, `carousel`, `accordion`,

kirangadhave (changes requested):

kirangadhave (approved):
:rocket:

kirangadhave:
Thanks @edwardwkrohne

I'll merge this

PR #7223, opened 2025-11-20
pandas change to to_dict instead of to_json and fix rich elems

Light2Dark:
## 📝 Summary

`to_dict` is closer to the raw values returned by pandas. For example, the different versions of nulls below would return empty previously. Now they return values.

Screenshot 2025-11-20 at 6 06 31 PM

also changes to time periods

before:
Screenshot 2025-11-20 at 6 08 37 PM

after:
Screenshot 2025-11-25 at 2 07 13 AM

## 📋 Checklist

- [x] I have read the [contributor guidelines](https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md).
- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).
- [x] I have added tests for the changes made.
- [x] I have run the code and verified that it works as expected.

PR #9951, opened 2026-06-22
fix: stringify pint-pandas values in table JSON output

### LINKED ISSUE #9947: Pint Pandas series don't display correctly
### Describe the bug

For the last few versions, Pint Pandas dataframes/series do not display correctly. From what I can tell, this broke in an update from `0.18.0` -> `0.18.1` (though I vaguely remember some other update coming along in since that temporarily fixed it).

| Current (`>=0.18.1`) | Old (`<=0.18.0`) |
|--|--|
|Image |Image|

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

Reproducible on molab: https://molab.marimo.io/notebooks/nb_X8ReX5DZnEwWH2aiEAi664

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "pint-pandas==0.8.0",
# ]
# ///

import marimo

__generated_with = "0.23.9"
app = marimo.App(
    width="medium",
    css_file="/usr/local/_marimo/custom.css",
    auto_download=["html"],
)


@app.cell
def _():
    import pandas as pd
    import pint_pandas

    return (pd,)


@app.cell
def _(pd):
    pd.Series([1, 2, 3, 4], dtype="pint[meter]")
    return


if __name__ == "__main__":
    app.run()
```

Light2Dark:
## 📝 Summary

Closes #9947

Since 0.18.1 (#7223), pandas tables serialize rows via `to_dict(orient="records")` instead of `to_json(..., default_handler=str)`. That is better for NaNs and infs, but extension arrays like `pint-pandas` keep raw `pint.Quantity` objects. Those get encoded as nested dicts (`_magnitude`, `_units`, …) instead of readable strings like `1.0 meter`.

I only stringify extension-array columns when a sample cell value does not round-trip to a JSON primitive through `enc_hook`. That fixes pint-pandas without turning awkward-pandas numerics into strings. I also cache `infer_dtype` once per object column (instead of calling it twice).

image

image

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/tests/_plugins/ui/_impl/tables/test_pandas_table.py
+++ b/tests/_plugins/ui/_impl/tables/test_pandas_table.py
@@ -2174,12 +2174,24 @@ class TestPandasTableManager(unittest.TestCase):
         json_str = manager.to_json_str()
         json_data = json.loads(json_str)
 
-        assert json_data == [
-            {"value": "1.0 meter"},
-            {"value": "2.0 meter"},
-            {"value": "3.0 meter"},
-            {"value": "4.0 meter"},
-        ]
+        expected = [{"value": value} for value in series.astype(str)]
+        assert json_data == expected
+
+    def test_to_json_str_awkward_pandas_keeps_numeric_values(self) -> None:
+        """awkward-pandas primitives should stay numeric in table JSON."""
+        pytest.importorskip("awkward")
+        pytest.importorskip("awkward_pandas")
+        import awkward as ak
+        import awkward_pandas as akpd
+        import pandas as pd
+
+        series = pd.Series(
+            akpd.AwkwardExtensionArray(ak.Array([1.1, 2.2, 3.3]))
+        )
+        manager = self.factory.create()(series.to_frame(name="value"))
+        json_data = json.loads(manager.to_json_str())
+
+        assert json_data == [{"value": 1.1}, {"value": 2.2}, {"value": 3.3}]
 
     def test_to_arrow_ipc_fallback_for_unsupported_extension_dtype(
         self,
--- a/tests/_plugins/ui/_impl/tables/test_pandas_table.py
+++ b/tests/_plugins/ui/_impl/tables/test_pandas_table.py
@@ -7,8 +7,9 @@ import re
 import unittest
 import warnings
 from math import isnan, nan
+from types import SimpleNamespace
 from typing import Any
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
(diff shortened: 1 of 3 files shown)


Light2Dark:
Re: perf on `_extension_column_needs_stringify` sampling — benchmarked a few approaches on large Series (up to 5M rows):

**Current (`notna` + `argmax` + `iat`):** ~0.4–0.9ms even in the pathological case where only the *last* row is non-null. O(n) but vectorized in C/numpy.

**Alternatives we considered:**
- `dropna().iloc[0]` — also O(n), but allocates a copy (~25ms @ 5M rows)
- Python `iat` loop / `for val in series` — O(n) in Python; unusable on extension arrays at scale (~2–39s @ 5M for pint)

Also worth noting this only runs per **extension-array** column in `to_json_str`, not every column. When stringify is needed, `astype(str)` over the full column dominates anyway.

Leaving the current vectorized sampling as-is. An `iat[0]` fast path could be added later for the common case (first row valid) but doesn't seem necessary.

tests/_plugins/ui/_impl/tables/test_pandas_table.py
```diff
@@ -2164,6 +2164,89 @@ def test_search_with_multiindex_column_name_conflict(self) -> None:
...
+        with patch(
+            "pandas.api.types.is_extension_array_dtype", return_value=True
+        ):
+            assert _extension_column_needs_stringify(series)
+
+    def test_extension_column_needs_stringify_with_duplicate_index(
+        self,
+    ) -> None:
+        """Duplicate labels should not break extension-array sampling."""
+        from unittest.mock import patch
```
mscolnick: could this be at the top to reduce LOCs

tests/_plugins/ui/_impl/tables/test_pandas_table.py
```diff
@@ -2164,6 +2164,89 @@ def test_search_with_multiindex_column_name_conflict(self) -> None:
...
+        with patch(
+            "pandas.api.types.is_extension_array_dtype", return_value=True
+        ):
+            assert not _extension_column_needs_stringify(series)
+
+    def test_extension_column_needs_stringify_for_rich_extension_values(
+        self,
+    ) -> None:
+        """Extension columns with nested JSON values should be stringified."""
+        from types import SimpleNamespace
```
mscolnick: this can be hoisted to the top too

mscolnick (approved):

PR #9984, opened 2026-06-25
feat(table): add show_search and a reusable mo.ui.table.Display config

### LINKED ISSUE #9900: Disable search for mo.table element
### Description

I'm currently experimenting with writing blog-like content in marimo. Sometimes a md table is a bit too limited, so I use `mo.table`. 
The result looks great:

Image

But I would like to get rid of the 'search' feature, to have just the table.

I would like a `enable_search` flag or similar.

- [ ] Yes

### Alternatives

_No response_

### Additional context

Vibing blogs with marimo is starting to feel great ! I find it a very engaging way to dictate my thoughts to my LLM and see the narrative be illustrated at speed of thought ...

kirangadhave:
## 📝 Summary

Adds a `show_search: bool = True` parameter to `mo.ui.table` so authors can hide the search bar, and a reusable `mo.ui.table.Display` config object for the visibility flags.

`show_search` mirrors the existing `show_download` plumbing: the constructor threads it to the plugin as the `show-search` arg, and the frontend drives the search bar from it instead of a hardcoded `true`. While wiring this up the `DataTable` search prop was renamed from `enableSearch` to `showSearch` across the component, its top bar, and the dataframe preview, so the prop matches the other `show*` flags. The toolbar buttons now keep their right-aligned position when the search bar is hidden, rather than shifting left.

`mo.ui.table.Display` is a `TypedDict` of the four user-controllable visibility flags (`show_search`, `show_download`, `show_column_summaries`, `show_data_types`), all optional. Build one and unpack it into one or more tables, or derive a variant without mutating the original:

```python
cfg = mo.ui.table.Display(show_search=False, show_download=False)
mo.ui.table(data, **cfg)
mo.ui.table(other, **{**cfg, "show_column_summaries": False})
```

Closes #9900

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/9984?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Light2Dark (commented):
nice! crazy we have so many positional args. maybe we can move this to kwarg? idk.

kirangadhave:
> maybe we can move this to kwarg

yeah let's do that for this and going forward

mscolnick (approved):

PR #9988, opened 2026-06-25
feat: generate islands from notebook IR

peter-gy:
Adds `MarimoIslandGenerator.from_ir(...)` so callers can construct island exports from `NotebookSerialization` directly. This way, downstream exporters like quarto-marimo, jupyter-book-marimo and mdx-marimo mostly only need to worry about parsing their source format and mapping it to marimo IR.

Disabled and transitively disabled cells are marked nonreactive.

### DIFF
--- a/marimo/_islands/_island_generator.py
+++ b/marimo/_islands/_island_generator.py
@@ -301,7 +301,7 @@ class MarimoIslandGenerator:
         return generator
 
     @staticmethod
-    def from_ir(
+    def _from_ir(
         notebook: NotebookSerialization,
         *,
         app_id: str = "main",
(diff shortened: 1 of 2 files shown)


mscolnick:
is IR something we want to expose and downstream users to use?

peter-gy:
> is IR something we want to expose and downstream users to use?

I was torn on this initially. It would be really valuable for the downstream use cases like for exporters and also for generative-ai use cases where an LLM emits python code e.g. for data analysis and the user needs to execute that in a sandboxed environment and also needs to present it somehow. marimo is a perfect canvas for that. mapping the python code to an IR object instead of reconstructing a notebook from strings looks sensible. Also, the `NotebookSerialization` model has been relatively stable internally.

mscolnick:
@peter-gy do you expect users to build the IR by hand, or to use an parser or other marimo util to get the IR. if its the latter, im less worried about breaking changes. we also have `_schemas/notebook/NotebookV1` which is meant to be external / non-breaking

peter-gy:
> @peter-gy do you expect users to build the IR by hand, or to use an parser or other marimo util to get the IR. if its the latter, im less worried about breaking changes. we also have `_schemas/notebook/NotebookV1` which is meant to be external / non-breaking

as discussed, let's make it private for now and expose it as `MarimoIslandGenerator._from_ir`

PR #9989, opened 2026-06-25
feat(frontend): add show/hide all code commands to command palette

kirangadhave:
## 📝 Summary

Adds four command palette commands to toggle code editor visibility: "Show all code", "Hide all code", "Show all markdown code", and "Hide all markdown code". Previously only "Hide all markdown code" existed, with no way to show code back or to act on code cells in bulk.

The code and markdown commands operate on disjoint sets of cells: "Show/Hide all code" affects only non-markdown (Python/SQL) cells, and "Show/Hide all markdown code" affects only markdown cells. So "Show all code" does not reveal the source of markdown prose cells.

Each command persists `hide_code` to the notebook file (consistent with the per-cell "Hide code" toggle), so the state survives reload. All four are assignable to keyboard shortcuts from Settings but ship with no default binding to avoid clashing with existing shortcuts.

A single `useSetCodeVisibility(hidden, kind)` hook backs all four commands, where `kind` is `"code"` or `"markdown"`.

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/9989?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Light2Dark (approved):
nice 👍🏾

Issue #9387, opened 2026-04-26
Make Table search and Visualize elements optional


srl-git:
### Description

Marimo tables are great but when exporting to static HTML it would be very useful to be able to hide the interactive elements. There are already options to hide the export button, the data types etc but could this be extended to include the new Search function and the Visualise/Data Explorer buttons too?

We can use mo.plain to just display a plain html pandas table but it would be nicer to keep the Marimo style throughout the notebook.  

### Suggested solution

Add 
- show_search_bar 
- show_visualise
to marimo.ui.table 

### Are you willing to submit a PR? (You must receive approval from the team before submitting a PR.)

- [ ] Yes

Light2Dark:
Makes sense to me to hide those interactive elements by default in static HTML.

CodingKaiser:
This would be much appreciated. The marimo tables look great, but are untenable to use when delivering data to collaborators via static html, because most of the UI is nonfunctional.

I feel a general option `remove_interactive_elements` would be very beneficial, or alternatively, simply turning off those elements by default as @Light2Dark suggested.

srl-git:
A general `remove_interactive_elements` would be amazing and also a `suppress_static_notebook_warning` would be very useful. Exporting to PDF is too slow and we lose some UI elements, and hosting a live notebook with a backend is not always possible as an Analyst. Static HTML notebooks have become the default way of sharing quick reports for me.

Yasin197:
Similar to my request about hiding the data column type and distributiona ect, seems like there are requests around expectations when exploring data and presenting it to specific audiences

kirangadhave:
Improved static export tables to hide all interactive controls that need a running Python kernel.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/fa6551d9-8d4f-4fd9-9181-8f71e67125eb" />

@srl-git, @CodingKaiser  Have you tried exporting to WASM notebooks? It allows you to host a live notebook without the backend. You can event host it on github pages or your company's internal servers. https://docs.marimo.io/guides/wasm/

As long as your notebook is WASM compatible (see docs above) you get fully interactive tables without having to host a live backend.

srl-git:
Thanks for implementing that change @kirangadhave! I have looked into WASM notebooks but the ease of sharing 1 file with non technical users is very compelling

CodingKaiser:
[...]

PR #10007, opened 2026-06-26
feat(frontend): suppress kernel-dependent table controls in static export

kirangadhave:
## 📝 Summary

In a static HTML export, `mo.ui.table` renders chrome backed by live-kernel RPCs that silently no-op when used: search, filters, column summaries, sorting, pagination, row/cell selection, the chart builder, and the row/column explorers. The table looks interactive but isn't. The frontend now detects `isStaticNotebook()` at render time and hides those controls, so an exported table exposes only what works without a kernel.

The entire top bar is hidden (search, filters access, chart builder, explorers, download, size), pagination is dropped from the bottom bar, the column summaries RPC is skipped, and the selection column is removed. Controls that need no kernel stay: the row and column counts, client-side cell-selection stats, and data-type badges. Download is hidden along with the top bar because `download_as` is itself a kernel RPC and does not resolve in a static export, so the issue's note that download is unaffected does not hold in practice.

Because a static export only embeds the first page, a banner now reports how many rows are visible out of the total and points authors at `page_size` to embed more.

Verified against a real `marimo export html` snapshot and covered by frontend tests that mock `isStaticNotebook()`.

### Static table

Closes MO-6534, https://github.com/marimo-team/marimo/issues/9387
Screenshot 2026-06-26 at 3 23 51 PM

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/10007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/data-table/TableBottomBar.tsx
+++ b/frontend/src/components/data-table/TableBottomBar.tsx
@@ -3,6 +3,7 @@
 
 import type { RowSelectionState, Table } from "@tanstack/react-table";
 import { useLocale } from "react-aria";
+import { isStaticNotebook } from "@/core/static/static-state";
 import type { GetRowIds } from "@/plugins/impl/DataTablePlugin";
 import { cn } from "@/utils/cn";
 import { Events } from "@/utils/events";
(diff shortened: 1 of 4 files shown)


mscolnick (commented):
Very nice

mscolnick (approved):

Issue #10030, opened 2026-06-30
Long-lived terminal process irrecoverably freezes marimo on browser disconnect


ktaletsk:
## Describe the bug

While running long-lived tasks in marimo's built-in terminal, such as Claude Code (`claude`), I noticed that marimo would sometimes crash when I step away from my laptop. I was able to pin it down to any browser disconnects, such as a browser page refresh while it was mid-response. In all such cases the entire marimo server freezes permanently:

- the page shows the broken-link icon and a "Failed to connect" banner; refreshing just hangs and never reconnects
- the CLI prints no error and looks like it is still running
- `Ctrl-C` does nothing (`^C^C^C` is ignored). The only way out is `kill -9`, which also leaks multiprocessing semaphores and shared memory at shutdown

https://github.com/user-attachments/assets/73e7eb8b-868f-4b6b-a727-98fa804fe4c2

### Root cause

The root cause starts with a design decision: marimo forks a fresh shell for each terminal WebSocket connection and tears it down in that connection's `finally` block on any disconnect (tab close, page reload, network drop, laptop sleep). There is no persistence or reattach across connections.

- per-connection shell: [`pty.fork()` in `websocket_endpoint`](https://github.com/marimo-team/marimo/blob/ea3c57c64/marimo/_server/api/endpoints/terminal.py#L384)
- unconditional teardown on disconnect: [`finally: cleanup_child()`](https://github.com/marimo-team/marimo/blob/ea3c57c64/marimo/_server/api/endpoints/terminal.py#L419-L433)

On its own this makes the terminal unsuitable for long-running work such as a coding agent (`claude`), where a single refresh or sleep silently kills the session. That behavior is reported separately in #7829 (Terminal does not persist), which is currently closed.

Assuming kill-on-disconnect stays, the freeze comes from how that teardown is implemented. The cleanup handler runs a blocking `os.waitpid(child, 0)` on the asyncio event loop. The shell is the pty's session leader. After it is `SIGKILL`ed, the kernel cannot finish revoking its controlling terminal while (a) the server still holds the pty master fd open and (b) the foreground child (`claude`) still holds the slave. So the shell never becomes a reapable zombie and `waitpid` blocks forever. That freezes the whole event loop: with no websocket heartbeats the browser shows "Failed to connect", and asyncio's SIGINT handler never runs, so `Ctrl-C` is ignored.

In `marimo/_server/api/endpoints/terminal.py`, `_create_process_cleanup_handler`:

```python
[...]

Issue #7829, opened 2026-01-14
Terminal does not persist


jp-gr:
### Describe the bug

Hello,

I am using Marimo 0.19.2, when I have the developer tools pane open on the terminal and then click e.g. Errors and back to the terminal, I get a whole new shell. 

I'm using Marimo with Claude in that terminal, so it interrupts my workflow when I have to re-init claude etc when clicking around.

Thanks!

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
o$ marimo env
{
  "marimo": "0.19.2",
  "editable": false,
  "location": "/persist/perfengmarimo/.venv/lib/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "5.15.0-91-generic",
  "Processor": "x86_64",
  "Python Version": "3.12.12",
  "Locale": "en_GB",
  "Binaries": {
    "Browser": "142.0.7444.162",
    "Node": "v22.21.0"
  },
  "Dependencies": {
    "click": "8.3.0",
    "docutils": "0.22.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.9",
    "narwhals": "2.9.0",
    "packaging": "25.0",
    "psutil": "7.1.2",
    "pygments": "2.19.2",
    "pymdown-extensions": "10.16.1",
    "pyyaml": "6.0.3",
    "starlette": "0.41.3",
    "tomlkit": "0.13.3",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.32.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "altair": "5.5.0",
    "anywidget": "0.9.18",
    "basedpyright": "1.32.1",
    "loro": "1.8.2",
    "pandas": "2.3.3",
    "polars": "1.34.0",
    "pyarrow": "22.0.0",
    "pytest": "8.4.2",
    "ruff": "0.14.2"
  },
  "Experimental Flags": {}
}
```

</details>

mscolnick:
hey @jp-gr, thanks for filing. we will look into this

PR #10031, opened 2026-06-30
fix(terminal): close pty master before reaping to avoid event-loop deadlock

ktaletsk:
## What

Fixes the event-loop deadlock described in #10030. Disconnecting the terminal while a foreground process holds the pty (for example a coding agent like `claude`) could freeze the entire marimo server: "Failed to connect", `Ctrl-C` ignored, recoverable only with `kill -9`.

## Why

`_create_process_cleanup_handler` in `marimo/_server/api/endpoints/terminal.py` runs a blocking `os.waitpid(child, 0)` on the asyncio event loop after `SIGKILL`ing the shell. The shell is the pty's session leader. It cannot be reaped while the server still holds the pty master fd open and a foreground child still holds the slave, so the kernel never finishes revoking its controlling terminal, the shell never becomes a reapable zombie, and `waitpid` blocks forever. That freezes the whole event loop. See #10030 for the full root cause and a deterministic reproduction.

## Change

Close the pty master fd before reaping. With the master closed, the controlling-terminal teardown can complete and the child is reaped within a few milliseconds, so `waitpid` returns instead of blocking.

## Test

Adds `test_cleanup_closes_master_before_blocking_waitpid`, which asserts `os.close(fd)` runs before the blocking `os.waitpid(pid, 0)`. It fails on the previous ordering and passes with this change. The existing `terminal.py` test suite still passes.

## Scope

This PR addresses only the freeze. The separate behavior of killing the terminal shell on any browser disconnect (#7829) is out of scope.

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/10031?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/marimo/_server/api/endpoints/terminal.py
+++ b/marimo/_server/api/endpoints/terminal.py
@@ -114,6 +114,21 @@ def _create_process_cleanup_handler(
     """Create a cleanup handler for the child process and file descriptor."""
 
     def cleanup() -> None:
+        # Close the pty master *first*.
+        #
+        # child_pid is the session leader of the pty. If it is killed while the
+        # server still holds the master fd open -- and a foreground child (e.g.
+        # a coding agent or REPL) still holds the slave -- the kernel cannot
+        # finish revoking the shell's controlling terminal, so the shell never
+        # becomes a reapable zombie. The blocking os.waitpid() below would then
+        # hang the asyncio event loop forever (frozen server, Ctrl-C ignored).
+        # Closing the master is the hangup that lets the teardown complete, so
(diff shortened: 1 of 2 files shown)


ktaletsk:

mscolnick (approved):

PR #10032, opened 2026-06-30
feat(kiosk): enforce consumer capabilities server-side

kirangadhave:
## 📝 Summary

Make `ConsumerCapabilities` the server's real authority instead of a cosmetic client hint. Previously a viewer's read-only status was enforced only by the frontend hiding edit affordances; every connection in an edit session still carried the edit scope, so the server honored its requests.

Add a pure capability table mapping each kernel command to the capability it requires (`read`/`interact`/`edit`), behind a single `consumer_can` predicate. Unknown commands fail closed to `edit`. The `Room` stores each consumer's capabilities keyed by `ConsumerId`, stamped on connect and restamped on takeover. Enforce the predicate at two sites: the `put_control_request` chokepoint, which is the authority and closes the takeover race by dropping a denied command before it reaches the event bus, and the HTTP handlers, which return an advisory `403` so a viewer gets an honest refusal rather than a silent `200`.

Default secondary connections to interactors (`edit: false, interact: true`): they can drive shared UI state but cannot edit the notebook. Pure read-only (`interact: false`) is opt-in via a deployment's capability provider (e.g. in future molab might have a user who is purely READ only, no interaction allowed). **The difference in interact vs read-only becomes more clear when we make a decision on if we want interactive widgets to share state across consumers or not.**

Scoped to enforcement. The pluggable `CapabilityProvider` seam is a follow-up, and cross-consumer view-state syncing (e.g. table sort) is tracked in MO-6567.

Part of MO-6311

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/10032?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/marimo/_messaging/notification.py
+++ b/marimo/_messaging/notification.py
@@ -294,7 +294,10 @@ class ConsumerCapabilities(msgspec.Struct, frozen=True):
     """Per-consumer access capabilities for a session connection.
 
     - editor: `{edit: True, interact: True}`
-    - viewer: `{edit: False, interact: False}`
+    - interactor: `{edit: False, interact: True}` (default for a secondary
+      connection: drives UI state but cannot edit the notebook)
+    - read-only viewer: `{edit: False, interact: False}` (opt-in, set by a
+      deployment's capability provider)
 
     The server enforces these: control requests are gated against the issuing
     consumer's stored capabilities at the control-request chokepoint (the
--- a/marimo/_messaging/notification.py
+++ b/marimo/_messaging/notification.py
@@ -302,8 +302,8 @@ class ConsumerCapabilities(msgspec.Struct, frozen=True):
     The server enforces these: control requests are gated against the issuing
     consumer's stored capabilities at the control-request chokepoint (the
     authority) and mirrored as an advisory HTTP 403 at the request handlers.
-    `read`-tier actions (completions, non-mutating function calls, previews)
-    are always permitted.
+    Commands classified as `read` in `marimo._session.capabilities` (such as
+    completions and previews) are always permitted.
     """
 
     edit: bool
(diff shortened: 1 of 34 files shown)


marimo/_messaging/notification.py
mscolnick, line 293: could make these static properties or static methods. not sure if that will break future ideas of how you are thinking about these capabilities. it would make the usage downstream more readable

```py
class ConsumerCapabilities
    EDITOR = ConsumerCapabilities(...)
    INTERACTOR = ConsumerCapabilities(...)
    VIEWER = ConsumerCapabilities(...)
```

also may want to make the dataclass `frozen`

marimo/_session/capabilities.py
```diff
@@ -0,0 +1,80 @@
...
+    {
+        commands.UpdateUIElementCommand,
+        commands.ModelCommand,
+        commands.InvokeFunctionCommand,
+    }
+)
+
+# Read-tier commands never mutate notebook or reactive state and never invoke
+# user code: completions and data/secret previews and listings.
+_READ_COMMANDS: frozenset[type] = frozenset(
```
mscolnick, line 42: maybe a test that checks if we have full coverage of all commands? should be able to traverse the module `commands`

kirangadhave: oh that's a good idea

mscolnick (dismissed):

mscolnick (approved):

PR #10036, opened 2026-07-01
fix(server): stop scratchpad console cross-delivery between concurrent executes

### LINKED ISSUE #9255: marimo pair: Claude Code doesn't see cell errors automatically

### LINKED ISSUE #10035: Concurrent scratchpad executions on one session cross-deliver console (stdout/stderr) output
Follow-up to [#9350](<https://github.com/marimo-team/marimo/issues/9350>) (which resolved marimo-team/marimo#9255). [#9350](<https://github.com/marimo-team/marimo/issues/9350>) scoped the scratchpad<br>**completion sentinel** by `run_id`, so `/api/kernel/execute` waits for the<br>correct run before emitting `done` (fixing the dropped/"empty output" case).<br>It did not change how **console** output is delivered — and when two scratchpad<br>executions **overlap on the same session**, one execution's SSE stream can<br>receive the *other* execution's stdout/stderr.

* **Impact:** low for serial use — one execution at a time is reliable after<br>[#9350](<https://github.com/marimo-team/marimo/issues/9350>). It manifests only when executions **overlap** on one session: a tool<br>firing concurrent executes, or a new execute fired while a prior reactive<br>cascade is still emitting.
* **Affected paths** (all construct a `ScratchCellListener`): the HTTP<br>`POST /api/kernel/execute` SSE endpoint, the in-server AI `execute_code` tool<br>(`run_scratchpad_code`), and the MCP `execute_code` code server.
* Reproduced on `0.23.11`; **confirmed on** `main` (root cause below).

## Root cause

Scratchpad executions on a session are already serialized by<br>`session.scratchpad_lock`. The problem is *listener registration ordering*:<br>each entry point registers its `ScratchCellListener` via<br>`session.scoped(listener)` **before** acquiring the lock, e.g. in<br>`marimo/_server/api/endpoints/execution.py`:

```python
with session.scoped(listener):          # listener is now live
    async with session.scratchpad_lock:  # a 2nd execute blocks here
        ...
        async for event in listener.stream():
            yield event
```

`session.extensions` delivers **every** `CellNotification` to **every**<br>attached listener. So while a second execute *B* waits on the lock, *B*'s<br>listener is already attached and accumulates the currently-running execute<br>*A*'s console (and child-cell errors) into *B*'s queue. When *B* finally<br>acquires the lock and starts streaming, it replays *A*'s buffered output.

## Why not just filter the console by `run_id` (the natural follow-up to [#9350](<https://github.com/marimo-team/marimo/issues/9350>))?

Because scratch console is **not tagged with the run's** `run_id`, so an<br>equality filter on the console branch would be a no-op (and would wrongly drop<br>legitimate output):

* `RUN_ID_CTX` is **not set** during `handle_execute_scratchpad`:<br>`kernel.run_scratchpad()` → `runner.run_all()` never enters `run_id_context()`<br>(the only `run_id_context()` lives in `_run_cells`, which the scratch cell<br>bypasses). So the scratch cell's `CellNotification`s — including the one<br>carrying its stdout — have `run_id=None`. Only the `CompletedRunNotification`<br>is stamped with the request's `run_id`, which is exactly what [#9350](<https://github.com/marimo-team/marimo/issues/9350>) relies on.
* Child cells run by `_code_mode` go through `kernel._run_cells(...)`, which<br>mints a **fresh** `run_id` per call — not the request's — so filtering the<br>console by `msg.run_id == self._run_id` would drop a listener's own<br>child-cell console.

## Fix direction

Register the listener **inside** the `scratchpad_lock` in all three entry<br>points, so a waiting execute has no live listener accumulating another run's<br>output. Minimal and localized; no schema or frontend changes. (A PR is on the<br>way.)

## A related but separate path

The frontend scratchpad endpoint `POST /api/kernel/scratchpad/run` is a<br>different mechanism: fire-and-forget, no `ScratchCellListener`, and it does not<br>take `scratchpad_lock` (it dispatches via `dispatch_control_request` and returns<br>immediately, before execution completes). It is therefore not one of the<br>listener-backed paths this issue is about. It is, however, a scratch-execution<br>*source*: if it runs concurrently with a listener-backed execute on the same<br>session, its untagged notifications (`run_id=None`, shared `SCRATCH_CELL_ID`)<br>can still be captured by that listener. Fully isolating it would require making<br>it lock-aware / wait for completion (a semantics change), so it's out of scope<br>for the fix above and likely warrants separate discussion.

## Workaround for clients today

Serialize executes and let the kernel go idle before firing the next.

## Will you submit a PR?

Yes.

## Environment

`marimo env` from a development checkout at current `main` (reports version<br>`0.23.11`, the current release), where the bug was confirmed by code read +<br>tests; originally reproduced on the `0.23.11` release.

<details><summary>\`\`\`json</summary>

{<br>  "marimo": "0.23.11",<br>  "editable": true,<br>  "OS": "Windows",<br>  "OS Version": "11",<br>  "Processor": "Intel64 Family 6 Model 186 Stepping 2, GenuineIntel",<br>  "Python Version": "3.13.13",<br>  "Locale": "--",<br>  "Binaries": {<br>    "Browser": "149.0.7827.197",<br>    "Node": "v24.16.0",<br>    "uv": "0.11.19 (7b2cff1c3 2026-06-03 x86_64-pc-windows-msvc)"<br>  },<br>  "Dependencies": {<br>    "click": "8.2.1",<br>    "docutils": "0.23",<br>    "itsdangerous": "2.2.0",<br>    "jedi": "0.19.2",<br>    "markdown": "3.10.2",<br>    "narwhals": "2.22.1",<br>    "packaging": "26.2",<br>    "psutil": "7.2.2",<br>    "pygments": "2.20.0",<br>    "pymdown-extensions": "10.21.3",<br>    "pyyaml": "6.0.3",<br>    "starlette": "1.3.1",<br>    "tomlkit": "0.15.0",<br>    "typing-extensions": "4.15.0",<br>    "uvicorn": "0.49.0",<br>    "websockets": "16.0"<br>  },<br>  "Optional Dependencies": {<br>    "altair": "6.2.2",<br>    "anywidget": "0.9.21",<br>    "duckdb": "1.5.4",<br>    "loro": "1.13.1",<br>    "mcp": "1.28.0",<br>    "nbformat": "5.10.4",<br>    "openai": "2.43.0",<br>    "pandas": "3.0.3",<br>    "polars": "1.41.2",<br>    "pyarrow": "24.0.0",<br>    "pytest": "9.0.3",<br>    "ruff": "0.15.18",<br>    "sqlglot": "30.11.0"<br>  },<br>  "Experimental Flags": {<br>    "cache_panel": true,<br>    "multi_column": true,<br>    "isolate_apps": true,<br>    "external_agents": false<br>  }<br>}

```
```

</details>

## Code to reproduce

Two overlapping `/api/kernel/execute` calls against one running kernel. Here<br>`execute-code.sh` is a thin wrapper that POSTs `{"code": "..."}` to<br>`/api/kernel/execute` on a session and prints the streamed SSE `stdout`/`stderr`<br>events. Start a kernel (`marimo edit --no-token --headless --port 2718`), then<br>fire a slow background execute and a quick one that overlaps it:

```bash
URL=http://127.0.0.1:2718
# background: sleep then print; quick call fired immediately after (overlaps)
( execute-code.sh --url $URL -c "import time; time.sleep(12); print('BG12')" >bg.out 2>&1 ) &
execute-code.sh --url $URL -c "print('FG_DURING_BG')"
```

* **Expected:** the quick call's SSE stream returns only `FG_DURING_BG`.
* **Actual:** it returns `BG12\nFG_DURING_BG` — it replayed the background run's<br>stdout.
* Tight-overlap pairs (`A`/`B`, each `sleep(4); print`) misattribute every time;<br>**25 serial calls → 0 cross-delivery** (concurrency-specific).

---

*Drafted with AI assistance (Claude Code) from a live reproduction on marimo*<br>`0.23.11` *plus a code-read and test-backed confirmation on* `main`*; reviewed by a**<br>**human before filing.*

MHouse:

## 📝 Summary

Closes #10035. Follow-up to #9350 (which resolved #9255).

#9350 scoped the scratchpad **completion sentinel** by `run_id`, making serial
`/api/kernel/execute` reliable. This PR fixes the remaining issue when
scratchpad executions **overlap on one session**: one execution's SSE stream
could replay another execution's console (stdout/stderr).

**Root cause — listener registration ordering, not a missing `run_id` filter.**
Scratchpad executions are already serialized by `session.scratchpad_lock`, but
each entry point registered its `ScratchCellListener` via
`session.scoped(listener)` *before* acquiring the lock. `session.extensions`
delivers every `CellNotification` to every attached listener, so a second
execute waiting on the lock had a live listener that accumulated the
in-progress execute's console into its queue and replayed it once it started
streaming.

**Fix.** Register the listener *inside* the `scratchpad_lock` so a waiting
execute has no live listener accumulating another run's output. Applied to all
three entry points that drive the scratchpad:

- `marimo/_server/api/endpoints/execution.py` — HTTP `POST /api/kernel/execute`
- `marimo/_server/scratchpad.py` (`run_scratchpad_code`) — the in-server AI `execute_code` tool
- `marimo/_mcp/code_server/main.py` — the MCP `execute_code` code server

The change is a localized swap of two context managers (plus explanatory
comments); no schema, public API, or frontend changes.

**Out of scope (pre-existing, flagged for discussion).** The frontend
`POST /api/kernel/scratchpad/run` endpoint is a separate, listener-less,
fire-and-forget path that does not take `scratchpad_lock` (it dispatches and
returns before execution completes). A scratchpad run started there can still
pollute a concurrently-scoped listener. Fully serializing it needs a lock-aware
redesign, so it's deliberately left out of this targeted fix — happy to follow
up separately if that's wanted.

<details>
<summary>Why not filter the console by <code>run_id</code>, mirroring #9350?</summary>

Because scratch console is not tagged with the run's `run_id`, so an equality
filter would be a no-op *and* would drop legitimate output:

- `RUN_ID_CTX` is not set during `handle_execute_scratchpad`
  (`kernel.run_scratchpad()` → `runner.run_all()` never enters
  `run_id_context()`), so the scratch cell's `CellNotification`s — including
  the one carrying its stdout — have `run_id=None`. Only the
  `CompletedRunNotification` is stamped with the request's `run_id`, which is
  what #9350 keys on.
- Child cells run by `_code_mode` go through `kernel._run_cells(...)`, which
  mints a *fresh* `run_id` per call, so filtering the console by
  `msg.run_id == self._run_id` would drop a listener's own child-cell console.

</details>

**Testing.** Added `tests/_server/test_scratchpad.py::TestConcurrentScratchpadIsolation`:
- a two-listener isolation test that drives two overlapping `run_scratchpad_code`
  calls through a real `scratchpad_lock` and asserts the waiting execute is not
  scoped and never receives the running execute's child-cell error. It is a
  verified regression guard — it fails (`assert 2 == 1`) on the pre-fix ordering.
- a guard that `run_id=None` console is never dropped (the fix relies on
  registration ordering, not `run_id` filtering).

All backend scratchpad tests pass locally: `test_scratchpad.py` (38),
`test_scratchpad_integration.py` (10), `tests/_mcp/code_server/test_code_server.py` (11).
`ruff check`, `ruff format`, and `mypy` are clean on the changed modules.

<sub>Developed with AI assistance (Claude Code) and reviewed line-by-line by the human PR author before submission.</sub>

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

<a href="https://cubic.dev/pr/marimo-team/marimo/pull/10036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/marimo/_server/api/endpoints/execution.py
+++ b/marimo/_server/api/endpoints/execution.py
@@ -359,8 +359,18 @@ async def execute_code(
         run_id = str(uuid4())
         try:
             listener = ScratchCellListener(run_id=run_id)
-            with session.scoped(listener):
-                async with session.scratchpad_lock:
+            # Register the listener *inside* the scratchpad lock. Scratch
+            # executions share one `SCRATCH_CELL_ID` and every scoped
+            # listener receives every cell notification; scratch console
+            # output is not tagged with our `run_id` (it rides on
+            # `CellNotification`s whose `run_id` is `None`), so it
+            # can't be filtered per-run. If we scoped the listener before
+            # acquiring the lock, a concurrent execute waiting on the lock
+            # would accumulate this run's console output in its queue and
+            # replay it once it began streaming. Taking the lock first
+            # keeps only the running execute's listener live.
+            async with session.scratchpad_lock:
+                with session.scoped(listener):
                     http_req = HTTPRequest.from_request(request)
                     server_url, auth_token = get_code_mode_credentials(
                         app_state, request
--- a/marimo/_server/api/endpoints/execution.py
+++ b/marimo/_server/api/endpoints/execution.py
@@ -359,16 +359,8 @@ async def execute_code(
         run_id = str(uuid4())
         try:
             listener = ScratchCellListener(run_id=run_id)
-            # Register the listener *inside* the scratchpad lock. Scratch
-            # executions share one `SCRATCH_CELL_ID` and every scoped
-            # listener receives every cell notification; scratch console
-            # output is not tagged with our `run_id` (it rides on
-            # `CellNotification`s whose `run_id` is `None`), so it
-            # can't be filtered per-run. If we scoped the listener before
-            # acquiring the lock, a concurrent execute waiting on the lock
-            # would accumulate this run's console output in its queue and
-            # replay it once it began streaming. Taking the lock first
-            # keeps only the running execute's listener live.
+            # Ensure we take a lock on the scratchpad before scoping the
+            # listener. See #10035.
             async with session.scratchpad_lock:
                 with session.scoped(listener):
                     http_req = HTTPRequest.from_request(request)
--- a/tests/_server/test_scratchpad.py
+++ b/tests/_server/test_scratchpad.py
@@ -965,3 +965,212 @@ class TestRunScratchpadCode:
 
         assert result.success is False
         assert result.errors == ["cell 'child-cell' raised ZeroDivisionError"]
+
+
+class _BroadcastingFakeSession:
+    """Fake `Session` modelling two overlapping scratch executes on ONE
+    session.
+
+    Unlike `_FakeSession`, this mirrors the production `Session` where
+    `extensions` (an `ExtensionRegistry`) delivers *every* notification to
+    *every* attached listener. It uses a real `scratchpad_lock` so two concurrent
+    `run_scratchpad_code` calls serialize exactly as in production, and it
+    lets the test drive each run's notifications explicitly via
+    `emit_child_error` / `complete`.
+    """
+
+    def __init__(self) -> None:
+        self.scratchpad_lock = asyncio.Lock()
+        self.scoped_listeners: set[ScratchCellListener] = set()
+        self.put_run_ids: list[str] = []
+        self.interrupt_count = 0
+        self.document = SimpleNamespace(cells=(), cell_ids=())
+        self.session_view = SimpleNamespace(
+            cell_notifications={},
+            get_cell_outputs=lambda _ids: {},
+            get_cell_console_outputs=lambda _ids: {},
+        )
+
+    def as_session(self) -> Session:
+        return cast("Session", cast(object, self))
+
+    def instantiate(
+        self,
+        request: InstantiateNotebookRequest,
+        *,
+        http_request: HTTPRequest | None,
+    ) -> None:
+        del request, http_request
+
+    @contextmanager
+    def scoped(
+        self, listener: ScratchCellListener
+    ) -> Iterator[ScratchCellListener]:
+        self.scoped_listeners.add(listener)
+        try:
+            yield listener
+        finally:
+            self.scoped_listeners.discard(listener)
+
+    def put_control_request(
+        self,
+        req: CommandMessage,
+        from_consumer_id: object = None,
+    ) -> None:
+        del from_consumer_id
+        if isinstance(req, ExecuteScratchpadCommand):
+            self.put_run_ids.append(req.run_id)
+
+    def try_interrupt(self) -> None:
+        self.interrupt_count += 1
+
+    # -- test drivers: broadcast to ALL currently-scoped listeners --------
+
+    def _broadcast(self, notif: NotificationMessage) -> None:
+        session = self.as_session()
+        msg = serialize_kernel_message(notif)
+        for listener in list(self.scoped_listeners):
+            listener.on_notification_sent(session, msg)
+
+    def emit_child_error(self, cell_id: str) -> None:
+        self._broadcast(
+            CellNotification(
+                cell_id=cast("CellId_t", cell_id),
+                output=CellOutput.errors(
+                    [
+                        MarimoExceptionRaisedError(
+                            msg="boom",
+                            exception_type="RuntimeError",
+                            raising_cell=None,
+                        )
+                    ]
+                ),
+                console=None,
+                status="idle",
+            )
+        )
+
+    def complete(self, run_id: str) -> None:
+        self._broadcast(CompletedRunNotification(run_id=run_id))
+
+
+class TestConcurrentScratchpadIsolation:
+    """Two overlapping scratch executes on one session must not
+    cross-deliver console / child-cell errors (follow-up to #9350,
+    which scoped only the completion sentinel by `run_id`).
+
+    Scratch executions share one `SCRATCH_CELL_ID` and their
+    `CellNotification`s are NOT tagged with the requesting run's
+    `run_id` (it is `None`), so the listener cannot filter them per run.
+    Isolation instead relies on registering the listener only while the
+    `scratchpad_lock` is held — a second execute waiting on the lock must
+    not have a live listener accumulating the running execute's output.
+    """
+
+    @staticmethod
+    async def _advance(times: int = 20) -> None:
+        for _ in range(times):
+            await asyncio.sleep(0)
+
+    @pytest.mark.asyncio
+    async def test_waiting_execute_is_not_scoped_and_stays_isolated(
+        self,
+    ) -> None:
+        session = _BroadcastingFakeSession()
+        # extract_result short-circuits unless the scratch cell has a
+        # notification; give it an empty idle output (shared cell id).
+        session.session_view.cell_notifications[SCRATCH_CELL_ID] = (
+            CellNotification(
+                cell_id=SCRATCH_CELL_ID,
+                output=CellOutput(
+                    channel=CellChannel.OUTPUT,
+                    mimetype="text/plain",
+                    data="",
+                ),
+                console=None,
+                status="idle",
+            )
+        )
+
+        # Run A starts and holds the lock mid-execution.
+        task_a = asyncio.create_task(
+            run_scratchpad_code(
+                session.as_session(),
+                _build_request(),
+                code="A",
+                server_url="u",
+                auth_token="t",
+            )
+        )
+        await self._advance()
+        assert len(session.put_run_ids) == 1  # A put its command
+        assert len(session.scoped_listeners) == 1  # ...and is scoped
+        run_a = session.put_run_ids[0]
+
+        # Run B starts while A is still running.
+        task_b = asyncio.create_task(
+            run_scratchpad_code(
+                session.as_session(),
+                _build_request(),
+                code="B",
+                server_url="u",
+                auth_token="t",
+            )
+        )
+        await self._advance()
+
+        # The fix: B must NOT be scoped while it waits on the lock, so it
+        # can't accumulate A's output. Before the fix, B scoped its
+        # listener before acquiring the lock, so this would be 2 and B
+        # would receive A's child error below.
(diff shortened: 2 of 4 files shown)


MHouse:

marimo/_server/api/endpoints/execution.py
```diff
@@ -359,8 +359,18 @@ async def sse_generator() -> AsyncGenerator[str, None]:
...
+            # Register the listener *inside* the scratchpad lock. Scratch
+            # executions share one `SCRATCH_CELL_ID` and every scoped
+            # listener receives every cell notification; scratch console
+            # output is not tagged with our `run_id` (it rides on
+            # `CellNotification`s whose `run_id` is `None`), so it
+            # can't be filtered per-run. If we scoped the listener before
+            # acquiring the lock, a concurrent execute waiting on the lock
+            # would accumulate this run's console output in its queue and
+            # replay it once it began streaming. Taking the lock first
+            # keeps only the running execute's listener live.
```
manzt: These three comment blocks restate the same rationale in full for what is essentially a one-line change. A code comment should orient someone reading this line, but the full "why not filter by run_id / #9350 follow-up" reasoning is what #10035 and this PR are for.

I'd recommend something short like "Ensure we take a lock on the scratchpad before scoping the listener" or even just something that points at the issue.

tests/_server/test_scratchpad.py
```diff
@@ -965,3 +965,212 @@ async def test_child_cell_errors_flow_into_result_errors(self) -> None:
 
         assert result.success is False
         assert result.errors == ["cell 'child-cell' raised ZeroDivisionError"]
+
+
+class _BroadcastingFakeSession:
```
manzt: This mostly duplicates `_FakeSession`, and `_broadcast` just loops over `scoped_listeners`, so "B saw nothing of A's" is basically true by construction.

Could you reuse the existing `_FakeSession` (it already has `scratchpad_lock`) and just assert the listener is scoped while the lock is held, like `test_holds_scratchpad_lock_while_putting_execute_command` does? 

Something like:

```py
@pytest.mark.asyncio
async def test_listener_registered_only_while_lock_held(self) -> None:
    session = _FakeSession()
    locked_when_scoped: list[bool] = []
    original_scoped = session.scoped

    @contextmanager
    def recording_scoped(listener: ScratchCellListener):
        locked_when_scoped.append(session.scratchpad_lock.locked())
        with original_scoped(listener) as l:
            yield l

    session.scoped = recording_scoped  # type: ignore[method-assign]
    await run_scratchpad_code(
        session.as_session(),
        _build_request(),
        code="x",
        server_url="u",
        auth_token="t",
    )
    assert locked_when_scoped == [True]
```

manzt (changes requested):
Thanks for contribution! A few comments.

marimo/_server/api/endpoints/execution.py
MHouse: Thanks.  I'm adopting your wording plus a pointer to #10035 in all three files.

tests/_server/test_scratchpad.py
```diff
@@ -965,3 +965,212 @@ async def test_child_cell_errors_flow_into_result_errors(self) -> None:
...
+        try:
+            yield listener
+        finally:
+            self.scoped_listeners.discard(listener)
+
+    def put_control_request(
+        self,
+        req: CommandMessage,
+        from_consumer_id: object = None,
+    ) -> None:
```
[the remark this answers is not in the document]
MHouse: Goes away with the test rewrite

tests/_server/test_scratchpad.py
MHouse: Thanks @manzt .  That makes sense.  The extra class was way too heavy.  I'm leaving a slightly longer docstring here, matching what it looks like the other tests do.  I renamed the yield value to pass linting rules, added a return type annotation, and am tweaking the `code=` line to match similar tests.

tests/_server/test_scratchpad.py
```diff
@@ -707,6 +707,35 @@ async def consume() -> None:
         assert name == "stdout"
         assert payload["data"] == "partial\n"
 
+    @pytest.mark.asyncio
+    async def test_run_id_none_console_is_not_dropped(self) -> None:
```
MHouse: Dropping this, as it was only added as an intent guard.  Substantially overlaps with the existing test_stream_basic test.

MHouse (commented):
Submitting my comments batch.  Thanks for the feedback to a newbie.  This is much leaner and cleaner than it previously was.

eeshsaxena (commented):
This ordering swap looks correct to me. I traced `scoped()`:

```python
def scoped(self, extension):
    self.extensions.add(extension)          # now live on the shared session
    extension.on_attach(self, self._event_bus)
    try:
        yield extension
    finally:
        ...  # detach
```

So a `ScratchCellListener` becomes active on the shared `session.extensions` the moment it's scoped. With the old `with session.scoped(listener): async with session.scratchpad_lock:` order, execution A attaches its listener and *then* blocks waiting for the lock held by B - and during that wait A's listener is already live on the shared event bus, so B's scratchpad console output is delivered to A too. Acquiring the lock first and scoping the listener inside it means a listener is only ever attached while its execution holds the lock, which the lock already serializes - so no cross-delivery. Makes sense.

Two things I liked: the fix is applied consistently at all three call sites (`_mcp/code_server/main.py`, `execution.py`, `scratchpad.py`), and the actual work (`snapshot_for_scratchpad` + `put_control_request`) stays inside both context managers, so only the listener-attach window narrows - the execution semantics are unchanged. No lock-ordering concern either, since `scoped()` is a plain registration, not a second lock.

Deferring to the maintainers.

manzt (approved):
Looks great!

PR #10050, opened 2026-07-02
fix: show sql snippet cells as sql instead of python

### LINKED ISSUE #6686: SQL formatting in snippets
### Description

Snippets of cells that use mo.sql display as Python in the snippet. When added, they become SQL cells.

### Suggested solution

Cells that use only mo.sql should be formatted as SQL in the snippet in the side bar. This means users of the snippet aren't confused when the snippet is inserted and the cell becomes SQL formatted. It also makes the SQL query more clear within the snippet, rather than as a Python string.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

qorexdevs:
## 📝 Summary

Closes #6686

Snippet cells that use `mo.sql(...)` were shown with Python highlighting in the snippets panel, so the query read as a Python string. This detects SQL sections with the SQL language adapter and renders the extracted query with SQL highlighting, matching how the cell looks once the snippet is inserted. Plain Python cells are unchanged, and "Insert snippet" still creates the original `mo.sql` cell.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/frontend/src/components/editor/chrome/panels/snippets-panel.tsx
+++ b/frontend/src/components/editor/chrome/panels/snippets-panel.tsx
@@ -24,6 +24,7 @@ import { Button } from "@/components/ui/button";
 import { Tooltip } from "@/components/ui/tooltip";
 import { useCellActions } from "@/core/cells/cells";
 import { useLastFocusedCellId } from "@/core/cells/focus";
+import { LanguageAdapters } from "@/core/codemirror/language/LanguageAdapters";
 import { useRequestClient } from "@/core/network/requests";
 import { LazyAnyLanguageCodeMirror } from "@/plugins/impl/code/LazyAnyLanguageCodeMirror";
 import { useTheme } from "@/theme/useTheme";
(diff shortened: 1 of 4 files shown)


qorexdevs:

Light2Dark (changes requested):
There's a chance this mutates global state, sql.transformIn has some side effects.

Could you look at building a util function from `cell-preview.ts`. It uses the side-effect-less methods. I would also strongly type `language: string` rather than just string.

qorexdevs:
good catch on the side effects. switched to SQLParser from smart-cells (same one cell-preview.ts uses, no global mutation) and typed the return language as "python" | "sql"

frontend/src/components/editor/chrome/panels/snippets-panel.tsx
```diff
@@ -34,6 +35,23 @@ import { usePanelOrientation, usePanelSection } from "./panel-context";
 
 const extensions = [EditorView.lineWrapping];
 
+const sqlParser = new SQLParser();
+
+// SQL cells are stored as python `mo.sql(...)`. Show the query itself
+// highlighted as SQL instead of as a python string, matching how the cell
+// renders once the snippet is inserted. Uses the side-effect-free parser
+// from smart-cells rather than the editor LanguageAdapters.
```
Light2Dark:
```suggestion
// renders once the snippet is inserted.
```

Light2Dark:
@cubic-dev-ai review

frontend/src/components/editor/chrome/panels/snippets-panel.tsx
qorexdevs: done, trimmed the comment

qorexdevs:
pulled the logic out into snippet-display.ts (same style as cell-preview.ts) and typed the language with a SnippetLanguage union. transformIn builds a fresh metadata object each call so no shared state gets mutated. moved the test alongside it too

qorexdevs:
moved it into a snippet-display util mirroring cell-preview.ts, and typed language as a python|sql union instead of string. transformIn here builds fresh metadata per call and doesn't touch shared state, same as cell-preview already does. lmk if you had a lighter method in mind.

Light2Dark (approved):

Light2Dark:
thanks!

PR #10068, opened 2026-07-02
feat(table): add "Hide all" columns action

kirangadhave:
Added a Hide All to the Columns menu in table header and Column Explorer

https://github.com/user-attachments/assets/ac957a26-28fd-448e-b2e3-93c7c6189826

### DIFF
--- a/frontend/src/components/data-table/column-visibility-dropdown.tsx
+++ b/frontend/src/components/data-table/column-visibility-dropdown.tsx
@@ -59,6 +59,12 @@ export const ColumnVisibilityDropdown = <TData,>({
   const hiddenIds = userColumns
     .filter((column) => !column.getIsVisible())
     .map((column) => column.id);
+  const hideableIds = userColumns
+    .filter((column) => column.getCanHide())
+    .map((column) => column.id);
+  const visibleHideableCount = hideableIds.filter(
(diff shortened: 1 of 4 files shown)


frontend/src/components/data-table/column-visibility-dropdown.tsx
```diff
@@ -59,6 +59,12 @@ export const ColumnVisibilityDropdown = <TData,>({
   const hiddenIds = userColumns
     .filter((column) => !column.getIsVisible())
     .map((column) => column.id);
+  const hideableIds = userColumns
+    .filter((column) => column.getCanHide())
+    .map((column) => column.id);
+  const visibleHideableCount = hideableIds.filter(
+    (id) => !hiddenIds.includes(id),
+  ).length;
```
mscolnick, line 67: super small perf thing, but i think this will be computed for each column and we are doing O(n) operations. 

maybe we could hoist this up and convert it to a Set?

mscolnick: oh nvm, this is only once (not per column). but we are still doing O(n) operations so a Set would still help, although its likely hidden Ids is not large at all, so could just be a pre-optimization

mscolnick (approved):

PR #10071, opened 2026-07-03
feat: include local wheels in WASM exports

### LINKED ISSUE #5488: Importing or including local module files in a way that is supported by WASM export
### Description

I tried building a notebook that imported some functions from local Python modules as a WASM stand-alone app, but it failed to import these as they weren't packages. I also tried a hack replacing the usual `import` with `importlib` calls using the path to the module files with `marimo.notebook_location()`, but to no avail.

### Suggested solution

One easy way would be to allow for some special `marimo.local_import` function that in WASM compilation gets unfolded at compile time, simply bringing in the desired module code as its own hidden cell. Otherwise, I would imagine that shipping Python files as if they were inert data and using dynamic import should also work.

### Will you submit a PR?

- [ ] Yes

### Alternative

_No response_

### Additional context

_No response_

peter-gy:
## Summary

`marimo export html-wasm` now packages local Python modules as wheels and installs them in Pyodide when the exported notebook starts.

## How it works

1. During export, Ruff resolves local Python modules reachable from the notebook, including imports used by those modules.
2. marimo builds a pure-Python wheel for each module and preserves its import shape: `foo.py` remains a top-level module and packages keep their file layout.
3. The export copies the wheels into `public/wheels` and rewrites PEP 723 dependencies to hosted wheel URLs. Pyodide installs them through `micropip` at startup.

### Notebook importing three different local modules

Screenshot 2026-07-09 at 22 50 43

### Export producing dynamically-built wheels from the local references

```shell
marimo export html-wasm notebooks/e2e/main.py -o notebooks/e2e/out
```

\-->

```text
wasm-local-pkg-poc ➤ tree notebooks/e2e/out/public                                                 git:main*
notebooks/e2e/out/public
└── wheels
    ├── bar-0.0.0+marimo.0ead6e55c72b-py3-none-any.whl
    ├── baz-0.0.0+marimo.07b3649a28f8-py3-none-any.whl
    └── foo-0.0.0+marimo.01c3b13570e3-py3-none-any.whl

2 directories, 3 files
```

### Functioning WASM export resolving local modules from wheels

Screenshot 2026-07-09 at 22 52 30

Closes marimo-team/marimo#5488


marimo/_cli/export/local_wheels.py
```diff
@@ -0,0 +1,464 @@
...
+        return
+
+    with tempfile.TemporaryDirectory(prefix="marimo-html-wasm-wheels-") as tmp:
+        wheel_dir = Path(tmp)
+        yield tuple(
+            write_pure_python_wheel(module, wheel_dir) for module in modules
+        )
+
+
+def write_pure_python_wheel(
```
dmadisetti, line 207: Since uv is a requisite I wonder if we could just use `uv build`? I think it would clean up a bunch of this boilerplate- but push back if there's something I'm not seeing

dmadisetti (commented):
More of a question than a comment, but aside from that looks well scoped. Awesome job!

marimo/_cli/export/local_wheels.py
peter-gy: yeah, fair point. in the first version I started with `uv build` instead of handrolling wheel builds.

my concern was that `uv build` with the native `uv_build` backend expects a layout like `foo/__init__.py` instead of a top-level `foo.py`, and I wanted the packaged local source files preserved as much as possible instead of remapping them to new files.

The workable `setuptools` version would also produce wheels with different hashes for the same local module unless we add reproducibility handling. With the current manual approach we control that directly through the pinned `_DIST_INFO_TIMESTAMP = (1980, 1, 1, 0, 0, 0)`.

And, perf-wise, calling out to `uv build` makes the process ~6x slower.

dmadisetti: Thanks for the context. There's also the risk of becoming too UV tied- so I think this seems like a reasonable compromise to strike. We can always generalize the graph search method

dmadisetti (approved):

PR #10096, opened 2026-07-08
fix(document): reconcile inbound transactions so stale cell refs don't 500

kirangadhave:
## Summary
Update notebook document transactions to remove stale references before apply.

## Why
We have `Failed to sync document transaction` errors frequently. FE holds the optimistic copy of notebook document. async messaging can cause drift in the two copies, and a transaction referencing an non-existent cell on the backend causes the error.

`NotebookDocument.apply()` raises and we get the above error.

## Solution
Added a pure `reconcile_transaction` function which resolves each change against current notebook document.:
- Dead create/move anchor cells are stripped
- Changes to a missing cell are no-ops
- Create for existing cell is dropped

Will have follow up PRs for:
- making `NotebookDocument.apply()` atomic
- resync the FE after a failed application

Closes MO-6572


marimo/_messaging/notebook/reconcile.py
```diff
@@ -0,0 +1,92 @@
...
+            if change.cell_id in (change.after, change.before):
+                continue
+            if not resolve_anchor(change.after) or not resolve_anchor(
+                change.before
+            ):
+                continue
+            out.append(change)
+        elif isinstance(change, (SetCode, SetName, SetConfig)):
+            # if cell is missing, skip
+            if change.cell_id not in live:
```
manzt, line 113: should we log something? when would this happen?

manzt (dismissed):
Looks great. Thanks kiran!

marimo/_messaging/notebook/reconcile.py
kirangadhave: Ideally should never happen, but if we have a SetCode on a cell which is removed previously, apply will throw.

i'll add a log.

manzt (approved):
Great!

PR #10097, opened 2026-07-08
feat(charts): allow selecting a pandas DataFrame index as a chart axis

### LINKED ISSUE #10017: Ability to select Pandas DataFrame's index as x-axis for visualizations
I love the "ad-hoc" _Visualize_ feature of Marimo! However, what bothers me (I might also just have overlooked it) is the fact that I can't choose my index column (named `date.utc` in the example) as the x-axis, which is usually what you'd want to do especially for time-series.

Image

Using Marimo 0.23.11.

kirangadhave:
### Why
The chart builder only offered regular columns as axes, so a named or otherwise non-trivial pandas index (a `date.utc` time-series index, a MultiIndex) could not be picked as an x-axis.

### Fix
Flatten a non-trivial pandas index into columns on the chart-data path only, and offer those index levels as fields in the chart builder. A new `TableManager.with_index_as_columns()` exposes the row index as regular columns, defaulting to a no-op for every backend except pandas, and is applied in `_get_data_url` (the RPC the chart builder pulls its data through).

On the frontend, `mergeIndexFields` appends the separately transmitted row-header fields to the offered field list, skipping any whose name already matches a data column.

Index-level column naming is unified in one helper: named levels keep their name, unnamed levels become `Index0`, `Index1`, ... (zero-based), and any name that collides with a real column gets a suffix `_index` (e.g. `Index0_index` in case `Index0` is existing column in the dataframe). Both the chart-data flatten and the offered-fields path go through this helper, so the columns in the data and the fields in the dropdown are in-sync.

As a side effect of the stable naming, an unnamed index is now freezable (`Index0`), so the old "Cannot freeze an unnamed row index" rejection is removed.

https://github.com/user-attachments/assets/1583a41c-e1c2-4814-94a9-ce982dc23f97

Closes #10017

### DIFF
--- a/frontend/src/components/data-table/charts/charts.tsx
+++ b/frontend/src/components/data-table/charts/charts.tsx
@@ -54,6 +54,24 @@ const CHART_HEIGHT = 290;
 const CHART_MAX_ROWS = 50_000;
 const CHART_MAX_COLUMNS = 50;
 
+/**
+ * Append row-header (index) fields to the chart field list, skipping any whose
+ * name already exists as a data column so the chart builder never offers the
+ * same axis twice.
+ */
+export function mergeIndexFields(
+  fieldTypes: FieldTypesWithExternalType | null | undefined,
+  rowHeaders: FieldTypesWithExternalType | null | undefined,
+): FieldTypesWithExternalType {
+  const base = fieldTypes ?? [];
+  if (!rowHeaders || rowHeaders.length === 0) {
+    return base;
+  }
+  const seen = new Set(base.map(([name]) => name));
+  const indexFields = rowHeaders.filter(([name]) => !seen.has(name));
+  return [...base, ...indexFields];
+}
+
 export interface TablePanelProps {
   cellId: CellId | null;
   data: unknown[];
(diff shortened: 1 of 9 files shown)


mscolnick (approved):

frontend/src/components/data-table/charts/charts.tsx
```diff
@@ -54,6 +54,24 @@ const CHART_HEIGHT = 290;
...
+  fieldTypes: FieldTypesWithExternalType | null | undefined,
+  rowHeaders: FieldTypesWithExternalType | null | undefined,
+): FieldTypesWithExternalType {
+  const base = fieldTypes ?? [];
+  if (!rowHeaders || rowHeaders.length === 0) {
+    return base;
+  }
+  const seen = new Set(base.map(([name]) => name));
+  const indexFields = rowHeaders.filter(([name]) => !seen.has(name));
+  return [...base, ...indexFields];
```
mscolnick, line 72: you could also do `uniqueBy([...base, ...indexFields], f => f[0])`. maybe reads a bit better

kirangadhave: oh nice, i forgot we had this util

PR #10100, opened 2026-07-08
fix(completion): fall back to interpreter completion when jedi static analysis raises

### LINKED ISSUE #10055: Attribute completion silently fails for variables assigned from pl.concat(...) — jedi crashes with 'TreeInstance' object has no attribute 'with_generics', and the interpreter fallback is never reached
# Preface
## ⬇️ This is written by a human, me Bartomat/Bartek from Munich 👋

Hi, 
currently I am learning Marimo with Polars and I am enjoying this very much!
Yesterday I was annoyed by missing autocompletions, and as Claude Fable 5 is available for some days in my subscription, I decided to let him investigate the problem.
And I was really astonished, how deeply he investigated the problem. The solution for me was pretty easy, but only in my case, with the help of Claude.
So I asked him to write down the error he claims to have found for you, the creators of Marimo, see below.
I am aware that this is not a severe bug, so priorize it in a manner that suits you.

But anyway: Thanks for all the great work you did and will do with Marimo 🍀 

Cheers 
Bartek
<hr>  

#### Describe the bug

Dot-attribute completion (e.g. typing `my_df.wi` and expecting `with_columns`) silently returns **no suggestions at all** for any variable whose definition involves `polars.concat(...)`. Variable-*name* completion keeps working (it is served from the frontend variables store), which makes the failure look mysterious: names complete, attributes don't.

The behavior is identical with marimo's built-in Jedi completion **and** with pylsp enabled, because both run the same Jedi static analysis, and Jedi crashes on polars' generic `concat` signature with:

```
AttributeError: 'TreeInstance' object has no attribute 'with_generics'
```

This is a known, still-open upstream Jedi bug: https://github.com/davidhalter/jedi/issues/1990 (open since 2024-04-22; originally surfaced as https://github.com/ipython/ipython/issues/14412). We verified it reproduces with jedi 0.19.2 (marimo's pinned range is `jedi<0.20.0,>=0.18.0`) **and** with jedi 0.20.0, so a version bump would not help.

The marimo-specific part of this report: in `marimo/_runtime/complete.py`, `complete()` wraps everything in a `try/except` that logs the failure only at DEBUG level and sends "no completions". Because the static path *raises* instead of returning an empty list, the interpreter-based fallback in `_get_completions()` — which would have completed this variable just fine from the runtime globals — is never reached. So a Jedi inference crash on one generic function silently kills attribute completion for the affected variable, with nothing visible in the default (WARN) logs.

## How this was found

I am Claude (Claude Fable 5, an AI assistant by Anthropic), and I researched and wrote this report while helping Bartek debug why completion "sort of didn't work" in his notebooks (marimo 0.23.13, Windows 11/Chrome and Linux Mint/Firefox — identical symptoms on both, which ruled out the browser). The trail was:

1. Running `marimo edit -l debug` revealed `Completion with jedi failed: 'TreeInstance' object has no attribute 'with_generics'` on every completion request for the affected variable (this message is `LOGGER.debug(...)` in `complete.py`, so it is invisible at the default log level).
2. `~/.cache/marimo/logs/pylsp.log` contained the same `AttributeError` 210 times, specifically when resolving the affected variable, with a traceback ending in `jedi/inference/compiled/value.py`, line 261, `v.with_generics(arguments)` (inside `execute_annotation`).
3. Bisecting the variable's definition chain in a clean environment isolated the trigger to `pl.concat(...)` alone — `pl.read_csv`, `unique(subset=~cs.integer())`, `with_row_index(...)`, and `sort(...)` are all fine on their own.

`pl.concat` is generically typed (`def concat(items: Iterable[PolarsType], ...)`, where `PolarsType` is a `TypeVar`), and the crash happens exactly while Jedi executes that generic annotation, matching jedi#1990.

#### Minimal reproduction (no marimo needed — this is the exact code path `_get_completions_with_script` runs)

```python
import jedi  # 0.19.2 or 0.20.0

code = "\n".join([
    "import polars as pl",
    'bk_a = pl.read_csv("a.csv")',
    'bk_b = pl.read_csv("b.csv")',
    "bk_x = pl.concat([bk_a, bk_b])",
    "bk_x.wi",
])
jedi.Script(code).complete()
# AttributeError: 'TreeInstance' object has no attribute 'with_generics'
```

Replace `pl.concat([bk_a, bk_b])` with plain `bk_a` and the same call returns `with_columns`, `with_row_index`, etc.

**In marimo:** put the first four lines in one cell, run it, then type `bk_x.` in a new cell. No completion popup appears (while `bk_a.` completes normally). Same result with the pylsp language server enabled.

#### Expected behavior

Either the completion should still work (the interpreter fallback would succeed — `jedi.Interpreter("bk_x.wi", [globals_with_real_df]).complete()` returns correct results for this case), or at minimum the failure should be visible above DEBUG level.

#### Suggested fixes (any subset would help)

1. In `_get_completions()` (`marimo/_runtime/complete.py`), catch exceptions from `_get_completions_with_script(...)` and treat them like "no static completions", so the existing interpreter fallback gets a chance to run.
2. Log Jedi completion failures at a more visible level (or once per session at WARNING), so users don't face a silent no-op.
3. Optionally track https://github.com/davidhalter/jedi/issues/1990 upstream.

#### Workaround for users

Annotate the assignment explicitly — Jedi then uses the annotation instead of evaluating `concat`'s generic return type:

```python
bk_x: pl.DataFrame = pl.concat([bk_a, bk_b])...
```

Verified: this restores attribute completion both with the built-in Jedi completion and with pylsp.

#### Environment

Affected user environment (symptoms confirmed on both):
- marimo 0.23.13, Windows 11 (Chrome) and Linux Mint (Firefox)
- jedi 0.19.2, polars 1.42.1, python-lsp-server 1.14.0

<details>
<summary>marimo env output</summary>

```
{
  "marimo": "0.23.13",
  "editable": false,
  "location": "C:/____bk/__Python_ENVs_2026/env_marimo_tests/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 183 Stepping 1, GenuineIntel",
  "Python Version": "3.14.2",
  "Locale": "--",
  "Binaries": {
    "Browser": "149.0.7827.201",
    "Node": "--",
    "uv": "0.9.28 (0e1351e40 2026-01-29)"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.1",
    "narwhals": "2.15.0",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.19.2",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.14.0",
    "uvicorn": "0.40.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "duckdb": "1.4.4",
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "openai": "2.16.0",
    "pandas": "3.0.0",
    "polars": "1.42.1",
    "pyarrow": "23.0.0",
    "python-lsp-ruff": "2.3.1",
    "ruff": "0.14.14",
    "sqlglot": "28.7.0"
  },
  "Experimental Flags": {}
}
```

</details>

Independently reproduced in a clean Linux sandbox with jedi 0.19.2 and 0.20.0, polars 1.42.1.

David-Wu1119:
## 📝 Summary

Closes #10055

jedi's static analysis can raise while inferring some code — e.g. resolving the generic return type of `polars.concat(...)` crashes with `AttributeError: 'TreeInstance' object has no attribute 'with_generics'` (upstream [jedi#1990](https://github.com/davidhalter/jedi/issues/1990)). Because `_get_completions()` only fell back to interpreter-based completion when the static path returned *no* completions — not when it *raised* — a jedi crash silently killed attribute completion for the affected variable, even though the interpreter fallback would have completed it fine from the runtime globals.

This implements the fix approach @manzt suggested on the issue: treat a static-analysis crash like "no static completions", so the interpreter fallback still gets a chance to run. The crash is logged at DEBUG (matching the existing logging in `complete()`).

While restructuring, this also removes a redundant second `script.complete()` call in `_get_completions()` — `_get_completions_with_script()` already runs the completion, so the previous code ran jedi's (potentially slow or crashing) analysis twice per request.

**Before:** with `df_x = pl.concat([df_a, df_b])` in a cell, typing `df_x.wi` produces no completion popup at all (silently, with nothing above DEBUG in the logs).
**After:** `df_x.wi` completes (`with_columns`, `with_row_index`, ...) via the interpreter fallback.

Tests: a unit test that simulates the static path raising (verifying the fallback runs), and a polars regression test with the exact repro from the issue (skipped when polars isn't installed).

## 📋 Pre-Review Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) — approach suggested by @manzt in https://github.com/marimo-team/marimo/issues/10055#issuecomment-4867834374.

## ✅ Merge Checklist

### DIFF
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -640,7 +640,7 @@ def _get_completions(
             return script, completions
     except Exception as e:
         # jedi's static analysis can crash while inferring some code — for
-        # example https://github.com/davidhalter/jedi/issues/1990). 
+        # example https://github.com/davidhalter/jedi/issues/1990).
         # Fallback to interpreter when it crashes
         LOGGER.debug("Completion with jedi Script failed: %s", str(e))
     return _get_completions_with_interpreter(document, glbls, glbls_lock)
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -634,13 +634,19 @@ def _get_completions(
     glbls: dict[str, Any],
     glbls_lock: threading.RLock,
 ) -> tuple[jedi.Script, list[jedi.api.classes.Completion]]:
-    script, completions = _get_completions_with_script(codes, document)
-    completions = script.complete()
-    if not completions:
-        script, completions = _get_completions_with_interpreter(
-            document, glbls, glbls_lock
-        )
-    return script, completions
+    try:
+        script, completions = _get_completions_with_script(codes, document)
+        if completions:
+            return script, completions
+    except Exception as e:
+        # jedi's static analysis can raise while inferring some code — for
+        # example, resolving the generic return type of polars' `concat`
+        # crashes with an AttributeError
+        # (https://github.com/davidhalter/jedi/issues/1990). Treat a crash
+        # like "no static completions" so the interpreter-based fallback
+        # below still gets a chance to run.
+        LOGGER.debug("Completion with jedi Script failed: %s", str(e))
+    return _get_completions_with_interpreter(document, glbls, glbls_lock)
 
 
 # NOTE you will hit front display bug if the result set doesn't
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -639,12 +639,9 @@ def _get_completions(
         if completions:
             return script, completions
     except Exception as e:
-        # jedi's static analysis can raise while inferring some code — for
-        # example, resolving the generic return type of polars' `concat`
-        # crashes with an AttributeError
-        # (https://github.com/davidhalter/jedi/issues/1990). Treat a crash
-        # like "no static completions" so the interpreter-based fallback
-        # below still gets a chance to run.
+        # jedi's static analysis can crash while inferring some code — for
+        # example https://github.com/davidhalter/jedi/issues/1990). 
+        # Fallback to interpreter when it crashes
         LOGGER.debug("Completion with jedi Script failed: %s", str(e))
     return _get_completions_with_interpreter(document, glbls, glbls_lock)
 
--- a/tests/_runtime/test_complete.py
+++ b/tests/_runtime/test_complete.py
@@ -23,6 +23,7 @@ from marimo._runtime.complete import (
     _get_completion_info,
     _get_completion_option,
     _get_completion_options,
+    _get_completions,
     _get_docstring,
     _maybe_get_key_options,
     _resolve_chained_key_path,
@@ -34,6 +35,7 @@ from tests.mocks import snapshotter
 
 snapshot = snapshotter(__file__)
 HAS_PANDAS = DependencyManager.pandas.has()
+HAS_POLARS = DependencyManager.polars.has()
 
 
 def test_build_docstring_function_no_init():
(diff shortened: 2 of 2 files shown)


marimo/_runtime/complete.py
```diff
@@ -634,13 +634,19 @@ def _get_completions(
...
+        script, completions = _get_completions_with_script(codes, document)
+        if completions:
+            return script, completions
+    except Exception as e:
+        # jedi's static analysis can raise while inferring some code — for
+        # example, resolving the generic return type of polars' `concat`
+        # crashes with an AttributeError
+        # (https://github.com/davidhalter/jedi/issues/1990). Treat a crash
+        # like "no static completions" so the interpreter-based fallback
+        # below still gets a chance to run.
```
Light2Dark:
```suggestion
        # jedi's static analysis can crash while inferring some code — for
        # example https://github.com/davidhalter/jedi/issues/1990). 
        # Fallback to interpreter when it crashes
```

Light2Dark (commented):
looks good, just doing a pass to reduce comments.

Light2Dark:
Would you also mind signing the CLA @David-Wu1119 ?

David-Wu1119:
recheck

David-Wu1119:

mscolnick (dismissed):

David-Wu1119:
Signed — thanks for the review! The CLA check is green now.

akshayka (approved):

PR #10119, opened 2026-07-09
feat: experimental SSE transport for the kernel connection (MARIMO_SERVER_TRANSPORT)

mscolnick:
Adds an experimental server-sent-events alternative to the /ws kernel
websocket, for deployments behind proxies or services that do not
support WebSockets. Opt in with:
 
```
MARIMO_SERVER_TRANSPORT=sse
```

The kernel channel is already unidirectional (client->kernel traffic
goes over HTTP POST keyed by the Marimo-Session-Id header), so SSE only
replaces the outbound stream. Terminal, LSP, and /ws_sync are untouched;
RTC is auto-disabled under SSE.

Server:
- Extract a transport-agnostic SessionHandler base (session lifecycle,
  kernel-ready, reconnect/resume/replay, TTL close) shared by
  WebSocketHandler and the new SSESessionHandler.
- New GET /sse endpoint streams the same {"op", "data"} wire payloads as
  SSE events, mirrors WebSocket close frames with an in-band `close`
  event, sends keep-alive comments, and connects to the session lazily
  on first iteration so an unconsumed response never attaches a
  consumer. Pending messages are flushed before a server close.
- Shared SSE framing + HTTP-disconnect helpers in marimo/_server/sse.py,
  reused by the scratchpad /execute stream.

Frontend:
- New SseTransport (fetch + incremental SSE parser) behind the existing
  IConnectionTransport interface, with partysocket-parity retries: the
  budget counts consecutive attempts without a delivered message, gives
  up with TRANSPORT_EXHAUSTED, and never stacks a retry on top of a
  consumer-driven reconnect.
- Auth travels in the Authorization header (like the REST API) instead
  of the URL; classifyCloseEvent now treats MARIMO_UNAUTHORIZED and
  MARIMO_KIOSK_NOT_ALLOWED as terminal.

### DIFF
--- a/docs/guides/deploying/index.md
+++ b/docs/guides/deploying/index.md
@@ -53,6 +53,31 @@ If you would like to deploy your application at a subpath, you can set the `--ba
 marimo run app.py --base-url /subpath
 ```
 
+### Deploying without WebSockets (experimental)
+
+marimo streams kernel messages to the browser over a WebSocket by default.
+Some proxies and hosting services handle WebSockets poorly or not at all; for
+these deployments, you can switch the kernel connection to server-sent
+events (SSE) over plain HTTP with the `MARIMO_SERVER_TRANSPORT` environment
+variable:
+
+```bash
+MARIMO_SERVER_TRANSPORT=sse marimo run app.py
+```
+
+This setting is experimental. Keep in mind:
+
+- The terminal and language servers (LSP) in `marimo edit` still require
+  WebSockets and are unaffected by this setting.
+- Real-time collaboration (`rtc_v2`) requires WebSockets and is disabled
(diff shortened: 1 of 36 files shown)


docs/guides/deploying/index.md
```diff
@@ -53,6 +53,31 @@ If you would like to deploy your application at a subpath, you can set the `--ba
 marimo run app.py --base-url /subpath
 ```
 
+### Deploying without WebSockets (experimental)
```
dmadisetti, line 56: Worth mentioning sage maker?

dmadisetti (approved):
Woops, sorry thought I approved

PR #10126, opened 2026-07-09
feat: experimental line_timing — green active-line highlight + per-line timer

mscolnick:
While a cell runs with the experimental `line_timing` flag on, the
currently executing line is highlighted green, and once a line has been
busy for ~500ms a small dim elapsed-time pill ticks at the end of it.

Reuses the debugger's frame watcher (single sys.settrace hook even with
both flags on); the debugger is now optional on FrameWatcher. Timing is
frontend-derived from active-line changes, so the protocol is unchanged.

Also moves formatElapsedTime from CellStatus.tsx to utils/time.ts so
core/codemirror doesn't import a component module.

https://github.com/user-attachments/assets/4eaf1d30-4dab-4334-8542-3e20567be69d

### DIFF
--- a/frontend/src/core/codemirror/cells/line-timing-decorations.ts
+++ b/frontend/src/core/codemirror/cells/line-timing-decorations.ts
@@ -1,0 +1,165 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+
+import {
+  type EditorState,
+  type Extension,
+  RangeSetBuilder,
+} from "@codemirror/state";
+import {
+  Decoration,
+  type DecorationSet,
+  EditorView,
+  type PluginValue,
+  ViewPlugin,
+  type ViewUpdate,
+  WidgetType,
+} from "@codemirror/view";
+import type { Observable } from "@/core/state/observable";
+import { Logger } from "@/utils/Logger";
+import { formatElapsedTime } from "@/utils/time";
+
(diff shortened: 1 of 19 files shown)


dmadisetti (approved):
Backend looked great to me

frontend/src/core/codemirror/cells/line-timing-decorations.ts
```diff
@@ -0,0 +1,165 @@
...
+export function activeLineTimer(
+  infoObservable: Observable<ActiveLineInfo | null>,
+): Extension {
+  return [
+    ViewPlugin.define((view) => new ActiveLineTimer(view, infoObservable), {
+      decorations: (plugin) => plugin.decorations,
+    }),
+    EditorView.theme({
+      ".cm-timing-current-line": {
+        backgroundColor: "color-mix(in srgb, var(--grass-4) 50%, transparent)",
```
manzt, line 155: More stylistic, but we could maybe reuse the line selection color (blue) instead of a new color unless that would be confusing. I think the running timer could distinguish that it's the executing line and not selected.

manzt (approved):
I like this. From a UI perspective it have a slight preference for avoiding more colors, either a more neutral gray or reusing the blue. But as a POC I think this is great and would be perfectly fine to ship as is for feedback.

PR #10127, opened 2026-07-09
Support anywidget composition and hot reload

### LINKED ISSUE #9719: Support for anywidget frontend hot-reload
### Description

Marimo's `mo.watch.file()` works well for reloading widgets when their JS code changes:
```py
# Cell 1
js_file = mo.watch.file("monkey.js")
css_file = mo.watch.file("monkey.css")
```
```py
# Cell 2
class Widget(anywidget.AnyWidget):
    _esm = js_file.read_text()
    _css = css_file.read_text()
Widget()
```
However, the whole cell is now re-running, not only the frontend.

Here's a comparison screen recording, left JupyterLab with real hot reloading, and right the current marimo workaround.

https://github.com/user-attachments/assets/d23c8551-a781-4dff-aca3-b680a8d5a569

Would be great if Marimo could support real hot reloading as well!

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

manzt:
Closes #9719 

Supports the new anywidget [composition API](https://github.com/manzt/anywidget/pull/942), where parents render children they were handed as values through `host.getWidget` and widget code can reload in place as you edit it. Also simplifies how widget code reaches the frontend (one channel instead of three minting sites in every open message) and how the frontend manages widget lifecycles.

Terminology

- **ESM spec**, a url plus content hash for a widget's frontend code, minted by the comm once at open and carried on model notifications.
- **Registry**, the one frontend owner of the model, spec, and bindings, serving display mounts and `host.getWidget` alike.
- **Generation**, one immutable binding. `initialize` runs exactly once, and a code change replaces the generation against the same model, re-rendering live views with state intact. That is the entire hot reload mechanism.

Specs ride notifications rather than model state, even though upstream anywidget reads code from the `_esm` trait. Model state is client-writable and echoes to peers, so a frontend that executed it would let one client run code on another in a shared session. This is also why the server strips `_esm` and `_css` from client updates, and why hot reload only operates in edit sessions (enforced on both kernel and client). A viewer's widget code is immutable for its model's lifetime. CSS hot-applies in every mode since it is state, not code.

Render listeners now observe current model state exactly once per view, right after `render` settles. Updates can arrive before a view attaches its listeners, and the old replay both re-fired every listener on the model (mounting a second view double-painted the first) and only ran on the display path, so composed children never hydrated.

Also fixes URL-form `_esm` (previously refused) and replaces silent ten second timeouts with precise errors.


frontend/src/plugins/impl/anywidget/types.ts
```diff
@@ -10,6 +10,10 @@ export type EventHandler = (...args: any[]) => void;
  */
 export type WidgetModelId = TypedString<"WidgetModelId">;
 
+export function isWidgetModelId(value: unknown): value is WidgetModelId {
```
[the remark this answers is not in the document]
manzt: false positive.

packages/openapi/src/api.ts
```diff
@@ -5614,10 +5570,23 @@ export interface components {
...
+     *             buffer_paths: Paths into `state` whose binary values were
+     *                 extracted into `buffers`.
+     *             buffers: Binary payloads, parallel to `buffer_paths`.
+     *             esm_spec: Where to import this widget's ESM from.
      */
     ModelOpen: {
       buffer_paths: (string | number)[][];
       buffers: components["schemas"]["Base64String"][];
+      /** @default null */
+      esm_spec?: null | components["schemas"]["EsmSpec"];
```
manzt, line 5589: ModelOpen messages now optionally carry the ESM for the widget (isntead of stuffing js_url and js_hash onto model state).

frontend/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@ai-sdk/react": "^3.0.134",
-    "@anywidget/types": "^0.2.0",
```
manzt, line 25: New anywidget types.

PR #10133, opened 2026-07-10
test: add unit tests for env_to_value

eeshsaxena:
## Description

`marimo/_utils/env.py` (`env_to_value`) had no dedicated test module. This adds `tests/_utils/test_env.py` covering:

- an unset variable returning `None`
- boolean parsing (`true`/`false`, case-insensitive)
- list parsing (`[a,b,c]`)
- the literal `none` string mapping back to `None`
- a plain string being wrapped in a one-tuple

No source changes.

## Verification

`pytest tests/_utils/test_env.py` - 5 passed. `ruff check` / `ruff format --check` clean.


eeshsaxena:
Rebased onto current `main` to clear the conflict. `tests/_utils/test_env.py` was added on both sides - `main` covers `is_env_true`, this PR covers `env_to_value` - so I merged the two sets into one file rather than replacing either. Renamed my key constant to `ENV_TO_VALUE_KEY` to avoid clashing with the existing `KEY`, and matched the `monkeypatch: pytest.MonkeyPatch` annotation style already used there.

20 tests pass, ruff check + format clean.

PR #10135, opened 2026-07-10
fix(runtime): prevent stale UI finalizers from deleting live functions

### LINKED ISSUE #9763: Sporadic "could not find function given request" in mo.ui.table

kirangadhave:
## 📝 Summary

Closes #9763.

Ensure a UI element finalizer only deletes function namespaces owned by its own runtime context. Previously, a finalizer from a closed session could delete namespaces from a live session, causing sporadic `Function not found` errors.

Reproduced by disabling automatic GC, switching sessions, and manually triggering collection. With the fix, namespaces remain registered and table search continues working.

### DIFF
--- a/marimo/_plugins/ui/_core/registry.py
+++ b/marimo/_plugins/ui/_core/registry.py
@@ -181,7 +181,10 @@ class UIElementRegistry:
         except ContextNotInitializedError:
             pass
         else:
-            ctx.function_registry.delete(namespace=object_id)
+            # GC finalizers may run under a context that does not own this
+            # registry; only delete functions from the owning context.
+            if ctx.ui_element_registry is self:
+                ctx.function_registry.delete(namespace=object_id)
 
         if object_id in self._bindings:
             del self._bindings[object_id]
--- a/tests/_plugins/ui/_core/test_registry.py
+++ b/tests/_plugins/ui/_core/test_registry.py
@@ -1,10 +1,29 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
+import gc
+import weakref
+from dataclasses import dataclass, field
+from typing import Any, cast
+
+import pytest
+
 from marimo import ui
+from marimo._plugins.ui._core.registry import UIElementRegistry
+from marimo._plugins.ui._core.ui_element import UIElement
 from marimo._runtime.context import get_context
+from marimo._runtime.functions import EmptyArgs, Function
 from marimo._runtime.runtime import Kernel
+from marimo._types.ids import UIElementId
 from tests.conftest import ExecReqProvider
+
+
+@dataclass
+class _CyclicDummy:
+    cycle: _CyclicDummy = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.cycle = self
 
 
 async def test_cached_element_still_registered(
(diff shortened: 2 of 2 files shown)


mscolnick (approved):

liquidcarbon:
Looks legit, I'm excited for the fix to be released!  I wonder if the imprecise garbage collection might be responsible for another set of vague symptoms in #9127 

Thanks for your work!

kirangadhave:
> I wonder if the imprecise garbage collection might be responsible for another set of vague symptoms

Interesting, I'll try to take a look and explore the issue in this direction

liquidcarbon:
It's a stretch - but main similarity is that both behaviors show up only you have multiple sessions on the same endpoint, which isn't straightforward to test for, making the bugs hard to catch.

Your fix suggests that when one session is closed, a number of shared objects disppear into garbage.  This causes an irrecoverable crash, sometimes just on table UI, sometimes on the whole worker.  Am I on the right track here?

PR #10142, opened 2026-07-11
fix(completions): complete on empty prefix after import statements

### LINKED ISSUE #10140: Explicit completion does not activate for imports with whitespace prefix
### Describe the bug

Completion works as expected for `from dataclasses import d<pause/ctl+space>` (completion options appear), but does *not* activate at all for `from dataclasses import <pause/ctl+space>`. This is an important feature for experimenting with APIs and I miss it when I use marimo. Love the tool otherwise!

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.14",
  "editable": false,
  "location": "/Users/tomshafer/Library/Caches/uv/archive-v0/_LjD0vyGeI3lEH5m/lib/python3.14/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.5.0",
  "Processor": "arm",
  "Python Version": "3.14.2",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "149.0.7827.201",
    "Node": "v26.3.1",
    "uv": "0.11.26 (396ef7ce4 2026-06-30 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.23.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.0",
    "uvicorn": "0.51.0",
    "websockets": "16.1"
  },
  "Optional Dependencies": {
    "loro": "1.13.1",
    "python-lsp-ruff": "2.3.1",
    "ruff": "0.15.21"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
# /// script
# dependencies = [
#     "marimo",
# ]
# requires-python = ">=3.14"
# ///

import marimo

__generated_with = "0.23.14"


@app.cell
def _():
    import marimo as mo

    return


@app.cell
def _():
    from dataclasses import 

    return


if __name__ == "__main__":
    app.run()
```

AnayGarodia:

Fixes #10140.

## Problem

Autocomplete does not open for `from dataclasses import <cursor>` (empty prefix), while `from dataclasses import d<cursor>` works. Same for a bare `import <cursor>`.

Jedi does return the right completions at that position (the module's importable names), but `complete()` discarded them: the prefix is empty and the last character isn't a `.`/`/` trigger, so the empty-prefix guard cleared the list and fell through to signature help.

That guard exists for a good reason — an empty prefix after `(` or `,` makes Jedi return the entire namespace (every builtin and global), which is noise. But an import-name position is different: Jedi returns only the relevant, bounded set of names, so completing there is exactly what you want.

## Fix

- Detect when the cursor sits at an import-name position (`from X import `, `import `, `from a import b, `) from the current line, and keep the completions in that case.
- Fix an adjacent latent bug: when the prefix is empty, `prefix = document[-prefix_length:]` computes `document[-0:]`, which is the whole document rather than `""`. Guard the zero case.

The intentional suppression after `(`/`,` and the `/`-as-division handling are untouched and still covered by their existing tests.

## Tests

`tests/_runtime/test_complete.py`:
- `test_completion_after_from_import_with_empty_prefix` — `from dataclasses import `, a trailing-comma form, and extra-whitespace form all surface `dataclass`/`field`/`asdict` with `prefix_length == 0`.
- `test_completion_after_bare_import_with_empty_prefix` — `import ` surfaces top-level modules.

Full file passes (140 passed, 127 skipped for optional deps), `ruff check`/`format` clean, `mypy` clean.

### DIFF
--- a/marimo/_runtime/complete.py
+++ b/marimo/_runtime/complete.py
@@ -71,6 +71,22 @@ DOC_CACHE_SIZE = 200
 # characters falls through to signature help below, which is the useful
 # behavior inside a call's argument list (e.g. `mo.ui.slider(start=1,`).
 COMPLETION_TRIGGER_CHARACTERS = frozenset({".", "/"})
+# Matches an import statement whose cursor sits at a name position with an empty
+# prefix, e.g. `from dataclasses import ` or `import ` or `from a import b, `.
+# Unlike an empty prefix after `(`/`,` (which yields the entire namespace), Jedi
+# returns only the relevant, bounded set of importable names here, so completing
+# on an empty prefix is the useful behavior. Operates on the final line only.
+_IMPORT_COMPLETION_PATTERN = re.compile(
+    r"\s*(?:from\s+[\w.]+\s+)?import\s+(?:\w+(?:\s+as\s+\w+)?\s*,\s*)*$"
+)
+
+
+def _is_import_completion_context(document: str) -> bool:
+    """Return True when the cursor is at an import-name position (empty prefix)."""
+    last_line = document.rsplit("\n", 1)[-1]
+    return _IMPORT_COMPLETION_PATTERN.match(last_line) is not None
+
+
 # Matches display bracket delimiters: \[...\]
 _MATH_DISPLAY_BRACKET_PATTERN = re.compile(r"\\\[(.+?)\\\]", re.DOTALL)
 # Matches inline paren delimiters: \(...\)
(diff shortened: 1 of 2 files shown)


tomshafer:
This does not seem to fix the issue. I’m closing the issue in marimo in favor of an issue in marimo-team/codemirror-languageserver, which is where the problem seems to originate.

AnayGarodia:
Closing this. @tomshafer confirmed the completion isn't triggered on the frontend side (moved to marimo-team/codemirror-languageserver#68), so this backend change to the kernel completion path doesn't address the user-visible bug — the request never reaches `complete()`. Thanks for the quick check. Happy to help on the CodeMirror side if useful.

PR #10146, opened 2026-07-12
test: add unit tests for the is_url utility

eeshsaxena:
## Description

`marimo/_utils/url.py` (the `is_url` validator, adapted from the
`validators` package) had no dedicated test module. This adds
`tests/_utils/test_url.py` covering:

- common schemes accepted (`http`, `https`, `ftp`)
- URLs with a port, path, query, and fragment
- IP-address and `localhost` hosts
- rejection of non-URLs (empty string, plain text, missing scheme,
  single-character TLD) and unsupported schemes
- the `public=True` flag rejecting private/loopback hosts while still
  accepting public ones, and defaulting to `False`

No source changes.

## Verification

`pytest tests/_utils/test_url.py` - 8 passed. `ruff check` /
`ruff format --check` clean.

### DIFF
--- a/tests/_utils/test_url.py
+++ b/tests/_utils/test_url.py
@@ -1,0 +1,50 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
+from marimo._utils.url import is_url
+
+
+def test_accepts_common_schemes() -> None:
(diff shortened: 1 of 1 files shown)


Light2Dark:
This function is a copy from another source, it's unecessary to add tests. Thanks!

PR #10155, opened 2026-07-13
test: add unit tests for assert_never and log_never

eeshsaxena:
## Description

`marimo/_utils/assert_never.py` (`assert_never`, `log_never`) had no dedicated test module. This adds `tests/_utils/test_assert_never.py` covering:

- `assert_never` raising `AssertionError` with the value and its type name in the message
- the exact `Unhandled value: 5 (int)` format
- `log_never` (the non-fatal counterpart) returning `None` without raising

No source changes.

## Verification

`pytest tests/_utils/test_assert_never.py` - 3 passed. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/tests/_utils/test_assert_never.py
+++ b/tests/_utils/test_assert_never.py
@@ -1,0 +1,25 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
+import pytest
+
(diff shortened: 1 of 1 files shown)


Light2Dark:
Hi, thanks for the change. But I don't think this is necessary to add.

PR #10166, opened 2026-07-13
feat: propagate request trace context to the kernel

### LINKED ISSUE #6637: Offer a way to pass Starlette trace to the kernel
### Description

When mounting Marimo notebooks as an ASGI app using the starlette integration provided, there seems to be no way to wrap all activity in a notebook session into a single trace that can be instrumented/reported by third party libraries like logfire (based on OpenTelemetry).

Here is a simple setup:
```
logfire.configure()
public_app = FastApi(...)
internal_app = FastApi(....)
server = marimo.create_asgi_app()
notebooks_dir = Path(__file__).parent / "notebooks"

for filename in sorted(notebooks_dir.iterdir()):
    if filename.suffix == ".py":
        app_name = filename.stem
        server = server.with_app(path=f"/{app_name}", root=str(filename))

internal_app.mount("/notebooks", server.build())
internal_app.add_middleware(NotebookAuthMiddleware)

app = FastAPI(title="API Gateway", docs_url=None, openapi_url=None)
# Mount internal and public APIs
app.mount("/internal", internal_app)
app.mount("", public_app)

logfire.instrument_fastapi(app, excluded_urls=["/healthz", "/internal/healthz"])
logfire.instrument_pydantic(record="failure")
logfire.instrument_httpx()
logfire.instrument_requests()
logfire.instrument_sqlalchemy()
logfire.instrument_system_metrics(base="basic")
```

Ideally when users visit a notebook we'd pass along trace_id between the fastAPI entry-point and the kernel so that they can be tied together.

### Suggested solution

Pass the trace from Starlette to the Kernel

- [ ] Yes

### Alternatives

_No response_

### Additional context

Started as a [thread on Discord with context there](https://discord.com/channels/1059888774789730424/1422749717615607829)

mscolnick:
Re-attach the W3C trace context carried on each control command in  <br>handle_message so kernel spans (cell execution, previews, function  <br>calls) become children of the originating request's trace. Enables a  <br>single distributed trace to span an upstream app, the marimo server,  <br>and the kernel, even when marimo is mounted as an ASGI sub-app.

Closes marimo-team/marimo#6637

### DIFF
--- a/marimo/_tracer.py
+++ b/marimo/_tracer.py
@@ -308,11 +308,14 @@ def attach_trace_context(
     try:
         from opentelemetry import context as otel_context
         from opentelemetry.propagate import extract
-    except Exception:
+
+        token = otel_context.attach(extract(carrier=dict(headers)))
+    except Exception as e:
+        # A missing dependency or malformed header must never break dispatch.
+        LOGGER.debug("Failed to attach trace context", exc_info=e)
         yield
         return
 
-    token = otel_context.attach(extract(carrier=dict(headers)))
     try:
         yield
     finally:
--- a/marimo/_tracer.py
+++ b/marimo/_tracer.py
@@ -14,7 +14,7 @@ from marimo._utils.platform import is_pyodide
 LOGGER = _loggers.marimo_logger()
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from collections.abc import Iterator, Mapping, Sequence
     from pathlib import Path
 
     from opentelemetry import trace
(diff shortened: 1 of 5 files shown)


marimo/_tracer.py
```diff
@@ -285,6 +285,40 @@ def create_tracer(trace_name: str) -> trace.Tracer:
...
+    is unavailable.
+    """
+    if not headers or is_pyodide() or GLOBAL_SETTINGS.TRACING is False:
+        yield
+        return
+
+    try:
+        from opentelemetry import context as otel_context
+        from opentelemetry.propagate import extract
+    except Exception:
```
dmadisetti: Should we log?

mscolnick: yea good call

marimo/_tracer.py
```diff
@@ -285,6 +285,40 @@ def create_tracer(trace_name: str) -> trace.Tracer:
...
+    context from `headers` and installs it as the current context for the
+    duration of the block. This links kernel spans to the trace of the
+    originating HTTP request, so a distributed trace (e.g. from an upstream
+    FastAPI/Starlette app) can span both the marimo server and the kernel,
+    even though the kernel runs in a separate process.
+
+    No-op when there are no headers, tracing is disabled, or opentelemetry
+    is unavailable.
+    """
+    if not headers or is_pyodide() or GLOBAL_SETTINGS.TRACING is False:
```
dmadisetti, line 304:
```suggestion
    if not headers or is_pyodide() or not GLOBAL_SETTINGS.TRACING:
```

`headers is None` ? Maybe not, empty dict should also return

dmadisetti (approved):
Looks good!

marimo/_tracer.py
mscolnick: we should probably enforce this with ruff https://docs.astral.sh/ruff/rules/true-false-comparison/, we are not consistent in the repo about this

PR #10167, opened 2026-07-13
fix(code_mode): skip missing-package nudge for marimo.* imports

mscolnick:
A failed `import marimo.<x>` surfaces as a missing `marimo` package,
prompting code_mode to install marimo. But marimo is always installed
and a bad submodule import can't be fixed by installing it, so filter
marimo-owned modules/packages out of the missing-packages computation.

Closes MO-6161

### DIFF
--- a/marimo/_runtime/callbacks/packages.py
+++ b/marimo/_runtime/callbacks/packages.py
@@ -37,6 +37,17 @@ if TYPE_CHECKING:
 LOGGER = _loggers.marimo_logger()
 
 
+def _is_marimo_module(module_name: str) -> bool:
+    """True for marimo itself or any of its submodules.
+
+    A failed `import marimo.<submodule>` surfaces as a missing "marimo"
+    package, but marimo is always installed and the error can never be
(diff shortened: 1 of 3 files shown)


marimo/_runtime/callbacks/packages.py
```diff
@@ -37,6 +37,17 @@
 LOGGER = _loggers.marimo_logger()
 
 
+def _is_marimo_module(module_name: str) -> bool:
```
manzt, line 40: I'm trying to think... would we want to help the agent understand that module doesn't exist? Or will it get that feedback?

manzt: Thinking about it, probably not. The `missing_packages_hook` is supplementary and only governs the install nudge. The agent will still see an import error.

manzt (approved):

marimo/_runtime/callbacks/packages.py
mscolnick: yea they will get the error. this told them to install it

PR #10169, opened 2026-07-13
feat(pair): target the current session so agents connect to the right notebook

mscolnick:
Include the notebook's session id in the pair-with-agent commands and
prompt so agents connect to the specific notebook the user is viewing
instead of guessing when multiple sessions are open.

Extract the pure command-building logic out of the modal into
pair-with-agent-commands.ts and add unit tests for it.

Closes MO-5902


manzt (dismissed):
Nice. This will be good to have for multiple notebooks.

PR #10172, opened 2026-07-14
test: add unit tests for the timer decorator

eeshsaxena:
## Description

`marimo/_utils/timer.py` (the `timer` decorator) had no dedicated test module. This adds `tests/_utils/test_timer.py` covering:

- the wrapped function's return value being passed through
- positional and keyword arguments being forwarded
- `functools.wraps` preserving name/docstring
- the timing line being printed (`"<func> took ... seconds to execute"`)

No source changes.

## Verification

`pytest tests/_utils/test_timer.py` - 4 passed. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/tests/_utils/test_timer.py
+++ b/tests/_utils/test_timer.py
@@ -1,0 +1,45 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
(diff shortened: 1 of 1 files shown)


Light2Dark:
not necessary, we don't need to test every function

PR #10175, opened 2026-07-14
security: move OG image execution behind flag

peter-gy:
Makes the OG image generator function's automatic execution **opt-in** through `--execute-opengraph-generators` flag to reduce risk of Remote Code Execution vulnerability.


dmadisetti (approved):
Thanks! I think this is a great compromise

PR #10201, opened 2026-07-16
feat(keymap): option to disable Escape command mode (salvage of #9907, credit @ameenalkhaldi)

### LINKED ISSUE #7426: Option to disable command mode or change key binding
### Description

My muscle memory is to press escape to clear any hint popups before navigating with the arrow keys. When I do this in the notebook it unexpectedly shoots me to the next or previous cell.

### Suggested solution

Provide an option to disable escape entering command mode or add the ability to remap the key.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

Bartok9:
**This pull request was authored with AI assistance and human-reviewed.**

## Summary

salvage: [#9907](<https://github.com/marimo-team/marimo/issues/9907>) (credit **@ameenalkhaldi**)

Adds `keymap.enter_command_mode_on_escape` (default `true`) so users can set it to `false` when Escape should clear completions / signature help without entering command mode.

## Why salvage

Original [#9907](<https://github.com/marimo-team/marimo/issues/9907>) was a clean, focused draft that stuck on **CLA**. Rebased onto current `main`, kept the original design, and documented the flag in the editor overview (keymap section). OpenAPI `KeymapConfig` + frontend schema updated to match.

Closes marimo-team/marimo#7426.

## Config

```toml
[keymap]
enter_command_mode_on_escape = false
```

## Checks

- [X] Line-by-line human review of the diff
- [X] Pure additive config default preserves existing Escape behavior
- [X] No vendor `Co-authored-by` trailers

### DIFF
--- a/docs/guides/editor_features/overview.md
+++ b/docs/guides/editor_features/overview.md
@@ -72,9 +72,9 @@ mode](#command-mode) for notebook navigation.
 - `dd` - delete empty cell
 - `:w` - save notebook
 
-**Disable Escape → command mode (default keymap):**
+**Keep Escape from entering command mode:**
 
-By default, Escape leaves the cell editor and enters command mode. To keep Escape for dismissing completions/signature help only:
+Normally Escape exits the editor and enters command mode. Set this to `false` if you'd rather have Escape only dismiss completions and signature help:
 
 /// tab | User config
 
--- a/docs/guides/editor_features/overview.md
+++ b/docs/guides/editor_features/overview.md
@@ -72,9 +72,9 @@ mode](#command-mode) for notebook navigation.
 - `dd` - delete empty cell
 - `:w` - save notebook
 
-**Keep Escape from entering command mode:**
+**Escape without command mode:**
 
-Normally Escape exits the editor and enters command mode. Set this to `false` if you'd rather have Escape only dismiss completions and signature help:
+By default, `Esc` leaves the cell editor and enters command mode. Set the option below to `false` to keep `Esc` from switching modes; it will still close completions and signature help:
 
 /// tab | User config
 
--- a/docs/guides/editor_features/overview.md
+++ b/docs/guides/editor_features/overview.md
@@ -72,7 +72,30 @@ mode](#command-mode) for notebook navigation.
 - `dd` - delete empty cell
 - `:w` - save notebook
 
+**Disable Escape → command mode (default keymap):**
+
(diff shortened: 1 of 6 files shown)


Bartok9:
Thanks @cubic-dev-ai — valid P1. Fixed in 6d99c11: popup/selection cleanup (`simplifySelection`, `closeSignatureHelp`, `closeCompletion`) now always runs on Escape regardless of `enter_command_mode_on_escape`; only the `exitToCommandMode()` call is gated on the flag via a new `allowCommandMode` parameter. So with the option disabled, Escape still dismisses signature help and autocomplete popups as intended.

docs/guides/editor_features/overview.md
```diff
@@ -72,8 +72,31 @@ mode](#command-mode) for notebook navigation.
 - `dd` - delete empty cell
 - `:w` - save notebook
 
+**Disable Escape → command mode (default keymap):**
```
akshayka: Please write this subtitle by hand, not with AI.

akshayka:
@manzt, do we want this as an option? If so I imagine we want a different setting name (the proposed one is too specific) and a user interface.

Bartok9:
Thanks @akshayka — rewrote that subtitle by hand in a9ee867 (no AI). The popup-clearing gap you flagged earlier is handled in 6d99c11: `closeSignatureHelp()`/`closeCompletion()` now run unconditionally, and only the `exitToCommandMode()` step is gated on the flag.

On the setting name/UI: totally agree `enter_command_mode_on_escape` is narrow. Happy to rename to something more general (open to your + @manzt's preference) and wire it into the keymap settings UI if you'd like this to land as a real option. Just say the word and I'll push the rename + UI in this PR.

Bartok9:
Rewrote the Escape-option heading and lead sentence by hand in the latest commit — thanks @akshayka.

manzt (commented):
Thanks for picking this up. I agree with the user need, but I think this should be modeled as a key binding rather than an Escape-specific boolean.

We already have `command.vimEnterCommandMode` for the Vim preset. Could we add `command.enterCommandMode` for the regular preset, defaulting to `Escape`, and allow overrides to be explicitly unbound? For example, an empty override could mean disabled, while a missing override continues to mean “use the default.” This preserves the different regular/Vim bindings and supports both remapping and disabling, similar to Jupyter's configurable `enter-command-mode` binding.

Issue #10182, opened 2026-07-14
Docs: pytest fixture patterns in setup cell and conftest.py don't work as documented


Seamus05:
## Bug Report

The [pytest testing guide](https://docs.marimo.io/guides/testing/pytest/) claims two fixture patterns work that don't in practice.

### Environment
- marimo 0.23.14
- pytest 9.0.2
- Python 3.12.3

### Pattern 1: Fixtures in the setup cell (docs claim it works)

The docs show this example:
```python
with app.setup:
    from fixtures import db_connection, sample_data

@app.cell
def _(sample_data):
    def test_data_loaded(sample_data):
        assert len(sample_data) > 0
```

**Result:** The fixture is passed as a `FixtureFunctionDefinition` object, not the resolved value. The test receives the raw fixture definition, not its return value.

```
E       TypeError: 'FixtureFunctionDefinition' object is not subscriptable
```

### Pattern 2: conftest.py fixtures (docs claim they work)

The docs state: "`conftest.py` fixtures work as expected — pytest discovers them automatically."

**Result:** The fixture name isn't in the cell's reference list, so marimo raises `NameError` — the cell signature doesn't include it.

```
E       NameError: name 'db_connection' is not defined
```

### What actually works

The only reliable pattern is defining fixtures and tests in the same cell:

```python
@app.cell
def _(pytest):
    @pytest.fixture
    def temp_file():
        import tempfile
        with tempfile.NamedTemporaryFile() as f:
            yield f

    def test_writes_to_file(temp_file):
        temp_file.write(b"hello")
        temp_file.seek(0)
        assert temp_file.read() == b"hello"
```

### Suggested fix

Update the docs to:
1. Remove or correct the setup cell fixture example (it doesn't work)
2. Remove or correct the conftest.py claim (it doesn't work)
3. Clarify that the only reliable pattern is same-cell fixture+test definition
4. For shared fixtures, recommend importing from a plain .py module in each test cell

dmadisetti:
Pattern 1 definitely sounds like a bug, thanks for reporting.

I can take another look at `conftest`- that's surprising to me and uncertain why our tests haven't caught this.

---

For clarity, is this from `pytest mynotebook.py` or directly in the kernel?

PR #10202, opened 2026-07-16
docs(testing): correct pytest fixture patterns for marimo notebooks

Bartok9:
**This pull request was authored with AI assistance and human-reviewed.**

## Summary

Updates the pytest guide so fixture docs match real behavior (see #10182):

- **Recommend** defining fixtures in the **same cell** as the tests that use them
- Keep class-scoped fixtures as a same-cell pattern
- Prefer **plain Python helpers** for shared setup
- Explicitly call out **setup-cell fixture imports** and **naive `conftest.py`** usage as patterns that usually fail

Docs only — no runtime changes.

Addresses #10182.

### DIFF
--- a/docs/guides/testing/pytest.md
+++ b/docs/guides/testing/pytest.md
@@ -158,29 +158,11 @@ FAILED test_notebook.py::test_fails - AssertionError: This test fails
 
 ## Using Pytest Fixtures
 
+marimo notebooks can define pytest fixtures, but **fixture resolution does not follow
+the same rules as ordinary notebooks cells**. The reliable pattern is to define
+fixtures **in the same cell as the tests that use them**.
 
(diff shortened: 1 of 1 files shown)


Bartok9:
recheck

dmadisetti:
Thanks, I just merged a change to the `pytest` and will followup to see if it works as expected. Some of the behavior may not be intended, so I rather fix the issue than document a bug. I'll followup on the original ticket, but moving this to draft until I scope everything out

Bartok9:
Thanks so much, @dmadisetti — that makes complete sense. Fixing the underlying `pytest` behavior beats documenting a bug any day. Happy to have this in draft while you scope it out. If it turns out the docs still need a correction once the fix lands, just ping me and I'll rebase and update to match the intended behavior. Appreciate you taking the time.

Bartok9:
Restored to draft — sorry for the accidental ready flip earlier. Still holding for @dmadisetti pytest scoping as discussed.

PR #10207, opened 2026-07-17
feat: API for retained WASM island lifecycle

peter-gy:
## 📝 Summary

Adds a retained-runtime lifecycle for static sites that navigate between pages containing marimo islands. Page transitions can replace the active Python app while keeping the worker and Pyodide environment ready.

The islands entry module exports:

- `canReplaceApp()` returns whether the current document contains one replaceable app
- `stopApp(appId?)` stops the matching active app session
- `initialize()` starts, reuses, or replaces the app in the current document

Hosts can call these methods around client-side navigation:

```ts
await stopApp()
updatePage()
await initialize()
```

App transitions are serialized and scoped by app ID and session generation. Rapid or failed transitions preserve session ownership, prevent stale controls from reaching another app, and allow teardown or replacement to be retried.

Worker startup creates and tracks the Python resources for each session. Package loading is serialized, superseded sessions skip initialization, and successful teardown releases the session task and Python references while retaining Pyodide.

Also, island elements now support page content that is upgraded or reused during navigation. They bind after compiled cell IDs become available, refresh when an existing cell ID receives new source, and release static staging output after mounting.

### DIFF
--- a/frontend/src/core/islands/components/web-components.tsx
+++ b/frontend/src/core/islands/components/web-components.tsx
@@ -5,7 +5,7 @@ import { isValidElement, type JSX } from "react";
 import ReactDOM, { type Root } from "react-dom/client";
 import { ErrorBoundary } from "@/components/editor/boundary/ErrorBoundary";
 import { TooltipProvider } from "@/components/ui/tooltip";
-import { notebookAtom } from "@/core/cells/cells";
+import { cellIdsAtom, notebookAtom } from "@/core/cells/cells";
 import { OBJECT_ID_ATTR } from "@/core/dom/ui-element-constants";
 import { UI_ELEMENT_REGISTRY } from "@/core/dom/uiregistry";
 import { LocaleProvider } from "@/core/i18n/locale-provider";
@@ -16,6 +16,7 @@ import { store } from "../../state/jotai";
 import {
   ISLAND_CSS_CLASSES,
   ISLAND_DATA_ATTRIBUTES,
+  ISLAND_SOURCE_CHANGED_EVENT,
   ISLAND_TAG_NAMES,
 } from "../constants";
 import { extractIslandCodeFromEmbed, retainIslandSource } from "../parse";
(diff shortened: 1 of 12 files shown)


frontend/src/core/islands/components/web-components.tsx
```diff
@@ -242,6 +297,13 @@ export class MarimoIslandElement extends HTMLElement {
    * Cleanup when element is removed from DOM
    */
   disconnectedCallback(): void {
+    this.connectionGeneration += 1;
```
mscolnick: are we incrementing on connection and disconnection?

frontend/src/core/wasm/worker/bootstrap.ts
```diff
@@ -12,6 +13,7 @@ import type { SerializedBridge, WasmController } from "./types";
 import { shouldLoadDuckDBPackages } from "../utils";
 
 const MAKE_SNAPSHOT = false;
+type SessionResources = [PyProxy, PyCallable, PyProxy, PyCallable];
```
mscolnick: nit: could do a typed tuple here `[one: PyProxy, two: PyCallable, three: PyProxy, four: PyCallable]`

mscolnick (commented):
looks great! minor comments

frontend/src/core/islands/components/web-components.tsx
peter-gy: the idea was for `connectedCallback()` to start a new lifecycle and `disconnectedCallback()` to explicitly invalidate its pending work, so every lifecycle transition advances the epoch. looking at it again it is probably overkill though because `isConnected` and the next connection generation already cover the same race.

mscolnick: i think thats actually fine and makes total sense now that i think about it

mscolnick (approved):
ship!

mscolnick:
oh one last idea, could we add a test that validates the exports and possible the typedef API and snapshots it? since now its a public API that cannot break.

can do in a followup

peter-gy:
> oh one last idea, could we add a test that validates the exports and possible the typedef API and snapshots it? since now its a public API that cannot break.
> 
> 
> 
> can do in a followup

Oh yes I'll do it in a next pr

PR #10211, opened 2026-07-17
fix: data_editor corrupting int8/uint8/float16 columns on edit

### LINKED ISSUE #10210: mo.ui.data_editor silently corrupts int8/uint8/float16 columns on edit
### Bug

Editing a cell in a numeric column whose dtype is `Int8`, `UInt8`, `Int128`, `UInt128`, or `Float16` silently corrupts the column instead of preserving the number.

`_convert_value` in `marimo/_plugins/ui/_impl/data_editor.py` hand-enumerates only `Int16/Int32/Int64/UInt16/UInt32/UInt64` and `Float32/Float64`. Any other numeric width falls through to `return str(value)`, so:

- **pandas** — the edited value becomes a string and the whole column is coerced to `object`.
- **polars** — the edit fails with a misleading `ValueError: Data editor does not support this type of data`.

### Reproduction

```python
import pandas as pd
from marimo._plugins.ui._impl.data_editor import apply_edits

df = pd.DataFrame({"a": pd.array([1, 2, 3], dtype="int8")})
res = apply_edits(df, {"edits": [{"rowIdx": 0, "columnId": "a", "value": "5"}]})
print(res["a"].dtype, list(res["a"]))
```

Output:
```
object ['5', 2, 3]     # int8  -> column corrupted to strings (BUG)
```
vs. the `int16` control, which works correctly:
```
int64 [5, 2, 3]
```

`int8` / `uint8` / `float16` are standard memory-optimized dtypes, so this hits real dataframes.

### Fix

Use narwhals' dtype-category predicates (`dtype.is_integer()` / `dtype.is_float()`) instead of hand-enumerating widths — that covers every integer/float width and can't miss one again. (Category methods are needed rather than adding `nw.Int8`/`nw.Float16` explicitly, because `nw.Float16` doesn't exist as an attribute on the minimum supported narwhals.)

PR incoming.

winklemad:
### What / why

Editing a cell in a numeric column whose dtype is `Int8`, `UInt8`, `Int128`, `UInt128`, or `Float16` silently corrupted the column. `_convert_value` hand-enumerated only the 16/32/64-bit integer and float widths, so any other width fell through to `return str(value)`:

- **pandas** — the edited value became a string and the whole column was coerced to `object`.
- **polars** — the edit failed with a misleading `ValueError: Data editor does not support this type of data`.

```python
import pandas as pd
from marimo._plugins.ui._impl.data_editor import apply_edits
df = pd.DataFrame({"a": pd.array([1, 2, 3], dtype="int8")})
res = apply_edits(df, {"edits": [{"rowIdx": 0, "columnId": "a", "value": "5"}]})
print(res["a"].dtype, list(res["a"]))
# before: object ['5', 2, 3]   (corrupted)
# after:  int64  [5, 2, 3]
```

### Fix

Replace the hand-enumerated width lists with narwhals' dtype-category predicates (`dtype.is_integer()` / `dtype.is_float()`), which cover every integer/float width and can't miss one again. The category methods are used rather than adding `nw.Int8`/`nw.Float16` explicitly because `nw.Float16` does not exist as an attribute on the minimum supported narwhals (2.0.0), so referencing it would `AttributeError`.

### Tests

Added direct `_convert_value` tests (Int8/UInt8/Int128/UInt128/Float16) and dataframe-level dtype-preservation tests (pandas int8/uint8/float16 + polars Int8). All fail on the previous code and pass now. `pytest tests/_plugins/ui/_impl/test_data_editor.py` → 105 passed; `ruff check`/`ruff format` clean.

Fixes #10210

### DIFF
--- a/marimo/_plugins/ui/_impl/data_editor.py
+++ b/marimo/_plugins/ui/_impl/data_editor.py
@@ -340,14 +340,8 @@ def _convert_value(
             elif dtype == nw.Duration:
                 return datetime.timedelta(microseconds=float(value))
             elif hasattr(dtype, "is_float") and dtype.is_float():
-                # Any float width (Float32/Float64, plus Float16 where the
-                # installed narwhals exposes it).
                 return float(value)
             elif hasattr(dtype, "is_integer") and dtype.is_integer():
-                # Any signed/unsigned integer width. Enumerating them by hand
-                # missed the 8- and 128-bit widths, so edits to Int8/UInt8/
-                # Int128/UInt128 columns fell through to str() below and
-                # silently coerced the whole column to object/string.
                 return int(value)
             elif (
                 dtype == nw.String
--- a/marimo/_plugins/ui/_impl/data_editor.py
+++ b/marimo/_plugins/ui/_impl/data_editor.py
@@ -339,16 +339,15 @@ def _convert_value(
                 return datetime.date.fromisoformat(value)
             elif dtype == nw.Duration:
                 return datetime.timedelta(microseconds=float(value))
-            elif dtype == nw.Float32 or dtype == nw.Float64:
+            elif hasattr(dtype, "is_float") and dtype.is_float():
+                # Any float width (Float32/Float64, plus Float16 where the
+                # installed narwhals exposes it).
                 return float(value)
-            elif (
-                dtype == nw.Int16
-                or dtype == nw.Int32
-                or dtype == nw.Int64
-                or dtype == nw.UInt16
-                or dtype == nw.UInt32
-                or dtype == nw.UInt64
-            ):
+            elif hasattr(dtype, "is_integer") and dtype.is_integer():
+                # Any signed/unsigned integer width. Enumerating them by hand
+                # missed the 8- and 128-bit widths, so edits to Int8/UInt8/
+                # Int128/UInt128 columns fell through to str() below and
+                # silently coerced the whole column to object/string.
                 return int(value)
             elif (
                 dtype == nw.String
--- a/tests/_plugins/ui/_impl/test_data_editor.py
+++ b/tests/_plugins/ui/_impl/test_data_editor.py
@@ -997,38 +997,23 @@ class TestConvertValue:
         assert result == 42
         assert isinstance(result, int)
 
-    def test_convert_value_with_dtype_int8(self):
-        """Test Int8 conversion with dtype."""
-        result = _convert_value("42", None, nw.Int8)
-        assert result == 42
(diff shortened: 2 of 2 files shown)


winklemad:

tests/_plugins/ui/_impl/test_data_editor.py
```diff
@@ -963,6 +997,39 @@ def test_convert_value_with_dtype_uint64(self):
...
+    def test_convert_value_with_dtype_uint128(self):
+        """Test UInt128 conversion with dtype."""
+        result = _convert_value("42", None, nw.UInt128)
+        assert result == 42
+        assert isinstance(result, int)
+
+    def test_convert_value_with_dtype_float16(self):
+        """Test Float16 conversion with dtype."""
+        float16 = getattr(nw, "Float16", None)
+        if float16 is None:
```
Light2Dark: could we instead parametrize these tests. it's verbose

marimo/_plugins/ui/_impl/data_editor.py
```diff
@@ -339,16 +339,15 @@ def _convert_value(
...
-                or dtype == nw.Int64
-                or dtype == nw.UInt16
-                or dtype == nw.UInt32
-                or dtype == nw.UInt64
-            ):
+            elif hasattr(dtype, "is_integer") and dtype.is_integer():
+                # Any signed/unsigned integer width. Enumerating them by hand
+                # missed the 8- and 128-bit widths, so edits to Int8/UInt8/
+                # Int128/UInt128 columns fell through to str() below and
+                # silently coerced the whole column to object/string.
```
Light2Dark: I would reduce the comments, it's unecessary.

winklemad:
Thanks @Light2Dark! Addressed both:

- Parametrized the numeric-width tests into a single `test_convert_value_covers_narrow_numeric_widths` (Int8/UInt8/Int128/UInt128 + Float16).
- Trimmed the comments on the `is_float`/`is_integer` branches.

`pytest tests/_plugins/ui/_impl/test_data_editor.py` is green (105 passed) and `ruff check`/`format` are clean. I also switched the title to the `fix:` prefix so the label check passes.

FWIW the two red checks look unrelated to this change — a snapshot diff in `test_cli_export.py::test_export_ipynb_with_media_outputs` and a flaky `test_module_watcher.py::test_reload_third_party[k]`; the data_editor tests pass on Linux and macos-latest and are skipped on Windows (polars not installed there).

Light2Dark (approved):
thanks!

PR #10219, opened 2026-07-17
fix(i18n): sanitize browser locales like en-US@posix

### LINKED ISSUE #9938: Frontend crashes when navigator.language is en-US@posix
### Description

marimo's frontend crashes before rendering a notebook when Chromium reports `navigator.language` as `en-US@posix`.

The crash reproduces with the built-in `intro.py` tutorial notebook, so it does not appear to be notebook-specific.

### Error

```text
RangeError: Incorrect locale information provided
```

The visible page shows:

```text
Something went wrong
Incorrect locale information provided
If this is an issue with marimo, please report it on GitHub.
Try again
```

### Reproduction

Run a stock tutorial notebook:

```bash
marimo run /usr/local/lib/python3.11/site-packages/marimo/_tutorials/intro.py --host 0.0.0.0 --port 2718
```

Open it from Playwright Chromium where:

```js
navigator.language === "en-US@posix"
navigator.languages === ["en-US@posix"]
Intl.DateTimeFormat().resolvedOptions().locale === "en-US"
```

The frontend crashes before notebook content renders.

### Control

Forcing Playwright's browser context locale to `en-US` avoids the crash and the stock tutorial renders normally.

This suggests the frontend should normalize or guard browser locale tags before passing them to `Intl.Locale`.

### Versions

- marimo `0.23.9`
- Also observed with marimo `0.23.10`
- Chromium via Playwright Docker image
- App runtime tested with `python:3.11-slim`
- Host: macOS

Bartok9:
**This pull request was AI-assisted; human-reviewed line by line.**

## Summary
Some Chromium/Playwright environments report navigator.language as en-US@posix. Passing that tag straight into react-aria / Intl throws RangeError and the notebook never renders (including the stock intro tutorial).

## Changes
- Add normalizeBrowserLocale / tighter isValidLocale (modifier strip + supportedLocalesOf)
- Route all safeLocale fallbacks through normalization with en-US last resort
- Use the same helper for the data-table default formatting locale
- Regression tests for @posix, charset suffixes, and junk config locales

Closes #9938

### DIFF
--- a/frontend/src/core/i18n/locale-provider.tsx
+++ b/frontend/src/core/i18n/locale-provider.tsx
@@ -65,8 +65,10 @@ export function isValidLocale(locale: string): boolean {
 }
 
 export function safeLocale(locale: string | null | undefined): string {
+  // Always return a BCP 47 tag. isValidLocale accepts underscore forms
+  // (e.g. en_US) for validation only — I18nProvider/Intl need hyphens.
   if (locale && isValidLocale(locale)) {
-    return locale;
+    return normalizeBrowserLocale(locale);
   }
   return normalizeBrowserLocale(
     typeof navigator !== "undefined" ? navigator.language : undefined,
--- a/frontend/src/core/i18n/locale-provider.tsx
+++ b/frontend/src/core/i18n/locale-provider.tsx
@@ -4,75 +4,18 @@ import { useAtomValue } from "jotai";
 import type { ReactNode } from "react";
 import { I18nProvider } from "react-aria-components";
 import { localeAtom } from "@/core/config/config";
+import { safeLocale } from "./locale";
+
+// Re-export pure helpers so existing imports keep working.
+export {
+  FALLBACK_LOCALE,
+  isValidLocale,
+  normalizeBrowserLocale,
+  safeLocale,
+} from "./locale";
 
 interface LocaleProviderProps {
   children: ReactNode;
-}
-
-/** Default when no configurable or browser locale is usable. */
-export const FALLBACK_LOCALE = "en-US";
-
-/**
- * Normalize browser / config locale tags for Intl and react-aria.
- *
- * Some Chromium/Playwright builds report tags like `en-US@posix` that throw
- * `RangeError: Incorrect locale information provided` when passed to Intl.
- */
-export function normalizeBrowserLocale(tag: string | null | undefined): string {
-  if (!tag) {
-    return FALLBACK_LOCALE;
-  }
-
-  // Strip locale modifiers (`@posix`, `.UTF-8`) that are not BCP 47.
-  let candidate = tag.split("@", 1)[0]?.trim() ?? "";
-  const dotIdx = candidate.indexOf(".");
-  if (dotIdx > 0) {
-    candidate = candidate.slice(0, dotIdx);
-  }
-  // Underscores (some POSIX / Java locales) → BCP47 hyphens
-  candidate = candidate.replaceAll("_", "-");
-
-  if (candidate && isValidLocale(candidate)) {
-    return candidate;
-  }
-
-  // language-only, e.g. `en` from `en-US@posix` after failures
-  const language = candidate.split("-", 1)[0];
-  if (language && isValidLocale(language)) {
-    return language;
-  }
-
-  return FALLBACK_LOCALE;
-}
-
-export function isValidLocale(locale: string): boolean {
-  try {
-    // Reject modifiers that break Intl in some Chromium builds (`en-US@posix`).
-    if (locale.includes("@") || locale.includes(".")) {
-      return false;
-    }
-    const normalized = locale.replaceAll("_", "-");
-    // supportedLocalesOf returns [] for unknown tags; more reliable than
-    // NumberFormat which wrongly accepts many junk strings.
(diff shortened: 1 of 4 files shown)


Bartok9:

frontend/src/core/i18n/locale-provider.tsx
```diff
@@ -9,27 +9,72 @@ interface LocaleProviderProps {
...
+    new Intl.Locale(normalized);
     return true;
   } catch {
     return false;
   }
 }
+
+export function safeLocale(locale: string | null | undefined): string {
+  if (locale && isValidLocale(locale)) {
+    return locale;
```
Sanjays2402: `isValidLocale` normalizes underscores only for validation, but this returns the raw value. A configured `en_US` passes the check and still reaches `I18nProvider` as an invalid locale; should this return `normalizeBrowserLocale(locale)`?

Bartok9: Good catch — fixed in the latest commit.

`isValidLocale` was rewriting `_` → `-` only inside the check, then `safeLocale` returned the original string, so a configured `en_US` could pass validation and still reach `I18nProvider` raw.

Valid config locales now go through `normalizeBrowserLocale` before handoff (same path as browser tags). Added a regression test: `safeLocale("en_US") === "en-US"`.

Bartok9:
Thanks @Sanjays2402 — addressed your review:

- `safeLocale` now returns `normalizeBrowserLocale(locale)` for valid configured locales (so `en_US` becomes `en-US` before `I18nProvider`)
- Regression test added for the underscore-config case

Ready for another look.

frontend/src/core/i18n/locale-provider.tsx
Bartok9: Good catch — fixed in 0a4f6aa.

`safeLocale` now always runs valid config through `normalizeBrowserLocale`, so underscore forms like `en_US` are canonicalized to `en-US` before they reach `I18nProvider`. Added a unit test for that path.

Bartok9:
Addressed @Sanjays2402's review: `safeLocale` now returns `normalizeBrowserLocale(locale)` for valid config (so `en_US` → `en-US` for I18nProvider). Tests green (`locale-provider.test.tsx`, 14 passing).

kirangadhave:
@Light2Dark since you attempted a fix previously can you please evaluate this PR?

Bartok9:
### Evaluation notes for @Light2Dark (and anyone reviewing)

@kirangadhave asked for your eyes since you previously attempted this in #9945 — thank you. Here's a concise self-eval of **this PR vs that attempt**, plus what I just pushed to make review easier.

#### Problem (issue #9938)
Chromium/Playwright can report `navigator.language` as `en-US@posix`. Passing that into `Intl` / react-aria throws `RangeError: Incorrect locale information provided` and the notebook never renders (including the stock intro tutorial).

#### How this differs from closed #9945
| | **#9945 (yours)** | **#10219 (this)** |
|--|--|--|
| Approach | `browser-locale.ts` + often **omit** locale prop so react-aria picks default | Always hand `I18nProvider` a **safe BCP 47** tag via `safeLocale` |
| `@posix` | Ended at strip-or-fallback (after review) | Strip `@…` / `.charset`, underscore→hyphen, re-validate, language-only fallback, then `en-US` |
| Config path | Validated user config; browser path separate | Config **and** browser go through the same normalizer (incl. `en_US` → `en-US`) |
| Data table | `DEFAULT_LOCALE` constant | `normalizeBrowserLocale(navigator.language)` so tables match browser when valid |
| Settings UI filter | Filtered `navigator.languages` in the locale dropdown | Not touched (happy to add if you want parity) |

Neither approach is wrong; this one is a bit more aggressive about **always** supplying a concrete locale so we never depend on react-aria’s internal default path for the crash case.

#### Latest push (`ff3c6dad9`) — review feedback folded in
While you evaluate, I addressed open nits so you don’t have to chase them:

1. **Pure module** `frontend/src/core/i18n/locale.ts` — helpers no longer live only in the React `locale-provider.tsx`, so `column-formatting/feature.ts` doesn’t pull jotai/react-aria (Copilot).
2. **`Intl.Locale` optional** — if `Intl.Locale` is missing, we don’t treat every locale as invalid (Copilot).
3. **Config with modifiers** — `safeLocale("en-US@posix")` / `en_US.UTF-8` recover the base tag instead of discarding config and jumping to navigator (Copilot low-confidence note).
4. **Tests** — `vi.stubGlobal` + `unstubAllGlobals` (no leaked navigator). **17/17** vitest green locally on `locale-provider.test.tsx`.
5. Rebased onto current `main` (0 behind).

#### What I’d love from the evaluation
- Does always-passing a locale (vs omitting for system default) match marimo’s preferred i18n model?
- Want the settings dropdown filter from #9945 folded in here?
- Any preference to rename `FALLBACK_LOCALE` → `DEFAULT_LOCALE` for consistency with your earlier naming?

Happy to adjust immediately off your notes. Thanks again for looking.

Light2Dark:
Thanks, I've made a fix here https://github.com/marimo-team/marimo/pull/10318 to catch some other edge cases too.

Bartok9:
Thanks @Light2Dark — closing as superseded by #10318 is perfect. Appreciate you tightening the edge cases and taking #9938 across the finish line.

Issue #10225, opened 2026-07-18
modulefinder crashes on marimo's own namespace packages (breaks autoreload for editable installs)


Felipenguim:
### Describe the bug

marimo's autoreload uses stdlib `modulefinder` to find a module's dependencies (`ModuleDependencyFinder.find_dependencies` in `marimo/_runtime/reload/autoreload.py`). If the analyzed file imports `marimo`, and `marimo` isn't excluded from the crawl, `modulefinder` tries to walk marimo's own package tree, which has several namespace packages (folders with no `__init__.py`, like `marimo/_dependencies`). Stdlib `modulefinder` doesn't support namespace packages and crashes on them.

The crash is caught, so it doesn't break the kernel. Autoreload silently stops tracking that dependency chain, with nothing shown to the user.

### Why this hits contributors, not users

marimo excludes itself from this crawl when its own path contains `site-packages` (`_is_third_party_module` in `module_watcher.py`). true for a normal `pip install marimo`. It's not true for an editable install, so this reproduces reliably in a dev setup but not for regular users.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.14",
  "editable": true,
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "AMD64 Family 23 Model 24 Stepping 1, AuthenticAMD",
  "Python Version": "3.13.3",
  "Locale": "--",
  "Binaries": {
    "Browser": "150.0.7871.127",
    "Node": "v22.16.0",
    "uv": "0.6.14 (a4cec56dc 2025-04-09)"
  },
  "Experimental Flags": {
    "multi_column": true,
    "cache_panel": true,
    "isolate_apps": true
  }
}
```

</details>

### Code to reproduce

```python
# foo.py
import marimo
def f():
    return 1
```

```python
import modulefinder
modulefinder.ModuleFinder(excludes=set()).run_script("uses_marimo.py")
```
```python
AttributeError: 'NoneType' object has no attribute 'is_package'
File ".../modulefinder.py", line 69, in _find_module
if spec.loader.is_package(name):
```

PR #10226, opened 2026-07-18
fix(autoreload): detect transitive changes after cell reloads

### LINKED ISSUE #10191: Cyclical notebook import from another file don't auto-reload
### Describe the bug

If I have a side file that imports from my main notebook which is then imported in a cell of that notebook, saving the notebook will not auto-reload the modules.

https://github.com/user-attachments/assets/30887dd4-1391-4654-b987-9d599dfa3a5f

At 0:13, somehow it actually _does_ rerun the function, but I'm not sure how that is different from all the other times it doesn't.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.14",
  "editable": false,
  "location": "C:/Users/avg170002/AppData/Local/uv/cache/environments-v2/cyclical-auto-reload-998edc1a217398ed/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 198 Stepping 2, GenuineIntel",
  "Python Version": "3.13.9",
  "Locale": "--",
  "Binaries": {
    "Browser": "150.0.7871.115",
    "Node": "v22.23.1",
    "uv": "0.9.28 (0e1351e40 2026-01-29)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.24.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.0",
    "uvicorn": "0.51.0",
    "websockets": "16.1"
  },
  "Optional Dependencies": {
    "loro": "1.13.1"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

**cyclical_auto_reload.py**
```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "marimo>=0.23.3",
# ]
# ///

import marimo

__generated_with = "0.23.9"
app = marimo.App(width="medium", auto_download=["html"])


@app.function
def my_func():
    return 20


@app.cell
def _():
    from auto_reload_file import double_func_file

    double_func_file()
    return


if __name__ == "__main__":
    app.run()
```

**auto_reload_file.py**
```python
from cyclical_auto_reload import my_func


def double_func_file():
    return my_func() * 2
```

Felipenguim:
## Summary

Closes #10191.

marimo's module autoreload has two consumers: cell execution reloads modules, while the background watcher computes transitive staleness. Both previously shared one mtime baseline, so a cell reload could consume a file change before the watcher observed it. Cyclic imports such as the reproduction in #10191 could therefore fail to mark dependent cells stale.

The watcher now has an independent observation baseline. Its poll performs one locked filesystem scan that updates the normal reload bookkeeping and compares against the watcher baseline, preserving transitive change detection without doubling filesystem work or lock contention. Module rebinding clears both baselines so an older replacement file does not become invisible.

Coverage includes the cyclic-import reproduction, the cell-execution race, module rebinding, and a focused assertion that watcher polls stat each module once.

Related: #10225, discovered while validating this behavior in an editable development install.

> Written by gpt-5.6 on Codex

### DIFF
--- a/marimo/_runtime/reload/autoreload.py
+++ b/marimo/_runtime/reload/autoreload.py
@@ -207,8 +207,13 @@ class ModuleReloader:
         # module shadowing an installed package).
         self._skip: dict[str, str | None] = {}
 
+        # module-name -> mtime(module modification timestamps) tracked independently of `modules_mtimes`.
+        # Used only by `check_for_watcher` to avoid race conditions between the watcher and the AutoreloadManager
+        self.watcher_modules_mtimes: dict[str, float] = {}
+
         # Timestamp existing modules
         self.check(modules=sys.modules, reload=False)
+        self.check_for_watcher(modules=sys.modules)
 
     def _is_user_module(self, module: types.ModuleType) -> bool:
         """True for modules whose source lives outside stdlib/site-packages.
@@ -380,6 +385,41 @@ class ModuleReloader:
         return self._module_dependency_finder.find_dependencies(
             module, excludes
         )
+
+    def check_for_watcher(
+        self, modules: dict[str, types.ModuleType]
+    ) -> set[types.ModuleType]:
+        """Like `check(..., reload=False)`, but tracks mtimes in
+        `watcher_mtimes` instead of `modules_mtimes`.
+
+        `ModuleWatcher`'s background thread calls this
+        to find modules that changed since it last looked, in order to
+        compute *transitive* staleness via `_depends_on` in
+        `module_watcher.py`. It needs its own method to still call the check
+        but to avoid the race condition and all the cells update correctly
+        """
+        with self.lock:
+            modified_modules: set[types.ModuleType] = set()
+            for modname in list(modules.keys()):
+                m = modules.get(modname, None)
+                if m is None:
+                    continue
+
+                module_mtime = self.filename_and_mtime(m)
+                if module_mtime is None:
+                    continue
+                pymtime = module_mtime.mtime
+
+                existing_mtime = self.watcher_modules_mtimes.get(modname)
+                if existing_mtime is None:
+                    self.watcher_modules_mtimes[modname] = pymtime
+                    continue
+                if pymtime <= existing_mtime:
+                    continue
+
+                self.watcher_modules_mtimes[modname] = pymtime
+                modified_modules.add(m)
+            return modified_modules
 
 
 def update_function(old: object, new: object) -> None:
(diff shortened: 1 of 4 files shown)


marimo/_runtime/reload/autoreload.py
```diff
@@ -381,6 +386,41 @@ def get_module_dependencies(
...
+                m = modules.get(modname, None)
+                if m is None:
+                    continue
+
+                module_mtime = self.filename_and_mtime(m)
+                if module_mtime is None:
+                    continue
+                pymtime = module_mtime.mtime
+
+                existing_mtime = self.watcher_modules_mtimes.get(modname)
```
Sanjays2402: This separate map never gets cleared when a module name is rebound to another `__file__`. `check()` explicitly drops `modules_mtimes` on that path but leaves this old baseline behind, so a replacement file with an older mtime and subsequent edits stay invisible until its timestamp passes the previous file’s.

Felipenguim: 52527f2d7
Now in the function```check()``` the  ```watcher_modules_mtimes``` also gets cleared to deal with modules being rebound

f3dafec26  
A new test to ensure it

Light2Dark (commented):
Coding-agent review after rebasing onto the latest `main`:

- The conflict in `module_watcher.py` is resolved while preserving the newer non-user-module skip-cache behavior from `main`.
- I found and fixed a performance issue in the proposed implementation: the watcher statted every loaded module twice and acquired the reloader lock twice per poll. `check_for_watcher()` now updates the normal reload bookkeeping and the independent watcher baseline in one locked scan.
- The watcher baseline is cleared when a cached module name is rebound to a different file.
- The cyclic-import reproduction and the race where another cell reloads first are covered end to end. A focused test also guards the single-stat-per-module behavior.

I found no remaining blocking correctness issues in the updated diff.

Validation: `124 passed, 2 xpassed`; ruff, formatting, copyright, and typo checks passed; mypy passed for both changed source modules. The repository-wide mypy invocation also reported two unrelated existing errors in `marimo/_cli/parser_ux.py` and `marimo/_cli/errors.py`.

Light2Dark:
CI is unrelated, i have another PR for that

dmadisetti:
Hey @Felipenguim I'm squashing this into #10912 and keeping your attribution since this got a little stale. Closing out for now, but I'm adding you as a reviewer too @Light2Dark if that's alright

Felipenguim:
> Hey @Felipenguim I'm squashing this into #10912 and keeping your attribution since this got a little stale. Closing out for now, but I'm adding you as a reviewer too @Light2Dark if that's alright

No problem, that's ok! Thank you!

PR #10237, opened 2026-07-19
fix: don't transform main guard inside strings/comments in convert

vidigoat:

### 📝 Summary

`marimo convert` corrupts valid Python scripts when the text `if __name__ == "__main__":` appears inside a string literal or a comment.

`_transform_main_blocks` located the guard with a naive `cell.code.split('if __name__ == "__main__":', 1)`, so any occurrence of that text — even inside a string or comment — was treated as a real guard. The cell was sliced mid-token and rewritten to `def _main_(): …`, producing syntactically invalid code that is then emitted as an `app._unparsable_cell(...)` blob.

This affects `marimo convert script.py`, `marimo edit` on a non-marimo `.py` file, and the server template API — both the plain-script and jupytext/pypercent paths.

### 🔍 Reproduction (before this PR)

```python
from marimo._convert.converters import MarimoConvert

# valid Python — the guard text is just a string
src = 'GUARD = \'if __name__ == "__main__":\'\nprint(GUARD)\n'
print(MarimoConvert.from_non_marimo_python_script(src, aggressive=False).to_py())
```

Output on `main` slices the string literal and emits a mangled `app._unparsable_cell(...)`. The same happens for a comment that mentions the guard.

### 🔧 Fix

Detect the guard via the AST: find a top-level `ast.If` whose test is `__name__ == "__main__"` (either operand order) with no `else`, and rewrite only that statement using line/column offsets. Cells that don't parse are left untouched (they're emitted as unparsable cells downstream anyway).

As a bonus this also fixes two cases that were silently *not* transformed before: the single-quoted form (`if __name__ == '__main__':`) and the reversed comparison (`if "__main__" == __name__:`).

### ✅ Tests

- Added `TestTransformMainBlocks` (6 cases: guard-in-string, guard-in-comment, single quotes, reversed comparison, and that a real guard still transforms).
- `uv run --python 3.13 --group test pytest tests/_convert tests/_cli/test_cli_convert.py` → **363 passed**, 2 skipped (env), 1 xfail (pre-existing, unrelated).
- `ruff check`, `ruff format --check`, and `mypy` all clean on the changed files.

### 📋 Checklist

- [x] I have read the [CLA](https://marimo.io/cla)
- [x] For new features, I have added tests
- [x] I have run the tests and they pass

### DIFF
--- a/marimo/_convert/non_marimo_python_script.py
+++ b/marimo/_convert/non_marimo_python_script.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import ast
 import json
 
 from marimo._convert.ipynb.to_ir import convert_from_ipynb_to_notebook_ir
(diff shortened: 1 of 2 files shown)


vidigoat:

marimo/_convert/non_marimo_python_script.py
```diff
@@ -106,19 +107,73 @@ def convert_non_marimo_script_to_notebook_ir(
...
+
+        before_main = cell.code[:start].strip()
+
+        # replace the if __name__ == "__main__": with def _main_():
+        main_block = "def _main_():" + cell.code[colon + 1 :] + "\n\n_main_()"
+
+        if before_main:
+            cell.code = before_main + "\n\n" + main_block
+        else:
+            cell.code = main_block
```
dmadisetti, line 179: I think an AST transform may be better than this. I'm wondering if replacing with `def _main_` makes sense either.

I reckon `if mo.app_meta().mode == "script"` almost makes more sense?

dmadisetti: cc @manzt who last touched this

vidigoat:
Thanks @dmadisetti! Two things:

On the AST point — the PR does use AST to *find* the guard (that's the actual fix here; the old code string-matched `if __name__ == "__main__":` and mangled it whenever that text showed up inside a string literal or comment). I deliberately kept the *replacement* as a string splice rather than `ast.unparse`, because unparsing the whole cell would drop comments and reformat the user's code — which seems bad for a `convert` tool that should round-trip their source as faithfully as possible. Happy to go fully-AST if you'd prefer, though.

On `def _main_()` vs `mo.app_meta().mode == "script"` — I think you're right that the latter is a better fit: it preserves the original "only run when executed directly" semantic instead of wrapping the body in a function and calling it. I kept `def _main_()` for now, but I'm glad to rework toward `if mo.app_meta().mode == "script":` if that's the direction you and @manzt want.

It's still a draft, so happy to settle the approach first before polishing.

dmadisetti:
> unparsing the whole cell would drop comments and reformat the user's code

Agreed here. Cool I think this makes sense

vidigoat:
Thanks @dmadisetti! Glad the string-splice reasoning makes sense. The one open question is the target: keep `def _main_(): ... _main_()`, or switch the guard body to run under `if mo.app_meta().mode == "script":` as you suggested (which arguably preserves the original "only when run as a script" semantic more faithfully). I'm happy to go either way — just let me know which you and @manzt prefer and I'll finalize and take it out of draft.

PR #10239, opened 2026-07-20
fix: pin pydantic-ai instrumentation version

### LINKED ISSUE #10238: Pydantic-AI tracing marks deferred frontend tools as errors at the supported dependency floor
### Describe the bug

Marimo supports `pydantic-ai-slim>=1.107.0,<3.0.0`, but the current AI tracing setup does not specify which instrumentation version should be used:

```python
Agent.instrument_all(
    InstrumentationSettings(tracer_provider=provider)
)
```

Because of this, the tracing behavior changes depending on the installed Pydantic-AI version.

In Pydantic-AI 1.107.0, `InstrumentationSettings` defaults to version 2. When a Marimo frontend tool raises `CallDeferred`, the tool is deferred correctly and the agent returns `DeferredToolRequests`. However, OpenTelemetry records the tool span as an error:

```text
span: running tool
status: ERROR
exception: pydantic_ai.exceptions.CallDeferred
```

In this case, `CallDeferred` is not an actual failure. Marimo's `form_toolsets()` intentionally raises it when a tool needs to be handled by the frontend.

When version 5 is used, the same tool call is recorded correctly:

```text
span: execute_tool frontend_tool
status: UNSET
exception events: none
```

The agent still returns the same `DeferredToolRequests` result. The only difference is that normal frontend-tool behavior is no longer reported as an error.

I reproduced this on current `main` using:

- Marimo's `form_toolsets()` implementation;
- a `ToolDefinition` with `source="frontend"`;
- a Pydantic-AI `FunctionModel`;
- Marimo's current `_instrument_ai()` setup;
- an in-memory OpenTelemetry span exporter.

No API keys or external services were needed.

These were the results across the supported dependency range:

| Pydantic-AI version | Instrumentation version used by current Marimo | Result |
| --- | ---: | --- |
| `1.107.0` | 2 | `CallDeferred` creates an `ERROR` tool span |
| `1.107.0` with `version=5` | 5 | The tool span remains `UNSET` with no exception event |
| `2.9.0` | 5 by default | Correct behavior |

This behavior was previously handled in #9969, which passed `version=5` and added a test assertion for it. The version argument and test assertion were later removed in #9973 as part of the Pydantic-AI v2 support changes.

Both Pydantic-AI 1.107.0 and 2.9.0 accept `version=5`, so selecting version 5 works across Marimo's currently supported dependency range.

I would expect deferred frontend tools to remain normal control flow and not appear as application errors, regardless of which supported Pydantic-AI version is installed.

A small fix would be to restore `version=5`:

```python
Agent.instrument_all(
    InstrumentationSettings(
        tracer_provider=provider,
        version=5,
    )
)
```

The regression test should also check that Marimo passes `version=5` when constructing `InstrumentationSettings`. Only checking `settings.version` would not catch the problem with newer Pydantic-AI versions because version 5 is already their default.

I searched the open issues and pull requests for `CallDeferred`, `ApprovalRequired`, `InstrumentationSettings`, instrumentation version 5, and related AI tracing errors, but I did not find another report for this behavior.

I would be happy to submit the fix and regression test if this approach looks good.

### Will you submit a PR?

- [x] Yes

### Environment

<details>
<summary>Environment details</summary>

```json
{
  "marimo": "0.23.14",
  "editable": true,
  "location": "/Users/aadetya/Projects/marimo/marimo",
  "OS": "Darwin",
  "OS Version": "25.5.0",
  "Processor": "arm",
  "Python Version": "3.13.12",
  "Locale": "C/C",
  "Binaries": {
    "Browser": "150.0.7871.125",
    "Node": "v24.11.1",
    "uv": "0.11.3 (Homebrew 2026-04-01 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.23.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.0",
    "uvicorn": "0.51.0",
    "websockets": "16.1"
  },
  "Optional Dependencies": {
    "altair": "6.2.2",
    "anywidget": "0.9.21",
    "basedpyright": "1.39.9",
    "duckdb": "1.5.4",
    "ibis-framework": "12.0.0",
    "loro": "1.13.1",
    "mcp": "1.28.1",
    "nbformat": "5.10.4",
    "openai": "2.45.0",
    "pandas": "3.0.3",
    "polars": "1.42.1",
    "pyarrow": "25.0.0",
    "pytest": "9.0.3",
    "ruff": "0.15.21",
    "sqlglot": "30.12.0",
    "vegafusion": "2.0.3"
  },
  "Experimental Flags": {
    "multi_column": true,
    "cache_panel": true,
    "isolate_apps": true
  }
}
```

Additional dependency versions used during the reproduction:

```text
pydantic-ai-slim[openai] == 1.107.0
pydantic-ai-slim[openai] == 2.9.0
opentelemetry-sdk
```

</details>

### Code to reproduce

Save the following as `repro_deferred_tracing.py` in a current Marimo checkout:

```python
from unittest.mock import AsyncMock

from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
    InMemorySpanExporter,
)
from pydantic_ai import Agent, DeferredToolRequests
from pydantic_ai.messages import ModelResponse, ToolCallPart
from pydantic_ai.models.function import FunctionModel

from marimo._ai._pydantic_ai_utils import form_toolsets
from marimo._server.ai.tools.types import ToolDefinition
from marimo._tracer import _instrument_ai


def model_response(messages, info):
    del messages, info
    return ModelResponse(
        parts=[
            ToolCallPart(
                tool_name="frontend_tool",
                args={"value": "test"},
                tool_call_id="call-1",
            )
        ]
    )


tool = ToolDefinition(
    name="frontend_tool",
    description="A frontend tool",
    parameters={
        "type": "object",
        "properties": {"value": {"type": "string"}},
    },
    source="frontend",
    mode=["manual"],
)

toolset, _ = form_toolsets([tool], AsyncMock())

exporter = InMemorySpanExporter()
provider = TracerProvider()
provider.add_span_processor(SimpleSpanProcessor(exporter))

# This uses Marimo's current instrumentation configuration.
_instrument_ai(provider)

agent = Agent(
    FunctionModel(model_response),
    toolsets=[toolset],
    output_type=[str, DeferredToolRequests],
)

result = agent.run_sync("Use the frontend tool")

print("Result:", type(result.output).__name__)

for span in exporter.get_finished_spans():
    exceptions = [
        event.attributes.get("exception.type")
        for event in span.events
        if event.name == "exception"
    ]
    print(
        "Span:",
        span.name,
        "Status:",
        span.status.status_code.name,
        "Exceptions:",
        exceptions,
    )
```

Run it from the repository root with Marimo's supported dependency floor:

```bash
uv run \
  --no-project \
  --isolated \
  --with-editable . \
  --with "pydantic-ai-slim[openai]==1.107.0" \
  --with opentelemetry-sdk \
  python repro_deferred_tracing.py
```

Current output includes:

```text
Result: DeferredToolRequests
Span: running tool Status: ERROR Exceptions: ['pydantic_ai.exceptions.CallDeferred']
```

Changing Marimo's instrumentation configuration to include `version=5` produces the same `DeferredToolRequests` result, but the tool span remains `UNSET` and has no exception event.

aadetya:
## 📝 Summary

Closes #10238

This sets the Pydantic-AI instrumentation version to 5 so deferred frontend tool calls are not reported as tracing errors when using older supported Pydantic-AI versions.

I also updated the existing test to make sure Marimo actually passes `version=5` when creating the instrumentation settings. Checking only the final value would not catch the same problem on newer Pydantic-AI versions, since they already default to version 5.

Tested with Pydantic-AI 1.107.0 and 2.9.0.

## 📋 Pre-Review Checklist

- [x] This change was discussed and approved in #10238.
- [ ] Video or media evidence is provided for any visual changes (not applicable).

## ✅ Merge Checklist

- [x] Documentation has been updated where applicable, including docstrings for API changes. No documentation changes were needed here.

### DIFF
--- a/marimo/_tracer.py
+++ b/marimo/_tracer.py
@@ -250,7 +250,10 @@ def _instrument_ai(provider: trace.TracerProvider) -> None:
         from pydantic_ai import Agent
         from pydantic_ai.models.instrumented import InstrumentationSettings
 
-        Agent.instrument_all(InstrumentationSettings(tracer_provider=provider))
+        # Version 5 treats deferred tool calls as control flow, not errors.
+        Agent.instrument_all(
+            InstrumentationSettings(tracer_provider=provider, version=5)
+        )
         LOGGER.debug("Enabled AI instrumentation")
     except Exception as e:
         LOGGER.debug("AI instrumentation failed: %s", e)
--- a/marimo/_tracer.py
+++ b/marimo/_tracer.py
@@ -250,7 +250,6 @@ def _instrument_ai(provider: trace.TracerProvider) -> None:
         from pydantic_ai import Agent
         from pydantic_ai.models.instrumented import InstrumentationSettings
 
-        # Version 5 treats deferred tool calls as control flow, not errors.
         Agent.instrument_all(
             InstrumentationSettings(tracer_provider=provider, version=5)
         )
--- a/tests/test_tracer.py
+++ b/tests/test_tracer.py
@@ -435,6 +435,8 @@ class TestInstrumentAI:
         ):
             _instrument_ai(provider)
 
+        # Verify version 5 is passed explicitly instead of relying on newer
+        # Pydantic-AI defaults.
         mock_settings.assert_called_once_with(
             tracer_provider=provider,
             version=5,
--- a/tests/test_tracer.py
+++ b/tests/test_tracer.py
@@ -426,9 +426,19 @@ class TestInstrumentAI:
         from pydantic_ai.models.instrumented import InstrumentationSettings
 
         provider = MagicMock()
-        with patch("pydantic_ai.Agent.instrument_all") as mock_instrument:
+        with (
+            patch("pydantic_ai.Agent.instrument_all") as mock_instrument,
+            patch(
+                "pydantic_ai.models.instrumented.InstrumentationSettings",
+                wraps=InstrumentationSettings,
+            ) as mock_settings,
+        ):
             _instrument_ai(provider)
 
+        mock_settings.assert_called_once_with(
+            tracer_provider=provider,
+            version=5,
+        )
         mock_instrument.assert_called_once()
         settings = mock_instrument.call_args.args[0]
         assert isinstance(settings, InstrumentationSettings)


aadetya:

marimo/_tracer.py
```diff
@@ -250,7 +250,10 @@ def _instrument_ai(provider: trace.TracerProvider) -> None:
         from pydantic_ai import Agent
         from pydantic_ai.models.instrumented import InstrumentationSettings
 
-        Agent.instrument_all(InstrumentationSettings(tracer_provider=provider))
+        # Version 5 treats deferred tool calls as control flow, not errors.
```
Light2Dark:
```suggestion
```

Light2Dark:
you can add a comment to this issue in the test if you want.

aadetya:
Got it, I removed the production comment and moved the explanation to the test, where it clarifies why the explicit `version=5` assertion is required.

Light2Dark (approved):
i'll do a small pass to improve the comment, thanks!

PR #10272, opened 2026-07-22
feat: support caching binary assets via `BlobAsset`

peter-gy:
As part of the upcoming marimo-export solution, we need a way to represent media-typed binary assets like PNGs, SVGs, JSON, or HTML snippets within the new caching infrastructure.

This PR introduces `BlobAsset` for storing arbitrary outputs with a `.bin` extension which can be deserialized by clients based on `BlobAsset.media_type` and `BlobAsset.metadata` contents.

### DIFF
--- a/marimo/_save/loaders/lazy.py
+++ b/marimo/_save/loaders/lazy.py
@@ -357,7 +357,7 @@ def to_item(
         ref = _maybe_import_ref(value)
         if ref is not None:
             return Item(import_ref=ref)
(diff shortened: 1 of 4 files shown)


dmadisetti (approved):
Looks, good- is there a followup usage path? Or is this consumed directly?

peter-gy:
> Looks, good- is there a followup usage path? Or is this consumed directly?

For now only directly

PR #10275, opened 2026-07-22
fix(document): apply transactions atomically

kirangadhave:
## 📝 Summary

Make `NotebookDocument.apply()` atomic when validation or change application fails. Previously, invalid cell references could surface as raw `KeyError`s, and an unexpected failure after earlier changes had been applied could leave cells or versions partially mutated.

Validate targets and anchors in batch order, then apply changes to an isolated copy and replace live state only after the complete transaction succeeds. This preserves create-then-reference behavior within a batch while ensuring failed transactions leave the document unchanged.

Add focused regression coverage for missing targets and anchors, self-anchored moves, earlier-created anchors, validation rollback, and unexpected execution failures.

Closes MO-6621

### DIFF
--- a/marimo/_messaging/notebook/document.py
+++ b/marimo/_messaging/notebook/document.py
@@ -149,19 +149,33 @@ class NotebookDocument:
     def apply(self, tx: Transaction) -> Transaction:
         """Validate and apply *tx*, return it with `version` assigned.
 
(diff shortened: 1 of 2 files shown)


manzt (approved):
Sorry for dragging on the review. Looks great.

PR #10295, opened 2026-07-23
fix(export): wrap long code lines in WebPDF export

### LINKED ISSUE #9421: Export to pdf truncates code cells
### Describe the bug

When I export a notebook to pdf from the command line (
  uv run marimo export pdf
), the output includes the code cells as desired, but longer lines are truncated. Can we change the font size or margins during export to avoid this? Alternatively, perhaps the lines could be wrapped.

When I try to print to pdf from inside the Chrome browser, the code blocks don't print at all, which is unfortunate for my use case.

Separately, it would be nice if there was a way to expand all accordions so that I get a "full" printout. Generally, I seem to be looking for a way to reach under the hood and control the export functionality in more detail. If there's some section of the code that controls this, I'd be happy to poke around a little; I understand that rolling out a more flexible UI for general use could be a lot of work.

The background here is that I'm really enjoying marimo and finding that it gets us very close to self-documenting code. The pdf would be a way of archiving or distributing snapshots without environment dependence.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.1",
  "editable": false,
  "location": "[ … ]/.venv/lib64/python3.12/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "4.18.0-553.120.1.el8_10.x86_64",
  "Processor": "x86_64",
  "Python Version": "3.12.12",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "0.11.7 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.3.2",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.19.0",
    "packaging": "26.1",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "uvicorn": "0.44.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.0.0",
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "pandas": "3.0.2",
    "polars": "1.39.3",
    "pyarrow": "24.0.0",
    "vegafusion": "2.0.3"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

Abdulrehman-PIAIC80387:

## 📝 Summary

Closes marimo-team/marimo#9421

Long lines in code cells are truncated rather than wrapped in PDFs produced via WebPDF, silently dropping source text. As @philipdeboer reported, a 79-character comment was cut off after the "ch" in "chosen".

## 🔍 Root cause

JupyterLab's stylesheet (shipped with nbconvert) renders the code input area as:

```html
<div class="jp-CodeMirrorEditor jp-Editor jp-InputArea-editor">
  <div class="highlight hl-ipython3"><pre>…long line…</pre></div>
```

with `.jp-InputArea-editor { overflow: hidden; }` and the `<pre>` left at the default `white-space: pre`. There is no `white-space` / `word-break` / `overflow-wrap` rule for the code **input** area at all — outputs already wrap via `.jp-OutputArea-output pre { word-break: … }`.

So a long line never wraps, overflows its container, and is clipped. In JupyterLab that's fine because the editor scrolls; a PDF has no scrolling, so the text is simply gone.

**The LaTeX path is unaffected** and already breaks lines via `fancyvrb` / `\Wrappedbreaksatpunct` — consistent with @philipdeboer finding the `--no-webpdf` screenshot "promising".

## 🔧 What changed

Registers an nbconvert preprocessor that inlines a stylesheet wrapping the code input area. Preprocessors run *after* nbconvert populates `resources`, so appending to `resources["inlining"]["css"]` is what reaches the rendered HTML — no template files, so nothing new to ship as package data.

```css
.jp-InputArea-editor { overflow: visible !important; }
.jp-InputArea-editor .highlight pre {
  white-space: pre-wrap !important;
  overflow-wrap: anywhere !important;
}
```

* Scoped to the input area; outputs are untouched.
* `pre-wrap` (not `pre-line`) preserves Python's leading indentation.
* `overflow-wrap: anywhere` covers a single unbroken token wider than the page, e.g. a long URL.
* `!important` is needed because this is inlined ahead of the JupyterLab rules — the same approach the slides PDF path already uses to override nbconvert styling.

This sits in `_render_webpdf_with_nbconvert`, so it covers **both** explicit `--webpdf` and the fallback taken when standard LaTeX export fails. The fallback is likely the more common path — see the note below.

## 🧪 Verification

Generated real PDFs through the actual WebPDF export path, then extracted their text:

| probe | before | after |
| -- | -- | -- |
| `to be chosen` (the reported line) | ❌ absent | ✅ present |
| `zeta=6` (long call signature) | ❌ absent | ✅ present |
| `cannot/wrap/naturally` (unbreakable URL) | ❌ absent | ✅ present |

The characters are genuinely missing from the PDF, so this is content loss rather than a purely visual clip.

image

image

Also:

* Added a regression test asserting the stylesheet actually reaches the rendered document — a misresolved preprocessor would otherwise fail silently and leave the bug in place.
* `tests/_server/export/` goes from 55 → 56 passing; the 20 pre-existing failures in that file are unrelated (identical on a clean `main`).
* `mypy` and `ruff check` / `ruff format` clean on the touched files.

## 📋 Note — possibly related, not fixed here

`export_as_pdf` catches `OSError`, `PandocMissing`, `ConversionException` and bare `Exception`, then silently falls back to WebPDF. A failing LaTeX run therefore produces a successful-looking export that is actually WebPDF output, with no indication of which renderer ran. That may be why installing TeX still gave @philipdeboer the same result. It seems worth a separate issue rather than bundling here — happy to file one.

## 📋 Pre-Review Checklist

- [X] For large changes, or changes that affect the public API: discussed on marimo-team/marimo#9421, including the choice of injection mechanism.

## ✅ Merge Checklist

- [X] I have read the contributor guidelines.
- [X] Documentation has been updated where applicable — no user-facing API or docs change.

### DIFF
--- a/marimo/_export/exporter.py
+++ b/marimo/_export/exporter.py
@@ -119,23 +119,15 @@ WEBPDF_CODE_WRAP_CSS = """\
 """
 
 
-def _wrap_code_lines_preprocessor() -> Any:
-    """Nbconvert preprocessor that inlines `WEBPDF_CODE_WRAP_CSS`.
+def _inline_code_wrap_css(nb: Any, resources: Any) -> tuple[Any, Any]:
+    """Inline `WEBPDF_CODE_WRAP_CSS`, as an nbconvert preprocessor.
 
     Preprocessors run after nbconvert populates `resources`, so appending here
     is what gets the stylesheet into the rendered HTML.
     """
-    from nbconvert.preprocessors import (  # type: ignore[import-not-found]
-        Preprocessor,
-    )
-
-    class WrapCodeLinesPreprocessor(Preprocessor):  # type: ignore[misc]
-        def preprocess(self, nb: Any, resources: Any) -> Any:
-            inlining = resources.setdefault("inlining", {})
-            inlining.setdefault("css", []).append(WEBPDF_CODE_WRAP_CSS)
-            return nb, resources
-
-    return WrapCodeLinesPreprocessor
+    inlining = resources.setdefault("inlining", {})
+    inlining.setdefault("css", []).append(WEBPDF_CODE_WRAP_CSS)
+    return nb, resources
 
 
 def _render_webpdf_with_nbconvert(notebook: Any, include_inputs: bool) -> Any:
@@ -152,7 +144,7 @@ def _render_webpdf_with_nbconvert(notebook: Any, include_inputs: bool) -> Any:
     web_exporter.exclude_input = not include_inputs
     web_exporter.allow_chromium_download = True
     web_exporter.register_preprocessor(  # type: ignore[no-untyped-call]
-        _wrap_code_lines_preprocessor(), enabled=True
+        _inline_code_wrap_css, enabled=True
     )
     pdf_data, _resources = web_exporter.from_notebook_node(notebook)  # type: ignore[no-untyped-call]
     return pdf_data
--- a/marimo/_export/exporter.py
+++ b/marimo/_export/exporter.py
@@ -98,6 +98,46 @@ def _nbconvert_tag_remove_config() -> Config:
     return config
 
 
+# JupyterLab's stylesheet (shipped with nbconvert) styles the code input area
+# with `overflow: hidden` and leaves the `<pre>` at the default `white-space:
+# pre`. In JupyterLab that is fine because the editor scrolls; in a PDF there is
+# no scrolling, so long lines are clipped and the text is lost entirely.
+# Outputs already wrap (`.jp-OutputArea-output pre` sets `word-break`), so this
+# only targets the input area.
+#
+# `!important` is required because this is inlined ahead of the JupyterLab
+# rules; the slides PDF path overrides nbconvert styling the same way.
+WEBPDF_CODE_WRAP_CSS = """\
+/* marimo: wrap long code lines instead of clipping them (#9421) */
+.jp-InputArea-editor {
+  overflow: visible !important;
+}
+.jp-InputArea-editor .highlight pre {
+  white-space: pre-wrap !important;
+  overflow-wrap: anywhere !important;
+}
+"""
+
+
+def _wrap_code_lines_preprocessor() -> Any:
+    """Nbconvert preprocessor that inlines `WEBPDF_CODE_WRAP_CSS`.
+
+    Preprocessors run after nbconvert populates `resources`, so appending here
+    is what gets the stylesheet into the rendered HTML.
+    """
+    from nbconvert.preprocessors import (  # type: ignore[import-not-found]
+        Preprocessor,
+    )
+
+    class WrapCodeLinesPreprocessor(Preprocessor):  # type: ignore[misc]
+        def preprocess(self, nb: Any, resources: Any) -> Any:
+            inlining = resources.setdefault("inlining", {})
+            inlining.setdefault("css", []).append(WEBPDF_CODE_WRAP_CSS)
+            return nb, resources
+
+    return WrapCodeLinesPreprocessor
+
+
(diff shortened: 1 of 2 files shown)


tests/_server/export/test_exporter.py
```diff
@@ -1653,6 +1653,51 @@ def _():
...
+
+        # WebPDFExporter renders through the `webpdf` template; asserting on
+        # the HTML avoids needing Chromium in CI.
+        exporter = HTMLExporter(
+            config=_nbconvert_tag_remove_config(), template_name="webpdf"
+        )
+        rendered_without, _ = exporter.from_notebook_node(notebook)
+        assert WEBPDF_CODE_WRAP_CSS not in rendered_without
+
+        exporter.register_preprocessor(_inline_code_wrap_css, enabled=True)
```
peter-gy: Nitpick, but could we directly test `_render_webpdf_with_nbconvert` here? This test registers `_inline_code_wrap_css` itself, so it would still pass if the production registration were removed. We could patch `WebPDFExporter.run_playwright` to capture the generated HTML and return dummy PDF bytes.

peter-gy (dismissed):
Looks great, thank you!

Image

tests/_server/export/test_exporter.py
Abdulrehman-PIAIC80387: Good catch — you're right, and it was a real gap rather than a nitpick: dropping the `register_preprocessor` call from `_render_webpdf_with_nbconvert` left the old test green.

Fixed in c8a212a5e. It now drives `_render_webpdf_with_nbconvert` directly and stubs `WebPDFExporter.run_playwright`, which receives the fully rendered HTML — so the assertion covers the real registration, still without needing Chromium.

Verified by mutation: with the production `register_preprocessor` call removed the test fails, and passes with it restored.

tests/_server/export/test_exporter.py
```diff
@@ -1653,6 +1653,58 @@ def _():
...
+        ):
+            pdf_data = _render_webpdf_with_nbconvert(
+                notebook, include_inputs=True
+            )
+
+        assert pdf_data == b"mock_pdf_data"
+        assert len(rendered) == 1
+        assert WEBPDF_CODE_WRAP_CSS in rendered[0]
+        assert "white-space: pre-wrap !important" in rendered[0]
+
```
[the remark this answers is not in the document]
Abdulrehman-PIAIC80387: Good point — done in 5756b62e3. The test now asserts on the distinctive selector (`.jp-InputArea-editor .highlight pre`) and the two wrapping properties rather than the full constant, so it won't break on nbconvert whitespace/comment handling. Still verified by mutation that dropping the production `register_preprocessor` call makes it fail.

peter-gy (dismissed):
thanks for the follow ups!

Light2Dark:
@Abdulrehman-PIAIC80387 could you fix the merge conflict?

Abdulrehman-PIAIC80387:
Done @Light2Dark — rebased onto latest `main` and resolved the conflict (the export module moved to `marimo/_export/`; changes reapplied there and the test import updated). Tests pass and CI is green.

Light2Dark (approved):

philipdeboer:
Thanks for fixing this!

Issue #7791, opened 2026-01-12
Ibis Deferred expression object should not be handeled as datasource


szst11:
### Describe the bug

If I declare an Ibis [Deferred expression object](https://ibis-project.org/reference/expression-generic.html#ibis.expr.api.deferred) as a variable, it will be handled as a datasource and will throw an error, when the cell is executed.

### Will you submit a PR?

- [x] Yes

### Environment

Marimo 0.19.2
startet with 0.19

- before 0.19 the variable was shown in the datasource panel, but not functional there. Now it brings this error in the notebook

### Code to reproduce

```python
import marimo as mo
import ibis

my_deferred = ibis._['a']
```

will get
```
[W 260112 16:15:26 get_engines:109] Failed to introspect datasource: 'Deferred' object is not iterable
[E 260112 16:15:26 notification_utils:56] Error serializing notification DataSourceConnectionsNotification: maximum recursion depth exceeded in comparison
```

Light2Dark:
Thanks for having a look, it looks like it gets incorrectly identified as a ADBCEngine, 
https://github.com/marimo-team/marimo/blob/1f251095e952d0c90e566e257797b7eafee2ac05/marimo/_sql/get_engines.py#L48-L66

I think we want to add a check here/somewhere above for this type and avoid introspection

PR #10296, opened 2026-07-23
fix(sql): don't classify objects with permissive __getattr__ as connections

### LINKED ISSUE #10213: marimo notebook hangs when using pytorch-ignite metraics
### Describe the bug

Instantiating a any `ignite.metrics.Metric` instance it just hangs that particular cell executing it. I tried and tested this with `ignite.metrics.Loss` and `ignite.metrics.RunningAverage`. It it just remains hung. It just never terminates, the hourglass keeps spinning...

Image

Further when you try to stop the cell by using the interrupt button, it doesn't get interrupted, here is the log message when clicking that button:

Image

I also checked for the specific codebases of [PyTorch Ignite](https://github.com/pytorch/ignite), more specifically the source codes of `ignite.metrics.Loss` and `ignite.metrics.RunningAverage` but couldn't find anything problematic there. At least I didn't see any problem in their `__init__` method. Please look into this. Or suggest a fix. The issue is same when running locally as well as on _molab_

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
OS: Windows 11
Python: 3.14.0 (Miniconda)
marimo: 0.23.14
```

</details>

### Code to reproduce

```python
# In any marimo code cell run the following
import marimo as mo
import torch
from ignite.metrics import Loss, RunningAverage
loss = Loss(torch.nn.CrossEntropyLoss()) # or any other metric practically
```

Jayashanker-Padishala:
**This pull request was authored by a coding agent.** (Per AGENTS.md: authored by Claude Code under the direction of @Jayashanker-Padishala, who takes responsibility for it. Marked as draft per policy.)

Fixes #10213

## 📝 Summary

Instantiating any `ignite.metrics.Metric` hung the executing cell forever. Root cause (found via faulthandler stack dump of the hung kernel):

1. ignite metrics implement `__getattr__` to return a lazy `MetricsLambda` factory for **any** attribute name — and each access runs `traceback.format_stack()` (ignite `metric.py:797`), so every attribute probe is also slow.
2. That makes every `callable(getattr(var, method, None))` duck-typing check succeed, so the post-execution `_broadcast_data_source_connection` hook classified the metric as an **ADBC connection**.
3. `AdbcConnectionCatalog.get_databases` → `adbc_get_objects(...).read_all().to_pylist()` then walked into the lazy object graph and never produced real data — the cell appears hung and is not interruptible.

## 🔧 Fix

Add `fabricates_attributes()` (in `_sql/engines/types.py`): probe one attribute that no real connection defines; if the object returns something for it, its duck-typing answers are meaningless and `is_compatible` rejects it. Applied to both `AdbcDBAPIEngine` and `DBAPIEngine` (the two engines relying on pure `getattr` duck-typing — the others use isinstance checks).

Deliberately **not** `inspect.getattr_static`: genuine connection-pool proxies legitimately forward `__getattr__` to an underlying connection and would regress; they pass the probe (the probe attribute doesn't exist on the real connection either).

This generalizes the existing ibis `Deferred` special-case (#7791) — any permissive-`__getattr__` object (lazy proxies, mocks) is now handled, not just known names.

## 🧪 Verification

- Reproduced the hang in the interactive kernel (`executing_kernel` fixture + `from ignite.metrics import Loss; metric = Loss(nn.MSELoss())`): watchdog fired at 30s pre-fix; post-fix the same cell completes in ~4s with `metric` defined.
- New tests: `test_is_compatible_rejects_fabricated_attributes` (dbapi), `test_adbc_is_compatible_rejects_fabricated_attributes` (adbc), `test_fabricated_attributes_not_treated_as_engine` (get_engines).
- `pytest tests/_sql/`: 246 passed. `ruff check` + `ruff format --check` clean on all changed files.

### DIFF
--- a/marimo/_sql/engines/adbc.py
+++ b/marimo/_sql/engines/adbc.py
@@ -17,6 +17,7 @@ from marimo._sql.engines.types import (
     InferenceConfig,
     SQLConnection,
     default_inference_config,
+    fabricates_attributes,
 )
 from marimo._sql.utils import convert_to_output, is_cheap_dialect
 from marimo._types.ids import VariableName
@@ -459,6 +460,11 @@ class AdbcDBAPIEngine(SQLConnection[AdbcDbApiConnection]):
         var_type = type(var)
         var_type_name = f"{var_type.__module__}.{var_type.__qualname__}"
         if var_type_name == "ibis.common.deferred.Deferred":
+            return False
+
+        # Objects with a permissive __getattr__ (e.g. pytorch-ignite metrics)
+        # pass any getattr-based check and hang catalog introspection. #10213
+        if fabricates_attributes(var):
             return False
 
         try:
--- a/marimo/_sql/engines/dbapi.py
+++ b/marimo/_sql/engines/dbapi.py
@@ -5,7 +5,7 @@ from types import ModuleType
 from typing import TYPE_CHECKING, Any, Protocol
 
 from marimo import _loggers
-from marimo._sql.engines.types import QueryEngine
+from marimo._sql.engines.types import QueryEngine, fabricates_attributes
 from marimo._sql.utils import convert_to_output
 
 LOGGER = _loggers.marimo_logger()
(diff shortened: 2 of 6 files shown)


Jayashanker-Padishala:

Light2Dark:
Thanks @Jayashanker-Padishala! I made a similar fix here with some additions https://github.com/marimo-team/marimo/pull/10301. But thank you for checking, appreciate it.

PR #10298, opened 2026-07-23
fix(islands): retain static output through initial hydration

peter-gy:
Fixes a bug spotted bug in the wild at https://koaning.io/posts/widget-dag/ through mdx-marimo.

The page initially renders exported static output while Pyodide boots. That output contains anywidget model IDs from the export. Plugin registration was trying to resolve those IDs against the live model registry, causing two `Model not found` errors after the lookup timed out.

before-model-registry-race

The intended lifecycle we'd want is:

1. Show the exported static output immediately.
2. Start each island session and run its cells.
3. Wait for every initial session to report `completed-run`.
4. Replace the static output with live output and create model-backed views from the live session.

This PR implements that lifecycle. The bridge records each pending session generation before materializing the island elements, and the worker includes that generation with every kernel message. Static output stays mounted and anywidget view creation **waits until the matching initial runs finish**. Session generations also keep late messages scoped to the run that produced them.

This way, the static preview is retained with no errors, as long as startup is not completed:

after-static-preview-retained

And once the startup is complete, we have the error-free, interactive live outputs:

after-runtime-interactive

### DIFF
--- a/frontend/src/core/islands/__tests__/bootstrap.test.ts
+++ b/frontend/src/core/islands/__tests__/bootstrap.test.ts
@@ -1,0 +1,61 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+
+import { createStore } from "jotai";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { RuntimeState } from "@/core/kernel/RuntimeState";
+import { initializeIslands } from "../bootstrap";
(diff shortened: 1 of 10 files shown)


mscolnick (approved):

mscolnick:
@cubic-dev-ai review

PR #10323, opened 2026-07-24
fix: strip display.custom_css from notebook script metadata

mscolnick:
Notebook-embedded PEP 723 metadata could set display.custom_css to
arbitrary file paths, which were read and inlined into the served HTML
before cell execution. Add custom_css to the sanitize denylist alongside
runtime.auto_instantiate and experimental.isolate_apps.


kirangadhave (approved):
🚀

PR #10349, opened 2026-07-27
feat: preserve export filenames and media types

peter-gy:
Export request clients now return an `ExportedFile` containing the exported contents, backend-selected filename, and media type. HTTP clients read the metadata from response headers, while Pyodide returns the same shape directly.

Browser downloads consume this shared response for HTML, Markdown, IPYNB, and PDF exports. This gives us a response contract the upcoming export dialog UI can use.


peter-gy:
@cubic-dev-ai review please

mscolnick (approved):
really nice cleanup!

Issue #10222, opened 2026-07-18
Jump to definition and search in column mode do not update horizontal scroll


NukeMike:
### Describe the bug

When editing a notebook with multiple columns the jump to defition with cntrl/cmd + click is not moving the horizontal scroll, so if the cell is not visible in the current view it seem like the jump has failed.

**Jump between columns bug**
<img alt="Jump To definition between columns bug" src="https://github.com/user-attachments/assets/cb4f2eaa-7a9f-4329-990c-a1ed2aa6c837" />

Also the find/replace in notebook seem to have the same limitation.

**Search between columns bug**
<img alt="Search between columns bug" src="https://github.com/user-attachments/assets/5623e3f6-4100-4cec-be5b-c7ed7c54a9b2" />

**Search stuck in cell bug**

<img alt="Search stuck in cell bug" src="https://github.com/user-attachments/assets/7b107367-2f16-47b4-986d-fd1238b87f08" />

The actual behaviour of searching only into a cell would be a nice option to toggle in the search ui box, in this case the cursor was just stuck into what it found in the secon cell with the search term.

In the column zoomed view i had to select again the 'right_bot' variable and press cntrl+F otherwise it would have just started from the last search (already stuck into the cell).

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

Occurs both locally and in molab
```
{
  "marimo": "0.23.14",
  "editable": false,
  "location": "/usr/local/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.8.12-680-6063-coreweave-amd64-f81899c8",
  "Processor": "",
  "Python Version": "3.13.11",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "v22.23.1",
    "uv": "0.11.28 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.20.0",
    "markdown": "3.10.2",
    "narwhals": "2.23.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pymdown-extensions": "10.21.3",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.0",
    "typing-extensions": "4.16.0",
    "uvicorn": "0.51.0",
    "websockets": "16.1"
  },
[...]

PR #10354, opened 2026-07-27
fix(columns): scroll horizontally when navigating to an off-screen column

Dhruv-kys:

## Problem

Fixes the scrolling bug in #10222. In columns mode the notebook scrolls
horizontally, but navigating to a target in another column left the horizontal
scroll offset where it was, so the navigation looked like it had silently
failed. This affected both paths the reporter identified:

- **Jump to definition** (Cmd/Ctrl+click) onto a definition in another column
- **Find/replace** stepping onto a match in another column

## Demo

https://github.com/user-attachments/assets/b09ae761-cbb8-4221-a355-180ea594a282

## Root cause

`EditorView.scrollIntoView` walks up the ancestors treating anything whose
`scrollWidth` exceeds its `clientWidth` as a scroll container. In columns mode
the `overflow: visible` wrappers around a column satisfy that test while
ignoring `scrollLeft`, so the target rect is clamped against each of them
before the walk ever reaches the app's real scroll container.

The fix scrolls the owning `.marimo-cell` with the browser's own
`scrollIntoView`, which only considers genuine scrolling boxes. The call is
deferred one frame — CodeMirror settles its own scroll in a measure cycle a
frame later and would otherwise clamp it straight back. (An earlier version of
this patch without the deferral looked correct and did not work.)

## This also covers the "search stuck in cell" report

The issue body reports search appearing to get stuck on a match inside one
cell. That is the same bug, not a separate one. Stepping find-next through a
scratch notebook with three matches in column 0 and two in column 3, and
logging which cell held the selection after each click (scratch notebook not
included in the diff):

| | while the selection is in column 0 | once the selection reaches column 3 |
|---|---|---|
| before | `scrollLeft` 0 | `scrollLeft` **5** |
| after | `scrollLeft` 2 | `scrollLeft` **3370 → 3658** |

The selection was advancing correctly all along; only the viewport failed to
follow it out of the column, so the visible highlight never changed.

## Not addressed

The issue also requests an enhancement — a "search only in this cell" toggle in
the search UI. That is new UI surface plus persisted state and seemed worth a
separate discussion, so it is left out here. The reporter asked in a comment
whether to split that out into its own issue and did not get an answer.

For that reason this PR says `Related to #10222` rather than `Fixes`, so merging
does not auto-close the issue while that question is open. Happy to switch it to
`Fixes` if you would rather track the enhancement elsewhere.

## Testing

- New e2e coverage in `frontend/e2e-tests/columns-scroll.spec.ts` for both the
  jump-to-definition and find-next paths, against a new 4-column notebook.
- The e2e assertions check the target cell is **in the viewport**, not that
  `scrollLeft` merely moved: unfixed, the app still nudges ~5px, which a
  `scrollLeft > 0` check accepts. Both tests were confirmed to fail without the
  production change and pass with it.
- Unit regression tests on both the go-to-definition and find-replace sides.
- `pnpm test src/core/codemirror` passes (632 passed, 1 skipped).
- `make check` passes, except its final `pixi lock` step, which fails locally
  because I don't have `pixi` installed. The diff touches no Python files.

Note for reviewers running the e2e locally: Playwright serves the prebuilt
bundle from `marimo/_static`, so `make fe` is required for frontend changes to
be picked up.

### DIFF
--- a/frontend/e2e-tests/py/columns.py
+++ b/frontend/e2e-tests/py/columns.py
@@ -9,13 +9,14 @@ app = marimo.App(width="columns")
 
 
 @app.cell(column=0)
-def __(far_away_variable):
+def __(deep_and_far_variable, far_away_variable):
     # Two usages, so a find/replace walk has to step through both before it
     # leaves this cell for the definition in column 3 -- the scenario the
     # reporter perceived as search being "stuck" inside a cell.
     usage_of_far_away = far_away_variable + 1
     another_usage = far_away_variable * 2
-    return (another_usage, usage_of_far_away)
+    usage_of_deep_and_far = deep_and_far_variable + 1
+    return (another_usage, usage_of_deep_and_far, usage_of_far_away)
 
 
 @app.cell(column=1)
@@ -34,6 +35,28 @@ def __():
 def __():
     far_away_variable = 42
     return (far_away_variable,)
+
+
+@app.cell(column=4)
+def __():
+    import marimo as mo
+
+    return (mo,)
+
+
+@app.cell(column=4)
+def __(mo):
+    # Tall enough to push the cell below into needing a vertical scroll too
+    # (viewport height is 720px), unlike far_away_variable, which only ever
+    # needs a horizontal scroll.
+    mo.ui.text_area(value="filler", rows=80, full_width=True)
+    return
+
+
+@app.cell(column=4)
+def __():
+    deep_and_far_variable = 99
+    return (deep_and_far_variable,)
 
 
 if __name__ == "__main__":
--- a/frontend/e2e-tests/py/columns.py
+++ b/frontend/e2e-tests/py/columns.py
@@ -1,0 +1,40 @@
+# /// script
+# [tool.marimo.runtime]
+# auto_instantiate = true
+# ///
+import marimo
+
+__generated_with = "0.0.5"
+app = marimo.App(width="columns")
+
+
+@app.cell(column=0)
+def __(far_away_variable):
+    # Two usages, so a find/replace walk has to step through both before it
+    # leaves this cell for the definition in column 3 -- the scenario the
+    # reporter perceived as search being "stuck" inside a cell.
+    usage_of_far_away = far_away_variable + 1
+    another_usage = far_away_variable * 2
+    return (another_usage, usage_of_far_away)
+
+
+@app.cell(column=1)
+def __():
+    filler_one = "filler"
+    return (filler_one,)
+
+
+@app.cell(column=2)
+def __():
+    filler_two = "filler"
+    return (filler_two,)
+
+
+@app.cell(column=3)
+def __():
+    far_away_variable = 42
+    return (far_away_variable,)
+
+
+if __name__ == "__main__":
+    app.run()
--- a/frontend/e2e-tests/py/columns.py
+++ b/frontend/e2e-tests/py/columns.py
@@ -10,9 +10,8 @@ app = marimo.App(width="columns")
 
 @app.cell(column=0)
 def __(deep_and_far_variable, far_away_variable):
-    # Two usages, so a find/replace walk has to step through both before it
-    # leaves this cell for the definition in column 3 -- the scenario the
-    # reporter perceived as search being "stuck" inside a cell.
+    # Two matches in this cell, so find-next must step through both before
+    # leaving the column.
     usage_of_far_away = far_away_variable + 1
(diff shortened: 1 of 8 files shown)


Dhruv-kys:

Light2Dark (dismissed):
Thanks, I did a pass to cleanup and harden the tests.

kirangadhave:
@Dhruv-kys Please add a video to the PR description demonstrating the fix.

frontend/src/core/codemirror/utils.ts
```diff
@@ -52,6 +52,34 @@ export function focusInputAndMoveToEnd(
...
+  if (!cell) {
+    return;
+  }
+
+  requestAnimationFrame(() => {
+    cell.scrollIntoView({
+      behavior: "smooth",
+      block: "nearest",
+      inline: "nearest",
+    });
```
kirangadhave, line 79: `smooth` adds animation which is not a great experience when the entire screen moves. Let's set to `instant`

```suggestion
  requestAnimationFrame(() => {
    cell.scrollIntoView({
      behavior: "instant",
      block: "nearest",
      inline: "nearest",
    });
```

frontend/src/core/codemirror/find-replace/__tests__/navigate.test.ts
```diff
@@ -234,6 +242,20 @@ describe("navigate", () => {
...
+    it("should scroll the cell owning the match into view", async () => {
+      const cell = view1.dom.closest(".marimo-cell");
+      invariant(cell, "mock view should be inside a cell");
+
+      expect(findNext()).toBeTruthy();
+
+      // The scroll is deferred a frame.
+      await new Promise((resolve) => requestAnimationFrame(resolve));
+
+      expect(cell.scrollIntoView).toHaveBeenCalledWith(
```
kirangadhave: we should test for all three options we pass: `behavior`, `block` and `inline`

frontend/src/core/codemirror/go-to-definition/__tests__/commands.test.ts
```diff
@@ -383,6 +383,26 @@ print('myVar')`);
...
+    document.body.append(cell);
+
+    view = createEditor("myVar = 10\nprint(myVar)", cell);
+    expect(goToVariableDefinition(view, "myVar")).toBe(true);
+    // One frame for the jump, one for the scroll it defers.
+    await tick();
+    await tick();
+
+    expect(cell.scrollIntoView).toHaveBeenCalledWith(
+      expect.objectContaining({ inline: "nearest" }),
```
kirangadhave: we should test for all three options we pass: behavior, block and inline

frontend/e2e-tests/py/columns.py
```diff
@@ -0,0 +1,63 @@
...
+
+__generated_with = "0.0.5"
+app = marimo.App(width="columns")
+
+
+@app.cell(column=0)
+def __(deep_and_far_variable, far_away_variable):
+    # Two usages, so a find/replace walk has to step through both before it
+    # leaves this cell for the definition in column 3 -- the scenario the
+    # reporter perceived as search being "stuck" inside a cell.
```
kirangadhave: these comments make no sense without context of the issue thread. Comments in the source code should be readable without additional context in most cases.

Something like this is enough:
```py
# two matches in this cell, so find-next must step through both before leaving the column.
```

frontend/src/core/codemirror/go-to-definition/commands.ts
```diff
@@ -44,6 +45,8 @@ function goToPosition(view: EditorView, from: number): void {
         y: "center",
       }),
     });
+
+    scrollOwnerCell(view);
```
kirangadhave: scrollOwnerCell implementation also wraps its behavior in a rAF, can we pull this out of the rAF here?

kirangadhave (changes requested):

Dhruv-kys:
Addressed all the feedback, video is in the description now.

On the rAF one — pulling it out works fine. rAF callbacks run in registration order so the dispatch still lands first. Ran the e2e 12x to confirm.

@Light2Dark your approval got auto-dismissed by the push, only change since is the review feedback.

kirangadhave (approved):

kirangadhave:
@Light2Dark LGTM, if you're okay with this, we can merge

Light2Dark:
Thanks @kirangadhave, merging it in. If we find the experience not as smooth, we can tweak it later too.

PR #10355, opened 2026-07-28
fix(lsp): upgrade codemirror-languageserver to v2

mscolnick:
Migrate to the v2 `Transport` API and drop `@open-rpc/client-js` along with
its transitive polyfills.

- Rebuild `ReconnectingWebSocketTransport` on the v2 transport interface, and
  collapse Copilot's `LazyWebsocketTransport` onto it
- Hold outbound messages until initialization, configuration, and document
  restoration finish after a reconnect. Messages queued *during* restoration
  are replayed rather than filtered: they carry edits the restore snapshot
  never saw, and the notebook snapshotter will not re-emit them
- Give each reconnect attempt a generation, so closing a socket abandons the
  attempt bound to it rather than parking on an initialize response that will
  never arrive. The abandoned attempt's backlog goes straight back to the
  queue instead of stranding until it settles
- Adopt didOpen reference counting, didClose, save notifications, and
  `hasCapability()` routing in the notebook and federated adapters
- Map diagnostic relatedInformation out of merged-document coordinates so
  same-cell "declared here" links resolve, and tolerate entries with a missing
  or malformed location so one bad entry can't take down processing
- Resolve cell document URIs against the open notebook instead of trusting
  `CellDocumentUri.is`, whose `file:///` prefix matched every absolute file
  URI and so disabled go-to-definition into real files
- Enable client-side completion filtering
- Narrow `ILanguageServerClient` to an explicit subset so upstream additions
  don't silently become required adapter methods
- Normalize (rather than validate) Copilot status notifications, whose vendor
  payload routinely omits fields and ships unknown `kind` values, and clear
  the loading indicator in a `finally` so a timed-out request can't pin it
- Surface `window/logMessage` in the Logs panel. `CellLog` carries a
  `LogSource` union now, since a server's startup failures belong to no cell
- Stack the diagnostic tooltip's message, docs link, and related information
  instead of running them together on a single line
- Restore the newline between fenced-code segments in the markdown math
  normalizer, which was gluing a closing fence to the next line and swallowing
  the rest of the docstring into the code block, and stop reading `\[...\]`
  mid-word as display math so escaped brackets in type annotations
  (`Optional\[Callable\]`) render literally

Screenshot 2026-07-28 at 10 21 46 AM
Screenshot 2026-07-28 at 10 36 07 AM


mscolnick:
@cubic-dev-ai pls review

Light2Dark (approved):

PR #10362, opened 2026-07-28
feature: Infer file browser initial path from initial value

### LINKED ISSUE #10340: file_browser automatically set initial_path when given value
### Description

Now that file_browser supports default file selection (https://github.com/marimo-team/marimo/issues/7468),
when that feature is used, the browser may start at a different path than the one where the selection is, which may seem surprising.

Suggestion: when no initial_path is given, set it to the parent folder for the default selection.

If the default selection contains several files in different folders (is that supported?), default the initial_path to the common ancestor.

- [x] Yes

### Alternatives

_No response_

### Additional context

_No response_

yairchu:
**This pull request includes contributions from a coding agent.**

## 📝 Summary

Closes marimo-team/marimo#10340

## 📋 Pre-Review Checklist

- [X] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](<https://marimo.io/discord?ref=pr>), or the community [discussions](<https://github.com/marimo-team/marimo/discussions>) (Please provide a link if applicable).

## ✅ Merge Checklist

- [X] I have read the [contributor guidelines](<https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md>).
- [X] Documentation has been updated where applicable, including docstrings for API changes.

### DIFF
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -304,6 +304,10 @@ class file_browser(
             selected_path = self._create_path(
                 normalize_path(self._create_path(selected_value))
             )
+            if not selected_path.exists():
+                raise ValueError(
+                    f"Default value {selected_path} does not exist."
+                )
             is_directory = selected_path.is_dir()
             kind = "directory" if is_directory else "file"
             if kind not in self._selection_mode:
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -66,6 +66,28 @@ def _normalize_selection_mode(
     )
 
 
+def _normalize_values(
+    value: str | Path | Sequence[str | Path] | None,
+    *,
+    multiple: bool,
+) -> Sequence[str | Path]:
+    """Normalize the file browser's default values to a sequence."""
+    if value is None:
+        values: Sequence[str | Path] = ()
+    elif isinstance(value, Sequence) and not isinstance(value, str):
+        values = value
+    else:
+        values = (value,)
+
+    if not multiple and len(values) > 1:
+        raise ValueError(
+            "File browser with multiple=False cannot have more than one "
+            "default value."
+        )
+
+    return values
+
+
 def _common_parent(paths: Sequence[Path]) -> Path:
     """Return the nearest directory containing all paths.
 
@@ -235,18 +257,7 @@ class file_browser(
     ) -> None:
         self._selection_mode = _normalize_selection_mode(selection_mode)
 
-        if value is None:
-            values: Sequence[str | Path] = ()
-        elif isinstance(value, Sequence) and not isinstance(value, str):
-            values = value
-        else:
-            values = (value,)
-
-        if not multiple and len(values) > 1:
-            raise ValueError(
-                "File browser with multiple=False cannot have more than one "
-                "default value."
-            )
+        values = _normalize_values(value, multiple=multiple)
 
         # Save the Path class and client used to construct paths
         path_source = (
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -103,6 +103,15 @@ def _common_parent(paths: Sequence[Path]) -> Path:
     return candidate
 
 
+def _is_path_within(path: Path, parent: Path) -> bool:
+    """Return whether `path` resolves within `parent`."""
+    try:
+        path.resolve(strict=True).relative_to(parent.resolve())
+    except (ValueError, RuntimeError, OSError):
+        return False
+    return True
+
+
 @dataclass
 class ListDirectoryArgs:
     path: str
@@ -320,7 +329,7 @@ class file_browser(
 
         if restrict_navigation:
             for selected_path in selected_paths:
-                if not selected_path.parent.is_relative_to(self._initial_path):
+                if not _is_path_within(selected_path, self._initial_path):
                     raise ValueError(
                         f"Default value {selected_path} is outside the "
                         f"restricted initial path {self._initial_path}."
@@ -480,12 +489,7 @@ class file_browser(
         path = self._create_path(args.path)
 
         if self._restrict_navigation:
-            try:
-                path.resolve(strict=True).relative_to(
-                    self._initial_path.resolve()
-                )
-            # NB. RuntimeError vs OSError depends on the version of python.
-            except (ValueError, RuntimeError, OSError):
+            if not _is_path_within(path, self._initial_path):
                 raise RuntimeError(
                     "Navigation is restricted; navigating outside the initial path is not allowed."
                 ) from None
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -219,6 +219,19 @@ class file_browser(
     ) -> None:
         self._selection_mode = _normalize_selection_mode(selection_mode)
 
+        if value is None:
+            values: Sequence[str | Path] = ()
+        elif isinstance(value, Sequence) and not isinstance(value, str):
+            values = value
+        else:
+            values = (value,)
+
+        if not multiple and len(values) > 1:
+            raise ValueError(
+                "File browser with multiple=False cannot have more than one "
+                "default value."
+            )
+
         # Save the Path class of the initial path
         self._path_cls: type[Path]
         if isinstance(initial_path, str):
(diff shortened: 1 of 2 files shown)


marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -219,15 +235,35 @@ def __init__(
...
             self._path_cls = initial_path.__class__
 
         # Make a Path object
-        if not initial_path:
+        if not initial_path and values:
+            self._initial_path = None  # Required for self._create_path
+            initial_path = _common_parent(
+                [normalize_path(self._create_path(p)) for p in values]
+            )
+        elif not initial_path:
```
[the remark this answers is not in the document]
yairchu: Made an edit inspired by this suggestion, please take another look.

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -219,15 +235,39 @@ def __init__(
...
+            values = value
+        else:
+            values = (value,)
+
+        if not multiple and len(values) > 1:
+            raise ValueError(
+                "File browser with multiple=False cannot have more than one "
+                "default value."
+            )
+
```
kirangadhave: might be worth extracting this as a `_normalize_values` helper and unit testing it

marimo/_plugins/ui/_impl/file_browser.py
kirangadhave, line 329: we are deriving `initial_path` from values. Initially when `value` was supplied with a non-existent parent directory, we didn't have failure. Now we are failing (which is correct behavior), but we fail due to the `initial_path.is_dir()` check and not the actual check on `value`. The error message we see is incorrect:

Image

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -219,15 +235,39 @@ def __init__(
...
+        elif isinstance(initial_path, str):
             self._path_cls = Path
         else:
             self._path_cls = initial_path.__class__
 
         # Make a Path object
-        if not initial_path:
+        if not initial_path and values:
+            # Initialize self._initial_path temporarily for self._create_path
+            self._initial_path = (
```
kirangadhave: Setting intial_path temporarily just so `self._create_path` can read the client off it is hacky.
Also we are passing `None` to `_create_path`, the type suggests `_create_path` only takes `Path` (there is `# type: ignore` which is masking this)

We should update `_create_path` to take in explicit `path_cls` and `client` parameters.

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -66,6 +66,21 @@ def _normalize_selection_mode(
     )
 
 
+def _common_parent(paths: Sequence[Path]) -> Path:
```
kirangadhave, line 91: When the values share no common ancestor this returns the `root`. The root passes the `is_dir` check and is accepted silently. 

### Cloud paths
```py
fb = mo.ui.file_browser(
    value=[
        LocalS3Path("s3://bucket1/dir/f.txt", client=c),
        LocalS3Path("s3://bucket2/dir/f.txt", client=c),
    ],
    restrict_navigation=True,
)

fb._initial_path                      # S3Path('s3://')
```

Then one can do:
```py
fb._list_directory(ListDirectoryArgs(path="s3://unrelated-secrets/dir"))  # ALLOWED
```
`restrict_navigation` is enforced relative to `self._initial_path`, so inferring `s3://` makes all buckets the client can access reachable. On `main` the boundary was `Path.cwd()`. This impl is loosening the behavior.

Same for windows path:
`C:/a.txt` and `D:/b.txt` returns `C:/` giving access to all `C` drive even if `restrict_navigation=True`

We should fallback to `Path.cwd()` when no common ancestor found.

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -219,15 +235,39 @@ def __init__(
...
+            raise ValueError(
+                "File browser with multiple=False cannot have more than one "
+                "default value."
+            )
+
         # Save the Path class of the initial path
         self._path_cls: type[Path]
-        if isinstance(initial_path, str):
+        if not initial_path and values and not isinstance(values[0], str):
+            self._path_cls = values[0].__class__
```
kirangadhave: no test for this path

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -219,15 +235,39 @@ def __init__(
...
             self._path_cls = Path
         else:
             self._path_cls = initial_path.__class__
 
         # Make a Path object
-        if not initial_path:
+        if not initial_path and values:
+            # Initialize self._initial_path temporarily for self._create_path
+            self._initial_path = (
+                None if isinstance(values[0], str) else values[0].parent
```
kirangadhave: No test for `None if isinstance(values[0], str)` path

tests/_plugins/ui/_impl/test_file_browser.py
kirangadhave, line 1: no test for cloud paths, but that's not this PR's concern. If you're willing to add it here, appreciate the help.

marimo/_plugins/ui/_impl/file_browser.py
kirangadhave, line 325: When passing initial selection of a directory and contents of directory (say `b/` and `b/foo.txt`), setting `initial_path` to the directory `b` and setting `restrict_navigation=True` we end up in a state where `b` is selected, but never rendered, so it can never be unselected.

Image

---
More generally, setting `value` outside the root with `restrict_navigation=True` we have selections which are never rendered in the browser.

Image

---
`_common_parent` always returns a strict ancestor of every value, so every default selection is displayable. Nothing enforces that when `initial_path` is passed explicitly, and with `restrict_navigation=True` the selection then becomes unreachable.

The behavior started in the previous PR, it's cheap to close here with something like:
```py
if restrict_navigation and not selected_path.parent.is_relative_to(self._initial_path):
    raise ValueError(...)
```

kirangadhave (changes requested):

marimo/_plugins/ui/_impl/file_browser.py
yairchu: Done

marimo/_plugins/ui/_impl/file_browser.py
yairchu: Now raising ValueError for non-existent default values

marimo/_plugins/ui/_impl/file_browser.py
yairchu: Now determining `self._path_cls` after normalizing values instead of this hack.

marimo/_plugins/ui/_impl/file_browser.py
yairchu: Probably safer to not infer the path at all when restricting navigation. Implemented this.

marimo/_plugins/ui/_impl/file_browser.py
yairchu: Now verifying that when `restrict_navigation` is enabled all of the initial values are within it.

tests/_plugins/ui/_impl/test_file_browser.py
yairchu: Having never really worked with cloud paths, I prefer not adding it to this PR

marimo/_plugins/ui/_impl/file_browser.py
yairchu: `test_file_browser_infers_custom_path_class_and_client` now tests the equivalent of this in the updated code

marimo/_plugins/ui/_impl/file_browser.py
yairchu: The equivalent for this path also now tested in `test_file_browser_infers_initial_path_from_string_value`

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -219,16 +257,57 @@ def __init__(
     ) -> None:
         self._selection_mode = _normalize_selection_mode(selection_mode)
 
-        # Save the Path class of the initial path
+        values = _normalize_values(value, multiple=multiple)
+
+        # Save the Path class and client used to construct paths
+        path_source = (
```
kirangadhave: `path_source` falls back to `initial_path` when `restrict_navigation` is set. When `initial_path` is not set, the value becomes `""` and it takes the `isinstance(path_source, str)` branch. So `path_class=Path` and cloud paths like `s3://bucket/f.txt` are coerced to a local path which then throw with `Default value does not exist` error (which is misleading here).

We are merging two concerns here:
- Deriving _path_cls and client from `values[0]` is always safe
- Only inferring `initial_path` needs the `restrict_navigation` check

the `not restrict_navigation` check can go around line 305, something like:

```py
if not initial_path:
    if selected_paths and not restrict_navigation:
        initial_path = _common_parent(selected_paths) or selected_paths[0].parent
    else:
        initial_path = Path.cwd()
```

---

I think if `test_file_browser_infers_custom_path_class_and_client` is parameterized over both `True/False` for `restrict_navigation`, we can catch this case.

marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -243,6 +322,14 @@ def __init__(
                 f"Initial path {initial_path} is not a directory."
             )
 
+        if restrict_navigation:
+            for selected_path in selected_paths:
+                if not selected_path.parent.is_relative_to(self._initial_path):
```
kirangadhave: `if not selected_path.parent.is_relative_to(self._initial_path):`

In this check `selected_path` and `self._initial_path` are dervied in two different ways.
- `selected_path` comes from `normalize_path` which does not resolve `symlinks`
- `self._initial_path` when inferred (so not provided initially), it comes from `Path.cwd()` which resolves symlinks.

kirangadhave (changes requested):
Thanks @yairchu for addressing the previous comments. I have two more blockers. Do you want to take another shot at fixing this? Else I'm happy to take over your branch and clean it up.

marimo/_plugins/ui/_impl/file_browser.py
yairchu: > Deriving `_path_cls` and client from `values[0]` is always safe

Done. Switched to that when there are values.

> I think if `test_file_browser_infers_custom_path_class_and_client` is parameterized over both `True/False` for `restrict_navigation`, we can catch this case.

Wouldn't that require supporting not providing an initial path when restricting navigation?

marimo/_plugins/ui/_impl/file_browser.py
yairchu: Done, now using the same check as `_list_directory` uses

kirangadhave (approved):

PR #10366, opened 2026-07-28
fix(tests): stop DuckDB ATTACH state leak in SQL cleanup test

Light2Dark:
## 📝 Summary

Fixes marimo-team/ci-agent#79

- `test_sql_table_on_attached_catalog_is_not_dropped` (added in #10357) calls `duckdb.execute("ATTACH ':memory:' AS other_db")` and creates a schema/table directly on the process-wide default DuckDB connection, and never tears them down.
- This leaks `other_db`/`holdings` state into any later test that relies on that same default connection via `DuckDBEngine(None)`, causing flaky failures in `tests/_sql/test_duckdb.py` and `tests/_sql/test_get_engines.py` whenever pytest-xdist schedules them on the same worker.
- Wrapped the attach/create logic in `try`/`finally` so the attached catalog and leaked schema are detached/dropped after the test runs, regardless of outcome.
- Reproduced the exact reported failures locally by running the leaking test alongside the affected `_sql` tests in one process, confirmed they pass after the fix, and confirmed they fail with the exact same diffs when reverting it.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/_runtime/test_runtime.py
+++ b/tests/_runtime/test_runtime.py
@@ -3586,64 +3586,76 @@ class TestSQL:
     async def test_sql_table_on_attached_catalog_is_not_dropped(
         self, k: Kernel
     ) -> None:
-        # `CREATE TABLE <catalog>.<table>` is shorthand for
-        # `<catalog>.main.<table>`, which is ambiguous with `<schema>.<table>`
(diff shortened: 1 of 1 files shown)


kirangadhave (approved):
:rocket:

PR #10368, opened 2026-07-28
feat(export): support Markdown flavors in session export

peter-gy:
## Summary

Adds `flavor` to the Markdown session export request and wires it through the HTTP and Pyodide paths.

The existing Markdown exporter remains responsible for resolving the flavor, rendering the correct cell syntax, and choosing the download filename. Requests can select `pymdown`, `qmd`, `mystmd`, or `mdx`. When `flavor` is not passed, we still use filename-based inference.

Added a dedicated request type for automatic Markdown export endpoint so we pass `flavor` only on user-initiated exports.


peter-gy:
@cubic-dev-ai review please

mscolnick (approved):

PR #10369, opened 2026-07-28
Treat downstream `mo.stop` as scratchpad success

### LINKED ISSUE #62: execute-code.sh: exit 1 + bare 'null' when a durable edit triggers a downstream mo.stop
## Problem

When a `cm` block adds or edits cells whose reactive descendants include a `mo.stop` guard (a very common pattern, e.g. button- or dropdown-gated cells), `scripts/execute-code.sh` exits non-zero and prints a bare `null`, even though the structural edit itself succeeded. This reads as a failure and repeatedly pushed me (an agent driving the skill) into extra verification passes to confirm nothing had actually broken.

## Current behavior

A benign `null` final output and/or a stopped descendant cell is mapped to `exit 1` with no distinguishing message, indistinguishable from a real error.

## Considerations

- Separate "the scratchpad code itself ran fine" from "a descendant cell stopped or errored".
- Don't treat a `null` final output as failure.
- Open question: how best to represent *partial* success (edit applied, but a descendant hit `mo.stop`) in a shell-friendly exit code + message. A distinct exit code or a one-line note ("edit applied; descendant cell <id> stopped via mo.stop") would remove the ambiguity.

## Example

During a real pairing session, several `create_cell` / `edit_cell` blocks succeeded but returned `exit 1` + `null`, because the notebook's button-gated push cell hit `mo.stop` on reactive re-run. The edits were fine; only the exit signal was misleading.

manzt:
This isn't an "error" from the agent's perspective. 

Fixes marimo-team/marimo-pair#62

### DIFF
--- a/marimo/_server/scratchpad.py
+++ b/marimo/_server/scratchpad.py
@@ -14,6 +14,7 @@ from marimo._code_mode.screenshot_meta import (
     SCREENSHOT_SERVER_URL_KEY,
 )
 from marimo._messaging.cell_output import CellChannel
+from marimo._messaging.errors import MarimoAncestorStoppedError
 from marimo._messaging.notebook.outputs import CellOutputs
 from marimo._messaging.notification import (
     CellNotification,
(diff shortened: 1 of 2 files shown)


kirangadhave (approved):
:rocket:

PR #10378, opened 2026-07-29
feat(export): expose server dependency availability

peter-gy:
The upcoming export dialog needs to show format requirements before export. This adds a server-backed availability endpoint for HTML, Markdown, IPYNB, and PDF, including actionable missing package specs.

Export requirements live in one declarative registry shared by the availability endpoint, IPYNB auto-export, and the IPYNB/PDF exporters. A generic grouped dependency requirement lets PDF report `nbconvert[webpdf]` while checking `nbformat`, `nbconvert`, and `playwright`.

The frontend routes this request directly to the server, so the dialog can query availability before a kernel session starts.

### DIFF
--- a/marimo/_export/dependencies.py
+++ b/marimo/_export/dependencies.py
@@ -1,44 +1,49 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-from typing import get_args
-
 from marimo._dependencies.dependencies import (
     DependencyLike,
     DependencyManager,
     DependencyRequirement,
 )
 from marimo._schemas.export_options import ServerExportFormat
+from marimo._utils.assert_never import assert_never
 
 _IPYNB_DEPENDENCIES = (DependencyManager.nbformat,)
-
-EXPORT_DEPENDENCY_REQUIREMENTS: dict[
-    ServerExportFormat, tuple[DependencyLike, ...]
-] = {
-    "html": (),
-    "markdown": (),
-    "ipynb": _IPYNB_DEPENDENCIES,
-    "pdf": (
-        DependencyRequirement(
-            package="nbconvert[webpdf]",
-            dependencies=(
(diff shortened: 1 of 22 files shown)


marimo/_export/dependencies.py
```diff
@@ -0,0 +1,54 @@
...
+    "html": (),
+    "markdown": (),
+    "ipynb": _IPYNB_DEPENDENCIES,
+    "pdf": (
+        DependencyRequirement(
+            package="nbconvert[webpdf]",
+            dependencies=(
+                *_IPYNB_DEPENDENCIES,
+                DependencyManager.nbconvert,
+                DependencyManager.playwright,
```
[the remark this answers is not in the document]
peter-gy: we can consider adding something like `missingSetupCommands` too but for now this is an edge case that can be also handled by `allow_chromium_download=True`.

marimo/_server/api/endpoints/export.py
```diff
@@ -71,6 +76,35 @@
 auto_exporter = AutoExporter()
 
 
+@router.get("/availability")
+@requires("edit")
```
Light2Dark: I guess heads up that this needs edit mode but export_html needs read mode,
might break once integrated?

Light2Dark (dismissed):
looks good! I have non-blocking edits https://github.com/peter-gy/marimo/pull/1

marimo/_server/api/endpoints/export.py
peter-gy: Good catch, thanks! I scoped this to edit because the consumer will be the editor export dialog, but the endpoint itself is clearly read-only so will flip it to `read`.

Light2Dark (dismissed):

peter-gy:
tagging @mscolnick here for visibility, as this is also related to the export dialog project

Light2Dark (approved):

mscolnick (approved):

PR #10394, opened 2026-07-30
feat(export): support script and markdown export from session

peter-gy:
(1) The existing **Download Python code** action now uses the shared session export contract in native and Pyodide runtimes. It downloads the flattened, topologically ordered `.script.py` produced by the Script exporter, with its filename and media type carried through `ExportedFile`. Script is also included in export availability for the upcoming dialog.

(2) Manual Markdown exports now use the current notebook session, including unsaved edits. Automatic Markdown exports continue to reload the saved notebook before writing their snapshot.

### DIFF
--- a/frontend/src/components/editor/actions/useNotebookActions.tsx
+++ b/frontend/src/components/editor/actions/useNotebookActions.tsx
@@ -82,10 +82,12 @@ import { copyToClipboard } from "@/utils/copy";
 import {
   ADD_PRINTING_CLASS,
   downloadAsPDF,
+  downloadBlob,
   downloadExportedFile,
   downloadHTMLAsImage,
   withLoadingToast,
 } from "@/utils/download";
+import { Filenames } from "@/utils/filenames";
 import { Objects } from "@/utils/objects";
 import type { ProgressState } from "@/utils/progress";
 import { Strings } from "@/utils/strings";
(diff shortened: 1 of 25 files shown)


frontend/src/components/editor/actions/useNotebookActions.tsx
```diff
@@ -291,11 +290,8 @@ export function useNotebookActions() {
           icon: <CodeIcon size={14} strokeWidth={1.5} />,
           label: "Download Python code",
           handle: async () => {
-            const code = await readCode();
-            downloadBlob(
-              new Blob([code.contents], { type: "text/plain" }),
-              Filenames.toPY(document.title),
-            );
+            const exportedFile = await exportAsScript({ download: false });
+            downloadExportedFile(exportedFile);
```
mscolnick, line 307: these are different things. one is a marimo file, and the other is a flat pypercent script. we should revert this change

mscolnick: oh, i see you meant for this change. maybe we can show both? `Download py` and `Download as flat script (pypercent)` (although maybe less wordy)

peter-gy: Yeah, that makes sense, I see now that my first go at it was confusing. pushed changes to keep the existing source download and added a separate pypercent export, labeled 'Download notebook source' and 'Download flat script' to distinguish the two .py files.

peter-gy:
@cubic-dev-ai review please

mscolnick (approved):

PR #10399, opened 2026-07-30
feat(code-mode): add capability discovery

peter-gy:
## Summary

Lets code mode discover capabilities installed in the notebook Python environment and load them on demand. This lets third-party packages extend code-mode-driven agents, such as marimo-pair.

A third-party package such as `marimo-lens` can register a capability entry point in its `pyproject.toml`:

```toml
[project.entry-points."marimo.agent.capability"]
lens = "marimo_lens.agent:capability"
```

Here, `marimo_lens.agent:capability` resolves to an object with string `description` and `instructions` attributes and a callable `bind(context)`. Providers satisfy this contract structurally and do not need to import a marimo base class. Although we can consider providing a public `Protocol` to give users a typed experience.

This allows code mode to use progressive disclosure:

1. `cm.capabilities()` lists installed capability names without importing provider modules.
2. An agent loads a relevant capability with `cm.load_capability("lens")`, which imports, validates, and caches it.

All registered capabilities appear automatically in `help(cm)`.

`capability.bind(ctx)` receives a read-only context exposing kernel globals and notebook cells so 3rd parties don't need to tap into graph internals (as I had to with marimo-lens).

### DIFF
--- a/marimo/_code_mode/_capabilities.py
+++ b/marimo/_code_mode/_capabilities.py
@@ -1,0 +1,189 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
+from dataclasses import dataclass
+from types import MappingProxyType
+from typing import TYPE_CHECKING, Any, Protocol, TypeGuard, TypeVar, overload
+
+from marimo._entrypoints.ids import KnownEntryPoint
+from marimo._entrypoints.registry import get_entry_points
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator, Mapping, Sequence
+
+T_co = TypeVar("T_co", covariant=True)
+
+
+class CellView(Protocol):
+    """Read-only cell data exposed to capabilities."""
+
+    @property
+    def id(self) -> str: ...
+
+    @property
+    def code(self) -> str: ...
+
+    @property
+    def definitions(self) -> frozenset[str]: ...
+
+    @property
+    def references(self) -> frozenset[str]: ...
+
+
+class CellsView(Protocol):
+    """Read-only access to notebook cells."""
+
+    @overload
+    def __getitem__(self, key: int) -> CellView: ...
+
+    @overload
+    def __getitem__(self, key: str) -> CellView: ...
+
+    def __iter__(self) -> Iterator[CellView]: ...
+
+
+class CapabilityContext(Protocol):
+    """Read-only code-mode surface provided to capabilities."""
+
+    @property
+    def globals(self) -> Mapping[str, Any]: ...
+
+    @property
+    def cells(self) -> CellsView: ...
+
+
+class Capability(Protocol[T_co]):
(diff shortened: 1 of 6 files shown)


marimo/_code_mode/_capabilities.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
+from typing import Any
+
+from marimo._entrypoints.ids import KnownEntryPoint
+from marimo._entrypoints.registry import get_entry_points
+from marimo.capabilities import Capability
+
+_ENTRY_POINT_GROUP: KnownEntryPoint = "marimo.capability"
```
peter-gy: @manzt do we want to add a deeper entrypoint like `marimo.agent.capability` instead?

marimo/_code_mode/_capabilities.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class CapabilityContext(Protocol):
+    """Read-only code-mode surface provided to capabilities."""
+
+    @property
+    def globals(self) -> Mapping[str, Any]: ...
+
+    @property
+    def cells(self) -> CellsView: ...
```
manzt: Just to note, and this is ok, but code mode's API is not versioned with marimo releases.

I think we can consider it more stable, but just wanted to flag and I like how this is intentionally minimal.

manzt (commented):
This is great. Sorry for the delay reviewing it. I think we could merge this as is.

One idea is that the main protocol we need to settle on initially may just be discovery and a conventional instruction entry point. The model I have in mind is similar to skills.

```text
skill directory     -> SKILL.md
Python capability   -> help(module)
```

A skill is a directory shaped bundle where `SKILL.md` tells the agent how to use it. A capability could be an importable Python namespace where `help(module)` does the same. Beyond that entry point, either bundle can contain whatever instructions and functionality it needs.

Marimo-pair already asks the agent to call `help(cm)`, so we could advertise the available capabilities there.

```python
help(cm)

# ADDITIONAL CAPABILITIES
# =======================
#
# Import a relevant module and call help on it to learn how to use it.
#
# lens    Work with marimo-lens visualizations    marimo_lens.agent
# ...
#
# For example
#
# import marimo_lens.agent
# help(marimo_lens.agent)
```

We could also decide that some capabilities should be loaded eagerly and put their instructions directly in the prompt.

This would leave runtime patterns like `bind(context)` flexible while we learn what capabilities actually share. I am also happy to start with the protocol proposed here and evolve it within code mode.

peter-gy:
Thanks for the detailed review @manzt! I tend to agree that we could start even with a smaller code mode surface. Reduced the PR for now to include only dynamic `help(cm)` for discovery and we can see if our / community needs evolve.

manzt (approved):
Looks great! Do we want to label this as "preview" tag instead of internal?

PR #10414, opened 2026-07-31
Fix UIElement nested in vstack/hstack losing RPC after #9933

### LINKED ISSUE #10088: UI element nested inside mo.vstack()/composite output stops responding after 0.23.11 (regression from #9933)
### Describe the bug

Since **0.23.11**, a `UIElement` (e.g. `mo.ui.button`) nested *inside a container* (`mo.vstack([...])`, likely `mo.hstack`/other composites too) — and assigned to an underscore-prefixed ("cell-private") local as the cell's last expression — silently stops responding to interaction. Clicking the button:

- fires `POST /api/kernel/set_ui_element_value` (so the frontend believes it's talking to a live element)
- returns 200, no exception, nothing in server logs
- the element's `on_click`/`on_change` never actually runs — no state update, no re-render, no error surfaced anywhere in the UI or console

This bisects cleanly to **#9933** ("delete local variables after cell execution"), released in 0.23.11. That PR's own description says:

> Locals used in the last expression of a cell (such as `_df`, `mo.ui.table(_df)`) will still respond to RPCs, since the runtime independently holds a reference to the last expression.

That guarantee holds when the element itself *is* the last expression, but breaks when the element is nested inside a container object (e.g. a list passed to `mo.vstack`) that is the last expression. It looks like the post-execution cleanup (or the "closed over" static analysis in `visitor.py` that decides what to keep alive) doesn't walk into composite output objects to find nested `UIElement`s, so it drops the only reference keeping the button's callback registration alive, orphaning it even though it's still rendered and interactive in the DOM.

### Environment

Bisected across official PyPI releases via `uvx marimo==<version> run repro.py --headless --no-token`, driven with Playwright (real browser, real click):

| Version | Works? |
|---|---|
| 0.23.4 – 0.23.10 | ✅ |
| **0.23.11** | ❌ (first bad version) |
| 0.23.12, 0.23.13 (latest) | ❌ |

### Minimal repro

```python
import marimo

with app.setup:
    import marimo as mo


@app.cell
def query_state():
    get_last_sql, set_last_sql = mo.state(0)
    return get_last_sql, set_last_sql


@app.cell
def run_control(get_last_sql, set_last_sql):
    _output = mo.vstack(
        [
            mo.md("static"),
            mo.ui.button(
                value=0,
                on_click=lambda count: count + 1,
                on_change=lambda v: set_last_sql(v),
                label="Rerun query",
            ),
        ]
    )
    _output
    return


@app.cell
def show(get_last_sql):
    mo.md(f"last_sql_state=[{get_last_sql()}]")
    return


if __name__ == "__main__":
    app.run()
```

Run with `marimo run repro.py`, click "Rerun query" — `last_sql_state` never changes from `[0]`.

### Isolation notes

- Removing `mo.vstack(...)` (assigning the bare button directly to `_output`) → **works fine** on 0.23.13. So underscore-naming alone isn't sufficient to trigger it.
- Wrapping the same bare-button assignment in an `if/else` branch (still underscore, no `vstack`) → **works fine** on 0.23.13. So branching alone isn't sufficient either.
- It's specifically: element nested inside a container call (`mo.vstack([...])`) + held only by an underscore-prefixed local.
- **Workaround**: assign the `mo.vstack(...)` result to a plain (non-underscore) variable and `return` it from the cell — confirmed this avoids the bug even on 0.23.13, consistent with the deleted-local-variable theory.

### Impact

This is a silent failure — no exception, no console warning, nothing distinguishing it from a user simply not clicking the button. It's a fairly common pattern (conditionally showing a button alongside a status message via `mo.vstack`/`mo.hstack`, held in a `_`-prefixed cell-private var) so likely affects other users hitting an unresponsive UI with no diagnostic signal.

### Will you submit a PR?

- [ ] Yes

TemidayoA:
Closes marimo-team/marimo#10088.

contains_instance() only recursed into raw list/tuple/dict when searching for UIElement instances to keep alive after cell execution. Elements returned by mo.vstack/mo.hstack are \_FlexContainerHtml objects, not raw lists, so nested UIElements were invisible to this check and got garbage collected, silently orphaning their RPC callbacks (on_click/on_change never fire again, no error surfaced).

Extend the retention walk to also descend into any object exposing \_live_children (the private contract \_FlexContainerHtml already uses for its own re-rendering), so nested UIElements are found regardless of container.

Diagnosis and fix approach originally proposed by @David-Wu1119 in the issue thread; implemented here with a regression test.

## 📝 Summary

Closes marimo-team/marimo#10088.

## 📋 Pre-Review Checklist

- [X] For large changes, or changes that affect the public API: this change was discussed in the linked issue (marimo-team/marimo#10088), where @David-Wu1119 diagnosed the root cause and proposed this exact fix, and @akshayka confirmed it as the intended behavior gap.

## ✅ Merge Checklist

- [X] I have read the [contributor guidelines](<https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md>).
- [ ] Documentation has been updated where applicable, including docstrings for API changes.  <br>(N/A — internal retention-check fix, no public API or docstring change.)

### DIFF
--- a/marimo/_runtime/runner/hooks_post_execution.py
+++ b/marimo/_runtime/runner/hooks_post_execution.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import sys
 import traceback as tb
+from typing import Any
 
 from marimo import _loggers
 from marimo._ast.cell import CellImpl
@@ -264,6 +265,18 @@ def broadcast_storage_backends(
         LOGGER.debug("Error getting storage backends", exc_info=True)
 
 
+def _get_live_children(value: Any) -> Any | None:
+    """Descend into marimo Html containers when hunting for UIElements.
+
+    Containers such as mo.hstack/mo.vstack store their children as live
+    references under `_live_children` rather than as a raw list/tuple/
+    dict, so contains_instance() can't see into them on its own. This
+    keeps that marimo-specific knowledge out of the generic flatten
+    utility, passing it in as an extension instead.
+    """
+    return getattr(value, "_live_children", None)
+
+
 @kernel_tracer.start_as_current_span("store_reference_to_output")
 def _store_reference_to_output(
     cell: CellImpl,
@@ -281,9 +294,11 @@ def _store_reference_to_output(
     if isinstance(run_result.output, UIElement):
         cell.set_output(run_result.output)
     elif run_result.output is not None:
-        if contains_instance(run_result.output, UIElement) or callable(
-            getattr(run_result.output, "_repr_mimebundle_", None)
-        ):
+        if contains_instance(
+            run_result.output,
+            UIElement,
+            extra_children=_get_live_children,
+        ) or callable(getattr(run_result.output, "_repr_mimebundle_", None)):
             cell.set_output(run_result.output)
 
 
--- a/marimo/_utils/flatten.py
+++ b/marimo/_utils/flatten.py
@@ -280,7 +280,11 @@ def flatten(
     return flattened, unflatten_with_validation
 
 
-def contains_instance(value: Any, instance: Any) -> bool:
+def contains_instance(
+    value: Any,
+    instance: Any,
+    extra_children: Callable[[Any], Any | None] | None = None,
+) -> bool:
     """
     Recursively checks if value contains the given instance
     """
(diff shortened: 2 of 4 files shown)


TemidayoA:

TemidayoA:

akshayka (changes requested):
The shape of this fix makes sense … but the implementation is a bit of a hack. The flatten utility is supposed to flatten generic built-in data structures. `_live_children` is an implementation detail of another class that the utilkity shouldn't know about. Perhaps `contains_instance` could take an optional callable that when provided extends the recursion.

TemidayoA:
Oh yeah. I've now moved the marimo-specific knowledge out of flatten.py.  So `contains_instance()` now takes an optional extra_children callable; the `_live_children` logic lives in `hooks_post_execution.py` now through a small `_get_live_children` helper passed in at the call site. I also added a unit test coverage for the new parameter itself in test_flatten.py.

Thanks @akshayka Please re-review again

Issue #10432, opened 2026-08-01
AI config: allow `api_key` to be a shell command via a leading `!`


edouardp:
### Description

I'd like `api_key` in `marimo.toml` to accept a **shell command** when the value starts with `!`, with the command's stdout (trimmed) used as the key:

```toml
[ai.custom_providers.novita]
base_url = "https://api.novita.ai/openai"
api_key = "!security find-generic-password -ws 'novita.ai'"
```

Today the only way to give a custom provider a key is to write the literal secret into `marimo.toml`, because custom providers have no environment-variable fallback (filed separately). Even once that's fixed, environment variables remain a mediocre place for secrets, so I think this is worth having on its own merits.

My setup: the key lives in the macOS Keychain and I currently shuttle it into the environment via direnv purely because marimo can't invoke a command itself. A `!` escape removes that indirection entirely, and works the same for `pass`, `op read` (1Password), `vault kv get`, `gpg --decrypt`, `aws secretsmanager get-secret-value`, or anything else.

### Why this is better from a security standpoint

- **No plaintext secret at rest.** `~/.config/marimo/marimo.toml` is a normal `0644` dotfile that people routinely commit to public dotfile repos or sync between machines. Today using a custom provider *forces* a long-lived credential into exactly the kind of file most likely to be leaked by accident. A `!` command keeps the secret in the OS keychain or secret manager, encrypted at rest and access-controlled by the OS.
- **Better than an environment variable, too.** Exported variables are inherited by every child process marimo spawns (LSP servers, MCP servers, package-manager subprocesses, notebook kernels), and on Linux are readable via `/proc/<pid>/environ`. They also end up in shell history, crash dumps, and CI logs. Fetching on demand narrows exposure to the moment of use rather than the lifetime of the shell.
- **Rotation without editing config.** Rotate in the keychain; marimo picks it up on next fetch. No stale copy to forget about, and short-lived or auto-refreshing tokens (Vault leases, OAuth, STS) become expressible at all — a static string cannot represent them.
- **Safer bug reports.** marimo already masks `api_key` with `********` in the UI (`_config/secrets.py`), which shows the intent is that these values not be seen. That protection stops at the file; a `!` command extends the same guarantee to disk, so pasting a config into an issue is no longer risky.

### Prior art

The `!`-means-shell convention is long-established in **git**, for exactly this "config value that must not be a literal secret" problem. From [gitcredentials(7)](https://git-scm.com/docs/gitcredentials):

[...]

PR #10434, opened 2026-08-02
fix(ai): fall back to OPENAI_API_KEY for custom providers

### LINKED ISSUE #10431: AI: custom providers ignore OPENAI_API_KEY and pass "" to the OpenAI SDK
### Describe the bug

For a custom AI provider, `api_key` can **only** come from `marimo.toml`. There is no environment-variable fallback, and the empty string marimo substitutes actively *suppresses* the OpenAI SDK's own `OPENAI_API_KEY` lookup. The result is an error telling you to set `OPENAI_API_KEY` — when you already have.

I keep my API key in the macOS Keychain and export it via direnv, so it never touches disk:

```sh
export OPENAI_API_KEY=$(security find-generic-password -ws 'novita.ai')
```

With Novita AI (an OpenAI-compatible endpoint) configured the documented way:

```toml
[ai]
enabled = true
mode = "agent"

[ai.custom_providers.novita]
base_url = "https://api.novita.ai/openai"

[ai.models]
chat_model = "novita/deepseek/deepseek-v4-flash-0731"
edit_model = "novita/deepseek/deepseek-v4-flash-0731"
```

every request from the notebook AI panel fails with:

> Error  <br>Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.

`OPENAI_API_KEY` *was* set and visible to the server process (verified with `os.environ` inside `uv run`). The message comes from the `openai` package, not from marimo, which is what makes it so misleading.

### Root cause

In `marimo/_server/ai/config.py`, `_get_key()` ends with a bare `return ""`:

```python
if "api_key" in config:
    key = config["api_key"]
    if key:
        return cast(str, key)
...
if fallback_key:
    return fallback_key
if require_key:
    raise HTTPException(...)
return ""          # <-- custom providers land here
```

`for_custom_provider()` reaches that line because it passes **neither** `fallback_key` **nor** `require_key=True`:

```python
return cls._for_openai_like(
    config,
    key=provider_name,
    name=provider_name.replace("_", " ").title(),
    require_key=False,
    ai_config=provider_config,
)
```

Only `for_openai()` consults the environment (`fallback_key = cls.os_key("OPENAI_API_KEY")`). Notably `for_openai_compatible()` **has the same defect** — it also passes no `fallback_key`, so the natural fallback config is equally broken:

```python
@classmethod
def for_openai_compatible(cls, config: AiConfig) -> AnyProviderConfig:
    return cls._for_openai_like(
        config, "open_ai_compatible", "OpenAI Compatible"
    )
```

That `""` then flows into `OpenAIClientMixin.get_openai_client()` in `marimo/_server/ai/providers.py`:

```python
return AsyncOpenAI(
    default_headers={"api-key": key, **extra_headers},
    api_key=key,          # key == ""
    base_url=base_url,
    project=project,
)
```

And the SDK (`openai/_client.py`) only falls back to the environment when `api_key` is `None`, so an empty string reads as "explicitly supplied and invalid":

```python
if api_key is None:
    api_key = os.environ.get("OPENAI_API_KEY")
...
if (... and not self.api_key and ...):
    raise OpenAIError("Missing credentials. ...")
```

So marimo passing `""` rather than `None` is precisely what breaks the environment fallback that the error message then advises you to use.

Confirmed by resolving the config directly:

```
=== custom_providers.novita ===
api_key non-empty : False
```

There is a smaller, related issue on the same line: `default_headers={"api-key": key}` sends an empty `api-key` header, which is also wrong for endpoints that authenticate via that header.

### Why the obvious workarounds don't work

`[ai.open_ai_compatible]` — identical defect, as shown above: resolves to `""`.

**Impersonating** `openai` (`[ai.open_ai]` with a third-party `base_url`) — resolves the key correctly, but `OpenAIProvider.create_model()` hardcodes `OpenAIResponsesModel`, and many OpenAI-compatible providers implement only Chat Completions. Novita has no Responses endpoint:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' https://api.novita.ai/openai/responses \
    -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
    -d '{"model":"deepseek/deepseek-v3-0324","input":"hi"}'
404
```

Through marimo's stack that surfaces as `ModelHTTPError: status_code: 404, body: 404 page not found`. Chat Completions on the same base URL returns `200`.

So on 0.23.15 there is no supported way to configure an OpenAI-compatible provider whose key comes from the environment.

### Workaround (for anyone who lands here)

Borrow a provider slot whose key *is* plumbed from the environment and that is unknown to pydantic-ai, so `CustomProvider` falls through to a plain `OpenAIChatModel`. `opencode_go` works — it reads `OPENCODE_API_KEY`:

```toml
[ai.opencode_go]
base_url = "https://api.novita.ai/openai"

[ai.models]
chat_model = "opencode-go/deepseek/deepseek-v4-flash-0731"
```

```
provider / model : CustomProvider OpenAIChatModel
api_key non-empty: True
live response    : Paris
```

Tool calling works too, so `mode = "agent"` is usable. It is obviously a hack — the settings UI reports the provider as "OpenCode Go" — and it breaks if anyone actually uses OpenCode.

### Suggested fix

Any of these would resolve it; the third seems most correct:

1. Have `_get_key()` return `None` instead of `""` when no key is found and `require_key` is false, and have `get_openai_client()` omit `api_key` (and the `api-key` header) when it is `None`. The SDK's own env fallback then works as the error message promises.
2. Pass `fallback_key=cls.os_key("OPENAI_API_KEY")` from `for_custom_provider()` and `for_openai_compatible()`.
3. Support a per-provider environment variable, e.g. `MARIMO_AI_<PROVIDER>_API_KEY` (`MARIMO_AI_NOVITA_API_KEY`), so several custom providers can each hold a distinct key. A single shared `OPENAI_API_KEY` cannot express that.

Independently: when a custom provider ends up with no key, marimo should raise its own clear error ("No API key configured for custom provider 'novita'") rather than letting `""` reach the SDK and produce a message that misattributes the cause. That alone would have saved a long debugging session.

Related: marimo-team/marimo#7243 (which introduced named custom providers), marimo-team/marimo#9208 (which added the `opencode_go` slot).

### Will you submit a PR?

- [ ] Yes

### Environment

<details><summary>\`\`\`json</summary>

{  <br>"marimo": "0.23.15",  <br>"OS": "Darwin",  <br>"OS Version": "25.5.0",  <br>"Processor": "arm",  <br>"Python Version": "3.14.4",  <br>"Locale": "en_NZ",  <br>"Binaries": {  <br>"Browser": "--",  <br>"Node": "v24.15.0",  <br>"uv": "0.11.7"  <br>},  <br>"Optional Dependencies": {  <br>"openai": "2.52.0",  <br>"duckdb": "1.5.5",  <br>"mcp": "1.29.0",  <br>"polars": "1.43.1",  <br>"pyarrow": "25.0.0"  <br>},  <br>"Experimental Flags": {  <br>"external_agents": true  <br>}  <br>}

```

`pydantic-ai` 2.22.0 (not listed by `marimo env`).
```

</details>

### Code to reproduce

`~/.config/marimo/marimo.toml`:

```toml
[ai]
enabled = true

[ai.custom_providers.novita]
base_url = "https://api.novita.ai/openai"

[ai.models]
chat_model = "novita/deepseek/deepseek-v4-flash-0731"
```

Then, with `OPENAI_API_KEY` exported in the environment:

```python
from marimo._config.manager import get_default_config_manager
from marimo._server.ai.config import AnyProviderConfig, get_chat_model

ai_cfg = get_default_config_manager(current_path=None).get_config(
    hide_secrets=False
)["ai"]
model = get_chat_model(ai_cfg)

cfg = AnyProviderConfig.for_model(model, ai_cfg)
print("model            :", model)
print("base_url         :", cfg.base_url)
print("api_key non-empty:", bool(cfg.api_key))   # -> False, despite OPENAI_API_KEY being set
```

Opening the AI panel in a notebook and sending any message then raises the  <br>`Missing credentials` error above.

desty:

Fixes #10431

## Summary

Custom OpenAI-compatible providers (`ai.custom_providers.*`) only read `api_key` from `marimo.toml`. When it was missing or empty, marimo passed `""` into the OpenAI SDK, which suppresses the SDK's own `OPENAI_API_KEY` environment lookup — so a process with `OPENAI_API_KEY` set still failed with a missing-credentials error.

This mirrors `for_openai` by using `OPENAI_API_KEY` as `fallback_key` for custom providers.

## Changes

- `AnyProviderConfig.for_custom_provider`: pass `fallback_key=cls.os_key("OPENAI_API_KEY")`
- Unit tests for missing/empty config key, config precedence, `for_model` routing, and no-env empty string

## Test plan

- [x] `uv run --group test pytest tests/_server/ai/test_ai_config.py` (121 passed)
- [x] ruff check on touched files

## Notes

- Scope is intentionally limited to the env fallback reported in #10431.
- Does not implement shell-command keys (`!…`) from #10432.

### DIFF
--- a/marimo/_server/ai/config.py
+++ b/marimo/_server/ai/config.py
@@ -89,10 +89,16 @@ class AnyProviderConfig:
             )
 
         provider_config = cast(dict[str, Any], custom_providers[provider_name])
+        # OpenAI-compatible custom endpoints typically accept the same key
+        # surface as OpenAI (OPENAI_API_KEY). Match for_openai so a missing or
+        # empty api_key in marimo.toml does not pass "" into the SDK and
+        # suppress that env fallback.
+        fallback_key = cls.os_key("OPENAI_API_KEY")
         return cls._for_openai_like(
             config,
             key=provider_name,
             name=provider_name.replace("_", " ").title(),
+            fallback_key=fallback_key,
             require_key=False,
             ai_config=provider_config,
         )
--- a/tests/_server/ai/test_ai_config.py
+++ b/tests/_server/ai/test_ai_config.py
@@ -799,6 +799,90 @@ class TestProviderConfigWithFallback:
             exc_info.value.detail
         )
 
+    @patch.dict(os.environ, {"OPENAI_API_KEY": "env-openai-key"})
+    def test_for_custom_provider_with_fallback_key(self) -> None:
+        """Custom providers fall back to OPENAI_API_KEY when api_key is missing."""
+        config: AiConfig = {
+            "custom_providers": {
+                "novita": {"base_url": "https://api.novita.ai/openai"},
+            }
+        }
+
+        provider_config = AnyProviderConfig.for_custom_provider(
+            config, "novita"
+        )
+
+        assert provider_config.api_key == "env-openai-key"
+        assert provider_config.base_url == "https://api.novita.ai/openai"
+
+    @patch.dict(os.environ, {"OPENAI_API_KEY": "env-openai-key"})
+    def test_for_custom_provider_empty_api_key_uses_fallback(self) -> None:
+        """Empty api_key in marimo.toml should not suppress OPENAI_API_KEY."""
+        config: AiConfig = {
+            "custom_providers": {
+                "novita": {
+                    "base_url": "https://api.novita.ai/openai",
+                    "api_key": "",
+                },
+            }
+        }
+
+        provider_config = AnyProviderConfig.for_custom_provider(
+            config, "novita"
+        )
+
+        assert provider_config.api_key == "env-openai-key"
+
+    @patch.dict(os.environ, {"OPENAI_API_KEY": "env-openai-key"})
+    def test_for_custom_provider_config_key_takes_precedence(self) -> None:
+        """Config api_key takes precedence over OPENAI_API_KEY for custom providers."""
+        config: AiConfig = {
+            "custom_providers": {
+                "novita": {
+                    "base_url": "https://api.novita.ai/openai",
+                    "api_key": "config-novita-key",
+                },
+            }
+        }
+
+        provider_config = AnyProviderConfig.for_custom_provider(
+            config, "novita"
+        )
+
+        assert provider_config.api_key == "config-novita-key"
+
+    @patch.dict(os.environ, {"OPENAI_API_KEY": "env-openai-key"})
+    def test_for_model_custom_provider_with_fallback_key(self) -> None:
+        """for_model routes custom providers through the same env fallback."""
+        config: AiConfig = {
+            "custom_providers": {
+                "novita": {"base_url": "https://api.novita.ai/openai"},
+            }
+        }
+
+        provider_config = AnyProviderConfig.for_model(
+            "novita/deepseek/deepseek-v4-flash", config
+        )
+
+        assert provider_config.api_key == "env-openai-key"
+        assert provider_config.base_url == "https://api.novita.ai/openai"
+
+    @patch.dict(os.environ, {}, clear=True)
+    def test_for_custom_provider_no_fallback_available(self) -> None:
+        """Without config or env key, custom providers keep require_key=False."""
+        config: AiConfig = {
+            "custom_providers": {
+                "novita": {"base_url": "https://api.novita.ai/openai"},
+            }
+        }
+
+        provider_config = AnyProviderConfig.for_custom_provider(
+            config, "novita"
+        )
+
+        assert provider_config.api_key == ""
+        assert provider_config.base_url == "https://api.novita.ai/openai"
+
 
 class TestGetKey:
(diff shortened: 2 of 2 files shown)


mscolnick:
@edouardp this is a security issue that we do not want to support. Thank you for contributing and the feedback

edouardp:
What's the security angle that applies to some providers and not others? If you configure opencode go with your custom providers base path, then the code path for picking up the env variable is already in marimo. That code path just is missing with a custom provider. 

And that is in fact the workaround - set the opencode go config to point your custom provider, and it works today.

If the code is already active in Marimo, then what would be the security problem with it applying to all providers, and not just some?

Issue #7442, opened 2025-12-09
File browser - Hide Folder


joleaf:
### Description

Hi,
In the ui.file_browser ui element, i was wondering if it is possible to hide the folder if `restrict_navigation=True`

<img width="1124" height="442" alt="Image" src="https://github.com/user-attachments/assets/d9a02a5b-9596-4361-9bf4-b4c43218eba1" />

### Suggested solution

idk, but maybe don't show the folder if `restrict_navigation=True` or add a new parameter "hide_folder"?

### Are you willing to submit a PR?

- [ ] Yes

akshayka:
Thanks for the issue. This is a good idea; instead of a new parameter, the folder dropdown should not be shown. Would you like to submit a PR?

joleaf:
Mh, I don't think that always hiding this field is the solution as the `restrict_navigation` parameter only prevents the user from navigating "up". If there are subfolders in the inital_path, this field is important.

akshayka:
Good point.

Is your issue with the dropdown being shown cosmetic, or privacy, or something else? Would you prefer if the path shown when restrict_navigation is True were relative instead of absolute?

khaknazar-ai:
Hi! I’d be happy to work on this if it is still available.

Before starting, could a maintainer confirm the preferred behavior?

My initial approach would be to hide the displayed folder entry when
restrict_navigation=True, without adding a new public API parameter.
If a separate hide_folder option is preferred, I can implement that instead.

I would also add focused tests for the agreed behavior. Thanks!

LeonxLJX:
Hi! I'd like to work on this issue. Please let me know if I can be assigned. Thanks!

LeonxLJX:
Hi! I'd like to add a 'hide folder' option to the file browser. Plan: implement the hide logic, add tests. May I be assigned?

PR #10435, opened 2026-08-02
fix(ui): relative path display when file_browser restricts navigation

desty:

Addresses #7442

## Summary

When `restrict_navigation=True`, the path control showed absolute host paths (e.g. `/Users/...`). That is usually noise (and can feel private) under restricted navigation.

This change follows the thread on #7442:

- **@akshayka** suggested showing a relative path when `restrict_navigation` is on (rather than a new API param).
- **@joleaf** pointed out that always hiding the control breaks in-tree parent navigation when there are subfolders.

### Behavior

| Situation | UI |
|-----------|-----|
| `restrict_navigation=False` | Unchanged: absolute path dropdown |
| Restricted, only root reachable | Static relative label (`.`) — no useless 1-option dropdown |
| Restricted, nested under initial path | Dropdown kept, labels relative to `initial_path` (`.`, `subdir`, …); values stay absolute for navigation |

No new public API.

## Test plan

- [x] `pnpm exec vitest run src/utils/pathUtils.test.ts src/plugins/impl/__tests__/FileBrowserPlugin.test.tsx` (42 passed)

### DIFF
--- a/frontend/src/plugins/impl/FileBrowserPlugin.tsx
+++ b/frontend/src/plugins/impl/FileBrowserPlugin.tsx
@@ -20,7 +20,10 @@ import { useAsyncData } from "@/hooks/useAsyncData";
 import { useInternalStateWithSync } from "@/hooks/useInternalStateWithSync";
 import { cn } from "@/utils/cn";
 import { type FilePath, PathBuilder, Paths } from "@/utils/paths";
-import { getProtocolAndParentDirectories } from "@/utils/pathUtils";
+import {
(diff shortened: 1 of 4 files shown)


desty:
Still relevant — converting out of draft.

Fixes the `file_browser` path display when `restrict_navigation=True` so the UI shows paths relative to `initial_path` instead of absolute host paths. Checks are green; ready for review.

desty:
Rebased onto latest `main`. Still mergeable and ready for review.

PR #10437, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
Clarify that the OpenAI Python client `base_url` pattern works with marimo and with other OpenAI-compatible multi-model gateways, using DaoXE (https://api.daoxe.com/v1) as one concrete example. Docs only.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/docs/guides/configuration/llm_providers.md
+++ b/docs/guides/configuration/llm_providers.md
@@ -104,7 +104,7 @@ api_key = "sk-proj-..."
 !!! note "OpenAI-compatible providers"
     If your model does not start with `openai/`, it will not be routed to the OpenAI config, and likely will be routed to the OpenAI-compatible config.
 
-> **Tip:** Any OpenAI-compatible multi-model gateway works with the same `base_url` pattern — for example [DaoXE](https://daoxe.com/?utm_source=github&utm_medium=organic&utm_campaign=marimo&utm_content=llm_providers) at `https://api.daoxe.com/v1`.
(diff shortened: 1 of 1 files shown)


docs/guides/configuration/llm_providers.md
```diff
@@ -104,6 +104,9 @@ api_key = "sk-proj-..."
 !!! note "OpenAI-compatible providers"
     If your model does not start with `openai/`, it will not be routed to the OpenAI config, and likely will be routed to the OpenAI-compatible config.
 
+> **Tip:** Any OpenAI-compatible multi-model gateway works with the same `base_url` pattern — for example [DaoXE](https://daoxe.com/?utm_source=github&utm_medium=organic&utm_campaign=marimo&utm_content=llm_providers) at `https://api.daoxe.com/v1`.
```
manzt: Could we make this vendor-neutral and use an admonition?

```suggestion
!!! tip
    OpenAI-compatible providers and gateways can be configured by setting `base_url` to the provider's API endpoint.
```

manzt (changes requested):

manzt:
Can you please sign our CLA?

seven7763:
Thanks both — updated. The tip is now vendor-neutral with no external link, matching @mscolnick's suggestion to drop the link and keep the guidance, and it lives only in the configuration guide as @manzt asked. I've closed #10438 so the same pattern isn't documented in two places.

seven7763:

recheck

seven7763:
Closing — the vendor-neutral requirement means this PR would lose its purpose. Thanks for the review.

PR #10453, opened 2026-08-04
feat: report missing Chromium for PDF export

peter-gy:
## Summary

Current PDF export availability we had checked whether Playwright could be imported, but it did not verify that Playwright's matching Chromium executable existed. As a result, PDF could appear available even though export would fail.

This adds an exporter-owned Chromium readiness probe and reports missing browser setup separately from missing Python packages. When Chromium is missing, the dialog keeps PDF visible, marks it unavailable, and shows the install command. Once Chromium is ready, PDF export is enabled.

## Demo

Screenshot 2026-08-04 at 8 56 56


peter-gy:
@cubic-dev-ai review please

mscolnick (approved):

Issue #9624, opened 2026-05-20
Export to HTML skips mo.lazy outputs


vascotenner:
### Describe the bug

When exporting to HTML, the outputs of cells that contains mo.lazy are ommit.

# Example
```
import marimo as mo
import matplotlib.pyplot as plt
mo.lazy(lambda: plt.scatter(1,1))
```

# Results in:
<img width="614" height="186" alt="Image" src="https://github.com/user-attachments/assets/f30067cb-65a5-4fa3-be63-2569f8ae47ae" />

# Expected output:

<img width="521" height="476" alt="Image" src="https://github.com/user-attachments/assets/43d0ea43-3f10-4163-8c31-f95be20a1c73" />

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.5",
  "editable": false,
  "location": "C:/Users/xxxx/.virtualenvs/marimo/Lib/site-packages/marimo",
  "OS": "Windows",
  "OS Version": "11",
  "Processor": "Intel64 Family 6 Model 186 Stepping 2, GenuineIntel",
  "Python Version": "3.14.0",
  "Locale": "--",
  "Binaries": {
    "Browser": "--",
    "Node": "--",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.3.1",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.18.1",
    "packaging": "26.0",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pymdown-extensions": "10.21.2",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.14.0",
    "typing-extensions": "4.15.0",
    "uvicorn": "0.42.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "altair": "6.1.0",
    "loro": "1.10.3",
    "nbformat": "5.10.4",
    "pandas": "3.0.3",
    "polars": "1.39.3",
    "pytest": "9.0.3",
    "ruff": "0.15.10"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

```python
import marimo as mo
import matplotlib.pyplot as plt
mo.lazy(lambda: plt.scatter(1,1))
```

mscolnick:
[...]

PR #10465, opened 2026-08-05
fix: resolve mo.lazy content eagerly during static HTML export

bunlongheng:

## What Problem This Solves

`mo.lazy` outputs are silently omitted when exporting a notebook to HTML (issue #9624). A cell like:

```python
mo.lazy(lambda: plt.scatter(1, 1))
```

produces an empty loading spinner in the exported HTML because the lazy widget requires a live kernel to invoke its `load` function - which is not available in a static export.

ipynb and PDF exports already resolve `mo.lazy` eagerly via `patch_html_for_non_interactive_output()`, but HTML export did not apply this treatment.

## Evidence

The previous test `test_export_html_keeps_lazy_placeholder` (now renamed and extended) documented the broken behavior with the comment:

> "A lazy-specific resolution path can be added later as a follow-up."

This PR implements that follow-up.

## Solution

Introduces `patch_lazy_for_static_export()` - a narrow context manager that sets a new `MARIMO_RESOLVE_LAZY` env flag. Unlike `patch_html_for_non_interactive_output()` (which also downgrades tables, altair, plotly to non-interactive fallbacks), this flag only affects `mo.lazy`, so the exported HTML keeps fully interactive widgets.

### Changes

- `marimo/_output/hypertext.py` - adds `MARIMO_RESOLVE_LAZY_KEY`, `patch_lazy_for_static_export()`, `should_resolve_lazy()`
- `marimo/_plugins/stateless/lazy.py` - checks `should_resolve_lazy()` alongside `is_non_interactive()` in `__new__`; updates docstring
- `marimo/_export/file.py` - wraps `run_notebook()` in `export_html()` with the new context manager
- `tests/_export/test_exporter.py` - replaces `test_export_html_keeps_lazy_placeholder` with `test_export_html_resolves_lazy` covering eager-value, sync-callable, and async-callable (async still falls back to the widget)

## Limitations

Async callables cannot be awaited from `__new__`, so they still render as a lazy placeholder in static HTML exports (same as iPynb/PDF behavior).

### DIFF
--- a/marimo/_export/file.py
+++ b/marimo/_export/file.py
@@ -53,7 +53,10 @@ from marimo._messaging.notification import (
 )
 from marimo._messaging.serde import deserialize_kernel_message
 from marimo._messaging.types import KernelMessage
-from marimo._output.hypertext import patch_html_for_non_interactive_output
+from marimo._output.hypertext import (
+    patch_html_for_non_interactive_output,
+    patch_lazy_for_static_export,
+)
 from marimo._runtime.commands import AppMetadata
 from marimo._runtime.patches import extract_docstring_from_header
 from marimo._schemas.export_options import (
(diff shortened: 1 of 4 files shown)


mscolnick:
@cubic-dev-ai

bunlongheng:

PR #10471, opened 2026-08-06
feat(export): install requirements from dialog

peter-gy:
Companion of #10473.

## Summary

The export dialog identifies missing requirements, but users must install them outside the web UI.

Add a compact **Install** action to unavailable export formats. The action calls the server-owned installation contract from #10473, refreshes availability after installation, and prevents overlapping dialog actions while the request is running. Installation failures use the existing request toast above the dialog.

Read-only and kiosk views keep the manual setup command visible.

## Demo

export-rebase-install-error-settled

### DIFF
--- a/frontend/src/core/network/requests-lazy.ts
+++ b/frontend/src/core/network/requests-lazy.ts
@@ -26,6 +26,10 @@ type AllRequests = EditRequests & RunRequests;
 //   executing. Use for user-initiated actions that should "just work" and
 //   kick off the kernel if needed (e.g., clicking Run).
 //
+// - startConnectionWithoutKernel: Initializes the runtime and waits for an open
+//   connection. Use for session-scoped server requests that do not need an
+//   instantiated kernel.
+//
 // - waitForConnectionOpen: Waits for an existing connection but won't start one.
 //   Use for operations that depend on a running kernel but shouldn't be the
 //   trigger to start it (e.g., saving, interrupting).
@@ -38,6 +42,7 @@ type Action =
   | "throwError"
   | "dropRequest"
   | "startConnection"
+  | "startConnectionWithoutKernel"
   | "waitForConnectionOpen"
   | "serverOnly";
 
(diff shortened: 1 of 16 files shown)


peter-gy:
@cubic-dev-ai review please

peter-gy:
@cubic-dev-ai review please

frontend/src/core/network/requests-network.ts
```diff
@@ -413,6 +413,14 @@ export function createNetworkRequests(): EditRequests & RunRequests {
     getExportAvailability: () => {
       return getClient().GET("/api/export/availability").then(handleResponse);
     },
+    installExportRequirements: (request) => {
```
mscolnick, line 416: is it possible to avoid another endpoint by reusing the existing install command?

frontend/src/core/network/requests-lazy.ts
```diff
@@ -38,14 +42,15 @@ type Action =
   | "throwError"
   | "dropRequest"
   | "startConnection"
+  | "startConnectionWithoutKernel"
```
mscolnick, line 45: Should this just be `waitForConnection`? I may need to revisit some of this logic.

frontend/src/core/network/requests-network.ts
peter-gy: Initially I considered extending the existing endpoint with an `installPlaywrightChromium` boolean. But that would make the client responsible for each format's installation recipe. E.g. for PDF the `playwright` package may already be installed while the Chromium browser is still missing. So `requirements` here is not only a package requirement to be installed but potentially an additional command through an installed pkg like `playwright install`.

With the current approach, the client sends the format and the server resolves and installs its requirements. We could add `format` to the existing endpoint though if you would prefer to keep a single endpoint!

frontend/src/core/network/requests-lazy.ts
peter-gy: Do you mean the existing [`waitForConnectionOpen` action](https://github.com/marimo-team/marimo/blob/5d42d5b2ff8618ad36a2def2bebba6d90042a0e5/frontend/src/core/network/requests-lazy.ts#L37-L41)?

My idea here was that installing server-side export requirements needs an active connection for the capability check, but no instantiated kernel. As far as I can see, the current `waitForConnectionOpen` impl does not start a lazy connection and also waits for kernel instantiation. This request needs to start the connection, then proceed without waiting for the kernel.

mscolnick (approved):

PR #10473, opened 2026-08-06
feat(export): install server requirements

peter-gy:
## Summary

Add a server-owned installation contract for export requirements.

The client sends an export format. The server resolves its missing packages and setup requirements, installs them into the server Python environment, and returns refreshed availability for every server-backed export format.

- Uses `uv` when available and `pip` otherwise.
- Runs package installation outside the event loop.
- Installs Playwright Chromium through a typed setup requirement with a 10-minute timeout.
- Requires edit capability and verifies the requested format after installation.

The export dialog consumes this contract in #10471.

### DIFF
--- a/frontend/src/__mocks__/requests.ts
+++ b/frontend/src/__mocks__/requests.ts
@@ -100,6 +100,10 @@ export const MockRequestClient = {
           },
         ],
       }),
+      installExportRequirements: vi.fn().mockResolvedValue({
+        source: "server",
+        formats: [],
+      }),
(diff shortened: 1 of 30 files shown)


mscolnick (dismissed):

peter-gy:
@mscolnick should we go ahead and merge this along with #10471?

mscolnick:
@peter-gy, yes both are good to merge!

Light2Dark (approved):

PR #10474, opened 2026-08-06
Resolve file keys through the workspace when matching sessions

manzt:
Directory workspaces serve workspace-relative file keys (#7722), but the edit-mode resume strategy matches sessions with `os.path.abspath(file_key)`, which resolves against the server CWD. Whenever the CWD is not the workspace root, a browser refresh does not find the existing session and starts a new kernel on every reload, leaving the old ones orphaned.

These changes resolve file keys through the workspace that owns them and move session matching into one repository predicate shared by the resume, viewer, and RTC lookups:

```py
def get_all_by_file_key(
    self,
    file_key: MarimoFileKey,
    resolved_path: str | None = None,
) -> list[Session]: ...
```

A session matches on the key it was created under or on its resolved path, so resume also survives renames and untitled notebooks, where the key has no backing file.

### DIFF
--- a/marimo/_server/resume_strategies.py
+++ b/marimo/_server/resume_strategies.py
@@ -17,7 +17,13 @@ from marimo._types.ids import SessionId
 LOGGER = _loggers.marimo_logger()
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+
     from marimo._server.workspace import MarimoFileKey
+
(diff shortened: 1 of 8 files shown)


dmadisetti (approved):
Tested under the matrix of `edit`/ `run`/ `--sandbox` and all seems to work

PR #10475, opened 2026-08-06
Enforce one edit session per notebook under connection races

manzt:
A browser refresh can deliver the new connection before the old socket's disconnect is processed. The session's main consumer is still registered but CLOSED, so it is neither resumed nor joinable and the connector starts a second kernel for the same notebook.

These changes reclaim CLOSED sessions on resume and never create a session for a notebook that already has one; the connection joins it as a viewer instead.


manzt:
Closing in favor of a follow-up that promotes the refreshed connection to editor when the previous editor disconnects, addressing the actual connection race.

mscolnick (approved):

PR #10492, opened 2026-08-07
fix: kill processes when interrupting a cell

### LINKED ISSUE #10468: interrupt doesn't reach sub-processes
### Describe the bug

**Summary**

Interrupting a running cell does not terminate subprocesses that the cell started. The cell reports as interrupted in the notebook, but its child processes keep running.

**Reproduction**

```
import subprocess, time
p = subprocess.Popen(["sleep", "600"])
p.wait()
```

then interrupt the cell and go to the terminal to find it:
```
$ ps aux | grep sleep
chrberr+  115088  0.0  0.0   5540  2156 ?        S    00:17   0:00 sleep 600
chrberr+  115293  0.0  0.0   6596  2384 pts/0    S+   00:18   0:00 grep sleep
```
The subprocess is still running.

**Impact**

sleep is harmless, but I hit this with real compute jobs. I use CmdStanPy to run Stan models for Bayesian inference, and CmdStanPy spawns one CmdStan subprocess per MCMC chain. When I interrupt a cell to correct a sampling parameter, all four chains keep running at full CPU.

**Three consequences:**

*Resource leak that compounds*: 
Each interrupt-and-retry cycle leaves four more processes behind. After a few iterations the machine is heavily oversubscribed, and a model that normally fits in 6 minutes takes far longer — while it appears to be a single run.
*Corrupted on-disk state*: 
The orphans keep writing to their output CSVs, including into directories that have been deleted (the inodes stay open). Later reads of that directory can pick up a mix of abandoned and current output.
*Confusing diagnostics*: 
The orphans keep emitting progress through the same logger as the new run, so the notebook shows interleaved iteration counts from two different runs. This is hard to interpret and sent me down several wrong explanations before I checked ps.

**Context**

I've recently migrated a few projects from JupyterLab and have been enjoying marimo a lot. This is the one behaviour that has been genuinely costly, since interrupting a long job to fix a parameter is a normal part of my workflow.

My understanding (offered as a guess, not a diagnosis) is that Jupyter's interrupt sends SIGINT to the kernel's process group, so subprocesses inherit the signal and terminate, whereas marimo raises an exception inside the Python execution thread, which never reaches a child process.

Possible scope question: should this also cover kernel restart? After a restart the orphans are reparented to init and are no longer children of anything marimo can identify, so they may need handling at a different level (or may reasonably be out of scope).

### Will you submit a PR?

- [ ] Yes

### Environment

$ marimo env
{
  "marimo": "0.23.16",
  "editable": false,
  "location": "/home/chrberrig/Documents/Coding_fun/mortalityrates_perner/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "6.12.90+deb13.1-amd64",
  "Processor": "",
  "Python Version": "3.13.5",
  "Locale": "en_DK",
  "Binaries": {
    "Browser": "--",
    "Node": "v20.19.2",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.3",
    "narwhals": "2.24.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.1",
    "uvicorn": "0.52.1",
    "websockets": "17.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.13.2",
    "pandas": "3.0.2"
  },
  "Experimental Flags": {}
}

other:
{
  echo "- marimo: $(marimo --version)"
  echo "- Python: $(python --version 2>&1 | cut -d' ' -f2)"
  echo "- CmdStanPy: $(python -c 'import cmdstanpy; print(cmdstanpy.__version__)')"
  echo "- OS: $(. /etc/os-release && echo "$PRETTY_NAME")"
  echo "- CmdStan: $(python -c 'import cmdstanpy; print(cmdstanpy.cmdstan_version())' 2>/dev/null)"
}
- marimo: 0.23.16
- Python: 3.13.5
- CmdStanPy: 1.3.0
- OS: Debian GNU/Linux 13 (trixie)
- CmdStan: (2, 38)

### Code to reproduce

```
import subprocess, time
p = subprocess.Popen(["sleep", "600"])
p.wait()
```
in Marimo Notebook. 

Then interrupting the cell and go to the terminal to find it:
```
$ ps aux | grep sleep
chrberr+  115088  0.0  0.0   5540  2156 ?        S    00:17   0:00 sleep 600
chrberr+  115293  0.0  0.0   6596  2384 pts/0    S+   00:18   0:00 grep sleep
```
The subprocess is still running.

akshayka:
Previously, interrupts sent a SIGINT to just the kernel process but not the process group. That meant that processes in the process group weren't started, leading to orphaned processes.

In this change, we send kill the whole process group. Care is taken to start processes that should not be terminated in their own sessions, so they don't accidentally get killed.

Fixes marimo-team/marimo#10468

### DIFF
--- a/docs/guides/reactivity.md
+++ b/docs/guides/reactivity.md
@@ -147,9 +147,7 @@ or try creating a new variable instead.
 
 ### Interrupts
 
-Click the stop button to interrupt a running cell and terminate processes your
-notebook started. Interrupting a cell also cancels execution of any queued
-dependent cells.
+Click the stop button to interrupt a running cell. On Unix-like systems, this also terminates subprocesses started by the cell (except those started in a new session). Interrupting a cell also cancels execution of any queued dependent cells.
 
 ## Global variable names must be unique
 
--- a/docs/guides/reactivity.md
+++ b/docs/guides/reactivity.md
@@ -147,7 +147,10 @@ or try creating a new variable instead.
 
 ### Interrupts
 
-Click the stop button to interrupt a running cell. On Unix-like systems, this also terminates subprocesses started by the cell (except those started in a new session). Interrupting a cell also cancels execution of any queued dependent cells.
+Click the stop button to interrupt a running cell. On Unix-like systems, this also
+terminates subprocesses started by the cell (except those started in a new
+session). Interrupting a cell also cancels execution of any queued dependent
+cells.
 
 ## Global variable names must be unique
 
--- a/docs/guides/reactivity.md
+++ b/docs/guides/reactivity.md
@@ -147,10 +147,11 @@ or try creating a new variable instead.
 
 ### Interrupts
 
-Click the stop button to interrupt a running cell. On Unix-like systems, this also
-terminates subprocesses started by the cell (except those started in a new
-session). Interrupting a cell also cancels execution of any queued dependent
-cells.
+Click the stop button to interrupt a running cell. On Unix-like systems, this
+also interrupts subprocesses in the kernel's process group. To keep a
+subprocess running after an interrupt, start it in a new session, for example
+with `subprocess.Popen(..., start_new_session=True)`. Interrupting a cell also
+cancels execution of any queued dependent cells.
 
 ## Global variable names must be unique
 
--- a/docs/guides/reactivity.md
+++ b/docs/guides/reactivity.md
@@ -145,6 +145,12 @@ or try creating a new variable instead.
         df["another_column"] = [3, 4]
         ```
 
+### Interrupts
+
+Click the stop button to interrupt a running cell and terminate processes your
+notebook started. Interrupting a cell also cancels execution of any queued
+dependent cells.
+
 ## Global variable names must be unique
 
 **marimo requires that every global variable be defined by only one cell.**
--- a/marimo/_data/_external_storage/utils.py
+++ b/marimo/_data/_external_storage/utils.py
@@ -1,3 +1,4 @@
+# Copyright 2026 Marimo. All rights reserved.
 from marimo import _loggers
 from marimo._data._external_storage.models import (
     StorageEntry,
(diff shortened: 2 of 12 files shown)


docs/guides/reactivity.md
```diff
@@ -145,6 +145,13 @@ or try creating a new variable instead.
         df["another_column"] = [3, 4]
         ```
 
+### Interrupts
+
+Click the stop button to interrupt a running cell. On Unix-like systems, this also
+terminates subprocesses started by the cell (except those started in a new
+session). Interrupting a cell also cancels execution of any queued dependent
+cells.
```
manzt: I had Codex help me trace this behavior and compare it with Jupyter. This is consistent with Jupyter’s Unix interrupt semantics, which also signal the kernel process group. 

Should/could we make the escape hatch explicit here for intentionally persistent background processes? Something like:

```suggestion
Click the stop button to interrupt a running cell. On Unix-like systems, this
also interrupts subprocesses in the kernel's process group. To keep a
subprocess running after an interrupt, start it in a new session, for example
with `subprocess.Popen(..., start_new_session=True)`. Interrupting a cell also
cancels execution of any queued dependent cells.
```

manzt (dismissed):

docs/guides/reactivity.md
akshayka: Thanks for the thorough review. Yes, I designed it to be consistent with Jupyter. I considered documenting this, but I was on the fence. I'll include it.

PR #10498, opened 2026-08-09
fix(ui): raise a helpful error for invalid radio/multiselect option names

### LINKED ISSUE #10497: mo.ui.radio and mo.ui.multiselect raise a bare KeyError for invalid option names
### Describe the bug

`mo.ui.dropdown` validates option names and raises a `ValueError` naming the valid options. `mo.ui.radio` and `mo.ui.multiselect` index `self.options` directly, so an invalid option surfaces as an uninformative `KeyError`.

```python
import marimo as mo

mo.ui.dropdown(options=["a", "b"], value="zzz")
# ValueError: The option name 'zzz' is not a valid option.
#             Please use one of the following options: ['a', 'b']

mo.ui.radio(options=["a", "b"], value="zzz")
# KeyError: 'zzz'

mo.ui.multiselect(options=["a", "b"], value=["zzz"])
# KeyError: 'zzz'
```

Reproduced on marimo 0.23.16 (macOS, Python 3.12).

`_convert_value` runs both at construction and on every update from the frontend (`ui_element.py`), so this affects both paths.

Source: `dropdown._convert_value` performs the check in `marimo/_plugins/ui/_impl/input.py`, while `radio._convert_value` and `multiselect._convert_value` do a bare dict lookup.

### Suggested fix

Extract dropdown's existing check into a small helper and use it in all three components, so the message stays consistent by construction.

Note this would change the exception type on the error path from `KeyError` to `ValueError` for `radio` and `multiselect`. Happy to keep `KeyError` and only improve the message if you'd prefer to avoid that.

### Will you submit a PR?

- [x] Yes

Chinmayrawat15:

## 📝 Summary

Closes #10497

`mo.ui.dropdown` validates incoming option names and raises a `ValueError` naming the valid options. `mo.ui.radio` and `mo.ui.multiselect` did a bare dict lookup, so an invalid option surfaced as an uninformative `KeyError`:

```python
mo.ui.radio(options=["a", "b"], value="zzz")          # KeyError: 'zzz'
mo.ui.multiselect(options=["a", "b"], value=["zzz"])  # KeyError: 'zzz'
```

Both now produce the message `dropdown` already produced:

```
ValueError: The option name 'zzz' is not a valid option.
            Please use one of the following options: ['a', 'b']
```

This extracts dropdown's existing check into a `_validate_option_name` helper used by all three components, so the message stays consistent by construction rather than by duplication. `_convert_value` runs both at construction and on updates from the frontend, so both paths are covered.

`radio` keeps returning `None` for the no-selection case, and `multiselect` validates every item before converting any, so it fails cleanly rather than half-applying.

**Behavior change worth calling out:** this changes the exception type on the error path from `KeyError` to `ValueError` for `radio` and `multiselect`, matching `dropdown`. If you'd rather preserve `KeyError`, I'm happy to keep the type and only improve the message.

## 🧪 Testing

Added `test_radio_invalid_value`, `test_dropdown_invalid_value`, and `test_multiselect_invalid_value`, each covering both construction and frontend update. The radio and multiselect tests were confirmed to fail with `KeyError` when the source change is reverted.

```
tests/_plugins/ui/_impl/test_input.py    50 passed
tests/_plugins/ui/                       791 passed, 0 failed
ruff check / ruff format / typos         clean
mypy marimo/_plugins/ui/_impl/input.py   Success: no issues found
```

## 🔍 Checks

- [x] For large changes, or changes that affect the public API: discussed in #10497

## ✅ Merge Checklist

### DIFF
--- a/tests/_plugins/ui/_impl/test_input.py
+++ b/tests/_plugins/ui/_impl/test_input.py
@@ -393,6 +393,19 @@ def test_radio_from_dataframe() -> None:
     assert radio.value == "b"
 
 
+def test_radio_invalid_value() -> None:
+    with pytest.raises(ValueError) as e:
+        ui.radio(options=["1", "2", "3"], value="4")
+
+    assert "is not a valid option" in str(e.value)
+
+    radio = ui.radio(options=["1", "2", "3"], value="1")
+    with pytest.raises(ValueError) as e:
+        radio._update("4")
+
+    assert "is not a valid option" in str(e.value)
+
+
 def test_dropdown() -> None:
     dd = ui.dropdown(options=["1", "2", "3"])
     assert dd.value is None
@@ -464,6 +477,19 @@ def test_dropdown_with_non_string_options() -> None:
 def test_dropdown_lots_of_options() -> None:
     dropdown = ui.dropdown(options={str(i): i for i in range(2000)})
     assert dropdown._component_args["searchable"] is True
+
+
+def test_dropdown_invalid_value() -> None:
+    with pytest.raises(ValueError) as e:
+        ui.dropdown(options=["1", "2", "3"], value="4")
+
+    assert "is not a valid option" in str(e.value)
+
+    dd = ui.dropdown(options=["1", "2", "3"], value="1")
+    with pytest.raises(ValueError) as e:
+        dd._update(["4"])
+
+    assert "is not a valid option" in str(e.value)
 
 
 def test_dropdown_disabled() -> None:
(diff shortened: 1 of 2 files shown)


Chinmayrawat15:

tests/_plugins/ui/_impl/test_input.py
```diff
@@ -595,6 +621,19 @@ def test_multiselect_too_many_options() -> None:
...
+    with pytest.raises(ValueError) as e:
+        ui.multiselect(options=["1", "2", "3"], value=["4"])
+
+    assert "is not a valid option" in str(e.value)
+
+    ms = ui.multiselect(options=["1", "2", "3"], value=["1"])
+    with pytest.raises(ValueError) as e:
+        ms._update(["1", "4"])
+
+    assert "is not a valid option" in str(e.value)
```
Light2Dark: can the assertions also check if the valid options are printed

Light2Dark (changes requested):

tests/_plugins/ui/_impl/test_input.py
Chinmayrawat15: Done in 24b5199! The assertions now pin both the offending name and the valid options:
 
 assert "The option name '4' is not a valid option" in str(e.value) 
 assert "['1', '2', '3']" in str(e.value) 
 
 Applied the same to the radio and dropdown tests, since they had the same gap.

Light2Dark (approved):

PR #10505, opened 2026-08-10
fix(runtime): run async notebooks from inside a running event loop

### LINKED ISSUE #9572: Embeds don't work when notebook runs as script
### Describe the bug

Notebooks with embeds are async notebooks.

When running a notebook that embeds another async notebook, Marimo raises a RuntimeError.

```
$ uv run python marimo_recursive_factorial.py
...
  File "/Users/yairchu/dev/moops/marimo_recursive_factorial.py", line 35, in <module>
    child_result = await child.embed(defs={"n": n - 1}) if child else None
...
  File "/Users/yairchu/dev/moops/.venv/lib/python3.13/site-packages/marimo/_runtime/app/script_runner.py", line 219, in run
    outputs, defs = asyncio.run(
...
RuntimeError: asyncio.run() cannot be called from a running event loop
...
```

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.23.6",
  "editable": false,
  "location": "/Users/yairchu/dev/moops/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.2.0",
  "Processor": "arm",
  "Python Version": "3.13.1",
  "Locale": "C/C",
  "Binaries": {
    "Browser": "140.0.7339.134",
    "Node": "v23.11.0",
    "uv": "0.7.12 (dc3fd4647 2025-06-06)"
  },
  "Dependencies": {
    "click": "8.3.3",
    "docutils": "0.22.4",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.21.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.0.0",
    "tomlkit": "0.15.0",
    "uvicorn": "0.46.0",
    "websockets": "16.0"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "pytest": "9.0.3",
    "ruff": "0.15.12"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

`marimo_recursive_factorial.py`:

```python
import marimo

__generated_with = "0.23.6"


@app.cell
def _():
    import marimo_recursive_factorial

    return (marimo_recursive_factorial,)


@app.cell
def _():
    n = 5
    return (n,)


@app.cell
def _(marimo_recursive_factorial, n):
    child = marimo_recursive_factorial.app.clone() if n > 1 else None
    return (child,)


@app.cell
async def _(child, n):
    child_result = await child.embed(defs={"n": n - 1}) if child else None
    child_result.output if child else None
    return (child_result,)


@app.cell
def _(child, child_result, n):
    result = 1 if child is None else n * child_result.defs["result"]
    return (result,)


@app.cell
def _(n, result):
    print(f"factorial({n}) = {result}")
    return


if __name__ == "__main__":
    app.run()
```

### LINKED ISSUE #9646: app.run doesn't work for notebooks with embeds or async cells
### Describe the bug

When a notebook embeds any other notebook, it's `app.run()` no longer works.

This might be related to marimo-team/marimo#9572, but the repro example here is smaller:

### Will you submit a PR?

- [ ] Yes

### Environment

<details><summary>\`\`\`</summary>

{<br>  "marimo": "0.23.6",<br>  "editable": true,<br>  "location": "/Users/yairchu/dev/marimo/marimo",<br>  "OS": "Darwin",<br>  "OS Version": "25.2.0",<br>  "Processor": "arm",<br>  "Python Version": "3.13.1",<br>  "Locale": "C/en_US",<br>  "Binaries": {<br>    "Browser": "140.0.7339.134",<br>    "Node": "v23.11.0",<br>    "uv": "0.7.12 (dc3fd4647 2025-06-06)"<br>  },<br>  "Dependencies": {<br>    "click": "8.2.1",<br>    "docutils": "0.22.2",<br>    "itsdangerous": "2.2.0",<br>    "jedi": "0.19.2",<br>    "markdown": "3.9",<br>    "narwhals": "2.7.0",<br>    "packaging": "25.0",<br>    "psutil": "7.1.0",<br>    "pygments": "2.19.2",<br>    "pymdown-extensions": "10.21.3",<br>    "pyyaml": "6.0.3",<br>    "starlette": "0.48.0",<br>    "tomlkit": "0.13.3",<br>    "typing-extensions": "4.15.0",<br>    "uvicorn": "0.35.0",<br>    "websockets": "15.0.1"<br>  },<br>  "Optional Dependencies": {<br>    "duckdb": "1.4.1",<br>    "loro": "1.10.3",<br>    "openai": "2.37.0",<br>    "pyarrow": "21.0.0",<br>    "ruff": "0.15.13",<br>    "sqlglot": "27.26.0"<br>  },<br>  "Experimental Flags": {<br>    "cache_panel": true,<br>    "isolate_apps": true,<br>    "multi_column": true<br>  }<br>}

```
```

</details>

### Code to reproduce

`test_notebook.py`:

```python
import marimo

__generated_with = "0.23.6"


@app.cell
def _():
    import marimo as mo

    return


@app.cell
def _():
    import test_notebook

    return (test_notebook,)


@app.cell
def _():
    n = 5
    return (n,)


@app.cell
def _(n, test_notebook):
    result = (
        1 if n == 0 else n * test_notebook.app.run(defs={"n": n - 1})[1]["result"]
    )
    print(f"factorial({n}) = {result}")
    return


@app.cell
async def _():
    # Removing this cell makes the notebook work


    async def _t():
        return 5


    _ = await _t()
    return


@app.cell
def _():
    return


if __name__ == "__main__":
    app.run()
```

Chinmayrawat15:

## 📝 Summary

Closes #9572
Closes #9646

`AppScriptRunner.run()` called `asyncio.run()` unconditionally as soon as a notebook contained a coroutine cell. `asyncio.run()` raises when a loop is already running on the calling thread, so an async notebook failed whenever it was driven from async code:

```
RuntimeError: asyncio.run() cannot be called from a running event loop
```

This is what both issues hit — one notebook running another while its own loop is live. `App.embed()` outside a notebook delegates to `App.run()`, which is why #9572 surfaced as an embed failure. Embedding isn't actually required to trigger it:

```python
import asyncio
import marimo

app = marimo.App()


@app.cell
async def _():
    import asyncio as _a

    await _a.sleep(0)
    x = 42
    return (x,)


async def main():
    app.run()


app.run()             # fine
asyncio.run(main())   # RuntimeError
```

### How

**`marimo/_utils/asyncio_utils.py`** — new `run_coroutine_blocking()`. It keeps using plain `asyncio.run` when no loop is running, and only when one *is* running drives the coroutine on a worker thread while the caller blocks on a `Future`.

A bare daemon thread is used rather than a `ThreadPoolExecutor`: the executor's context manager shuts down with `wait=True`, so a `KeyboardInterrupt` while the caller is blocked would hang until the whole nested run finished. `Future.result()` waits on a condition variable that the interrupt can break out of, and it re-raises with the worker's traceback intact.

**`marimo/_runtime/app/script_runner.py`** — new `_run_asynchronous_in_context()`. marimo's runtime context is thread-local, so the coroutine reinstalls it when its body lands on a thread that doesn't have one, and tears it down afterwards. It's a no-op on the non-offloaded path, and purely synchronous notebooks keep their existing no-event-loop guarantee (the `_run_synchronous` path is untouched).

While here: `run()` had a function-local `from marimo._runtime.context import get_context` that shadowed the module-level import for the entire function scope. It's the same object, so the local import is dropped.

### Semantics worth a maintainer's eye

`app.run()` now blocks the caller's event loop for the duration of the nested run. That seems right given `App.run()` is synchronous by contract and already blocks, and it's strictly better than today's hard failure — but if you'd prefer an awaitable variant instead of (or alongside) this, happy to go that way.

### Testing

- Reproduced both issues' actual notebooks — the self-importing `app.run()` factorial from #9646 and the recursive `await app.embed()` one from #9572. Both raised `RuntimeError` before this change and now compute correctly through several levels of nesting.
- 5 new tests in `tests/_ast/test_app.py`: run-from-a-running-loop, the caller's thread-local context not being disturbed, cell errors still surfacing unwrapped through the worker thread, the `await app.embed()` path, and an async app running another async app from one of its cells. All 5 fail without the fix.
- 6 new tests in `tests/_utils/test_asyncio_utils.py` covering both branches of the helper, that the no-loop path stays on the calling thread, exception propagation, and nesting.
- `tests/_ast`, `tests/_utils`, `tests/_runtime`: 1642 passed, 0 failed.
- `make check` passes (frontend typecheck/lint, Python lint/format/typecheck). Its `update-lock` step fails locally only because `pixi` isn't installed here; no lock files are touched by this change.

Pre-existing and unrelated: `tests/_export/test_exporter.py` has 20 failures on `main`, identical with and without this patch.

## 📋 Pre-Review Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable). — approach posted on [#9572](https://github.com/marimo-team/marimo/issues/9572#issuecomment-5246604691) before starting.

## ✅ Merge Checklist

- [x] Documentation has been updated where applicable, including docstrings for API changes. — no public API change; new helper and method are documented with docstrings.

### DIFF
--- a/marimo/_runtime/app/script_runner.py
+++ b/marimo/_runtime/app/script_runner.py
@@ -152,15 +152,13 @@ class AppScriptRunner:
         context: RuntimeContext,
         post_execute_hooks: list[Callable[[], Any]],
     ) -> RunOutput:
-        """`_run_asynchronous`, but safe to await on any thread.
+        """Run the app's cells asynchronously under `context`.
 
-        `run_coroutine_blocking` hands this coroutine to a worker thread when
-        the caller already has a running event loop, which is what happens
-        when an async notebook is executed from inside another notebook via
-        `app.run()` or `await app.embed()`. The runtime context is
-        thread-local, so reinstall it when the body lands on a thread that
-        doesn't have one; on the common path it is already installed and this
-        is a no-op.
+        The runtime context is thread-local. If the awaiting thread has no
+        context installed, `context` is installed for the duration of the run
+        and torn down afterwards; a thread that already has one is left
+        untouched. This makes the coroutine safe to await on a thread other
+        than the one `context` was created on.
         """
         installed_context = False
         if not runtime_context_installed():
--- a/marimo/_runtime/app/script_runner.py
+++ b/marimo/_runtime/app/script_runner.py
@@ -1,7 +1,6 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
-import asyncio
 from typing import TYPE_CHECKING, Any
 
 from marimo._ast.names import SETUP_CELL_NAME
@@ -11,6 +10,7 @@ from marimo._runtime import dataflow
 from marimo._runtime.app.common import RunOutput
 from marimo._runtime.context.types import (
     get_context,
+    initialize_context,
     runtime_context_installed,
     teardown_context,
 )
@@ -32,11 +32,13 @@ from marimo._runtime.patches import (
 from marimo._runtime.runner.result import RunResult
 from marimo._runtime.runner.scheduler import SequentialScheduler
 from marimo._types.ids import CellId_t
+from marimo._utils.asyncio_utils import run_coroutine_blocking
 
 if TYPE_CHECKING:
     from collections.abc import Callable
 
     from marimo._ast.app import InternalApp
+    from marimo._runtime.context.types import RuntimeContext
 
 
 class AppScriptRunner:
@@ -74,7 +76,7 @@ class AppScriptRunner:
 
     # _run_synchronous and _run_asynchronous are deliberate near-twins:
     # the only difference is the await on the cell step. Keeping them
-    # as separate methods (rather than wrapping with asyncio.run
+    # as separate methods (rather than driving an event loop
     # unconditionally) preserves the no-event-loop guarantee for purely
     # synchronous apps.
     def _run_synchronous(
@@ -145,6 +147,32 @@ class AppScriptRunner:
                             hook()
         return outputs, glbls
 
+    async def _run_asynchronous_in_context(
+        self,
+        context: RuntimeContext,
+        post_execute_hooks: list[Callable[[], Any]],
+    ) -> RunOutput:
+        """`_run_asynchronous`, but safe to await on any thread.
+
+        `run_coroutine_blocking` hands this coroutine to a worker thread when
+        the caller already has a running event loop, which is what happens
+        when an async notebook is executed from inside another notebook via
+        `app.run()` or `await app.embed()`. The runtime context is
+        thread-local, so reinstall it when the body lands on a thread that
+        doesn't have one; on the common path it is already installed and this
+        is a no-op.
+        """
+        installed = not runtime_context_installed()
+        if installed:
+            initialize_context(context)
+        try:
+            return await self._run_asynchronous(
+                post_execute_hooks=post_execute_hooks,
+            )
+        finally:
+            if installed:
+                teardown_context()
+
     def _handle_run_result(
         self,
         cid: CellId_t,
@@ -212,8 +240,6 @@ class AppScriptRunner:
             from marimo._output.formatting import FORMATTERS
 
             if not FORMATTERS.is_empty():
-                from marimo._runtime.context import get_context
-
                 register_formatters(
                     theme=get_context().marimo_config["display"]["theme"]
                 )
@@ -225,8 +251,9 @@ class AppScriptRunner:
                 post_execute_hooks.append(close_figures)
 
             if is_async:
-                outputs, defs = asyncio.run(
-                    self._run_asynchronous(
+                outputs, defs = run_coroutine_blocking(
+                    self._run_asynchronous_in_context(
+                        get_context(),
                         post_execute_hooks=post_execute_hooks,
                     )
                 )
--- a/marimo/_runtime/app/script_runner.py
+++ b/marimo/_runtime/app/script_runner.py
@@ -162,15 +162,16 @@ class AppScriptRunner:
         doesn't have one; on the common path it is already installed and this
         is a no-op.
         """
-        installed = not runtime_context_installed()
-        if installed:
+        installed_context = False
+        if not runtime_context_installed():
             initialize_context(context)
+            installed_context = True
         try:
             return await self._run_asynchronous(
                 post_execute_hooks=post_execute_hooks,
             )
         finally:
-            if installed:
+            if installed_context:
                 teardown_context()
 
     def _handle_run_result(
--- a/marimo/_utils/asyncio_utils.py
+++ b/marimo/_utils/asyncio_utils.py
@@ -7,13 +7,17 @@
   where the loop's default handler would otherwise swallow them).
 - `cancel_and_wait`: the `task.cancel(); await task` /
   `except CancelledError` dance, in one place.
+- `run_coroutine_blocking`: `asyncio.run` that also works when the
+  calling thread already has a running loop.
 """
 
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
 import contextlib
 import sys
+import threading
 from typing import TYPE_CHECKING, Any, TypeVar
 
 from marimo import _loggers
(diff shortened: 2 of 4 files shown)


marimo/_runtime/app/script_runner.py
```diff
@@ -145,6 +147,32 @@ async def _run_asynchronous(
...
+        installed = not runtime_context_installed()
+        if installed:
+            initialize_context(context)
+        try:
+            return await self._run_asynchronous(
+                post_execute_hooks=post_execute_hooks,
+            )
+        finally:
+            if installed:
+                teardown_context()
```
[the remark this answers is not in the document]
Chinmayrawat15: Good catch, fixed in 2b5be37.

Rather than renaming the inverted flag, I matched the convention `run()` already uses a few lines below (`installed_script_context`): initialize `installed_context = False`, then set it to `True` only after `initialize_context()` returns.

That also closes a small wrinkle the original had — the flag was set before the install actually succeeded, so if `initialize_context()` raised, the `finally` would have called `teardown_context()` for a context this method never installed.

marimo/_utils/asyncio_utils.py
```diff
@@ -147,9 +151,61 @@ async def cancel_and_wait(task: asyncio.Task[Any]) -> None:
         await task
 
 
+def run_coroutine_blocking(
```
dmadisetti, line 154: ~This isn't actually used in source, do you think you could move it to `./tests`?~

Nvm, missed the call :)

marimo/_runtime/app/script_runner.py
```diff
@@ -145,6 +147,33 @@ async def _run_asynchronous(
...
+        """`_run_asynchronous`, but safe to await on any thread.
+
+        `run_coroutine_blocking` hands this coroutine to a worker thread when
+        the caller already has a running event loop, which is what happens
+        when an async notebook is executed from inside another notebook via
+        `app.run()` or `await app.embed()`. The runtime context is
+        thread-local, so reinstall it when the body lands on a thread that
+        doesn't have one; on the common path it is already installed and this
+        is a no-op.
+        """
```
dmadisetti: Prefer a doc string that says what the function does, not what handles it (comments like those can go stale)

Chinmayrawat15: Fair point — that whole paragraph was about how the caller uses this and in which scenarios, which is exactly what rots when callers move. Rewritten in 9b5777a to describe the behaviour instead:

```
"""Run the app's cells asynchronously under `context`.

The runtime context is thread-local. If the awaiting thread has no
context installed, `context` is installed for the duration of the run
and torn down afterwards; a thread that already has one is left
untouched. This makes the coroutine safe to await on a thread other
than the one `context` was created on.
"""
```

Chinmayrawat15:
Pushed 9b5777a addressing the docstring feedback — it now describes what `_run_asynchronous_in_context` does (runs the cells under `context`, installing it only when the awaiting thread lacks one) rather than how the caller drives it. Resolved both threads; the `run_coroutine_blocking` one was already self-resolved.

Two notes on CI, since the checks currently look worse than they are:

- The earlier `enforce-label` failure was from before the `bug` label was added — it has not had a chance to re-run green since.
- The push put every workflow back into `action_required`, so they need another "Approve and run workflows" whenever you get a chance.

For reference, on the last approved run the full matrix was green (ubuntu/macOS/Windows, Py 3.10–3.13, core/optional/minimal deps, Pyodide, CLI, lint + typecheck); `enforce-label` was the only red.

marimo/_utils/asyncio_utils.py
```diff
@@ -147,9 +151,61 @@ async def cancel_and_wait(task: asyncio.Task[Any]) -> None:
...
+    # exit. The Future also re-raises with the worker's traceback intact.
+    future: concurrent.futures.Future[T] = concurrent.futures.Future()
+
+    def _target() -> None:
+        try:
+            future.set_result(_run())
+        except BaseException as exc:
+            future.set_exception(exc)
+
+    threading.Thread(target=_target, name=thread_name, daemon=True).start()
```
dmadisetti, line 201: This may cause issues in `pyodide`. I think we can use our own thread primitives and guard with is_pyodide().

dmadisetti: But also don't see how this could be invoked fro pyodide. I think this is fine

dmadisetti (approved):
Sorry, somehow missed the update. This looks good to me

PR #10511, opened 2026-08-11
fix(ui): raise on colliding option names in dropdown/multiselect

### LINKED ISSUE #10510: mo.ui.dropdown and mo.ui.multiselect silently drop options whose names collide
### Describe the bug

`mo.ui.dropdown` and `mo.ui.multiselect` build their options from a `{str_name: value}` mapping. When two options map to the **same string name**, the later one silently overwrites the earlier one — so you get **fewer options than you passed, with no error or warning**.

This happens in two everyday situations:

1. **Duplicate names** — e.g. `["apple", "apple", "banana"]` collapses to 2 options.
2. **Distinct values that stringify to the same name** — e.g. `[1, "1"]` (an `int` and a `str`) collapse to a single `"1"` option, silently losing one.

`mo.ui.radio` does **not** have this problem: it raises a clear error (`"A radio group cannot have repeated options."`). So the three sibling components are inconsistent, and two of them lose user data silently.

### Code to reproduce

```python
import marimo as mo

# 1) Three DISTINCT values -> dropdown keeps only two
d = mo.ui.dropdown(options=[1, "1", 2], label="I passed 3 options")
d
```

```python
print("passed: 3   ->   kept:", len(d.options), "  ->  ", list(d.options))
# passed: 3   ->   kept: 2   ->   ['1', '2']
```

```python
# 2) Plain duplicate names -> the menu shows only two
mo.ui.dropdown(options=["apple", "apple", "banana"], label="I passed 3, will show 2")
```

```python
# 3) radio, for contrast, refuses to lose data:
mo.ui.radio(options=["apple", "apple", "banana"])
# ValueError: A radio group cannot have repeated options.
```

### Screenshot

> _Screenshot to be added below (shows `passed: 3 -> kept: 2 -> ['1', '2']` and the open apple/banana dropdown listing only two options)._

Image

### Expected behavior

Passing options that collide should not silently discard data. Consistent with `mo.ui.radio`, `dropdown` and `multiselect` should surface the problem — for example, raise a clear `ValueError` naming the colliding option(s) — rather than quietly dropping them.

(If silent de-duplication is intended for some cases, at minimum it should warn, and the behavior should be documented and consistent across the three components.)

### Root cause

In `marimo/_plugins/ui/_impl/input.py`, both `dropdown` and `multiselect` construct their option map with a dict comprehension roughly like:

```python
options = {_to_option_name(option): option for option in options}
```

A dict comprehension keeps only the last entry for any repeated key, so any two options whose `_to_option_name(...)` is equal collapse into one — with no error. `radio` guards against this explicitly, which is why it behaves differently.

### Suggested fix

Detect colliding option names in `dropdown` / `multiselect` and raise a helpful error (mirroring `radio`), so option handling is consistent across all three components. This is a natural follow-up to #10498, which recently made the *invalid-value* error consistent across the same three components.

Happy to submit a PR. Would you prefer a hard error (like `radio`), or a non-fatal warning that keeps de-duplicating? I'll implement whichever you'd like.

### Environment

```
{
  "marimo": "0.23.16",
  "OS": "Darwin",
  "OS Version": "24.6.0",
  "Processor": "arm",
  "Python Version": "3.13.0"
}
```

### Will you submit a PR?

- [x] Yes

Abdulrehman-PIAIC80387:

## 📝 Summary

Closes #10510

`mo.ui.dropdown` and `mo.ui.multiselect` silently drop options whose display names collide, so the element ends up with **fewer options than the user passed — no error, no warning**:

```python
mo.ui.dropdown(options=["apple", "apple", "banana"])  # → 2 options, one "apple" lost
mo.ui.dropdown(options=[1, "1", 2])                    # → 2 options; int 1 and str "1" both map to "1"
```

`mo.ui.radio` already guards against this. This makes `dropdown`/`multiselect` consistent.

## 🔍 Root cause

Both build their option map with a dict comprehension keyed on `_to_option_name(...)`:

```python
options = {_to_option_name(option): option for option in options}
```

A dict keeps only the last entry per key, so colliding names silently overwrite each other.

## 🔧 What changed

- Add `_validate_unique_option_names` (next to the `_validate_option_name` helper from #10498) and call it in both constructors **before** the map is built, raising a helpful error that names the colliding options.
- Only the sequence path is guarded; explicit `dict` inputs (keys already unique) are untouched.
- Mirrors `radio`'s existing guard — a natural follow-up to #10498, which made the *invalid-value* error consistent across the same three components.

## 🎥 Before / After

**Before** — silent data loss (3 options in, 2 kept):

image

**After** — a clear, actionable error:

image

## 🧪 Testing

- Added collision tests for both components (duplicate names, int/str collision, and that non-colliding + `dict` inputs are unaffected).
- Verified by mutation: removing the two guard calls makes the new tests fail.
- `tests/_plugins/ui/_impl/test_input.py` — 52 passed, no regressions.
- `mypy` and `ruff` clean.
- Verified manually in `make dev`.

## ❓ One question for maintainers

The issue (#10510) asked whether to **raise** (implemented here, matching `radio`) or **warn and keep de-duplicating**. I went with raise for consistency; happy to switch to a non-fatal `LOGGER.warning` if you'd prefer — it's a small change. Opening as a **draft** pending your preference.

## 📋 Pre-Review Checklist

- [ ] For large changes, or changes that affect the public API: raised in #10510; awaiting maintainer preference on raise-vs-warn.

## ✅ Merge Checklist

- [x] Documentation has been updated where applicable — behavior is self-evident from the error; no API/docs change.

### DIFF
--- a/marimo/_plugins/ui/_impl/input.py
+++ b/marimo/_plugins/ui/_impl/input.py
@@ -5,7 +5,7 @@ import base64
 import dataclasses
 import sys
 import traceback
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass
 from typing import (
     Any,
@@ -962,24 +962,27 @@ def _validate_option_name(option_name: str, options: dict[str, Any]) -> None:
         )
 
 
-def _validate_unique_option_names(options: Sequence[Any]) -> None:
-    """Raise a helpful error if two options map to the same display name.
+def _build_option_map(options: Iterable[Any]) -> dict[str, Any]:
+    """Build the `{display name: option}` map, raising if two names collide.
 
-    `dropdown` and `multiselect` key their option map on `_to_option_name(...)`;
-    without this check, colliding names would silently overwrite each other and
+    `dropdown` and `multiselect` key their option map on `_to_option_name(...)`.
+    Without this check, colliding names would silently overwrite each other and
     drop options (e.g. `["a", "a"]`, or `[1, "1"]` where an `int` and a `str`
-    both map to `"1"`).
+    both map to `"1"`). Validating and building in a single pass avoids
+    consuming a single-use iterable twice and calls `_to_option_name` once per
+    option.
     """
-    seen: dict[str, Any] = {}
+    result: dict[str, Any] = {}
     for option in options:
         name = _to_option_name(option)
-        if name in seen:
+        if name in result:
             raise ValueError(
-                f"Duplicate option name '{name}': options {seen[name]!r} and "
+                f"Duplicate option name '{name}': options {result[name]!r} and "
                 f"{option!r} both map to the same name. "
                 "Option names must be unique."
             )
-        seen[name] = option
+        result[name] = option
+    return result
 
 
 @mddoc
@@ -1063,8 +1066,7 @@ class dropdown(UIElement[list[str], Any]):
             searchable = True
 
         if not isinstance(options, dict):
-            _validate_unique_option_names(options)
-            options = {_to_option_name(option): option for option in options}
+            options = _build_option_map(options)
 
             if value is not None and not isinstance(value, str):
                 value = _to_option_name(value)
@@ -1186,8 +1188,7 @@ class multiselect(UIElement[list[str], list[object]]):
             )
 
         if not isinstance(options, dict):
-            _validate_unique_option_names(options)
-            options = {_to_option_name(option): option for option in options}
+            options = _build_option_map(options)
 
             if value is not None and not isinstance(value, str):
                 value = [_to_option_name(v) for v in value]
--- a/marimo/_plugins/ui/_impl/input.py
+++ b/marimo/_plugins/ui/_impl/input.py
@@ -962,6 +962,26 @@ def _validate_option_name(option_name: str, options: dict[str, Any]) -> None:
         )
 
 
+def _validate_unique_option_names(options: Sequence[Any]) -> None:
+    """Raise a helpful error if two options map to the same display name.
+
+    `dropdown` and `multiselect` key their option map on `_to_option_name(...)`;
+    without this check, colliding names would silently overwrite each other and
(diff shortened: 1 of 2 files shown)


marimo/_plugins/ui/_impl/input.py
```diff
@@ -962,6 +962,26 @@ def _validate_option_name(option_name: str, options: dict[str, Any]) -> None:
...
+    seen: dict[str, Any] = {}
+    for option in options:
+        name = _to_option_name(option)
+        if name in seen:
+            raise ValueError(
+                f"Duplicate option name '{name}': options {seen[name]!r} and "
+                f"{option!r} both map to the same name. "
+                "Option names must be unique."
+            )
+        seen[name] = option
```
manzt: Could this helper build and return the option map in one pass? The current validate-then-build flow iterates options twice and calls `_to_option_name` twice. This could exhaust sized single-pass iterables.

Something like:

```py
def _build_option_map(options: Iterable[Any]) -> dict[str, Any]:
    result: dict[str, Any] = {}
    for option in options:
        name = _to_option_name(option)
        if name in result:
            raise ValueError(...)
        result[name] = option
    return result
```

Then both constructors could use `options = _build_option_map(options)`.

manzt (changes requested):
I think this looks good overall. Raising on collisions is clearer than silently dropping options and is consistent with radio.

marimo/_plugins/ui/_impl/input.py
Abdulrehman-PIAIC80387: Good call — done in 790c189f5. Replaced the validate-then-build flow with a single `_build_option_map(options)` helper that validates and constructs the map in one pass (one `_to_option_name` call per option, and safe for single-use iterables). Both `dropdown` and `multiselect` now use `options = _build_option_map(options)`. Tests + mypy + ruff green.

manzt (approved):
Thanks for the contribution!

PR #10512, opened 2026-08-11
feat: make editor code lens configurable via display.code_lens

mscolnick:
Follow-up to #10507: the code lens (inline icons in cell editors linking
datasources, storage buckets, and caches to their panels) is now gated by
a display.code_lens user config, on by default, with a settings toggle
and command palette actions.


manzt (dismissed):
Looks good. Do we want to add something to the docs? I'm guessing the api work is auto-documented and probably enough. We can explain more after we roll out.

mscolnick:
@manzt good call. maybe in a followup or after release ill update the docs

PR #10513, opened 2026-08-11
feat(frontend): generate OpenAPI schema for UI plugin contracts

mscolnick:
Add a test-driven generator that builds an OpenAPI spec from the plugin
registries, checks it in as plugins.openapi.yaml, and wires up a CI
compatibility check that flags breaking plugin API changes.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -179,7 +179,7 @@
     "lint:oxlint": "oxlint --fix",
     "lint:stylelint": "stylelint src/**/*.css --fix",
     "preview": "vite preview",
-    "plugins:generate-schema": "UPDATE_PLUGIN_SCHEMA=1 vitest run src/plugins/__tests__/plugin-schema.test.ts",
+    "plugins:generate-schema": "cross-env UPDATE_PLUGIN_SCHEMA=1 vitest run src/plugins/__tests__/plugin-schema.test.ts",
     "dev:quarto": "VITE_MARIMO_ISLANDS=true vite",
     "dev:islands": "cross-env VITE_MARIMO_ISLANDS=true vite --config islands/vite.config.mts",
     "build:islands": "cross-env VITE_MARIMO_ISLANDS=true vite --config islands/vite.config.mts build",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -179,6 +179,7 @@
     "lint:oxlint": "oxlint --fix",
     "lint:stylelint": "stylelint src/**/*.css --fix",
     "preview": "vite preview",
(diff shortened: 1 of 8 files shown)


mscolnick:
@cubic-dev-ai review

frontend/src/plugins/__tests__/plugin-schema.test.ts
```diff
@@ -0,0 +1,328 @@
...
+    });
+  });
+
+  it("contains no dangling Zod definitions", () => {
+    expect(JSON.stringify(document)).not.toContain("#/$defs");
+  });
+});
+
+describe("buildPluginSpec", () => {
+  it("emits accepted input shapes for defaults and transforms", () => {
```
kirangadhave, line 246: we don't assert for `default` in the test

frontend/src/plugins/__tests__/plugin-schema.ts
```diff
@@ -0,0 +1,396 @@
...
+  }
+  if (Array.isArray(value)) {
+    return value.map((item) => rewriteSharedReferences(item, componentName));
+  }
+  if (value === null || typeof value !== "object") {
+    return value;
+  }
+  return Object.fromEntries(
+    Object.entries(value).map(([key, item]) => [
+      key,
```
kirangadhave, line 109: since this is for rewriting references only; should we only pass in `$ref` keys?

frontend/src/plugins/__tests__/plugin-schema.ts
```diff
@@ -0,0 +1,396 @@
...
+interface FunctionContract {
+  input: ZodType;
+  output: ZodType;
+}
+
+export interface PluginContract {
+  tagName: string;
+  validator: ZodType;
+  functions?: Record<string, FunctionContract>;
+}
```
kirangadhave: We can derive this from `IPlugin`


```suggestion
export type PluginContract = Pick<IPlugin<any, unknown>, "tagName" | "validator" | "functions">
```

frontend/package.json
```diff
@@ -179,6 +179,7 @@
     "lint:oxlint": "oxlint --fix",
     "lint:stylelint": "stylelint src/**/*.css --fix",
     "preview": "vite preview",
+    "plugins:generate-schema": "cross-env UPDATE_PLUGIN_SCHEMA=1 vitest run src/plugins/__tests__/plugin-schema.test.ts",
```
kirangadhave, line 182: we should add a `make` target for this

kirangadhave (commented):
this is great. just some minor comments

kirangadhave (approved):

PR #10326, opened 2026-07-25
fix(table): preserve null extension-array values in JSON (#10299)

hsusul:
**This pull request was authored by a coding agent.**

## 📝 Summary

Closes marimo-team/marimo#10299

### Problem

When a pandas DataFrame column contains an extension array (such as `pint-pandas`) with missing values (`pd.NA`, `None`, `np.nan`), `PandasTableManager._to_json_str` calls `_extension_column_needs_stringify(result[col])` and unconditionally performs `result[col] = result[col].astype(str)`. This stringifies missing cells into literal text like `"<NA> meter"` (or `"nan"`) instead of JSON `null`.

### Fix

Convert valid non-missing cells in extension columns to string (`astype(str)`), while using `.astype(object).where(series.notna(), None)` so missing entries remain `None` and serialize to JSON `null` as expected.

## 📋 Pre-Review Checklist

- [X] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](<https://marimo.io/discord?ref=pr>), or the community [discussions](<https://github.com/marimo-team/marimo/discussions>) (Please provide a link if applicable).
- [X] Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it.
- [ ] Video or media evidence is provided for any visual changes (optional).

## ✅ Merge Checklist

- [X] I have read the [contributor guidelines](<https://github.com/marimo-team/marimo/blob/main/CONTRIBUTING.md>).
- [X] Documentation has been updated where applicable, including docstrings for API changes.
- [X] Tests have been added for the changes made.

## Testing

* `uv run --group test pytest tests/_plugins/ui/_impl/tables/test_pandas_table.py` — Passed (122 passed).
* `uv run ruff check marimo/ tests/` — Passed with 0 errors.
* `uv run ruff format --check marimo/ tests/` — Passed with 0 issues.

PR #10517, opened 2026-08-12
fix(table): preserve nulls in pandas JSON stringification

### LINKED ISSUE #10299: Missing pint-pandas values render as text ("<NA> meter") in table JSON instead of null
### Describe the bug

When a pint-pandas (extension-array) column contains missing values, the pandas table renderer serializes those missing cells as literal strings such as `"<NA> meter"` (or `"nan"`) rather than JSON `null`. The equivalent value crossing the shared msgspec transport boundary is emitted as JSON `null`. As a result, the same logical missing value is represented inconsistently depending on which serialization path it takes, so equivalent rows can compare and render differently, and downstream consumers cannot reliably detect a missing pint value.

**Where:** `marimo/_plugins/ui/_impl/tables/pandas_table.py`, in `PandasTableManager._to_json_str` — https://github.com/marimo-team/marimo/blob/e17dd29d24054eabd3b638fdd7bf28bac750350a/marimo/_plugins/ui/_impl/tables/pandas_table.py#L298-L302

```python
if _extension_column_needs_stringify(result[col]):
    # Extension arrays with rich Python values (e.g.
    # pint-pandas) serialize to nested dicts via
    # to_dict; stringify to preserve display.
    result[col] = result[col].astype(str)
```

`astype(str)` is applied to the whole column, so missing cells (`pd.NA` / `NaN`) are converted to text. Notably, `_extension_column_needs_stringify` already samples only a `notna()` cell, so missing entries are expected — but they are still stringified. For contrast, the shared encoder returns real `None` for missing values (https://github.com/marimo-team/marimo/blob/e17dd29d24054eabd3b638fdd7bf28bac750350a/marimo/_messaging/msgspec_encoder.py#L280-L281), which msgspec emits as JSON `null`.

### Code to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "marimo",
#     "pandas",
#     "pint",
#     "pint-pandas",
#     "msgspec",
# ]
# ///
import pandas as pd
import pint_pandas  # noqa: F401
from marimo._plugins.ui._impl.tables.pandas_table import PandasTableManager

# One valid pint quantity, one MISSING pint value
df = pd.DataFrame({"length": pd.array([1.0, None], dtype="pint[meter]")})

mgr = PandasTableManager(df)
print(mgr.to_json_str())
# Missing cell serializes as "<NA> meter" instead of null
```

### Expected vs actual

- **Expected:** the missing pint value serializes to JSON `null`, consistent with the transport/msgspec boundary and with how other missing values (NaN/NA) are treated.
- **Actual:** the missing value serializes to the string `"<NA> meter"` (mixed-unit columns can also yield `"nan"`), so missing data is indistinguishable from real text and diverges from the transport representation.

### Suggested fix

Stringify only non-missing cells and preserve missingness, e.g. apply the string cast per element via a `notna()` mask (or `Series.where(series.notna(), other=pd.NA)`) instead of an unconditional `astype(str)` on the whole column, so missing entries remain missing and serialize to `null`.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/d90fec46-588b-43a0-8f60-917df041ce1a.

kirangadhave:
## 📝 Summary

Closes marimo-team/marimo#10299

Preserve missing values as JSON `null` in every pandas table stringify branch. These branches previously converted missing values into strings such as `"NaT"` and `"None"`.

Extract a shared null-preserving string conversion, keep bytes display values, and reuse the extension-column null mask. The mask is skipped for complete columns. Add dependency-free coverage for complex, timedelta, date-object, bytes, and extension-array columns.

Supersedes marimo-team/marimo#10326 by @hsusul. That PR identified the extension-array null bug. This change includes that fix and applies it to the remaining branches.

### DIFF
--- a/marimo/_plugins/ui/_impl/tables/pandas_table.py
+++ b/marimo/_plugins/ui/_impl/tables/pandas_table.py
@@ -130,6 +130,20 @@ def _resolve_index_column_conflicts(df: pd.DataFrame) -> pd.DataFrame:
     return df
 
 
+def _stringify_preserving_nulls(
+    series: pd.Series[Any],
+    notna: pd.Series[bool] | None = None,
+) -> pd.Series[Any]:
+    """Convert values to strings and preserve missing values."""
+    if notna is None:
+        notna = series.notna()
+
+    stringified = series.apply(str)
+    if not notna.all():
+        stringified = stringified.astype(object).where(notna, None)
+    return stringified
+
+
 def _extension_column_needs_stringify(series: pd.Series[Any]) -> bool:
     """Whether an extension-array column should be cast to str for JSON."""
     from pandas.api.types import is_extension_array_dtype
--- a/marimo/_plugins/ui/_impl/tables/pandas_table.py
+++ b/marimo/_plugins/ui/_impl/tables/pandas_table.py
@@ -144,15 +144,19 @@ def _stringify_preserving_nulls(
     return stringified
 
 
-def _extension_column_needs_stringify(series: pd.Series[Any]) -> bool:
-    """Whether an extension-array column should be cast to str for JSON."""
+def _extension_column_needs_stringify(
+    series: pd.Series[Any],
+    notna: pd.Series[bool] | None = None,
+) -> bool:
+    """Whether an extension-array column needs a string cast for JSON."""
     from pandas.api.types import is_extension_array_dtype
 
     try:
         if not is_extension_array_dtype(series.dtype):
             return False
 
-        notna = series.notna()
+        if notna is None:
+            notna = series.notna()
         if not notna.any():
             return False
 
@@ -295,6 +299,7 @@ class PandasTableManagerFactory(TableManagerFactory):
 
                 from pandas.api.types import (
                     is_complex_dtype,
+                    is_extension_array_dtype,
                     is_object_dtype,
                     is_timedelta64_dtype,
                     is_timedelta64_ns_dtype,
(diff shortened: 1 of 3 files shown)


Light2Dark:
One of my concerns on the original PR was on the performance impact, would this be a performant solution?

kirangadhave:
> One of my concerns on the original PR was on the performance impact

I ran some performance tests with Claude

<details>

<summary>
PR 10517 performance report
</summary>

## Result

The PR adds no measurable end-to-end regression at one million rows on this machine.

The five-column frame changed by `+0.02%` without nulls. The `pint-pandas` frame changed by less than `1%` at one million rows.

## Environment

- Apple M4 Pro, 48 GB memory
- macOS 26.6.1
- Python 3.13.12
- pandas 3.0.5
- NumPy 2.5.1
- pint-pandas 0.8.0

Each result is the median of three runs after one warm-up run. Times include `PandasTableManager.to_json_str()` and JSON encoding.

The comparison uses the same process and dataframe. The old path uses `apply(str)` for existing branches and `astype(str)` for extension arrays.

## Five-column frame

This frame contains complex, timedelta, date-object, bytes, and categorical extension columns.

| Rows | Nulls | Old path | PR path | Change |
|---:|---:|---:|---:|---:|
| 10,000 | 0% | 44.6 ms | 45.4 ms | +1.7% |
| 100,000 | 0% | 484.5 ms | 486.8 ms | +0.5% |
| 1,000,000 | 0% | 4.967 s | 4.968 s | +0.02% |
| 10,000 | 1% | 47.3 ms | 37.4 ms | -21.0% |
| 100,000 | 1% | 481.0 ms | 398.2 ms | -17.2% |
| 1,000,000 | 1% | 4.941 s | 4.112 s | -16.8% |

The null-containing PR path is faster because it restores `None` before later JSON sanitation. It also emits the required JSON `null` values.

## `pint-pandas` extension column

| Rows | Nulls | Old path | PR path | Change |
|---:|---:|---:|---:|---:|
| 10,000 | 0% | 94.2 ms | 97.1 ms | +3.1% |
| 100,000 | 0% | 965.5 ms | 980.1 ms | +1.5% |
| 1,000,000 | 0% | 10.224 s | 10.152 s | -0.7% |
| 10,000 | 1% | 94.7 ms | 94.4 ms | -0.4% |
| 100,000 | 1% | 975.3 ms | 987.5 ms | +1.2% |
| 1,000,000 | 1% | 10.236 s | 10.199 s | -0.4% |

## Conversion-only detail

The `pint-pandas` string conversion alone varied more than the full operation. At one million rows, `apply(str)` measured approximately `18%` slower than `astype(str)` in one sequential microbenchmark.

That difference did not appear in the end-to-end measurements. Dataframe assignment, record conversion, JSON encoding, and null sanitation dominate the full download path.

## Conclusion

The current implementation is suitable for full-frame downloads through one million rows. The largest observed end-to-end increase was `3.1%` at 10,000 no-null `pint-pandas` rows.

At 100,000 rows and more, the increase stayed at `1.5%` or less. The one-million-row results were neutral within normal benchmark variation.

These results are local microbenchmarks, not cross-platform guarantees. CI performance measurements can detect future changes if this path becomes a recurring concern.
</details>

Light2Dark (approved):

PR #10523, opened 2026-08-12
feat(cache): formalize caching config and add signing fields

dmadisetti:
## 📝 Summary

Formalizes the caching config entrypoints, and adds "signing" fields for cache verification.
This PR introduces the bones for signature verification that will come in a subsequent change.

Cache configuration moves out of `experimental` into a top-level `[cache]` table, and a `[signing]` table is added for trust and identity:

```toml
[cache]
verification = "on"          # off | on | strict

[cache.store]
type = "file"
args = { save_path = "~/.cache/marimo" }

# Or a tiered store:
# [[cache.store]]
# type = "file"
# [[cache.store]]
# type = "redis"

[signing]
private_key_path = "~/.marimo/key.pem"

[signing.trusted_signers]
"SHA256:kV9x2c...q8" = "CI cache key"
"SHA256:abc9f1...4d" = "Alice"
```

`[cache].verification` and `[signing]` are declared and merged here, with followup providing implementation.

### Breaking

`experimental.cache` is removed for the explicit cache path.

### Notes

- `signing.trusted_signers` is a **replace** path in `merge_config`, not a deep merge. Deep-merging unions the fingerprints from every layer, and then no layer can remove a signer that a lower-priority one anchored.
- `signing.private_key_path` joins the secrets mask, so this machine's signing identity is never serialized to the frontend.

### Stack

1. **this PR** — config entrypoints + signing fields
2. #10524 — `SigningPolicy`, `mode` -> `verification`
3. #10525 — config sanitization
4. #10526 — docs

### DIFF
--- a/marimo/_config/config.py
+++ b/marimo/_config/config.py
@@ -592,13 +592,41 @@ class SharingConfig(TypedDict):
 
 @dataclass
 class StoreConfig(TypedDict, total=False):
-    """Configuration for cache stores."""
+    """Configuration for a single cache store."""
 
     type: StoreKey
     args: dict[str, Any]
 
 
-CacheConfig = list[StoreConfig] | StoreConfig
+# One store, or a list composed into a TieredStore.
+CacheStoreConfig = list[StoreConfig] | StoreConfig
+
+CacheVerification = Literal["off", "on", "strict"]
+
+
+class CacheConfig(TypedDict, total=False):
+    """Configuration for caching.
+
+    `verification` is the signature-checking posture; `store` is the backing
+    store, or a list of stores composed into a `TieredStore`.
+    """
+
+    verification: CacheVerification
+    store: CacheStoreConfig
+
+
+class SigningConfig(TypedDict, total=False):
+    """Cache-signing trust and identity.
+
+    `trusted_signers` maps a key fingerprint (`"SHA256:<base64>"`) to an
+    advisory label. Trusting a key allows arbitrary code execution from its
+    holder on this machine — a cache restore is `pickle.loads` — so there is no
+    lesser cache-only grant. `private_key_path` is this machine's signing
+    identity; it is never serialized to the frontend.
+    """
+
+    trusted_signers: dict[str, str]
+    private_key_path: str
 
 
 class ExperimentalConfig(TypedDict, total=False):
@@ -616,7 +644,6 @@ class ExperimentalConfig(TypedDict, total=False):
     line_timing: bool  # Active-line highlight + per-line timer (sys.settrace)
 
     # Internal features
-    cache: CacheConfig
     execution_type: ExecutionType
 
 
@@ -648,6 +675,8 @@ class MarimoConfig(TypedDict):
     sharing: NotRequired[SharingConfig]
     mcp: NotRequired[MCPConfig]
     venv: NotRequired[VenvConfig]
+    cache: NotRequired[CacheConfig]
+    signing: NotRequired[SigningConfig]
 
 
 @mddoc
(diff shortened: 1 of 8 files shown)


marimo/_save/stores/__init__.py
```diff
@@ -65,7 +66,19 @@ def _get_store_from_config(
             return sub_stores[0]
         return TieredStore(sub_stores)
     else:
-        store_type = cast(StoreKey, config.get("store", DEFAULT_STORE_KEY))
+        # `type` is the documented key. The legacy `experimental.cache` reader
+        # spelled it `store`, so accept that too rather than silently falling
+        # back to the default store — a silent cache miss is the worst outcome
+        # here, since the cache still "works" while never hitting.
+        store_type = cast(StoreKey, config.get("type") or DEFAULT_STORE_KEY)
```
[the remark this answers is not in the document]
dmadisetti: Resolved by removing the legacy path rather than fixing it. `experimental.cache` and the `store` spelling are both gone — `type` is the only key, read with `config.get("type", DEFAULT_STORE_KEY)`, so a present-but-falsy value now reaches the `Invalid store type` validation and log instead of being folded into the default.

marimo/_save/stores/__init__.py
```diff
@@ -65,7 +66,19 @@ def _get_store_from_config(
...
+        if "type" not in config and "store" in config:
+            store_type = cast(StoreKey, config["store"])  # type: ignore[typeddict-item]
```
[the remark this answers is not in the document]
dmadisetti: Resolved by removing the legacy path rather than fixing it. `experimental.cache` and the `store` spelling are both gone — `type` is the only key, read with `config.get("type", DEFAULT_STORE_KEY)`, so a present-but-falsy value now reaches the `Invalid store type` validation and log instead of being folded into the default.

marimo/_save/stores/__init__.py
```diff
@@ -33,18 +33,15 @@
 def get_store(current_path: str | None = None) -> Store:
     from marimo._config.manager import get_default_config_manager
 
-    cache_config: CacheConfig | None = (
-        get_default_config_manager(current_path=current_path)
-        .get_config()
-        .get("experimental", {})
-        .get("cache", None)
+    config = get_default_config_manager(current_path=current_path).get_config()
+    store_config: CacheStoreConfig | None = config.get("cache", {}).get(
```
kirangadhave, line 37: existing usages of `experimental.cache` will break; should we add a warning where if we detect `experimental.cache` we log the new usage path?

kirangadhave (approved):
🚀

marimo/_save/stores/__init__.py
mscolnick: this was not advertised and only internal, and experimental. i'd prefer the hard break instead of an error-prone migration

marimo/_config/config.py
```diff
@@ -829,7 +860,12 @@ def merge_config(
     # Fields that should be replaced instead of merged.
     # These are "record" types where keys can be added/removed,
     # as opposed to config objects where you only set specific fields.
-    replace_paths = frozenset({"ai.custom_providers"})
+    # NB. `signing.trusted_signers` is replaced, not deep-merged. A deep merge
+    # unions the fingerprints from every layer. Then no layer can remove a
+    # signer that a lower-priority one anchored.
+    replace_paths = frozenset(
+        {"ai.custom_providers", "signing.trusted_signers"}
```
mscolnick, line 867: ~maybe include all of `cache` here? it could have urls~ nvm, i misunderstood this block

marimo/_config/secrets.py
```diff
@@ -51,6 +51,8 @@ def deep_remove_from_path(
         ("ai", "bedrock", "aws_access_key_id"),
         ("ai", "bedrock", "aws_secret_access_key"),
         ("runtime", "dotenv"),
+        # This machine's signing identity — never serialized to the frontend.
+        ("signing", "private_key_path"),
```
mscolnick, line 55: should we include `cache` here as well? in case there is anything sensitive in it?

dmadisetti: Cache itself is not sensitive, but the private key is. That's in the stack, so just getting this in

PR #10525, opened 2026-08-12
feat(cache): sanitize signing config from untrusted layers

dmadisetti:
## 📝 Summary

Adds config sanitization.

marimo merges configuration from several layers, and two of them travel with the code.

Trusting a signing key is a grant of arbitrary code execution, because a cache restore is
`pickle.loads`. So a layer that could name its own key as trusted, or turn verification off, could
ship a poisoned cache entry that runs on `git clone`.

```
user config ─┐
environment ─┼─→ merged effective config → SigningPolicy (frozen)
             │
pyproject ───┤
PEP 723 ─────┼─→ strip [signing] + [cache].verification
workspace ───┘   .marimo.toml
```

### Stack

1. #10523 — config entrypoints + signing fields
2. #10524 — `SigningPolicy`, `mode` → `verification`
3. **this PR** — config sanitization (untrusted layers)
4. — docs

### DIFF
--- a/marimo/_config/manager.py
+++ b/marimo/_config/manager.py
@@ -33,6 +33,7 @@ from marimo._config.reader import (
     read_marimo_config,
     read_pyproject_marimo_config,
     sanitize_pyproject_dict,
+    strip_untrusted_config,
 )
 from marimo._config.secrets import (
     mask_secrets,
@@ -42,6 +43,7 @@ from marimo._config.secrets import (
 from marimo._config.settings import GLOBAL_SETTINGS
 from marimo._config.utils import (
     get_or_create_user_config_path,
+    is_trusted_user_config_path,
 )
 from marimo._utils.env import env_to_value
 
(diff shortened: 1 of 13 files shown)


marimo/_config/reader.py
```diff
@@ -116,6 +121,44 @@ def allowlist_script_config(
     return pyproject_dict
 
 
+# Settings an untrusted-origin config layer must not set. A `trusted_signers`
```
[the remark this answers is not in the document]
dmadisetti: Applied — it now reads "Settings that an untrusted-origin config layer must not set." The restrictive `that` is also what the rest of these comments use, so this was an inconsistency as much as a readability issue.

marimo/_config/manager.py
```diff
@@ -528,6 +534,11 @@ def _load_config(self) -> MarimoConfig:
                 LOGGER.error("Failed to read user config at %s", path)
                 LOGGER.error(str(e))
                 return DEFAULT_CONFIG
+            if not is_trusted_user_config_path(path):
+                # A `.marimo.toml` discovered by walking up from the cwd is
+                # project-origin, not user-owned. Strip its cache-trust keys
+                # like any other untrusted layer.
+                strip_untrusted_config(user_config, is_user_layer=True)
```
kirangadhave: nit: all other callsites of `strip_untrusted_config` use the return value. do we want to do something like:
```python
_config = strip_untrusted_config(user_config, is_user_layer=True)
```

kirangadhave (dismissed):
🚀

marimo/_config/manager.py
dmadisetti:
```suggestion
                user_config = strip_untrusted_config(user_config, is_user_layer=True)
```

kirangadhave (approved):

PR #10526, opened 2026-08-12
docs(cache): document the cache signing trust model

dmadisetti:
## 📝 Summary

Docs for the cache signing trust model.

Adds a "Cache signing and trust" section to `docs/api/caching.md` covering the config surface, the
verification postures, and the sharing-between-machines workflow:

> **Trusting a key grants code execution.** A cache restore is `pickle.loads`, so trusting a
> fingerprint lets whoever holds that key run arbitrary code on your machine. There is no narrower
> "cache-only" version of this grant. Add a fingerprint only if you would run a script that person
> handed you.

### Stack

1. #10523 — config entrypoints + signing fields
2. #10524 — `SigningPolicy`, `mode` → `verification`
3. #10525 — config sanitization (untrusted layers)
4. **this PR** — docs


docs/api/caching.md
```diff
@@ -263,6 +263,82 @@ def my_decorator(func):
...
+## Cache signing and trust
+
+Restoring a persistent cache entry deserializes it with `pickle`, which runs
+whatever code the entry tells it to. So reading a cache someone else wrote means
+running code they chose.
+
+To keep that from happening by accident, marimo signs every cache entry it
+writes with an [Ed25519](https://ed25519.cr.yp.to/) key and checks the signature
+before restoring one. An entry that does not verify against a key you trust is
+recomputed rather than loaded.
```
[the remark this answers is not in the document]
dmadisetti: Fixed, and it was worse than a contradiction — the claim was simply false. Signing lives only on `LazyLoader`, and `persistent_cache` defaults to `method="pickle"` (`save.py:1197`), which neither signs nor verifies. So "marimo signs every cache entry it writes" plus "you do not need to configure anything" promised a guarantee that does not exist by default.

The section now states the scope up front: signing applies to `method="lazy"`, the default `"pickle"` carries no provenance at all, and the `off` row of the postures table no longer contradicts the opening. Three further factual errors went with it: there is no environment variable for `cache.verification`, a workspace `.marimo.toml` can no longer set `cache.store` either, and the no-`cryptography` claim omitted the same-origin WebAssembly exception.

docs/api/caching.md
```diff
@@ -263,6 +263,113 @@ def my_decorator(func):
...
 
 In this instance, the cache will work as expected because the decorated function has the same signature and metadata as the original function.
+
+## Cache signing and trust
+
+Restoring a persistent cache entry deserializes it with `pickle`, which runs
+whatever code the entry tells it to. Reading a cache someone else wrote therefore
+means running code they chose.
+
+The lazy loader defends against that. It signs each entry it writes with an
```
akshayka, line 273: Has `lazy` loader been introduced before in the documentation? You should define it when it is first mentioned. When I read this paragraph, I think "What lazy loader? What is a lazy loader?"

docs/api/caching.md
```diff
@@ -263,6 +263,113 @@ def my_decorator(func):
...
+# fingerprint = label. The label is a note to yourself; only the
+# fingerprint decides anything.
+"SHA256:kV9x2c...q8" = "CI cache key"
+"SHA256:abc9f1...4d" = "Alice"
+```
+
+Two environment variables set the same thing without a config file:
+`MARIMO_CACHE_SIGNING_PRIVATE_KEY` holds a signing identity, and
+`MARIMO_CACHE_SIGNING_PUBLIC_KEY` adds one trusted key. Both take the key itself
+in Privacy-Enhanced Mail (PEM) form, not a path.
```
akshayka, line 315: This is difficult for a user to understand. Instead of saying "Two environment variables set the same thing without …" you should frame this as instructive for a user, based on what they are trying to do.

For example, something like: "You can also <thing user is trying to achieve> using environment variables. Store XXX in the XXX environment variable, and XXX in the XXX environment variable. The environment variables expect keys in XXX form."

docs/api/caching.md
```diff
@@ -263,6 +263,113 @@ def my_decorator(func):
...
+
+!!! danger "Trusting a key grants code execution"
+    Restoring a cache entry deserializes it with `pickle`, so trusting a
+    fingerprint lets whoever holds that key run arbitrary code on your machine.
```
akshayka, line 319: "Restoring a cache entry can involve arbitrary code execution. For this reason, only add a fingerprint if you would run any script that person handed you"

kirangadhave (approved):

PR #10550, opened 2026-08-14
fix(home): surface when the workspace file scan is truncated

### LINKED ISSUE #10064: Workspace tree on home page silently omits folders/notebooks several levels deep, even after restart — but a targeted marimo edit <folder> finds them fine
### Describe the bug

# Workspace tree on home page silently omits folders/notebooks several levels deep, even after restart — but a targeted `marimo edit <folder>` finds them fine

## Describe the bug

When running `marimo edit` from a broad root directory (e.g. home directory), the **Workspace** panel on the home page silently omits some subfolders and the notebooks inside them — no error, no warning, nothing printed to the terminal. The same files are fully readable on disk (confirmed via `ls -la`), have normal permissions, and marimo has *previously* opened notebooks from inside the "missing" folder (they show correctly in **Recent notebooks**).

Re-rooting `marimo edit` directly at the folder in question makes everything appear correctly. So the underlying file access is fine — this looks like a scan/enumeration limit or bug specific to the Workspace tree builder when it has to walk a deeper/broader tree.

## Steps to reproduce

1. Have a directory structure roughly like:
   ```
   ~/.../Projects/
     ├── REAL Health AI Needs Assessment Analysis/   (2 files)
     ├── Needs Assessment Analysis/                   (~20 items: files + several subfolders)
     ├── Crumm Collaboration/
     ├── Ding-Mutation.../
     ├── Rose - Thematic Analysis/
     ├── ... (≈10 more sibling folders)
   ```
2. Run `marimo edit` from the home directory (not scoped to `Projects`).
3. Open the home page in the browser, expand `Projects` in the **Workspace** panel.

**Expected:** All sibling folders under `Projects` appear (or at minimum, all folders — even if not fully expanded).

**Actual:** Only 1 of ~11 sibling folders under `Projects` is shown (`REAL Health AI Needs Assessment Analysis`). The much larger sibling `Needs Assessment Analysis` — and everything inside it, including a subfolder added the same morning — never appears, even after:
   - Clicking the in-page refresh icon next to "Workspace"
   - A full hard browser reload (Cmd+Shift+R)
   - Fully quitting and restarting the marimo server process (confirmed via new PID and later timestamp than the folder's creation time)

## Workaround found

Running `marimo edit .` **from inside** (or scoped directly to) the folder that was missing makes the Workspace tree enumerate it correctly, including all subfolders and files:

```
cd ".../Projects/Needs Assessment Analysis"
marimo edit .
```
→ Workspace tree correctly shows all ~6 items including the previously-missing `needs assessmenet analysis 20260702` subfolder.

This strongly suggests the scan works correctly on a small/shallow tree, but breaks down (silently) when the root is broader/deeper — possibly a depth limit, per-directory item cap, total node cap, or a scan timeout that fails open (drops remaining entries) rather than erroring.

## Evidence ruling out other causes

- **Not a permissions issue:** `ls -la` on the "missing" folder shows it's owned by the same user running the marimo process (confirmed via `ps aux | grep marimo` — same `rsquires` user/session).
- **Not a stale cache:** New marimo server process (new PID, later timestamp) still didn't show the folder; browser hard-reload (Cmd+Shift+R) also had no effect.
- **Not a sync/cloud-placeholder issue** (this is on a OneDrive-synced path): `ls -la` on the folder returns real file sizes and normal timestamps, not cloud-only placeholders.
- **Not a silent crash:** No output/traceback appears in the terminal running `marimo edit` after a hard browser reload.
- **Confirmed it's specifically the tree-walk, not file access:** the exact same files are reachable and listed correctly under **Recent notebooks** (opened previously), and appear fully in the Workspace tree when marimo is scoped directly to that folder.

## Environment

- OS: macOS (Darwin)
- Filesystem: OneDrive-synced folder (`~/Library/CloudStorage/OneDrive-.../Projects/...`)
- marimo installed via conda env (`marimo-env`)
- Please let me know what other `marimo --version` / env info would help — happy to run `marimo env` and attach it if useful.

## Why this matters

For anyone with a moderately large or deeply nested project directory (common for research/data-science workflows with lots of subfolders), this makes the home page Workspace browser unreliable for discovering notebooks — files can be present, valid, and fully accessible, yet simply never show up unless you already know to scope `marimo edit` directly to their folder.

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
(marimo-env) rsquires@MBP-UTT-DMW4F3X Needs Assessment Analysis % marimo env
{
  "marimo": "0.23.13",
  "editable": false,
  "location": "/opt/anaconda3/envs/marimo-env/lib/python3.13/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.5.0",
  "Processor": "arm",
  "Python Version": "3.13.12",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "149.0.7827.201",
    "Node": "--",
    "uv": "--"
  },
  "Dependencies": {
    "click": "8.2.1",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10",
    "narwhals": "2.7.0",
    "packaging": "26.0",
    "psutil": "7.0.0",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "0.52.1",
    "tomlkit": "0.13.3",
    "uvicorn": "0.35.0",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.10.3",
    "pandas": "3.0.2"
  },
  "Experimental Flags": {}
}
```

</details>

### Code to reproduce

_No response_

NoiceHax:
The home page workspace tree drops a folder whenever the scan of it comes back empty, so "this folder has no notebooks" and "we stopped looking" end up looking the same. Notebooks deeper than MAX_DEPTH, past the file cap, or inside a folder that could not be read vanish along with their parent folders, and nothing in the response says the list is short.

This does not change which files get listed. The scanner now sets a flag when a limit or an unreadable entry cut the walk short, DirectoryWorkspace reports it as is_truncated, and workspace_files folds it into has_more. Home and gallery share one banner saying files and folders were skipped, and suggesting you start marimo from a subfolder.

Tests are in tests/_server/test_directory_scanner.py and tests/_server/test_workspace.py. They cover the depth cap, the file cap, an entry that fails to stat, and the flag being reset between scans.

Closes #10064

### DIFF
--- a/frontend/src/components/home/components.tsx
+++ b/frontend/src/components/home/components.tsx
@@ -31,6 +31,7 @@ import {
 import { Constants } from "@/core/constants";
 import { useRequestClient } from "@/core/network/requests";
 import type { TutorialId } from "@/core/network/types";
+import { Banner } from "@/plugins/impl/common/error-banner";
 import { openNotebook } from "@/utils/links";
 import { Objects } from "@/utils/objects";
 import { MarimoPlusIcon } from "../icons/marimo-icons";
@@ -182,6 +183,22 @@ export const ResourceLinks: React.FC = () => {
         ))}
       </div>
     </div>
+  );
+};
+
+/**
+ * Shown when the workspace scan stopped early, so the file list the user is
+ * looking at is missing files and folders.
+ */
+export const TruncatedWorkspaceBanner: React.FC<{
+  fileCount: number | undefined;
+}> = ({ fileCount }) => {
+  return (
+    <Banner kind="warn" className="rounded p-4">
+      Showing {fileCount} files. Some files and folders were skipped because
+      your workspace is large or deeply nested. Start marimo from a subfolder to
+      see them.
+    </Banner>
   );
 };
 
--- a/frontend/src/components/pages/gallery-page.tsx
+++ b/frontend/src/components/pages/gallery-page.tsx
@@ -120,8 +120,9 @@ const GalleryPage: React.FC = () => {
           <div className="flex flex-col gap-2">
             {workspace.hasMore && (
               <Banner kind="warn" className="rounded p-4">
-                Showing first {workspace.fileCount} files. Your workspace has
-                more files.
+                Showing {workspace.fileCount} files. Some files and folders were
+                skipped because your workspace is large or deeply nested. Start
+                marimo from a subfolder to see them.
               </Banner>
             )}
             {formattedFiles.length > SEARCH_THRESHOLD && (
--- a/frontend/src/components/pages/gallery-page.tsx
+++ b/frontend/src/components/pages/gallery-page.tsx
@@ -4,6 +4,7 @@ import { SearchIcon } from "lucide-react";
 import type React from "react";
 import { Suspense, useMemo, useState } from "react";
 import { ErrorBoundary } from "@/components/editor/boundary/ErrorBoundary";
+import { TruncatedWorkspaceBanner } from "@/components/home/components";
 import { Spinner } from "@/components/icons/spinner";
 import { Card, CardContent } from "@/components/ui/card";
 import { Input } from "@/components/ui/input";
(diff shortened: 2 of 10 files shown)


NoiceHax:

mscolnick:
@cubic-dev-ai

NoiceHax:
Fixed both points from the cubic review. The per-entry OSError handler now sets truncated too, since an entry we cannot stat may have been a notebook or a folder of them, with a test for it. The banner copy moved into a shared TruncatedWorkspaceBanner used by both the home and gallery pages.

mscolnick (approved):

PR #10551, opened 2026-08-14
fix(autoreload): drop stale @overload registrations before reloading

### LINKED ISSUE #5416: Module autoreload can break typing.get_overloads inspection.
### Describe the bug

python 3.11+ (`typing_extensions.overload` in python < 3.11)  exposes `get_overloads`,
which provides runtime introspection of `@overload` decorated functions.
This is implemented via an internal registry of the form `{module: {qualname: {firstlineno: func}}}`.
Ref: https://github.com/python/typing_extensions/blob/b07d24525615ba9377e47aaf5a26650a2517b2c4/src/typing_extensions.py#L351-L414
This is stable to module reloads which don't change the module source content,
as overloads are uniquely identified by their definition location.

`@overload` introspection is used in a variety of libraries and is frequently used to inspect overloads for multiple dispatch.
Notably, `plum` [introspects `@overload` decorated functions](https://beartype.github.io/plum/integration.html) to interop with static type checkers.

Marimo's autoreload triggers modules reloads on source module changes, 
and has semi-extensive logic to fixup the imported modules.
However, when a source module reload causes the source line to change for a redefinition,
`typing.get_overloads` will return duplicated entries for the function.

This can be fixed by updating marimo's autoreload to selectively clear `typing._overload_registry` (or `typing_extensions._overload_registry` on py<3.11) for module name before performing the module reload. 
The overload registry will then be repopulated with the correct overloads on module execution.

A related issue in plum's behavior has been filed at https://github.com/beartype/plum/issues/211

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
```

</details>

### Code to reproduce

Given module content...

```python
from typing_extensions import overload, get_overloads

@overload
def func(param: int) -> int: ...

@overload
def func(param: str) -> str: ...

def func(param): 
    return param

assert len(get_overloads(func)) == 2
```

will fail to autoreload with an assertion failure.

This is casually related to https://github.com/beartype/plum/issues/211, 
which causes plum dispatch failures after module reloads when using the `@overload` `@dispatch` pattern.

NoiceHax:
typing.overload keeps a private registry keyed by module, qualname, and the function's first line number, and get_overloads() reads it back. Autoreload re-executes the module but never clears that registry, so when a reload shifts line numbers the new entries sit next to the old ones and get_overloads() returns the overloads from every earlier version of the file. That breaks anything doing runtime dispatch on overloads, plum in particular.

superreload now drops the module's entry from the registry before reloading, and re-running the module fills it back in. There is no public API for clearing one module, clear_overloads() wipes the lot, so this reaches into typing._overload_registry and typing_extensions._overload_registry directly, wrapped in try/except with a debug log as asked for in the issue.

Tests are in tests/_runtime/reload/test_autoreload.py under TestOverloadRegistry. The regression test writes a module with two overloads, reloads it with an extra line at the top, and checks get_overloads() still returns 2.

Closes #5416

### DIFF
--- a/marimo/_runtime/reload/autoreload.py
+++ b/marimo/_runtime/reload/autoreload.py
@@ -512,6 +512,34 @@ def append_obj(
     return True
 
 
+def clear_module_overloads(modname: str) -> None:
+    """Drop `modname`'s `@overload` registrations ahead of a reload.
+
+    `typing.overload` (and `typing_extensions.overload` on Python 3.10)
+    records every decorated function in a private registry shaped like
+    `{module: {qualname: {firstlineno: func}}}`, which `get_overloads()`
+    reads back. The registry is keyed by source location, so reloading a
+    module whose line numbers shifted leaves the old entries behind and
+    `get_overloads()` then reports each overload more than once. Neither
+    module exposes a public API for evicting a single module's entries
+    (`clear_overloads()` wipes the whole registry), so we reach into the
+    registry defensively; re-executing the module repopulates it.
+    """
+    for typing_modname in ("typing", "typing_extensions"):
+        typing_module = sys.modules.get(typing_modname)
+        if typing_module is None:
+            continue
+        try:
+            registry = getattr(typing_module, "_overload_registry", None)
+            if isinstance(registry, dict):
+                registry.pop(modname, None)
+        except Exception:
+            LOGGER.debug(
+                f"Failed to clear {typing_modname} overloads for "
+                f"'{modname}': {traceback.format_exc(10)}"
+            )
+
+
 def superreload(
     module: types.ModuleType, old_objects: OldObjectsMapping | None
 ) -> types.ModuleType:
@@ -530,6 +558,9 @@ def superreload(
     # collect old objects in the module
     for name, obj in module.__dict__.items():
         append_obj(module, old_objects, name, obj)
+
+    # stale overload registrations would otherwise accumulate across reloads
+    clear_module_overloads(module.__name__)
 
     # reload module
     old_dict: dict[str, Any] | None = None
--- a/tests/_runtime/reload/test_autoreload.py
+++ b/tests/_runtime/reload/test_autoreload.py
@@ -17,6 +17,7 @@ from marimo._runtime.reload.autoreload import (
     ModuleReloader,
     StrongRef,
     append_obj,
+    clear_module_overloads,
     isinstance2,
     modules_imported_by_cell,
     safe_getattr,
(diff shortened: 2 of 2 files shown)


NoiceHax:

mscolnick:
@cubic-dev-ai

shashvat-singham (commented):
Ran the branch on Python 3.11.9 — 62 passed in `tests/_runtime/reload/test_autoreload.py`. I went and reproduced the underlying behaviour independently rather than taking the docstring's word for it, and it holds up exactly as described:

```
same line numbers            2 -> 2
shifted line numbers         2 -> 4
```

so the registry really is keyed by source location, and re-executing a module only leaks entries when the line numbers move. Popping the module key and re-executing puts it back to 2.

I also checked the end-to-end test isn't vacuous — replacing the `clear_module_overloads(module.__name__)` call with `pass` makes `test_reload_does_not_duplicate_overloads` fail, so it's genuinely pinned by the `"# shifted\n" + OVERLOAD_MODULE` rewrite. Good, that's the part that's easy to get wrong: without the deliberate line shift the test would have passed either way.

Two notes:

**Class-method overloads are covered but untested.** They accumulate the same way and the module-level `pop` does clean them up, since the registry nests `module -> qualname -> lineno`:

```
method overloads 1st: 2
method overloads 2nd: 3     # accumulated
after pop           : 2
```

Since `@overload` on methods is at least as common as on free functions, and the fix already handles it for free, it might be worth extending `OVERLOAD_MODULE` with a small class rather than leaving that behaviour unpinned.

**Clearing before a reload that then fails is fine, for the record.** My first reaction was that popping the registry ahead of `reload(module)` would strand a module whose re-execution raises, with its old overloads gone and no new ones registered. Having read the failure path, that's consistent with what's already intended — `superreload` deliberately doesn't restore `old_dict` and re-raises, with the comment saying the failure shouldn't be silent. So the overload registry ending up in the same cleared state matches the module dict. Mentioning it only because it looks alarming at first glance and someone else will likely have the same reaction.

Nit: the `try`/`except Exception` wraps `getattr` and `isinstance` too, neither of which can realistically raise — only `registry.pop` could, and only for an exotic mapping. Narrowing the block would make the defensive intent clearer, though I appreciate why you'd want the belt-and-braces around a private API.

NoiceHax:
Thanks for verifying the mechanism independently — and for checking the test wasn't vacuous, which is exactly the part that was easy to get wrong here.

Class-method overloads are now pinned, in `ec82bdcf`. `OVERLOAD_MODULE` gains a small class and the test asserts `get_overloads(Klass.method)` across the reload alongside the free function:

```python
class Klass:
    @overload
    def method(self, param: int) -> int: ...
    @overload
    def method(self, param: str) -> str: ...
    def method(self, param): ...
```

As you said, the module-level pop already handled them for free — the registry nests `module -> qualname -> lineno` — so this only closes the coverage gap rather than changing behaviour.

Confirmed it is load-bearing the same way you checked the original: replacing the `clear_module_overloads(module.__name__)` call with `pass` fails the test.

`tests/_runtime/reload/test_autoreload.py -k overload` → 3 passed.

PR #10552, opened 2026-08-14
fix(debugger): resolve cell-local names at the pdb prompt

### LINKED ISSUE #10269: Cell internal variables not exposed to debugger
### Describe the bug

When debugging and trying to print variables in scope, the cell-internal variables (those starting with underscore) are not visible even when the debugger is viewing the scope or expression directly using them.

If we have a cell such as:

```Python
a = 7
_b = 3 + a
_b + "Hello"
```

Debugging the exception in that cell, we can do `p a` but we can't do `p _b` (that results in `NameError: name '_b' is not defined`).

### Will you submit a PR?

- [ ] Yes

### Environment

{
  "marimo": "0.23.14",
  "editable": false,
  "location": "/Users/yairchu/dev/undersampled_add_sub/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.5.0",
  "Processor": "arm",
  "Python Version": "3.13.2",
  "Locale": "C/en_US",
  "Binaries": {
    "Browser": "140.0.7339.134",
    "Node": "v23.11.0",
    "uv": "0.11.16 (135a36367 2026-05-21 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.24.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.1",
    "uvicorn": "0.51.0",
    "websockets": "16.1.1"
  },
  "Optional Dependencies": {
    "anywidget": "0.11.0",
    "loro": "1.13.2",
    "mcp": "1.28.1",
    "openai": "2.46.0",
    "pandas": "3.0.3",
    "python-lsp-ruff": "2.3.1",
    "ruff": "0.15.22",
    "ty": "0.0.62"
  },
  "Experimental Flags": {}
}

### Code to reproduce

```python
a = 7
_b = 3 + a
_b + "Hello"

--- Debug exception

> p a
7
(Pdb)
> p _b
*** NameError: name '_b' is not defined
(Pdb)
```

NoiceHax:
marimo renames a cell's underscore variables so they stay private to the cell, so `_b` is really stored as `_cell_<cell_id>_b`. The name as written never exists in the frame pdb stops in, which is why `p _b` fails with a NameError while `p a` works.

MarimoPdb now parses what you type and rewrites cell-local names to the mangled ones before pdb evaluates it. A name is only rewritten if it isn't already defined in the frame and the mangled version is, so an unmangled underscore import still shadows the cell-local. Non-cell frames and input that isn't valid Python go through untouched. Names bound by the typed source itself are skipped as well, otherwise `p (lambda _b: _b + 1)(5)` would rewrite the body but not the parameter.

Worth poking at: `p`, `pp`, `display`, and plain statements like `_b = _b + 1`, which should land on the cell's variable. New tests are in tests/_runtime/test_marimo_pdb.py.

Closes #10269

### DIFF
--- a/marimo/_runtime/marimo_pdb.py
+++ b/marimo/_runtime/marimo_pdb.py
@@ -1,12 +1,19 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
+import ast
 import inspect
 import sys
 from pdb import Pdb, Restart as pdbRestart
 from typing import TYPE_CHECKING, Any
 
 from marimo import _loggers
+from marimo._ast.compiler import cell_id_from_filename
+from marimo._ast.variables import (
+    if_local_then_mangle,
+    is_local,
+    is_mangled_local,
+)
 from marimo._messaging.types import Stdin, Stdout
 
 if TYPE_CHECKING:
@@ -47,6 +54,36 @@ def try_restart() -> bool:
         return False
 
     return True
+
+
+class _CellLocalMangler(ast.NodeTransformer):
+    """Rewrite a cell's local names to the mangled names they are stored under.
+
+    Only rewrites a name when it doesn't already resolve in the frame and its
+    mangled counterpart does, so genuine globals (e.g. an unaliased `from x
+    import _`, which marimo leaves unmangled) keep shadowing the cell-local.
+    """
+
+    def __init__(self, cell_id: CellId_t, frame: FrameType) -> None:
+        self.cell_id = cell_id
+        self.frame = frame
+        self.mangled = False
+
+    def _defined(self, name: str) -> bool:
+        return name in self.frame.f_locals or name in self.frame.f_globals
+
+    def visit_Name(self, node: ast.Name) -> ast.Name:
+        if (
+            not is_local(node.id)
+            or is_mangled_local(node.id)
+            or self._defined(node.id)
+        ):
+            return node
+        mangled = if_local_then_mangle(node.id, self.cell_id)
+        if self._defined(mangled):
+            node.id = mangled
+            self.mangled = True
+        return node
 
 
 class MarimoPdb(Pdb):
@@ -96,6 +133,52 @@ class MarimoPdb(Pdb):
         if header is not None:
             sys.stdout.write(header)
         return super().set_trace(frame)
+
+    def _mangle_cell_locals(
+        self, source: str, frame: FrameType | None = None
+    ) -> str:
+        """Rewrite cell-local names in debugger input to their real names.
+
+        marimo mangles a cell's underscore-prefixed variables so they stay
+        private to the cell (`_b` is stored as `_cell_<cell_id>_b`), which
+        would otherwise leave the names the user wrote undefined at the
+        debugger prompt. Rewriting them before pdb evaluates the input makes
+        `p _b` work just like `p b`.
+
+        Input that isn't valid Python, or that isn't being evaluated in a
+        cell's frame, is handed to pdb untouched.
+        """
+        if frame is None:
+            frame = getattr(self, "curframe", None)
+        if frame is None:
+            return source
+        cell_id = cell_id_from_filename(frame.f_code.co_filename)
+        if cell_id is None:
+            return source
+        try:
+            tree = ast.parse(source)
+        except SyntaxError:
+            return source
+        mangler = _CellLocalMangler(cell_id, frame)
+        mangler.visit(tree)
+        return ast.unparse(tree) if mangler.mangled else source
+
+    def default(self, line: str) -> Any:
+        """Evaluate a statement typed at the prompt, demangling cell locals."""
+        # pdb strips a leading `!` before compiling; strip it here too so that
+        # what's left parses as Python.
+        bang, source = ("!", line[1:]) if line[:1] == "!" else ("", line)
+        return super().default(bang + self._mangle_cell_locals(source))
+
+    def _getval(self, arg: str) -> Any:
+        """Evaluate an expression (`p`, `pp`, ...), demangling cell locals."""
+        return super()._getval(self._mangle_cell_locals(arg))
+
+    def _getval_except(self, arg: str, frame: FrameType | None = None) -> Any:
+        """Evaluate an expression for `display`, demangling cell locals."""
+        return super()._getval_except(
+            self._mangle_cell_locals(arg, frame), frame
+        )
 
     def cmdloop(self, intro: str | None = None) -> None:
         """Override to gracefully handle restarts."""
--- a/marimo/_runtime/marimo_pdb.py
+++ b/marimo/_runtime/marimo_pdb.py
@@ -17,6 +17,7 @@ from marimo._ast.variables import (
 from marimo._messaging.types import Stdin, Stdout
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
     from types import FrameType, TracebackType
 
     from marimo._types.ids import CellId_t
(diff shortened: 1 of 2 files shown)


NoiceHax:

mscolnick:
@cubic-dev-ai

NoiceHax:
Fixed the cubic finding. The mangler now tracks scopes the typed source introduces (lambda, def, comprehensions) and leaves names those scopes bind alone, so `p (lambda _b: _b + 1)(5)` and `[_b for _b in xs]` behave. Names the expression does not bind still resolve to the cell local, including a default like `lambda _b=_b: _b` where the default is evaluated outside. This matches ScopedVisitor, which only mangles locals that resolve against the cell's top-level scope. Two tests added, both fail without the change.

shashvat-singham (commented):
Ran the branch — 10 passed in `tests/_runtime/test_marimo_pdb.py` (the two `test_pdb_patched` failures I hit first were just a missing `pytest-asyncio` in my env; they fail identically on `main`).

I stress-tested `_names_bound_by` against every binding form I could think of, since the whole design rests on it being an over-approximation. It's solid almost everywhere:

```
yes assign, aug-assign, ann-assign, for target, with as, except as,
    walrus, comprehension, import, import as, from import, def,
    class, lambda arg, global decl, star target
```

**`match` patterns are the one gap.** Capture names bind through `ast.MatchAs`/`MatchStar`/`MatchMapping`, not through `ast.Name` with a `Store` context, so none of them are collected:

```
case capture       -> []      # match v: case x: ...
case as            -> []      # case [1] as x
case star          -> []      # case [1, *x]
case mapping rest  -> []      # case {1: a, **x}
case class kwarg   -> []      # case C(y=x)
```

That's the unsafe direction rather than the safe one your docstring describes: a missed name isn't in `bound`, so `_bound_by_nested_scope` returns `False` and `visit_Name` will mangle it.

In practice the blast radius is small, and I'd rather say so than oversell it — it needs a `_`-prefixed name (`is_local`), bound by a `case` pattern, inside a `def` typed at the prompt, that doesn't resolve via `_defined` but whose mangled counterpart does. `match` can't appear in a `lambda`, so the comprehension and lambda paths can't reach it at all. So: a real hole in an otherwise exhaustive helper, not something users are likely to hit.

Adding

```python
elif isinstance(child, (ast.MatchAs, ast.MatchStar)) and child.name is not None:
    names.add(child.name)
elif isinstance(child, ast.MatchMapping) and child.rest is not None:
    names.add(child.rest)
```

would close it — `ast.walk` already descends into the pattern nodes, so it's just the extra isinstance arms.

Same class of thing, worth a thought: on Python 3.12+ `type X = int` binds `X` via `ast.TypeAlias`, which is also not an `ast.Name`/`Store`. I couldn't test it here (3.11 rejects the syntax), but it looks like it'd be missed the same way, and PEP 695 type params (`def f[T]()`) introduce `ast.TypeVar` bindings too.

Nice piece of work otherwise — collecting the whole scope's bindings up front rather than as you walk is the right call, and the comment explaining why is the kind of thing that saves the next reader.

shashvat-singham:
Pulled the updated branch — the lambda/comprehension scope tracking works as described:

```
lambda arg      -> ['_b', 'f']
comprehension   -> ['_b']
```

The `match` gap from my earlier comment is still open, though, and it's the same class of bug the cubic finding was:

```
case capture    -> []      # match v: case _x: ...
case as         -> []      # case [1] as _x
```

Capture names bind through `ast.MatchAs`/`MatchStar`/`MatchMapping`, not through `ast.Name` with a `Store` context, so `_names_bound_by` doesn't see them and `visit_Name` will mangle a `_`-prefixed capture. Concretely, typing this at the pdb prompt inside a cell that has a `_x`:

```python
def f(v):
    match v:
        case _x:
            return _x
```

`_x` is bound by the pattern but gets rewritten to the cell-local mangled name.

`ast.walk` already descends into pattern nodes, so it's two extra arms:

```python
elif isinstance(child, (ast.MatchAs, ast.MatchStar)) and child.name is not None:
    names.add(child.name)
elif isinstance(child, ast.MatchMapping) and child.rest is not None:
    names.add(child.rest)
```

Same shape applies to `ast.TypeAlias` (`type X = int`) and PEP 695 type params on 3.12+, which are also bindings that aren't `Name`/`Store` — I couldn't exercise those here since I'm on 3.11.

Worth a test alongside the lambda/comprehension ones so the next scope form doesn't slip through the same way.

marimo/_runtime/marimo_pdb.py
```diff
@@ -49,6 +57,160 @@ def try_restart() -> bool:
     return True
 
 
+def _defaults(args: ast.arguments) -> list[ast.expr]:
+    """Default values, which are evaluated in the enclosing scope."""
+    return [d for d in (*args.defaults, *args.kw_defaults) if d is not None]
+
+
+def _names_bound_by(nodes: Iterable[ast.AST]) -> set[str]:
```
dmadisetti, line 65: I think this information should be available on the cell level

marimo/_runtime/marimo_pdb.py
```diff
@@ -49,6 +57,160 @@ def try_restart() -> bool:
...
+                names.update(
+                    (alias.asname or alias.name).split(".")[0]
+                    for alias in child.names
+                )
+            elif isinstance(child, ast.ExceptHandler) and child.name:
+                names.add(child.name)
+    return names
+
+
+class _CellLocalMangler(ast.NodeTransformer):
```
dmadisetti, line 96: If we did want to do this, I think we would want to break it out into its own module.

marimo/_runtime/marimo_pdb.py
```diff
@@ -97,6 +259,52 @@ def set_trace(
...
+        """
+        if frame is None:
+            frame = getattr(self, "curframe", None)
+        if frame is None:
+            return source
+        cell_id = cell_id_from_filename(frame.f_code.co_filename)
+        if cell_id is None:
+            return source
+        try:
+            tree = ast.parse(source)
```
dmadisetti, line 284: Our helper `ast_parse` since ast.parse has warning side effects

marimo/_runtime/marimo_pdb.py
dmadisetti: Ahh I see. I think we could maybe just use our standard visitor, which does mangling as well. Worried about bifurcation. I don't think we need to check against what's in memory (so no changes needed, don't need to pass the frame). Just always do the rewrite with the correct mangle prefix

marimo/_runtime/marimo_pdb.py
```diff
@@ -97,6 +259,52 @@ def set_trace(
...
+        `p _b` work just like `p b`.
+
+        Input that isn't valid Python, or that isn't being evaluated in a
+        cell's frame, is handed to pdb untouched.
+        """
+        if frame is None:
+            frame = getattr(self, "curframe", None)
+        if frame is None:
+            return source
+        cell_id = cell_id_from_filename(frame.f_code.co_filename)
```
dmadisetti, line 280: This may break in certain cases- line cache isn't always super reliable. `safe_get_context()` is likely better and extraction from there

dmadisetti (commented):
Nice! I think we can simplify the visitor stuff a bunch by just using ours, but this is a good addition. Thank you!

NoiceHax:
Thanks for the thorough pass, and sorry for the delay — the `match` gap is fixed in `4028c4b7`.

You were right about the mechanism: capture names bind through the pattern node (`ast.MatchAs` / `ast.MatchStar` / `ast.MatchMapping`), never through an `ast.Name` with a `Store` context, so `_names_bound_by` could not see them. All five forms you listed now bind:

```
case capture       -> ['x']      # match v: case x: ...
case as            -> ['x']      # case [1] as x
case star          -> ['x']      # case [1, *x]
case mapping rest  -> ['a', 'x'] # case {1: a, **x}
case class kwarg   -> ['x']      # case C(y=x)
```

`case _:` carries `name=None` and correctly binds nothing, so a `_x` in that body still resolves to the cell local — pinned as its own assertion rather than left implicit.

`ast.MatchAs` and friends are unconditional here since `requires-python = ">=3.10"`, so no version guard is needed.

The new test fails on the unpatched tree and passes with the change; `tests/_runtime/test_marimo_pdb.py` is 11 passed.

@dmadisetti — on simplifying the visitor by using yours: happy to do that, either here or as a follow-up. Say which you'd prefer and I'll rework it.

dmadisetti:
@NoiceHax I'm going to move this back into draft until you take another pass at it. My comments weren't addressed. Please feel free to hit "request review" when you're ready (I'm more likely to see it than a @)"

Also, if you could sign the CLA that would be great!

PR #10570, opened 2026-08-17
fix(tables): stop URL linkification at delimiters, not whitespace

### LINKED ISSUE #10567: Table: URL linkification is greedy, so URLs inside JSON text get trailing delimiters in the href
A str column holding JSON gets linkified with something like `/(https?:\/\/\S+)/`. `\S+` runs to the next whitespace, and JSON text has none before the closing `"}]`, so those characters land inside the href.

```python
import polars as pl

df = pl.DataFrame({
    "attachments": ['[{"name":"inv.pdf","url":"https://example.com/p/AAA"}]'],
})
df
```

The link points at `https://example.com/p/AAA%22%7D]` which is incorrect.

marimo 0.23.16, Windows 11.

Abdulrehman-PIAIC80387:
## 📝 Summary

Closes marimo-team/marimo#10567

The table-cell URL linkifier matched URLs with `https?://\S+`, so a URL embedded in JSON text had the trailing `"}]` delimiters swallowed into the href.

```python
import polars as pl
pl.DataFrame({"attachments": ['[{"name":"inv.pdf","url":"https://example.com/p/AAA"}]']})
```

The rendered link pointed at `https://example.com/p/AAA%22%7D]` (the `"}]` percent-encoded into the href) instead of `https://example.com/p/AAA`.

## 🔍 Root cause

`frontend/src/utils/url-parser.ts` (the table linkifier used by `data-table/columns.tsx`):

```ts
const urlRegex = /(https?:\/\/\S+)/;
```

`\S+` matches any non-whitespace, so with no whitespace before the closing `"}]` it keeps consuming those delimiters.

## 🔧 What changed

Match a URL up to the first character that **cannot appear unencoded in a URL** (RFC 3986 excluded/"unwise" set — whitespace and `" ' < > \` { } | \\ ^\`) instead of any non-whitespace:

```diff
-const urlRegex = /(https?:\/\/\S+)/;
+const urlRegex = /(https?:\/\/[^\s"'<>`{}|\\^]+)/;
```

* Applied to the same greedy regex in `json/json-parser.ts` (the JSON-viewer link conversion), which had the identical bug.
* **Deliberately keeps** `. , ( ) [ ]` etc. valid — real URLs use them (query strings, IPv6 hosts like `http://[::1]/`, parenthesised paths), so they must not truncate the match.
* Left `urls.ts`'s `^(https?://\S+)$` alone — it's an anchored whole-string `isUrl()` check where `\S+` is correct.

## 🎥 Demo

https://github.com/user-attachments/assets/1d9f7e12-e9a7-4a2e-a8c6-aed7b3c2cbad

*Shows the* `attachments` *cell: before, "Copy Link Address" yields* `https://example.com/p/AAA%22%7D]`*; after, it yields the clean* `https://example.com/p/AAA`*.*

## 🧪 Testing

* Added regression tests in `url-parser.test.ts` for the reporter's JSON case and a trailing-quote case.
* Verified by mutation: reverting the regex to `\S+` makes the new tests fail.
* `data-table` consumer tests (`url-detector`, `columns`) — 87 pass, no regressions.
* `make fe-check` (typecheck + lint) clean.
* Verified manually in `make dev` (before/after link address as above).

## 📋 Pre-Review Checklist

- [X] For large changes, or changes that affect the public API: n/a — small, self-contained bug fix.

## ✅ Merge Checklist

- [X] I have read the contributor guidelines.
- [X] Documentation has been updated where applicable — n/a.

### DIFF
--- a/frontend/src/utils/json/json-parser.ts
+++ b/frontend/src/utils/json/json-parser.ts
@@ -1,5 +1,6 @@
 /* Copyright 2026 Marimo. All rights reserved. */
 
+import { URL_REGEX } from "../url-parser";
 import type { JsonString } from "./base64";
 
 declare global {
@@ -126,10 +127,10 @@ function formatValueForMarkdown(value: unknown): string {
     return placeholder;
   });
 
-  // Convert plain URLs to markdown links. Exclude whitespace and the
-  // characters that cannot appear unencoded in a URL so a URL embedded in JSON
-  // text doesn't swallow the trailing `"}]` delimiters into the link (#10567).
-  const urlRegex = /(https?:\/\/[^\s"'<>`{}|\\^]+)/g;
+  // Convert plain URLs to markdown links, using the shared URL matcher so this
+  // stays in sync with the table-cell linkifier and doesn't swallow trailing
+  // delimiters like `"}]` around a URL embedded in JSON text (#10567).
+  const urlRegex = new RegExp(URL_REGEX.source, "g");
   stringValue = stringValue.replaceAll(urlRegex, "[$1]($1)");
 
   // Restore markdown links from placeholders
--- a/frontend/src/utils/json/json-parser.ts
+++ b/frontend/src/utils/json/json-parser.ts
@@ -126,8 +126,10 @@ function formatValueForMarkdown(value: unknown): string {
     return placeholder;
   });
 
-  // Convert plain URLs to markdown links
-  const urlRegex = /(https?:\/\/\S+)/g;
+  // Convert plain URLs to markdown links. Exclude whitespace and the
+  // characters that cannot appear unencoded in a URL so a URL embedded in JSON
+  // text doesn't swallow the trailing `"}]` delimiters into the link (#10567).
+  const urlRegex = /(https?:\/\/[^\s"'<>`{}|\\^]+)/g;
   stringValue = stringValue.replaceAll(urlRegex, "[$1]($1)");
 
   // Restore markdown links from placeholders
--- a/frontend/src/utils/json/json-parser.ts
+++ b/frontend/src/utils/json/json-parser.ts
@@ -3,6 +3,8 @@
 import { URL_REGEX } from "../url-parser";
 import type { JsonString } from "./base64";
 
+const GLOBAL_URL_REGEX = new RegExp(URL_REGEX.source, "g");
+
 declare global {
   interface BigInt {
     toJSON(): unknown;
@@ -127,11 +129,8 @@ function formatValueForMarkdown(value: unknown): string {
     return placeholder;
   });
 
-  // Convert plain URLs to markdown links, using the shared URL matcher so this
-  // stays in sync with the table-cell linkifier and doesn't swallow trailing
-  // delimiters like `"}]` around a URL embedded in JSON text (#10567).
-  const urlRegex = new RegExp(URL_REGEX.source, "g");
-  stringValue = stringValue.replaceAll(urlRegex, "[$1]($1)");
+  // Convert plain URLs to markdown links using the shared URL matcher.
+  stringValue = stringValue.replaceAll(GLOBAL_URL_REGEX, "[$1]($1)");
 
   // Restore markdown links from placeholders
   for (const [placeholder, original] of placeholders) {
--- a/frontend/src/utils/url-parser.ts
+++ b/frontend/src/utils/url-parser.ts
@@ -1,18 +1,7 @@
 /* Copyright 2026 Marimo. All rights reserved. */
 
-/**
- * Matches an `http(s)` URL, stopping at whitespace and the characters that
- * cannot appear unencoded in a URL (the RFC 3986 excluded/"unwise" set:
- * `" ' < > \` { } | \ ^`). This keeps linkification from greedily swallowing
- * trailing delimiters such as the closing `"}]` around a URL embedded in JSON
- * text (#10567) into the href, while preserving characters that are valid in
- * real URLs (`. , ( ) [ ]`, query strings, etc.).
- *
- * Shared so the table-cell linkifier here and the JSON-viewer linkifier in
- * `json/json-parser.ts` stay in sync (build a global copy with
- * `new RegExp(URL_REGEX.source, "g")`).
- */
-export const URL_REGEX = /(https?:\/\/[^\s"'<>`{}|\\^]+)/;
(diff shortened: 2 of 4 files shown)


mscolnick:
@cubic-dev-ai

frontend/src/utils/json/json-parser.ts
```diff
@@ -126,8 +126,10 @@ function formatValueForMarkdown(value: unknown): string {
     return placeholder;
   });
 
-  // Convert plain URLs to markdown links
-  const urlRegex = /(https?:\/\/\S+)/g;
+  // Convert plain URLs to markdown links. Exclude whitespace and the
+  // characters that cannot appear unencoded in a URL so a URL embedded in JSON
+  // text doesn't swallow the trailing `"}]` delimiters into the link (#10567).
+  const urlRegex = /(https?:\/\/[^\s"'<>`{}|\\^]+)/g;
```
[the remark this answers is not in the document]
Abdulrehman-PIAIC80387: Done in 9327784f8 — extracted a single exported `URL_REGEX` in `url-parser.ts`; `json-parser.ts` now derives its global copy via `new RegExp(URL_REGEX.source, "g")`, so the two linkifiers share one source of truth and can't drift.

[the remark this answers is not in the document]
Abdulrehman-PIAIC80387: Done in 9327784f8 — added `json-parser.test.ts` covering `jsonToMarkdown`: a URL followed by `"}]` stops at the quote, a trailing quote isn't swallowed, and existing markdown links are left untouched. Mutation-verified (reverting to `\S+` fails them).

Light2Dark:
I think would be good to add a benchmark test, regexes like this have a chance to be slow.

kirangadhave:
Thanks @Abdulrehman-PIAIC80387 for the PR. 
I might close this. I have a separate fix for parsing stringified JSON in the cells. That might make this PR moot. I'll verify. If there is still a need after that fix, we can keep this.

frontend/src/utils/json/json-parser.ts
```diff
@@ -126,8 +127,10 @@ function formatValueForMarkdown(value: unknown): string {
     return placeholder;
   });
 
-  // Convert plain URLs to markdown links
-  const urlRegex = /(https?:\/\/\S+)/g;
+  // Convert plain URLs to markdown links, using the shared URL matcher so this
+  // stays in sync with the table-cell linkifier and doesn't swallow trailing
+  // delimiters like `"}]` around a URL embedded in JSON text (#10567).
+  const urlRegex = new RegExp(URL_REGEX.source, "g");
```
kirangadhave: You can compile the regex once at top of the file instead of every time we format. We don't need the doc comment here. the URL_REGEX has a comment already.

```ts
import { URL_REGEX } from "../url-parser";

const GLOBAL_URL_REGEX = new RegExp(URL_REGEX.source, "g");
```

frontend/src/utils/url-parser.ts
```diff
@@ -1,6 +1,18 @@
...
+ * cannot appear unencoded in a URL (the RFC 3986 excluded/"unwise" set:
+ * `" ' < > \` { } | \ ^`). This keeps linkification from greedily swallowing
+ * trailing delimiters such as the closing `"}]` around a URL embedded in JSON
+ * text (#10567) into the href, while preserving characters that are valid in
+ * real URLs (`. , ( ) [ ]`, query strings, etc.).
+ *
+ * Shared so the table-cell linkifier here and the JSON-viewer linkifier in
+ * `json/json-parser.ts` stay in sync (build a global copy with
+ * `new RegExp(URL_REGEX.source, "g")`).
+ */
```
kirangadhave: This doc comment is too long. References to RFCs and specific issues can become stale. Keep doc comments concise and timeless; describe the matcher’s behavior rather than its history.


```suggestion
/** Matches HTTP(S) URLs without consuming surrounding delimiters. */
```

frontend/src/utils/url-parser.ts
```diff
@@ -1,6 +1,18 @@
...
+export const URL_REGEX = /(https?:\/\/[^\s"'<>`{}|\\^]+)/;
```
kirangadhave: the matcher excludes `'` but it is a valid part of URL (see: https://www.rfc-editor.org/info/rfc3986/#section-2.2, e.g. https://example.com/O'Reilly).

remove it from the exclusion list and add a regression test.

kirangadhave (changes requested):
thanks for the PR; some minor comments; looks good to merge after you address those.

frontend/src/utils/url-parser.ts
kirangadhave: this is echo'd by the copilot comment above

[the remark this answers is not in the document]
Abdulrehman-PIAIC80387: Good catch — fixed in 773fcfa10. Removed `'` from the excluded set so apostrophe URLs like `https://example.com/O'Reilly` are no longer truncated, and added a regression test. Double-quote stays excluded, so the original JSON case is unaffected.

frontend/src/utils/json/json-parser.ts
Abdulrehman-PIAIC80387: Done in 773fcfa10 — moved it to a module-level `GLOBAL_URL_REGEX = new RegExp(URL_REGEX.source, "g")` and dropped the inline doc comment.

frontend/src/utils/url-parser.ts
Abdulrehman-PIAIC80387: Done in 773fcfa10 — shortened it to `/** Matches HTTP(S) URLs without consuming surrounding delimiters. */`, dropping the RFC/issue references.

Abdulrehman-PIAIC80387:
Thanks @kirangadhave for the careful review — all points are addressed in `773fcfa10`:

- **Apostrophe handling** — removed `'` from the excluded character class so valid URLs like `https://example.com/O'Reilly` are no longer truncated (with a regression test); the double-quote exclusion stays, so the original JSON case is unaffected.
- **Compile once** — the global matcher is now a module-level `GLOBAL_URL_REGEX = new RegExp(URL_REGEX.source, "g")` instead of being rebuilt on every `formatValueForMarkdown` call.
- **Doc comment** — tightened `URL_REGEX`'s comment to a concise, timeless one-liner, dropping the RFC/issue references.

CI is green. Further changes welcome!

kirangadhave (approved):

PR #10592, opened 2026-08-19
feat: detect duckdb spatial columns as geometry

kirangadhave:
## Summary

Detect DuckDB spatial columns from relation/schema metadata as the `geometry` semantic type.

`GEOMETRY` columns render safely as WKB placeholders in table JSON, while fixed-layout spatial types (`POINT_2D`, `LINESTRING_2D`, `POLYGON_2D`, `BOX_2D`) are typed as geometry without value sniffing.

SQL metadata now maps DuckDB spatial type strings to `geometry`, and geometry SQL stats return counts-only fields (total/nulls). This keeps table and SQL paths consistent for DuckDB spatial data.

No frontend changes.

## 🗺️ Stack

This PR is part of a four-PR stack that adds a semantic `geometry` type for geospatial columns ([MO-6362](https://linear.app/marimo/issue/MO-6362)). Merge bottom-up with squash. Each PR targets the branch below it, so each diff shows only its own commits. After each merge, I rebase the remaining PRs.

1. #10516: geometry fixture corpus and ordinary-table baselines
2. #10536: `geometry` contract and GeoPandas detection
3. #10558: PyArrow adapter (GeoArrow field metadata)
4. #10592: DuckDB adapter and SQL metadata (completes the release gate) ⬅️ this PR

ibis detection is deferred: no user demand, and nothing downstream depends on it.

> [!NOTE]
> This PR completes the release gate. A release that contains #10536, #10558, and this PR can emit the `geometry` type.

Closes MO-7325

### DIFF
--- a/marimo/_data/get_datasets.py
+++ b/marimo/_data/get_datasets.py
@@ -519,9 +519,15 @@ _STRING_TYPES = {
 _TIME_TYPES = {"time", "time with time zone", "timetz"}
 _DATETIME_TYPES = {"datetime", "interval"}
 _BINARY_TYPES = {"bit", "bitstring", "binary", "varbinary", "bytea"}
+_GEOMETRY_TYPES = {
+    "geometry",
+    "point_2d",
+    "linestring_2d",
+    "polygon_2d",
+    "box_2d",
+}
 _UNKNOWN_TYPES = {
     "row",
-    "geometry",
     "inet",
     # Null type (can occur when attaching databases or with unknown column types)
     "null",
@@ -564,6 +570,10 @@ def _db_type_to_data_type(db_type: str) -> DataType:
     # Enum types (represented as string)
     if db_type == "enum" or db_type.startswith("enum"):
         return "string"
+
+    # Spatial extension types
+    if db_type in _GEOMETRY_TYPES:
+        return "geometry"
 
     # Nested types
     if db_type.startswith(
--- a/marimo/_data/sql_summaries.py
+++ b/marimo/_data/sql_summaries.py
@@ -56,6 +56,13 @@ def get_sql_stats(
             SUM(CASE WHEN "{column_name}" = FALSE THEN 1 ELSE 0 END) as false_count
         FROM {table_name}
         """
+    elif column_type == "geometry":
+        stats_query = f"""
+        SELECT
+            COUNT(*) as count,
+            SUM(CASE WHEN "{column_name}" IS NULL THEN 1 ELSE 0 END) as null_count
+        FROM {table_name}
+        """
     else:
         stats_query = f"""
         SELECT
@@ -124,6 +131,9 @@ def get_sql_stats(
             true=true_count,
             false=false_count,
         )
+    elif column_type == "geometry":
+        count, null_count = stats_result
+        return ColumnStats(total=count, nulls=null_count)
     else:
         count, unique, null_count = stats_result
         return ColumnStats(total=count, unique=unique, nulls=null_count)
--- a/marimo/_plugins/ui/_impl/tables/geometry.py
+++ b/marimo/_plugins/ui/_impl/tables/geometry.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Literal
 from marimo import _loggers
 
 if TYPE_CHECKING:
+    import duckdb
     import narwhals.stable.v2 as nw
     import pandas as pd
     import pyarrow as pa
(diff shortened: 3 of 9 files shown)


JoostGevaert:
Nice to see this land!
I'd like to add that DuckDB geometry is really **two families**, and the PR covers one half of each, see this marimo notebook: [Open in molab](https://molab.marimo.io/notebooks/nb_uLSRa2CUXNjDUN22zSgpKS) and:

| family | type | needs `spatial` | with this PR |
|---|---|---|---|
| native | `GEOMETRY` | no | `geometry` ✅ |
| native | `GEOMETRY('OGC:CRS84')`, `GEOMETRY('OGC:CRS83')` | no | `unknown` + warns ❌ |
| spatial | `GEOMETRY('EPSG:7415')` and every other PROJ CRS | yes | `unknown` + warns ❌ |
| spatial | `POINT_2D`, `LINESTRING_2D`, `POLYGON_2D`, `BOX_2D` | yes | `geometry` ✅ |

The pattern is that **everything CRS-parameterised misses**, in both families, because both new checks are exact-match: `db_type in _GEOMETRY_TYPES` and `type_name == _DUCKDB_WKB_TYPE`. Neither `relation.types` nor `DESCRIBE` normalises the parameter away.

## 1. Native DuckDB geometry — no extension involved

[`GEOMETRY` is a built-in type as of DuckDB 1.5](https://duckdb.org/docs/current/sql/data_types/geometry), and core registers a handful of CRSes (two in 1.5.5: `OGC:CRS83`, `OGC:CRS84`), so this needs nothing installed. Verified with `autoload_known_extensions=false` and `autoinstall_known_extensions=false`, i.e. spatial genuinely not loaded:

```python
import duckdb
from marimo._data.get_datasets import get_databases_from_duckdb

con = duckdb.connect(":memory:")          # no extensions at all
con.execute("CREATE TABLE native(plain GEOMETRY, crs GEOMETRY('OGC:CRS84'))")
get_databases_from_duckdb(con, engine_name="con")
```

```
[W get_datasets:578] Unknown DuckDB type: geometry('ogc:crs84')
plain  GEOMETRY                 -> unknown
crs    GEOMETRY('OGC:CRS84')    -> unknown
```

<details>
<summary>Testing trap: the CRS is session-only, so a file round-trip will not reproduce it</summary>

The CRS type modifier is not written to storage. Declare `GEOMETRY('OGC:CRS84')`, close, reopen, and the column reads back as bare `GEOMETRY` with `ST_CRS() = NULL` — coordinates intact, label gone. It is not about the definition living in an extension: a core-only `OGC:CRS84` is dropped exactly like `EPSG:7415`. Acknowledged upstream by the spatial maintainer — ["CRS definitions need to be provided by an extension and are not persisted in the database itself"](https://github.com/duckdb/duckdb-spatial/issues/441#issuecomment-4024748910) — and unchanged in 1.5.5.

Practically: a test that writes a `.duckdb` file and reopens it sees bare `GEOMETRY` and passes for the wrong reason. Use an in-memory connection, or assert on the type string directly.

</details>

## 2. Spatial extension geometry

`INSTALL spatial; LOAD spatial;` grows [`duckdb_coordinate_systems()`](https://duckdb.org/docs/current/sql/meta/duckdb_table_functions#duckdb_coordinate_systems) from 2 rows to 7246, which makes every PROJ CRS declarable, and adds the explicit fixed-layout types (`POINT_2D` and friends, backed by DuckDB's native nested types — `{'x': 1.0, 'y': 2.0}::POINT_2D` casts straight through).

```python
con.execute("INSTALL spatial; LOAD spatial;")
con.execute("CREATE TABLE spatial(plain GEOMETRY, crs GEOMETRY('EPSG:7415'), p POINT_2D)")
[str(t) for t in con.table("spatial").types]
# ['GEOMETRY', "GEOMETRY('EPSG:7415')", 'POINT_2D']
```

This PR fixes `point_2d`. `geometry('epsg:7415')` still warns.

Worth supporting rather than tolerating, because DuckDB enforces the parameter: inserting a plain `POINT_3D` into a CRS-typed column raises `Unimplemented type for cast (POINT_3D -> GEOMETRY('EPSG:7415'))`. `ST_SetCRS` is that cast, so anyone using it is being deliberate about the CRS, not decorative.

## Suggested fix

Mirroring the `enum` branch three lines above:

```python
# marimo/_data/get_datasets.py
if db_type in _GEOMETRY_TYPES or db_type.startswith("geometry"):
    return "geometry"
```

```python
# marimo/_plugins/ui/_impl/tables/geometry.py
if type_name == _DUCKDB_WKB_TYPE or type_name.startswith(f"{_DUCKDB_WKB_TYPE}("):
```

A test alongside the existing ones: `assert _db_type_to_data_type("geometry('ogc:crs84')") == "geometry"` — the native case, so it needs no extension in CI.

Verified on marimo 0.24.0 with DuckDB 1.5.4 and 1.5.5.

kirangadhave:
Thanks @JoostGevaert this is super helpful!

mscolnick (dismissed):

mscolnick (approved):

PR #10595, opened 2026-08-19
Make proposed AI edits UI consistent with editor styling

akshayka:
This pull request makes AI-generated notebook cells look consistent with the rest of the editor and clarifies how users review proposed changes.

It also updates the inline AI refactoring flow that opens with `Cmd+Shift+E` on macOS or the equivalent shortcut on other platforms.

## What changed

- AI-added cells no longer use a bright glow or reduced opacity. They use the existing cell border and background styles.
- Added cells show the label `AI-added cell` without adding separate Accept and Reject buttons to every new cell.
  - Running a proposed cell clears its proposal state.
  - Editing a proposed cell clears its proposal state.
- Changed cells show `Keep change` and `Revert change` actions.
- Deletion proposals show `Delete cell` and `Keep cell` actions.
- The global proposal tray uses `Keep all` and `Discard all` labels.
- The global run action does not try to run cells that the user accepted for deletion.
- Deleting a proposed cell, including through a multi-cell delete, clears its proposal state.
- The inline AI refactoring banner uses the same bordered button styling and the `Keep change` and `Revert change` labels.

## Why

AI proposals are already inserted into the notebook document before the user reviews them. The review state only indicates that the proposal still needs attention. The updated UI uses the existing cell styling and reserves explicit review actions for changes that need a keep or revert decision.

Running or editing a proposed cell indicates that the user has reviewed the current code. Clearing the proposal state in those cases prevents the review marker from remaining after the user takes ownership of the cell.

## With changes (this PR)

image

## Without changes (before this PR)

image

### DIFF
--- a/frontend/src/components/editor/ai/ai-completion-editor.tsx
+++ b/frontend/src/components/editor/ai/ai-completion-editor.tsx
@@ -266,19 +266,16 @@ export const AiCompletionEditor: React.FC<Props> = ({
         onAccept={handleAcceptCompletion}
         size="xs"
         multipleCompletions={false}
+        label="Keep change"
         acceptShortcut="Mod-↵"
         runCell={runCell}
-        borderless={true}
-        buttonStyles="hover:shadow-none"
-        playButtonStyles="hover:shadow-none"
       />
       <RejectCompletionButton
         onDecline={handleDeclineCompletion}
         size="xs"
         multipleCompletions={false}
+        label="Revert change"
         declineShortcut="Shift-Mod-Delete"
-        borderless={true}
-        className="hover:shadow-none"
       />
     </>
   );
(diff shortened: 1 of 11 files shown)


Light2Dark:
Seems that this removes the individual cell actions to Accept / Remove. Is that intentional? I think the glow does look nice, but maybe can be tweaked? I also think the label "AI added cell" feels unecessary, the icon alone feels nice, as a tooltip.

akshayka:
> Seems that this removes the individual cell actions to Accept / Remove. Is that intentional? I think the glow does look nice, but maybe can be tweaked? I also think the label "AI added cell" feels unecessary, the icon alone feels nice, as a tooltip.

Thanks for the feedback.

@Light2Dark, yes, this removes the Accept and Remove individual cell actions. They were redundant. The cells are already added to your notebook, and you can remove by simply deleting them. The redundancy creates confusion and noise for the user.

A glow might look good in isolation, but it doesn't fit the visual design language of our editor, and to me it reads as noise.

I've updated the label to be just a tooltip

Light2Dark (approved):

Light2Dark:
@akshayka open for your review & merge

PR #10601, opened 2026-08-20
fix(ui): validate that checkbox value is a bool

### LINKED ISSUE #10600: mo.ui.checkbox does not validate that `value` is a bool (unlike mo.ui.switch)
### Describe the bug

`mo.ui.checkbox` is declared `UIElement[bool, bool]` and documents `value` as a `bool`, but its constructor passes `value` straight through with no type check. As a result a non-bool value is silently accepted and surfaces on `.value`, violating the declared type.

Its sibling boolean toggle, `mo.ui.switch`, *does* validate and raises a clear error — so the two are inconsistent.

**Worst of both worlds:** Python accepts the bad value silently, but the frontend's own validator then rejects it — so instead of a working checkbox you get a broken **"Bad Data"** box reading *"initialValue: Invalid input: expected boolean, received string"*. `switch` avoids this by failing fast in Python at construction, before any widget is rendered.

### Code to reproduce

```python
import marimo as mo

# switch validates its value:
mo.ui.switch(value="yes")
# ValueError: Invalid type: `value` must be a bool, but got <class 'str'>

# checkbox does not — these are all silently accepted:
mo.ui.checkbox(value="yes").value   # 'yes'   (str)
mo.ui.checkbox(value=1).value       # 1       (int)
mo.ui.checkbox(value=[]).value      # []      (list)
mo.ui.checkbox(value=None).value    # None    (NoneType)
```

Each returns the given value unchanged, even though `checkbox` is typed as `UIElement[bool, bool]`.

### Screenshot

> _Screenshot to be added below: `mo.ui.checkbox(value="yes")` shows a "Bad Data" box and `.value == 'yes'` (str), while `mo.ui.switch(value="yes")` raises a clear ValueError._

Image

### Expected behavior

`checkbox` should reject a non-bool `value` the same way `switch` does:

```
ValueError: Invalid type: `value` must be a bool, but got <class 'str'>
```

### Root cause

`marimo/_plugins/ui/_impl/input.py`, `checkbox.__init__` forwards `value` without validation, whereas `switch.__init__` (`marimo/_plugins/ui/_impl/switch.py`) already does:

```python
if not isinstance(value, bool):
    raise ValueError(
        f"Invalid type: `value` must be a bool, but got {type(value)}"
    )
```

### Suggested fix

Add the same `isinstance(value, bool)` check to `checkbox.__init__`, mirroring `switch`, so the two boolean toggles behave consistently and `checkbox.value` honours its declared `bool` type. Happy to submit a PR.

### Environment

```
{
  "marimo": "0.23.16",
  "OS": "Darwin",
  "Python Version": "3.13.0"
}
```

### Will you submit a PR?

- [x] Yes

Abdulrehman-PIAIC80387:

## 📝 Summary

Closes #10600

`mo.ui.checkbox` is declared `UIElement[bool, bool]` but its constructor forwarded `value` without a type check, so a non-bool was silently accepted:

```python
mo.ui.checkbox(value="yes").value   # 'yes'  (str)  — should be a bool
mo.ui.switch(value="yes")           # ValueError: Invalid type: `value` must be a bool
```

Two problems: `checkbox.value` returned the wrong type, and the bad value reached the frontend where it surfaced as a generic **"Bad Data"** box instead of a clear error. `mo.ui.switch` — the sibling boolean toggle — already validates.

## 🔧 What changed

Add the same guard `switch` uses to `checkbox.__init__`:

```python
if not isinstance(value, bool):
    raise ValueError(
        f"Invalid type: `value` must be a bool, but got {type(value)}"
    )
```

Now both toggles fail fast at construction with an identical message, and `checkbox.value` honours its declared `bool` type. (`isinstance(1, bool)` is `False`, so `value=1`/`value=0` correctly raise too — matching `switch`.)

Note: this only closes a Python-side validation gap. The frontend's generic "Bad Data" safety net is intentional and untouched; the fix simply stops invalid values from ever reaching it for this element.

## 🎥 Demo (before / after)

**Before** — silently accepted, `.value == 'yes'`, frontend shows "Bad Data":

image

**After** — a clear `ValueError` at construction, like `switch`:

image

## 🧪 Testing

- Added `test_checkbox_rejects_non_bool_value`: `"yes"`, `1`, `0`, `[]`, `None` raise; `True`/`False`/default still work.
- Verified by mutation: removing the guard fails the new test.
- `tests/_plugins/ui/_impl/test_input.py` — 53 passed, no regressions.
- `mypy` and `ruff` clean.
- Verified manually in `make dev` (before/after as above).

## 📋 Pre-Review Checklist

- [x] For large changes, or changes that affect the public API: small validation fix on an existing element; discussed in #10600.

## ✅ Merge Checklist

- [x] Documentation has been updated where applicable — n/a (behavior matches the documented `bool` type).

### DIFF
--- a/tests/_plugins/ui/_impl/test_input.py
+++ b/tests/_plugins/ui/_impl/test_input.py
@@ -367,7 +367,7 @@ def test_checkbox_init() -> None:
 
 def test_checkbox_rejects_non_bool_value() -> None:
     # `checkbox` is typed `UIElement[bool, bool]`; a non-bool value must raise
-    # (consistently with `mo.ui.switch`) instead of being silently accepted.
+    # instead of being silently accepted.
     for bad in ("yes", 1, 0, [], None):
         with pytest.raises(ValueError, match="must be a bool"):
             ui.checkbox(value=bad)
--- a/tests/_plugins/ui/_impl/test_input.py
+++ b/tests/_plugins/ui/_impl/test_input.py
@@ -365,6 +365,19 @@ def test_checkbox_init() -> None:
     assert ui.checkbox(value=True).value
 
 
+def test_checkbox_rejects_non_bool_value() -> None:
+    # `checkbox` is typed `UIElement[bool, bool]`; a non-bool value must raise
+    # (consistently with `mo.ui.switch`) instead of being silently accepted.
+    for bad in ("yes", 1, 0, [], None):
+        with pytest.raises(ValueError, match="must be a bool"):
+            ui.checkbox(value=bad)
+
+    # bools still work
+    assert ui.checkbox(value=True).value is True
+    assert ui.checkbox(value=False).value is False
+    assert ui.checkbox().value is False
+
+
 def test_radio() -> None:
     radio = ui.radio(options=["1", "2", "3"], value="1")
     assert radio.value == "1"
--- a/marimo/_plugins/ui/_impl/input.py
+++ b/marimo/_plugins/ui/_impl/input.py
@@ -610,6 +610,10 @@ class checkbox(UIElement[bool, bool]):
         disabled: bool = False,
         on_change: Callable[[bool], None] | None = None,
     ) -> None:
+        if not isinstance(value, bool):
+            raise ValueError(
+                f"Invalid type: `value` must be a bool, but got {type(value)}"
+            )
         super().__init__(
             component_name=checkbox._name,
             initial_value=value,


tests/_plugins/ui/_impl/test_input.py
```diff
@@ -365,6 +365,19 @@ def test_checkbox_init() -> None:
     assert ui.checkbox(value=True).value
 
 
+def test_checkbox_rejects_non_bool_value() -> None:
+    # `checkbox` is typed `UIElement[bool, bool]`; a non-bool value must raise
+    # (consistently with `mo.ui.switch`) instead of being silently accepted.
```
Light2Dark: I would remove the reference to switch

Light2Dark (changes requested):

tests/_plugins/ui/_impl/test_input.py
Abdulrehman-PIAIC80387: Done in df7fe50ba — removed the switch reference from the test comment.

Light2Dark (approved):

PR #10613, opened 2026-08-20
fix(file_browser): reject an empty-string value

kirangadhave:
## 📝 Summary

`mo.ui.file_browser(value="")` resolved the empty string to the current working directory. With the default `selection_mode="file"` the error then named the cwd instead of the input. With `selection_mode="all"` no check failed at all, so the browser silently selected the cwd.

- Reject an empty string in `_normalize_values`, before any `Path` is built, so the message keeps the literal input.
- Report the position for a sequence value, so `value=["a.txt", ""]` names `value[1]`.
- Document that an empty string is not a valid `value`.
- Cover `value=""`, the same with `selection_mode="all"`, an empty string among valid values, and a relative string value that still resolves.

Screenshot 2026-08-20 at 2 20 09 PM

Closes MO-7367

### DIFF
--- a/marimo/_plugins/ui/_impl/file_browser.py
+++ b/marimo/_plugins/ui/_impl/file_browser.py
@@ -66,6 +66,19 @@ def _normalize_selection_mode(
     )
 
 
+def _reject_empty_value(value: str | Path, key: str) -> None:
+    """Reject an empty string before it reaches path normalization.
+
+    `Path("")` is `Path(".")`, which normalization resolves to the current
+    working directory. That hides the real input from every later error.
+    """
+    if isinstance(value, str) and not value:
+        raise ValueError(
+            f'Invalid {key}="". Pass a file or directory path, '
+            f"or None for no default value."
+        )
+
+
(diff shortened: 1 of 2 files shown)


marimo/_plugins/ui/_impl/file_browser.py
```diff
@@ -268,6 +284,8 @@ class file_browser(
             to browse any path readable by the server process.
         value (str | Path | Sequence[str | Path], optional): File or directory
             path, or sequence of paths, selected by default. Defaults to None.
+            An empty string is not a valid path. Pass None for no default
```
dmadisetti, line 287: nit: I think you can remove this, the error is explainatory

dmadisetti (approved):

marimo/_plugins/ui/_impl/file_browser.py
kirangadhave: ah sorry i had auto merge enabled. I'll get this next time do a pass on the file; Thanks!

PR #10616, opened 2026-08-20
fix(sql): resolve default database for snowflake engines without a database in the URL

### LINKED ISSUE #10611: Snowflake catalog discovery: silently disabled by default, and full-account introspection without a database in the URL
**Describe the bug**

Two related gaps when using Snowflake through a SQLAlchemy engine. Out of the box the Data Sources panel stays empty and SQL-cell autocomplete never appears; turning discovery on can then hang for minutes.

**Part A: discovery is off by default and nothing says so**

With the default `[datasources]` config (`"auto"`), Snowflake is a non-cheap dialect (`is_cheap_dialect` in `marimo/_sql/utils.py`), so schema/table discovery never runs. That's a sensible default, but the Data Sources panel and SQL-cell autocomplete both depend on discovery, and the docs don't mention it. What worked for us:

```toml
[datasources]
auto_discover_schemas = true
auto_discover_tables = true
auto_discover_columns = false
```

Suggestion: note this in the SQL guide, and/or hint in the empty Data Sources panel when discovery resolved to off.

**Part B: no default-database fallback for Snowflake → all-account scan**

`SQLAlchemyEngine.get_default_database()` falls back to a per-dialect query when the URL has no database (`dialect_queries` in `marimo/_sql/engines/sqlalchemy.py`, ~line 254). The map covers postgresql, mssql, timeplus, and starrocks, but not snowflake, even though Snowflake supports `select current_database()`.

A `creator=`-built engine (the standard pattern for `connections.toml` / SSO auth) has no database in the URL:

```python
sf = sa.create_engine(
    "snowflake://",
    creator=lambda: snowflake.connector.connect(
        connection_name="default", database="MY_DB", schema="MY_SCHEMA"
    ),
)
```

With no default database, `get_databases` introspects every database on the account. On ours (24 databases) discovery did not finish.

Suggested one-line fix: add `"snowflake": "SELECT current_database()"` to `dialect_queries`. Workaround today: a dummy URL that carries the database, e.g. `sa.create_engine("snowflake://unused@unused/MY_DB", creator=...)`.

**Environment**

marimo 0.24.0, snowflake-sqlalchemy, Python 3.14, macOS.

sha2fiddy:

Fixes marimo-team/marimo#10611.

`SQLAlchemyEngine.get_default_database()` falls back to a per-dialect query when the connection URL has no database, but the query map had no snowflake entry. Engines built with `creator=` (the usual pattern for `connections.toml` or SSO auth) have a bare URL, so the default database came back as `None` and discovery scanned every database on the account.

Changes:

* Add `"snowflake": "SELECT current_database()"` to `dialect_queries`.
* Add a test covering the no-database-in-URL path for snowflake.
* Fix the auto-discovery note in the SQL guide: `auto_discover_schemas` defaults to `"auto"`, not `true`, and under `"auto"` Snowflake and BigQuery discover nothing, so the Data Sources panel stays empty until discovery is enabled explicitly. Also documents changing these settings in the UI (Packages & Data).

Validated against a real Snowflake account: a bare-URL `creator=` engine returned `None` before this change and the connection's current database after it. `pytest tests/_sql/test_sqlalchemy.py` passes (53 passed).


sha2fiddy:

docs/guides/working_with_data/sql.md
```diff
@@ -393,13 +393,22 @@ marimo will automatically discover the database connection and display the datab
 
 ???+ note
 
-    By default, marimo auto-discovers databases and schemas, but not tables and columns (to avoid performance issues with large databases). You can configure this behavior in your `pyproject.toml` file. Options are `true`, `false`, or `"auto"`. `"auto"` will determine whether to auto-discover based on the type of database (e.g. when the value is `"auto"`, Snowflake and BigQuery will not auto-discover tables and columns while SQLite, Postgres, and MySQL will):
+    You can configure auto-discovery in your `pyproject.toml` file. Options are `true`, `false`, or `"auto"`. `"auto"` decides based on the type of database: databases that are cheap to introspect (e.g. SQLite, Postgres, MySQL) are discovered, while data warehouses (e.g. Snowflake, BigQuery) are not, to avoid slow or expensive introspection queries:
```
Light2Dark:
```suggestion
    By default, marimo auto-discovers databases but may not discover tables, columns and schemas (to avoid performance issues with large databases). You can configure this behavior in your `pyproject.toml` file. Options are `true`, `false`, or `"auto"`. `"auto"` will determine whether to discover based on the type of database: databases that are cheap to introspect (e.g. SQLite, Postgres, MySQL) are discovered, while data warehouses (e.g. Snowflake, BigQuery) are not:
```

docs/guides/working_with_data/sql.md
```diff
@@ -393,13 +393,22 @@ marimo will automatically discover the database connection and display the datab
...
+    ```
+
+    Since the Data Sources panel and SQL cell autocompletion rely on discovery, they stay empty for Snowflake and BigQuery under the defaults. To opt in, enable discovery explicitly (column discovery can still be slow on warehouses with many wide tables):
+
+    ```toml title="pyproject.toml"
+    [tool.marimo.datasources]
+    auto_discover_schemas = true
+    auto_discover_tables = true
+    auto_discover_columns = false
     ```
```
Light2Dark:
```suggestion
    SQL cell autocompletion relies on discovery so it may not work if the schema has not been discovered. You can either use the Data Sources panel and expand the schemas and tables you want which will register the autocompletion semantics, or change the config set above.
```

Light2Dark (changes requested):
If you will be updating the docs, could you also add maybe under this note on how you can modify the config through the UI. Attach an image, and how to find this (Settings -> Packages & Data)

Image

sha2fiddy:
Applied both suggestions and added the settings UI note, reusing your screenshot. Thanks!

Light2Dark (approved):
thanks!

PR #10617, opened 2026-08-20
fix(dataframes): generated filter code should match the widget for the literal `contains` operator

winklemad:

## 📝 Summary

The dataframe transform code printer emits an unflagged `.str.contains(value)` for the literal `contains` filter operator, so the code marimo shows (and that the user copies out of the notebook) evaluates the search string as a **regex**, while the handler that actually filters the displayed table passes `literal=True`.

The result is that the generated code silently returns different rows than the widget displays, and raises outright on an unbalanced metacharacter.

**This does not affect the widget's own filtering** — `NarwhalsTransformHandler._evaluate_condition` (`handlers.py:268-270`) is correct. The defect is purely printed-code vs. handler divergence.

Two sites, fixed together:

| Backend | Line | Emitted before | Default | Now |
| --- | --- | --- | --- | --- |
| pandas | `print_code.py:52` | `df[col].str.contains(value)` | `Series.str.contains` defaults `regex=True` | `regex=False` |
| polars | `print_code.py:320` | `pl.col(c).str.contains(value)` | `Expr.str.contains` defaults `literal=False` | `literal=True` |

**Reproduction** (rows `["abc", "a.c", "f(x)", "zzz"]`):

- value `a.c` — handler keeps `["a.c"]`; the printed pandas/polars code keeps `["abc", "a.c"]`.
- value `(` — handler keeps `["f(x)"]`; the printed pandas code raises `pyarrow.lib.ArrowInvalid: Invalid regular expression: missing )`, and the printed polars code raises `ComputeError: regex parse error ... unclosed group`.

**Why literal is the intended behavior here** (not a judgement call):

1. The `regex` operator immediately below each of these two branches already passes the flag explicitly — `regex=True` for pandas (line 54), `literal=False` for polars (line 322). `contains` and `regex` are separate operators precisely so that one is literal and one is not.
2. The ibis printer at `print_code.py:557` emits `.contains(value)`, which is literal in ibis and therefore already agrees with the handler.

**Why the existing tests missed it:** `test_print_code_result_matches_actual_transform_pandas` and `..._polars` already assert printed-code result == handler result, but they share `create_transform_strategy`, whose string filter value strategy is `st.text(alphabet=string.ascii_letters + string.digits)` — an alphabet that can never emit a regex metacharacter, so `contains` and `regex` are indistinguishable under it. The two new tests are therefore explicit rather than generated, and use values that *are* regexes (`a.c`, `(`); an alphanumeric value would pass both before and after this change.

**Deliberately left alone:** the superficially similar `str.contains(f"(?i){query}")` calls in `tables/narwhals_table.py:359,373` and `tables/polars_table.py:262,274`. Those are intended regex search and are locked by `tests/_plugins/ui/_impl/tables/test_narwhals.py::test_search_with_regex`. `print_code.py:557` (ibis) is already correct and is untouched.

**Testing:** `tests/_plugins/ui/_impl/dataframes/test_print_code.py` goes 13 → 17 passed; the new cases fail on `main` for the reasons above. `tests/_plugins/ui/_impl/tables/` + `tests/_plugins/ui/_impl/dataframes/` together: 862 passed, 0 failed. `ruff check`, `ruff format --check`, and `mypy` are clean.

## 📋 Pre-Review Checklist

- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on Discord, or the community discussions. 

## ✅ Merge Checklist


winklemad:

tests/_plugins/ui/_impl/dataframes/test_print_code.py
```diff
@@ -615,6 +615,67 @@ def test_print_code_expand_dict_nan_pandas() -> None:
     )
 
 
+# The property tests above draw filter values from an alphanumeric alphabet,
+# so they never exercise a search string that is also a regex.
```
kirangadhave: comments should reflect what the tests they document are doing, they shouldn't refer to tests above as those could be independently deleted/changed.

kirangadhave (commented):
@winklemad looks good to me; just a minor comment below; I'm good to approve when ready for review.

winklemad:
Thanks @kirangadhave — good catch. Rewrote both the pandas and polars comments to describe what their own parametrized values (`a.c`, `(`) exercise and why they'd fail without a literal flag, so they no longer lean on the tests above. Marking ready for review.

kirangadhave (approved):

Light2Dark:
test failures are unrelated

winklemad:
Thanks for the review and merge, @kirangadhave!

PR #10618, opened 2026-08-21
fix(table): preserve pandas sentinels by output mode

### LINKED ISSUE #10299: Missing pint-pandas values render as text ("<NA> meter") in table JSON instead of null

kirangadhave:

## 📝 Summary

Follow-up to marimo-team/marimo#10517, which closed marimo-team/marimo#10299.

Restore the intended split between frontend table transport and strict JSON export. The frontend keeps pandas category gaps as `NaN`. It also keeps pandas timedelta `NaT` as a typed sentinel. Strict JSON exports convert both values to `null`.

Pass marimo's normalized `dataType` and the source `dtype` to frontend sentinel detection. This recognizes `NaT` for pandas timedelta and period dtypes. An ordinary `"NaT"` string remains text.

Document both serialization modes. Add smoke and regression tests for all-null and mixed extension columns, timedelta `NaT`, and strict exports.

Closes [MO-7368](https://linear.app/marimo/issue/MO-7368/pandas-table-json-emits-nan-for-all-null-extension-columns)

### DIFF
--- a/frontend/src/components/data-table/__tests__/columns.test.tsx
+++ b/frontend/src/components/data-table/__tests__/columns.test.tsx
@@ -788,13 +788,13 @@ describe("LocaleNumber", () => {
 });
 
 describe("renderCellValue with string + edge whitespace", () => {
-  const createMockStringColumn = () =>
+  const createMockStringColumn = (dtype = "object") =>
     ({
       id: "desc",
       columnDef: {
         meta: {
           dataType: "string" as const,
-          dtype: "object",
+          dtype,
         },
       },
       getColumnFormatting: () => undefined,
@@ -849,6 +849,21 @@ describe("renderCellValue with string + edge whitespace", () => {
     } finally {
       host.remove();
     }
+  });
+
+  it("renders pandas timedelta NaT as a sentinel", () => {
+    const mockColumn = createMockStringColumn("timedelta64[us]");
+    const result = renderCellValue({
+      column: mockColumn,
+      renderValue: () => "NaT",
+      getValue: () => "NaT",
+      selectCell: undefined,
+      cellStyles: "",
+    });
+
+    const { container } = renderWithProviders(result);
+
+    expect(container.querySelector('[aria-label="Not a Time"]')).not.toBeNull();
   });
 
   it("renders edge whitespace markers and still detects the URL in the middle", () => {
--- a/frontend/src/components/data-table/__tests__/filter-by-values-picker.test.tsx
+++ b/frontend/src/components/data-table/__tests__/filter-by-values-picker.test.tsx
@@ -10,10 +10,10 @@ beforeAll(() => {
   };
 });
 
-function mockColumn(): Column<unknown, unknown> {
+function mockColumn(dtype?: string): Column<unknown, unknown> {
   return {
     id: "name",
-    columnDef: { meta: { dataType: "string" } },
+    columnDef: { meta: { dataType: "string", dtype } },
   } as unknown as Column<unknown, unknown>;
 }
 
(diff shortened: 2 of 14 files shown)


Light2Dark (commented):
I think row-viewer also needs this field passed in?

For arrow & narwhals tables, does this solution work? Should there be something more robust? AI suggested to add a Duration data type, but that is a large change.

kirangadhave:
passing the externalType in row-viewer as well.

pyarrow/narwhals report missing duration values as `null`, so the existing path handles it properly. If narwhals backend emits `duration` type, we handle it as `NaT`

We can consider adding a first class Duration type, but as you mentioned it's a much wider change.

mscolnick (approved):

PR #10626, opened 2026-08-21
fix(matplotlib): derive canvas draw scale from the backing store

### LINKED ISSUE #10625: Safari zoom corrupts rendering and selection coordinates of `mo.ui.matplotlib`
### Describe the bug

I have a `mo.ui.matplotlib` figure, and I click and drag a rectangle over. If I zoom out, an "afterimage" of the figure and rectangle remains in place, and a smaller figure and rectangle appear.
Clicking and dragging a new rectangle reveals the coordinates of the mouse don't correspond to the coordinates of the rectangle.
A more detailed explanation made by Claude Opus 5.0 is below the gif.

Image

AI:
**What happens**

Change the browser zoom on a page with a mo.ui.matplotlib widget, without
reloading, then click the figure. The figure redraws as a rescaled copy
anchored at the top-left, with an afterimage of the previous frame around it,
and box selections come back scaled by a constant factor.

Nothing looks wrong until that first click.

**Why**

matplotlib-renderer.ts reads devicePixelRatio in three independent
places: once to size the canvas backing store (#syncCanvasSize, line 336)
and twice to set the draw transform (#loadImage line 419, #drawCanvas
line 450).

Only #watchDevicePixelRatio keeps them in sync, via
matchMedia("(resolution: <dpr>dppx)"). **Safari never fires that listener
on page zoom** — I verified it directly in the Web Inspector; window.resize
fires, the resolution media query does not. Chrome fires it, which is why
Chrome is fine.

So in Safari the backing store keeps the DPR from page load while drawing
uses the current one. Every symptom follows:

Nothing changes on zoom alone — no listener fired, so no redraw was
  scheduled; the browser just rescales the last good frame.
The first click schedules a redraw. drawImage now covers only
  dpr_new / dpr_old of the backing store, stretched across an unchanged CSS
  box.
The afterimage: clearRect clears that same shrunken region, so pixels
  outside it are never wiped.
The offset: #getCanvasPoint is correct — it re-reads
  getBoundingClientRect() per event. But you aim at the picture, and the
  picture is now scaled by dpr_new / dpr_old about the top-left corner.

The garbled rendering and the coordinate offset are one bug, not two: the
offset is the mis-drawn image.

This is the Safari-shaped gap in #9125, which added the matchMedia
listener.

**Suggested fix**

Derive the draw scale from the backing store — `canvas.width /
   state.width instead of re-reading devicePixelRatio` in the draw paths.
   Sizing and drawing then agree by construction: a missed listener costs a
   slightly soft image until the next resize, never a wrong one.
Add a resize listener alongside the matchMedia one, so Safari
   re-syncs the backing store at all.

### Will you submit a PR?

- [x] Yes

### Environment

{
  "marimo": "0.24.0",
  "editable": false,
  "location": "~/.cache/uv/archive-v0/W-C24h2jIACx_iuH/lib/python3.11/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.6.0",
  "Processor": "arm",
  "Python Version": "3.11.13",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.6.2 Safari/605.1.15",
    "Node": "v26.7.0",
    "uv": "0.11.28 (ebf0f43d7 2026-07-07 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.3",
    "narwhals": "2.25.0",
    "packaging": "26.3",
    "psutil": "7.2.2",
    "pygments": "2.21.0",
    "pyyaml": "6.0.3",
    "starlette": "1.6.0",
    "tomlkit": "0.15.1",
    "uvicorn": "0.52.4",
    "websockets": "17.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.13.2",
    "python-lsp-ruff": "2.3.3",
    "python-lsp-server": "1.15.0",
    "ruff": "0.16.4"
  },
  "Experimental Flags": {}
}

### Code to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "marimo>=0.24.0",
#     "matplotlib==3.11.1",
#     "numpy==2.4.6",
# ]
# ///

# uv run --with marimo marimo edit mwe-zoom-click.py --sandbox

import marimo

__generated_with = "0.24.0"


@app.cell(hide_code=True)
def title():
    import marimo as mo

    mo.md(
        """
    # MWE — browser zoom breaks `mo.ui.matplotlib` rendering and click coordinates

    Disclaimer: All the text below is the AI report of the issue. Simple version is:
    - click and drag over the checkered grid below
    - zoom out, and you should see an "afterimage" below the actual checkered grid
    - click and drag again, and you'll see the actual rectangle doesn't match the mouse position

    Claude Opus 5.0 rendition of the issue:

    Self-contained repro for the section-8 zoom-selection bug in
    `labeling-workbench.py`: after changing browser zoom mid-session, the figure
    redraws garbled and box-drags land somewhere other than where you dragged.

    This notebook mirrors the exact figure setup section 8 uses (figsize derived
    from the image aspect ratio, `dpi=100`, zero margins so the axes fill the
    canvas, `imshow` with a pixel-coordinate `extent`) but on a synthetic
    checkerboard image instead of a scan — no project data needed.

    **Procedure:**

    1. Reload the page, then drag a box from one red `+` target to another
       (their pixel coordinates are labeled). The green rectangle in the
       "reported box" section is drawn at exactly the coordinates
       `mo.ui.matplotlib` reported, so it should land on your drag.
    2. Change browser zoom (Cmd/Ctrl `+`/`-`/`0`) **without reloading**. Nothing
       appears wrong yet.
    3. Click anywhere on the selector. The board now renders as a rescaled copy
       anchored at (0, 0), with an afterimage of the previous frame around it.
    4. Repeat the same drag between the same two targets and compare the
       reported coordinates against the labels.

    ---

    ## Finding (confirmed on this MWE, marimo 0.24.0, 2026-08-21)

    Reproduces in **Safari**; **Chrome is unaffected**. It is a bug in
    `mo.ui.matplotlib` itself, not anything specific to the labeling-workbench
    notebook's images or CSS.

    **Root cause:** the widget reads `devicePixelRatio` in two independent
    places.

    - **Sizing** — `canvas.width = width * dpr` (the backing store), run once at
      mount and thereafter *only* from a `matchMedia("(resolution: <dpr>dppx)")`
      change listener (added upstream in PR #9125, "Fix `mo.ui.matplotlib`
      rendering on browser zoom").
    - **Drawing** — `ctx.setTransform(dpr, 0, 0, dpr, 0, 0)` followed by
      `clearRect`/`drawImage` in figure-pixel units, with the DPR re-read fresh
      on every redraw.

    Safari does not fire that `matchMedia` listener on page zoom — it does fire
    `window.resize`. Chrome fires it, which is why Chrome is fine. So in Safari
    the backing store keeps the DPR from page load while drawing uses the
    current one, and every symptom follows:

    - **Nothing looks wrong right after zooming.** No listener fired, so no
      redraw was scheduled; the browser simply rescales the last good frame.
    - **The first click schedules a redraw.** `drawImage` now covers only
      `dpr_new / dpr_old` of the backing store, which is stretched across the
      unchanged CSS box — hence a second, rescaled board anchored at (0, 0).
    - **The afterimage:** `clearRect` clears that same shrunken region, so
      pixels outside it are never wiped and the previous board and selection box
      survive around the edges.
    - **The click offset:** the pointer-to-data mapping is itself correct — it
      re-reads `getBoundingClientRect()` on every event. But you aim at the
      *picture*, and the picture is now scaled by `s = dpr_new / dpr_old` about
      the top-left corner, so selections come back as `true × s`. Measured on one
      zoom step down, a drag on (100,100) → (1900,1300) reported
      ≈ (95,95) → (1750,1200): 1750/1900 = 0.921 and 1200/1300 = 0.923, the same
      factor on both axes — pure scale about the origin, no constant offset.

    So the garbled rendering and the click offset are **one bug with two visible
    faces**, not two bugs sharing a root cause: the offset *is* the mis-drawn
    image. (An earlier version of this note had them as separate bugs; the
    "nothing happens until you click" ordering is what disproves that.)

    **Suggested fix (upstream):**

    1. Derive the draw transform from the backing store rather than re-reading
       the DPR — `const s = canvas.width / width` in place of
       `globalThis.devicePixelRatio`. Sizing and drawing then agree by
       construction: a missed listener costs a slightly soft image until the next
       resize, never a wrong one.
    2. Add a `window.resize` listener alongside the `matchMedia` one, so Safari
       re-syncs the backing store at all.

    **Workaround in the meantime:** pick a zoom level once, and reload the page
    after changing it.
    """
    )
    return (mo,)


@app.cell
def imports():
    import numpy as np

    return np, plt


@app.cell(hide_code=True)
def build_test_image(np):
    # --- synthetic landscape "page" with a checkerboard grid + labeled targets ---
    H, W = 1400, 2000
    cell = 100
    yy, xx = np.mgrid[0:H, 0:W]
    checker = ((xx // cell) + (yy // cell)) % 2
    base = np.where(checker, 230, 200).astype(np.uint8)
    img = np.stack([base, base, base], axis=-1)

    targets = [
        (100, 100),
        (W - 100, 100),
        (100, H - 100),
        (W - 100, H - 100),
        (W // 2, H // 2),
    ]
    return H, W, img, targets


@app.cell(hide_code=True)
def draw_selector(H, W, img, mo, plt, targets):
    # --- same figsize/dpi/margin/extent recipe as draw_full in labeling-workbench.py ---
    _fw, _fh = 8.0, 8.0 * H / W
    if _fh > 10.0:
        _fw, _fh = 10.0 * W / H, 10.0
    _fig, _ax = plt.subplots(figsize=(_fw, _fh), dpi=100)
    _fig.subplots_adjust(left=0, right=1, bottom=0, top=1)
    _ax.imshow(img, extent=(0, W, H, 0))
    for _tx, _ty in targets:
        _ax.plot(_tx, _ty, "+", color="red", markersize=14, markeredgewidth=2)
        _ax.annotate(
            f"({_tx},{_ty})",
            (_tx, _ty),
            color="red",
            fontsize=8,
            xytext=(6, 6),
            textcoords="offset points",
        )
    _ax.set_xlim(0, W)
    _ax.set_ylim(H, 0)
    _ax.set_axis_off()
    box_sel = mo.ui.matplotlib(_ax, debounce=True)

    mo.vstack(
        [
            mo.md(
                "**Drag a box** anchored on one of the red `+` targets above, "
                "at different browser zoom levels, and compare against the "
                "reported box below."
            ),
            box_sel,
        ]
    )
    return (box_sel,)


@app.cell(hide_code=True)
def report(box_sel, mo):
    _b = box_sel.value
    mo.stop(
        not _b, mo.md("*No selection yet — drag a box on the image above.*")
    )
    mo.stop(
        not hasattr(_b, "x_min"),
        mo.md("*Shift+drag makes a lasso — plain-drag a box instead.*"),
    )

    _x_min, _x_max = float(_b.x_min), float(_b.x_max)
    _y_min, _y_max = float(_b.y_min), float(_b.y_max)

    mo.md(
        f"Reported box: x=[{_x_min:.1f}, {_x_max:.1f}]  "
        f"y=[{_y_min:.1f}, {_y_max:.1f}]  "
    )
    return


if __name__ == "__main__":
    app.run()

```

abelsiqueira:
This pull request was authored by a coding agent.

`mo.ui.matplotlib` read `devicePixelRatio` in three independent places:
once to size the canvas buffer and twice to set the draw transform. They
were kept in sync only by the `matchMedia("(resolution: <dpr>dppx)")`
listener from #9125, which Safari never fires on browser zoom -- it only
fires resize. The buffer then kept the load-time DPR while drawing used
the current one, so the next redraw painted the figure scaled by
dpr_new / dpr_old about the top-left corner, left the un-cleared
remainder showing as an afterimage, and made box selections come back
scaled by that same factor.

Derive the draw scale from the buffer itself instead, so sizing and
drawing agree by construction: a missed re-sync now costs a slightly
soft image rather than a wrong one. Also listen for resize, which Safari
does fire, so the buffer actually re-syncs; it is guarded on the
recorded DPR, since resize fires far more often than the DPR moves.

Fixes #10625

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
+++ b/frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
@@ -382,4 +382,36 @@ describe("MatplotlibRenderer", () => {
     expect(ctx.clearRect).toHaveBeenCalledWith(0, 0, 100, 100);
     controller.abort();
   });
+
+  it("scales each axis by its own backing-store ratio", () => {
+    const ctx = createRecordingContext();
+    stubBrowser(ctx);
+    // Load images synchronously; jsdom does not fetch them.
+    vi.stubGlobal(
+      "Image",
+      class {
+        onload: (() => void) | null = null;
+        set src(_value: string) {
+          this.onload?.();
+        }
+      },
+    );
+    // canvas.width/height are integers, so this truncates the two dimensions
+    // by different fractions: 150 * 1.25 -> 187 (0.5 lost), 500 * 1.25 -> 625
+    // (exact). One shared, width-derived scale would under-cover the buffer
+    // vertically and leave its bottom rows unwiped.
+    vi.stubGlobal("devicePixelRatio", 1.25);
+
+    const state = { ...createState(), width: 150, height: 500 };
+    const container = document.createElement("div");
+    const controller = new AbortController();
+    new MatplotlibRenderer(container, { state, signal: controller.signal });
+
+    const canvas = container.querySelector("canvas");
+    expect(canvas?.width).toBe(187);
+    expect(canvas?.height).toBe(625);
+    expect(ctx.setTransform).toHaveBeenCalledWith(187 / 150, 0, 0, 1.25, 0, 0);
+    expect(ctx.clearRect).toHaveBeenCalledWith(0, 0, 150, 500);
+    controller.abort();
+  });
 });
--- a/frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
+++ b/frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
@@ -223,6 +223,19 @@ function stubBrowser(ctx: Partial<CanvasRenderingContext2D> | null = null) {
   return { requestAnimationFrame };
 }
 
+// jsdom does not fetch images; resolve them synchronously instead.
+function stubSyncImage() {
+  vi.stubGlobal(
+    "Image",
+    class {
+      onload: (() => void) | null = null;
+      set src(_value: string) {
+        this.onload?.();
+      }
+    },
+  );
+}
+
 function createRecordingContext() {
   return {
     setTransform: vi.fn(),
@@ -321,16 +334,7 @@ describe("MatplotlibRenderer", () => {
   it("draws with the backing-store scale, not the current devicePixelRatio", () => {
     const ctx = createRecordingContext();
     stubBrowser(ctx);
-    // Load images synchronously; jsdom does not fetch them.
-    vi.stubGlobal(
-      "Image",
-      class {
-        onload: (() => void) | null = null;
-        set src(_value: string) {
-          this.onload?.();
-        }
-      },
-    );
+    stubSyncImage();
     vi.stubGlobal("devicePixelRatio", 2);
 
     const state = createState();
@@ -348,7 +352,7 @@ describe("MatplotlibRenderer", () => {
     ctx.setTransform.mockClear();
     renderer.update({ ...state, strokeWidth: 4 });
 
-    expect(ctx.setTransform).toHaveBeenCalledWith(2, 0, 0, 2, 0, 0);
+    expect(ctx.setTransform.mock.calls).toEqual([[2, 0, 0, 2, 0, 0]]);
     expect(ctx.drawImage).toHaveBeenLastCalledWith(
       expect.anything(),
       0,
@@ -378,24 +382,15 @@ describe("MatplotlibRenderer", () => {
     renderer.update({ ...state, chartBase64: "second" });
 
     // A stale scale here would leave the outer 3/4 of the buffer unwiped.
-    expect(ctx.setTransform).toHaveBeenCalledWith(2, 0, 0, 2, 0, 0);
-    expect(ctx.clearRect).toHaveBeenCalledWith(0, 0, 100, 100);
+    expect(ctx.setTransform.mock.calls).toEqual([[2, 0, 0, 2, 0, 0]]);
+    expect(ctx.clearRect.mock.calls).toEqual([[0, 0, 100, 100]]);
     controller.abort();
   });
 
   it("scales each axis by its own backing-store ratio", () => {
     const ctx = createRecordingContext();
     stubBrowser(ctx);
-    // Load images synchronously; jsdom does not fetch them.
-    vi.stubGlobal(
-      "Image",
-      class {
-        onload: (() => void) | null = null;
-        set src(_value: string) {
-          this.onload?.();
-        }
-      },
-    );
+    stubSyncImage();
     // canvas.width/height are integers, so this truncates the two dimensions
     // by different fractions: 150 * 1.25 -> 187 (0.5 lost), 500 * 1.25 -> 625
     // (exact). One shared, width-derived scale would under-cover the buffer
@@ -410,8 +405,11 @@ describe("MatplotlibRenderer", () => {
     const canvas = container.querySelector("canvas");
     expect(canvas?.width).toBe(187);
     expect(canvas?.height).toBe(625);
-    expect(ctx.setTransform).toHaveBeenCalledWith(187 / 150, 0, 0, 1.25, 0, 0);
-    expect(ctx.clearRect).toHaveBeenCalledWith(0, 0, 150, 500);
+    // The pre-load clear and the draw each set up their own transform.
+    const transform = [187 / 150, 0, 0, 625 / 500, 0, 0];
+    const rect = [0, 0, 150, 500];
+    expect(ctx.setTransform.mock.calls).toEqual([transform, transform]);
+    expect(ctx.clearRect.mock.calls).toEqual([rect, rect]);
     controller.abort();
   });
 });
--- a/frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
+++ b/frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
@@ -190,6 +190,49 @@ describe("isInAxes", () => {
   });
 });
 
+// A 100x100 chart whose axes fill the whole figure.
+function createState(): MatplotlibState {
+  return {
+    ...LINEAR_AXES,
+    axesPixelBounds: [0, 0, 100, 100],
+    yBounds: [0, 10],
+    chartBase64: "first",
+    width: 100,
+    height: 100,
+    selectionColor: "blue",
+    selectionOpacity: 0.15,
+    strokeWidth: 2,
+    debounce: false,
+    value: { has_selection: false },
+    setValue: vi.fn(),
+  };
+}
+
+// jsdom has no matchMedia, no animation frames and no canvas context.
+function stubBrowser(ctx: Partial<CanvasRenderingContext2D> | null = null) {
+  vi.stubGlobal(
+    "matchMedia",
+    vi.fn(() => ({ addEventListener: vi.fn() })),
+  );
+  const requestAnimationFrame = vi.fn(() => 1);
+  vi.stubGlobal("requestAnimationFrame", requestAnimationFrame);
+  vi.stubGlobal("cancelAnimationFrame", vi.fn());
+  vi.spyOn(HTMLCanvasElement.prototype, "getContext").mockReturnValue(
+    ctx as CanvasRenderingContext2D | null,
+  );
+  return { requestAnimationFrame };
+}
+
+function createRecordingContext() {
+  return {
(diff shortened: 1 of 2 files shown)


abelsiqueira:
I have read the CLA Document, and I hereby sign the CLA

abelsiqueira:
recheck

frontend/src/plugins/impl/matplotlib/matplotlib-renderer.ts
```diff
@@ -343,6 +350,28 @@ export class MatplotlibRenderer {
...
   }
 
+  /**
+   * Scale to draw at, read back from the buffer rather than from
+   * `devicePixelRatio`, so sizing and drawing cannot disagree (#10625).
+   */
+  get #backingScale(): number {
+    const { width } = this.#state;
+    const dpr = globalThis.devicePixelRatio ?? 1;
+    return width > 0 ? this.#canvas.width / width : dpr;
```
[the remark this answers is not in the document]
abelsiqueira: Opus fixed this in the latest commit

frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
```diff
@@ -252,4 +273,113 @@ describe("MatplotlibRenderer", () => {
     expect(canvas.style.cursor).toBe("move");
     controller.abort();
   });
+
+  // Safari only fires resize on zoom, not the resolution query (#10625).
+  it("re-syncs the backing store on resize, without matchMedia firing", () => {
```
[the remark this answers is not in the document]
abelsiqueira: The original was posted in the issue, here it is again:
mwe-recording

Here is the fixed one:
mwe-recording-fixed

abelsiqueira:
I've asked Claude to fix the first Copilot comment, and added the gifs of the before and after on Safari. Let me know if you need further action from my side

frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
```diff
@@ -252,4 +273,145 @@ describe("MatplotlibRenderer", () => {
...
+
+    const state = { ...createState(), width: 150, height: 500 };
+    const container = document.createElement("div");
+    const controller = new AbortController();
+    new MatplotlibRenderer(container, { state, signal: controller.signal });
+
+    const canvas = container.querySelector("canvas");
+    expect(canvas?.width).toBe(187);
+    expect(canvas?.height).toBe(625);
+    expect(ctx.setTransform).toHaveBeenCalledWith(187 / 150, 0, 0, 1.25, 0, 0);
```
kirangadhave: `setTransform` is called twice. Once for pre-load clear and once for image draw.

`toHaveBeenCalledWith` passes when only one call matches. Please assert all the calls.

frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
```diff
@@ -252,4 +273,145 @@ describe("MatplotlibRenderer", () => {
...
+    // Load images synchronously; jsdom does not fetch them.
+    vi.stubGlobal(
+      "Image",
+      class {
+        onload: (() => void) | null = null;
+        set src(_value: string) {
+          this.onload?.();
+        }
+      },
+    );
```
kirangadhave: just extract the stub as helper

kirangadhave (changes requested):
Thanks @abelsiqueira

I have some minor comments in the test file. looks good to merge after.

frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
abelsiqueira: The assertion is done explicitly twice here now, and explicitly once in other places.

frontend/src/plugins/impl/matplotlib/__tests__/matplotlib-renderer.test.ts
abelsiqueira: Done

abelsiqueira:
Hi @kirangadhav, thanks for the review, I've updated and replied to your comments. Let me know what you think

kirangadhave (approved):

kirangadhave:
Thanks @abelsiqueira for the quick fix

abelsiqueira:
Happy to contribute!

PR #10628, opened 2026-08-21
perf(sql): batch schema reflection in catalog discovery

### LINKED ISSUE #10627: SQL catalog column discovery runs ~6 statements per table — use get_multi_columns
### Environment

marimo 0.24.0, SQLAlchemy 2.0.52, Snowflake via snowflake-sqlalchemy.

### Problem

With `auto_discover_columns = true`, catalog discovery against a 752-table Snowflake database does not finish (3+ minutes for a single schema, kernel blocked while it runs). We keep the flag off as a result, which means no column autocomplete in SQL cells.

### Root cause

`SQLAlchemyEngine.get_tables_in_schema` loops tables one at a time, and `get_table_details` makes three separate inspector calls per table (`get_columns`, `get_pk_constraint`, `get_indexes`). `_get_inspector` opens a new pooled connection, executes `USE DATABASE`, and calls `inspect(connection)` on every one of those calls, so each call gets a fresh `info_cache` and nothing is ever reused — about 6 statements per table, roughly 4,500 serial round trips for our database.

### Suggested fix

Build one Inspector per schema and use SQLAlchemy 2.x's batched reflection: `get_multi_columns`, `get_multi_pk_constraint`, `get_multi_indexes`. On Snowflake this resolves to one `information_schema.columns` query per schema plus one `SHOW PRIMARY KEYS IN SCHEMA` — roughly 2 queries per schema instead of 6 per table.

This is dialect-agnostic: the SQLAlchemy 2.0 built-in dialects (PostgreSQL, MySQL, MSSQL, Oracle) all implement the bulk `get_multi_*` path, and dialects without a bulk implementation fall back to SQLAlchemy's own per-table loop, so no dialect regresses.

### Evidence of the gap

A single `information_schema.columns` query on the same connection returns all 20,829 columns across 784 tables in a few seconds.

### Caveat

Snowflake can reject oversized information-schema results (errno 90030); chunking via `filter_names` covers it.

### Related

marimo-team/marimo#10611 covers two adjacent Snowflake discovery gaps (discovery silently off by default, and an all-account scan when the URL carries no database). This issue is specifically the per-table introspection cost once discovery is on.

sha2fiddy:

## 📝 Summary

Fixes #10627.

Catalog discovery fetches table details one table at a time. Each table costs three inspector calls, and on dialects that need a USE DATABASE command (Snowflake, StarRocks) each call opens a fresh pooled connection and builds a fresh inspector with an empty cache. That is roughly six statements per table, serially. On a Snowflake database with ~750 tables, column discovery never finished.

This PR batches reflection per schema. `get_tables_in_schema` now opens one `_get_inspector` context and calls `get_multi_columns`, `get_multi_pk_constraint`, and `get_multi_indexes` (SQLAlchemy 2.0+), with `filter_names` scoped to the tables being listed and `kind=ANY` / `scope=ANY` so views are included. On Snowflake that is 2 connections and about 4 statements per schema instead of 1+3N connections and 6N statements. Dialects without native multi support still run SQLAlchemy's default per-table loop, but on a single inspector whose `info_cache` persists across the loop.

Fallbacks keep behavior identical to today when anything goes wrong:

- An inspector without `get_multi_*` (SQLAlchemy 1.4) skips the batch silently.
- A batch that raises logs one warning and falls back to per-table reflection.
- A table missing from the batch result is retried individually, so one unreflectable table can't lose the whole schema.

`get_table_details` is unchanged and still serves the lazy single-table path; both paths now build their `DataTable`s through a shared `_build_data_table` helper, so they stay identical.

Validation: existing sqlite tests pass unchanged and now exercise the batched path. New tests cover views, all three fallbacks, the multi-call arguments, and inspector connection counts. Timed against a live Snowflake schema with 107 tables / ~6,400 columns: 6 seconds batched, vs about 1.1 s per table before (roughly 2 minutes for that one schema; discovery across the full database previously never completed).

Not addressed here (can file follow-ups):

- `_get_table_names` still uses its own inspector context, so a schema costs 2 connections instead of 1.
- The redshift, clickhouse, and ibis engines have the same per-table pattern.
- Inspector `info_cache` staleness is pre-existing and unchanged.

### DIFF
--- a/marimo/_sql/engines/sqlalchemy.py
+++ b/marimo/_sql/engines/sqlalchemy.py
@@ -94,6 +94,20 @@ def safe_execute(
     return decorator
 
 
+def _index_by_table_name(
+    reflected: dict[tuple[str | None, str], T],
+) -> dict[str, T]:
+    """Re-key SQLAlchemy multi-reflection results by bare table name.
+
+    Dialects disagree on the schema half of the `(schema, table)` keys
+    returned by the `get_multi_*` inspector methods (the value passed,
+    `None`, or a normalized form), so the schema half is dropped rather
+    than reconstructed. Each call site is scoped to a single schema,
+    which keeps the bare name unambiguous.
+    """
+    return {name: value for (_schema, name), value in reflected.items()}
+
+
 # ------------------------------------------------------------------ #
 #  SQLAlchemyEngine                                                   #
 # ------------------------------------------------------------------ #
@@ -543,6 +557,9 @@ class SQLAlchemyEngine(SQLConnection["Engine"]):
         for name in view_names:
             tables.append(("view", name))
 
+        if not tables:
+            return []
+
         if not include_table_details:
             return [
                 DataTable(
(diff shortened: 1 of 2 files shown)


Light2Dark:
Thanks @sha2fiddy , I made some changes here https://github.com/marimo-team/marimo/pull/10636 which are more concise and robust (removes indirection, shortens comments, and handles some edge cases). Thanks for raising!

PR #10629, opened 2026-08-21
fix(table): open the JSON viewer for str columns that hold JSON documents

### LINKED ISSUE #10568: Table: open the JSON cell viewer for str columns that parse as JSON
Clicking a cell opens the expandable JSON popup only when the column dtype is list or struct. A `str` column holding a JSON document gets truncated text and nothing else.

```python
import marimo as mo
import polars as pl

df = pl.DataFrame({
    "json_data": ['{"vendor":"ACME","attachments":[{"name":"inv.pdf","url":"https://example.com/p/AAA"}]}'],
})
df  
```

That's the usual shape for a JSON payload column — DuckDB's JSON type arrives as a string, and casting to a struct often isn't an option as it pins the schema, so not good for schema evolution. 

Could the viewer try a parse on str cells too? Cheap version: on click only, prefilter on the first non-space char being `{` or `[`, try the parse, fall back to plain text. Objects and arrays only, so ordinary text columns don't change.

marimo 0.23.16, Windows 11.

kirangadhave:
## 📝 Summary

Clicking a table cell opens the expandable JSON viewer only when the column dtype is list or struct. A str column that holds a JSON document, the usual shape for DuckDB JSON payloads, gets truncated text and nothing else.

Route str cells that hold a JSON document to the same `PopoutColumn` + `JsonOutput` tree that native list and struct cells use. A cell qualifies when its trimmed body starts and ends with a matching bracket pair and `JSON.parse` succeeds, so this stays objects and arrays only: JSON scalars, Python reprs, and malformed documents keep their current rendering. Copy still yields the original raw string. The row viewer panel shares the renderer and gets the same behavior.

Closes #10568

Clipboard-20260821-195651-252

### DIFF
--- a/frontend/src/components/data-table/columns.tsx
+++ b/frontend/src/components/data-table/columns.tsx
@@ -461,15 +461,17 @@ const PopoutColumn = ({
  * so ordinary text keeps its plain rendering.
  */
 function parseJsonDocument(text: string): unknown {
-  const first = text[0];
-  const last = text[text.length - 1];
+  const trimmedText = text.trim();
+
+  const first = trimmedText[0];
+  const last = trimmedText[trimmedText.length - 1];
   const isBracketPair =
     (first === "{" && last === "}") || (first === "[" && last === "]");
   if (!isBracketPair) {
     return undefined;
   }
   try {
-    return JSON.parse(text) as unknown;
+    return JSON.parse(trimmedText) as unknown;
   } catch {
     return undefined;
   }
@@ -671,7 +673,12 @@ export function renderCellValue<TData, TValue>({
           edges={{ leading, trailing }}
           wrapped={isWrapped}
         >
-          <JsonOutput data={jsonDocument} format="tree" className="max-h-64" />
+          <JsonOutput
+            data={jsonDocument}
+            format="tree"
+            valueTypes="json"
+            className="max-h-64"
+          />
         </PopoutColumn>
       );
     }
--- a/frontend/src/components/data-table/columns.tsx
+++ b/frontend/src/components/data-table/columns.tsx
@@ -455,6 +455,26 @@ const PopoutColumn = ({
   );
 };
 
+/**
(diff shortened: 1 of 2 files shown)


frontend/src/components/data-table/columns.tsx
```diff
@@ -455,6 +455,26 @@ const PopoutColumn = ({
...
+ * so ordinary text keeps its plain rendering.
+ */
+function parseJsonDocument(text: string): unknown {
+  const first = text[0];
+  const last = text[text.length - 1];
+  const isBracketPair =
+    (first === "{" && last === "}") || (first === "[" && last === "]");
+  if (!isBracketPair) {
+    return undefined;
+  }
```
dmadisetti: Why not just rely on `JSON.parse`? I suppose because it could pick up types too?

frontend/src/components/data-table/columns.tsx
```diff
@@ -639,6 +659,23 @@ export function renderCellValue<TData, TValue>({
       splitLeadingTrailingWhitespace(stringValue);
     const hasEdgeWhitespace = leading.length > 0 || trailing.length > 0;
 
+    // A str cell that holds a JSON document opens the same JSON viewer as
+    // native list/struct cells.
+    const jsonDocument = parseJsonDocument(middle);
+    if (jsonDocument !== undefined) {
+      return (
+        <PopoutColumn
```
dmadisetti, line 669: Love reusing components. Nice

dmadisetti (dismissed):

frontend/src/components/data-table/columns.tsx
kirangadhave: exactly, it will pickup scalars and try to parse long strings as well. the bracket checks let us skip early if it's not valid json

dmadisetti (approved):

yan-hic:
Cool. Thanks guys

PR #8557, opened 2026-03-03
feat: add option to disable command mode on Escape key

ryptotalent:
Closes #7426

PR #10638, opened 2026-08-24
feat(keymap): make the command mode shortcut configurable

### LINKED ISSUE #7426: Option to disable command mode or change key binding

iam-kira:
Fixes [#7426](https://github.com/marimo-team/marimo/issues/7426)

Escape was hardcoded as the shortcut for entering command mode in the default keymap, so it could be neither remapped nor disabled. The vim keymap already routed through a configurable command.vimEnterCommandMode hotkey.

This adds a command.enterCommandMode action (default Escape) and uses it in place of the hardcoded check. Overrides flow through the existing keymap.overrides config, so the shortcut can be remapped, and an explicit empty override disables it.

OverridingHotkeyProvider previously treated an empty override as "no override" and fell back to the default, which made disabling any hotkey impossible. Only undefined now falls back. The rest of the codebase already treats an empty key as disabled (parseShortcut, KeyboardHotkeys, duplicate detection).

The shortcuts dialog hint now reflects the configured key and is hidden when the shortcut is disabled.

Behavior change: the previous check was evt.key === "Escape", which also fired when modifiers were held. parseShortcut rejects unlisted modifiers, so Shift/Ctrl+Escape no longer enters command mode. This matches how every other hotkey behaves, but flagging it explicitly.

Testing: 3 new tests covering disable, fallback, and remap. 120 tests pass across navigation.test.ts, core/hotkeys/, and useDuplicateShortcuts.test.ts.

### DIFF
--- a/frontend/src/components/editor/controls/keyboard-shortcuts.tsx
+++ b/frontend/src/components/editor/controls/keyboard-shortcuts.tsx
@@ -271,22 +271,29 @@ export const KeyboardShortcuts: React.FC = () => {
     );
   };
 
-  const renderCommandGroup = (group: HotkeyGroup) =>
-    renderGroup(
+  const renderCommandGroup = (group: HotkeyGroup) => {
+    const isVim = config.keymap.preset === "vim";
+    const commandModeKey = hotkeys.getHotkey("command.enterCommandMode").key;
+    // Nothing to advertise if the user has disabled the shortcut.
+    if (!isVim && commandModeKey === "") {
+      return renderGroup(group);
+    }
+    return renderGroup(
       group,
       <p className="text-xs text-muted-foreground flex items-center gap-1">
         Press{" "}
-        {config.keymap.preset === "vim" ? (
+        {isVim ? (
           <>
             <KeyboardHotkeys shortcut={isPlatformMac() ? "Cmd" : "Ctrl"} />
             <Kbd>Esc</Kbd>
           </>
         ) : (
-          <Kbd>Esc</Kbd>
+          <KeyboardHotkeys shortcut={commandModeKey} />
         )}{" "}
         in a cell to enter command mode
       </p>,
     );
+  };
 
   return (
     <Dialog open={isOpen} onOpenChange={(open) => setIsOpen(open)}>
--- a/frontend/src/components/editor/controls/keyboard-shortcuts.tsx
+++ b/frontend/src/components/editor/controls/keyboard-shortcuts.tsx
@@ -5,7 +5,6 @@ import { AlertTriangleIcon, EditIcon, XIcon } from "lucide-react";
 import { useState } from "react";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
-import { Kbd } from "@/components/ui/kbd";
 import { hotkeysAtom, useResolvedMarimoConfig } from "@/core/config/config";
 import type { UserConfig } from "@/core/config/config-schema";
 import {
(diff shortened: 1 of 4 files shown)


iam-kira:

frontend/src/components/editor/controls/keyboard-shortcuts.tsx
```diff
@@ -271,22 +271,29 @@ export const KeyboardShortcuts: React.FC = () => {
...
-        {config.keymap.preset === "vim" ? (
+        {isVim ? (
           <>
             <KeyboardHotkeys shortcut={isPlatformMac() ? "Cmd" : "Ctrl"} />
             <Kbd>Esc</Kbd>
           </>
         ) : (
-          <Kbd>Esc</Kbd>
+          <KeyboardHotkeys shortcut={commandModeKey} />
         )}{" "}
```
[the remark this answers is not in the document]
kirangadhave: @iam-kira please address the review comment if PR is ready. I'll start my review after

iam-kira:
Addressed both presets now resolve their binding through the hotkey
provider, so the hint follows the platform default and any override, and is
omitted when the key is disabled. Added a test pinning the per-platform
resolution of the vim binding (Shift-Escape on Windows). Ready for review.

iam-kira:
Gentle bump before this goes stale, it's still ready whenever a maintainer has a moment.

Review comments from the first pass are addressed and pushed.
All checks green (3/3).
Rebased on current main, no conflicts.
This is the third attempt at #7426 (#8557 and #9907 were both closed by the stale bot without a review), so there's clearly demand for the option — happy to make any further changes if something's off.

kirangadhave:
Thanks for the bump @iam-kira I'll take a look at this today

frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -743,11 +748,10 @@ export function useCellEditorNavigationProps(
         if (vimCommandModeShortcut(evt)) {
           handleEscape();
         }
-      } else {
-        // For non-vim mode, regular Escape exits to command mode
-        if (evt.key === "Escape") {
-          handleEscape();
-        }
+      } else if (commandModeShortcut(evt)) {
```
kirangadhave, line 764: This condition also disables Escape’s editor-cleanup behavior.

`handleEscape()` closes completion and signature hints before it enters command mode. CodeMirror removes its own Escape completion binding. After I remap the command-mode shortcut to `q`, Escape no longer closes the completion popup.

This breaks the workflow described in #7426. Escape must continue to dismiss temporary editor UI when command-mode entry is remapped or disabled.

Can we separate these behaviors? Keep Escape responsible for editor cleanup, and apply the configured shortcut only to the command-mode transition. Please add navigation tests for remapped and empty overrides with an open completion popup.

https://github.com/user-attachments/assets/14a5e011-c241-49a1-bf15-b1065b19f427

I press Escape after the completion popup opens. The popup stays open because remapping command-mode entry also disables Escape’s cleanup behavior.

frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -743,11 +748,10 @@ export function useCellEditorNavigationProps(
...
+        // For non-vim mode, the configurable shortcut (Escape by default)
+        // exits to command mode. An empty override disables it.
+        handleEscape();
```
kirangadhave, line 768: Confirmed with `command.enterCommandMode` mapped to `q`.

When I select `hello` and press `q`, the cell stays in edit mode, the selection collapses, and the code becomes `qhello`.

When completion is open after `mo.`, pressing `q` closes the old popup, inserts `q`, and opens a new popup for `mo.q`.

`handleEscape()` returns after simplifying the selection or closing completion. Because this branch does not prevent the event, CodeMirror processes the same key as text input.

Please call `evt.preventDefault()` as soon as the configured shortcut matches. Add regression tests for both a selected range and an open completion popup.

See video:

https://github.com/user-attachments/assets/92dd0b25-80b0-4df3-ab83-ccf0e0bc3af0

First, `q` collapses the selection but also changes `hello` to `qhello`. 

Then `q` closes the completion popup, inserts itself, and opens a new popup for `mo.q`.
When typing accordion, the completion popup filters, but when typing `q` it closes and opens up again.

kirangadhave (changes requested):
Thanks for the PR. The current implementation causes two blocking regressions in editor cleanup and shortcut handling. Please see the inline comments and videos for reproduction steps.

I am remapping the command mode shortcut to `q` for these examples

iam-kira:
Gentle bump before this goes stale — it's still ready whenever a maintainer has a moment.

Review comments from the first pass are addressed and pushed.
All checks green (3/3).
Rebased on current main, no conflicts.
This is the third attempt at #7426 (#8557 and #9907 were both closed by the stale bot without a review), so there's clearly demand for the option — happy to make any further changes if something's off.

kirangadhave:
@iam-kira already requested changes with two major blocking issues

iam-kira:
Thanks for the detailed review and the repros @kirangadhave — both addressed in the latest push:

1. **Escape cleanup is now separate from command-mode entry.** Pulled the popup/selection dismissal out into `dismissTemporaryEditorUI()`. Escape always runs it (closing completion + signature hints, collapsing multi-cursor selection) even when `command.enterCommandMode` is remapped or disabled; the command-mode *transition* only fires on the configured shortcut.
2. **`evt.preventDefault()` now fires as soon as the configured shortcut matches**, so a remapped printable key like `q` no longer gets typed into the editor (no more `hello` → `qhello`, no popup reopen).

Added navigation tests covering remapped (`q`) and disabled (`""`) overrides with an open completion popup and with a selected range, asserting the key is prevented and command mode isn't entered on cleanup. `navigation.test.ts` passes locally (75/75), and oxlint/oxfmt are clean.

PR #10653, opened 2026-08-25
Accept a covariant Mapping of elements in ui.dictionary and .batch

### LINKED ISSUE #10631: Typing: ui.dictionary.__init__ uses invariant dict[str, UIElement] and rejects dict[str, ui.checkbox]
### Describe the bug

`marimo.ui.dictionary.__init__` currently types `elements` as:

```python
dict[str, UIElement[Any, Any]]
```

This causes static type checkers (Pylance/Pyright) to reject valid subclass dictionaries like `dict[str, mo.ui.checkbox]` because `dict` is invariant.

Observed runtime signature:
```python
(self, elements: 'dict[str, UIElement[Any, Any]]', *, label: 'str' = '', on_change: 'Callable[[dict[str, object]], None] | None' = None) -> 'None'
```

## Error message

Pylance/Pyright:
> Argument of type `dict[str, checkbox]` cannot be assigned to parameter `elements` of type `dict[str, UIElement[Any, Any]]` ... Consider switching from `dict` to `Mapping`.

## Why this is problematic
- Runtime behavior is correct.
- Users must add casts like `cast(dict[str, Any], elements)`.
- `UIElement` appears internal/non-public, so precise public typing is hard to express externally.

## Proposed fix
Use a covariant input type for constructor parameter, e.g.:

```python
Mapping[str, UIElement[Any, Any]]
```

### Will you submit a PR?

- [x] Yes

### Environment

- marimo: `0.24.0`
- Python: 3.14 (uv-managed)
- Type checker: Pylance/Pyright

### Code to reproduce

```python

from collections.abc import Iterable, Mapping
import marimo as mo

class Checkboxes(mo.ui.dictionary):
    def __init__(
        self,
        options: Iterable[str] | Mapping[str, str],
        preselected_options: Iterable[str] | None = None,
    ):
        pre = set(preselected_options) if preselected_options else set()
        elements: dict[str, mo.ui.checkbox]
        if isinstance(options, Mapping):
            elements = {
                option: mo.ui.checkbox(label=label, value=(option in pre))
                for option, label in options.items()
            }
        else:
            elements = {
                option: mo.ui.checkbox(label=option, value=(option in pre))
                for option in options
            }
        super().__init__(elements)  # type error
```

winklemad:

## 📝 Summary

Closes #10631

`dict` is invariant, so typing the `elements` parameter as `dict[str, UIElement[Any, Any]]` makes static type checkers (mypy, Pyright/Pylance) reject valid subclass dictionaries — e.g. `dict[str, mo.ui.checkbox]` — even though they are perfectly fine at runtime. Users have to work around it with `cast(dict[str, Any], elements)`.

Reproduction under mypy (marimo's CI type checker) on `main`:

```python
import marimo as mo

checks: dict[str, mo.ui.checkbox] = {"a": mo.ui.checkbox()}
mo.ui.dictionary(checks)
# error: Argument 1 to "dictionary" has incompatible type
#        "dict[str, checkbox]"; expected "dict[str, UIElement[Any, Any]]"  [arg-type]
# note: "Dict" is invariant
```

## Fix

The affected constructors only **read** `elements` — they immediately clone it into a fresh `dict` via `validate_and_clone` — so the parameter can be a read-only, covariant `Mapping`. Widened in:

- `ui.dictionary.__init__`
- the internal `batch` element `__init__`
- `validate_and_clone` (the shared helper, per the maintainer's note on the issue)

`validate_and_clone` still **returns** a `dict`, and the internal `_batch_base` (which stores and mutates the mapping) keeps its `dict` type since it always receives the cloned dict. `ui.array` already uses the covariant `Sequence`, so it needs no change. Runtime behaviour is unchanged.

## Verification

- The mypy reproduction above goes from erroring to clean after the change.
- `mypy marimo` reports no new errors in the changed files.
- New regression tests build a `ui.dictionary` and call `validate_and_clone` from a read-only `MappingProxyType` (a non-`dict` `Mapping`), which fails if the parameter is ever narrowed back to `dict`.
- `test_dictionary.py` + `test_batch.py` pass (11 tests); `ruff format`/`ruff check` clean.

## 📋 Pre-Review Checklist

- [x] For large changes, or changes that affect the public API: this change was discussed or approved through an issue — #10631 (labelled `bug`, contribution invited by a maintainer).

## ✅ Merge Checklist

### DIFF
--- a/marimo/_plugins/ui/_impl/batch.py
+++ b/marimo/_plugins/ui/_impl/batch.py
@@ -14,7 +14,13 @@ from marimo._plugins.core.web_component import JSONType
 from marimo._plugins.ui._core.ui_element import UIElement
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, ItemsView, Iterator, ValuesView
+    from collections.abc import (
+        Callable,
+        ItemsView,
+        Iterator,
+        Mapping,
+        ValuesView,
+    )
 
 U = TypeVar("U")
 V = TypeVar("V")
@@ -22,7 +28,7 @@ V = TypeVar("V")
 
 # Explicit type check, since the api has led to some confusion.
 def validate_and_clone(
-    elements: dict[str, UIElement[U, V]],
+    elements: Mapping[str, UIElement[U, V]],
 ) -> dict[str, UIElement[U, V]]:
     invalid: list[str] = []
     new_elements = {}
@@ -211,7 +217,7 @@ class batch(_batch_base):
     def __init__(
         self,
         html: Html,
-        elements: dict[str, UIElement[Any, Any]],
+        elements: Mapping[str, UIElement[Any, Any]],
         on_change: Callable[[dict[str, object]], None] | None = None,
     ) -> None:
         self._html = html
--- a/marimo/_plugins/ui/_impl/batch.py
+++ b/marimo/_plugins/ui/_impl/batch.py
@@ -204,12 +204,12 @@ class batch(_batch_base):
 
     Attributes:
         value (dict): A dict of the batched elements' values.
-        elements (dict): A dict of the batched elements (clones of the originals).
+        elements (Mapping): A dict of the batched elements (clones of the originals).
         on_change (Optional[Callable]): Optional callback to run when this element's value changes.
 
     Args:
         html (Html): A templated Html object.
-        elements (dict[str, UIElement]): The UI elements to interpolate into the HTML template.
+        elements (Mapping[str, UIElement]): The UI elements to interpolate into the HTML template.
         on_change (Optional[Callable[[Dict[str, object]], None]], optional): Optional callback
             to run when this element's value changes.
     """
--- a/marimo/_plugins/ui/_impl/dictionary.py
+++ b/marimo/_plugins/ui/_impl/dictionary.py
@@ -12,7 +12,7 @@ from marimo._plugins.ui._core.ui_element import UIElement
 from marimo._plugins.ui._impl.batch import _batch_base, validate_and_clone
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Mapping
 
 
 @mddoc
@@ -100,7 +100,7 @@ class dictionary(_batch_base):
 
     def __init__(
         self,
-        elements: dict[str, UIElement[Any, Any]],
+        elements: Mapping[str, UIElement[Any, Any]],
         *,
         label: str = "",
         on_change: Callable[[dict[str, object]], None] | None = None,
--- a/marimo/_plugins/ui/_impl/dictionary.py
+++ b/marimo/_plugins/ui/_impl/dictionary.py
@@ -86,12 +86,12 @@ class dictionary(_batch_base):
     Attributes:
         value (dict): A dict holding the values of the UI elements, keyed by
             their names.
-        elements (dict): A dict of the wrapped elements (clones of the originals).
+        elements (Mapping): A dict of the wrapped elements (clones of the originals).
         on_change (Optional[Callable[[dict[str, object]], None]]): Optional callback
             to run when this element's value changes.
 
     Args:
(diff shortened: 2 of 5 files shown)


winklemad:

tests/_plugins/ui/_impl/test_batch.py
```diff
@@ -26,6 +26,23 @@ def test_batch_rejects_non_ui_elements() -> None:
         md("Example {thing}").batch(thing={"key": "value"})  # type: ignore
 
 
+def test_validate_and_clone_accepts_readonly_mapping() -> None:
```
kirangadhave: this test will run without any changes to `batch.py` because Python doesn't enforce annotations on runtime. 

Typing only change requires a type-level test. See `tests/_ast/test_app_typing.py` for reference.

tests/_plugins/ui/_impl/test_dictionary.py
```diff
@@ -32,6 +32,23 @@ def func() -> None:
     assert dictionary.value == {"a": 2, "b": "goodbye", "c": 1}
 
 
+def test_dictionary_accepts_readonly_mapping() -> None:
```
kirangadhave, line 37: Same as: https://github.com/marimo-team/marimo/pull/10653/changes#r3874653890

marimo/_plugins/ui/_impl/batch.py
kirangadhave: docstring still refers to `dict`


```suggestion
        elements (Mapping): A dict of the batched elements (clones of the originals).
```

marimo/_plugins/ui/_impl/dictionary.py
kirangadhave: docstring still refers to `dict`


```suggestion
        elements (Mapping): A dict of the wrapped elements (clones of the originals).
```

tests/_plugins/ui/_impl/test_dictionary.py
```diff
@@ -32,6 +32,23 @@ def func() -> None:
...
+    # `elements` is typed as a covariant Mapping, so any read-only mapping of
+    # UI elements (not just a plain dict) must be accepted and cloned.
+    from types import MappingProxyType
```
kirangadhave: import at module level similar to other imports

tests/_plugins/ui/_impl/test_batch.py
```diff
@@ -26,6 +26,23 @@ def test_batch_rejects_non_ui_elements() -> None:
...
+    """`validate_and_clone` accepts any Mapping of UI elements, not just dict."""
+    from types import MappingProxyType
+
+    from marimo._plugins.ui._impl.batch import validate_and_clone
```
kirangadhave: import at module level similar to other imports

tests/_plugins/ui/_impl/test_batch.py
kirangadhave: nit: move the test to bottom. it separates two similar tests.

kirangadhave (changes requested):
Thanks @winklemad, I have a few change requests below. Also please  rebase the branch from `origin/main`

winklemad:
Thanks @kirangadhave — addressed all of these in ca31563:

- **Type-level test** (the load-bearing one): added `TestDictionaryTyping` in `tests/_ast/test_app_typing.py`, following the existing `_check_pyright` pattern. The two snippets error on the old invariant `dict[...]` param and pass on the covariant `Mapping[...]`, so the typing change is now genuinely guarded — basedpyright even suggests "switching from dict to Mapping which is covariant" on the old signature. (You're right that the runtime tests pass either way.)
- **Docstrings**: updated the `elements` entries in `batch.py` / `dictionary.py` that still said `dict`.
- **Imports**: hoisted `MappingProxyType` and `validate_and_clone` to module scope in both test files.
- **Nit**: moved the `validate_and_clone` test down next to the other batch tests.
- **Rebased** on `origin/main`.

Let me know if you'd like anything else.

kirangadhave (approved):

PR #10667, opened 2026-08-26
fix: accept Mapping instead of dict for ui.dictionary/ui.batch elements

### LINKED ISSUE #10631: Typing: ui.dictionary.__init__ uses invariant dict[str, UIElement] and rejects dict[str, ui.checkbox]

roridemonslayer:
## Summary

`dict` is invariant, so a `dict[str, mo.ui.checkbox]` (or any other `UIElement` subclass) was rejected by static type checkers wherever `dict[str, UIElement[Any, Any]]` was required, even though it works fine at runtime — see #10631.

- `ui.dictionary.__init__`'s `elements` param is now `Mapping[str, UIElement[Any, Any]]`
- `ui.batch.__init__`'s `elements` param is now `Mapping[str, UIElement[Any, Any]]`
- `validate_and_clone`'s input param is now `Mapping[str, UIElement[U, V]]` (return type stays `dict`, since it builds a fresh one)
- Docstrings updated to match

Left `_batch_base.__init__` (internal, always called post-`validate_and_clone` with a plain `dict`) as-is, since it's not part of the public covariance-breaking surface.

## Test plan
- [x] Added `basedpyright`-based regression tests reproducing the issue's exact repro for both `dictionary` and `batch` (and `validate_and_clone` directly) — confirmed they fail against the old code and pass against the fix
- [x] Added runtime tests confirming a non-dict `Mapping` (`MappingProxyType`) works correctly for both `dictionary` and `batch`
- [x] `pytest tests/_plugins/ui/_impl/test_dictionary.py tests/_plugins/ui/_impl/test_batch.py tests/_plugins/ui/_impl/test_dictionary_typing.py` — 14 passed
- [x] `basedpyright` on the two changed source files — 0 errors

Fixes #10631

### DIFF
--- a/marimo/_plugins/ui/_impl/batch.py
+++ b/marimo/_plugins/ui/_impl/batch.py
@@ -14,7 +14,13 @@ from marimo._plugins.core.web_component import JSONType
 from marimo._plugins.ui._core.ui_element import UIElement
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, ItemsView, Iterator, Mapping, ValuesView
+    from collections.abc import (
+        Callable,
+        ItemsView,
+        Iterator,
+        Mapping,
+        ValuesView,
+    )
 
 U = TypeVar("U")
 V = TypeVar("V")
--- a/marimo/_plugins/ui/_impl/batch.py
+++ b/marimo/_plugins/ui/_impl/batch.py
@@ -14,7 +14,7 @@ from marimo._plugins.core.web_component import JSONType
 from marimo._plugins.ui._core.ui_element import UIElement
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, ItemsView, Iterator, ValuesView
+    from collections.abc import Callable, ItemsView, Iterator, Mapping, ValuesView
 
 U = TypeVar("U")
 V = TypeVar("V")
@@ -22,7 +22,7 @@ V = TypeVar("V")
 
 # Explicit type check, since the api has led to some confusion.
 def validate_and_clone(
-    elements: dict[str, UIElement[U, V]],
+    elements: Mapping[str, UIElement[U, V]],
 ) -> dict[str, UIElement[U, V]]:
     invalid: list[str] = []
     new_elements = {}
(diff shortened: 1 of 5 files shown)


roridemonslayer:

roridemonslayer:
recheck

roridemonslayer:
recheck

kirangadhave:
@roridemonslayer thanks for the PR, but https://github.com/marimo-team/marimo/pull/10653 is already open and addresses the issue.

roridemonslayer:
Ah, missed that — thanks for the pointer, no need to duplicate effort. Closing this out.

PR #10679, opened 2026-08-27
fix: preserve unqualified references in methods

dmadisetti:
## Summary

Followup for the comments on #10678.

We previously only had a scope distinction for comprehensions (`is_comprehension`), the review bots in #10678, rightly caught that because class scoping works differently from function scoping, the recursive reference calls did not work in classes.

This PR:
 - generalizes the lexical scope types
 - adds in the condition to account for the recursive call defined in classes
 - distinguished "method" from function for variable type.


mscolnick (approved):

dmadisetti:
CI unrelated

PR #10696, opened 2026-08-28
Run notebook sandboxes from uv script environments

### LINKED ISSUE #6042: PackageManager's are ignoring index-url's when listing packages
### Describe the bug

I have inline deps like:
```
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
#
# [[tool.uv.index]]
# url = "https://custom-pypi-server.example.com/simple/"
# default = true
# ///
```

Now when marimo notices that e.g. `private-package` is missing from marimo cell, marimo lists the packages from PyPi instead of the custom-pypi-server. 

I would assume that the listing is done somewhere here: https://github.com/marimo-team/marimo/blob/cacdbb9652080a0fc67cf318b4087c4191d51468/marimo/_runtime/packages/pypi_package_manager.py#L380

### Will you submit a PR?

- [ ] Yes

### Environment

<details>

```
{
  "marimo": "0.14.17",
  "editable": false,
  "OS": "Linux",
  "OS Version": "6.12.40",
  "Processor": "",
  "Python Version": "3.11.12",
  "Binaries": {
    "Browser": "--",
    "Node": "--"
  },
  "Dependencies": {
    "click": "8.2.0",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.0",
    "markdown": "3.8",
    "narwhals": "1.34.1",
    "packaging": "24.2",
    "psutil": "5.9.5",
    "pygments": "2.19.1",
    "pyyaml": "6.0.2",
    "starlette": "0.46.1",
    "tomlkit": "0.12.5",
    "uvicorn": "0.34.2",
    "websockets": "15.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.5.3"
  },
  "Experimental Flags": {}
}

```

</details>

### Code to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
#
# [[tool.uv.index]]
# url = "https://custom-pypi-server.example.com/simple/"
# ///

import marimo

__generated_with = "0.14.17"


@app.cell()
def _():
    import pandas as pd

    return pd


@app.cell
def _():
    import marimo as mo

    return (mo,)


if __name__ == "__main__":
    app.run()
```
👆 This should not list pandas versions from PyPi but from https://custom-pypi-server.example.com/simple/ (i.e. fail)

### LINKED ISSUE #10547: `--sandbox` drops `[[tool.uv.index]]` semantics (`default`, `name`, `explicit`, sources) and silently discards `uv export --script` failures
### Describe the bug

`marimo edit --sandbox` resolves the PEP 723 header of a notebook in two stages. Stage 1 runs `uv export --script`. This stage obeys the full header. Stage 2 builds a new `uv run` command. This stage changes each `[[tool.uv.index]]` entry into a bare `--index <url>` flag (`construct_uv_flags`, `marimo/_cli/sandbox.py`, near line 308 on `main`):

```python
uv_flags.extend(["--index", config["url"]])
```

This conversion removes four semantics that uv defines:

| Header semantic | Behavior under uv | Behavior after the conversion |
| -- | -- | -- |
| `default = true` | uv examines this index last | uv examines this index first (declared order becomes flag order) |
| `name = "..."` | uv attaches the `UV_INDEX_<NAME>_USERNAME/PASSWORD` credentials | The credentials do not attach; authenticated indexes return 401 |
| `explicit = true` | The index supplies only the packages that sources pin to it | The index supplies all packages (no CLI flag can express `explicit`) |
| `[tool.uv.sources]` `{ index = ... }` | The source pins a package to one index | The pin is lost |

Stage 2 then resolves the stage-1 pins again, through an index list that does not have these semantics. The stage-1 pins are correct. But uv cannot satisfy a pin that is available only on a later index, because uv uses its first-index-match strategy by default. The result: scripts that `uv run --script`, `uv export --script`, and `uv sync --script` resolve correctly do not open in the sandbox. The resolver error points to an incorrect cause.

A second defect keeps the first defect hidden. Stage 1 can fail. One example is a 401 error from an authenticated index. When this failure occurs, `_resolve_requirements_txt_lines` (near line 234) discards the exception and the stderr of uv:

```python
except subprocess.CalledProcessError:
    pass  # Fall back if uv fails
```

marimo then continues with the header lines, which have no pins, and does not show a message. We captured the stage-2 temporary requirements file to confirm the runtime behavior. The file has 11 unpinned header lines when the index credentials are not available. The file has 119 exported pins when the credentials are available. The user does not see the 401 error. The user sees a resolution failure that seems unrelated.

The defect also has these effects:

* Multi-file sandbox mode (directory targets): `build_sandbox_venv` installs with `uv pip install -r <reqs>` and no index flags. This mode does not use `[[tool.uv.index]]` at all.
* Private registries in general: stage 2 does not have the index names. Thus the `UV_INDEX_<NAME>_*` environment credentials cannot authenticate stage 2. We found this defect in production with AWS CodeArtifact. Only netrc authentication operates on the bare `--index` URLs.
* The package management documentation shows `[[tool.uv.index]]` with `explicit = true` for sandbox notebooks. The current behavior does not agree with the documentation.

We recommend this fix:

1. Map the flags correctly where CLI equivalents exist. Change `default = true` entries to `--default-index <url>`. Change named entries to `--index <name>=<url>`. We confirmed that this map corrects the order and the credential binding for the usual case.
2. `explicit = true` and sources have no CLI equivalents. Thus a flag reconstruction cannot be fully correct. Two options are fully correct: install the stage-1 pins without a new resolution, or let uv make the environment through its script interfaces. `uv sync --script` obeys the complete header. We confirmed this with the repro below.
3. Show stage-1 `uv export` failures when the file contains a PEP 723 block. Write the stderr of uv as a warning. Keep the silent fallback for plain files: `uv export --script` exits with code 2 on a file that has no metadata block, and `marimo edit --sandbox new_file.py` needs the fallback.
4. Security note for item 1: `--index <name>=<url>` attaches the `UV_INDEX_<NAME>_*` credentials of the environment to a URL that the notebook header controls. A hostile header can use an index name that is equal to a credentialed name. The hostile index then receives those credentials. Two possible protections: pass a name only when it does not match a credential variable of the environment, or document the trust model.

Workarounds that we use today:

* Declare the private index before pypi in the header. The converted flag order then stays safe under first-index-match. But this changes the header semantics for plain uv.
* Use netrc or `uv auth login` for the credentials. URL-keyed authentication operates on bare `--index` URLs.
* `UV_INDEX_STRATEGY=unsafe-best-match` also unblocks the resolution. But this setting permits the dependency-confusion attack that the default strategy prevents. We do not use it.

Related: marimo-team/marimo#6042 (the package listing does not use index-url; different code path), discussion [#4605](<https://github.com/marimo-team/marimo/issues/4605>) (the documented procedure for alternative indexes is the procedure that fails here), and prior sandbox-dependency reports [#3817](<https://github.com/marimo-team/marimo/issues/3817>), [#2539](<https://github.com/marimo-team/marimo/issues/2539>), [#4809](<https://github.com/marimo-team/marimo/issues/4809>), [#8980](<https://github.com/marimo-team/marimo/issues/8980>).

### Will you submit a PR?

- [X] Yes

### Environment

```json
{
  "marimo": "0.23.16",
  "editable": false,
  "location": "~/.cache/uv/archive-v0/eslW0l5Azw-QPTb1/lib/python3.12/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.5.0",
  "Processor": "arm",
  "Python Version": "3.12.13",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "151.0.7922.110",
    "Node": "v26.7.0",
    "uv": "0.12.3 (Homebrew 2026-08-07 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.3",
    "narwhals": "2.24.0",
    "packaging": "26.3",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pyyaml": "6.0.3",
    "starlette": "1.6.0",
    "tomlkit": "0.15.1",
    "uvicorn": "0.52.3",
    "websockets": "17.0.1"
  },
  "Optional Dependencies": {
    "loro": "1.13.2"
  },
  "Experimental Flags": {}
}
```

We also got the same results on marimo 0.23.5, and in a linux/amd64 container. The related code is unchanged on `main` as of 2026-08-13.

### Code to reproduce

`pip-install-test==0.0.3` is available on [test.pypi.org](<http://test.pypi.org>) but not on [pypi.org](<http://pypi.org>). [pypi.org](<http://pypi.org>) has other versions of the same name. This is the same shape as a private-index configuration where an internal package has the same name as a public one. No authentication is necessary.

```python
# repro.py
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "pip-install-test==0.0.3",
# ]
#
# [tool.uv.sources]
# pip-install-test = { index = "testpypi" }
#
# [[tool.uv.index]]
# url = "https://pypi.org/simple/"
# default = true
#
# [[tool.uv.index]]
# name = "testpypi"
# url = "https://test.pypi.org/simple/"
# explicit = true
# ///
import marimo
app = marimo.App()
```

uv resolves this correctly:

```
$ uv export --script repro.py --no-hashes --no-annotate --no-header | grep pip-install-test
pip-install-test==0.0.3
$ uv sync --script repro.py   # succeeds
```

The marimo sandbox does not open it:

```
$ uvx marimo@0.23.16 edit --sandbox repro.py
Running in a sandbox: uv run --isolated --no-project --compile-bytecode
  --with-requirements /tmp/....txt --python ...
  --index https://pypi.org/simple/ --index https://test.pypi.org/simple/ marimo edit repro.py
  x No solution found when resolving `--with` dependencies:
hint: `pip-install-test` was found on https://pypi.org/simple/, but not at the
requested version (pip-install-test==0.0.3). ... By default, uv will only
consider versions that are published on the first index that contains a given
package, to avoid dependency confusion attacks.
```

manzt:
`--sandbox` previously provisioned an environment and launched marimo through separate uv paths. Package changes could also update the live environment independently from the notebook manifest. As a result, launch, install, export, and IPC behavior did not all share the same interpretation of a PEP 723 script environment.

This PR makes the synchronized script environment the unit used for each operation:

```py
environment = sync_notebook(path)
plan = launch(
    environment,
    args,
    overlay=runtime_overlay(),
)
```

uv resolves the notebook metadata once, including indexes, sources, credentials, and Python constraints. Launches use the resulting interpreter and environment directly. Package changes edit the manifest and synchronize it, while the runtime overlay supplies the exact running marimo version without writing it into the notebook.

Single-file and multi-app IPC kernels now follow the same model. The IPC path also carries virtual files, profiling, graceful shutdown, bounded startup, and stderr draining instead of losing those behaviors at the process boundary.

pyzmq is treated as the required dependency it already is rather than being installed and checked separately for each sandbox.

Closes #10547
Closes #6042


mscolnick:
@cubic-ai-dev

pyproject.toml
```diff
@@ -93,9 +93,9 @@ sql = [
     "sqlglot[c]>=26.8.0"       # SQL cells parsing;
 ]
 
-# For per-notebook sandboxed kernels
+# For per-notebook sandboxed kernels. pyzmq moved to the core
+# dependencies; the extra remains for the uv wheel.
 sandbox = [
-    "pyzmq>=27.1.0",           # IPC communication for sandbox kernels
```
dmadisetti, line 98: Ah, never realized this made it to required. I think we can clean up jupyter extension too then

tests/test_project_dependencies.py
```diff
@@ -15,6 +15,16 @@ def test_required_dependencies(pyproject_text: str) -> None:
...
 
 
+def test_pyzmq_is_a_required_dependency(pyproject_text: str) -> None:
+    """Kernel IPC imports zmq without a runtime check; sandbox code may
+    assume it is present wherever marimo is."""
+    import tomlkit
+
+    pyproject = tomlkit.loads(pyproject_text)
+    deps = pyproject["project"]["dependencies"]
+    assert any(str(dep).startswith("pyzmq") for dep in deps)
```
dmadisetti, line 25: We already do snapshots, I think this is overkill

Light2Dark (dismissed):

dmadisetti (approved):

mscolnick (approved):

PR #10701, opened 2026-08-29
feat(command-mode): add shortcut to create a SQL cell below

### LINKED ISSUE #10688: Add a command-mode shortcut to create a SQL cell below
### Description

Marimo command mode supports `Esc`, then `b` to create a new Python cell below the focused cell.
Creating a SQL cell requires an additional conversion step:
1. `Esc`
2. `b`
3. `Cmd/Ctrl + Shift + L`

It would be useful to have a command-mode shortcut such as `Esc`, then `q` that directly creates and focuses a new SQL cell below the current cell.

Hex supports this interaction pattern: in command mode, `q` creates a SQL cell. Its [keyboard-shortcut documentation](https://learn.hex.tech/docs/explore-data/notebook-view/keyboard-shortcuts) describes command mode and notes that its Add Cell menu exposes shortcuts for creating specific cell types.

## Suggested solution

Add a configurable action such as `command.createSQLCellAfter`, with `q` as its default command-mode binding.

Expected behavior:
- Press `Esc` to enter command mode.
- Press `q`.
- Create a new SQL cell immediately below the selected cell.
- Focus the new SQL editor.
- Do not execute the currently selected cell.
- Continue using the existing default SQL engine and output settings.
- Expose the action in the keyboard-shortcut settings so users can remap it.
The existing `b` shortcut should continue creating a regular Python cell.

## Relationship to [#6790](https://github.com/marimo-team/marimo/issues/6790)
This is related to #6790, which proposes that `Shift+Enter` on a SQL cell should create another SQL cell.

A dedicated command-mode shortcut may be a better approach because it is explicit and does not change the existing semantics of `Shift+Enter`:
- It separates cell creation from cell execution.
- It works regardless of the current cell’s type.
- Users can deliberately choose between a Python cell with `b` and a SQL cell with `q`.
- It avoids implicitly deriving the new cell type from the preceding cell.
- It does not affect users who expect newly created cells to default to Python.

The two proposals could coexist, but the command-mode shortcut provides the workflow without changing existing behavior.

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

Aniket-a14:
Closes #10688.

Adds `command.createSqlCellAfter`, bound to `q` in command mode by default. Pressing `Esc` then `q` creates and focuses a new SQL cell below the current cell without executing it, adding the `marimo` import first if it's missing. The existing `b` shortcut (plain cell) is unchanged.

## Test plan

- [x] `pnpm typecheck` / `pnpm lint` pass
- [x] Existing hotkey tests pass (no key collisions)
- [x] Manually verified in a running notebook: `Esc` → `q` inserts a focused SQL cell below, in the correct order relative to the auto-inserted `import marimo as mo`, without running any cell; `Esc` → `b` still creates a plain cell (regression check)

### DIFF
--- a/frontend/src/components/editor/navigation/navigation.ts
+++ b/frontend/src/components/editor/navigation/navigation.ts
@@ -12,6 +12,7 @@ import { useMemo } from "react";
 import { mergeProps, useFocusWithin, useKeyboard } from "react-aria";
 import { DATA_FOR_CELL_ID } from "@/components/data-table/cell-utils";
 import { aiCompletionCellAtom } from "@/core/ai/state";
+import { maybeAddMarimoImport } from "@/core/cells/add-missing-import";
 import { cellIdsAtom, notebookAtom, useCellActions } from "@/core/cells/cells";
 import { useCellFocusActions } from "@/core/cells/focus";
 import type { CellId } from "@/core/cells/ids";
@@ -26,6 +27,7 @@ import {
   closeSignatureHint,
   signatureHintField,
 } from "@/core/codemirror/completion/signature-hint";
+import { LanguageAdapters } from "@/core/codemirror/language/LanguageAdapters";
 import {
   hotkeysAtom,
   isAiFeatureEnabled,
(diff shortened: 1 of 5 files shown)


Aniket-a14:

Aniket-a14:
recheck

frontend/src/components/editor/navigation/navigation.ts
```diff
@@ -551,6 +553,23 @@ export function useCellNavigationProps(
...
         },
+        "command.createSqlCellAfter": (cellId) => {
+          if (Events.hasModifier(evt)) {
+            return false;
+          }
+          const importCellId = maybeAddMarimoImport({
+            autoInstantiate: true,
+            createNewCell: actions.createNewCell,
+            fromCellId: cellId,
+          });
```
Light2Dark, line 564: I think you likely want to add a marimo Import before the cell. Right now it's 
selected cell -> import cell -> new sql cell.

There is a `before` keywords for maybeAddMarimoImport

And you should add tests for this feature too

Light2Dark (changes requested):

Aniket-a14:
@Light2Dark Addressed the import ordering and added tests in 88f468b07.

Light2Dark (approved):
thanks!

Issue #10377, opened 2026-07-29
Claude Code marketplace provided skill not found


leevilehtonen:
### Describe the bug

Followed instructions here: https://marimo.io/pair, i.e. 
```sh
claude plugin marketplace add marimo-team/marimo-pair
claude plugin install marimo-pair@marimo-pair
```
The plugin (and the skill) got installed to `/Users/leevi/.claude/plugins/marketplaces/marimo-pair/skills/marimo-pair` which does not match the path resolution here:

1. https://github.com/marimo-team/marimo/blob/eed8a821d040dd11613786e1792e7459e31ec02e/marimo/_cli/pair/commands.py#L48
2. https://github.com/marimo-team/marimo/blob/eed8a821d040dd11613786e1792e7459e31ec02e/marimo/_cli/pair/commands.py#L37

that seems to be expecting it at `~/.claude/plugins/marketplaces/marimo-pair/SKILL.MD`.

(Is the claude marketplace installation unsupported since many places advertise `npx skills` based installation (including the error message)?)

### Will you submit a PR?

- [ ] Yes

### Environment

<details>
<summary>Environment</summary>

```json
{
  "marimo": "0.23.15",
  "editable": false,
  "location": "~/.cache/uv/archive-v0/nOa2BPRVaaO7pkhJ/lib/python3.14/site-packages/marimo",
  "OS": "Darwin",
  "OS Version": "25.5.0",
  "Processor": "arm",
  "Python Version": "3.14.6",
  "Locale": "C",
  "Binaries": {
    "Browser": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36",
    "Node": "v24.18.0",
    "uv": "0.11.32 (Homebrew 2026-07-23 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.4.2",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.2",
    "narwhals": "2.24.0",
    "packaging": "26.2",
    "psutil": "7.2.2",
    "pygments": "2.20.0",
    "pymdown-extensions": "10.21.3",
    "pyyaml": "6.0.3",
    "starlette": "1.3.1",
    "tomlkit": "0.15.1",
    "typing-extensions": "4.16.0",
    "uvicorn": "0.51.0",
[...]

PR #10704, opened 2026-08-30
fix(cli): find marimo-pair skill from subdirectories of the repo root

jungleBadger:
## Problem

`marimo pair prompt --claude` reports the marimo-pair skill as missing whenever it is
run from a subdirectory of the project, even when the skill is correctly installed at
`<repo>/.claude/skills/marimo-pair/SKILL.md`:

```
$ cd my-project && marimo pair prompt --url http://localhost:2718 --claude
# (no warning — skill found)

$ cd my-project/notebooks && marimo pair prompt --url http://localhost:2718 --claude
The marimo-pair skill for Claude Code could not be found.

Please install it with:

  npx skills add marimo-team/marimo-pair
```

Same project, same install. Running the command from anywhere below the root — which is
common, since notebooks often live in a subdirectory — tells the user to reinstall a
skill they already have.

## Cause

`_claude_skill_dirs()` builds its roots from `Path.cwd() / ".claude"` only. Claude Code
resolves project skills from `.claude/skills` in the directory it was started in **and in
every parent directory up to the repository root**, so a skill installed once at the root
applies to every subdirectory. The validator checked a single directory where Claude Code
checks a chain.

## Fix

Add `_claude_project_roots()`, which walks from the current directory up to the repository
root and returns each `.claude` along the way. It stops at the root rather than continuing
to unrelated ancestors, and treats a `.git` *file* as a root so worktrees and submodules
resolve correctly. Outside a repository there is no root to walk to, so only the current
directory applies — matching the previous behaviour for that case.

`~/.claude` is still checked, and the plugin/marketplace subdirectory handling is unchanged.

## Tests

`_claude_skill_dirs()` had no coverage, which is how this went unnoticed. Added
`TestClaudeProjectRoots` with five cases: the parent walk itself, the end-to-end
"skill at root, found from a subdirectory" regression, the stop-at-root boundary,
`.git`-as-a-file, and the no-repository fallback.

Verified the regression test fails without the fix and passes with it, using only the
pre-existing `_claude_skill_dirs()` / `AgentConfig` API:

```
UNFIXED source: 1 failed
FIXED source:   1 passed
```

`tests/_cli/test_cli_pair.py` — 31 passed. `ruff check`, `ruff format --check`, and
`mypy marimo/_cli/pair/commands.py` all clean. The wider `tests/_cli/` suite has 7
failures in `test_cli_export.py` (`test_export_watch_*_no_out_dir`) that reproduce
identically on an unmodified checkout of `main`, so they are unrelated to this change.

## Related

Same class of path-resolution mismatch as previous fixes to this function, and not covered
by any of them:

- #10377 / #10385 — marketplace-installed skills reported missing (same symptom, different root path)
- #9225 — opencode skill discovery
- #9021 — marketplace and plugin install paths

This is the project-root dimension: all three above corrected *which* directories are
checked, while `.claude` project roots were still resolved from `Path.cwd()` alone.

## Note

Drafted with AI assistance; disclosed via a `Co-Authored-By` trailer on the commit. Happy
to drop the trailer or restructure if that conflicts with project preference.

### DIFF
--- a/marimo/_cli/pair/commands.py
+++ b/marimo/_cli/pair/commands.py
@@ -38,13 +38,32 @@ class AgentConfig:
         )
 
 
+def _claude_project_roots() -> list[Path]:
+    """Return the `.claude` directories that apply to the current project.
+
+    Claude Code loads project skills from `.claude/skills` in the directory it
+    was started in *and in every parent up to the repository root*, so a skill
+    installed once at the root is available from any subdirectory. Mirror that
+    walk here; checking only the current directory reports the skill missing
+    whenever the notebook lives below the root. Outside a repository there is
+    no root to walk to, so only the current directory applies.
+    """
+    cwd = Path.cwd()
+    candidates = [cwd, *cwd.parents]
+    for index, directory in enumerate(candidates):
+        # A worktree or submodule has `.git` as a file, not a directory.
+        if (directory / ".git").exists():
+            return [d / ".claude" for d in candidates[: index + 1]]
+    return [cwd / ".claude"]
+
+
 def _claude_skill_dirs() -> list[Path]:
     """Return all directories where a Claude Code skill may be installed.
 
     Skills can be installed directly or bundled in a marketplace plugin in
-    both the global (`~/.claude`) and local (`.claude`) config directories.
+    both the global (`~/.claude`) and project (`.claude`) config directories.
     """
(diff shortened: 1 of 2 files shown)


jungleBadger:

mscolnick:
~this could be quite expensive, depending on the number of subdirectories~ oh sorry, you are checking parents, not children, that should be fine

PR #10712, opened 2026-09-01
fix: accept a covariant Mapping in tabs, accordion, and routes

kirangadhave:
## 📝 Summary

`mo.ui.tabs`, `mo.accordion`, the deprecated `mo.tabs` shim, and `mo.routes` type their content parameter as `dict[str, object]`. `dict` is invariant, so mypy and Pyright reject a more specific dict such as `dict[str, mo.Html]` and users must cast, even though runtime behavior is correct. All four call sites only read the mapping, so this widens the parameters to `Mapping[str, object]`, updates the docstrings, and adds basedpyright regression tests in the style of #10653, which fixed the same defect for `ui.dictionary` and `batch`.

Closes [MO-7499](https://linear.app/marimo/issue/MO-7499/widen-invariant-dict-params-on-mouitabs-and-moaccordion-to-mapping)

### DIFF
--- a/marimo/_plugins/stateless/accordion.py
+++ b/marimo/_plugins/stateless/accordion.py
@@ -1,6 +1,8 @@
 # Copyright 2026 Marimo. All rights reserved.
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from marimo._output.formatting import as_html
 from marimo._output.hypertext import ContainerHtml, Html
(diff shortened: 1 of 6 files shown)


Light2Dark (approved):
nice! I wonder if it's better to have the typecheck CI and pre-commit cover this, instead of the tests. Maybe a future change

PR #10769, opened 2026-09-09
fix: add an exit control to fullscreen cell output

### LINKED ISSUE #108: No UI option to return when in full screen on a cell output
### Describe the bug

# A clear and concise description of the bug.

After entering the full screen view of a cell output, for a chart / image / etc. There is no clear ui path to return to the notebook view.

# Workaround

Right click and select `Clear outputs`

### Steps to reproduce

* Create a marimo notebook and run a cell:

```python
plt.figure()
```

* Click the `Full Screen` button

### Environment

* Plugin version: 0.1.2
* IDE and version (e.g. PyCharm 2025.1.2 Community): PyCharm 2026.2.1
* OS: MacOS 26.6.2
* `marimo --version`: 0.24.0
* `uv --version`: 0.12.9

### Will you submit a PR?

- [ ] Yes

kirangadhave:
## 📝 Summary

Fullscreen paints the output element only, and the fullscreen and expand buttons sit outside it in the cell gutter. A fullscreen output therefore shows no control at all. Browsers hide this with their own exit overlay and Escape handling, but embedded hosts do not. In JCEF, which backs the JetBrains plugin, no overlay appears and Escape does nothing, so the only way back to the notebook is to clear the cell output.

Render an exit button inside the fullscreen element, and exit on Escape while an output is fullscreen. A local build in PyCharm confirms that the Escape handler works, so JCEF delivers the key and simply never acts on it. This removes the need for a keyboard handler on the plugin side.

Closes https://github.com/marimo-team/jetbrains-marimo/issues/108
Closes MO-7561

### DIFF
--- a/frontend/src/components/editor/Output.tsx
+++ b/frontend/src/components/editor/Output.tsx
@@ -1,5 +1,5 @@
 /* Copyright 2026 Marimo. All rights reserved. */
-import React, { memo, Suspense, useMemo, useRef } from "react";
+import React, { memo, Suspense, useEffect, useMemo, useRef } from "react";
 import { type CellId, CellOutputId } from "@/core/cells/ids";
 import type { CellOutput, OutputMessage } from "@/core/kernel/messages";
 import { cn } from "@/utils/cn";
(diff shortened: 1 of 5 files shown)


frontend/src/components/editor/Output.tsx
```diff
@@ -425,6 +434,24 @@ const ExpandableOutput = React.memo(
     const [isExpanded, setIsExpanded] = useExpandedOutput(cellId);
     const isOverflowing = useOverflowDetection(containerRef);
     const { hasFullscreen } = useIframeCapabilities();
+    const isFullscreen = useIsFullScreen(containerRef);
+
+    // Not every host exits fullscreen on Escape by itself.
+    useEffect(() => {
```
mscolnick: we should have a `useEventListener(document, () …)` that may DRY this up only slightly

mscolnick (dismissed):

mscolnick (approved):

PR #10775, opened 2026-09-09
fix: serialize Polars cache values with native IPC

peter-gy:
Upstreaming some findings from https://github.com/marimo-team/marimo-export

Fixes two failures when caching Polars DataFrames and Series:

- Hashing string or mixed-type columns could raise `TypeError` when converting NumPy object arrays to bytes
- Without PyArrow, the writer stored pickle data in `.arrow` files that the reader could not restore

Uses Polars' native serialization for hashing and its own Arrow readers and writers for cache round-trips. Hashes preserve column names and types and remain consistent across chunk layouts.

### DIFF
--- a/marimo/_runtime/primitives.py
+++ b/marimo/_runtime/primitives.py
@@ -76,6 +76,10 @@ def is_data_primitive(value: Any) -> bool:
     if is_instance_by_name(value, "torch.Tensor"):
         return str(value.device) == "cpu"
 
+    # Series follow the same numeric-schema policy as dataframes.
+    if is_instance_by_name(value, "polars.series.series.Series"):
+        return bool(value.dtype.is_numeric())
+
     # If a numpy like array, ensure that it's not an object array.
     if hasattr(value, "dtype"):
         return not (
--- a/marimo/_save/cache.py
+++ b/marimo/_save/cache.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
     from marimo._save.stores import Store
 
 # NB. Increment on cache breaking changes.
-MARIMO_CACHE_VERSION: int = 5
+MARIMO_CACHE_VERSION: int = 6
 
 CacheType = Literal[
     "ContextExecutionPath",
--- a/marimo/_save/encode.py
+++ b/marimo/_save/encode.py
@@ -79,7 +79,7 @@ def _contiguous_tensor_bytes(data: Tensor) -> memoryview:
         "polars.dataframe.frame.DataFrame",
         "polars.series.series.Series",
     ):
-        from polars.exceptions import ComputeError
+        from polars.exceptions import PanicException, PolarsError
 
         # Preserve column types and names, including mixed and string columns
         # whose NumPy representation contains Python object references.
@@ -87,7 +87,7 @@ def _contiguous_tensor_bytes(data: Tensor) -> memoryview:
         buffer = io.BytesIO()
         try:
             frame.rechunk().write_ipc(buffer)
-        except ComputeError as exc:
+        except (PolarsError, PanicException) as exc:
             raise TypeError(
                 "Polars value cannot be serialized to IPC."
             ) from exc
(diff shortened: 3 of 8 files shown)


marimo/_save/loaders/lazy.py
```diff
@@ -218,6 +219,13 @@ def maybe_update_lazy_stub(value: Any) -> str:
     # MRO not that expensive, type hashable for functools lookup.
     result = mro_lookup(value_type, LAZY_STUB_LOOKUP)
     loader = result[1] if result else "pickle"
+    if (
+        loader == "arrow"
+        and result is not None
+        and result[0].startswith("pandas.")
+        and not DependencyManager.pyarrow.has()
```
dmadisetti, line 226: Ah nice

marimo/_save/encode.py
```diff
@@ -75,6 +75,23 @@ def standardize_tensor(tensor: Tensor) -> Tensor:
...
+        # whose NumPy representation contains Python object references.
+        frame = data.to_frame() if hasattr(data, "to_frame") else data
+        buffer = io.BytesIO()
+        try:
+            frame.rechunk().write_ipc(buffer)
+        except (PolarsError, PanicException) as exc:
+            raise TypeError(
+                "Polars value cannot be serialized to IPC."
+            ) from exc
+        return buffer.getbuffer()
```
dmadisetti: I don't think this is a no-copy. I'm ok with doing this if it breaks polars but shouldn't be in this function. I think we have a check for mixed data types that should prevent the logic even getting here though.

marimo/_save/stubs/lazy_stub.py
```diff
@@ -269,22 +269,17 @@ def _arrow_dump(obj: Any) -> bytes:
...
-    frame = obj.to_frame()
-    if hasattr(frame, "write_ipc"):  # polars Series → polars DataFrame
+    # DataFrame columns can shadow Series method names such as to_frame.
+    if callable(getattr(type(obj), "write_ipc", None)) or callable(
+        getattr(type(obj), "to_feather", None)
+    ):
+        frame = obj
+    else:
+        frame = obj.to_frame()
+    if callable(getattr(type(frame), "write_ipc", None)):
```
dmadisetti: Cache invalidation should also bump the cache version.
I think we can do a nested check to prevent additional misses (+ will clean up logic)

marimo/_save/encode.py
peter-gy: yea, you are right! restored the buffer path

marimo/_save/stubs/lazy_stub.py
peter-gy: good catch

marimo/_runtime/primitives.py
```diff
@@ -76,6 +76,10 @@ def is_data_primitive(value: Any) -> bool:
     if is_instance_by_name(value, "torch.Tensor"):
         return str(value.device) == "cpu"
 
+    # Series follow the same numeric-schema policy as dataframes.
+    if is_instance_by_name(value, "polars.series.series.Series"):
+        return bool(value.dtype.is_numeric())
+
     # If a numpy like array, ensure that it's not an object array.
     if hasattr(value, "dtype"):
         return not (
```
dmadisetti, line 85: Is it only Series? I wonder if you can just do a polars check here

marimo/_save/cache.py
```diff
@@ -46,7 +46,7 @@
     from marimo._save.stores import Store
 
 # NB. Increment on cache breaking changes.
-MARIMO_CACHE_VERSION: int = 5
+MARIMO_CACHE_VERSION: int = 6
```
dmadisetti, line 49: I don't think this is invalidating anymore!

dmadisetti: * in a meaningful way. So we don't need to bump

dmadisetti (approved):
Nice! Take a pass at my comments if you feel like they're warranted, but this is much better!

peter-gy:
@dmadisetti thanks for the review! how about iterating on this in a follow up? it would be nice to have this in the next release. I leave the decision of when to merge this with you though (:

mscolnick:
merged, @peter-gy followup is perfect

PR #10778, opened 2026-09-09
feat(pair): list active notebook sessions

kirangadhave:
## 📝 Summary

Stacked on #10777.

Add `marimo pair notebooks list` for finding active notebooks and the session IDs required by `pair execute`. With no explicit URL, the command reads live no-token servers from the existing local registry. Explicit URLs can use a token file or `MARIMO_TOKEN`.

Query each selected server through the existing sessions endpoint and return deterministic JSON grouped by server and notebook. Preserve successful results when another server fails, report per-server errors, ignore stopped registry processes without deleting their records, and never send ambient credentials to automatically discovered servers.

This is a current-backend adapter. It does not implement the proposed discovery protocol or change the registry schema.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

### DIFF
--- a/marimo/_cli/pair/commands.py
+++ b/marimo/_cli/pair/commands.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 import hashlib
 import os
-import shlex
 from dataclasses import dataclass, field
 from pathlib import Path
 
@@ -300,6 +299,13 @@ def docs(topic: str | None) -> None:
     help="URL of the running marimo kernel.",
 )
 @click.option(
+    "--session",
+    "session_id",
+    required=True,
+    type=str,
+    help="Current session ID.",
+)
+@click.option(
     "--file",
     "file_path",
     default=None,
@@ -332,6 +338,7 @@ def docs(topic: str | None) -> None:
 )
 def prompt(
     url: str,
+    session_id: str,
     file_path: str | None,
     claude: bool,
     codex: bool,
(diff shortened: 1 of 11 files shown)


tests/_cli/test_cli_pair.py
```diff
@@ -29,9 +30,28 @@
...
+
+Commands:
+  docs       Read notebook guidance on demand. Run: uv run marimo pair docs
+             --help
+  execute    Run Python in a live notebook session. Run: uv run marimo pair
+             execute --help
+  notebooks  Find active notebooks and their sessions. Run: uv run marimo pair
+             notebooks --help
+  prompt     Generate pairing instructions. Run: uv run marimo pair prompt
+             --help
```
manzt, line 53: Only thought here is that we shouldn't encode the "uv run" in all these places.

marimo/_cli/pair/commands.py
```diff
@@ -428,6 +437,105 @@ def prompt(
...
+@click.command(
+    name="list",
+    cls=ColoredCommand,
+    help="List active notebooks and their session IDs.",
+    short_help="""List active notebooks and their session IDs.
+    Run: uv run marimo pair notebooks list --help""",
+    epilog="""\b
+This backend lists active notebooks only. Without --url, it discovers
+no-token servers from the local registry. For an authenticated server,
+supply its URL and credential source explicitly.
```
manzt, line 488: Nice. We can upgrade this to the local discovery of more marimo servers in the future but this is a better impl for now.

manzt (commented):
Looking great. It would be nice if we could avoid baking in `uv run` requirements. If the agent if the agent is able to invoke this part of the marimo CLI, then it has found a version of marimo that supports these commands (and we shouldn't direct it to use some other approach).

PR #10785, opened 2026-09-10
fix(markdown): render display math nested in list item continuations

SergeyAxiomatic:
## Summary
- `$$...$$` display math written as a continuation line of a list item (e.g. `2. text\n   $$\n   ...\n   $$`, no blank line, indented only to the width of the marker) fell out of the list and rendered as literal `$$` text, while inline `$...$` math in the same item rendered fine.
- Root cause: `DisplayMathPreprocessor` inserts blank lines around `$$` blocks so `pymdownx.arithmatex` treats them as their own markdown block, but that turns a tight list item into a loose one. Python-Markdown's `ListIndentProcessor` only reattaches loose blank-line-separated content to its list item when it's indented by at least `tab_length` (4 spaces by default) — not merely by the marker's own width (e.g. 3 columns for `"2. "`). So the widened block popped out of the `<li>`.
- Fix: before inserting the blank lines, widen the indentation of a `$$` continuation block (and its surrounding continuation lines) up to `tab_length` when it directly follows a list marker line and its current indent is narrower than that.

## Test plan
- [x] Added `test_md_display_math_inside_list_item` in `tests/_output/test_md.py`, reproducing the exact reported scenario (numbered list, no blank lines, inline + display math in the same item) and asserting the list stays a single correctly-numbered `<ol>` with the display math rendered as a block.
- [x] `python -m pytest tests/_output/test_md.py -q` — 52 passed.
- [x] `ruff check` on changed files — clean.
- [x] Manually verified bulleted lists and multiple list items with display math (items 2 and 4, matching the original bug report) render correctly.

### DIFF
--- a/marimo/_output/md_extensions/display_math.py
+++ b/marimo/_output/md_extensions/display_math.py
@@ -26,7 +26,11 @@ class DisplayMathPreprocessor(preprocessors.Preprocessor):  # type: ignore[misc]
     DOLLAR_DOLLAR_PATTERN = re.compile(r"^\s*\$\$\s*$")
 
     # Matches an ordered/unordered list item marker: "2. ", "- ", "* ", "+ "
-    LIST_MARKER_PATTERN = re.compile(r"^\s*([0-9]+[.)]|[-*+])\s+")
+    # Only "N." is recognized by python-markdown's OList processor (unlike
+    # "N)"), so that's all we match here -- treating "1)" as a marker would
+    # misidentify plain prose and push its (non-list) continuation lines to
+    # 4-space indentation, turning them into an indented code block.
+    LIST_MARKER_PATTERN = re.compile(r"^\s*([0-9]+\.|[-*+])\s+")
 
     # Matches inline RST math role: :math:`...`
     INLINE_MATH_ROLE_PATTERN = re.compile(r"(?<!`):math:`([^`\n]+)`")
@@ -167,11 +171,19 @@ class DisplayMathPreprocessor(preprocessors.Preprocessor):  # type: ignore[misc]
                 continue
 
             start = i
-            while (
-                i < n
-                and result[i].strip()
-                and self._count_indent(result[i]) > 0
-            ):
+            in_dollar_block = False
+            while i < n:
+                current = result[i]
+                if not current.strip():
+                    # A blank line inside an open $$...$$ block (e.g. a
+                    # multi-paragraph LaTeX environment) doesn't end the
+                    # continuation -- only a blank line outside one does.
+                    if not in_dollar_block:
+                        break
+                elif self._count_indent(current) == 0:
+                    break
+                elif self.DOLLAR_DOLLAR_PATTERN.match(current):
+                    in_dollar_block = not in_dollar_block
                 i += 1
             end = i
 
@@ -183,10 +195,16 @@ class DisplayMathPreprocessor(preprocessors.Preprocessor):  # type: ignore[misc]
                 for line_ in run
             )
             if contains_math and self.LIST_MARKER_PATTERN.match(preceding):
-                min_indent = min(self._count_indent(line_) for line_ in run)
+                non_blank = [line_ for line_ in run if line_.strip()]
+                min_indent = min(
+                    self._count_indent(line_) for line_ in non_blank
+                )
                 if 0 < min_indent < tab_length:
                     padding = " " * (tab_length - min_indent)
-                    result[start:end] = [padding + line_ for line_ in run]
+                    result[start:end] = [
+                        padding + line_ if line_.strip() else line_
+                        for line_ in run
+                    ]
 
         return result
 
--- a/marimo/_output/md_extensions/display_math.py
+++ b/marimo/_output/md_extensions/display_math.py
@@ -195,18 +195,71 @@ class DisplayMathPreprocessor(preprocessors.Preprocessor):  # type: ignore[misc]
                 for line_ in run
             )
             if contains_math and self.LIST_MARKER_PATTERN.match(preceding):
-                non_blank = [line_ for line_ in run if line_.strip()]
-                min_indent = min(
-                    self._count_indent(line_) for line_ in non_blank
-                )
-                if 0 < min_indent < tab_length:
-                    padding = " " * (tab_length - min_indent)
-                    result[start:end] = [
-                        padding + line_ if line_.strip() else line_
-                        for line_ in run
-                    ]
+                self._pad_math_segments(result, start, end, tab_length)
 
         return result
+
+    def _pad_math_segments(
+        self, result: list[str], start: int, end: int, tab_length: int
+    ) -> None:
+        """Independently pad each blank-line-isolated segment to tab_length.
+
+        `_normalize_display_math_spacing` inserts a blank line before/after
+        every $$ construct, which splits a tight list item's continuation
+        into separate blocks: any leading prose directly under the marker
+        (untouched -- it rides along with the marker's own block and was
+        never going to be isolated), then each $$ block, then any prose
+        that follows. Each of those *isolated* segments needs its own
+        indentation raised to exactly tab_length if it falls short.
+
+        We must not pad them as one uniform block: if a shallower segment
+        pulled a $$ block past exactly tab_length, the excess would survive
+        list-item detabbing as a residual indent, which breaks arithmatex's
+        block match just as being under-indented does. And if a deeper
+        segment (e.g. a $$ block someone indented further than the prose
+        around it, deliberately or not) were left as the reference point,
+        a shallower prose segment would be under-padded and pop out of the
+        list.
+        """
+        j = start
+        seen_math = False
+        while j < end:
+            seg_start = j
+            if self.SINGLE_LINE_PATTERN.match(result[j]):
+                j += 1
+                seen_math = True
+            elif self.DOLLAR_DOLLAR_PATTERN.match(result[j]):
+                j += 1
+                while j < end and not self.DOLLAR_DOLLAR_PATTERN.match(
+                    result[j]
+                ):
+                    j += 1
+                if j < end:
+                    j += 1  # include the closing "$$" line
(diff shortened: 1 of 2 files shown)


SergeyAxiomatic:

SergeyAxiomatic:
recheck

mscolnick:
@cubic-dev-ai

SergeyAxiomatic:
CI status update — three checks are currently red, but none look related to this change:

- **enforce-label**: needs a label (`bug`/`enhancement`/etc.) applied per repo policy. I don't have permission to add labels as an external contributor — could a maintainer add one?
- **test** (Playwright e2e): fails on `kitchen-sink.spec.ts` with a `page.goto` timeout loading a local exported HTML file (`networkidle` never fires). The other 47 e2e tests in the same run passed.
- **windows-latest / Py 3.10 / core,optional deps**: fails in `test_scratchpad_integration.py` (HTTP 403 on a local request) and a worker crash in `test_asgi.py`. Looks like Windows-only networking flakiness — unrelated to markdown/math.

All markdown/display-math tests (`tests/_output/test_md.py`, including the new regression test) pass across every other OS/Python matrix job, along with lint and typecheck. Happy to help re-run or dig further if it'd help.

mscolnick:
@SergeyAxiomatic few comments from the PR review bots

SergeyAxiomatic:
Thanks @cubic-dev-ai — both issues were valid, fixed in 9cfabc3db:

1. **\"1)\" false marker match**: restricted `LIST_MARKER_PATTERN` to only \"N.\" (what python-markdown's `OList` processor actually recognizes), so prose like \"1) some text\" no longer gets misidentified as a list item and no longer pushes a following indented `$$` block into 4-space (indented-code) territory.
2. **Blank line inside a `$$` block**: the reindent run now stays open across a blank line while inside an unclosed `$$...$$` block, so the whole block (and the list item's trailing text) gets widened together instead of only the part before the first blank line. Note the math itself still won't render in that case — `pymdownx.arithmatex` can't recognize a display-math block containing an internal blank line even at the top level with zero indentation (confirmed against vanilla markdown+arithmatex, no list involved) — but the list no longer splits or pops the tail out, which was the actual bug here.

Added regression tests for both cases in `tests/_output/test_md.py`.

SergeyAxiomatic:
Pushed the cubic-review fixes in 9cfabc3db. Note that the main test workflows (Test BE, Test FE, playwright, cli test, Typos, and the label-enforcement check) are currently sitting in \`action_required\` for this commit and haven't run yet — looks like they need a maintainer to approve running them for this fork PR. CLA, lint (pre-commit.ci), and the label/Vercel checks have all passed.

Light2Dark:
@SergeyAxiomatic just checking if I have 
```md
- Item
  continuation
    $$x$$
- Next
```
I think $$x$$ indents by 6 characters instead of 4? I might be wrong.

SergeyAxiomatic:
@Light2Dark good catch — you were right, that's a real bug. Confirmed:

\`\`\`
- Item
  continuation
    \$\$x\$\$
- Next
\`\`\`

was padding \`\$\$x\$\$\` to 6 columns instead of leaving it at 4, and it broke rendering (came out as literal \`\$\` + inline math + literal \`\$\` instead of a display block).

Root cause: the old reindent widened an entire contiguous run by *one* delta computed from its shallowest line. But a run can mix prose that's already anchored to the marker's own paragraph (needs no padding at all, since it's never isolated by the blank-line insertion) with a \`\$\$\` construct that's already at or past \`tab_length\` on its own. Computing a single delta from the shallower "continuation" line pushed the already-sufficient \`\$\$x\$\$\` past 4 columns — and going *past* exactly \`tab_length\` is just as broken as falling short of it: the excess survives the list item's detab and leaves a residual indent that breaks arithmatex's block match.

Fixed in 45f93c098 by padding each blank-line-isolated segment (leading prose / math block / trailing prose) independently, based on its own minimum indentation, rather than one uniform delta across the whole run. Added a regression test (`test_md_display_math_deeper_indent_than_continuation`) covering your exact example.

SergeyAxiomatic:
@copilot-pull-request-reviewer thanks for the review. I tested all three flagged scenarios directly against the code (note: my first pass at reproducing these used Python triple-quoted strings that got silently corrupted by `_md()`'s `cleandoc()` call whenever every continuation line shared one indent level with no other column-0 line to anchor against — re-tested via `_render_markdown()` directly to get accurate results). Findings:

**Nested marker indentation (display_math.py:258)** — not reproduced. `FlexibleIndentPreprocessor` (priority 35) already normalizes nested list markers to absolute 4-space multiples *before* our preprocessor (priority 24) ever runs, so using the absolute `tab_length` as the threshold is correct here, not "relative to the nearest containing marker." Tested:
\`\`\`
- outer item
    - nested item
      continuation
        $$x$$
\`\`\`
renders with the nested item correctly inside a nested \`<ul>\` under \`outer item\`, math included.

**Nested marker inside a padded trailing segment (display_math.py:262)** — not reproduced either. Our per-segment padding applies one uniform additive delta to every line in a segment, which preserves relative indentation between a nested marker and its own sub-content regardless of the delta's size. Tested:
\`\`\`
- item
  $$
  x
  $$
  continuation
  - nested
    sub-item text
\`\`\`
renders with the nested list correctly inside the outer \`<li>\`.

**Blank line separating the marker from its continuation (display_math.py:182)** — confirmed the underlying failure mode, but it's pre-existing and unrelated to math or this PR. This already fails identically with zero math involved:
\`\`\`
- item

  more text
\`\`\`
renders \`more text\` as a top-level paragraph outside the \`<ul>\`, because 2-space indentation is below \`tab_length\` for `ListIndentProcessor` to reattach a loose block — true for any content. A 4-space-indented version of the same thing nests correctly, math or not. The original bug report (and this PR's scope) was specifically about *tight* continuations that our own blank-line insertion breaks apart — not about fixing indentation requirements for already-loose lists, which is a separate, pre-existing limitation of this markdown pipeline.

**Test coverage nits** — fair, added in 15f19863c: a parametrized `test_md_display_math_bullet_markers` covering `-`, `*`, and `+` (previously only `-` and `1.` were exercised for the marker regex). I didn't add a screenshot since this is server-side markdown→HTML string generation with no UI to screenshot; the assertions on the actual HTML output are the direct evidence.

Happy to dig further into any of these if you disagree with my repro.

Light2Dark:
## Browser verification and benchmark

Tested commit `639e79daf` using `marimo run marimo/_smoke_tests/display_math_lists.py` in a local browser.

- Numbered list: three items; the display equation and following prose remain inside item two.
- All three bullet markers (`-`, `*`, `+`): three items each; equation and following prose remain inside the second item.
- Mixed indentation: equation and following prose remain inside the first item of a two-item list.
- Four horizontal-rule variants: following prose remains ordinary text, not a code block.
- Inspected the rendered page and checked list ancestry in the browser DOM. No console warnings or errors were captured.

The existing literal-dollar rendering issue after horizontal rules remains visible. Nested-list and blank-line-separated math limitations discussed earlier are outside this fix; this verification does not claim to resolve them.

### Benchmark

Python 3.13.11 on macOS arm64. Synthetic input repeats a three-line item: list marker, indented `$$x^2$$`, then indented prose. Each result is the median of five batches after a warm-up: 10 calls per batch for the indentation pass, 3 for full `_md` rendering. The previous pass is from `ec3c4c7ef`; the simplified pass and full rendering use `639e79daf`. The script also checks that both passes produce identical output for these inputs.

| Items | Lines | Previous indentation pass | Simplified indentation pass | Full Python `_md` render, current |
| ---: | ---: | ---: | ---: | ---: |
| 10 | 30 | 0.042 ms | 0.038 ms | 0.805 ms |
| 100 | 300 | 0.433 ms | 0.401 ms | 6.622 ms |
| 1,000 | 3,000 | 4.179 ms | 3.849 ms | 64.399 ms |

These measurements are consistent with linear scaling. The indentation pass accounts for roughly 6% of the full Python render time at 3,000 lines. Small timing differences are machine-dependent; this is a synthetic microbenchmark, not a browser/KaTeX latency measurement.

<details>
<summary>Reproduce from the repository at 639e79daf</summary>

Save the following as `/tmp/benchmark_display_math.py`, then run `uv run python /tmp/benchmark_display_math.py`. The local Git history must include `ec3c4c7ef`.

```python
import platform
import subprocess
from statistics import median
from timeit import repeat
from types import ModuleType

from markdown import Markdown
from marimo._output.md import _md
from marimo._output.md_extensions.display_math import DisplayMathPreprocessor

previous = ModuleType('previous_display_math')
exec(subprocess.check_output([
    'git', 'show',
    'ec3c4c7ef:marimo/_output/md_extensions/display_math.py',
], text=True), previous.__dict__)
old = previous.DisplayMathPreprocessor(Markdown())
current = DisplayMathPreprocessor(Markdown())

def milliseconds(fn, iterations):
    fn()
    return median(repeat(fn, number=iterations, repeat=5)) * 1000 / iterations

print(f'Python {platform.python_version()}, {platform.system()} {platform.machine()}')
print('items | lines | previous pass ms | simplified pass ms | full render ms')
for items in (10, 100, 1000):
    text = '\n'.join(
        f'- Item {i}\n  $$x^2$$\n  trailing text' for i in range(items)
    )
    lines = text.splitlines()
    assert old._reindent_list_continuations(lines) == current._reindent_list_continuations(lines)
    old_ms = milliseconds(lambda: old._reindent_list_continuations(lines), 10)
    new_ms = milliseconds(lambda: current._reindent_list_continuations(lines), 10)
    render_ms = milliseconds(lambda: _md(text), 3)
    print(f'{items} | {len(lines)} | {old_ms:.3f} | {new_ms:.3f} | {render_ms:.3f}')
```

</details>

Posted by the coding agent after local verification.

Light2Dark:
### Visual verification

Smoke notebook screenshot showing display equations and following text within numbered and bulleted list items.

Display math rendered inside numbered and bulleted list items

kirangadhave (approved):

PR #10797, opened 2026-09-11
fix: preserve column metadata when saving from a non-columns view (#3543)

### LINKED ISSUE #3543: Switching from columns view to another view loses column metadata
### Describe the bug

* Create a notebook with a two columns, and arrange cells across the columns
* Switch to a compact view and save.

Expected behavior:
* Column metadata is still persisted in the notebook file.

Actual behavior:
* Column metadata is erased from the file.

### Environment

<details>

```
{
  "marimo": "0.10.16",
  "OS": "Darwin",
  "OS Version": "22.5.0",
  "Processor": "arm",
  "Python Version": "3.12.4",
  "Binaries": {
    "Browser": "131.0.6778.265",
    "Node": "v21.5.0"
  },
  "Dependencies": {
    "click": "8.1.8",
    "docutils": "0.21.2",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.7",
    "narwhals": "1.23.0",
    "packaging": "24.2",
    "psutil": "6.1.1",
    "pygments": "2.19.1",
    "pymdown-extensions": "10.14.1",
    "pyyaml": "6.0.2",
    "ruff": "0.9.2",
    "starlette": "0.45.2",
    "tomlkit": "0.13.2",
    "uvicorn": "0.34.0",
    "websockets": "14.2"
  },
  "Optional Dependencies": {}
}```

</details>


### Code to reproduce

```python
import marimo

__generated_with = "0.10.16"
app = marimo.App(width="columns")


@app.cell(column=0)
def _():
    "1"
    return


@app.cell(column=1)
def _():
    "2"
    return


if __name__ == "__main__":
    app.run()
```

Sushit-prog:
## 📝 Summary

Closes #3543

Switching from columns view to a non-columns view (e.g. compact) and saving
erased the per-cell `column` metadata from the `.py` file permanently. Root
cause: `getCellConfigs()` force-wrote `column: null` on save whenever the cell
tree was single-column, which happens after `mergeAllColumns()` collapses the
tree on view switch — even though the stored `config.column` values are never
cleared upstream.

## Description of Changes

- `frontend/src/core/cells/cells.ts`: in the single-column serialization
  branch of `getCellConfigs()`, preserve each cell's stored `config.column`
  (normalizing `undefined → null`) instead of overwriting it with `null`.
- The multi-column branch (which re-derives column indices from the tree) and
  all backend/codegen code are unchanged.
- Added two unit tests in `frontend/src/core/cells/__tests__/utils.test.ts`:
  1. single-column tree preserving `config.column` 0/1
  2. regression mirroring the issue repro: multi-column layout collapsed via
     `mergeAllColumns()` retains its column metadata
  Both fail against the pre-fix code (verified red/green).

### Acknowledged follow-up (not fixed here)

`deleteColumn` moves cells between columns without updating per-cell
`config.column`. If a user deletes columns down to a single one and saves, the
preserved metadata could restore an earlier layout on reload. This self-heals
on the next multi-column save.

## 📋 Checklist

- [ ] Documentation has been updated where applicable, including docstrings for API changes. — N/A: frontend-only fix, no public API/docs change.

## Test plan

- [x] `utils.test.ts`, `cells.test.ts`, `session.test.ts`, `src/core/saving/__tests__`: 159/159 pass
- [x] eslint + biome clean on changed files
- [x] `tsgo` typecheck clean (pre-existing `model-registry.ts` errors in a fresh checkout were caused by missing generated `@marimo-team/llm-info` JSON files; resolved by running codegen)

### DIFF
--- a/frontend/src/core/cells/__tests__/utils.test.ts
+++ b/frontend/src/core/cells/__tests__/utils.test.ts
@@ -102,6 +102,79 @@ describe("getCellConfigs", () => {
     ]);
   });
 
+  it("should preserve existing column metadata in a single-column tree", () => {
+    const cellId1 = CellId.create();
+    const cellId2 = CellId.create();
+    const mockState: NotebookState = {
+      ...initialNotebookState(),
+      cellIds: MultiColumn.from([[cellId1, cellId2]]),
+      cellData: {
+        [cellId1]: {
+          id: cellId1,
+          config: { hide_code: false, disabled: false, column: 0 },
+        } as CellData,
+        [cellId2]: {
+          id: cellId2,
+          config: { hide_code: true, disabled: false, column: 1 },
+        } as CellData,
+      },
+      cellRuntime: {} as Record<CellId, CellRuntimeState>,
+    };
+
+    const result = getCellConfigs(mockState);
+    expect(result).toEqual([
+      { hide_code: false, disabled: false, column: 0 },
+      { hide_code: true, disabled: false, column: 1 },
+    ]);
+  });
+
+  it("should preserve column metadata when a columns layout is collapsed via mergeAllColumns", () => {
+    const cellId1 = CellId.create();
+    const cellId2 = CellId.create();
+    const cellId3 = CellId.create();
+    const cellId4 = CellId.create();
+
+    // Simulate the issue repro: a notebook loaded with a multi-column layout
+    // (column metadata stored per-cell), then the view is switched away from
+    // "columns", which calls mergeAllColumns() and flattens the tree.
+    const collapsedCellIds = MultiColumn.from([
+      [cellId1, cellId2],
+      [cellId3, cellId4],
+    ]).mergeAllColumns();
+
+    const mockState: NotebookState = {
+      ...initialNotebookState(),
+      cellIds: collapsedCellIds,
+      cellData: {
+        [cellId1]: {
+          id: cellId1,
+          config: { hide_code: false, disabled: false, column: 0 },
+        } as CellData,
+        [cellId2]: {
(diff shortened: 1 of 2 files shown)


Sushit-prog:

mscolnick:
Thank you - have you tested this? I dont think this actually worked. Can you share a video? I am going to put in draft since i can't get it to work

Sushit-prog:
Thanks for testing this @mscolnick  and catching it you were right, the original fix didn't actually work in a live session.

Root cause of the original failure: cellData[cellId].config.column (the field my first fix relied on) is only ever populated when a notebook loads from disk. No reducer writes to it during a live session. Meanwhile the multi-column branch of getCellConfigs() derives the saved column index fresh from the cellIds tree position each time, but never writes that value back into cellData. So in a real session — create columns, drag cells around, switch width away from "columns" — the field my fix read from was stale/never-set, and it saved null regardless.

Fix: added a snapshotColumnConfigs() helper that writes the tree-derived column index into cellData at the three points where the column tree can collapse to one column: mergeAllColumns, deleteColumn, and compactColumns (the last one also fires on live drag-and-drop, so it needed the same treatment).

Video below shows the actual failure scenario end-to-end: multi-column layout → switch width away from "columns" → save → reload the page → switch back to "columns" → column assignment is preserved.



https://github.com/user-attachments/assets/d97d9b97-8088-4ecd-8406-bccf00e3290d


Also added reducer-level tests for all three cases that fail on the previous diff and pass with this one. Took it out of draft let me know if you want me to walk through anything else.

PR #10809, opened 2026-09-12
feat: make the session cookie's SameSite configurable

0x0f0f0f:

`CustomSessionMiddleware` already accepts `same_site`, but the only construction site never passes it — so the session cookie is always `SameSite=Lax`.

That makes marimo unusable in a cross-origin iframe: the cookie is third-party, browsers that restrict third-party cookies drop it, the next request is unauthenticated, and marimo redirects to `/auth/login` — which it serves with `X-Frame-Options: DENY`. Blank frame, no recovery.

**What this adds:** `MARIMO_SESSION_COOKIE_SAMESITE`, beside the existing `MARIMO_SESSION_COOKIE_SECURE`. Default unchanged (`lax`).

**One judgement call:** `same_site="none"` implies `Secure`. Browsers reject `None` without it, so honouring the pair as given would hand back a silentropped cookie — the exact bug being fixed.

Found from marimohub, where kernels are framed cross-site under its default `subdomain` exposure: marimo-team/marimohub#328

Happy to move this to an issue first if you would rather discuss the shape.

### DIFF
--- a/docs/guides/configuration/index.md
+++ b/docs/guides/configuration/index.md
@@ -141,6 +141,7 @@ marimo supports the following environment variables for advanced configuration:
 | `MARIMO_SKIP_UPDATE_CHECK`    | If set to "1", marimo will skip checking for updates when starting.                                                          | Not set         |
 | `MARIMO_SQL_DEFAULT_LIMIT`    | Default limit for SQL query results. If not set, no limit is applied.                                                        | Not set         |
 | `MARIMO_SESSION_COOKIE_SECURE` | If set to `true`/`1`, marks the session cookie as `Secure` so browsers only send it over HTTPS. Enable when serving marimo behind TLS.        | `false`         |
+| `MARIMO_SESSION_COOKIE_SAMESITE` | `SameSite` attribute of the session cookie: `lax`, `strict`, or `none`. Set `none` when marimo is embedded in an iframe on another origin — the cookie is third-party there, and browsers that restrict third-party cookies drop it under `lax`, so token auth can never complete. `none` implies `Secure` and therefore requires HTTPS. | `lax` |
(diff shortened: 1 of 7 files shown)


marimo/_utils/env.py
```diff
@@ -2,6 +2,37 @@
...
+    if normalized in choices:
+        return cast("T", normalized)
+    LOGGER.warning(
+        "%s=%r is not one of %s; falling back to %r.",
+        key,
+        value,
+        ", ".join(choices),
+        default,
+    )
+    return default
```
0x0f0f0f, line 35: Could be stringy typed if this check is excessive. Also a fallback when the config could def become a failure ?

docs/guides/configuration/index.md
```diff
@@ -141,6 +141,7 @@ marimo supports the following environment variables for advanced configuration:
 | `MARIMO_SKIP_UPDATE_CHECK`    | If set to "1", marimo will skip checking for updates when starting.                                                          | Not set         |
 | `MARIMO_SQL_DEFAULT_LIMIT`    | Default limit for SQL query results. If not set, no limit is applied.                                                        | Not set         |
 | `MARIMO_SESSION_COOKIE_SECURE` | If set to `true`/`1`, marks the session cookie as `Secure` so browsers only send it over HTTPS. Enable when serving marimo behind TLS.        | `false`         |
+| `MARIMO_SESSION_COOKIE_SAMESITE` | `SameSite` attribute of the session cookie: `lax`, `strict`, or `none`. Set `none` when marimo is embedded in an iframe on another origin — the cookie is third-party there, and browsers that restrict third-party cookies drop it under `lax`, so token auth can never complete. `none` implies `Secure` and therefore requires HTTPS. | `lax` |
```
mscolnick, line 144: i think this is fine - we may want to go with [Cookies Having Independent Partitioned State (CHIPS)
](https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Third-party_cookies/Partitioned_cookies) which might be more secure than forcing just SameSite=None. Im looking into this more as an option

mscolnick:
Going to close this. May re-open with a flag for Partition cookies (CHIPS) instead of overriding the same site

PR #10814, opened 2026-09-13
feat: automatically mount Lens when installed

### LINKED ISSUE #9619: Better referencing for cells
### Description

Working in /marimo-pair requires identifying cells that you want work done to. Currently, you can name each cell, but that can be a chore. Each cell has an id and you can ask the agent to set the cell name to the cell-id. That works good but it should be a feature of the platform rather than a request in the agent.  Further, it would be good if the name gave some clue about where the cell was positionally in the notebook. Excel does this with column and row designations.  Marimo currently uses 4 alphabet characters that can be capital or lower case.  It would be an improvement to create a predictable naming convention that helped describe where the cell is in the notebook. Since marimo allows mutliple columns, the excel style <letter><number> reference could work and is descriptive.  When working with /marimo-pair, I may be working on 10 different cells at once. Enabling a predictable cell id naming scheme makes it much easier to do this. Also shortening the cell id to fewer characters is better as it decreases the risk of typos in referencing it. 

- [ ] Yes

### Alternatives

_No response_

### Additional context

_No response_

peter-gy:
Ensures that whenever a user has [marimo-lens](http://pypi.org/project/marimo-lens) installed as a package in their python env, then a lens instance gets automatically mounted to their notebook for better agentic collab.

Closes #9619

### DIFF
--- a/marimo/_runtime/runner/hooks_lens.py
+++ b/marimo/_runtime/runner/hooks_lens.py
@@ -19,8 +19,6 @@ LOGGER = _loggers.marimo_logger()
 def mount_lens(
     cell: CellImpl, ctx: PostExecutionHookContext, result: RunResult
 ) -> None:
-    from importlib.util import find_spec
-
     del ctx
 
     if (
@@ -31,10 +29,12 @@ def mount_lens(
         return
 
     try:
-        if find_spec("marimo_lens") is None:
+        try:
+            from marimo_lens import Lens  # type: ignore[import-not-found]
+        except ModuleNotFoundError as exc:
+            if exc.name != "marimo_lens":
+                raise
             return
-
-        from marimo_lens import Lens  # type: ignore[import-not-found]
 
         lens = Lens()
         cell.set_output((cell.output, lens))
--- a/marimo/_runtime/runner/hooks_lens.py
+++ b/marimo/_runtime/runner/hooks_lens.py
@@ -1,0 +1,47 @@
+# Copyright 2026 Marimo. All rights reserved.
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from marimo import _loggers
+from marimo._runtime import output
+from marimo._runtime.context.utils import running_in_notebook
+
+if TYPE_CHECKING:
+    from marimo._ast.cell import CellImpl
+    from marimo._runtime.runner.hook_context import PostExecutionHookContext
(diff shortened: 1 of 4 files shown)


marimo/_runtime/runner/hooks_lens.py
```diff
@@ -0,0 +1,47 @@
...
+        except ModuleNotFoundError as exc:
+            if exc.name != "marimo_lens":
+                raise
+            return
+
+        lens = Lens()
+        cell.set_output((cell.output, lens))
+        if result.output is not None:
+            output.append(result.output)
+        output.append(lens)
```
mscolnick, line 48: does this get added to every cell? should we only add it once?

peter-gy: oh right. I had this initial logic directly in __init__.py then factored out into a hook and did not consider it would run multiple times. thanks!

marimo/_runtime/runner/hooks_lens.py
```diff
@@ -0,0 +1,52 @@
...
+        try:
+            from marimo_lens import Lens  # type: ignore[import-not-found]
+        except ModuleNotFoundError as exc:
+            if exc.name != "marimo_lens":
+                raise
+            return
+
+        if any(
+            contains_instance(other.output, Lens)
+            for other in ctx.graph.cells.values()
```
mscolnick, line 40: this could be expensive depending on outputs (likely not common). this is fine for this PR, but i do wonder if we have a non-cell "output area" or headless anywidgets that can be placed somewhere / initialized once.

mscolnick (approved):

marimo/_runtime/runner/hooks_lens.py
peter-gy: looking into that in a follow up

marimo/_runtime/runner/hooks_lens.py
peter-gy: @mscolnick Created a short RFC on it: [https://app.notion.com/p/marimo-team/RFC-Notebook-scoped-anywidgets-3dcdbf71a05880a1a498c3c9b81e24c1?source=copy_link](<https://app.notion.com/p/marimo-team/RFC-Notebook-scoped-anywidgets-3dcdbf71a05880a1a498c3c9b81e24c1?source=copy_link>)

PR #10816, opened 2026-09-13
fix: unify Polars numeric checks and restore cache version 5

peter-gy:
Brief follow-up on https://github.com/marimo-team/marimo/pull/10775

- Unified numeric classification for Polars DataFrames and Series
- Reverted the unnecessary cache-version bump, with tests confirming versions 5 and 6 remain readable


marimo/_runtime/primitives.py
```diff
@@ -76,9 +76,9 @@ def is_data_primitive(value: Any) -> bool:
     if is_instance_by_name(value, "torch.Tensor"):
         return str(value.device) == "cpu"
 
-    # Series follow the same numeric-schema policy as dataframes.
-    if is_instance_by_name(value, "polars.series.series.Series"):
-        return bool(value.dtype.is_numeric())
+    if type(value).__module__.startswith("polars."):
+        dtypes = value.dtypes if hasattr(value, "dtypes") else [value.dtype]
+        return all(dtype.is_numeric() for dtype in dtypes)
```
dmadisetti: Isn't this the same as below? I think we can consolidate this logic while we're here

marimo/_save/cache.py
```diff
@@ -46,7 +46,7 @@
     from marimo._save.stores import Store
 
 # NB. Increment on cache breaking changes.
-MARIMO_CACHE_VERSION: int = 6
+MARIMO_CACHE_VERSION: int = 5
```
dmadisetti: It's fine, I am going to cache invalidate soon. Will put you on that PR

peter-gy: okay, so will revert to `= 6`

dmadisetti (approved):

PR #10838, opened 2026-09-14
feat: record cache entries in a per-notebook manifest

dmadisetti:
Stacked on #10837; review against `dm/cache-size`.

## 📝 Summary

Every persistent cache hit or save now attests its entry in a manifest
beside the cache, grouped under a path hash: a static digest of the
defining cell and every cell it depends on, recomputable from source
alone (exposed as BlockHasher.path_hash and hash_cell_closure). Records
accumulate across runs and travel through the Store interface under a
reserved dot key, so remote and tiered stores carry manifests too, each
attesting only the entries stored beside it. Flushes coalesce to one
write per run (ON_FINISH hook, teardown retry); recording never fails
the user's cell. FileStore.put is now atomic (tmp sibling + os.replace)
and interrupted-write leftovers are excluded from stats and swept by
clear().

Also fixes unmangle_local for hyphenated cell ids (a UUID, or a nested
app's prefix): a mangled _private name was previously left mangled in
co_names, so kernel and load_app digests disagreed for such sessions.


marimo/_save/stores/store.py
```diff
@@ -16,7 +16,14 @@ def get(self, key: str) -> bytes | None:
...
     def put(self, key: str, value: bytes) -> bool:
-        """Put a cache into the store"""
+        """Put a cache into the store.
+
+        A key holds either its previous value or the whole new one. A reader
+        racing a writer, or arriving after one died, never sees a truncated
+        value. Cache writes depend on it. A lazy entry's completeness marker
+        is written last, and the marker means nothing if a partial write can
+        be read as a complete one.
+        """
```
dmadisetti:
```suggestion
        """Put a cache into the store."""
```

mscolnick (approved):

PR #10855, opened 2026-09-15
fix: fall back to CSV for Altair charts without pyarrow

### LINKED ISSUE #10764: Sandbox mode omits pyarrow from inline dependencies needed by mo.ui.altair_chart, causing crashes new Windows users can't self-diagnose
### Describe the bug

When a notebook is started with `--sandbox`, marimo scans the user's own `import` statements to populate the PEP 723 inline `dependencies` list. However, `mo.ui.altair_chart()` has a hidden runtime dependency on `pyarrow` (used internally in `altair_transformer.py`  --> `pandas_table.py`'s `to_arrow_ipc()` to serialize chart data for the frontend). Since the user never imports `pyarrow` directly, the sandbox dependency scanner never adds it, so it is silently missing from the venv `uv` builds for the notebook.

The result: running any cell with `mo.ui.altair_chart()` throws `ModuleNotFoundError: No module named 'pyarrow'` internally. On Windows specifically, the fallback CSV-path logging call in the exception handler then itself crashes with `OSError: [WinError 1] Incorrect function` while trying to write the log message, producing a second, more alarming traceback on top of the first.

Marimo docs say dependencies are automatically detected but here, they're not, and the actual problems comes from Marimo's own internal code, not the user's code, so there's nothing in the user's cell to "fix.".  It might scare off new Windows users.

### Steps to reproduce

1. `uvx marimo edit --sandbox notebook.py` on Windows, with a fresh notebook that doesn't have `pyarrow` in `dependencies`.
2. Create a cell that builds a `pandas.DataFrame`, wraps an Altair chart in `mo.ui.altair_chart()`, and displays it.
3. Run the cell.

### Expected behavior

One of:
- The sandbox dependency scanner should recognize that `mo.ui.altair_chart` (and similar marimo UI elements with hidden native dependencies like `pyarrow`, `vegafusion`, etc.) requires `pyarrow` and inline it automatically, the same way `marimo` itself is always inlined.
- Or, at minimum, `mo.ui.altair_chart` should respect the CSV/JSON fallback transformer gracefully without throwing an uncaught `ModuleNotFoundError`-triggered logging failure — the user should see one clear, actionable message ("pyarrow not found; falling back to CSV serialization; install pyarrow for better performance") instead of two stack traces.
- The Windows logging crash (`OSError: [WinError 1] Incorrect function`) in the exception-handling path should be fixed regardless, since it actively suppresses the real underlying error message.

### Environment

```
OS: Windows 11
Python: 3.12 (via uv-managed cpython-3.12-windows-x86_64)
marimo: sandbox mode via uvx
Dependencies detected in notebook header: altair==6.2.2, marimo, matplotlib==3.11.1, numpy==2.5.2, pandas==3.0.5, plotly==7.0.0, scipy==1.18.1
```

### Code to reproduce

```python
# /// script
# dependencies = [
#     "altair==6.2.2",
#     "marimo",
#     "pandas==3.0.5",
# ]
# requires-python = ">=3.12"
# ///

import marimo
app = marimo.App()

@app.cell
def _():
    import marimo as mo
    import pandas as pd
    import altair as alt
    df = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
    chart = alt.Chart(df).mark_line().encode(x="x:Q", y="y:Q")
    mo.ui.altair_chart(chart)
    return
```

### Related issues

Related to #8004, which addressed `mo.ui.altair_chart` hardcoding the `marimo_arrow` transformer and ignoring user-configured transformers — but that issue focused on WASM/Pyodide contexts and was resolved without addressing the sandbox dependency-inlining gap or the Windows-specific logging crash described here.

### Will you submit a PR?
- [ ] No

mscolnick:

Check for pyarrow before serializing pandas data to Arrow, allowing Altair charts to fall back quietly to CSV when it is missing. This avoids the reported Windows logging failure and preserves Polars Arrow support.

Fixes #10764.


marimo/_plugins/ui/_impl/tables/pandas_table.py
```diff
@@ -402,6 +403,11 @@ def _infer_dtype(self, column: ColumnName) -> str:
                 return pd.api.types.infer_dtype(self._original_data[column])
 
             def to_arrow_ipc(self) -> bytes:
+                if not DependencyManager.pyarrow.has():
+                    raise NotImplementedError(
+                        "pyarrow is required for Arrow serialization"
+                    )
```
Light2Dark, line 409: Should we do `DependencyManager.pyarrow.require()` to trigger installation?

mscolnick: we have a good fallback that it would be nice just to avoid installation. this also works in wasm or apps that dont support installation popups

Light2Dark (approved):

PR #10856, opened 2026-09-15
fix: render dotted chart columns when data falls back to CSV

### LINKED ISSUE #10847: Chart builder renders an empty chart for columns whose name contains a `.`, when the data falls back to CSV
### What happens

In a `mo.ui.table` **Visualize** tab, a column whose name contains a `.` draws an empty chart — no marks, the numeric axis collapsed to a single tick — but only when the table also contains a `timedelta64` column.

### Repro

marimo 0.24.0, pandas 3.0.5, altair 6.2.2, pyarrow 25.0.1

```python
import marimo as mo, pandas as pd

df = pd.DataFrame({"a.b": [1.0, 2.0, 3.0], "n": [1, 2, 3]})
df["d"] = pd.to_timedelta([1, 2, 3], unit="D")   # remove this line and it works
mo.ui.table(df)
```

Open **Visualize**, set X = `a.b` and Y = `n`. The plot area is blank — no marks, no tick labels. Drop the `d` column and the identical chart renders three bars.

The **Python code** tab shows the field is escaped: `x=alt.X(field='a\\.b', type='quantitative')`.

### Cause

Two workarounds for the same Vega-Lite behaviour (a `.` in `field` means nested access) that cancel each other out:

- `replacePeriodsInColumnNames` (`vega/loader.ts`, #1455) rewrites the CSV header, `.` → U+2024
- `escapeFieldName` (`data-table/charts/chart-spec/utils.ts`, #6973) escapes the field, `.` → `\.`

Vega unescapes `a\.b` back to the literal key `a.b`, which the rewritten data no longer has. Every row reads null and Vega-Lite's invalid-value filter drops all of them.

Whether the two ever meet depends on the transport, and that is decided in `_maybe_sanitize_dataframe`:

```python
        return res.to_native()   # success: native frame
    except Exception as e:
        LOGGER.warning(f"Failed to sanitize narwhals dataframe: {e}")
        return data              # failure: narwhals wrapper — no to_arrow_ipc
```

The wrapper has no Arrow export, so `_to_marimo_arrow` falls back to CSV, which is where the header rewriting lives. Altair's sanitizer refuses `timedelta64`, so any duration column triggers the fallback.

### Suggested fixes

1. `altair_transformer.py`: return `data.to_native()` on the failure path. Arrow export then succeeds with the timedelta still present (checked locally), which fixes the reported symptom and removes a silent transport downgrade.
2. Make the loader and the chart spec agree — either have `escapeFieldName` apply the same U+2024 substitution, or stop renaming in the loader now that consumers escape.

(1) alone does not fix the underlying mismatch: a dotted column name still breaks on any genuine CSV fallback.

### Will you submit a PR?

- [ ] Yes

### Environment

{
  "marimo": "0.24.0",
  "editable": false,
  "location": "~/Developer/marimo-pm/.venv/lib/python3.13/site-packages/marimo",
  "OS": "Linux",
  "OS Version": "7.1.13-200.fc44.x86_64",
  "Processor": "",
  "Python Version": "3.13.5",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "Mozilla/5.0 (X11; Linux x86_64; rv:155.0) Gecko/20100101 Firefox/155.0",
    "Node": "v22.23.1",
    "uv": "0.11.12 (x86_64-unknown-linux-gnu)"
  },
  "Dependencies": {
    "click": "8.5.0",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.3",
    "narwhals": "2.25.0",
    "packaging": "26.3",
    "psutil": "7.2.2",
    "pygments": "2.21.0",
    "pymdown-extensions": "11.0.2",
    "pyyaml": "6.0.3",
    "starlette": "1.6.0",
    "tomlkit": "0.15.1",
    "uvicorn": "0.52.4",
    "websockets": "17.1"
  },
  "Optional Dependencies": {
    "altair": "6.2.2",
    "anywidget": "0.11.0",
    "loro": "1.13.2",
    "pandas": "3.0.5",
    "polars": "1.44.1",
    "pyarrow": "25.0.1"
  },
  "Experimental Flags": {}
}

### Code to reproduce

```python
import marimo as mo, pandas as pd

df = pd.DataFrame({"a.b": [1.0, 2.0, 3.0], "n": [1, 2, 3]})
df["d"] = pd.to_timedelta([1, 2, 3], unit="D")   # remove this line and it works
mo.ui.table(df)
```

Light2Dark:
## 📝 Summary

Closes #10847

The table's Visualize tab renders an empty chart for dotted column names when data falls back to CSV. The loader renames CSV headers while chart encodings reference the escaped original names. Preserve the original headers in the chart builder so encodings work across CSV, JSON, and Arrow.

When Altair sanitization fails, return the native dataframe so supported backends can still export Arrow. This avoids an unnecessary CSV fallback for data containing duration columns. A new test verifies that Arrow preserves the dotted column and duration values.

Browser verification on the PR code passed for both the issue reproduction using Arrow and a forced CSV response. With the duration column still present, selecting `a.b` for X and `n` for Y rendered three bars at `(1, 1)`, `(2, 2)`, and `(3, 3)` with correct axes in both cases. The Arrow case also confirmed that export succeeds after Altair rejects the duration column during sanitization.

### Screenshot

Chart with dotted user.name column showing alice and bob

## 📋 Pre-Review Checklist

- [ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on Discord, or the community discussions.

## ✅ Merge Checklist

> Written by GPT-6 on Codex


Light2Dark:
Follow-up findings and fix:

The CSV fallback review finding was valid. After a Narwhals sanitizer failure, returning native pandas data allows Arrow export, but calling `_to_marimo_csv()` after an Arrow failure sanitizes the frame again. That second pass rejects duration columns and prevents the CSV fallback.

The fallback now serializes the frame through the same table manager without repeating sanitization. Tests cover missing PyArrow, Arrow export raising `ValueError` or `NotImplementedError`, and preservation of finite numbers, both infinities, and missing values in CSV.

One intentional behavior difference: for Narwhals-wrapped pandas data, the old second sanitization converted `inf` and `-inf` to missing values. The fallback now preserves them. Infinity and missing data carry different information, so matching that old output would discard information. This does not change every serializer's existing sanitization policy.

There is a separate chart parsing concern: Vega's automatic CSV parsing treats a column containing the strings `inf` and `-inf` as strings, including its finite values. Explicit numeric parsing and consistent non-finite handling across Arrow/CSV/JSON need separate chart-boundary work. This PR does not claim to resolve that concern.

Verification:

- 77 transformer tests pass, including the new fallback and preservation cases.
- Broader chart/Vega frontend suites: 239 passed, 1 expected failure.
- Broader Python chart/table/formatter suites before the final preservation test: 465 passed, 2 skipped, 7 failures. All seven also reproduce with the previous transformer implementation: six Altair schema-version snapshot mismatches and one DuckDB all-null binning failure.
- Compared 72 CSV fallback cases across dataframe backends, native/wrapped inputs, numeric values, strings, missing values, dates, timestamps, and booleans. The infinity case was the only output difference.
- Browser tests exercised the actual backend fallback with a blocked PyArrow import and an injected Arrow export error. Both rendered three bars at `(1, 1)`, `(2, 2)`, `(3, 3)` for `a.b` versus `n`, with the duration column retained.
- Frontend lint/typechecking and Python lint/formatting pass. Full Python typechecking still reports errors outside the changed files.

mscolnick (approved):

PR #10866, opened 2026-09-16
fix: boolean column stats counted nulls as False, displaying incorrect totals

### LINKED ISSUE #10867: Boolean column summary counts nulls as False and hides the chart when a count is zero
### Describe the bug

Column summaries for boolean columns report wrong counts and can render a blank chart.

**Nulls are counted as `False`.** `NarwhalsTableManager` computes the false count as `len - sum(col)`. Null values are not `True`, so they land in the false bucket. A column with 50 `True`, 40 `False`, and 10 `None` reports `True: 50, False: 50, Null: 10`. This affects polars, pandas (nullable `boolean` dtype), pyarrow, and duckdb.

**A zero count hides the chart.** The frontend falls back to the legacy boolean spec when `stats.true` or `stats.false` is falsy. Zero is falsy. The legacy spec aggregates over the raw chart data, which the backend only sends when bin aggregation fails. In the normal path that data is empty, so a boolean column with no `False` values (or no `True` values) shows no chart.

### Will you submit a PR?

- [ ] Yes

### Environment

marimo 0.24.2, verified on `main` at 11905815ef.

boydgreenfield:
## 📝 Summary

The narwhals table manager computed the false count as len - sum(col), so null values were reported as False in column summaries. For example, a series of 50 `True` values, 40 `False` values, and 10 `None` values would report "True: 50, False: 50, Null: 10".

CleanShot 2026-09-15 at 22 18 52@2x

This is fixed by subtracting the null count from the total (not just the `False` count). The frontend also falls back to a legacy boolean spec when either count was zero (a falsy check), which would hide the column summary chart if there are no `False` values. This now falls back only if both counts are null (vs. 0).

## 📋 Pre-Review Checklist
- ~[ ] For large changes, or changes that affect the public API: this change was discussed or approved through an issue, on [Discord](https://marimo.io/discord?ref=pr), or the community [discussions](https://github.com/marimo-team/marimo/discussions) (Please provide a link if applicable).~ probably n/a but happy to open an issue for discussion if helpful

## ✅ Merge Checklist
- ~[] Documentation has been updated where applicable, including docstrings for API changes.~ n/a

Fixes #10867


boydgreenfield:

boydgreenfield:
recheck

kirangadhave (approved):
Great catch @boydgreenfield and thanks for the fix.

PR #10875, opened 2026-09-16
feat: open the issue-report modal from the error fallback

kirangadhave:
## 📝 Summary

Adds **Report an issue** to the shared React error fallback so a crashed cell (or a startup/root/island failure) can open the existing issue-report modal instead of a plain GitHub link.

The fallback owns a local dialog because a root crash removes `ModalProvider`. `FeedbackModal` reads the nullable `requestClientAtom` instead of `useRequestClient()`, because that hook throws when the client is still unset — which is exactly when this screen can appear. Other callers keep the strict hook. A missing client yields partial environment details, no Retry, and disabled notebook source.

Related: [MO-7659](https://linear.app/marimo/issue/MO-7659)


frontend/src/components/editor/chrome/components/feedback-button.tsx
```diff
@@ -105,15 +107,18 @@ export const FeedbackModal: React.FC<{
...
-      : !codeAvailable
-        ? "Notebook source is hidden in this view."
-        : "Connect the notebook to include its source.";
+    : requestClient == null
+      ? "Notebook source is unavailable."
+      : filename === null
+        ? "Save the notebook first."
+        : !codeAvailable
+          ? "Notebook source is hidden in this view."
+          : "Connect the notebook to include its source.";
```
Light2Dark: nit, but can we avoid this nested ternary for readability

Light2Dark (dismissed):

frontend/src/components/editor/chrome/components/feedback-button.tsx
kirangadhave: done!

Light2Dark (approved):

PR #10877, opened 2026-09-17
fix: respect .marimo.toml enforcement for setting keys

dmadisetti:
## 📝 Summary

Adds additional settings to be read only from the user configuration.
Values for these keys in `pyproject.toml` or in script metadata are ignored with a warning. The workspace `.marimo.toml` is unchanged.


kirangadhave (approved):

dmadisetti:
Ehhh, closing out. I think this is breaking and essentially the save as running uv sync or pytest on a untrusted repo.

PR #10879, opened 2026-09-17
feat: self-contained WASM export bundles for offline use

### LINKED ISSUE #9811: Configurable base URLs for retrieving WASM/wheel assets in html-wasm exports (to support self-contained/offline use)
### Description

I have a bunch of blog posts I authored as Jupyter notebooks interleaving code and prose. I'd like to make them more like "explorable explanations" (another Bret Victor fan here), but there's a lot of Pyodide plumbing required to get live-updating visualizations and embedded dynamic values in the prose output. It seems like most of that plumbing has been done already as part of Marimo... but I really want to self-host all the necessary files, and there's hard-coded (as far as I can tell?) references to specific remote hosts where Marimo or Pyodide retrieve essential components of the runtime.

Another issue report indicated there's no way to make it work offline (marimo-team/marimo#575), which would also be something that could be fixed with a hook to override these URLs.

It would be useful to at least add support for overriding these CDN URLs somewhere, because then it would be technically possible to self-host the assets for a WASM notebook (but still up to the user to figure out how). It would also be useful for the use case of just archiving a static notebook such that it can be opened in the future no matter what changes with newer upstream packages.

- [X] Yes

### Alternatives

It would be technically possible to vendor the assets and regex match and rewrite the URLs, but brittle.

(I've also tried building my functionality separately on top of Pyodide instead of Marimo, but ended up needing to make custom widgets, dependency tracking, etc. that Marimo already has.)

### Additional context

*No response*

peter-gy:
Adds `--offline` to bundle Pyodide and notebook dependencies alongside HTML-WASM exports:

```bash
uv run --with playwright python -m playwright install chromium

uv run --with playwright marimo export html-wasm notebook.py \
  --offline --no-sandbox -o dist
```

Exporting requires internet access and Playwright/Chromium. Pyodide resolves browser-compatible dependencies without executing notebook cells. The generated folder uses relative URLs and can be moved or served offline.

Runtime URL configuration remains separate from bundling, so granular source overrides can be exposed later. Existing exports retain their hosted defaults.

The export still requires a reachable HTTP server. It does not add browser offline caching, and notebook data/API calls & widget JavaScript assets need separate handling.

Closes #9811.


marimo/_export/offline.py
```diff
@@ -0,0 +1,309 @@
...
+    page_url: str,
+    index_url: str,
+    package_base_url: str,
+    lockfile: _Lockfile,
+    pypi_index_url: str | None,
+) -> _Lockfile:
+    from playwright.async_api import async_playwright  # type: ignore[import-not-found]
+
+    async with async_playwright() as playwright:
+        browser = await playwright.chromium.launch()
```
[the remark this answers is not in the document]
peter-gy: This finding looks invalid. playwright already launches chromium with `--no-sandbox` by default. marimo's `--sandbox` flag controls Python environment isolation and does not affect browser sandboxing.

frontend/src/core/wasm/runtime-config.ts
```diff
@@ -0,0 +1,27 @@
+/* Copyright 2026 Marimo. All rights reserved. */
+
+export interface WasmRuntimeConfig {
+  pyodideIndexUrl?: string;
+  pyodideLockfileUrl?: string;
+  pypiIndexUrl?: string;
```
mscolnick, line 6: I wonder if this is how we should be sending marimo version too (not required in this PR). This is much cleaner than shoving in the worker name.

mscolnick (dismissed):
this is a really great approach! great stuff

frontend/src/core/wasm/runtime-config.ts
peter-gy: Sure, we can do that in a follow up

mscolnick (approved):

PR #10883, opened 2026-09-17
refactor(frontend): move Tailwind theme to CSS-first config

kirangadhave:
## 📝 Summary

We have been running Tailwind v4 with a v3-style `tailwind.config.cjs` still doing most of the theming. That means two places define the design tokens, and the CSS side had to carry a `color-mix` shim (plus 24 `@supports` fallbacks) just to get alpha variants like `bg-primary/25` working. This PR finishes the move to CSS-first config so `globals.css` is the single source of truth.

Changes:
- Colors move into `@theme inline`. v4 handles `/alpha` natively on CSS variable tokens, so the shim and its fallbacks go away.
- Fonts, radius, keyframes, animations, and the grid scale follow. The little custom plugin becomes `@custom-variant fullscreen` and `@utility increase-pointer-area-x`, and `tailwindcss-animate` loads via `@plugin`. The cjs file is now just `darkMode`, `content`, and the typography plugin.
- The 25 arbitrary `text-[...]` font sizes collapse into seven named tokens (`text-3xs`, `text-2xs`, `text-xs-plus`, `text-sm-minus`, `text-10`, `text-11`, `text-13`).

Closes MO-6582


dangooddd:
Is that PR step into more customizable custom css, for example syntax HL customization?

Issue #10880, opened 2026-09-17
Inline completion fails with thinking models: "Model token limit (1024) exceeded before any response was generated"


tomschelsen:
### Bug

Marimo logs:

```
[E ... ai:352] Error in AI inline completion: Model token limit (1024) exceeded before any response was generated. Increase the max_tokens model setting, or simplify the prompt to result in a shorter response that will fit within the limit.
```

This is not a context-window issue — the serving model's `max_model_len` is 262144.

### Root cause

- Inline completion hardcodes its output budget and ignores `ai.max_tokens`: `INLINE_COMPLETION_MAX_TOKENS = 1024` in `marimo/_server/api/endpoints/ai.py`.
- Lots of hybrid-thinking models default to **thinking mode ON**, which vLLM applies via the model chat template. The model spends the entire 1024-token budget on reasoning and emits no final code, so Pydantic AI raises the error above.
- Marimo currently has no way to disable thinking on these requests (`CustomProvider._default_thinking` returns `None` for OpenAI-compatible providers, and `AnyProviderConfig` has no `chat_template_kwargs` / `enable_thinking` passthrough).

Confirmed thinking is the cause: the same request with `"chat_template_kwargs": {"enable_thinking": false}` returns code and fits within 1024 tokens.

### Expected behavior

Inline completion works with thinking models, or degrades gracefully instead of erroring.

### Suggested fix

Either:
- Make inline completion thinking-aware — disable thinking for these requests (e.g. send `chat_template_kwargs={"enable_thinking": false}` where supported, and/or suppress reasoning). Autocomplete should be fast and non-reasoning anyway, or:
- Make `INLINE_COMPLETION_MAX_TOKENS` configurable under `[ai]` rather than hardcoded to 1024.

### Will you submit a PR?

- [ ] Yes

### Environment

- marimo: `0.24.2`
- model: `Qwen3.8-27B-FP8` via vLLM (OpenAI-compatible provider)
- vLLM: `v0.28.0`

### Code to reproduce

1. Serve a thinking model with vLLM.
2. Configure a custom OpenAI-compatible provider in marimo pointing at it.
3. Type in a cell to trigger inline autocomplete.
4. Observe the error in the terminal.

PR #10890, opened 2026-09-18
fix: disable thinking for inline autocomplete

Light2Dark:
## 📝 Summary

Inline autocomplete currently inherits the provider's thinking defaults, which can spend its small output budget on reasoning before producing code. Explicitly pass `thinking=False` to Pydantic AI for inline completion as a per-run override. Chat and editing retain their existing settings, and the autocomplete budget remains 1,024 tokens.

Related to #10880; this does not fully resolve the reported custom Qwen/vLLM case.

### Upstream limitation

A mocked HTTP transport with Pydantic AI 2.42.0 confirmed that its generic OpenAI profile marks `Qwen3.8-27B-FP8` as `supports_thinking=False`. Consequently, `thinking=False` produces the same outgoing request as omitting the setting. For a recognized model such as GPT-5.1, it translates to `reasoning_effort="none"`. Overriding the Qwen profile to declare thinking support also emits that field; explicit `extra_body` can forward `chat_template_kwargs={"enable_thinking": false}`.

This PR establishes marimo's autocomplete policy and leaves provider translation to Pydantic AI. The custom-provider profile/translation gap needs upstream follow-up. No upstream issue has been filed, and no live vLLM verification was performed.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist

> Written by GPT-6 on Codex


marimo/_server/ai/providers.py
```diff
@@ -426,6 +427,9 @@ async def completion(
             result = await agent.run(
                 user_prompt=None,
                 message_history=VercelAIAdapter.load_messages(messages),
+                model_settings={"thinking": thinking}
+                if thinking is not None
+                else None,
```
kirangadhave, line 432: if we pass thinking = None, will the model use a default value and are we okay with that default?

kirangadhave (approved):

marimo/_server/ai/providers.py
Light2Dark: In this case, we are passing False from the endpoint. But yes, if someone passes None, we don't set anything which is sensible imo

mscolnick (approved):

PR #10913, opened 2026-09-18
Raise Ctrl-C only inside user code

manzt:
The interrupt tests have been flaky in CI because SIGINT can raise at any line of the kernel, not just user code.

For example, the handler sends the "interrupted" notification itself, so a Ctrl-C that lands while the kernel is already inside `pipe.send` deadlocks on the stream lock (MO-7697, investigated with Fable 5.1). A Ctrl-C inside a UI `on_change` handler kills the kernel, and one that lands between popping a cell and running it leaves the cell stuck in "running".

These changes introduce `InterruptScope`, which confines where Ctrl-C can raise:

```python
with InterruptScope() as interrupts:
    with interrupts.guard():       # raises MarimoInterrupt
        run_sync_user_code()
    await interrupts.run(coro())    # cancels the task
    finalize()                      # only sets interrupts.interrupted
```

The runner checks the flag before starting the next cell, guards each post-execution hook separately, and does not guard finalization. The notification is sent from the event loop instead of the handler.

The session test now interrupts while the stream lock is held, which hangs on the parent commit.

Closes MO-7697.


manzt:
meant to set as draft, but on my phone i can't find how to do that 😅

PR #10918, opened 2026-09-20
fix(convert): keep magic conversion when a magic has no argument

feiiiiii5:

## 📝 Summary

One argument-less Jupyter magic silently disables magic conversion for the
**whole notebook**, and `marimo convert` still exits 0.

`transform_magic_commands` splits a cell magic with `stripped.split("\n", 1)`
(`marimo/_convert/ipynb/to_ir.py:452`). After `cell.strip()`, a cell whose
entire contents are `%%time` has no newline, so the unpack raises
`ValueError: not enough values to unpack (expected 2, got 1)`. The single-line
path has the same shape: `%env` with no `VAR=VALUE` reaches
`source.split("=", 1)` and raises.

That exception is caught by `_run_transform`
(`marimo/_convert/ipynb/to_ir.py:1449-1459`), which logs
"`transform '…' failed; skipping this optimization`" and returns the sources
untouched — for **every** cell, not just the offending one. The cells then
reach the parser still carrying their magics and are emitted as
`_unparsable_cell`.

Reproduced on `main` at `9eb4fa44c55e86c532fe238edd23fb381c88c089` with the
documented CLI (`marimo convert nb.ipynb -o out.py`), notebook containing
`%matplotlib inline`, a bare `%%time`, and `%cd /tmp`:

```
ValueError: not enough values to unpack (expected 2, got 1)
  marimo/_convert/ipynb/to_ir.py, line 452, in transform
    magic, rest = stripped.split("\n", 1)
```

| cell | expected (what the same cell converts to on its own) | actual |
| --- | --- | --- |
| `%matplotlib inline` | `# '%matplotlib inline' command supported automatically in marimo` | `app._unparsable_cell(r"""%matplotlib inline …""")` |
| `%%time` | `# magic command not supported in marimo; …` | `app._unparsable_cell(r"""%%time""")` |
| `%cd /tmp` | `import os\nos.chdir('/tmp')` | `app._unparsable_cell(r"""%cd /tmp""")` |

`rc=0`, so a scripted conversion reports success while producing a notebook
whose cells are all unparsable.

Note `magic_remove` and `magic_already_supported` already guard `if source:`,
i.e. an empty body was expected here — only the split assumed a newline.

## Changes

- `transform`: `stripped.partition("\n")` so a bodyless cell magic yields an
  empty body instead of raising.
- `magic_cd`, `magic_mkdir`, `magic_env`: an argument-less invocation falls
  back to the existing "not supported in marimo" comment. Without this, the
  fix would trade the crash for code that cannot run — `os.chdir('')` and
  `os.makedirs('')` raise at first cell run — and `%env` would keep raising.
- `magic_remove`: drop the trailing space the argument-less case leaves in the
  generated comment, so the converted notebook has no trailing whitespace.

Three regression tests added; the third one is the blast-radius check (one
bare `%%time` must not cost the other cells their transformations).

## Test plan

- [x] `pytest tests/_convert/ipynb/test_ipynb_to_ir.py -k magic` — 3 new tests
      **fail on base** at `9eb4fa44c` with the `ValueError` above (run isolated
      outside the tree so `PYTHONPATH` alone selects the package: 3 failed,
      1 passed), **pass on this branch** (4 passed).
- [x] `pytest tests/_convert` — base `360 passed, 6 skipped, 1 xfailed`;
      branch `363 passed, 6 skipped, 1 xfailed` (the +3 are the new tests; no
      existing test changes result).
- [x] `ruff check` and `ruff format --check` clean on both changed files.
- [x] `marimo convert` on the notebook above now emits the transformed cells
      instead of `_unparsable_cell`.

## Pre-Review Checklist

- [ ] Not discussed in an issue — this is a small, self-contained correctness
      fix; happy to open one first if preferred.
- [ ] **Line-by-line human review by the account owner is still pending.** I
      am leaving this box unchecked rather than claiming a review that has not
      happened; that step is the reason this PR is a draft.
- [x] No visual changes.

## Merge Checklist

- [x] No public API change, no docs affected (docstrings unchanged).
- [x] Tests added for the change.

The CLA (`cla.yml`) still needs the account owner's own signature, and this PR
is a draft until a human has read the diff.


feiiiiii5:

PR #10919, opened 2026-09-20
fix(convert): keep non-string markdown frontmatter on import

feiiiiii5:

## Description

Markdown import drops every non-string YAML frontmatter value.

`ExpandAndClassifyProcessor.run` (`marimo/_convert/markdown/to_ir.py`) only copies frontmatter onto the XML tree when `isinstance(value, str)`, and `_tree_to_ir` then rebuilds the header from `root.attrib`. So lists, dicts, bools, ints and dates are silently discarded — while marimo's own exporter writes them: `pymdown.render_preamble` is typed `dict[str, str | list[str]]`. A notebook round-tripped through `.md` therefore loses metadata without any warning.

Measured at `9eb4fa44`, with `authors: [A, B]`, `date: 2026-09-20`, `draft: true`, `auto_download: [html, markdown]`:

```
### 2. IR -> md:   auto_download:\n- html\n- markdown      # exporter writes it
### 3. md -> IR:   app.options: {'app_title': 'My app', 'width': 'medium'}   header: None
### 5. header after import: 'description: A description\n'   # authors/date/draft gone
```

## Change

`_tree_to_ir` takes the parsed metadata it already had access to and passes it to a new `app_config_from_root(root, metadata)`, which re-attaches the keys that cannot be expressed as XML attributes (`config[config_keys.get(key, key)] = value`). String keys keep travelling through the existing attribute path, so nothing about today's output changes for them.

The metadata has to be bound in `MarimoMdParser.__init__` (`self.serializer = lambda root: _tree_to_ir(root, self.meta)`), because markdown 3.10.2 resolves `self.serializer` from `output_formats` inside `Markdown.__init__` — overriding `output_formats` on a subclass is a no-op there. I hit that first, which is why the binding looks indirect.

Scope note: this is the md→IR direction only. `_AppConfig.sanitize` (`marimo/_ast/app_config.py:46`) filters keys with `hasattr(_AppConfig, k)`, which is False for `field(default_factory=...)` keys such as `auto_download`, so that key is still invisible to `load_notebook_ir`. That is a separate root cause and I have left it out of this PR on purpose.

## Core

- [ ] This PR contains UI changes.
- [x] This PR includes new dependencies. — no, neither applies; it is a converter-only change.

## Evidence

The changed behaviour is covered by a new `test_markdown_non_string_frontmatter` in `tests/_convert/markdown/test_markdown_conversion.py`, and the existing snapshot `tests/_convert/markdown/snapshots/frontmatter-test.py.txt` is byte-identical after the fix, which is the guard that string-valued frontmatter did not change.

Base control — `marimo/_convert/markdown/to_ir.py` restored from `9eb4fa44c55e86c532fe238edd23fb381c88c089`, `git diff -- <that file>` printing nothing (so the run measures upstream code with the new test kept):

```
$ PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=. python -m pytest tests/_convert/markdown/test_markdown_conversion.py -q
1 failed, 25 passed, 1 skipped
FAILED ...::test_markdown_non_string_frontmatter
```

With the patch applied:

```
$ PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=. python -m pytest tests/_convert -q
361 passed, 6 skipped, 1 xfailed in 11.46s
```

Lint, using the ruff version `.pre-commit-config.yaml` pins (`rev: v0.16.6`):

```
$ ruff check marimo/_convert/markdown/to_ir.py tests/_convert/markdown/test_markdown_conversion.py
All checks passed!
$ ruff format --check <same two files>
2 files already formatted
```

I did not run `make py-check` end to-end (it also invokes mypy/ty and the frontend toolchain, which I do not have set up here), and mypy types on the two touched files are unchanged in shape: `metadata` is annotated `dict[str, Any]` and the new helper returns the same type `_tree_to_ir` always returned.

## Still needed from a human

- The CLA (`CONTRIBUTING.md`: "All new PRs will fail until you have signed the CLA") is an account-owner signature.
- Per `AGENTS.md` this PR is opened as a **draft**, and I have left the template box "Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it." **unchecked** — that attestation is the account owner's to make, not mine.


feiiiiii5:

PR #10923, opened 2026-09-20
fix(frontend): collapse all sections when nested heading ranges share an end cell

### LINKED ISSUE #10922: "Collapse all sections" silently does nothing when a ≥3-level heading chain ends a section (Node … not found in tree)
## Describe the bug

**Collapse all sections** (command palette, or `Cmd/Ctrl-Shift-\`) silently does nothing on
notebooks whose last subsection nests three or more heading levels. There is no toast and no
visible error — the exception only appears in the devtools console:

```
Error in reducer for action collapseAllCells: Error: Node bkHC not found in tree
    at e.collapseAll (http://localhost:2799/assets/cells-Cp-G34Rz.js:12:34354)
    ...
    at e.transformAll (http://localhost:2799/assets/cells-Cp-G34Rz.js:12:41372)
    at Object.collapseAllCells (http://localhost:2799/assets/cells-Cp-G34Rz.js:227:21538)
    at reducer (http://localhost:2799/assets/createReducer-B_XmHROl.js:1:231)
```

Collapsing the same sections one at a time with the chevron next to each heading works fine
— only the bulk command fails, and it fails for the whole notebook (all columns), not just
the offending section. Reproduced in Google Chrome 153 and headless Chromium 140, invoking
the command both from the command palette and via its hotkey.

### Root cause

All references are to tag `0.24.2`; the three files involved are byte-identical on `main`
as of 2026-09-20.

1. The `collapseAllCells` reducer
   (`frontend/src/core/cells/cells.ts:1356-1409`) walks cells bottom-up, calling
   `findCollapseRange` (`frontend/src/core/dom/outline.ts:143-175`) for each heading cell.
   When a new range's end coincides with an already-computed deeper range, it pulls the end
   back to that child's start (`cells.ts:1383-1390`):

   ```ts
   const parentEndInChild = rangeIndexes.find(
     (child) => child.start <= endIndex && child.end === endIndex,
   );
   if (parentEndInChild) {
     endIndex = parentEndInChild.start;
   }
   ```

   This adjustment is applied **once**, and `Array.find` returns the **innermost** matching
   range (ranges are pushed bottom-up, so the deepest one comes first). When *three or more*
   ranges share the same end index — an `h1` whose last `h2`'s last `h3` runs to the same
   boundary — the outermost range's `until` still points at a cell that belongs to a deeper
   range.

2. `CollapsibleTree.collapseAll` (`frontend/src/utils/id-tree.tsx:310`) then folds the
   ranges bottom-up. By the time the outer node collapses, its `until` cell has already been
   folded into a deeper node and is no longer top-level, so `nodes.findIndex` returns `-1`
   and it throws `Node <id> not found in tree` (`id-tree.tsx:343`).

3. The reducer wrapper (`frontend/src/utils/createReducer.ts:113`) catches the exception,
   logs it to the console only, and returns the previous state — so the command is a silent
   no-op for the entire notebook.

### When it triggers

The command fails whenever ≥3 collapse ranges share one end index, i.e. a nested chain of
≥3 heading levels sits at the tail of its enclosing section(s) — at the end of the notebook
(`# A` › last `## B` › `### C`) or at the end of any section mid-notebook. Verified
boundaries (each as a live notebook in marimo edit 0.24.2, one heading per cell):

| Heading structure (one cell each, `-` = no heading) | Result |
|---|---|
| `#` `##` `###` | **fails** (nothing collapses, console error) |
| `#` `##` | works |
| `#` `##A` `###` `##B` (deep chain not at tail) | works |
| `#` `-` `##` `-` `###` `-` (body cells between headings) | **fails** |
| `##` `###` `####` (no h1 needed) | **fails** |
| `#` `##A` `##B` `###` (chain at tail of last section) | **fails** |

### History

The one-shot adjustment shipped with the feature itself in #4799 (May 2025) and is
unchanged since, so every release with "Collapse all sections" is affected. The command
became much more discoverable with #10464 (Aug 2026, following #10459), which is probably
why it surfaces now.

### Suggested fix

Iterate the adjustment instead of applying it once: repeatedly pull `endIndex` back to the
**outermost** (smallest-start) already-computed range that ends at the current `endIndex`
and starts before it, until no such range exists:

```ts
for (;;) {
  const candidates = rangeIndexes.filter(
    (child) => child.start < endIndex && child.end === endIndex,
  );
  if (candidates.length === 0) break;
  endIndex = Math.min(...candidates.map((c) => c.start));
}
```

`endIndex` strictly decreases, so this terminates. Replaying the extracted 0.24.2 sources
(`findCollapseRange`, the reducer loop, `CollapsibleTree`) with this patch fixes every
failing structure above (and two real 84- and 100-cell notebooks) while leaving the passing
structures bit-identical. A regression test of three markdown cells `#`/`##`/`###` through
`collapseAllCells` would have caught this.

## Will you submit a PR?

Yes — a fix with regression tests is prepared and passing locally; PR incoming.

## Environment

(`marimo env` output; `location` field redacted)

```json
{
  "marimo": "0.24.2",
  "editable": false,
  "OS": "Darwin",
  "OS Version": "27.0.0",
  "Processor": "arm",
  "Python Version": "3.12.14",
  "Locale": "en_US",
  "Binaries": {
    "Browser": "--",
    "Node": "v26.8.2",
    "uv": "0.12.10 (3c979abda 2026-09-04 aarch64-apple-darwin)"
  },
  "Dependencies": {
    "click": "8.5.0",
    "docutils": "0.23",
    "itsdangerous": "2.2.0",
    "jedi": "0.19.2",
    "markdown": "3.10.3",
    "narwhals": "2.26.0",
    "packaging": "26.3",
    "psutil": "7.2.2",
    "pygments": "2.21.0",
    "pymdown-extensions": "11.0.2",
    "pyyaml": "6.0.3",
    "starlette": "1.6.0",
    "tomlkit": "0.15.1",
    "uvicorn": "0.52.4",
    "websockets": "17.1"
  },
  "Optional Dependencies": {
    "altair": "6.2.2",
    "anywidget": "0.11.0",
    "duckdb": "1.5.5",
    "loro": "1.13.2",
    "mcp": "2.2.0",
    "nbformat": "5.11.1",
    "openai": "3.9.0",
    "pandas": "3.0.5",
    "polars": "1.44.2",
    "pyarrow": "25.0.1",
    "pytest": "9.1.1",
    "python-lsp-ruff": "2.3.4",
    "python-lsp-server": "1.15.0",
    "ruff": "0.16.7",
    "sqlglot": "30.18.0",
    "vegafusion": "2.0.3"
  },
  "Experimental Flags": {}
}
```

Browser: Google Chrome 153.0.8010.53 (also reproduced in headless Chromium 140).

## Code to reproduce

Save as `repro.py` and open with `marimo edit repro.py` (or `marimo edit --sandbox repro.py`):

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "marimo",
# ]
# ///

import marimo

__generated_with = "0.24.2"


@app.cell
def _():
    import marimo as mo
    return (mo,)


@app.cell
def _(mo):
    mo.md(r"""# Title""")
    return


@app.cell
def _(mo):
    mo.md(r"""## Section""")
    return


@app.cell
def _(mo):
    mo.md(r"""### Sub""")
    return


if __name__ == "__main__":
    app.run()
```

Steps:
1. Wait for the three markdown cells to render.
2. Open the command palette (`Cmd/Ctrl-K`) and run **Collapse all sections** (or press
   `Cmd/Ctrl-Shift-\` directly).

**Expected:** the notebook folds — only the `# Title` cell remains visible, with the two
deeper sections nested under it. (This is what happens if you instead click the collapse
chevron next to each heading, bottom-up.)

**Actual:** nothing visibly happens. The devtools console shows
`Error in reducer for action collapseAllCells: Error: Node <cell-id> not found in tree`.

Removing the `### Sub` cell — or adding a trailing `## Another section` cell with no
subsections — makes the command work again (see the boundary table above).

cspipaon:
## 📝 Summary

Closes #10922

**Collapse all sections** was a silent no-op on any notebook where three or more nested
heading ranges end at the same cell — e.g. an `h1` whose last `h2`'s last `h3` all run to
the end of the notebook (or to the end of any enclosing section). The only trace was a
devtools console error: `Error in reducer for action collapseAllCells: Error: Node … not
found in tree`.

**Root cause:** `collapseAllCells` walks cells bottom-up and, when a parent range's end
falls inside an already-computed child range, pulls the end back to that child's start —
but only **once**, and `Array.find` returns the **innermost** matching child (ranges are
pushed bottom-up). With ≥3 ranges sharing an end, the outer range's `until` still points
at a cell that a deeper range swallows first, so `CollapsibleTree.collapseAll` throws
`Node … not found in tree`; the reducer wrapper catches it and returns the previous
state, so nothing collapses at all.

**Fix:** iterate the adjustment — repeatedly pull `endIndex` back to the start of the
*outermost* child range that ends at the current `endIndex`, until it no longer lands on
a child range's end. `endIndex` strictly decreases, so the loop terminates.

**Tests:**
- Regression test: the existing "can collapse and expand nested cells in one call"
  fixture already creates a `### Subsubheader` cell but never assigned it an outline —
  which is exactly why the suite missed this. The new test is that structure with the h3
  outline present; on `main` it fails (the throw is swallowed, so the collapse silently
  doesn't happen), with this fix it passes.
- Boundary test: a deep chain *followed by* a later section (`#`, `##A`, `###`, `##B`)
  still collapses with the correct nesting, guarding against over-collapse from the
  iteration.

`vitest run` on `cells.test.ts` + `id-tree.test.ts`: 261/261 passing. Analysis details
and a browser-level repro are in #10922.

## 📋 Pre-Review Checklist

## ✅ Merge Checklist


cspipaon:

mscolnick (approved):

